PR #1646, opened 2015-03-20
Delete log files only after five days

### LINKED ISSUE #1631: Archive prior set of logs on game restart
So this came up after the recent logging changes, and I think we already hit the first use case where it would really matter :-)

@minnesotags' 24/7 server crashed, he restarted it, all was well. Well .. except now we had no crash logs :D

So simply instead of deleting old logs on startup simple move any into a timestamped directory instead. No real need to only keep one set either - can just keep adding new directories. Very easy for users to script in deletion of dirs on a regular basis if they want to clean up stuff - no impact on the files or risk of something being in use.

msteiger:
This PR does two things:

* Log files are put into a timestamped sub-folder now.
* Only folders that are older than 5 days are deleted on game start.

Fixes #1631

### DIFF
--- a/engine/src/main/java/org/terasology/engine/LoggingContext.java
+++ b/engine/src/main/java/org/terasology/engine/LoggingContext.java
@@ -22,6 +22,9 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.text.ParseException;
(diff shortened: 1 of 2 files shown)


Cervator:
Tested log creation for client and server real quick, looks good. Haven't tried to fake my clock enough to see if expiration / deletion works, but hey :-)

That's probably another thing we should expose in a logging config somewhere in the menu. Maybe something both game and launcher could config, perhaps another entry or two in @shartte's mini config file?

PR #1648, opened 2015-03-21
Request game server info without actually joining

msteiger:
This PR adds a new feature to the JoinGameScreen: a connection is established to the selected server trying to request some information such as world time, selected modules, etc.

The largest beef was about dealing with failed connections and exceptions in a multi-threaded environment. It should all work now, though ..


Cervator:
Very nice - I especially like the module highlight which makes potential issues clear:

terasology _ pre alpha_2015-03-21_13-17-35

However, actually trying to use "Add New" to add localhost or a test entry didn't get any entries added for me :( No errors in the log.

minnesotags:
I'm not sure what that is.... I think I created that VM last December?

Cervator:
@minnesotags That's in-game time, hehe :-)

Several game days in a RL day :D

msteiger:
Ahh right, the list of local servers was not updated. This should be fixed now.

Iirc, one real wall-clock day equals 48 in-game days.

Cervator:
Retested and all good! Merged

Also: First official engine build done using the new dynamic droplets, woo! Done at the same time as three other test engine builds and a good dozen modules or so :D

PR #1650, opened 2015-03-21
Use localhost as default server entry for new config files

msteiger:
As mentioned / suggested by @Cervator


Cervator:
Thanks ! :-)

PR #1651, opened 2015-03-23
Tune-up for height-map based world generator

msteiger:
Assets can now be loaded in WorldViewer, which now allows for loading the HeightMap world:

image

Surprisingly, it is by far the slowest world generator of all. I'm not sure why the code was like it was, but it offered many possibilities for performance improvements.

This PR performs a few of them leading to a speed-up of about 20x-50x at equal quality. Moreover, it makes the surface provider independent of chunk size and uses a dynamic scale factor instead. I also lowered the sea level for additional land mass (see the screenshot). 

In the default setting, the terrain repeats after 2048 blocks, but biomes/vegetation/etc. is procedural and thus different.

### DIFF
--- a/modules/Core/src/main/java/org/terasology/core/world/generator/facetProviders/HeightMapSurfaceHeightProvider.java
+++ b/modules/Core/src/main/java/org/terasology/core/world/generator/facetProviders/HeightMapSurfaceHeightProvider.java
@@ -15,47 +15,39 @@
  */
 package org.terasology.core.world.generator.facetProviders;
 
-import com.google.common.collect.Sets;
-import com.google.common.math.IntMath;
+import java.nio.ByteBuffer;
+import java.nio.IntBuffer;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.asset.Assets;
-import org.terasology.math.ChunkMath;
-import org.terasology.math.Rect2i;
-import org.terasology.math.Region3i;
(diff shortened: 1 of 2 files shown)


Cervator:
Test this please

(Seeing if the PR builder thing works - might not since it is an existing PR)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/1/
Jenkins Success Message

msteiger:
Awesome!

Cervator:
Merged, thanks :-)

Didn't have much of a reference point for speed, been a long while since I've been running around in a height map world for any substantial period of time. But I flew around for a few kilometers and the world was keeping up fine.

One issue I noted - I think the change in sea level threw off the level at which grass begins:

terasology-150324205847-1152x720

If you can find blocks high enough then grass begins normally :-)

The world in general seemed changed scale-wise - it used to be very smooth and gradual about height, now it feels a lot sharper. The tiny stray islands seem to be weirdly contrasted although that could be from just never having seen them before. Didn't seem to start on the "mainland" I'm used to:

terasology-150324205721-1152x720

Still, it is clearly the height map world with its interesting tectonic features, which is the most important part :-)

msteiger:
Hmm .. that's interesting. You should spawn at the same point as before. Will look into that in the frame of reworking spawning positions.

The scale in height and x/z are configurable now. We could even add the asset name and the sea level to the set of parameter to allow for entirely different worlds.

PR #1662, opened 2015-03-31
Tune-Up for SubSampledNoise

msteiger:
This replaces calls to `size().x` and `size().y` in `SubSampledNoise` with `sizeX` and `sizeY` respectively.

This avoids the unneeded allocation of Vector2i instances. Since this was done very often, it quickly amortized causing intense GC cleanup cycles.

The effect can be observed e.g. in JVisualVM. The number of allocated Vector2i instance decreases:
image

### DIFF
--- a/engine/src/main/java/org/terasology/utilities/procedural/SubSampledNoise.java
+++ b/engine/src/main/java/org/terasology/utilities/procedural/SubSampledNoise.java
@@ -74,11 +74,11 @@ public class SubSampledNoise extends AbstractNoise {
     }
 
     private float[] getSubset(float[] fullData, Rect2i fullRegion, Rect2i subRegion) {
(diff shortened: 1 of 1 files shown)


Cervator:
Nice stats!

We get these every once in a while when somebody gets into a profiling tool. But even though we have free access to YourKit I'm not sure we have any good instructions anywhere for a quick setup to use said profiling tools to find things.

Up for throwing a few quick instructions in the wiki sometime, maybe we can link to it from the forum in some sticky thread somewhere? And do you have/want YourKit access?

(Also: it is totally possible we do have instructions somewhere and I'm just missing them. But it seems like a topic we could probably improve)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/15/

Cervator:
Feel free to merge this, by the way. Or can somebody else merge/review? My main PC is down and I may be mostly out of commission for a few more days

PR #1663, opened 2015-04-03
Switched to Java 8

immortius:
* Set JDK version to 1.8
* Fixed residual javadoc errors
* Updated to gestalt 2.3.0-RC1 and latest versions of javassist and reflectasm

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -214,7 +214,7 @@ idea {
 
     project {
         // Set JDK
-        jdkName = '1.7'
+        jdkName = '1.8'
         wildcards -= '!?*.groovy'
 
         ipr {
@@ -231,7 +231,7 @@ idea {
 
             // Sets sourceCompatibility within IntelliJ (without this root build having the Java plugin applied)
             whenMerged {project ->
-                project.jdk.languageLevel = 'JDK_1_7'
+                project.jdk.languageLevel = 'JDK_1_8'
             }
         }
     }
--- a/config/gradle/common.gradle
+++ b/config/gradle/common.gradle
@@ -8,6 +8,9 @@ apply plugin: 'checkstyle'
 apply plugin: 'pmd'
 apply plugin: 'findbugs'
 
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
 // We use both Maven Central and our own Artifactory instance, which contains module builds, extra libs, and so on
 repositories {
     mavenCentral()
--- a/engine-tests/src/test/java/org/terasology/utilities/WhiteNoiseTest.java
+++ b/engine-tests/src/test/java/org/terasology/utilities/WhiteNoiseTest.java
@@ -22,7 +22,7 @@ import org.terasology.utilities.procedural.WhiteNoise;
 import org.terasology.utilities.random.FastRandom;
 
 /**
- * A simple validity test for {@link FastNoise}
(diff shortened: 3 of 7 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/16/

**Build Log**
*last 5 lines*

 ```
[...truncated 3215 lines...]
[ANALYSIS-COLLECTOR] Computing warning deltas based on reference build #15
Setting status of 1813b15388760a036ed91e6e4d89c67a59bd92e3 to FAILURE with url http://jenkins.terasology.org/job/TerasologyPRs/16/ and message: Merged build finished.
```

**Comment by msteiger** [comment]
Test this please

**Comment by msteiger** [comment]
This build was run with Java7 on Jenkins. I've manually set the JDK version to 8_40, but it would be good if the build would explicitly define the required Java version in the root `build.gradle` like so:

```gradle  
sourceCompatibility = 1.8
targetCompatibility = 1.8
```

[targetComp. defaults to sourceComp](https://gradle.org/docs/current/dsl/org.gradle.api.plugins.JavaPluginConvention.html), so it's not really required.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/18/

Cervator:
Yeah we need to manually update everything to use the Java 8 configured in Jenkins. Pretty sure there's a simple Groovy example on how to do that in bulk though, so we should be able to run that before merging this then rebuilding all modules etc.

I'll try to prep a stable build tonight or tomorrow then we can merge this into develop :-)

msteiger:
I would assume that even if engine is compiled with Java 8, modules can still be loaded even if they are Java 7. 

I think that @shartte had a small snippet for the Jenkins console that would set the JDK to a specific version. Then you would just need to iterator over all job configs.

Cervator:
@msteiger yeah I think the JDK swap Groovy snippet will be easy enough and indeed the modules should keep working :-)

I adjusted the engine job for the develop branch (plus the applet builder) and everything looks good. Will sort out the rest of Jenkins soonish and hopefully get a stable release post up tomorrow (ended up skipping a couple numbers due to failures - another good reason to decouple engine version from stable release)

Thanks both for all the Java 8 prep. We are officially there! :D

msteiger:
I tried to build PolyWorld on Jenkins yesterday, but it failed. I assume that most modules are failing and this is why:

Setting `sourceCompatibility = 1.8` in `common.gradle` is picked up by Core's build script, which is also used to compile the other modules. This *should* be ok, but since the publish task depends on javadoc, this task is run before. Unfortunately, Oracle switched it into paranoid mode and the build fails because of that.

I figure that we could either
* somehow convert most javadoc errors back to warnings ( I assume that `doclint:none` turns it off entirely? )
* continue the build even if javadoc fails
* fix Javadoc for all modules now ( too much work I guess )
* set Java8 compatibility for engine and PC facade only (made even set it in the root build.gradle?)

Cervator:
I figured that would happen, but didn't think too much of it, imagining few modules have a lot of Javadoc to update. With a stable release just out we can probably take a couple weeks to slowly fix it all?

After all we want to eventually straighten out those sorts of things, so might as well get it over with :-) Maybe we should've enabled Javadoc nagging via Checkstyle earlier to catch a bunch of them before Java 8. But oh well.

PR #1665, opened 2015-04-07
Refactor ModuleManager

msteiger:
This PR extracts an interface out of the `ModuleManager` class. It also separates the querying of module metadata, which is independent from the actual implementation of `ModuleManager`.

As a consequence, the module manager can no longer directly override the retrieval of this metadata. However, it can customize the loading process and thus retain its influence.

Only the first 4 lines in the constructor of `ModuleManagerImpl` were changed. GitHub gets confused with the renaming/modification.

### DIFF
--- a/engine-tests/src/main/java/org/terasology/testUtil/ModuleManagerFactory.java
+++ b/engine-tests/src/main/java/org/terasology/testUtil/ModuleManagerFactory.java
@@ -19,6 +19,7 @@ import com.google.common.collect.Sets;
 
 import org.terasology.engine.TerasologyConstants;
 import org.terasology.engine.module.ModuleManager;
+import org.terasology.engine.module.ModuleManagerImpl;
 import org.terasology.module.ClasspathModule;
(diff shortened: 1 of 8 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/19/

Cervator:
Bit outside my area of familiarity, what's the ultimate purpose / reasoning? Something related to WorldViewer or something? Matching some tweaks to gestalt-module somehow? Preparation for an eventual external module tool?

/me pokes @immortius if interested

msteiger:
Yeah, I want to be able to load modules at runtime in WorldViewer, too. However, it needs to be done a little bit different. There is no module folder, for example.

Using interfaces instead of concrete classes gives a lot of flexibility in general, in particular if they are used as some kind of service in many different places (through `CoreRegistry`). Refactoring the access to module metadata could be seen as a general code cleanup process.

There are no functional changes.

PR #1666, opened 2015-04-11
Added a StorageQualifier parameter to the Material methods where it is r...

LinusVanElswijk:
**Summary:**
This PR adds support for attribute parameters in shaders/materials. This is achieved by adding a storage qualifier parameter to the Material interface methods.

As a consequence of this, all `Material.set*(...)` calls have been updated. Previously, these calls could be assumed to set uniforms. This assumption now has to be made explicit using the new parameter.

**Motivation:**
Attributes are required for instanced rendering of particles.
Attributes are useful in general where similar objects are rendered (in batch) using the same shader, but with different input values for each object.


GooeyHub:

LinusVanElswijk:
Closed; per vertex data belongs in Meshes, not Materials

PR #1670, opened 2015-04-16
Fixing issue with being unable to start Terasology on new downloads due ...

MarcinSc:
...to module version problems.

### DIFF
--- a/modules/CoreSampleGameplay/module.txt
+++ b/modules/CoreSampleGameplay/module.txt
@@ -5,7 +5,7 @@
     "description" : "This module introduces the core Terasology content",
     "dependencies" :
     [
-        {"id": "Core", "minVersion": "0.1.0"}
+        {"id": "Core", "minVersion": "0.1.0", "maxVersion": "1.0.0"}
     ],
     "isGameplay" : "true",
     "defaultWorldGenerator" : "Core:FacetedPerlin"


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/21/

immortius:
It might be better to keep Core's version in line with engine's, rather
than make a given version of Core compatible with such a wide range.

Cervator:
Yeah this one is sort of another one of the surprises that keeps coming with not being entirely on top of formal release management :-)

In this case Core and CoreSampleGameplay are special in that they're always bundled with the engine anyway, so the current half-done release management clashes with the practice we've picked to let *minor* be the breaking version limiter by default while < 1.0.0

All other modules should be able to follow that fine, in this case I should indeed probably just start incrementing the min version for CoreSampleGameplay's dependency on Core each stable release - you'll always get a new one anyway.

I can straighten that out a bit more by next stable or maybe when we sort out Iota/Omega more (so we need the Core modules less)

Cervator:
So on second look this actually gets a little trickier.

Fundamentally it is a release management issue because we're not using SemVer correctly yet - not that it actually mandates that `minor` can be used as `major` in pre-release. But we're going by that and that's a good practice IMHO.

However, I'm incrementing `minor` on engine regardless, which is bad, and needs to stop in favor of SemVer correct bumps when needed + separate release version handling for "stable releases"

Trick is - even if we start doing that correctly it'll result in `minor` releases at appropriate times and then all the modules with an affected dependency will stop being accepted by the game until they have their versioning adjusted (and maybe code if there was an actual functionality break - but I expect that effect to be rare)

That'll be a bit of a pain, especially until the automated releasing in Jenkins is ready. Neither making Gestalt more lenient or using max version 1.0.0 is a good solution (though either would work). But how about we make the game simply issue a warning if a dependency is present but too new? Rather than block the use of the module.

I'll re-summarize this over at http://forum.terasology.org/threads/more-version-increments-to-avoid-dependency-issues.1270 - I think the game leniency is a good option since it gives us space to use Gestalt and SemVer correctly, without adding needless overhead. And it meshes with some other ideas there like post-release analytics that can hint at how likely issues are with using too-new dependencies.

PR #1672, opened 2015-04-18
LwjglRenderingProcess no longer a singleton, Dimension instances used more extensively

emanuele3d:
The [first commit](https://github.com/MovingBlocks/Terasology/commit/47ec5d348448357c9dc59bce418aa738882576d8) is an easy and hopefully uncontroversial one:
* it takes advantage of FBO.Dimensions instances throughout to simplify FBO generation code and dimensions handling, making more explicit the sizes of the FBOs involved and any change to them.
* it also add a couple of small method wrapping unsightly glViewport() calls into setViewportToFullScale() and setViewportTo().

The [second commit](https://github.com/MovingBlocks/Terasology/commit/09a6c9203d256a0e7d212c0171bd238e190166ed) is only a little more involved because it changes the LwjglRenderingProcess class from a singleton obtainable via its getInstance() method to an object instantiated by the WorldRenderer upon construction, and then made available throughout the application via the CoreRegistry. 

This allows for a more explicit control of the life cycle of the instance. Previously the instance was implicitely created by ShaderParameters instances within ShaderManager.initShaders() during Display initialization, well before the rendering process (or the shaders for that matter) were actually needed. Shaders are now initialized just after the rendering process is initialized, during WorldRenderer construction.

This second commit deals, perhaps a bit bluntly, also with shader initialization in headless mode. The headless version of the ShaderManager was used to call its initShaders() method a number times in headless. initShaders() is however an empty method and I removed calls to it. In hindsight, as I write this, this was perhaps unnecessary and maybe even wrong. An headless mode smoke test didn't reveal any problem with the changes though. And what do shaders do in headless mode anyway?

Finally, as a reference to where I am heading, you can take advantage of the last commit on [this branch](https://github.com/emanuele3d/Terasology/tree/RendererChangesBKP).

### DIFF
--- a/engine-tests/src/main/java/org/terasology/DisplayEnvironment.java
+++ b/engine-tests/src/main/java/org/terasology/DisplayEnvironment.java
@@ -155,9 +155,6 @@ public class DisplayEnvironment extends HeadlessEnvironment {
                 return new UISkin(uri, data);
             }
         });
-
-        // TODO: move somewhere else
-        CoreRegistry.put(ShaderManager.class, new ShaderManagerLwjgl()).initShaders();
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/editor/properties/SceneProperties.java
+++ b/engine/src/main/java/org/terasology/editor/properties/SceneProperties.java
@@ -38,9 +38,9 @@ public class SceneProperties implements PropertyProvider {
         if (backdropRenderer != null) {
             result.addAll(new ReflectionProvider(backdropRenderer).getProperties());
         }
-        LwjglRenderingProcess postRenderer = LwjglRenderingProcess.getInstance();
-        if (postRenderer != null) {
-            result.addAll(new ReflectionProvider(postRenderer).getProperties());
+        LwjglRenderingProcess renderingProcess = CoreRegistry.get(LwjglRenderingProcess.class);
+        if (renderingProcess != null) {
+            result.addAll(new ReflectionProvider(renderingProcess).getProperties());
         }
         return result;
     }
--- a/engine/src/main/java/org/terasology/engine/modes/loadProcesses/InitialiseGraphics.java
+++ b/engine/src/main/java/org/terasology/engine/modes/loadProcesses/InitialiseGraphics.java
@@ -37,8 +37,6 @@ public class InitialiseGraphics extends SingleStepLoadProcess {
 
     @Override
     public boolean step() {
-        CoreRegistry.get(ShaderManager.class).initShaders();
-
         NUIManager nuiManager = CoreRegistry.get(NUIManager.class);
         ((NUIManagerInternal) nuiManager).refreshWidgetsLibrary();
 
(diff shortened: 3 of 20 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/24/

msteiger:
I find the change of `LwjglRenderingProcess` towards a registered service and away from a singleton instance quite valuable.

Would it make sense to extract an interface `RenderingProcess` that can be used by callers instead? I reckon that it would enable the use of different rendering systems (in particular LWJGL 3, which is supposedly quite different from v2).

emanuele3d:
@msteiger, in short: I haven't looked that far.

A large number of methods will be moved into two new classes, a [GraphicState](https://github.com/emanuele3d/Terasology/blob/RendererChangesBKP/engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java), taking care of all opengl-specific state changes (mostly used in the first part of the rendering pipeline), and a [PostProcessor](https://github.com/emanuele3d/Terasology/blob/RendererChangesBKP/engine/src/main/java/org/terasology/rendering/opengl/PostProcessor.java), taking care of, guess what, post-processing and eventually generating the final image shown on screen. 

The current LwjglRenderingProcess will be renamed to [BuffersManager](https://github.com/emanuele3d/Terasology/blob/RendererChangesBKP/engine/src/main/java/org/terasology/rendering/opengl/BuffersManager.java) as that's almost all the functionality it will be left with: initializing, storing, providing and disposing of FBOs. It will also retain the functionality to take screenshots as that's based on reading the content of an FBO and writing it into a file. 

The already existing [WorldRendererImpl](https://github.com/emanuele3d/Terasology/blob/RendererChangesBKP/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java) class will become the hub of all this, orchestrating the construction, initialization and disposal of these support classes and taking advantage of them to indirectly generate the final image (see its render() method, the idea is that it should read like a complete index of what the renderer does every frame). 

In this context, switching the renderer to LWJGL 3 would be as "simple" as updating the innards of some of these classes to be LWJGL 3-based, maintaining their public methods as they are. In this scenario alone interfaces perhaps might not be strictly necessary, as the only customer of these classes is the WorldRendererImpl class which is already accessible through an interface.

I imagine interfaces might be needed if we wanted modules to have some deeper access to the rendering engine. But I'm not sure that would make sense in the current architecture: the renderer goes through an hardcoded pipeline that accepts a number of predefined inputs and produces a final image. Right now I can't think why a module would want to interact with it, as it stands. 

It would definitely make sense however if the pipeline could be modified at runtime, i.e. by injecting rendered content (i.e. layers) or adding/removing post-processes (i.e. flters) . Sometime ago I hypothesized a node-based architecture for this but it'd be quite a bit of work. Perhaps it will be necessary to move in that direction in the context of Cameras as components. As I eventually begin work on the new Skysphere, and thinking also about the possibility of portals and "security cameras", it is likely it will become necessary to address these issues.

Sorry if I've been too verbose. Let me know what you think, and if you really think there are uses for interfaces for these classes I'll make them available.

Cervator:
I agree with @GooeyHub - All is well + Jenkins likes the taste of this PR. Works fine locally too :-)

Merge when happy, I figure I'll leave it be for now in case of more discussion.

msteiger:
Sounds great. Looking forward to this!

PR #1681, opened 2015-04-21
Use code metric config files from Artifactory

msteiger:
This PR removes the configuration files for the code metrics tools PMD, FindBugs and CheckStyle.
Instead, gradle's new `TextResource` class is used to resolve and read the files from Artifactory.

This allows for reusing the config files across different projects, versioning and easier maintenance.

**EDIT**: this PR also updates the gradle wrapper to version 2.3.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/32/

**Build Log**
*last 5 lines*

 ```
[...truncated 2020 lines...]
	at hudson.model.Executor.run(Executor.java:240)
[ANALYSIS-COLLECTOR] Skipping publisher since build result is FAILURE
Setting status of 97df66b215b023e5354bfe35da235ae49173bf81 to FAILURE with url http://jenkins.terasology.org/job/TerasologyPRs/32/ and message: Merged build finished.
```

**Comment by GooeyHub** [comment]
http://jenkins.terasology.org/job/TerasologyPRs/34/

**Comment by Cervator** [comment]
After spending a couple hours looking at this more closely I'm conflicted. The nerd in me wants to gleefully embrace the shiny new text resource, but its use seems a bit quirky. 

Extracting the config file into each (sub)project's build dir, then cheating about the timing for that so a clean wouldn't hit before it gets used ... eep. Quirky Gradle looks quirky. Works though, and is still neat.

It also makes it trickier (though probably not unsolvable) to configure IDE analytics. Does it work like this in Eclipse without manual config? In IntelliJ for instance config/gradle/ide.gradle customizes the Checkstyle config to use the config/checkstyle stuff, which I guess now would be missing (did you not delete the actual checkstyle.xml there for some reason?)

I puzzled at it for a bit and realized there might be a far less geeky and way more boring way to do it. Here's a quick (well, okay, not so quick, am apparently not on top of my Gradle-fu tonight, this took way too long and could be done better) proof of concept:

https://github.com/MovingBlocks/Terasology/commit/b59211a208f1d7eed70c95389582be5c7056d59e

That task isn't tied to anything yet, and I don't know if it would hit similar timing issues. Could we add it as a dependency to the idea task and the code analytics config tasks? Would it need a doFirst? Etc

It is based off the extractNatives task, which admittedly is only needed at game runtime, so this might not work after all :/

In a similar vein I noticed Gradle 2.2 adds support for doing something like idea.project.cvs = "Git" instead of the hack we use in ide.gradle, but since that's a project level property it doesn't work for our nested Git roots so can't use that either :(

Going to take a break from the Gradle for now, let me know what you think. Maybe we can just copy via text resource into a single specific directory and configure to hit that both in Gradle land and via IDE? Rather than goofy unzip of dependency. Not sure either option is "better" - one is just geekier and new, me want :-)

In other news: It takes approximately 17 minutes to run "gradlew check" with near 70 modules in source form locally.

knowing

**Comment by msteiger** [comment]
I really like your solution - I actually prefer it over mine, since it is consistent with the previous file structure and thus still supports IDE integration nicely.

A [similar question was asked in the Gradle forums](https://discuss.gradle.org/t/right-way-to-copy-contents-from-dependency-archives/7449/13?u=msteiger) with the following answer:

```gradle
task extractApi(type: Copy) {
    from { // use of closure defers evaluation until execution time
        configurations.api.collect { zipTree(it) }
    }
    into "$buildDir/api/"
}
```

I suggest making the checkStyle* tasks depend on this copy task. Not sure how to do it for all source sets. My solution sets it manually for `main` and `test` and once more in engine for `dev`.

Cervator:
Ah, so that's the right syntax! I was hunting for a clean collect/ziptree version but just couldn't find it last night. Thanks for digging that up.

I can try to hook it up tonight and see how IntelliJ and such deals with it.

Issue #1379, opened 2014-11-20
Allow world generators to define spawnable locations/areas


msteiger:
Currently, the player spawns very close to the world's origin (center of chunk 0/0 I believe). I'd like to have this more customizable. Two examples:

* In PolyWorld, players should start at the island shore rather than 10 miles out in the sea.
* In Cities, player should spawn in a town/settlement to get started quickly.

I guess this could be done with `SpawnableAreaFacet` and/or `SpawnableLocationFacet` and engine picks the closest match with respect to the last-known player location. For worlds without that facets, the default (current) spawning procedure is performed.

Any other suggenstions/ideas/thoughts?

msteiger:
I think that the world generator needs some (facet-based) meta information to find good spawning points. This information is generated per (chunk) region and unfortunately not available after the chunk is rasterized.

I envision spawning point providers that write to a facet that is collected at the end of the chunk generation process. The world gen. would store them in a `Map<Chunk, Collection<Vector3i>>`. Then, this map could be queried by a method

    Vector3i getClosestSpawningPoint(int worldX, int worldZ);

which would iterate in a spiral around the target region until a valid spawning point is found. In a last step, the worldProvider must be checked for air blocks at the spawning location. 

In the simplest case, `getClosestSP()` would just return the queried location (at the correct height).

Maybe one spawning point per chunk is enough? Then, `Optional` could be used instead of Collection.

What do you guys think?

Cervator:
Sounds fine to me but then I'm not picky :-)

I've been wondering how to update the Spawning module to find suitable spawning sites for creatures with various demands, but I suppose that's entirely different (has the luxury of running on a timer after chunks have loaded)

Or alternatively should we consider a pre-spawning game state (think scoreboard in typical FPSes, or even optionally in an artificial lobby if the gameplay warrants that) so chunk gen can run for a while and make it easier to detect good spawning sites for either players or creatures?

Although it seems like there are two topics here

1. Finding good world scope "home" location for certain actions, like spawning the player. PolyWorld could go "Generate partial world map, then find suitable land mass, start game there, generate chunks, place player
1. Determine suitable block locations within generated chunks fitting a particular context. Could be the Spawning module, even for players.

Would that split approach help lay the groundwork for supporting gameplay with custom start settings? Like Polyworld, a large world possibly needing a start spot far from the default world center, or L&S where the player needs to spawn in a suitable village, maybe even with clickable options on what creature to spawn as

[...]

PR #1683, opened 2015-04-21
Support different height maps

msteiger:
This PR adds an alternative height map for HeightMapWorldGenerator and introduces different WrapModes that define the area outside the heightmap. Here's an example based on the new "Opposing Islands" map with WrapMode "CLAMP" (border pixels are repeated, creating an infinite ocean).

image

Note that you will spawn in the ocean on that map. Configuring spawning positions is still an open issue (#1379).


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/35/

Cervator:
Oooohhh that other noise type image of the opposing islands .... so pretty!

Merged in, reminded self of https://github.com/MovingBlocks/Terasology/pull/1651#issuecomment-85768901 which is still a thing (grass blocks start at a wrong level). We should probably throw that in a quick issue unless fixing soon so we don't forget again :-)

msteiger:
That should be fixed already. I found a piece of grass on the beach just one level above the water surface:

image

Maybe the grass density could be increased for the `BEACH` biome.

Cervator:
Oh that's a grass *billboard* they seem fine - look at the dirt blocks in the background. Those should likely be grass *blocks* which becomes very apparent in some "fields of dirt" you might normally expect to be plains :-)

PR #1686, opened 2015-04-24
Alternative approach to #1681

Cervator:
Instead of using the shiny new nerdy approach with text resource (geeky sadface) we at least can use the zip to extract and place the config files in the workspace.

Pro: No changes needed in Jenkins
Con: Didn't spot an easy way to guarantee the new extractConfig task has run if the user goes straight to running an individual analytic task instead of the top level "gradlew check"

I tried to attach extractConfig to the [analytic]Main tasks in common.gradle so it would correctly trigger in any case but this dramatically slowed down the configuration phase for Gradle

Possible follow-up to make the smaller projects load via zip instead of getting the files via engine harness in Jenkins and maybe try some other approaches to making it generic without making Gradle sad


config/gradle/common.gradle
```diff
@@ -25,6 +25,13 @@ repositories {
     }
 }
 
+dependencies {
```
Cervator: @msteiger - curiously where'd this come from? It looks unusual. I guess it sets the version to use but the format looks different?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/41/

Cervator:
Still not totally happy with this, but it works for now assuming it is exceedingly rare somebody will grab the project and immediately proceed to run gradlew :modules:Core:checkstyleMain without preparing for an IDE or otherwise triggering extractConfig

Leaving a todo for later to see if we can tweak it a bit more

@msteiger: Want to copy this approach to the other projects able to use the zip?

config/gradle/common.gradle
msteiger: Each of the plugins has its own `configuration` that can be used to explicitly define the artifact that is used to run the check.
The term in the bracket is a shortened  expression for `(group: <X>, name: <Y>, version:<1.0>)`.

engine/build.gradle
```diff
@@ -244,3 +244,8 @@ idea {
         downloadSources = true
     }
 }
+
+// Make sure our config file for code analytics get extracted (vulnerability: non-IDE execution of single analytic)
+ideaModule.dependsOn rootProject.extractConfig
+tasks.eclipse.dependsOn rootProject.extractConfig
+check.dependsOn rootProject.extractConfig
```
msteiger: I believe that this could fail randomly, since it does not guarantee that `extractConfig` is run before the checkstyle tasks. This only ensures that it's run before the (no-op) task `check`.
I assume that the solution that was posted on the gradle forums is the very slow one?

msteiger: Maybe the reason for the slowdown is that the extraction of file from the zip archive is done in the configuration phase?

In any case, it might be safer to make the root projects `checkstyleMain` and `checkstyleTest` tasks depend on the extraction. This should be fast and reliable, but fails if someone cleans and runs checkstyle on engine only. This seems to be a less serious issue to me.

Cervator: Yep, it absolutely could fail, that's what I dislike about it :D

I wonder if me adding an if caused the slowdown. I had checkstyleMain etc trigger extractConfig in common.gradle, but *only* if it wasn't done in the root project - that way module builds wouldn't do it (they'd get the config files via engine harness)

If they tried to call exactConfig it would fail since they wouldn't have that task (they'd start with Core's build.gradle). Although ... maybe I could put the task itself into common.gradle.

I figured even with the if the task would only get added to the execution once, but it slowed each module tasks config step by about 5 seconds. At that point the zipping shouldn't be happening yet so I wouldn't think that would be the slowdown.

In any case I'll try tweaking at it more later, and yeah at least using [analytic]Main instead of check would capture the three analytics independently rather than just the global version (but still not at the per-module level)

PR #1692, opened 2015-04-25
Fix block plaement near player.

flo:
Without this patch it is not possible to place blocks in certain
situations:
1. You couldn't place a block at character height if you walked towards
another path blocking block that is below or above the placement position.
e.g. place a block infront of you, walk towards it till limit.
Try placing a block on top of it.

2. Blocks couldn't be placed in a neighbour block that is 1 lower than
the current character position even when the character wasn't lowered
already from the edge (and capsule collision shape).

If in doubt the block placement will now be allowed which can lead in
certain corner situations to blocks that collide with the character.
e.g. when you walk very far on the edge and then plae a block below you.
You will then sink in the block till you hit the next block below it so
it is no big deal.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
+++ b/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
@@ -63,7 +63,7 @@ public class KinematicCharacterMover implements CharacterMover {
     /**
      * The amount of horizontal penetration to allow.
      */
-    private static final float HORIZONTAL_PENETRATION = 0.03f;
+    public static final float HORIZONTAL_PENETRATION = 0.03f;
 
     /**
      * The amount of extra distance added to horizontal movement to allow for penetration.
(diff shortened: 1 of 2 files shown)


Cervator:
Curious question: any particular interesting notes on this topic in multiplayer? IIRC if something sinks into a block lighting turns off - would a monkeyhead in that situation go black if a block is placed funny?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/51/

flo:
No idea, haven't tested it in multiplayer as it wasn't a multiplayer related change.

Cervator:
Okies, I'll check it out. Not exactly a big deal even if it does turn off the lights in a rare situation, we got bigger issues to fix :-)

Cervator:
Tested it out in multiplayer and yeah I can sort of place a block on myself and get stuck inside it, where indeed your model turns all-black when viewed by other players. But it is kinda tricky to pull off - most the time no problem.

I think you're right and that the usability benefit outweighs the rarely occurring quirk. We can worry about that come Beta one day ... :-) I'll leave a reminder around for later so we don't rediscover it in a year and get all confused about how that happened, heh.

PR #1691, opened 2015-04-25
Introduce initialize() method for FacetProvider

msteiger:
I would actually prefer to move the empty method implementations into a new class `AbstractFacetProvider` or `FacetProviderAdapter`, but this would require changing dozends of files. 

Also, default methods were introduced specifically for *interface evolution*, so this is probably a good fit - it just feels strange.

PR #1706, opened 2015-05-01
New spawning system

msteiger:
This PR reworks the `PlayerSystem` class quite a bit, mostly cleaning up legacy code. The *desired* spawning position is queried from a new interface `Spawner`.

Before an entity is actually created, air blocks at and around the desired spawning position are searched to ensure that there's enough space.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/60/

engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
```diff
@@ -70,13 +71,19 @@
 @RegisterSystem(RegisterMode.AUTHORITY)
 public class PlayerSystem extends BaseComponentSystem implements UpdateSubscriberSystem {
 
+    private static final Logger logger = LoggerFactory.getLogger(PlayerSystem.class);
+
+    private final float playerHeight = 1.5f;
```
Cervator: I see this got moved from spawnPlayer - any chance we can just load it from the player prefab?

Cervator:
Looks good. Tested around a bit both in single player and multiplayer :-)

I wonder if we could eventually merge this into a cleaned up Spawning module, when Iota is more in place (minimal runtime dependencies of the PC facade). Looks like the two narrowly avoid clashing in naming, so all is well. Spawning and LightAndShadow outright clash, both have a SpawnerSystem.

Way out of scope for this PR though! But curious what you think about the player prefab bit.

Cervator:
On testing it more closely with other worlds I realized they now do not implement WorldGenerator fully, and there was a bunch that would end up with boilerplate code. Java 8 to the rescue! I changed it to a default interface implementation. I feel so fancy now.

Took inspiration from https://github.com/MovingBlocks/Terasology/pull/1691/files - hopefully I did it right! https://github.com/MovingBlocks/Terasology/commit/3a2dad2f79c6a2445be9fdbe733b181d01f53c85

Even added javadoc! 

I put in a separate follow-up for later on the player height thing

Thanks as usual :-)

PR #1717, opened 2015-05-05
Two small changes

MarcinSc:
1. New component and system that allows for easy and powerful module configuration (for "modpack" developers). Decided to put it in engine, to have just one way of doing it, rather than 1 per module/module-author.
2. Preparation for optional modules - make sure that if a system is being created and it depends on a module that is optional, Terasology still starts instead of crashes with NoClassDefFoundError and just logs an error. Probably has to stay being an error, since it might be due to some other problems. Maybe just a log error message should be improved?

### DIFF
--- a/engine/src/main/java/org/terasology/logic/config/ModuleConfigSystem.java
+++ b/engine/src/main/java/org/terasology/logic/config/ModuleConfigSystem.java
@@ -71,6 +71,11 @@ public class ModuleConfigSystem extends BaseComponentSystem implements ModuleCon
         return getVariable(moduleName, propertyName, Float::parseFloat, defaultValue);
     }
 
+    @Override
+    public boolean getBooleanVariable(String moduleName, String propertyName, boolean defaultValue) {
+        return getVariable(moduleName, propertyName, Boolean::parseBoolean, defaultValue);
+    }
+
     private <T> T getVariable(String moduleName, String propertyName, Function<String, T> extractFunction, T defaultValue) {
         Map<String, String> moduleProperties = propertiesPerModule.get(moduleName);
         if (moduleProperties == null) {
--- a/engine/src/main/java/org/terasology/logic/config/ModuleConfigSystem.java
+++ b/engine/src/main/java/org/terasology/logic/config/ModuleConfigSystem.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2015 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/66/

Cervator:
How closely tied is this to https://github.com/MovingBlocks/gestalt/pull/18 ? Should we wait for it to settle and get released first?

MarcinSc:
Don't really have to, until the optional stuff is handled by gestalt, it is just going to be a required dependency.

engine/src/main/java/org/terasology/logic/config/ModuleConfigSystem.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(ModuleConfigSystem.class);
+
+    @In
+    private EntityManager entityManager;
+
+    private Map<String, Map<String, String>> propertiesPerModule = new HashMap<>();
+
+    @Override
+    public void preBegin() {
+        for (EntityRef entityRef : entityManager.getEntitiesWith(ModuleConfigComponent.class)) {
```
Josharias: As this only happens once and doesnt get any changes that an entity gets,  would it be better suited to find all prefabs with ModuleConfigComponent?  
Now that I have tried it, how do you create the entities at all before preBegin? I tried the AutoCreateComponent,  but that didnt work.  I must be missing something.
Using prefabs instead of entities made it work well for my little test.  Could be a good way going forward.

Josharias:
This has much more potential than the config idea I had a ways back.  I like it.

engine/src/main/java/org/terasology/logic/config/ModuleConfigSystem.java
MarcinSc: Ah, yes, I meant looking for prefabs, used the wrong thing...

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/67/

MarcinSc:
I've updated it to use PrefabManager, rather than EntityManager.

Cervator:
Thanks for code & thanks for merge :-)

Cervator:
Bump - ran back across this during change-log write-up for the release, should we have this more highlighted somewhere, like with a tutorial and some examples? It sounds to be worth more than a few notes in a PR

For instance does this help us get closer to configuring hot keys specific to modules? What's next on the quest to better support module conflicts and better extensibility?

@MarcinSc @Josharias @msteiger - ping!

Josharias:
I am using it in EventualSkills.
Config:
https://github.com/Terasology/EventualSkills/blob/master/assets/prefabs/Config.prefab
Usage in code:
https://github.com/Terasology/EventualSkills/blob/master/src/main/java/org/terasology/eventualSkills/systems/EventualSkillsCommonSystem.java#L117

Alas, there is no grimoire for using this feature...

Issue #71, opened 2012-01-28
GroovyConsole usability


Cervator:
We have a console in-game able to interact with game state via Groovy - which has immense potential, but isn't yet very user friendly! It might be nice to have a better text area to type into (multi-line, etc) and some convenience commands like "help" and a list of common commands

PR #1721, opened 2015-05-10
Introduce the concept of a Context to slowly get rid of CoreRegistry.

flo:
Reasoning: The core registry is just like static variables and it prevents
cool stuff like having a client and headless server running in the same VM.

### DIFF
--- a/engine-tests/src/main/java/org/terasology/Environment.java
+++ b/engine-tests/src/main/java/org/terasology/Environment.java
@@ -16,11 +16,13 @@
 
 package org.terasology;
 
+import com.google.common.collect.Sets;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.terasology.context.Context;
+import org.terasology.context.internal.ContextImpl;
 import org.terasology.naming.Name;
-
-import com.google.common.collect.Sets;
+import org.terasology.registry.CoreRegistry;
 
 import java.io.IOException;
 import java.util.Set;
@@ -34,12 +36,13 @@ public class Environment {
 
     private static final Logger logger = LoggerFactory.getLogger(Environment.class);
 
+    protected Context context;
+
     /**
      * Default setup order
      * @param moduleNames a list of module names
      */
     public Environment(Name ... moduleNames) {
-
         try {
             reset(Sets.newHashSet(moduleNames));
         } catch (Exception e) {
@@ -49,6 +52,8 @@ public class Environment {
     }
 
     protected void reset(Set<Name> moduleNames) throws Exception {
+        this.context = new ContextImpl();
+        CoreRegistry.setContext(context);
 
         setupPathManager();
 
@@ -139,8 +144,12 @@ public class Environment {
      * @throws Exception if something goes wrong
      */
     public void close() throws Exception {
-        // nothing to do
+        CoreRegistry.setContext(null);
+        context = null;
     }
 
+    public Context getContext() {
+        return context;
+    }
 }
 
--- a/engine/src/main/java/org/terasology/context/Context.java
+++ b/engine/src/main/java/org/terasology/context/Context.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2015 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.context;
+
+/**
+ * Provides classes with the utility objects that belong to the context they are running in.
+ *
+ * Use dependency injection or this interface to get at the objects you want to use.
+ *
+ *
+ * From this class there can be multiple instances. For example we have the option of letting a client and server run
+ * concurrently in one VM, by letting them work with two separate context objects.
+ *
+ * This class is intended to replace the CoreRegistry and other static means to get utility objects.
+ *
+ * Contexts must be thread save!
+ */
+public interface Context {
+
+    /**
+     * @return the object that is known in this context for this type.
+     */
+    <T> T get(Class<? extends T> type);
+
+    /**
+     * Makes the object known in this context to be the object to work with for the given type.
+     */
+    <T, U extends T> void put(Class<T> type, U object);
+
+
+    /**
+     * Removes the object of the given type from this context. Future get calls may however still return
+     * an object from the parent context for that type.
+     */
+    void remove(Class<?> type);
+}
--- a/engine/src/main/java/org/terasology/context/Context.java
+++ b/engine/src/main/java/org/terasology/context/Context.java
@@ -26,7 +26,7 @@ package org.terasology.context;
  *
  * This class is intended to replace the CoreRegistry and other static means to get utility objects.
  *
- * Contexts must be thread save!
+ * Contexts must be thread safe!
  */
 public interface Context {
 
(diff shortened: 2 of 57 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/72/

**Build Log**
*last 5 lines*

 ```
[...truncated 4381 lines...]
[ANALYSIS-COLLECTOR] Computing warning deltas based on reference build #71
Setting status of 0a4b47f420da4365ca702b9ef798a3f2dd963d0b to FAILURE with url http://jenkins.terasology.org/job/TerasologyPRs/72/ and message: Merged build finished.
```

**Comment by GooeyHub** [comment]
http://jenkins.terasology.org/job/TerasologyPRs/73/

**Comment by GooeyHub** [comment]
http://jenkins.terasology.org/job/TerasologyPRs/74/

**Comment by msteiger** [comment]
It would be great if you could summarize the changes in a few sentences as the PR is quite large.

As I see it, a `context` is set in constructors of various classes and then used to replace the CoreRegistry? Is that the idea?

**Comment by Cervator** [comment]
Tests out OK locally, although yeah I only get the general idea of the PR and the overall goal :-)

Relates to http://forum.terasology.org/threads/making-singleplayer-a-special-case-of-multiplayer.1291

The only module compile error I noted is in TinyEnvironment in PolyWorld, where the CoreRegistry.putPermanently seems to be gone - what would the replacement be?

`CoreRegistry.putPermanently(WorldGeneratorPluginLibrary.class, new WorldGeneratorPluginLibrary() {`

**Comment by Cervator** [comment]
PolyWorld should be happy with https://github.com/Terasology/PolyWorld/issues/3 merged after. Thanks @flo !

Pinging @immortius and @MarcinSc for further review :-)

**Comment by flo** [comment]
@msteiger the long term goal of this pull request series is to get rid of static variables so that it is for example possible to introduce a test mode in which a client connects to a headless server in a single JVM, possibly with configurable delay.

The main change of this pull request is the intrduction of a Context class.

To make the transition smooth the CoreRegistry class and it's most frequently used method "put" and "get" will remain. To make it possible to use either the non static Context object, or the static CoreRegistry methods the CoreRegistry will now internally work with a Context object.

The Context object provides non static put and get methods that equal in it's functionality those of CoreRegistry. The Context objects are hirachical. The TerasologyEngine class and only that class has access to a engine Context object which stores all objects that exist for the whole lifetime of the TerasologyEngine instance. All objects that got previously added to the CoreRegistry via putPermanently will go into that Context. The putPermanently method got removed from CoreRegistry. It got only used by a test of the PolyWorld module that is not yet considered stable.

For it's other states the TerasologyEngine offers now a method to create a child context. If a child context does not find an object then it will look it up at it's parent. Thus all objects that get for example registered during the loading of the game will only be in the child context that lifes as long as the game goes. The next game has a new context and will thus no longer contain the objects that got added to the context of the previous game.

Until the CoreRegistry usage has been removed completly, it will be necessary to set a context for it. During the initialization of the engine it is set to the engine context. Afterwards it is set to the context owned by the game states.

The main focus of the patch is the introduction of the Context object, it is however also removing a bit of CoreRegistry usage as a side effect.

**Comment by emanuele3d** [comment]
A couple of side questions:

As far as I understand the CoreRegistry only allowed to register one instance per class. Would it make sense for these Context objects to be more flexible, i.e. able to register any object against a simple string?

Also, in line with my recent suggestion of making Config Observable, it would be useful perhaps for these Context objects to be observable too? It would be useful to be informed if a context-held object has been replaced, removed or has even become available.

**Comment by MarcinSc** [comment]
I'm not sure if Context objects are supposed to be changing, its main purpose is going to be to set the @In dependencies in systems, and mutating those might require a lot of diligence in each system to be able to respond to those changes.

**Comment by flo** [comment]
@emanuele3d  I thought about string keys too, but in the end I think it's better to register just one instance per class: It is simple and nice solution (e.g. type savety, no need to enforce name conventions) and if you really need to have multiple instances of a class then you can still register a Manager class that allows you to get instances by string key. e.g. like a AssetManager.

Yes and I agree with @MarcinSc. They are supposed to be static due to their use in @In and the big question what would happen with existing injected classes when you change it later on.

**Comment by emanuele3d** [comment]
Good point about registering Managers to obtain instances or other types of objects by string. 

Regarding the registered objects being unchanging: good to know that's the aim. I'll double-check what I'm doing in renderland to make sure the objects I register do not change.

Thank you both for your replies.

**Comment by msteiger** [comment]
Thanks - sounds good to me.

**Comment by Cervator** [comment]
Bump - @immortius: got a moment to look at this sometime please? :octocat:

File: engine-tests/src/main/java/org/terasology/Environment.java [hunk 1]
```diff
@@ -34,12 +36,13 @@
 
     private static final Logger logger = LoggerFactory.getLogger(Environment.class);
 
+    protected Context context;
```
immortius: Given their is a public getContext() method I would suggest making this private, unless subclasses are replacing the context?

engine/src/main/java/org/terasology/context/Context.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ *
+ * From this class there can be multiple instances. For example we have the option of letting a client and server run
+ * concurrently in one VM, by letting them work with two separate context objects.
+ *
+ * This class is intended to replace the CoreRegistry and other static means to get utility objects.
+ *
+ * Contexts must be thread save!
+ */
+public interface Context {
```
immortius: This is probably really pedantic, by I would suggest renaming to Registry or similar, since there is nothing particularly contextual about this class, just the way it is used.

engine/src/main/java/org/terasology/context/Context.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Use dependency injection or this interface to get at the objects you want to use.
+ *
+ *
+ * From this class there can be multiple instances. For example we have the option of letting a client and server run
+ * concurrently in one VM, by letting them work with two separate context objects.
+ *
+ * This class is intended to replace the CoreRegistry and other static means to get utility objects.
+ *
+ * Contexts must be thread save!
```
immortius: -save +safe

engine/src/main/java/org/terasology/registry/InjectionHelper.java
```diff
@@ -31,7 +32,28 @@
 public final class InjectionHelper {
     private static final Logger logger = LoggerFactory.getLogger(InjectionHelper.class);
 
-    private InjectionHelper() {
+    public InjectionHelper(Context context) {
```
immortius: This is a static helper class, it shouldn't have a constructor?

immortius:
Ran my eyes over it. I see no fundamental issues with the approach, but found a couple of areas needing tweaks. Also suggest name change from Context to Registry (or similar), as I feel Registry describes what the class is better, and the contextual nature of it is how it is used - the class itself does not represent a context as such.

flo:
I think Context is a good name, as:
1. you have one instance per context. e.g. server and client context
2. it contains the objects that makes up the context e.g. the entity manager that belongs to that context
3. the name context implies that there is another instance per context, which is exactly what I want to express. Just calling it Registry would you make assume that there is just one instance.

engine-tests/src/main/java/org/terasology/Environment.java
flo: I had it first private too, but the field gets used a lot in the sub class HeadlessEnvironment and I think:
context.put(...)
reads better than:
getContext().put(...)

engine/src/main/java/org/terasology/context/Context.java
flo: thanks good catch.

engine/src/main/java/org/terasology/registry/InjectionHelper.java
flo: oops, that was a leftover from a refactor to make it a non static helper class which I reverted later on.

immortius:
I don't really like naming classes by how they are used rather than what they do (List doesn't have anything to say about whether there is one or many of them) and disagree that Registry implies singleton, but not really bothered enough to argue further.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/76/

Cervator:
Any further comments on this or should we go ahead and merge?

I can see both sides on Context vs. Registry. There is only one solution - Google Fight!

http://www.googlefight.com/index.php?word1=context&word2=registry

Context wins by about double and for some reason I'm getting results in Spanish.

Joking aside I really don't want to pick a winner on arbitrary grounds, but as long as work gets done I'm happy :-)

PR #1727, opened 2015-05-16
Remove CoreRegistry usage: Part II

flo:
Part II of a possibly long series to get rid of CoreRegistry usage.

This pull request requires #1721 to be merged first. See #1721 and its comments for an explanation of why the CoreRegistry usage gets removed.

It contains no big refactorings except maybe for the commit https://github.com/MovingBlocks/Terasology/commit/1cbac8a7d1549026ca8627965a9cbb68db291546 that converts EntitySystemBuilder into a static utility class as suggested by a TODO statement.

### DIFF
--- a/engine-tests/src/main/java/org/terasology/DisplayEnvironment.java
+++ b/engine-tests/src/main/java/org/terasology/DisplayEnvironment.java
@@ -31,9 +31,6 @@ import org.terasology.entitySystem.prefab.Prefab;
 import org.terasology.entitySystem.prefab.PrefabData;
 import org.terasology.entitySystem.prefab.internal.PojoPrefab;
 import org.terasology.naming.Name;
-import org.terasology.registry.CoreRegistry;
-import org.terasology.rendering.ShaderManager;
-import org.terasology.rendering.ShaderManagerLwjgl;
 import org.terasology.rendering.assets.animation.MeshAnimation;
 import org.terasology.rendering.assets.animation.MeshAnimationData;
 import org.terasology.rendering.assets.animation.MeshAnimationImpl;
@@ -87,7 +84,7 @@ public class DisplayEnvironment extends HeadlessEnvironment {
 
         try {
             Display.setDisplayMode(new DisplayMode(1, 1));
-            Display.create(CoreRegistry.get(Config.class).getRendering().getPixelFormat());
+            Display.create(context.get(Config.class).getRendering().getPixelFormat());
         } catch (LWJGLException e) {
             throw new IllegalStateException(e);
         }
@@ -95,8 +92,8 @@ public class DisplayEnvironment extends HeadlessEnvironment {
 
     @Override
     protected void setupAssetManager() {
-        AssetManager assetManager = new AssetManagerImpl(CoreRegistry.get(ModuleManager.class).getEnvironment());
-        CoreRegistry.put(AssetManager.class, assetManager);
+        AssetManager assetManager = new AssetManagerImpl(context.get(ModuleManager.class).getEnvironment());
+        context.put(AssetManager.class, assetManager);
         AssetType.registerAssetTypes(assetManager);
 
         assetManager.setAssetFactory(AssetType.PREFAB, new AssetFactory<PrefabData, Prefab>() {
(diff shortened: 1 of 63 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/80/

Cervator:
Just merged  #1721 so we can get started on this. It *looks* like the revision graph is still looking good, so we can avoid duplicate commits from part 1. Or can you try to see what would happen if you rebase this on top of develop?

Not to squash commits, just to avoid any funny merges.

flo:
it should merge fine (it's git after all) but I did a rebase for you.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/81/

Cervator:
Thanks! I appreciate it. I know it'll merge, I just like being able to minimize the merge commits, offers some conveniences :-)

... * some time passes * ...

Annd back, to actually test this out and submit this comment. Works fine locally with all the lineup modules :-)

@immortius @MarcinSc @msteiger @emanuele3d any time to review and throw some comments on here?

Cervator:
Bump - paging @immortius & @MarcinSc please pick up the courtesy phone, you have a collect call, do you accept the charges? :-)

(weekend is fine if needed, just fishing for a response)

emanuele3d:
I just went through it. Can't see anything obviously wrong with it. 

Question: when a child context goes out of scope, i.e. because of a GameState change I imagine, what happens to the objects registered with it? Does the context get invalidated or something? Do those objects get informed somehow? Sorry if I'm missing the obvious.

MarcinSc:
I guess these objects will get out of scope as well and be cleaned up by the VM as well.

emanuele3d:
...mumble...mumble... mumble...

I imagine some of those objects might need their dispose() method called for a proper cleanup. Within the context of a GameState I imagine it still the task of the GameState.dispose() method to do everything. I just wonder about more "casual" uses of children context objects. But perhaps I'm already thinking of abuses of something that is mostly designed to have only a couple of levels? To allow for client/servers in the same VM rather than, say, multiple context within a module?

flo:
@emanuele3d Yes some of the objects that got put into the context need a disposal. The conversion from CoreRegistry to Context does not change how it is done currently. Currently the game states just dispose what they are creating.

I think doing multiple changes at once in the same area is a bad idea, so I suggest we discuss this topic again when the refactoring is done. There are probably reasons against and for automatic disposal at the end of the lifetime of a context.

Currently there are a lot of small changes still to be done before the last CoreRegistry usage has been removed from the game.

@Cervator This pull request is btw. pretty minor. As there is really nothing special about it. It is just a follow up.

Cervator:
Righto. Will merge when I get a little time, or somebody else can test or merge and beat me to it.

emanuele3d:
@flo: yes, I understand and I agree with your suggestion. If you can, please let us know in which PR it would be most convenient to bring up the disposal issue again.

Cervator:
Even better: Submit an issue for later with whatever details we can think of so we won't forget :-)

MarcinSc:
I strongly disagree on the automatic disposal. The fact that some context goes out of scope and it happens to have an object in it, does not necessarily mean, that the object should be disposed.

It's the responsibility of the class that places the object in the context, to clean up when the context goes out of scope. Only that class knows, if the object it has added to context was created and this class is a sole owner of that object - therefore being able to manage its life-cycle, or if it is shared with other objects and it should not dispose it.

Cervator:
Could we introduce something like a Context AutoCloseable? So that if an object *should* auto-dispose if it goes out of context it'll do so?

Probably there is similar potential elsewhere. Although yeah that's for a later PR

emanuele3d:
@MarcinSc: just to be clear, personally I wasn't suggesting automatic disposal. But either registered/registering objects must be aware that a context might be no longer valid and have a way to check for it or, perhaps better, the context could inform observers that it is being disposed, so that they can respond to it appropriately.

MarcinSc:
The class that creates the child context (and implicitly or explicitly gets rid of it) is the only one that can pass it to other objects, therefore it's that class responsibility to notify them of the context being removed, if they need to be, though they will most likely also live in that context therefore will disappear with it.

I don't think this responsibility lies anywhere near the context itself.

emanuele3d:
Ok, understood.

Cervator:
Tested with the main module lineup, no issues noted, unit tests still looking fine including all the modules :-)

Made #1737 for later reference. Thanks as usual @flo and bring on part 3!

immortius:
This merge ruins my branch unfortunately. :(

Cervator:
Doh, sorry about that. Gotta speak up in time :(

How do we best resolve? There are no merged commits on top yet (although part 3 hanging off it) so we could possibly revert, merge, and redo, although I see @flo looks to have a part 3 + assets merge on the graph

flo:
@Cervator  I already helped @immortius  with merging. So it's not an issue.

Cervator:
Awesome, I was wondering if maybe I missed some action on IRC :-) Thanks!

PR #1738, opened 2015-05-23
Compute unit test coverage using JaCoCo

msteiger:
This PR adds JaCoCo to the engine-tests project, which computes code coverage. The resulting exec file can be visualized by the Jenkins JaCoCo plugin, for example:

http://jenkins.terasology.org/job/TerasologyTemp/lastBuild/jacoco/

I excluded the `o.t.protobuf` package from both JaCoCo and the Jenkins plugins (which reports 0% coverage otherwise).

### DIFF
--- a/engine-tests/build.gradle
+++ b/engine-tests/build.gradle
@@ -3,7 +3,11 @@
 // Grab all the common stuff like plugins to use, artifact repositories, code analysis config
 apply from: "$rootDir/config/gradle/artifactory.gradle"
 
+// Computes code coverage of (unit) tests
+apply plugin: 'jacoco'
+
 import groovy.json.JsonSlurper
+
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 // Java Section                                                                                                      //
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -48,4 +52,24 @@ dependencies {
     compile group: 'junit', name: 'junit', version: '4.12'
     compile group: 'org.mockito', name: 'mockito-all', version: '1.10.19'
     compile group: 'org.jboss.shrinkwrap', name: 'shrinkwrap-depchain-java7', version: '1.1.3'
+}
+
+jacoco {
+    toolVersion = "0.7.5.201505241946"
+}
+
+test {
+    jacoco {
+        append = false
+        excludes = ["org.terasology.protobuf.*"]
+    }
+}
+
+jacocoTestReport {
+    dependsOn test // why is this not by default?
+    reports {
+        xml.enabled false
+        csv.enabled false
+        html.enabled false
+    }
 }


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/94/

emanuele3d:
I like little histograms with numbers around them!!!

(...nervously shuffles his feet...)

But what do I do with them? Are they edible?

MarcinSc:
I heard that pie charts are edible, not sure about histograms...

emanuele3d:
=D

Cervator:
I hear histograms taste like aged graham crackers, not much to write home about ...

And woo, we're number ... 26.9th percentile! Honestly I thought coverage was going to be worse :-)

Thanks for kicking this off @msteiger, looking forward to playing with it some more. Been trying to get code coverage to do stuff in IntelliJ.

Cervator:
I haven't gotten further on this yet, have just looked at both https://coveralls.io and https://codecov.io which is also a thing (and even has a browser addon for GitHub). Both look like good options.

On a subtly related note I ran across http://www.yegor256.com/award.html on Twitter - a contest for code quality principled projects, heh.

msteiger:
I'm happy with whatever fits best :smile:

msteiger:
I wasn't able to configure an organization repo such as `MovingBlocks/master-server` in coveralls (maybe I overlooked the right switch), so I tried codecov.io and it worked out of the box.

The [instruction guide](https://codecov.io/guide#java) suggests installing the JaCoCo plugin and uploading the generated files to codecov.io.

So I guess that running JaCoCo on Jenkins is a reasonable step in any case.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/148/

Cervator:
Finally merged this locally for some testing and figuring out how IntelliJ handles code coverage.

It has its own style in addition to JaCoCo which is somehow supposed to be fancier I guess, but it sure takes a while - just past 23 minutes to run all tests, yikes!

Switched to JaCoCo and it ran in two minutes, same as via Gradle.

There is a run config where you can set JaCoCo and I'd like to generate that along with the other run configs, but oddly I don't see an option to exclude packages like protobuf

@msteiger did you get any special/good options in Eclipse or just tried Gradle and Jenkins?

One nice bit for IntelliJ is that it highlights lines according to the coverage, so that's pretty cool. Doesn't really matter much if it includes protobuf, can just ignore that.

Also curious to check this out with modules then will merge already and check out the coverage sites :-)

Cervator:
Ah-hah! Sneaky sneaky.

Sorted this out. Applied the setup globally instead of specifically for engine-tests, that way we get coverage in module builds now too with a simple tweak in the build template :-)

But huhwha! All coverages were suddenly 0%. Yet the TerasologyTemp job clearly showed that it worked! Turns out the Jenkins plugin isn't updated to use JaCoCo 0.7.5 yet, however that was what the config was set to use. Downgraded to 0.7.4 and it worked. Smells like a rebase after the initial test at 0.7.4? :D

A cool side-effect: Unlike unit test reporting JaCoCo can be enabled in a generic fashion, it won't break the build if there are no unit tests (or even no Java code). If somebody then adds unit tests suddenly there'll be a test coverage percentage on the [Statistics tab](http://jenkins.terasology.org/view/Statistics) - giving us a clue to enable unit test displaying on that job :-)

Actually happened as I hadn't realized MarkovChains had tests, I had just included it because I knew NameGenerator did and MC was a dependency, haha. And such coverage too, our current record holder at 93.65% !

In a similar way modules without any unit test will list at a glorious 0% coverage, making it clear who the guilty parties are. Like me ...

It does also manage to list 0% for pure asset modules without code though, but ... eh, details details.

Cervator:
@msteiger bit of a bump here on an old issue to keep this centralized. I played with JaCoCo and Codecov some today, but with very little luck. JaCoCo [art works fine via module template in Jenkins, but the step to upload to codecov is giving me a major headache.

I get what the bash curl thing is supposed to do, but the way Jenkins wraps the contents of a shell step into a temporary .sh file seems to screw it up. And that's where my limited knowledge of Linux hurts, trying to Google the use of specific symbols in a bash script with an apparently still fairly obscure tool (codecov) encountering weird errors is rough. 

While Codecov's site is great and polished it really only seems skin deep with little to explain the inner workings when stuff goes wrong. I eventually found the GitHub repo that actually contains the script the curl downloads, and saw some parameters in there, but had to resort to copying the whole thing to a file on the Jenkins server

Then even at that point it didn't work as the uploader didn't find anything in the module job I ran as a test (NanoMarkovChains) - the JaCoCo report gets written out to an .exec file but the script is seemingly looking for .gnco files (huh). Searching for alternatives to output JaCoCo was fruitless, the options in Gradle to write out XML etc didn't do anything useful for Codecov.

What did you do that let it work out of the box? Magic? :-)

http://jenkins.terasology.org/view/Nanoware/job/NanoMarkovChains/10/console

PR #1745, opened 2015-05-26
Introducing the PostProcessor

emanuele3d:
This PR introduces a new class in support of the renderer: the **PostProcessor**. 

The **[first commit](https://github.com/emanuele3d/Terasology/commit/969412a975f08e07f2bbb4725467ff42d4e34ddc)** simply extracts the post-processing functionality from class LwjglRenderingProcess into the new PostProcessor class, while the **[second commit](https://github.com/emanuele3d/Terasology/commit/09a2a6d9917fdbff38ea51d9c70a7b5145fb43fb)** patches everything back so that it can compile and run.

In the **[third commit](https://github.com/emanuele3d/Terasology/commit/bd418e2776850ca3b0a153ce7b094888f37f4864)** materials used by the PostProcessor are fetched only once, shortly (but not immediately) after the class is instantiated. Previously materials were fetched once per frame and in some cases the same material was fetched multiple times during the same frame.

In the **[fourth commit](https://github.com/emanuele3d/Terasology/commit/16a991c6a4def54b074b0ea312981a539e3119a0)** the most important change is that FrameBuffers used by the PostProcessor are stored in internal references. When appropriate, the LwjglRenderingProcess class triggers the PostProcessor to refresh its references. Some additional refactoring continues the work on WorldRendererImpl.render() method, to make it a complete high-level index of what happens each frame to produce the final image on screen.

The **[fifth commit](https://github.com/emanuele3d/Terasology/commit/f1b9dcaba516d93939ea4ebed07ff1bdf59c8cdb)** primarily deals with refactoring the final stages of post-processing. In particular it simplifies the code by separating early the paths for mono and stereo (OculusVR really) rendering. I seem to remember this incidentally fixes a number of problems when taking screenshots, especially in OculusVR mode. I didn't log this in the commit description and it is too late at night to check, hence my uncertainty. ;)

The **[sixth commit](https://github.com/emanuele3d/Terasology/commit/3b7eec17041944cfc1ce8e81f5ff5d1f02bda9bb)** just renames some variables and methods as I got to know them better. It also introduce a PostProcessor.dispose() method. 

The **[seventh commit](https://github.com/emanuele3d/Terasology/commit/53e649200570be4b5acfe629a25cc04a9557a73b)** shuffles a number of PostProcessor lines around, so that most post-processing methods follow the same order of operations: dealing with materials, dealing with textures, FBO-related state changes and then rendering.

The **[eight commit](https://github.com/emanuele3d/Terasology/commit/eb1fb682274e6307972377701c723ecb7adf42ac)** adds javadocs for most public methods, comments and TODOs for (near) future work. There is also some very limited, minor refactoring.

**In conclusion**: this PR is far from the final destination for this batch of renderer refactoring. Even the PostProcessor class needs further work. But I felt this is a good point where to pause and merge into develop, as future work on the PostProcessor will be more arcane and delicate.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/109/

Cervator:
This may be one of the best documented PRs of all time :D The per-commit thing really works well.

Thank you for keeping up with that by the way! I think your PRs are probably the most consistently well-documented we have. Even if there isn't always somebody able to truly review it line by line this is the way to keep changes understandable to a wide audience.

Naturally this code is still waaay beyond my comfort zone to review but I wanted to at least say that :-)

emanuele3d:
Thanks Cervator! I'm just assuming people do not have the time to check everything and this way they can zoom in on something that picks their interest. At the same time some commits are not really that interesting, i.e 1, 6 and 7, and I hope people will notice they have a lower priority.

Regarding comfort zones... I hear ya! I banged my head so many times on this code it certainly feels uncomfortable!!!! =D

immortius:
I will have a look this weekend.

Cervator:
Hey @emanuele3d - we should be able to get this moving again now. Can you check on what conflicts? Hopefully fairly simple changes in imports / asset related method names.

emanuele3d:
Yep, will look into it.

emanuele3d:
@Cervator: I did a couple of tests, one rebasing the branch of this PR on top of develop, the other a merge.

If I push the merged version of the branch, a bunch of commits unrelated to this PR will become listed here in the PR, right? On the other hand my commit-links above will still point to the right commits.

If, on the other hand, I push the rebased version of the branch, this PR will get a new set of commits and I'll just have to update the links. I would therefore be inclined to rebase, to keep the PR clean. Also,  there are currently no commit-specific comments, so I don't think a rebase in this circumstance would break anything. Does this make sense?

Additionally @immortius, I've just noticed a change regarding the retrieval of assets in the WorldRendererImpl class. One of the affected lines follows and the PostProcessor introduced by this PR would probably also need some of its lines modified in a similar way:

```java
chunkShader = Assets.getMaterial("engine:prog.chunk").orElseThrow(() -> new RuntimeException("Failed to resolve required engine material"));
```

It looks very ugly. Can't I just check if the returned object is null using the .get() method? Also, if I recall correctly a shader problem would throw an exception during compilation/linking, before the shaders are registered as engine assets. Why would this type of line ever fail then?

Cervator:
@emanuele3d yep that sounds correct, and yeah rebase is better :-)

I'm still not totally on top of the fine grained details of how well Github handles commit-comments in a PR that then get rebased. Same with links to specific commits. Lets find out!

immortius:
> Can't I just check if the returned object is null using the .get() method?

No, get() throws an exception if the value is not present. You can use the isPresent() method to check whether the value is null.  So you can use this to have an if statement if desired. 

That said I don't think it is ugly - it is new, since we haven't had lambdas until recently. But they are going to be much more common going forwards.

>Also, if I recall correctly a shader problem would throw an exception during compilation/linking, before the shaders are registered as engine assets. Why would this type of line ever fail then?

It fails if the asset could not be loaded for any reason - the files are not found or could not be loaded for instance. This can happen before compilation/linking.

Of course, since this is required to run the game it *should* never fail, hence the runtime exception. Also why it isn't worth 5 lines of code.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/145/

emanuele3d:
Rebased and [updated](https://github.com/MovingBlocks/Terasology/commit/62d88108e755417db3daaf8fe0aefe2f32865afe) for post gestalt-asset integration.

Also updated the links in the PR description to point to the appropriate post-rebase commits.

Cervator:
@immortius think there is anything left to review and tweak here or time to merge?

immortius:
All looks good to me.

Cervator:
Merged at last. Huzzah!

emanuele3d:
Thank you!

PR #1750, opened 2015-05-31
Read entity height from player prefab

### LINKED ISSUE #1711: Follow-up to #1706 - use player height from player.prefab
Tiny easy tweak on top of #1706 for later. See https://github.com/MovingBlocks/Terasology/pull/1706#discussion-diff-29544636 for the spot, player.prefab has a character height that could be used in the calculation instead of a hard coded value. That way a future giant race player won't spawn under a tiny tree!

msteiger:
This was more effort that I first thought since..
* The height of an entity depends on the type of collision object
* The entity was created by `PlayerFactory` only after the final spawning pos. was set

So I moved a few bits from `PlayerSystem` to `PlayerFactory`, which also looks cleaner to me now. The system deals with events and entities while the factory deals with creating new player instances. Just as the names incline.

Fixes #1711

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java
@@ -33,7 +33,6 @@ import org.terasology.network.Replicate;
  * @author Immortius
  */
 public final class CharacterComponent implements Component {
-    public Vector3f spawnPosition = new Vector3f();
     public float eyeOffset = 0.6f;
     /**
      * Specifies the maximium range at which this character is able to interact with other objects.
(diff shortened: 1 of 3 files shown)


Cervator:
Nice, I remember this. Reminds me of the grow/shrink potions we need to make already, plus a weird world to test where the player would spawn at different sizes :D

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/119/

Cervator:
This one could use a quick rebase now, probably not too bad :-)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/139/

PR #1758, opened 2015-06-03
Fix character encoding issues in source files

msteiger:
I stumbled over this when one of the Jenkins slave agents complained about invalid chars in java source files. So I replaced all non-ASCII characters and used escaped unicode instead. Easy fix.

That said, explicitly defining the encoding to `UTF-8` for `javac` should fix the same issue at least for the java compiler.

While I think that sticking with plain ASCII is reasonable for source files (translation files are certainly different), I didn't want to enforce this and went for `UTF-8` instead.

### DIFF
--- a/config/gradle/common.gradle
+++ b/config/gradle/common.gradle
@@ -11,6 +11,12 @@ apply plugin: 'findbugs'
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
+javadoc.options.encoding = 'UTF-8'
+
+tasks.withType(JavaCompile) {
+    options.encoding = 'UTF-8'
+}
+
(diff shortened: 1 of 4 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/127/

Cervator:
Heh, okay, saw what this caused first-hand, no more broken builds from Javadoc for a bit :D

Interesting side effect from the new builder!

Note that I reconfigured it to have only one executor + only build jobs matching its restrictions. Otherwise the applet job would start there and not have the usual Gradle prop file with signing info etc and two builds at once might hit other trouble :-)

PR #1735, opened 2015-05-22
Reducing the number of Rotations from 64 to the actual 24 possible, b…

MarcinSc:
…ut for backward compatibility allowing the use of the 40 remaining using a duplicate mapping.

Introduced a new block family that supports all the possible 24 rotations of the block.

PR #1768, opened 2015-06-09
Fixed #1764, possibly fixed #1767 (needs to be looked up) and some other improvements

### LINKED ISSUE #1764: Prefabs and entities with Collection (List/Set) of @ManagedContainer type objects are not read correctly
Entities with component that has a collection of @ManagedContainer objects throw exception when getting deserialized (with old asset code, did not check new one).

Also, prefabs with such configuration cannot be loaded (throw an exception).

### LINKED ISSUE #1767: PreBegin method is called after world is loaded, contrary to what JavaDoc says
PreBegin method of ComponentSystem is called after world is loaded, which is the opposite of what the javadoc for that method says.

MarcinSc:
The other improvements are:
- reduced the number of Rotations (to only those possible),
- removed AllSidesBlockFamily - doesn't make sense, as the "top" and "bottom" one does not have well defined sides,
- added method "getSide(Block)" to SideDefinedBlockFamily to avoid calling "Block::getDirection()" that will most likely be removed.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/135/

Cervator:
I figure this is the second coming of #1735 now that the big merge is done, right? And will probably be what helps fix the few broken modules likely set up to take advantage of this.

In any case it has architecture written all over it so casting Summon @immortius for review :-)

@emanuele3d was interested last go-around too so ping!

engine/src/main/java/org/terasology/math/Rotation.java
```diff
@@ -79,19 +82,40 @@ public static Rotation rotate(Yaw yaw, Pitch pitch, Roll roll) {
         return ROTATIONS.get(indexFor(yaw, pitch, roll));
     }
 
+    public static Rotation findReverse(Rotation rotation) {
+        Side frontResult = rotation.rotate(Side.FRONT);
```
immortius: I would suggest the reverse of a rotation should be the rotation that undoes the rotation. I'm not sure this code achieves this? If it does I think there is a likely a more direct way of achieving it.

I also assume that null shouldn't be a legitimate result, I would suggest a RuntimeException thrown instead.

engine/src/main/java/org/terasology/math/Rotation.java
```diff
@@ -17,24 +17,27 @@
...
  * <br><br>
  * Uses the fly weight pattern to cache the 64 combinations of rotation.
+ * Note (Marcin Sciesinski): there are actually only 24 possible combinations, the remaining 40 are just duplicates
  *
  * @author Immortius
  */
 public final class Rotation {
 
     private static final TByteObjectMap<Rotation> ROTATIONS;
+    private static final TByteByteMap DUPLICATE_ROTATIONS;
```
immortius: Is this needed? Why not just put the same rotation into multiple positions of the main map?

engine/src/main/java/org/terasology/math/Rotation.java
```diff
@@ -17,24 +17,27 @@
...
 
 import java.util.List;
 
-import org.terasology.math.geom.Quat4f;
-
 /**
  * Rotation provides easy access to 90 degree increments of rotations - intended for block-related rotations.
  * <br><br>
  * Uses the fly weight pattern to cache the 64 combinations of rotation.
+ * Note (Marcin Sciesinski): there are actually only 24 possible combinations, the remaining 40 are just duplicates
```
immortius: This assumes you don't care about the actual yaw/pitch/roll, only the resultant rotation. Is this a safe assumption? Yaw/pitch/roll are visible attributes.

engine/src/main/java/org/terasology/persistence/serializers/EntityDataJSONFormat.java
```diff
@@ -364,8 +364,8 @@ public JsonElement serialize(EntityData.Value src, Type typeOfSrc, JsonSerializa
                 for (JsonElement element : jsonArray) {
                     if (element.isJsonArray()) {
                         value.addValue((EntityData.Value) context.deserialize(element, EntityData.Value.class));
-                    } else if (json.isJsonObject()) {
-                        extractMap(json, context, value);
+                    } else if (element.isJsonObject()) {
```
immortius: Good catch

engine/src/main/java/org/terasology/math/Rotation.java
immortius: Ok, I've had a closer look. This does achieve that result.

engine/src/main/java/org/terasology/math/Rotation.java
```diff
@@ -17,24 +17,27 @@
...
  *
  * @author Immortius
  */
 public final class Rotation {
```
immortius: I'm going to suggest it would probably clean up the code to change this to calculate all the rotations up front - with each new rotation, checking to see if any of the previous rotations have the same value and using that existing rotation instead. Additionally, after all the rotations are created determining the opposite rotation for each one and setting that as a value in the rotations.

engine/src/main/java/org/terasology/persistence/typeHandling/protobuf/ProtobufPersistedData.java
```diff
@@ -40,9 +40,11 @@
 import org.terasology.protobuf.EntityData;
 
 import java.nio.ByteBuffer;
+import java.util.HashMap;
```
immortius: Seems to be a couple of unused imports here.

engine/src/main/java/org/terasology/math/Rotation.java
immortius: I might even suggest keeping the 64 rotations, but providing a normalise method that provides the "base" rotation which has identical effect, if a person wants to work with just the 24 resultant rotations. ;)

engine/src/main/java/org/terasology/world/block/family/SideDefinedBlockFamily.java
```diff
@@ -20,4 +20,5 @@
 
 public interface SideDefinedBlockFamily extends BlockFamily {
```
immortius: While not a matter for this PR, I don't feel this interface conveys its meaning very well. By side here we actually mean what, the facing direction of the block? Would this conveyed better if it was called SideFacingBlockFamily, and the methods be ```Block getBlockFacing(Side)``` and ```Side getFacingSideOf(Block)```

immortius:
> Moved FullRotationFamily to Signalling, though I'm sure it will be copy/pasted to many mods.

This is an argument for it being in a common module, not being in engine. The point of engine is to provide the minimum required to support the game, not to provide features that of common or even universal appeal. I know there is a lot of cruft in engine, and we have not always done well at keeping to this principle, but I still feel we should aspire to it.

I suggest we should create a new "common" module, siphon anything relevant out of core into it, add your FullRotationFamily, and start maintaining this as a source for commonly useful but piecemeal features, at least until enough stuff builds up that it makes sense to split them out into separate modules.  Would need to come up with some inclusion guidelines along the lines of guava (need to ensure things are commonly useful and not just some neat feature which one person needs).

immortius:
>Would this be a convenient time to also mark Block.getDirection() and Block.setDirection(...) as deprecated (or I suppose even remove it)? To me it seemed like this getSide(...) functionality was its intention.

@Josharias Maybe. Should we chuck get/setRotation into Block, and not bother with the getSide()?

MarcinSc:
I'm ok with modifying the Block to have "get/setRotation" instead of "getSide", or just dropping the "getSide" all together.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/136/

Cervator:
A related common misc module was made last time around :-)

https://github.com/Terasology/MarcinScIncubator

Until we find a more permanent home for utility stuff, anyway. Yeah we could do a more common(ly named) module instead, but no big difference for now.

MarcinSc:
@Cervator That is in fact, where the FullRotationFamily (and associated classes) currently resides.

Josharias:
I now agree with @MarcinSc about taking out Block.getDirection() in favor of the block family implementation of what direction a block is facing.  Using the block family is the only way that modules can attach extra metadata to a particular block (in this case, the facing direction) because the Block class is final.  The Block class should probably keep to the purpose of determining how a block is displayed in the world.

The other direction we could go is to allow subclasses of Block to be created and returned from the block family.  That would allow creation of a BlockThatFacesADirection class that would then contain the method that returns the direction.  However, I am unsure if that would cause all sorts of other problems in the rest of the engine.

immortius:
Extra information can be attached to blocks via their prefabs as well - although this is awkward currently. Ideally blocks will ultimately just be entities or prefabs and the block class will disappear.

The block family is not an intended method of adding data, only to group blocks and provide the algorithms to decide which to use in different situations. Sides/rotations/directions tie into this.

Cervator:
So is that a Yea or Nay on this PR's approach for now? Or more related to future architectural plans with this PR good to merge?

immortius:
I am happy with the PR, yes. :)

PR #1772, opened 2015-06-13
Added new mode for the DebugOverlay (the stats cycled with F4)

emanuele3d:
Added a new mode to the DebugOverlay, specialized on execution time running means for rendering code only. 

While functional, the new mode filters out (does not display) activities registered with the performance monitor that are not prefixed by the strings "WorldRenderer::" or "PostProcessor::". At the time of this commit no activity is prefixed with those strings and as such the new mode displays only its title. Those prefixes will be added some time after PR #1745 has been dealt with.

Additionally, differently from various time-related modes already in place, the new mode sorts the list of post-filter activities alphabetically, and formats each item to list the time value on the left and the activity name on the right, after a separator. I.e.:

```
  0.31ms - PostProcessor::Activity 1
  0.17ms - WorldRenderer::Activity 2
  1.87ms - WorldRenderer::Activity 3
```

Some issues: 
* I'm using String.format() to format the time value and also attempt to align it. If the number of digits is not the same however (i.e. 1.00ms vs 10.00ms) alignment doesn't work, most likely because the font used by the specific UI element is not a mono-spaced font. Any way I could change that?
* I've also been wondering if it'd be possible to add color to specific words or lines. This would allow the highlighting of times and activities that fulfill some criteria, i.e. when a running mean goes above 20ms it gets marked in orange and when it goes above 40ms it gets marked in red.
* For the past couple of weeks I experimented with a much more generic solution, i.e. one where arbitrary objects for sourcing, filtering, sorting and formatting the data could be provided upon construction and even set at runtime. Eventually I opted for the simpler, inflexible solution seen here, for the time being. EDIT: I opened issue #1773 on the matter.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
@@ -68,7 +68,8 @@ public class DebugOverlay extends CoreScreenLayer {
     private WorldProvider worldProvider;
 
     private List<MetricsMode> metricsModes = Lists.newArrayList(new NullMetricsMode(), new RunningMeansMode(), new SpikesMode(),
-            new AllocationsMode(), new RunningThreadsMode(), new WorldRendererMode(), new NetworkStatsMode());
+            new AllocationsMode(), new RunningThreadsMode(), new WorldRendererMode(), new NetworkStatsMode(),
+            new RenderingExecTimeMeansMode("Rendering - Execution Time: Running Means - Sorted Alphabetically"));
     private int currentMode;
     private UILabel metricsLabel;
 
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/RenderingExecTimeMeansMode.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/RenderingExecTimeMeansMode.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2015 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/140/

Cervator:
I forget if we've added support for custom fonts yet, that might be nice to pick a monospace font in places like that. Colored text we have in the console but could it be used in the debug text? Unsure.

@msteiger and/or @flo have played with colored text IIRC. @immortius would probably know about font assets.

msteiger:
You can use [FontColor](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/FontColor.java) to encode color information in plain strings. This isn't great and I plan to replace the functionality with a new `ColoredString` class, but that not a top priority atm for me.

We're using Noto Sans, which isn't monospaced. There could be a monospaced variant though, I haven't checked. You can use ["MarkeEigenbau" font](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/fonts/default.fnt). Other (experimental) fonts are in [TeraMisc/fonts](https://github.com/MovingBlocks/TeraMisc/tree/master/fonts), which you can try. 

An alternative would be an implementation of `\t` (tab) to align text fragments across multiple lines. Not sure if that is a good idea, though.

emanuele3d:
Thank you @msteiger for the leads. Regarding the font, I know nothing about the UI system in Terasology. How do I find where to set it? Is it something that can be set on a per-widget basis?

msteiger:
I don't think it's possible to define such properties for individual widget. You need to define a skin for that, which you can apply to widgets.

For example, the mainmenu skin [defines a family "highlight" that uses the **bold** Noto font](https://github.com/MovingBlocks/Terasology/blob/99418f5d43060850e99490c5fdf1f12474f02869/engine/src/main/resources/assets/skins/mainmenu.skin#L42). Maybe you could add a family "debug" that uses a monospaced font and apply that to the debug overlay widget.

immortius:
If you want the entire text to be a particular color, this can be handled through skinning. If you want to change the font used, that can also be handled through skinning (see http://www.angelcode.com/products/bmfont/ for details on how to create a custom font).

emanuele3d:
@msteiger: thank you for the hint about defining a new "debug" family in the mainmenu skin. I'll see what I can do. 

Meanwhile perhaps this PR can just be merged in. I don't think I'll want to add the font-related frills right away.

Cervator:
Tested this out - barely. Had my PC crash right before, browser and its hundreds of tabs survived, but my IntelliJ system settings went corrupt ...

Mah shortcuts, look & feel, and other customizations! /cry

Issue #1771, opened 2015-06-13
Terasology no longer compiles on eclipse


msteiger:
With #1740 merged, Terasology does no longer compile on eclipse. The following line fails:

```java
assetTypeManager.registerCoreAssetType(BlockTile.class, BlockTile::new, "sounds");
```
*The target type of this expression is not a well formed parameterized type due to bound(s) mismatch*

I first thought that this was a bug in the eclipse compiler: [Eclipse Bug 425278](https://bugs.eclipse.org/bugs/show_bug.cgi?id=425278), but if I understood it correctly this is supposed to happen.

One of the compiler developers [reported a bug for OpenJDK](https://bugs.openjdk.java.net/browse/JDK-8033810) that `javac` compiles the code fine, but it shouldn't.

Cervator:
Huh! Now that's interesting.

I do find the :: a bit scary still, haven't taken the time to get used to it yet :D Java is getting complicated.

Cervator:
Three merged PRs later is this one resolved?

msteiger:
Yup - thanks!

PR #1774, opened 2015-06-14
Explicitly specify lambda types for asset factories

msteiger:
Since Terasology no longer compiles on eclipse (see #1771), I decided to explicitly specify the mapping of lambdas to `AssetFactory` instances to avoid type inference altogether. This circumvents the [OpenJDK "bug" 8033810](https://bugs.openjdk.java.net/browse/JDK-8033810).

This is certainly more verbose than before, probably less elegant, but maybe more instructive and it compiles fine on both compilers.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
+++ b/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
@@ -19,8 +19,10 @@ package org.terasology.engine;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.Queues;
 import com.google.common.collect.Sets;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.terasology.assets.AssetFactory;
 import org.terasology.assets.management.AssetManager;
 import org.terasology.assets.module.ModuleAwareAssetTypeManager;
 import org.terasology.config.Config;
@@ -37,6 +39,7 @@ import org.terasology.engine.subsystem.DisplayDevice;
 import org.terasology.engine.subsystem.EngineSubsystem;
 import org.terasology.engine.subsystem.RenderingSubsystemFactory;
 import org.terasology.entitySystem.prefab.Prefab;
+import org.terasology.entitySystem.prefab.PrefabData;
 import org.terasology.entitySystem.prefab.internal.PojoPrefab;
 import org.terasology.game.Game;
 import org.terasology.identity.CertificateGenerator;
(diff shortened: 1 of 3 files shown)


MarcinSc:
Any chance of making the lambda expressions working in modules

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/141/

Cervator:
Lambdas working in modules - as in, they don't presently, or they too break in Eclipse? Since it seems this PR just tweaks some usages in the engine, but do not change anything fundamentally?

MarcinSc:
Last time I tried using lambda in a module I got a security exception.

Cervator:
Okay, so separate from this PR then :-)

We should probably submit an issue on that separately and get it addressed.

immortius:
@MarcinSc Please do formally report issues. That is the only way that we will know (and remember) to do something about them.

@msteiger Is the casting needed for every one, or is there something in particular about the BlockTiles?

immortius:
I am investigating changing the signature for the AssetFactory interface to avoid the problem.

msteiger:
As I understand it, this issue is generic.

(Not performing) type inference for lambda expressions seems to be cause for this. I removed (bounded) wildcard expressions (similar to https://github.com/MovingBlocks/gestalt/commit/fc1cee7a297db09fcf7ade24c2c45ad22bf9baef), but it still wouldn't work. I will try again with gestalt 4.0.3 tonight.

@MarcinSc lambdas should work fine in modules (since f8350845929)

immortius:
It appeared to work with gestalt 4.0.3 for me. #1776 is a PR with the switch to the new version.

msteiger:
Unfortunately, #1776 doesn't work for me (eclipse Luna SR2). That particular error disappears, but despite the missing wildcard bounds, the types are still not inferred. For the `Atlas` class factory I get:

    The method registerCoreAssetType(Class<T>, AssetFactory<T,U>, String...) in the type 
    ModuleAwareAssetTypeManager is not applicable for the arguments (Class<Atlas>,Atlas::new,String)

immortius:
Hmm, yes. Although it has the advantage that it now fails under the standard java compilers - so we won't have an issue where it works fine everywhere but eclipse - it fails everywhere. IntelliJ doesn't report the issue except when compiling though.

So the changes mean that registerCoreAssetType now expects an AssetType where T is bound to the type of the asset type, not a subtype. So this can be handled by casting:

```java
assetTypeManager.registerCoreAssetType(Prefab.class, (AssetFactory<Prefab, PrefabData>) PojoPrefab::new, false, "prefabs");
```

Or expanded to a lambda statement

```java
assetTypeManager.registerCoreAssetType(Prefab.class, (urn, assetType, data) -> {
            return new PojoPrefab(urn, assetType, data);
        }, false, "prefabs");
```

Or possibly wrapped in a method to convert the generic params.

msteiger:
Yes, this PR is doing just that (using typecasts). I just tried with gestalt 4.0.3 and this PR and it still works fine.

PR #1797, opened 2015-06-27
Extract ThreadManager and introduce EngineStatus

immortius:
First step of the engine lifecycle redux.

This step 
* moves threading related elements of the engine into a ThreadManagementSubsystem
* introduces EngineStatus and EngineStatus subscription. This allowed the SplashScreen to be moved into the PC facade.
* tweaks the various broad states of the engine (initialised/running/shutdown/disposed)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/153/

Cervator:
Poking potentially interested parties: @msteiger @MarcinSc @prestidigitator @flo @synopia

flo:
It's better then what we had be fore. So I am up for merging it.

engine/src/main/java/org/terasology/engine/EngineStatus.java
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.engine;
+
+/**
+ * Engine Status provides the current status of the engine - what it is doing, at a higher granularity than just running. This can be used by external and internal observers
+ * to report on the state of the engine, such as splash screens/loading screens.
```
emanuele3d: Looking here on github there seems to be some newline issue here. Not a problem when browsing the javadoc via browser I imagine but would be nice to fix for people dealing with the source.

engine/src/main/java/org/terasology/engine/EngineStatus.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @author Immortius
```
emanuele3d: I thought the author tag was no longer in vogue?

engine/src/main/java/org/terasology/engine/GameEngine.java
```diff
@@ -28,6 +28,23 @@
 public interface GameEngine extends AutoCloseable {
 
     /**
+     * @return The current, fine-grained status of the engine.
+     */
+    EngineStatus getStatus();
```
emanuele3d: We would now have a getState() and a getStatus(). I think this leads to confusion. If getStatus is a fine grained version of getState() let's find a way to communicate this fact. If getStatus() is meaningful only during the "running" state (as hinted by the EngineStatus javadoc) let's make sure that is also easy to latch on.

I thought of a few options (getRunningState(), getFineState()) but perhaps getSubState() is the most appropriate. Obviously you'd have to change EngineStatus into EngineSubState and everything associated with it.

engine/src/main/java/org/terasology/engine/GameEngine.java
```diff
@@ -45,19 +62,14 @@
...
     /**
-     * @return Whether the engine has been uninitialized
+     * @return Whether the engine is running - this is true from the point run() is called to the point shutdown is complete
      */
-    boolean isUninitialized();
-
-    /**
-     * @return Whether the engine has been initialized
-     */
-    boolean isInitialized();
```
emanuele3d: Please confirm: there is no "initializing" state because no party external to the engine would be interested in this phase, right?

engine/src/main/java/org/terasology/engine/GameEngine.java
emanuele3d: Hmmmm... I re-realized getState() is about GameStates and getStatus() is about engine status. And on top of that there are the isRunning/isShuttingDown/isDisposed methods which inform about an engine state that is neither a GameState and an EngineStatus.

How about:

* getState()
* getSubState()
* getGameState()

and do away with the is*() methods?

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
```diff
@@ -131,11 +128,14 @@
...
+    private final List<EngineStatusSubscriber> statusSubscriberList = new CopyOnWriteArrayList<>();
 
-    private EngineState engineState = EngineState.UNINITIALIZED;
-
-    private final TaskMaster<Task> commonThreadPool = TaskMaster.createFIFOTaskMaster("common", MAX_NUMBER_THREADS);
+    private transient boolean initialised;
+    private transient boolean shutdownRequested;
+    private transient boolean shuttingDown;
+    private transient boolean running;
+    private transient boolean disposed;
```
emanuele3d: This block of transient booleans is interesting.

 First I wonder why they have been made transient. I seem to understand this is to avoid them being serialized. But is the engine ever serialized?

Second, they all seem to be mutually exclusive. Which is what led me months ago to make an enum out of their predecessors. Why the revert to separate booleans then?

engine/src/main/java/org/terasology/engine/subsystem/ThreadManagerSubsystem.java
```diff
@@ -0,0 +1,130 @@
...
+import org.terasology.monitoring.ThreadMonitor;
+import org.terasology.utilities.concurrency.ShutdownTask;
+import org.terasology.utilities.concurrency.Task;
+import org.terasology.utilities.concurrency.TaskMaster;
+
+import java.util.concurrent.RejectedExecutionException;
+
+/**
+ * @author Immortius
+ */
```
emanuele3d: I wonder if this class or the method submitTask() could benefit from a javadoc specifying a little what happens to tasks that have been submitted. I.e. can any assumption be made on when the task will run or guarantees of the order in which they will run? I can also see that they are run with minimum priority. That notion perhaps could also go in the javadoc. 

Finally, could perhaps the number of threads be made a public variable (still static/final though) for any interested party? Somebody might attempt to submit 3417 task at once and not quite understand why they are not started all at once. Obviously.

engine/src/main/java/org/terasology/engine/subsystem/ThreadManagerSubsystem.java
```diff
@@ -0,0 +1,130 @@
...
+            commonThreadPool.put(new Task() {
+                @Override
+                public String getName() {
+                    return name;
+                }
+
+                @Override
+                public void run() {
+                    Thread.currentThread().setPriority(Thread.MIN_PRIORITY);
+                    Thread.currentThread().setName("Engine-Task-Pool");
```
emanuele3d: So, every time a task is executed, the thread that is executing it is given minimum priority and renamed to Engine-Task-Pool. No way to set thread priority for all 16 threads in the pool? And wouldn't it be (slightly) more useful to set the name to, perhaps, "Engine-Task-Pool::<name>" where name is the name of the task?

facades/PC/src/main/java/org/terasology/engine/splash/SplashScreenCore.java
```diff
@@ -16,6 +16,8 @@
 
 package org.terasology.engine.splash;
 
+import org.terasology.engine.EngineStatusSubscriber;
```
emanuele3d: This seems to be an unused import?

engine/src/main/java/org/terasology/engine/EngineStatus.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public interface EngineStatus {
+
+    /**
+     * @return The description of the status
+     */
+    String getDefaultDescription();
```
emanuele3d: getDefaultDescription() seems to hint that there is also a NON-default description, which doesn't seem to be the case. Could this just be getDescription() then?

engine/src/main/java/org/terasology/engine/GameEngine.java
emanuele3d: P.S. in the scenario described in the comment above changeState() could be renamed to changeGameState().

engine/src/main/java/org/terasology/engine/EngineStatus.java
Cervator: Probably just still embedded in people's IDE templates :-)

engine/src/main/java/org/terasology/engine/EngineStatus.java
immortius: You mean the line length? It is within our coding standard's max line length. Unless we want to change that length.

engine/src/main/java/org/terasology/engine/EngineStatus.java
immortius: Yeah, I haven't changed my IDE settings. I don't suppose that can be set up by gradle? I'll fix that anyway.

engine/src/main/java/org/terasology/engine/EngineStatus.java
immortius: I wa thinking for language support a user could either have some sort of enum->string per language conversion or otherwise have alternate descriptions - but not as part of the interface. I'm happy to change it.

engine/src/main/java/org/terasology/engine/GameEngine.java
immortius: State will eventually be gone, don't worry too much about it (will be a little interim confusion). There will only be getStatus() and the few high level states (running, disposed, etc)

engine/src/main/java/org/terasology/engine/GameEngine.java
immortius: Correct. I also don't think the idea of initialization carrying on between shut down and disposal needs to be exposed either.

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
immortius: Sorry, should be volatile, not transient.

engine/src/main/java/org/terasology/engine/subsystem/ThreadManagerSubsystem.java
immortius: A subject for another PR - I didn't alter the methods, just moved them.

engine/src/main/java/org/terasology/engine/subsystem/ThreadManagerSubsystem.java
immortius: Again, another thread.

engine/src/main/java/org/terasology/engine/EngineStatus.java
emanuele3d: Now I understand. Disregard this comment then.

engine/src/main/java/org/terasology/engine/EngineStatus.java
emanuele3d: Indeed I thought it came with the project....

engine/src/main/java/org/terasology/engine/EngineStatus.java
emanuele3d: I guess we'd have to decide what's the scope of language support. I.e. one thing is to make sure that the user interacts with a software in his/her language. It's another thing to ensure that developers can interact with a software in their own language, i.e. for debugging purposes.

If the status changes are mostly intended for developing/debugging purposes perhaps we can stick to English-only unless we want to go multilanguage at a deeper level. If the statuses are also intended to be meaningful for the user then I'd be ok with it I guess.

engine/src/main/java/org/terasology/engine/GameEngine.java
emanuele3d: I was just thinking that perhaps handling the game state here now felt a little out of place.  I'm now curious where you are gonna put it but I'll hold my tongue and wait (im)patiently...=D

Ultimately, ok for the interim confusion. =)

engine/src/main/java/org/terasology/engine/GameEngine.java
emanuele3d: Ok.

engine/src/main/java/org/terasology/engine/subsystem/ThreadManagerSubsystem.java
emanuele3d: Hmmmm... I feel you are missing the opportunity to do the right thing here. The extraction of the thread-related functionality is the first point of this PR. It is indeed a small thing code-wise but it is potentially a pretty important system application-wise. Unless you have plans to rework on this functionality and expand it or change it dramatically, the perfect time to add those javadocs would be now that it has been touched.

engine/src/main/java/org/terasology/engine/subsystem/ThreadManagerSubsystem.java
emanuele3d: Another approach to this would be to revert the extraction and deal with it in a separate PR, including adding  javadocs and potentially the features suggested here.

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
emanuele3d: I just went through the lifecycle. 

Let me get this straight: basically the engine is either "running" or "disposed". Within the context of "running" it can be "initialized", a shutdown may have been requested or it is actually in the process of shutting down. Then within the context of "disposed" everything else is false.

If I'm correct, I'm wondering if you could, for starters, remove the "initialized" variable. You use it only to avoid re-initialization but if that's the case could we perhaps prevent the re-execution of run()?

The other thing is, I appreciate the sophistication of having a shutdownRequested and a shuttingDown. But from the code having just "shuttingDown" set by shutdown() and checked by the main loop seem it would work just as well. In this context (eliminated shutdownRequested and initialized), the states "running", "shuttingDown" and "disposed" seem mutually exclusive and could go back into an enum, with the addition of an initial "NotYetStarted" value. This would make the whole lifecycle much easier to define and keep in mind.

Or am I missing something?

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
```diff
@@ -501,31 +529,24 @@ private void mainLoop() {
...
-        // In all other circumstances the EngineState is already set to
-        // INITIALIZED by the time the flow gets here.
-        engineState = EngineState.INITIALIZED;
+        changeStatus(StandardGameStatus.SHUTTING_DOWN);
     }
 
     private void cleanup() {
         logger.info("Shutting down Terasology...");
+        shuttingDown = true;
+        changeStatus(StandardGameStatus.SHUTTING_DOWN);
```
emanuele3d: It seems to me that this line is perhaps redundant with the identical line at the end of mainLoop().

engine/src/main/java/org/terasology/engine/TerasologyEngineStatus.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package org.terasology.engine;
+
+/**
+ * @author Immortius
+ */
+public enum TerasologyEngineStatus implements EngineStatus {
+
+    LOADING_CONFIG("Loading config..."),
+    PREPPING_SUBSYSTEMS("Preparing Subsystems..."),
```
emanuele3d: Cute, but I'd suggest "PREPARING_SUBSYSTEMS".

engine/src/main/java/org/terasology/engine/TerasologyEngineStatus.java
```diff
@@ -0,0 +1,51 @@
...
+    INITIALIZING_ASSET_MANAGEMENT("Initializing Asset Management..."),
+    INITIALIZING_SUBSYSTEMS("Initializing Subsystems..."),
+    INITIALIZING_MODULE_MANAGER("Initializing Module Management..."),
+    INITIALIZING_REFLECTION("Initializing high performance reflection..."),
```
emanuele3d: Is there a "low performance reflection"? I'm just wondering if there is an interest in knowing this is a particularly high performance thing.

engine/src/main/java/org/terasology/engine/EngineStatus.java
Cervator: We do have a few IDE tweaks like the copyright template. No Javadoc template yet though. Likely goes with a general overhaul of Checkstyle config for Javadoc with updated templates that'll match our conventions.

engine/src/main/java/org/terasology/engine/TerasologyEngineStatus.java
Cervator: I first mistook this as water style reflection, of which we have multiple types. Then I got into thinking about making a philosophical reflection joke, and realized my mistake.

This comment has no purpose.

engine/src/main/java/org/terasology/engine/EngineStatus.java
immortius: Status changes are intended to potentially be displayed to users - both in the splash screen (already implemented) and other load screens.

I probably will just make it english only, and remove default from the name. Translation can either use the common technique of using the english string to map to a language appropriate string, or can use the status object to map to a language appropriate string. Or the strings can come from language property files eventually. Either way there won't be any other methods so no need to have default in the name.

engine/src/main/java/org/terasology/engine/GameEngine.java
immortius: Essentially what is the difference between being in a menu and being in game? There's no world and no player in the menu... although there is a player who is interacting with the menu, just no character they control. And having the world (or worlds) driven by entities is already in our todo. So in the end there are really only two states - loading (switching environment and everything that goes with that) and loaded (into a menu, or into a game). That is where I plan to head. It may be a twisty road.

engine/src/main/java/org/terasology/engine/TerasologyEngineStatus.java
immortius: I struggled a bit with this name. Essentially this stage is setting up the systems that do runtime code generation to access fields/getter and setters, as a replacement for reflection. I didn't want to simply call it reflection due to that being associated with poor performance and not being technically correct.

Perhaps it could be called something like "Initializing Code Generation Magic" or something, dunno. Maybe "Initializating persistence framework" would do, though not quite correct.

engine/src/main/java/org/terasology/engine/subsystem/ThreadManagerSubsystem.java
immortius: There's already another PR dealing with the threading system, I don't want to go changing things until that is done or discarded.

engine/src/main/java/org/terasology/engine/subsystem/ThreadManagerSubsystem.java
immortius: There is already another PR with threading support changes. This change is focusing on moving threading handling out of engine and the engine api. Reworking thread management is scope creep.

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
immortius: You are right that it can be simplified, but I think I'll go further.

The only state is running - true or false. No initialised, no disposed. Part of shutdown will be disposal. Once run the engine can be rerun, or discarded.

Request shutdown will remain - it isn't a state, just a flag to trigger shutdown. Initialised/shutting down will be removed.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/160/

engine/src/main/java/org/terasology/engine/TerasologyEngineStatus.java
Cervator: "Reflecting on the value of objects" - philosophical, yet relevant ;-)

engine/src/main/java/org/terasology/engine/subsystem/ThreadManagerSubsystem.java
emanuele3d: As you wish.

engine/src/main/java/org/terasology/engine/GameEngine.java
emanuele3d: Understood. Thank you for explaining this.

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
emanuele3d: Sounds good!

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/161/

**Build Log**
*last 5 lines*

 ```
[...truncated 91 lines...]
	at hudson.model.Executor.run(Executor.java:240)
[ANALYSIS-COLLECTOR] Skipping publisher since build result is FAILURE
Setting status of 12edab53aa30ec79088403c8ec73576ddd026952 to FAILURE with url http://jenkins.terasology.org/job/TerasologyPRs/161/ and message: Merged build finished.
```

**Comment by GooeyHub** [comment]
http://jenkins.terasology.org/job/TerasologyPRs/162/

**Build Log**
*last 5 lines*

 ```
[...truncated 2427 lines...]
[ANALYSIS-COLLECTOR] Computing warning deltas based on reference build #160
Setting status of 58892c5510b6e1bfddf956252a61a37291c4bcfd to FAILURE with url http://jenkins.terasology.org/job/TerasologyPRs/162/ and message: Merged build finished.
```

Cervator:
I wish we could distinguish between unstable (test failures) and build failure in Jenkins on these PR tests (at least the log snippet gives a clue). The "failure" is from a single unit test seemingly breaking

http://jenkins.terasology.org/job/TerasologyPRs/162/testReport/junit/org.terasology.identity/CertificateTests/checkGeneratedSelfSignedCertificateValid/

Might fail in the main job too, still running. Probably an easy fix :-)

Thanks for all the work and review!

PR #1800, opened 2015-06-27
Rework join game dialog

### LINKED ISSUE #1786: Rename "Online" to "Listed" on the server join screen?
After merging #1784 I got to thinking - if the tab says "Online" yet I'm looking through servers that are mostly offline that could be read kinda funny.

Maybe a better option would be to name the tab "Listed" - as in, servers officially listed by us via the listing service. They may be running or not at any given time (online/offline)

@msteiger what do you think?

msteiger:
This PR deals with a few related issues around the join game dialog:

* The entry "Online" was renamed to "Listed" (now at the left and default) - Fixes #1786 
* Servers are now retrieved as a stream (plus an artificially delay to achieve an animation)
* Add the "Owner" field - editable for custom entries
* Evaluate the "active" field. Inactive servers are not listed.
* Add `extendLast` layout hint for `ColumnLayout`: the last entry takes all the excess space.
* The module list is no longer truncated and scrollable instead

image

### DIFF
--- a/engine/src/main/java/org/terasology/config/ServerInfo.java
+++ b/engine/src/main/java/org/terasology/config/ServerInfo.java
@@ -16,6 +16,8 @@
 
 package org.terasology.config;
 
+import com.google.common.base.Preconditions;
+
 
 /**
  * @author Immortius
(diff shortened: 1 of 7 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/155/

Cervator:
Looking good :-)

Should we leave @Calinou's entry as active even knowing it isn't? For being able to test a for-sure offline server in the list.

Issue #1777, opened 2015-06-17
Incomplete reflection cache files for 0.54 engine/Core jars


msteiger:
After upgrading to engine 0.54.0, some of the `FacetLayers` were no longer discovered in WorldViewer. I noticed that the `reflections.cache` file in the snapshot jars specified only some implementations.

Example: [engine-0.54.0-6](http://artifactory.terasology.org/artifactory/terasology-snapshot-local/org/terasology/engine/engine/0.54.0-SNAPSHOT/engine-0.54.0-20150617.090056-6.jar) reflections:

According to the cache file, the only subtypes of `AbstractFacetLayer` are `NominalFacetLayer` and `SeaLevelFacetLayer`, but in fact there are more - even in the same package (e.g. `FieldFacetLayer`).

How is that possible?

immortius:
This is the issue I've been struggling with. It only happens in a gradle
context - the same code for gathering reflections works fine external to
gradle. Some of the symptoms are really odd, like it discovering inner
classes but not the outer class for GLSLShaderFormat.

Perhaps reflections caching needs to be in a gradle plugin, or run in a
separate jvm?

msteiger:
Just a wild guess: Could it be related to gradle now compiling projects in parallel? I don't see a direct link, but maybe that's still buggy. Commit https://github.com/MovingBlocks/Terasology/commit/6c1fe4b5522b618458b1c9442797c49f3e0a1fe7 enabled it. The [Gradle docs](https://docs.gradle.org/current/userguide/multi_project_builds.html) "56.8 Parallel project execution" gives some background info.

immortius:
It doesn't look like it, but if the reflections caching is happening concurrently with compilation that could explain it. Although I think it is too consistent on what classes are and aren't included.

Cervator:
Actually that commit is to a template file, it isn't being used. I manually copy it to my local workspace to enable that kind of stuff and override destinations if I'm testing Artifactory type stuff.

Jenkins should still be compiling old school. We probably *should* actually consider changing that around a little sometime to optimize Gradle.

Would it be something to try running the build and the caching in separate Gradle executions? Like first compile the code, finish that Gradle execution, then start another to cache, then dist?

immortius:
Any caching in gradle, even if you do nothing but caching after previously
compiling, fails to find all the classes. Would need to debug in to work
out why.

PR #1801, opened 2015-06-28
Fix for #1777

immortius:
Fix for #1777
* Updated gradle build to use reflections-0.9.9 and the new dom4j dependency
* Tweaks to engine cacheReflections because it randomly started thinking it was always up-to-date on me.
* Core and CoreSampleGamePlay also updated to use the new reflections, this will need to be pushed out to other modules and possibly improved in structure.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,8 @@ buildscript {
         classpath 'org.ajoberstar:gradle-git:0.6.3'
 
         // Needed for caching reflected data during builds
-        classpath 'org.reflections:reflections:0.9.9-RC1'
+        classpath 'org.reflections:reflections:0.9.9'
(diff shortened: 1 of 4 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/156/

msteiger:
Awesome! I had this one on my TODO-List for today. Tests out fine on my machine. WorldViewer works again, too.

Cervator:
Excellent. Thank you for taking the time to figure this out :-)

The modules should be fine, since they build the same way (and with the same build.gradle) as Core. Might need to wait for some to get rebuilt, but TTA for instance works fine via binary now (at least the very basics)

PR #1807, opened 2015-07-03
Fix for certificate test failure.

immortius:
There was a 1 in 256 chance of certificates failing to verify as signed/self signed due to incorrect conversion between BigInteger and byte[].


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/163/

msteiger:
I assume that this is because the random number generator could create a number that starts with a leading 0x00 byte. This would lead to a shorter array and thus to authentication failure.

A short comment wouldn't hurt, but the code looks good to me.

Great find, btw.

immortius:
That is correct. I should have spelled it out further in the initial PR comment.

PR #1812, opened 2015-07-05
Remove CoreRegistry usage: Part V

flo:
Another pull request to reduce CoreRegistry usage.

It also fixes the PojoPrefabManager#exists method which was broken. I am not sure if that was causing any noticable bugs.

The place where the prefab formats get now registered isn't ideal, but it's the only place where it currently can be done. See discusion with @immortius in https://github.com/MovingBlocks/gestalt/issues/35 for more information.

Also noteworthy about this pull request is that the patch 465cd39e7aa157c7e4bc4bb65b84a7f9d3bc8447 moves the `applyBinds` call from `RegisterInputSystem#step` to `InputSystem#initialise`. That  way the `InputSystem` is initialized when the `applyBinds` method gets called. 

I also think it might be a good idea to move the  `applyBinds` logic from BindsConfig to InputSystem as the former is to me more a data holding object. Also we might want to introduce a InputBindingsChangedEvent that gets procssed by InputSystem.

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/prefab/internal/PojoPrefabManager.java
+++ b/engine/src/main/java/org/terasology/entitySystem/prefab/internal/PojoPrefabManager.java
@@ -15,13 +15,14 @@
  */
 package org.terasology.entitySystem.prefab.internal;
 
+import com.google.api.client.repackaged.com.google.common.base.Strings;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
-import org.terasology.asset.Assets;
 import org.terasology.assets.management.AssetManager;
+import org.terasology.context.Context;
 import org.terasology.entitySystem.Component;
 import org.terasology.entitySystem.prefab.Prefab;
 import org.terasology.entitySystem.prefab.PrefabManager;
-import org.terasology.registry.CoreRegistry;
 
 import java.util.Collection;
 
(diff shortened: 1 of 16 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/166/

**Build Log**
*last 5 lines*

 ```
[...truncated 2102 lines...]
[ANALYSIS-COLLECTOR] Computing warning deltas based on reference build #165
Setting status of 465cd39e7aa157c7e4bc4bb65b84a7f9d3bc8447 to FAILURE with url http://jenkins.terasology.org/job/TerasologyPRs/166/ and message: Merged build finished.
```

**Comment by GooeyHub** [comment]
http://jenkins.terasology.org/job/TerasologyPRs/167/

File: engine/src/main/java/org/terasology/entitySystem/prefab/internal/PojoPrefabManager.java [hunk 1]
```diff
@@ -34,13 +35,21 @@
...
+
+
     /**
      * {@inheritDoc}
      */
     @Override
     public Prefab getPrefab(String name) {
         if (!name.isEmpty()) {
-            return Assets.getPrefab(name).orElse(null);
+            Preconditions.checkArgument(!Strings.isNullOrEmpty(name));
```
immortius: Uh, you would put this inside an ```if(!name.isEmpty())``` block? I would have just put ```Preconditions.checkNotNull``` on the outside, since name should not be null.

engine/src/main/java/org/terasology/entitySystem/prefab/internal/PojoPrefabManager.java
```diff
@@ -51,7 +60,8 @@ public Prefab getPrefab(String name) {
      */
     @Override
     public boolean exists(String name) {
-        return Assets.getPrefab(name) != null;
+        Preconditions.checkArgument(!Strings.isNullOrEmpty(name));
```
immortius: I would argue it is ok if name is empty, so the contract can be loosened a little.

immortius:
Made a couple of small comments, the rest is good :)

flo:
Thanks for the review. I made now both method handle null and empty names without failure for consistency.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/168/

immortius:
Well, ok, I guess. I think null should actually be a failure - if null was allowed, then the methods would have taken ```Optional<String>``` rather than ```String``` as a parameter. But not an issue really. PrefabManager should probably be deprecated anyway, since it doesn't add anything over AssetManager.

PR #1809, opened 2015-07-03
World Atlas auto-reloading

immortius:
Implemented auto reloading of the world atlas texture after a block tile is reloaded.

PR #1814, opened 2015-07-07
Codecity-like visualization of a project as a Terasology world

seanorar:
merge de Coloring


GooeyHub:

Cervator:
Hi guys, nice, are you all done with your project then? Did it work out alright? Any instructions for how to test it? :-)

seanorar:
Sorry, this pull request was a mistake. We apologize and please don't pull, thanks xD

Cervator:
Awww, I was getting all excited! :D

How are you doing? Anything we can help with? Any questions?

sking32:
Hi, we are having some trouble with blocks which can be recolored after placed, we can already place colored blocks but not modify the color. Do you know how we could solve this problem? Thank you very much for your help

Cervator:
Sure. Generally you don't want to try changing the texture on an existing block in the world, although there are some ways to dynamically modify (like icons in Throughout the Ages / JoshariasSurvival - ping @Josharias & @MarcinSc) or reload (#1809) existing textures.

Instead you usually just make multiple unique blocks with the looks you want, then replace the blocks entirely. So if you have a "Red" block and want to change that to a "Green" block you simply use the BlockManager to place a new green block in that location.

If you can point to a code example we'll be able to tell more exactly what you're trying to do and what way would work the best :-)

Issue #1761, opened 2015-06-07
Chest inventory screen (and maybe others?) closed with 'ESC' causes need for 'e' twice to open


Cervator:
I think somebody else noticed this very recently but forgot who / where. Ran into it myself while testing and checked with @immortius who identified it as being from the use of `esc` to close a screen. May relate to NUI interactions with the asset system.

Not very serious so it can wait till a convenient time after the big PR is merged.

Cervator:
Closing as fixed by #1825

PR #1825, opened 2015-07-09
Give NUIManager a better fitting life cycle and context instance.

flo:
This pull request gives NUIManager a better fitting life cycle and context instance.

The NUIManager is no longer in the engine context.

Where previously it just got "refreshed" now a new instance gets created.

Namely the first NUIManager like the first sub context for the duration
of the main menu.

When the starts loading a new NUIManager gets created that lifes again
like it's owning Context till the end of the game.

This patch fixes the issue that caused interations require two activations
when a window got closed with escape (#1761).

The first patch moves the mouse grap triggering from LwjglInput to NUIManager so that LwjglInput does not require a  NUIManager instance (which is no longer in the engine context).

@immortius I hope this change is acceptable for you and hopefully you even like it :)

### DIFF
--- a/engine/src/main/java/org/terasology/engine/modes/loadProcesses/InitialiseGraphics.java
+++ b/engine/src/main/java/org/terasology/engine/modes/loadProcesses/InitialiseGraphics.java
@@ -22,6 +22,7 @@ import org.terasology.engine.TerasologyConstants;
 import org.terasology.math.geom.Vector4f;
 import org.terasology.naming.Name;
 import org.terasology.rendering.nui.NUIManager;
+import org.terasology.rendering.nui.internal.CanvasRenderer;
 import org.terasology.rendering.nui.internal.NUIManagerInternal;
 import org.terasology.rendering.primitives.Tessellator;
 import org.terasology.rendering.primitives.TessellatorHelper;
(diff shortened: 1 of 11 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/174/

flo:
I just noticed that the refreshWidgetsLibrary method is scanning the modules. I changed it back so that it gets called again after the modules got loaded.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/175/

engine/src/main/java/org/terasology/engine/modes/loadProcesses/InitialiseGraphics.java
```diff
@@ -44,6 +44,7 @@ public String getMessage() {
 
     @Override
     public boolean step() {
+        // Refresh widget library after modules got laoded:
```
immortius: Not quite how loaded is spelt.

engine/src/main/java/org/terasology/engine/subsystem/lwjgl/LwjglInput.java
```diff
@@ -61,17 +58,10 @@ public void postInitialise(Context context) {
 
     @Override
     public void preUpdate(GameState currentState, float delta) {
-        NUIManager nuiManager = context.get(NUIManager.class);
```
immortius: Good idea.

immortius:
I'm not totally enamored by the change, but that is mainly because it involves the states and I would like to get rid of those. But still an improvement.

I guess I'm still thinking about how NUIManager should fit into the lifecycle, particularly because there are UI screens that need to live across loading... Or possible not, hmm.

PR #1831, opened 2015-07-12
Avoid causing crashes when a component has a null string.

Josharias:
This needs some discussion. 

I would like to be a little more flexible with components and strings.  Currently if you let a null string be sent to the serializer for saving a world, it crashes the game.  It would lower the bar to entry for new module developers if we allowed for null values without crashes, less things to know and code around.

Could we be a little more lenient in these cases?  Is there any other considerations to making this kind of change?  Possibly this could be extended to other types as well?

### DIFF
--- a/engine/src/main/java/org/terasology/persistence/typeHandling/StringRepresentationTypeHandler.java
+++ b/engine/src/main/java/org/terasology/persistence/typeHandling/StringRepresentationTypeHandler.java
@@ -31,12 +31,20 @@ public abstract class StringRepresentationTypeHandler<T> implements TypeHandler<
 
     @Override
     public PersistedData serialize(T value, SerializationContext context) {
-        return context.create(getAsString(value));
+        String stringValue = getAsString(value);
+        if (stringValue == null) {
+            return context.createNull();
+        } else {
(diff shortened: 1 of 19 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/179/

Cervator:
Bump - @immortius @flo @MarcinSc anybody else interested?

immortius:
I generally agree, but the only way to ensure you don't have crashes is not to have null values, ultimately. If something is intended to be null, I'ld suggest ```Optional``` should be used instead.

That said, I would agree with either what you've done, or having null strings serialize/deserialize to empty string. Null is safer in case existing code depends on it.

Josharias:
Should I apply the same null allowance to the other type handlers as well?  There are some that handle null well,  and some that dont.

flo:
I think making the other type handlers null save is a good idea.

Josharias:
Here is the more complete version touching other type handlers that could also crash when null.  Did I miss anything?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/181/

PR #1833, opened 2015-07-13
Improvements to the BlockRegion system

Josharias:
- Allow block regions to drop all the blocks in the region instead of the single block that they act like (opt in).
- Automatically add health to the region entity similar to the ActAsBlock block would have.
- Simplify the ActAsBlock CreateBlockDropsEvent handler into a single event handler.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/regions/BlockRegionSystem.java
+++ b/engine/src/main/java/org/terasology/world/block/regions/BlockRegionSystem.java
@@ -16,6 +16,7 @@
 package org.terasology.world.block.regions;
 
 import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.event.EventPriority;
 import org.terasology.entitySystem.event.ReceiveEvent;
 import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.entitySystem.systems.RegisterMode;
@@ -38,7 +39,8 @@ public class BlockRegionSystem extends BaseComponentSystem {
     @In
     private BlockManager blockManager;
 
-    @ReceiveEvent
+    // trivial priority so that all other logic can happen to the region before erasing the blocks in the region
+    @ReceiveEvent(priority = EventPriority.PRIORITY_TRIVIAL)
     public void onDestroyed(DoDestroyEvent event, EntityRef entity, BlockRegionComponent blockRegion) {
         for (Vector3i blockPosition : blockRegion.region) {
             worldProvider.setBlock(blockPosition, blockManager.getBlock(BlockManager.AIR_ID));
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
@@ -107,18 +107,26 @@ public class BlockEntitySystem extends BaseComponentSystem {
     }
 
     @ReceiveEvent(priority = EventPriority.PRIORITY_TRIVIAL)
-    public void defaultDropsHandling(CreateBlockDropsEvent event, EntityRef entity, ActAsBlockComponent blockComponent, LocationComponent locationComp) {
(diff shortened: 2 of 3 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/180/

Cervator:
This made me realize I am not actually sure what BlockRegions do - the Javadoc on the Component isn't exactly super enlightening, and checking a few hits found no implementation Javadoc either :-)

Able to add some in while touching it anyway? 

I figure it could relate to stuff like MultiBlocks or situations like chopping down a tree in TTA where you don't want every single block in the tree to necessarily drop as singles?

Seems a straight forward enough change, just hunting further knowledge :-)

engine/src/main/java/org/terasology/world/block/regions/BlockRegionSystem.java
```diff
@@ -38,10 +43,25 @@
...
+    // trivial priority so that all other logic can happen to the region before erasing the blocks in the region
+    @ReceiveEvent(priority = EventPriority.PRIORITY_TRIVIAL)
     public void onDestroyed(DoDestroyEvent event, EntityRef entity, BlockRegionComponent blockRegion) {
         for (Vector3i blockPosition : blockRegion.region) {
             worldProvider.setBlock(blockPosition, blockManager.getBlock(BlockManager.AIR_ID));
         }
     }
+
+    @ReceiveEvent
+    public void onActivatedActAsBlockAddHealth(OnActivatedComponent event, EntityRef entity, BlockRegionComponent blockRegion, ActAsBlockComponent actAsBlock) {
```
immortius: I don't really agree with this. If the ActAsBlock entity should have health, the entity should just be given a Health component I feel. Health is too tied to blocks at the moment as is.

Josharias: This is to help a module developer maintain consistency for block health values.  Currently you would have to ensure that both the block and the region's entity are synchronized to get a similar feel.  This would get even more tricky if a block was overridden by another module.

I agree that block health really needs to get pulled out of engine to avoid polluting the engine code with this kind of gameplay specific code.

MarcinSc:
My suggestion is to mark BlockRegion as deprecated. The fact that these entities are always stored in memory, because they do not have a single chunk to be assigned to, is a no-go for me. This is why I created the MultiBlock2 to avoid that issue altogether.

Josharias:
Hrm... interesting observation.  I would have thought that these entities would have magic sprinkled in because they have a LocationComponent.  I havent looked at that part of the codebase in a while, will look now.

Josharias:
It *seems* like because it has a LocationComponent, it will stay relevant with a chunk.  

MarcinSc, I suppose your comment is about when the BlockRegion straddles a loaded chunk and an unloaded chunk?  So the only way to avoid complications is to have the entity always stay relevant?

MarcinSc:
Correct

Josharias:
MarcinSc, is your multiBlock2 code multiplayer ready?

MarcinSc:
I have not tested it in multiplayer yet.

Josharias:
Darn.  I would definitely be tempted to leave the BlockRegion behind if there was a better system that worked in multiplayer.

MarcinSc:
I didn't say it doesn't work - I said I did not tested it, but I expect it would work. :)

Cervator:
Bump. @Josharias were you looking at other implementation options for this? PR still valid or should we close with a later redo?

Josharias:
Still on the burner.  Just out of time.  The plan is to take out the Immortius disagreement part, and then roll with the rest. Thanks for the patience.

Josharias:
Ready for acceptance.  I have taken out the debatable part of automatically adding health.

To test:
1. remove ```BlockDropGrammar``` from the door.prefab
2. add ```"dropBlocksInRegion": true``` to the ```ActAsBlock``` component on door.prefab
3. place and break the door,  it will drop both the top and bottom pieces.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/223/

Cervator:
Looks good! Thanks for the testing instructions :-)

PR #1845, opened 2015-07-25
Fix #1832

### LINKED ISSUE #1832: ViewDistance is not able to be changed after recent Context changes (even crashes the game)
This binding in the video screen no longer works because there is no setter method like it expects...
https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/VideoSettingsScreen.java#L74

Probably is on someones todo list already. Making sure it was mentioned.

flo:
This pull request fixes 2 issues with bindings that got mentioned in #1832:
* The instant crash was caused by the movment of the fullscreen property to a differnt class and was an easy fix.
* The view distance was broken, because the setter had now an additional argument that was necessary to perform some logic. This pull request lets the binding send an new ViewDistanceChangeRequest that gets procssed still on the client side. The hanlder of the ViewDistanceChangeRequest event then sends a ViewDistanceChangeEvent to the server.

Any logic that wants to change the view distance can now simply send a ViewDistanceChangeRequest to the local player's client entity. The DebugControlSystem also uses that event to change the view distance when you press the keys `[End]` or `[Home]`.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/195/

engine/src/main/java/org/terasology/rendering/world/viewDistance/ViewDistance.java
```diff
@@ -13,14 +13,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.terasology.rendering.world;
+package org.terasology.rendering.world.viewDistance;
```
immortius: Hmm, this probably should be called WorldViewDistance or something for that matter. *shrug*

Cervator:
Awesome! Thanks for this :-)

PR #1863, opened 2015-07-28
Fix for Console input

immortius:
* Fixes missing key event consumption resulting in the player moving about while typing in the console
* Also reorders add component event order to reduce errors caused by removing a component during its add event.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/202/

Cervator:
Tests out - at least one scenario where the player could move is now handled right. However, there is another: if you click in the console (not in the input bar, above it) the blinking cursor moves up there and you can again move around in the world (although not jump)

Incidentally this fixed the sapling/seed planting in TTA, was that intended or a happy coincidence? :-)

Thanks!

Issue #207, opened 2012-03-10
Every path containd in variables shoudl end with '/'


ghost:
Every path containd in variables shoudl end with '/' we have places littered with '/' concatenations, really ugly.

PR #1869, opened 2015-07-31
More engine lifecycle rework

immortius:
Moved most of the remaining logic that shouldn't be in engine into subsystems, and it is now handled via generic lifecycle events.

Some particular focus on time-related behavior. Added support for altering the rate of game time. Fixed hibernation when alt-tabbed from the game.

### DIFF
--- a/engine-tests/src/main/java/org/terasology/TerasologyTestingEnvironment.java
+++ b/engine-tests/src/main/java/org/terasology/TerasologyTestingEnvironment.java
@@ -36,6 +36,7 @@ import org.terasology.engine.modes.loadProcesses.LoadPrefabs;
 import org.terasology.engine.module.ModuleManager;
 import org.terasology.engine.paths.PathManager;
 import org.terasology.entitySystem.entity.internal.EngineEntityManager;
+import org.terasology.game.Game;
 import org.terasology.naming.Name;
 import org.terasology.network.NetworkSystem;
 import org.terasology.network.internal.NetworkSystemImpl;
@@ -98,6 +99,7 @@ public abstract class TerasologyTestingEnvironment {
         mockTime = mock(EngineTime.class);
         context.put(Time.class, mockTime);
         NetworkSystemImpl networkSystem = new NetworkSystemImpl(mockTime, context);
+        context.put(Game.class, new Game());
         context.put(NetworkSystem.class, networkSystem);
         EntitySystemSetupUtil.addReflectionBasedLibraries(context);
         EntitySystemSetupUtil.addEntityManagementRelatedClasses(context);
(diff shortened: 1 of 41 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/208/

**Build Log**
*last 5 lines*

 ```
[...truncated 2057 lines...]
[ANALYSIS-COLLECTOR] Computing warning deltas based on reference build #207
Setting status of 3d789029dfb88e97a247a5037ef2d55124274ac3 to FAILURE with url http://jenkins.terasology.org/job/TerasologyPRs/208/ and message: Merged build finished.
```

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/209/

flo:
Overall nice TerasologyEngine simplifiation/purification.

flo:
I just tried to build with it, it breaks the Pathfinding module.

flo:
I noticed the engien version is now 0.54.1-SNAPSHOT, so we would have to increment it to 0.55.0-SNAPSHOT

flo:
hmm the modules haven't been released as stable version.

When we increment the engine version we would need a script or build task or something to do an automatic maxVersion incrment for all modules. @Cervator is something ready?

msteiger:
One option would be to set maxVersion to `1.0.0` to avoid this issue without the need to update the value on every engine release. Or is there a reason not to do that?

immortius:
Yes. If the maxVersion of a module was 1.0.0 and there is a breaking change that breaks it, then it won't be correctly marked as incompatible.

Cervator:
Related forum thread - http://forum.terasology.org/threads/more-version-increments-to-avoid-dependency-issues.1270 :-)

Lots of wonderful ideas on how to handle everything, less so on the time availability to implement. Still on my short-list

Also, time dilation? Neat! Now we just need that IRL :-) I take it we in theory could push that to a neato EVE-like lag handling technique if needed.

flo:
When the game does not have the focus when the loading finishes then the game is permanently paused. Might this be a bug introduced by the refactoring?

immortius:
Hibernation restores the pause state from before hibernation began. This is perhaps incorrect. It might be best for hibernation to not pause the game, not sure.

Cervator:
I'm personally a fan of being able to leave the window alone but active while working on other stuff. And wouldn't that have to be the case if connected to a server anyway? The client won't be able to pause time then.

But maybe I'm not entirely on top of what gets paused

immortius:
The network system disables hibernation explicitly. But yeah, I think I'll just change it to do the frame rate limiting and remove the auto-pausing.

PR #1875, opened 2015-08-02
Module download

msteiger:
**Internal**

TODO: support download

PR #1884, opened 2015-08-12
Add support for downloading modules at runtime.

### LINKED ISSUE #1754: Resolve and download modules at runtime
I think that it would be great if Terasology could resolve and download wanted or required artifacts (from Artifactory?) automatically.

- [x] **TODO:** Make Artifactory list all module jars

* The GAVC (group, artifact, version, classifier) search returns all file types (pom, jar, asc)
* The GAVC search without classifier finds all classifiers instead of none (javadoc, sources)
* By default, [Artifactory lists only 500 search results](http://artifactory.terasology.org/artifactory/api/search/gavc?g=org.terasology.modules&repos=terasology-release-local,terasology-snapshot-local), but this can be increased.

- [x] **TODO:** Retrieve module information for each entry
* One way would be extracting the module.txt from each jar file
* Another would be encoding the important data in the `pom.xml` file: minVersion, maxVersion (maybe name, description and authors)

- [x] **TODO:** Create a ModuleRegistry based on the entry data
* Maybe [master-server](https://github.com/MovingBlocks/master-server) could provide a cached and JSON-ized list of `ModuleMetaInfo`.

- [x] **TODO:** Resolve deps. and download artifacts
* Maybe using a Maven plugin
* Probably easier through gestalt-module
* Downloading must be done into a folder with write-access

- [ ] **TODO:** Search the target download folder for modules on startup and after downloads

### LINKED ISSUE #1787: Add legend for colors on the module page
This could be either squeezed in somewhere in the UI as a general note:

* Blue = active gameplay template
* Yellow = required dependency
* Green = user-activated module (may or may not be a dependency)
* Red = missing dependency

Or noted in the display for the actively selected module. This might be preferable to keep things compact and could yield more info like what something is a dependency of:

* (Green:) This module is a required dependency of module X. It has also been activated manually [and will be retained if the gameplay template is changed]

Or maybe even listed in a tooltip when hovering over a module in the list.

*Tricky part:* The square bracket bit was the original meaning but isn't actually accurate anymore. Changing the gameplay template now resets the module selection. So I'm not sure there is a purpose to having an activate/deactivate button on a dependency right now. That might reduce confusion a bit too. This has come up before but I couldn't find prior notes with a super quick search.

That part may get into #1754 and ModuleManager somewhat. But in any case it would be nice with a simple display of the basic meaning, we can tweak it later as the meanings change.

msteiger:
The information about modules based on their `module.txt` is retrieved from [meta.terasology.org](http://meta.terasology.org). On demand, modules can be downloaded (from Artifactory) and installed without need to restart the game.

image

Fixes #1754 and fixes #1787 (supercedes #1875).

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/RemoteModuleExtension.java
+++ b/engine/src/main/java/org/terasology/engine/module/RemoteModuleExtension.java
@@ -23,7 +23,7 @@ import org.terasology.module.ModuleMetadata;
 
 /**
  * A set of module extensions for remote modules.
- * NOTE: this is copy&paste from master-server.
+ * NOTE: this is copy&paste from meta-server.
  */
 public enum RemoteModuleExtension {
 
--- a/engine/src/main/java/org/terasology/engine/module/RemoteModuleExtension.java
+++ b/engine/src/main/java/org/terasology/engine/module/RemoteModuleExtension.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2015 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.terasology.engine.module;
+
+import java.net.URL;
+import java.util.Date;
+
+import org.terasology.module.ModuleMetadata;
+
+/**
+ * A set of module extensions for remote modules.
+ * NOTE: this is copy&paste from master-server.
+ */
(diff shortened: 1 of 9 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/224/

Cervator:
Awesome looking! Can't wait to play with this some :-)

It always defaults to the latest available I take it? Hitting Artifactory for now isn't a problem (although it occasionally runs out of memory - I've got alerting enabled for catching that), maybe we can play with Bintray for later use away from our infrastructure.

@immortius any comments on the architecture side?

immortius:
All looks great to me.

Cervator:
Works great! And latest Polyworld looks better than ever :D

Brings one question to mind though: how do we handle old modules that exist in Artifactory but have been removed from the standard lineup? There are a few like TerraTech, LandsOfTerra, and Rails that'll outright crash or make no sense.

Should we have the meta server consider the Omega list? If a module is available online but not in Omega mark it a darker gray than "not downloaded yet" to flag for "Caution" and have the game disclaim as much in the description box?

It would also be nice with a dependency listing in the description. Grabbed FlightModeToggleButton and was perplexed why it wasn't satisfied until I remembered HUDToggleButtons.

I know a lot of that stuff will come over time and will merge this now since it is an awesome step forward and we can tweak a bit more before next release still :-)

Issue #1886, opened 2015-08-14
125 chunks need to be loaded just to make one visible?


flo:
A chunk gets only scheduled for light merging by RemoteChunkProvider when all chunks around it are loaded:
```
    private boolean makeChunkAvailable(final Chunk chunk) {
        for (Vector3i pos : Region3i.createFromCenterExtents(chunk.getPosition(), 1)) {
            if (chunkCache.get(pos) == null) {
                return false;
            }
        }

        lightMerger.beginMerge(chunk, chunk);
        return true;
    }
```

Chunks get marked as ready when the light merging is done in RemoteChunkProvider:
```
    public void completeUpdate() {
        Chunk chunk = lightMerger.completeMerge();
        if (chunk != null) {
            chunk.markReady();
            updateAdjacentChunksReadyFieldOf(chunk);
            updateAdjacentChunksReadyFieldOfAdjChunks(chunk);
            listener.onChunkReady(chunk.getPosition());
            worldEntity.send(new OnChunkLoaded(chunk.getPosition()));
        }
    }
```

And the method in RenderableWorldImpl allows only the rendering of a chunk when all adjacent chunks are ready:
```
    public boolean isChunkValidForRender(RenderableChunk chunk) {
        return chunk.isReady() && chunk.areAdjacentChunksReady();
    }
```

flo:
It is the same for the LocalChunkProvider too.

flo:
I updated the upper description, we need to load even more chunks at a server to transfer them to a client.

immortius:
Ideally, the behavior we want is:

* A chunk can be made ready for use as soon as it is generated
* This includes sending to a remote client
* A chunk can be rendered as soon as it is surrounded by other, loaded chunks. This is both so that its lighting can be adequately generated, and because rendering of a block requires the surrounding blocks (at the moment just to work out what parts need to be rendered, but ultimately the rendering of each block should depend on surrounding blocks to support things like fences connecting and potentially walking cube implementations).

Looking at the code at the moment though....

LightMerging current requires a chunk to be surrounded by chunks. This seems to be partially a performance thing (it is expensive doing light merging piece-meal) and partially a logic thing because doing ad hoc light merging can spill out of the immediate chunks in some situations. On the other hand, light merging could actually be put off until later. So we could change from:

1. Generate chunk
2. Wait until surrounded by generated chunks
3. Merge lighting
4. Mark ready
[...]

Issue #1574, opened 2015-02-07
Logging improvements for better 24/7 headless support


Cervator:
So we have another user interested in doing a 24/7 server to help test, which revealed a fun little issue: the log file never rolls and we have a few scenarios that can trigger log spam. So we need a bit more usability there :-)

- [x] Move the log file to use a RollingFileAppender from logback - http://logback.qos.ch/manual/appenders.html#RollingFileAppender - maybe roll daily + in 5, 10, or 20 mb size chunks?
- [ ] Allow the user to easily configure logging, ideally in config.cfg. Maybe on initial run write the log settings from logback.groovy into a logging section in config.cfg and if it exists on future startups use the log settings from there?
- [x] Minimize log spam from the issue in #1503 in a clean fashion (might be out of scope to fix it, but maybe move to FINE or DEBUG?)
- [ ] Minimize log spam from "received too much input" in a multiplayer session
- [x] Minimize any other common log spam. Only one I can think of is a "brush" thing I vaguely remember from Cities
- [x] Tweak the CrashReporter if needed to deal with log file changes. Fancy: Could add back/forward buttons to load individual rolled log files

@msteiger might be taking a look :-)

msteiger:
Changing the logging style to rolling logs is fairly easy: https://github.com/msteiger/Terasology/commit/849998fec2d9eb20cf59a7d61ef4d5a684d09edb. This will continue to write to the `Terasology.log` file (by setting both the file and archive fileNamePattern), but it has a few drawbacks:
* Appending to previous log files is mandatory :cry: We would need to delete old log files on startup first.
* It checks log file size (requires file read permissions) at every logging operation. This seems to be fairly fast though. Maybe @immortius can clarify if/how module permissions would have to be changed to make rolling logs work for modules? Why does writing to the log file even work? I saw that the `slf4j.Logger` interface was added in `ModuleManager` to the permission set, but there must be more to it.
* Rolling log files eventually throw out old log events. This is ok, but it would often be nice to retain the log of  the init phase.

Maybe we should just postpone the rolling log thing until it becomes a problem? Deal with the log spam and maybe activate zipping should keep log files reasonable small for a long time.

Having that said, I suggest configuring log file names and root log level through Java. This would lead to:
* Remove dependency to groovy entirely (smaller jar, yay!)
* Faster startup (no need to load groovy runtime and compile/run the logfile "code")
* Configuration of encoder / specific log levels, etc. in XML
[...]

PR #1888, opened 2015-08-14
Better handling of unloaded blocks

flo:
There is currently a lot of spam about attempts to access unloaded blocks.  e.g.
* right at start, when the chunk isn't loaded yet but the game already performs a ray cast to check which block the player is currently looking at
* When you throw an item into a chunk that isn't loaded there will be infinite spam about unloaded block access as the phsyic engine tries to handle the interaction of the item with the world.

This pull requests solves both issues by
* Firstly not spamming warnings about it
* Introducing a new block type engine:unloaded that all unloaded blocks have. Unloaded blocks can neither be penetrated nor be destroyed.

Thus if you try to throw an item in an unloaded chunk it will just bounce back.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/modes/loadProcesses/InitialiseRemoteWorld.java
+++ b/engine/src/main/java/org/terasology/engine/modes/loadProcesses/InitialiseRemoteWorld.java
@@ -68,7 +68,7 @@ public class InitialiseRemoteWorld extends SingleStepLoadProcess {
         RemoteChunkProvider chunkProvider = new RemoteChunkProvider(blockManager, localPlayer);
 
         WorldProviderCoreImpl worldProviderCore = new WorldProviderCoreImpl(gameManifest.getWorldInfo(TerasologyConstants.MAIN_WORLD), chunkProvider,
-                blockManager.getBlock(BlockManager.AIR_ID), context);
+                blockManager.getBlock(BlockManager.UNLOADED_ID), context);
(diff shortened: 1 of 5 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/226/

Cervator:
Excellent :-)

The log spam fix makes this related to #1574. I still wonder if there's any scenario where we might *want* to log, especially if we can block most the issues with these "unloaded" blocks? But then it would hardly be the first place where we have something that works without having hunted down every possible issue. Not a bad deal.

@immortius - look good to you? Seems like you liked the idea of this fix in the past.

Cervator:
Tested around some, seems good. Little uncertain how to test it right.

I'm also wondering if an issue I see is reproducible by others. On legally blind (at least) I often do not get adjacent blocks made visible (probably related to #1886) until I move into them. Yet I seem able to throw stuff into adjacent chunks fine so the meshes/blocks have loaded alright. Maybe need some hackery to truly provoke a fully unloaded chunk with the new unloaded blocks.

If alternatively I go `hspeed` and `ghost` then expand the view distance specifically to provoke the game into falling behind so I can fly beyond visible world gen I can still seemingly throw stuff - but maybe that's again just from the chunks being there just still not visible.

flo:
There is always a 1 chunk thick layer around the player that is invisible but loaded as the rendering requires the adjacent chunks to be loaded.

The way I tested it was by setting a thread only breakpoint in the chunk generation/loading.

immortius:
Looks good.

Issue #1770, opened 2015-06-12
Engine Initialisation and Lifecycle improvement


immortius:
This issue will deal with a restructure of the central engine initialization and lifecycle. The goal of this restructure is:

* Assist in the removal of CoreRegistry
* Fix a variety of interdependency and ordering issues, such as the dependencies between serializers and systems that may not exist until later. 
* Centralize core lifecycle concerns such as the ModuleEnvironment, AssetManager and EntitySystem lifecycles.
* Move concerns to Subsystems where appropriate.
* Expand on the subsystem lifecycle events, and ensure each event has a well defined meaning and usage.

Some additional possibilities:
* Handle the main menu and ingame states more generically. May move main menu into a module, and have the game start with a default module set. Starting a game is just a transition to a different module set. Loading a game is a transition to a different module set, but also providing a save state. So essentially all major state changes involve a module set change and optionally a save state.
* Removal of the State classes.

immortius:
Firstly the GameEngine needs some cleanup. It has a number of odd methods that it really shouldn't in it. Additionally, there is an obvious need for monitoring the status of engine from outside (currently the Terasology engine has a hardcoded linkage to the splashscreen which is questionable.

Something like this would satisfy external needs.

```java
public interface GameEngine extends AutoCloseable {
    /**
     * Registers a Subsystem
     * @throws IllegalStateException if the engine is running or disposed
     */
    void register(EngineSubsystem subsystem);

    /**
     * The current status of the engine (e.g. loading config, updating asset system, etc)
     */
    EngineStatus getStatus();

    /**
     * Subscribe for notification of engine status changes.
     */
    void register(EngineStatusSubscriber subscriber);

    /**
     * Runs the engine, which will block the thread.
     * Invalid for a disposed engine
     */
    void run(Set<Name> initialModules);

    /**
     * Request the engine to stop running
     */
    void shutdown();

    /**
     * Cleans up the engine. Can only be called after shutdown.
     * This method should not throw exceptions.
     */
    void close();

    /**
     * @return Whether the engine is running
     */
    boolean isRunning();

    /**
     * @return Whether the engine is shutting down
     */
    boolean isShuttingDown();

    /**
     * @return Whether the engine has been disposed
     */
    boolean isDisposed();
}
```

Internally I think the only methods needed are to trigger environment switch, and to access the current context:

```java
    Context getCurrentContext();

    void switchEnvironment(Set<Name> modules);
```

[...]

PR #1893, opened 2015-08-15
Update world gen. configuration with Component clones

msteiger:
This PR changes the current property binding scheme for world gen. configuration. Previously, fields were modified directly through PropertyProvider's reflection system. This has the drawback that world gens are not notified (or only afterwards) that a parameter has changed.

This PR uses copies of configuration components and calls worldConfig.setProperty() as usual. This allows the world gen. to update accordingly (reload a height map, clear caches) or even decline the new parameter set.

I had to change `handleSwitchToPreviewEnvironment.handleSwitchToPreviewEnvironment` so it would update the `ComponentLibrary` with the new module environment. It also uses the sub-context now. Unfortunately, this update cannot be undone atm. Since this is using the sub-context only, it should be ok though.

To test this try the HeightMap world gen. or PolyWorld. Both should work nicely now even when the height map / graph density is modified.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/bootstrap/EnvironmentSwitchHandler.java
+++ b/engine/src/main/java/org/terasology/engine/bootstrap/EnvironmentSwitchHandler.java
@@ -129,11 +129,16 @@ public final class EnvironmentSwitchHandler {
     }
 
 
-    public void handleSwitchToPreviewEnivronment(Context context, ModuleEnvironment environment) {
+    public void handleSwitchToPreviewEnvironment(Context context, ModuleEnvironment environment) {
         cheapAssetManagerUpdate(context, environment);
+        ComponentLibrary library = new ComponentLibrary(context);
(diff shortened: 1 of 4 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/228/

Cervator:
Tested around some, making config changes with heightmap and previewing a bunch. Looks fine. Does that cover the changes?

Pinging @flo if interested in commenting on the fun with contexts

flo:
This will propably collide with the changes @immortius  made about the engine life cycle.

See his last comment on https://github.com/MovingBlocks/Terasology/issues/1770

immortius:
Probably not worth worrying about my branch though, who knows when it will be ready.

Cervator:
Went ahead and merged it then, at least it isn't a huge number of changes to deal with :-)

PR #1898, opened 2015-08-18
Name tags

flo:
This pull request adds:
* A FloatingTextComponent that can be used to make an entity show a text at it's current location in 3d space
* A NameTagComponent that when added to make it simpler to show a text with a offset over an entity. It uses an entity with a FloatingTextComponent to do that internally.
* A system that uses the NameTagComponents to show name tags over players. 
* A command for renaming users (to test that the name tags update)

The system is currently implemented in the engine, but in future we can move it out into a module. That way it could become possible to implement a different name tag generation rule that also uses the NameTagComponent. e.g. one that shows team name tags.

Also it is possible to replace the NameTagComponent visualization if we wanted to do that as it is an abstraction.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java
@@ -33,6 +33,10 @@ import org.terasology.network.Replicate;
  * @author Immortius
  */
 public final class CharacterComponent implements Component {
+    /**
+     * Recommended height from center at which name tags should be placed if there is one.
+     */
+    public float nameTagOffset = 0.8f;
     public float eyeOffset = 0.6f;
     /**
      * Specifies the maximium range at which this character is able to interact with other objects.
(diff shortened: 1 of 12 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/237/

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/238/

Cervator:
Good stuff :-)

Observations:

* Name tags are visible through obstacles (plenty acceptable for now I'd think, option to improve later when appropriate/needed)
* Name tags are affected by blur (rings a bell - something else unexpected is affected by it too IIRC)

Actually that might be it. Seems solid! Tested headless, client-server, and with the player rename. I'm good to merge whenever. Keeping it in engine temporarily seems fine with me.

Pinging @msteiger @immortius @Josharias if anybody else wants to check it out :-)

msteiger:
That's great - I've wanted this for a long time. Will have a look soon.
Suggestion for a follow-up: Give the text some depth and rotate it around the y-axis. Would be nice for quests.

image

Cervator:
Went ahead and merged since it seems solid :-)

Thanks again @flo !

immortius:
@msteiger I suggest that is not a use case for the floating text, but rather a use case for creating a 3d question mark mesh and attaching it using a child entity with a RotationComponent.

Issue #1571, opened 2015-02-07
Headless launched via jar still shows splash screen?


Cervator:
I was testing the headless launch from jar just to make sure I was giving the right command to a user:

`java -Xms256m -Xmx1024m -jar Terasology.jar -headless -homedir=server`

But running that on my Win7 desktop still popped up the splashscreen? I would expect not to see it.

#1501 took it out from running from source in a Jenkins run, but this is technically a different launch scenario. Unless maybe #1550 brought it back somehow? Any clues @msteiger ?

Seems low priority, user ran server fine on a hosted machine and I was able to connect :-)

msteiger:
Oh well, I made that on purpose, because I did not see a reason to deactivate it :smile:
It is displayed whenever a graphics environment is available and the splash image is passed as jvm argument (either explicitly or in the `MANIFEST.MF` entry in the jar file).

Would you rather have it deactivated it when the runtime arg `-headless` is passed?

Cervator:
Yeah, in part because it never closes :D Or at least I didn't notice it doing so. Seems counter-intuitive anyway since it is a "head" for headless of sorts :-)

Cervator:
Hitching a ride on this issue instead of making a new one - running the game with the `-help` flag also shows the splash screen, that's probably not desired (it flickers briefly since the help returns very quickly), but unsure how easy it is to distinguish?

Noticed while testing #1575 which also moved the jar so the command is now slightly different when running out of a binary package:

`java -jar libs\Terasology.jar -help`

On a somewhat related note the new .exes aren't showing the splash screen but that's probably just temporary till @shartte figures it out

There we've also talked a bit about the .exe probably being stuck *either* as a GUI app or a console app, but not both, leading to a sensible use case where headless is *not* actually window less - it would make sense in that case to run a log (or even help text) in a tiny basic window. Later that could be turned into a proper server admin tool. So maybe headless showing a splash screen isn't bad if it is part of loading a window anyway.

That scenario is specific to the Windows .exes though. Running via .jar is fine - you can get the logging in the console yet have the game window pop up.

Cervator:
Bumping this rather than making a new issue: splash screen now seems to trigger in any case (other than when no graphics environment is available, I assume). Still a minor issue at most.

msteiger:
The library also contains a pure Java implementation of the splash screen. We could remove the native AWT version altogether and configure it's visibility through Java code. This would also fix this issue here. Will have a look..

[...]

PR #1901, opened 2015-08-21
Move splash screen code to external library

msteiger:
This was necessary since the splash screen code was moved to the PC facade and thus no longer available for WorldViewer.

There are no functional changes. Starting the game through
* gradle run / startscripts / launch4j exe
* as runnable jar
* IDE

works as before (i.e. show the splash screen if not disabled).


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/241/

Cervator:
Tested and merged, although the splash screen seems active all the time now? Which isn't really a big issue but somewhat renders the "Splash enabled" run config pointless :-)

Reminds me of #1571 where the splash screen would trigger even with headless (if done with a graphics head available)

Not worth holding this back for or anything, but if you were expecting for the splash screen to only be active in some cases this seems to be overly inclusive :-)

PR #1501, opened 2015-01-07
Disable splashscreen in headless mode

msteiger:
This PR fixes the headless server that can be run by Jenkins:

Before:
http://jenkins.movingblocks.net/job/TerasologyHeadlessServer/1/console

After:
http://jenkins.movingblocks.net/job/TerasologyHeadlessServer/2/console

PR #1903, opened 2015-08-23
Replace native splash screen with Java library

msteiger:
I think I overdid it, but see for yourself: new fancy splash screen with lots of animations.
Borderless window, incremental loading animation, better text rendering, etc.

The breathing image is too much though, imho. Removing the `Breathing` from the class name to see what it looks as a static image.

Fixes #1501


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/246/

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
```diff
@@ -237,6 +233,7 @@ private void logEnvironmentInfo() {
      * Gives a chance to subsystems to do something BEFORE managers and Time are initialized.
      */
     private void preInitSubsystems() {
+        changeStatus(TerasologyEngineStatus.PREPARING_SUBSYSTEMS);
```
immortius: This is pointless, since the status is changed again immediately afterwards?

immortius: Oh I see, you're using it as a trigger.

Cervator:
Very nice! I love the unusual look of the segmented images with some see through bits. Big fan of non-square windows but then maybe I'm weird :-)

Yeah the breathing logo text is pretty crazy, haha. Although I only noticed it the first or second time and that's because I was looking. I could take it or leave it, but maybe it is splashy enough without.

Should splash still default to on in all kinds of cases like headless? Or only be on when executed via "headful" binary client or a run config turning it on? No strong opinion there either, really.

(to be exact: i tried the headless launch config and gradlew start and both showed the splash screen - I figure launching headless via binary would do it too but haven't confirmed)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/252/

msteiger:
* Merge conflicts are resolved now!
* Breathing image replaced with static image - looking good now imho
* I have a slight preference towards showing the splash screen on headless runs. It tells me that it's actually doing something when I click the icon. I'd be willing to change that if desired though.

Cervator:
Yeah one nicety with the new splash screen is that it isn't forced to stay in the foreground, so you can just make it go to the background if desired. I'm happy with that :-)

And eventually we might *want* a UI for controlling the headless server anyway, which could be nice when hosting on a normal PC

immortius:
The one thing you need to be careful of is that headless runs can occur in environments without UI at all (e.g. linux console). But that can be handled with the existing commandline options easily enough.

Cervator:
I think that's covered, @msteiger mentioned earlier that the splash screen is auto-disabled if the system has no available graphics head :-)

PR #1907, opened 2015-08-27
Usability tweaks for module and world gen. selection

msteiger:
Modules that are explicitly selected are currently not preserved across restart, because the selection of the gameplay module was overwriting it.

Also, the list of available world generators was not refreshed when the module selection screen is closed.

I also moved the updating of the gameplay dropdown box to the onOpened() method. It was calling `set()` from within the `get()` method, which can easily cause infinite loops.

**Steps to test**: Select Core Gameplay and Cities. Going back to the Create Game screen does (not) list the Cities world gen. until restart. On restart the Cities module is (not) selected.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
@@ -114,27 +114,6 @@ public class CreateGameScreen extends CoreScreenLayer {
 
             @Override
             public Module get() {
-                // try and be smart about auto selecting a gameplay
-                if (selected == null) {
-                    // get the default gameplay module from the config.  This is likely to have  a user triggered selection.
-                    Module defaultGameplayModule = moduleManager.getRegistry().getLatestModuleVersion(
-                            new Name(config.getDefaultModSelection().getDefaultGameplayModuleName()));
-                    if (defaultGameplayModule != null) {
-                        set(defaultGameplayModule);
-                        return selected;
-                    }
-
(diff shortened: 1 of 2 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/251/

immortius:
Perhaps gameplay template should be replaced with "Custom" if manual module selection has occurred?

msteiger:
Sounds like a good follow-up TODO to me: #1911

Cervator:
Took a quick look with a bit of testing, looks to be working well. We've been going back and forth on what this behavior should be and this (plus the custom thing) seems a good fit :-)

One oddity: for me L&S is the default gameplay template now instead of Core - is that just a quirk on my system or happens for others too? That's a bit of a drawback as L&S is broken at the moment pending a quick world fix :-)

msteiger:
Hmm .. the gameplay selection works for me at least. Core is the default entry.

Cervator:
Do you have a working or broken version of L&S available locally? I'm wondering if it being broken on my system impacts the selection. Although it still gets listed so ..

Think you can check out its world gen real quick sometime? The changes in Cities impacted it, probably minimally with an argument or two out of whack or something. If it is working again testing would be easier - I can't force compile my local workspace yet I can't very well see if it still shows as default if I remove it :D

Cervator:
Went ahead and just merged this. We can tweak at the default selection if it is an issue before the next release :-)

After all latest unstable isn't exactly guaranteed to be stable!

msteiger:
Thanks - I can have a look at L&S soon. Just need to sort out a few other TODOs first.

PR #1909, opened 2015-08-27
Allow widgets inside the browser to still be interacted with

Josharias:
Previously, the BrowserWidget would asume all the interactions done inside regardless if one of its children had an interaction region specified.  This prevented custom widgets rendered inside from showing tooltips and being clicked on.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/browser/ui/BrowserWidget.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/browser/ui/BrowserWidget.java
@@ -41,9 +41,6 @@ public class BrowserWidget extends CoreWidget {
     @Override
     public void onDraw(Canvas canvas) {
         hyperlinkBoxes.clear();
-        if (displayedPage != null) {
-            DocumentRenderer.drawDocumentInRegion(displayedPage, canvas, canvas.getCurrentStyle().getFont(), canvas.getCurrentStyle().getTextColor(), canvas.size(), register);
-        }
         canvas.addInteractionRegion(
                 new BaseInteractionListener() {
                     @Override
@@ -61,6 +58,9 @@ public class BrowserWidget extends CoreWidget {
                         return true;
                     }
                 });
+        if (displayedPage != null) {
+            DocumentRenderer.drawDocumentInRegion(displayedPage, canvas, canvas.getCurrentStyle().getFont(), canvas.getCurrentStyle().getTextColor(), canvas.size(), register);
+        }
     }
 
 


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/253/

MarcinSc:
What is the usecase for having widgets INSIDE of the browser?

Josharias:
I would like to put images of items/blocks inside the browser.  It is related to the stuff I am doing with the in game item help dialog.  It would be great if these items could provide tooltips like they do when they are outside the browser:
http://forum.terasology.org/attachments/moaricons-png.2094/

MarcinSc:
Maybe better than would be to have additional listener "hyperlinkHover" or something like that that provides Tooltip to display? You won't be able to place a widget inside and have it correctly positioned using document anyway.

Josharias:
The only problem with adding more hover plumbing on the browser is the complication that gets involved when you have nested widgets inside of the widget that is placed in the browser. 

I havent had any big issues positioning widgets so far,  are there some known times it will be bad?

What is the risk that you see in interacting with the widgets that are in the browser?  What am I missing?

MarcinSc:
Are the widget laid out by the document or are you doing it by adding them by code/*.ui externally?

MarcinSc:
The issue is, that it was not designed for that, so it might break at some point, or produce memory leaks due to not being disposed correctly if document changes?

Josharias:
They are being added by code using this: https://github.com/Terasology/InGameHelp/blob/master/src/main/java/org/terasology/inGameHelp/ui/WidgetFlowRenderable.java

So they should follow normal disposal of paragraph data.

MarcinSc:
ParagraphData does not have dispose methods, does it?

If so, it might be a problem, as some of the widgets might load textures and other resources that might not get disposed.

immortius:
Which touches on the future work I need to do around automatically collecting and allowing the disposal of garbage collected assets. I've actually done this for the next version of gestalt-asset-core, but it uses finalisers which is bleh. I'm thinking of changing the API a bit to allow for something better.

Josharias:
If disposal of assets is a potential problem for adding widgets, then it is not a new problem introduced with this PR as the images added currently could also suffer potential cleanup problems.

Cervator:
Sooo ... yea/nae on merging as-is?

Josharias:
(back from 2 day power outage)  Layering in more plumbing to accommodate this functionality inside the browser seems like more effort that is needed.  I do not see any advantage to adding more of a sandbox for this widget compared to the other widgets, as you will have similar disposal problems in all cases.  Immortius has a long term plan for better asset disposal that we can hook into later on.

Interaction regions will naturally default to the browser provided interaction region.  I think this is a good fit by allowing UIs take control when needed.  Obviously this also comes with the caveat of, "with great power,  comes great responsibility".

Cervator:
Thought then: With great responsibility comes great documentation? :-)

Do we have any real docs / tutorials on writing UI pieces? Beyond any of our usually rare javadoc, anyway. Maybe we should keep a wiki tutorial that can highlight these caveats and mention/link to later projects?

I'm thinking I'll merge this tonight if nobody does/objects earlier, but figured the doc angle would be a nice extra

Cervator:
All set. Good luck avoiding any further fun with power outages :-)

Issue #1810, opened 2015-07-04
Crash on startup - Linux Mint 17 with GLib


Cervator:
Report from IRC, same user (Froward) and system as https://github.com/MovingBlocks/DestinationSol/issues/40

Crashed on game startup, seems to be an error in GLib, specific to Linux systems. This is on a laptop with two GPUs.

Game log: http://pastebin.com/yjEXueE6

Snippet: 

`(process:18611): GLib-CRITICAL **: g_slice_set_config: assertion 'sys_page_size == 0' failed`

System details: http://pastebin.com/CZRxh8Jm

Snippet:

```
00:02.0 VGA compatible controller: Intel Corporation Mobile 4 Series Chipset Integrated Graphics Controller (rev 07)
...
01:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] RV620/M82 [Mobility Radeon HD 3450/3470]
```

I found the same error related to a Firefox bug - some links
- http://ubuntuforums.org/showthread.php?t=2175102
- https://bugs.launchpad.net/ubuntu/+source/firefox/+bug/1160569
- http://ubuntuforums.org/showthread.php?t=2238827
- https://bugzilla.mozilla.org/show_bug.cgi?id=8331170

I'm not quite sure what we can do here since it seems specific to the OS, distro, GLib, or maybe video card drivers? One of the links got into the proprietary vs. open source GPU device drivers we've hit on occasion in the past.

May also be difficult to troubleshoot when system specific like this.

@emanuele3d any idea if this could be related to initializing something graphics related or nah?

Edit: It is in the log but to highlight this is for stable release 55, so unrelated to recent work

Edit2: May also want to follow up with http://forum.terasology.org/threads/failed-install.1381/ if progress is made
##

Cervator:
It looks like this *may* be the same issue encountered in Destination Sol just spiraling into a weird error. Froward reports that using the same patched LWJGL 2.9.2 that fixed DestSol also fixed Terasology.

The fix in question, https://github.com/LWJGL/lwjgl/commit/0b83683ed155f089af61b2f4497cfda8181e333a, actually got merged to LWJGL master. However, I don't see it released in a 2.9.4 anywhere. It may exist unofficially / in a nightly build.

We could likewise build it ourselves or use a nightly, then use that version for both Terasology and Destination Sol to fix both issues (its use is transitive though, so we'd need to override it there). 

[...]

PR #1923, opened 2015-09-01
Update most library versions to latest

msteiger:
In order to avoid complications with reflections, the gestalt library dependencies should be updated as well.

An alternative to updating simultaneously would be relying on gestalt including a set of "basic" libraries such as reflections, gson, guava, slf4j.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,7 +67,7 @@ ext {
     // Lib dir for use in manifest entries etc (like in :engine). A separate "libsDir" exists, auto-created by Gradle
     subDirLibs = 'libs'
 
-    LwjglVersion = '2.9.2'
+    LwjglVersion = '2.9.3'
 }
 
 // Declare remote repositories we're interested in - library files will be fetched from here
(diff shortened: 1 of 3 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/259/

Cervator:
If memory serves there's something interesting in the unreleased LWJGL 2.9.4. Think it was found in DestSol then a similar issue came up with Terasology. Looks like #1810

Might be worth looking at using that one / reviewing the applet issue to see if we can get all that out of the way with one round of updates/testing?

@immortius thoughts on the Gestalt angle?

Cervator:
@immortius + @msteiger: ping

Cervator:
Addition: We could bump TeraMath to a new non-snapshot release along with this (needs to be done first). Seems it would fit with this PR nicely.

Cervator:
Done for now, will worry about LWJGL and Gestalt later. It sounds like we'll see an update of Gestalt land soon enough anyway :-)

PR #1925, opened 2015-09-02
Use large Noto font in floating billboards

msteiger:
This adds Noto Sans in font size 52 (the current one is 20) with an identical character set and sets it as default font for `FloatingTextRenderer`. The different fonts look like this:

noto

Note that the old "default" font is unable to display the full text. A second use case for the large font is to replace the current title font, which isn't ready for i18n text either.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/260/

Cervator:
Short and sweet, good stuff. Nice screenies :-)

PR #1927, opened 2015-09-10
Compose component name for nested classes

### LINKED ISSUE #1924: Copying Components fails for some class names
The Cities module has several facet providers, which all have a
```java
private static class Configuration implements Component
```
Starting the game with Cities currently fails because those config components cannot be copied. Renaming them to unique names seems to resolve the issue. Some debugging revealed:

* In CreateWorldEntity the world gen. components are added to the world entity
* The ReadWriteStorageManager is notified and delegates to EntitySetDeltaRecorder
* In ESDR.onEntityComponentChange() the entity mgrs. ComponentLibrary is used to copy components
* The AbstractClassLibrary.classLookup map does not contains an entry for the class

I think that it's caused by the fact that their `ComponentMetadata` share the same URI. The AbstractClassLibrary uses this URI in a lookup table. 

This could be fixed if the URI  also reflected the outer class if present. I *believe* that all(?) component class names are generated through MetadataUtil.getComponentClassName().

So maybe add a check for `componentClass.getEnclosingClass()` and concatenate if it exists? Unsure if the  default `$` should be put in between then.

msteiger:
Currently, component names are solely defined by their class names. This can be a problem for world generator configurations. The facet providers' configuration data is attached as a component to the world entity so it can be saved/loaded through the entity system. These names need to be unique - see #1924.

This leads to some rather long names such as `PerlinHillsAndMountainsProvider.PerlinHillsAndMountainsProviderConfiguration`

This PR resolves the issue by adding the enclosing class name (if available) to the component name. The nested class name could then simply be `Configuration` for more than one facet provider.

This PR also adds a warning if two different classes are registered for the same component name.

Closes #1924

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/metadata/MetadataUtil.java
+++ b/engine/src/main/java/org/terasology/entitySystem/metadata/MetadataUtil.java
@@ -29,6 +29,11 @@ public final class MetadataUtil {
 
     public static String getComponentClassName(Class<? extends Component> componentClass) {
         String name = componentClass.getSimpleName();
+        Class<?> outer = componentClass.getEnclosingClass();
+        if (outer != null) {
+            name = outer.getSimpleName() + name;
+        }
+
         int index = name.toLowerCase(Locale.ENGLISH).lastIndexOf("component");
         if (index != -1) {
             return name.substring(0, index);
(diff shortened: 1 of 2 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/263/

immortius:
What if the outer class is also a static inner class?

msteiger:
I think that would be a rather exotic scenario - similar to your argumentation in #1924. Still, it would work if any of two inner classes names' differ. If not, a warning is reported. Thus, I believe that it is covered well enough.

Cervator:
Merge time? @immortius any further comments?

PR #1935, opened 2015-09-16
Only add the CoreGameplay module if the module selection is empty

Josharias:
This will allow headless servers to run without CoreGameplay,  but still avoid breaking if there are no modules selected.  For the most part,  the default.cfg will take care of ensuring CoreGameplay gets selected at minimum.  This should still catch the times where there is invalid module names in the config file.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/269/

Cervator:
Makes sense to me. Merged!

PR #1941, opened 2015-09-20
Clear canvas FBO only when created

msteiger:


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/272/

Cervator:
Tested this out together with the rewritten Minimap but I still get nothing but black on my minimap (or gray if I zoom out and show unloaded chunks).

I suspect the bug you fixed is real but the bug I have is bigger and hides it. I never actually tried to break any blocks - the map started black so the smaller bug never had a chance to show itself.

Probably it'll work for you and at least some others now without the block updating causing blackness, but something on my system is likely leaving it broken. At least this way we can start having others test it :-)

Specific build that should exercise it: http://jenkins.terasology.org/job/DistroOmega/264/artifact/distros/omega/build/distributions/TerasologyOmega.zip

PR #1944, opened 2015-09-23
Harden EventSystem against failures in modules

msteiger:
Currently, the entire game crashes if a `@ReceiveEvent` method in a module throws an exception.

I think that it should log the exception and continue instead. This leads to improved game experience and is also easier to debug the broken code.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/274/

Cervator:
Sounds good to me. Any sense in capturing the more detailed exceptions first for any enhanced logging or really no point? Maybe throw a comment on there to clarify. Although that's probably nitpicking and this could just be merged :-)

PR #1947, opened 2015-09-26
Fix serialization bug with lists in components

### LINKED ISSUE #1945: Support complex components
I'm currently re-writing the Quests module and designed the following format to define player quests:
```json
"Quest" : {
    "shortName" : "TestQuest",
    "tasks" : [
        {
            "id" : "collectDirt",
            "type" : "CollectBlocksTask",
            "data" : { "amount" : 3 }
        },
        {
            "id" : "returnHome",
            "type" : "GoToBeaconTask",
            "dependsOn" : ["collectDirt"],
            "data" : {  }
        }
    ]
}
```
The different parts of a quests (=tasks) are nested inside a single quest component, because different instances of a class can exist. For example: collect 3 dirt blocks and 2 wood blocks.

This is not yet  working,  because the `tasks` element are not parsed (no entry in type serialization library?) even though a default Gson instance can read it just fine.

Also, the factories to convert `type` and `data` json into actual `Task` implementations need to be manually added the because ModuleEnvironment.getSubtypesOf() method is not available for modules.

Ideas? Feedback? Sending a ping to @nh-99 @MarcinSc @Josharias  and @immortius.

msteiger:
Partially fixes #1945

### DIFF
--- a/engine-tests/src/main/java/org/terasology/entitySystem/stubs/ListOfObjectComponent.java
+++ b/engine-tests/src/main/java/org/terasology/entitySystem/stubs/ListOfObjectComponent.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2013 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/276/

immortius:
In what way does this "partially" fix the problem? What is fixed and what is not?

msteiger:
The list bug is fixed by this commit alone. 

To deserialize "complex" components such as quests, other means are required as well, for example #1948.

PR #1948, opened 2015-09-26
Add GenericObject for delayed conversion

msteiger:
Simple and straightforward.

Completes #1945 together with #1947.

### DIFF
--- a/engine/src/main/java/org/terasology/persistence/GenericObject.java
+++ b/engine/src/main/java/org/terasology/persistence/GenericObject.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2015 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/277/

immortius:
This seems a rather lazy way to avoid making TypeHandler definition available to modules.

Cervator:
Is that a good thing or a bad thing? Should that be done instead or is it a problem to expose that?

(thanks for taking a look in any case!)

immortius:
I guess I was envisioning modules being able to declare type handlers. I'm not sure how well GenericObject would work on components, particularly in the future - components are going to either be continuously copied for thread safety, or serialized into some sort of in-memory datastore, so there isn't really scope partially deserialized components unless you don't mind systems working with GenericObject directly.

Maybe I'm missing how GenericObject would be used though?

msteiger:
The "lazy" part is very true actually. I've briefly sketched a concept for custom type adapters, but deferred it as it is not trivial. For now, the `GenericObject` approach fits well enough:

The `QuestSystem` has a hard-coded list of `TaskFactory` instances. Each one of them gets a `PersistedDataMap` and creates a certain type of `Task` from it. See [Tasks/QuestSystem](https://github.com/Terasology/Tasks/blob/develop/src/main/java/org/terasology/tasks/systems/QuestSystem.java#L159) for details.

This limits the definition of new task types to the QuestSystem class though - a generic approach based on custom type adapters would certainly help.

PR #1953, opened 2015-10-01
Add support for custom type handlers

msteiger:
This PR adds the `@RegisterTypeHandler` annotation that allows modules to register custom type handlers.

A simple implementation: [Tasks/TimedTaskTypeHandler](https://github.com/Terasology/Tasks/blob/develop/src/main/java/org/terasology/tasks/persistence/TimedTaskTypeHandler.java)

Delegates to custom implementations, similar to UIFormat: [Tasks/TaskTypeHandler](https://github.com/Terasology/Tasks/blob/develop/src/main/java/org/terasology/tasks/persistence/TaskTypeHandler.java)

### DIFF
--- a/engine/src/main/java/org/terasology/engine/bootstrap/ClassMetaLibrary.java
+++ b/engine/src/main/java/org/terasology/engine/bootstrap/ClassMetaLibrary.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2015 MovingBlocks
+ *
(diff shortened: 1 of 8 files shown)


msteiger:
We could actually revert #1948 since this supersedes it.

Cervator:
I was wondering if I should merge or wait on that one - oh well :-) Want to just do the delete in a new commit on this PR?

Cervator:
Merged then undid the changes from #1948, thanks as always :-)

PR #1958, opened 2015-10-04
Send chunks to client before light calculation is done. (Fix #1859)

### LINKED ISSUE #1859: Headless server + solo legally blind client = no visible chunks
Pretty sure this has been noted before with @minnesotags' server, but on there with pre-alpha release 55 chunks did eventually become visible by a client. Testing just now I waited a minute or two without having any visible blocks in the world. Yet the world *has* generated fine, I can walk around, place things, throw torches, etc.

A hosting player + client player doesn't have the same problem, world is visible immediately.

flo:
Send chunks to client before light calculation is done. (Fix #1859)

This fixes #1859 and improves (but not fully solves) the situation described in #1886

Reasoning:

The light calculation gets only performed once the adjacent chunks got
loaded. So if wait for the light calculation before we mark a chunk as
relevant for a client then we would transfer less chunks to the client
then the relevance region is large. the client would then again perform
the light calculation too based on that reduced chunk count and would
reduce the view distance again. That is why it makes sense to detect
chunks as relevant even when no light calculation has been performed yet.

### DIFF
--- a/engine/src/main/java/org/terasology/world/chunks/internal/ChunkRelevanceRegion.java
+++ b/engine/src/main/java/org/terasology/world/chunks/internal/ChunkRelevanceRegion.java
@@ -168,7 +168,17 @@ public class ChunkRelevanceRegion {
         return Objects.hashCode(entity);
     }
 
-    public void chunkReady(Chunk chunk) {
+    /**
+     * Checks if the chunk belongs to this relevance region and adds it to it if it is relevant.
+     *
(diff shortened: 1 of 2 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/283/

Cervator:
Nice! Tested it out and looks good. Many thanks :-)

PR #1961, opened 2015-10-06
Avoid blank player names

### LINKED ISSUE #1957: Players shouldn't be able to pick an empty name
This causes a lack of nametag in multiplayer for instance. Can't even tell from the "Players Online" box who else is on (other than maybe deducing player number from its size). Although I suppose if a second player named "" joins they'll end up being named "1" :-)

Probably an easy fix.

Josharias:
Resolves #1957

- Adds a OK/Cancel button to the player settings screen,  only committing the changes if OK is pressed.
- Adds validation to the player name to not allow blank names.
- Allows buttons to show tooltips even if disabled (so that the validation message could be shown)
- Shows the validation message when hovering over the OK button or player name text box.
- Uses the translation system for the validation message.

image
image
image

Controversial things, please comment on:
- Using tooltips for validation messages
- There is no validation message if the mouse is not over the OK button or name text box.
- Tooltips on both text box and button?  or just one?  or none?

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/PlayerSettingsScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/PlayerSettingsScreen.java
@@ -28,7 +28,6 @@ import org.terasology.rendering.nui.Color;
 import org.terasology.rendering.nui.CoreScreenLayer;
 import org.terasology.rendering.nui.UIWidget;
 import org.terasology.rendering.nui.WidgetUtil;
-import org.terasology.rendering.nui.databinding.BindHelper;
 import org.terasology.rendering.nui.databinding.DefaultBinding;
 import org.terasology.rendering.nui.widgets.ActivateEventListener;
 import org.terasology.rendering.nui.widgets.UIImage;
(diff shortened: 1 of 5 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/284/

Cervator:
Thanks!

For the "controversial" note should we just have a hidden label for validation messages, and it shows "You must enter a name" if needed? Rather than tooltips. This could be a general pattern used elsewhere as well. Like a whole validation box sitting outside the main UI box on either side (so it can be invisible and not mess up alignment) that only shows validation messages if the UI has determined one is needed.

msteiger:
Cool stuff! I think that the tooltip for the disabled button makes sense. I can't press it, so I hover it to find out why.

I wouldn't expect that for the UIText field, though. How about adding a checkmark/cross at the end to indicate the status?

We could add the Symbols version of the Noto Sans font, which contains [unicode chars](http://www.fileformat.info/info/unicode/char/2714/index.htm) for these. This would make it easy to include and could be used in button texts as well.

image

Cervator:
I like the idea of adding an x or checkmark. That could work with an on-hover too.

Cervator:
Merged both PRs. Should we put in a reminder issue for later to improve validation further or nah, we'll just get to it naturally one day?

Josharias:
Working on the things mentioned.  However I was having difficulty placing the "x" in the right spot.  Will keep on plunking on it as I am also not all the way sold on how tooltips need to be enabled on the button.

msteiger:
From memory: the player name line should be a `RowLayout`. I would add another `UILabel` and set the text to either `\u2714` (heavy check mark) or `\u2718` (heavy ballot X).

PR #1963, opened 2015-10-07
Integrate symbol font into regular font

msteiger:
This should make it a lot easier to work with special characters.
Unfortunately `Enclosed Alphanumeric Supplements` had to be removed, because they were combined (more than one char) symbols.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/286/

msteiger:
It can also be combied with the font color characters:
image

Cervator:
Nice! Will merge momentarily. Able to put it to use somewhere so we have a nice example?

(Btw: Tasks still use the GenericObject, able to upgrade that sometime? I merged the PR without realizing, probably pushed and didn't notice it was checked)

msteiger:
Tasks should be fixed already. Maybe I forgot to push the changes from last night. 

Maybe an entry in the Wiki on fonts/text could be useful?

Issue #1890, opened 2015-08-14
Show dependency information of modules


msteiger:
It would be nice to have a dependency listing in the description of modules. For example, when you download `FlightModeToggleButton` it won't work until `HUDToggleButtons` is present as well. This information should be communicated.

Taken from #1884.

andrelago13:
Hi!

I would like to try to do this within the [SelectModulesScreen.java](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/SelectModulesScreen.java) file. I was thinking the dependency information could be show in the same space as the module's description, but I don't know what you prefer.

andrelago13:
Alright, since it was a quite simple task I just decided to go on and tweak around a bit and I got this done the way I suggested above. Simply put, it checks wether the module has any dependencies with help from ModuleMetadata methods, and print their names if they exist.

Do you think this is a good approach or were you expecting a better and more sophisticated solution?

Below is what my solution looks like:

![imagem1](https://cloud.githubusercontent.com/assets/9060644/11703708/4355699e-9ed9-11e5-9897-ddea3a8acdca.png)

Cervator:
Hey @andrelago13 and sorry for not getting back to you sooner, I sometimes only get time to respond to these things in the evening US East time :-)

That's a great improvement! I think that covers this issue well enough to consider it complete.

Bonus points for:

* Color the text of dependencies based on what's available, not of the right version, not present, etc, like the modules on the left would.
* Make the description panel into a tabbed panel and show description on one tab and dependencies on another. I'm sure we can find something to use a third tab for like a preview graphic (#1487) or optional dependencies, whenever either of those two are actually implemented

But really those are optional, the change is sufficient as it is :-)

msteiger:
I agree - I would leave it as is for now as it improves the situation significantly. A slightly more sophisticated solution could be adding a new row (below "Status"?) listing all dependencies.

The server list dialog does something very similar: it lists all modules on the server and color-code them in red/green whether they are present on the client(=local) machine.

Cervator:
Adding a row to status would probably overflow the box on modules with bigger dependency lists though. No scrolling in there :-)

I think tabs will be the way forward but in the main description box at least we get scroll so that can support really long lists if needed.

And yeah good reference to the server list, nice thing to review.

andrelago13:
[...]

PR #1971, opened 2015-10-11
Make the download button in the module selection screen also download all required modules

flo:
This pull request makes the "Download" button in the module selection screen also download all required modules.

Before the download starts the user is asked to confirm that he indeed wants to download that number of modules.

Only modules that haven't been downloaded in the newest version will be downloaded. If a module is however at a snapshot version it will always be redownloaded, as we can't tell if it has changed or not.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/291/

Cervator:
Awesome! Thanks for this nice extra. This partially finishes / works around #1890, up for finishing it off entirely and making the dependencies list in the UI somewhere? :-)

Issue #1940, opened 2015-09-20
Usability tweak for meta-server module scan: ad-hoc instead of auto


Cervator:
Currently when you open the module list on the create new world screen the meta-server module scan starts adding entries to the list immediately, potentially making it scroll around. This can be a little distracting especially if you started with fairly few modules (like in the case of outright testing a smaller set of modules)

Rather than automatically start adding new modules to the list maybe we should have a button to trigger an online check for new / updated modules? Similar to not generating a map preview until the user requests it.

Edit: It also resets and re-triggers after you've been in-world and go look at the module screen again. It should probably be cached and re-used in the future.

PR #1972, opened 2015-10-11
Module list loading speedup

flo:
Reasoning:

Waiting for the list to load is annoying.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/292/

Cervator:
Oh I didn't even realize we had a pause in there and that was causing the longer loading. Nice and easy little tweak. If it was meant to do an animation of sorts while downloading we can probably reassess that when we actually have enough modules to take a while to get info on :-)

Thanks! This fixes #1940 IMHO so will close that too

Issue #1304, opened 2014-10-23
Pathfinding + spawning L&S creatures + unload chunks == boom!


Cervator:
Found while testing world loading/unloading. 

* Start a Light & Shadow world
* Place a spawner block or two to get some L&S creatures
* Fly far away so those chunks unload
* Mark an area for one of the pathfinding orders
* Wait a sec - crash! :D

First time I was just flying around furiously (with hspeed) and the stacktrace included a permission error. Second time I just tested exactly like above and provoked a similar error without permission error

First:

```java
23:29:55.622 [Chunk-Generator-3] WARN  o.t.cities.raster.ChunkBrush - Y value of 4 not in range [192..255]
23:30:05.185 [main] INFO  org.terasology.work.WorkBoard - Pathfinding:walkToBlock: open=0 requestable=0 targets=0 
23:30:05.185 [main] INFO  org.terasology.work.WorkBoard - Pathfinding:removeBlock: open=0 requestable=0 targets=0 
23:30:05.185 [main] INFO  org.terasology.work.WorkBoard - Pathfinding:buildBlock: open=0 requestable=0 targets=0 
23:30:10.197 [main] INFO  org.terasology.work.WorkBoard - Pathfinding:walkToBlock: open=0 requestable=0 targets=0 
23:30:10.198 [main] INFO  org.terasology.work.WorkBoard - Pathfinding:removeBlock: open=0 requestable=0 targets=0 
23:30:10.198 [main] INFO  org.terasology.work.WorkBoard - Pathfinding:buildBlock: open=0 requestable=0 targets=0 
23:30:12.428 [Workboard-0] ERROR o.t.u.concurrency.TaskProcessor - Error in thread Workboard-0
java.lang.NullPointerException: null
	at org.terasology.work.WorkBoard$FindWorkTask.run(WorkBoard.java:317) ~[na:na]
	at org.terasology.utilities.concurrency.TaskProcessor.run(TaskProcessor.java:50) ~[na:na]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145) ~[na:1.7.0_55]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615) ~[na:1.7.0_55]
	at java.lang.Thread.run(Thread.java:745) ~[na:1.7.0_55]
Unexpected exception
[...]

PR #1979, opened 2015-10-15
Make Behavior assets available for modules

msteiger:
Also adds a `null` check to avoid unnecessary NPEs.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/behavior/BehaviorComponent.java
+++ b/engine/src/main/java/org/terasology/logic/behavior/BehaviorComponent.java
@@ -17,12 +17,14 @@ package org.terasology.logic.behavior;
 
 import org.terasology.entitySystem.Component;
 import org.terasology.logic.behavior.asset.BehaviorTree;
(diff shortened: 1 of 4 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/293/

Cervator:
Tried out WildAnimals as a quick test and merged.

Am wondering if the NPE fix relates at all to the crash in #1304? I can't seem to reproduce the crash now, although I can still get some stacktraces from the unstable work system. But maybe that was fixed by something else in the recent past

Found out what went wrong last I tried testing. Dialogs didn't have a master branch yet, just develop, which will be the right way to go eventually, but at present Gradle assumes master, so it wasn't fetching anything :-)

With the quick version tweak basic L&S works again, yay. Tried testing the new stuff but got compile errors related to Cities (with both it and L&S running develop, which may well not be ready for testing). But that's fine as long as we can release, I'm just eager to play with the new stuff :D

PR #1992, opened 2015-10-25
Fix various issues in JoinGameScreen

### LINKED ISSUE #1918: Issues on the join server custom screen related to the server entries
For starters the game will crash if you click the "Edit" button without an entry selected. Pretty easy oversight and I figure it should be an easy fix too :-)

For an additional slight bit of convenience maybe make the UI auto-select the first entry, which is always there from a pristine launch (the localhost entry). However since a user can remove all the entries that's still not a reason to not simply disable the "Edit" button with no entry selected

But oh wait, the Remove button doesn't actually seem to work so you can't remove the "localhost" entry (actually, it looks like it disappears after restarting the game). And the "Add New" doesn't seem to result in creating an visible entry either (although you can see the new entry's info until you select another), huh. So there might be a few related issues there. Is some of this just happening for me?

In short: the Edit & Remove buttons should be ghosted/disabled if no entry is selected, and actually using Remove & Add New should visibly result in a change right away :-)

msteiger:
Fixes #1918

### DIFF
--- a/engine/src/main/java/org/terasology/config/NetworkConfig.java
+++ b/engine/src/main/java/org/terasology/config/NetworkConfig.java
@@ -17,10 +17,11 @@
 package org.terasology.config;
 
 import com.google.common.collect.Lists;
+
 import org.terasology.engine.TerasologyConstants;
 
+import java.util.Collections;
 import java.util.List;
-import java.util.Optional;
 
 /**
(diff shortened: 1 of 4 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/300/

PR #1994, opened 2015-10-28
More seamless tooltip functionality for custom widgets

Josharias:
The main meat of this is adding a base implementation of onDraw into AbstractWidget.  This allows tooltips to be displayed even if there is no other interaction region added from the subclass.  Without this, it becomes confusing why setting a tooltip on your custom widget does by default show a tooltip until you also add an interaction region.

The UIButton is back to its former self by not adding an interaction region when it is disabled.  Tooltips on a disabled buttons now happen naturally and no longer need this special treatment.

The other noise in here is getting all other widgets to call this base onDraw method.  There could be controversy on whether it is actually needed on the everything.  I did the everything for completeness sake without regard to usefulness.  Advice welcomed on any additional strategy.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/internal/CanvasImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/internal/CanvasImpl.java
@@ -30,17 +30,18 @@ import org.terasology.input.MouseInput;
 import org.terasology.input.device.KeyboardDevice;
 import org.terasology.input.device.MouseDevice;
 import org.terasology.math.Border;
-import org.terasology.math.geom.Rect2i;
 import org.terasology.math.TeraMath;
 import org.terasology.math.geom.BaseVector2i;
-import org.terasology.math.geom.Vector2i;
 import org.terasology.math.geom.Quat4f;
+import org.terasology.math.geom.Rect2i;
+import org.terasology.math.geom.Vector2i;
 import org.terasology.math.geom.Vector3f;
 import org.terasology.rendering.assets.font.Font;
 import org.terasology.rendering.assets.material.Material;
 import org.terasology.rendering.assets.mesh.Mesh;
 import org.terasology.rendering.assets.texture.Texture;
 import org.terasology.rendering.assets.texture.TextureRegion;
+import org.terasology.rendering.nui.BaseInteractionListener;
 import org.terasology.rendering.nui.Color;
 import org.terasology.rendering.nui.HorizontalAlign;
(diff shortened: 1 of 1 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/302/

msteiger:
One option would be moving the code from `onDraw()` to `CanvasImpl.drawWidget()`. This would avoid the (admittedly not very strict) necessity to call the super method in the implementations.

Somewhat related: https://en.wikipedia.org/wiki/Call_super

Josharias:
I looked in CanvasImpl first to find a good place to plug the default tooltip behavior, I did not recognize one.  Could you elaborate more?  I agree it would be nicer not to call super everywhere.

There is some benefit to putting this in a call to super in that you can have the flexibility to let the Abstract widget's interaction region to be layered above or below the inner content.  This could let your tooltip take precedence over the inner content's tooltips.   However, this is a small win and I am not sure how much merit it should hold.

msteiger:
I've [created a tiny sketch where the call is moved](https://github.com/MovingBlocks/Terasology/commit/59beaeef07be5bfc257b8189cca9617a6e75dfb2) to `CanvasImpl`, but I don't know if it's "better" this way round.

Maybe @immortius has an opinion on this.

Cervator:
Bump. Just about caught up with PRs again, woohoo :-)

@Josharias - thought on update? For either of you: Should we go ahead and merge?

Josharias:
Hoping to hear from @immortius if he has any insight. It is not urgent, wait a handful more days?

Cervator:
Sure thing. I think we're about stable enough for another patch release to fix Cities etc but this doesn't seem critical to that effort :-)

immortius:
I agree with @msteiger assessment. It is problematic relying on calls to super for common behavior, so would either do as he has and put the behavior back in canvas, or would have a final draw() method in AbstractWidget that is called by canvas instead of onDraw() and which in turn calls the abstract onDraw() method, so that implementations cannot forget to include the behavior.

Cervator:
Thanks for the feedback! :)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/311/

Josharias:
Thanks for the code snippet mstieger.  Default tooltips support now in the CanvasImpl.

msteiger:
Sure thing. Looks good to me.

Cervator:
Merged! :-)

PR #2001, opened 2015-10-31
Move TeraEd to new facade project

msteiger:
Also replaces the `@EditorRange` annotation with the more general `@Range` annotation. Maybe it could be moved outside the NUI package at some point of time.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/305/

Cervator:
Works! Kinda :-)

I was able to tweak at the sky shader dynamically while in-game, making this dark and dreary, yet also sort of sunny scene:

terasology-151108011014-1280x720

Ping @emanuele3d - new toy! If you haven't had a chance to play with it much before. I made a run config for it in IntelliJ so it is easy to start. Just pick one of the shaders from the window menu to expose properties to play with.

I am not sure they all work, and the window doesn't dispose when you try to exit. Nor does it seem to respect -homedir, it used my local OS save dir instead. But it has been a long time since it was looked at so I'm sure it needs some more attention to behave again.

But this is a good first step! :-)

emanuele3d:
Ping acknowledged!

PR #2002, opened 2015-10-31
Deployment Diagram

gtugablue:


GooeyHub:

emanuele3d:
This issue is happening often. Is there anything we can do to make their
life easier? Some project setting variable we can keep empty and needs to
be configured manually on project setup to avoid defaulting to Terasology?

Cervator:
PRs default to the upstream repo you forked from. So in theory when a working group starts up on a fork with several personal forks beyond that if the personal forks are all coming from the "group home" fork then every PR created should default that way instead of MovingBlocks.

That probably comes up pretty rarely though, so likely isn't well-known. We're around the 2000th most forked repo on GitHub (yep, really! 300ish if just counting Java) so even a project with a network this large is rare in the first place. Then also having sub-groups work independently like this? Very rare. Which is kinda cool :-)

PR #2016, opened 2015-11-22
Use new Rect2i class

msteiger:
Requires changes in:

* AnotherWorld
* Behaviors
* Cities
* ComputerMonitors
* InGameHelp
* PolyWorld
* TutorialWorldGeneration


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/312/

Cervator:
Oho, more extractions. Will take good care of this one soon. Do we have any other major refactoring like this to do before we're fully on TeraMath, so to say? We should probably try to do that before going v1.0.0 in that case. Let me know :-)

msteiger:
This should be it.

There are still two versions of `Matrix4i` and a few minor issues I'd like to fix before `1.0`, but we're getting there.

Cervator:
Tested it out, tweaked a few more things. Seems to work well, need to do some rebuilding in Jenkins now to clear out old binaries :-)

For PolyWorld I'm ending up pulling in another commit to master but it seems benign. Cities I'll leave in develop since that's what builds in Jenkins anyway.

Made a reminder for further refactoring so I don't forget and forge ahead without it: #2019

Thanks!

Cervator:
Oh, @Josharias I added a commit: https://github.com/Terasology/JoshariasSurvival/commit/78f50134d0ba5752e30f026cc3ec03d515d49c5d - I dunno how that was in there, thought we would've cleared that out a while ago. FYI!

Issue #2014, opened 2015-11-20
org.terasology.registry.CoreRegistry get method inconsistency (possible software fault)


Sk0ut:
Whilst working on another issue, I found that the [CoreRegistry class](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/registry/CoreRegistry.java) static method get has an if statement but calls the same method independently of it.

Although the call returns the expected value because of checking the same condition tested in the refered if statement, it makes the return value of the get method dependent on the [Context class](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/context/Context.java) implementation of the get method for the argument Context.class.

Should the CoreRegistry get method return its Context when the argument is Context.class or simply pass the argument to the get method of its Context?

After the answer, I'll create a PR with the fix according to it.

Cervator:
@flo might know (ping!). It could be a temporary thing while Context is slowly phasing out CoreRegistry.

flo:
@Sk0ut: good catch!

The original intention was to return the context field of the CoreRegistry when asked for the Context.class. However it is no longer necessary to do so since the Context implementation called ContextImpl is doing now a similar trick by itself. e.g. a calll of  context.get(Context.class) returns the context itself.

So the get method of CoreRegistry could be simplified to just 
`return context.get(type);`

Sk0ut:
@flo
I thought about that alternative. However, if there was another class implementing the Context interface that didn't return itself when the argument was Context.class, wouldn't that break the functionality of the method get of the CoreRegistry class?

Cervator:
Bump - any final feedback @flo?

I'm happy to merge it as-is if the difference is minor :-)

flo:
well, the intention is to remove the CoreRegistry. It only exists for backward compability so that we have time to move from CoreRegistry to Context usage.

What we might want consider to do instead is merging the Context interface and Context implementation instead into a single class. I made them separate originally as I though Immoritus want it like that, but it turned out that Immoritus would have gone for a single class too.

Cervator:
I read that as "Sure, can merge as-is then go further in the next cleanup PR" :-)

Let me know if I'm off, will merge soon otherwise

Cervator:
Closing as completed by #2014 :-)

@flo if you're low on time yet have enough to help direct would it be something to keep our new friends engaged a bit longer with some architectural vision and ideas? And maybe a link or two to catch them up.

[...]

PR #2017, opened 2015-11-22
Bugfix #2014

Sk0ut:
Fix for issue #2014 assuming that the CoreRegistry class should act independently of its Context when the argument for the get method is Context.class

### DIFF
--- a/engine/src/main/java/org/terasology/registry/CoreRegistry.java
+++ b/engine/src/main/java/org/terasology/registry/CoreRegistry.java
@@ -18,7 +18,7 @@ package org.terasology.registry;
 import org.terasology.context.Context;
 
 /**
- * Registry giving access to major singleton systems, via the interface they fulfil.
+ * Registry giving access to major singleton systems, via the interface they fulfill.
  *
  * @author Immortius
  */
@@ -36,6 +36,8 @@ public final class CoreRegistry {
      * @param <T>
      */
     public static <T, U extends T> U put(Class<T> type, U object) {
+        if (context == null)
+            return null;
         context.put(type, object);
         return object;
     }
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/registry/CoreRegistry.java
```diff
@@ -36,6 +36,8 @@ private CoreRegistry() {
      * @param <T>
      */
     public static <T, U extends T> U put(Class<T> type, U object) {
+        if (context == null)
```
Cervator: Brackets please - part of our code convention to always use them :-)

Cervator:
Thanks! So many delicious unit tests, more are sorely needed, much appreciated :-)

Looks fine to me but then I'm not an architect, hoping @flo or @immortius could spare a minute to check this.

I noted one breach of our code convention for use of brackets. You can usually run Checkstyle with a plugin in an IDE on changed files and it'll tell you if anything clashes with our conventions. If something else in a file is out of order with our conventions and you feel confident in fixing it please feel free! Often it is simple stuff like brackets, whitespace, javadoc (for instance removing author tags) etc.

You can simply fix stuff in an additional commit on the same branch and GitHub will update this PR automatically. Handy!

Gooey: Test this please

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/313/

Cervator:
Thanks for the extra fixes :-)

Keeping an eye on #2014 in case anything else comes out of that discussion before we merge. If you do end up making any additional changes feel free to tweak the year in the license header on the test class, it says 2013 still :D But that's not really a big enough deal to fix on its own

Cervator:
Merged - thanks again, especially for the unit tests! Also added @Sk0ut to the credits, again if more people were involved just let me know. Or keeping making PRs and you'll all get in there eventually :D

While we might eventually discontinue the CoreRegistry this seems a good exercise on the way there, especially if it helps our friends in the Porto group :-)

PR #2046, opened 2015-12-09
Rename gradle run task for shader editor

### LINKED ISSUE #2045: Block or rename the "run" Gradle task for the TeraEd facade
If you execute "gradlew run" it'll actually hit both the task in the PC facade and the TeraEd facade, starting the game twice.

msteiger:
This is a more of a workaround rather than a proper fix. Maybe we should file a feature request over at gradle?

Adapted from http://stackoverflow.com/questions/28728488/rename-gradle-task

Fixes #2045

### DIFF
--- a/facades/TeraEd/build.gradle
+++ b/facades/TeraEd/build.gradle
@@ -1,4 +1,3 @@
-apply plugin: 'application'
 // The Editor facade is responsible for the (shader) editor - a plain Java application runnable on PCs
 
 // Grab all the common stuff like plugins to use, artifact repositories, code analysis config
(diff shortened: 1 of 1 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/333/

Cervator:
Seems fine to me. Sure would be nice if you could supply override values more cleanly, but if it works it works :-)

Alternatively, if I understand it right it is applying the `application` plugin that provides the run task in the first place. Yet TeraEd isn't really an application (or a distribution - which is implicitly applied for application). We could in theory just take off that plugin and make a Java Exec manually of a name we choose?

msteiger:
Yes, exactly. Maybe we should just do that... will have a look.

msteiger:
Done!

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/334/

Cervator:
Thanks :-)

Now we hit another fun quirk, which is a repeat from the past. `gradlew run` now became ambiguous, as it could be short for either `runPC` or `runEditor` - which is the exact same thing that happened when we first added the server task (`runServer` clashed with `runPC` and thus why it became `startServer`)

And that was a hack fix itself, what's the difference between "run" and "start" ? Completely arbitrary, it just worked. So finally I've taken a stab at naming that makes sense:

* `gradlew game` - starts the game. Replaces the `runPC` task
* `gradlew server` - starts the headless server. Replaces the `startServer` task
* `gradlew editor` - starts the editor (this PR)

Since "run" is taken by the application plugin we might as well try to stay away from anything custom starting with that. Less confusion. Starting after everybody realizes we changed names :-)

Oh but wait, couldn't editor one day clash with a world editor launched via Gradle? Well, I guess maybe, but it might be a thought to roll all that stuff into TeraEd. Make it the developer/modder "advanced" way to run the game to tweak anything.

We could add a `gradlew player2` task for launching a second client in its own dir as well, like the run config for that in IntelliJ. But since Gradle blocks until completion that's a little less useful - you'd need several terminals to pull it off. But it might be worth the little bit of utility if we want to throw that in one day.

Maybe `gradlew monitor` could launch with the advanced monitor or that could be in TeraEd as well, but I'm getting carried away ...

For now I'll merge this + the three new names and probably aim to do a release over this weekend as a checkpoint for December with the new names highlighted if they sound OK. Next weekend will likely be the final peak of craziness here at work then the next weekend after that I may be traveling and the weekend after *that* we release v1.0.0. Fingers crossed :-)

msteiger:
Sounds good to me. We could use `shaderEditor` for TeraEd since this is what's it all about, afaics.
Why is having multiple commands start with `run` a problem? More typing? Btw: gradle supports camel case completion like the in-game console. So `shEd` should resolve to `shaderEditor` (after pressing `ENTER`)

I guess we should add the new command definitions to the `README` file.

Cervator:
Oh, ya, totally need to adjust the readme! Thanks for reminding :-)

More typing, yup, can't abbreviate if the result would be ambiguous. Camel case shortening is cool, hadn't thought about that, but is probably less intuitive.

I'd love to see the editor go beyond shaders, becoming a true "developer facade", rather than end up with a bunch of specialized versions :-)

Cervator:
Edited both readme and wiki as part of the recent release. Highlighted the new names in the announcement thread. Hopefully that'll do it!

PR #2048, opened 2015-12-11
Replace menu title image

msteiger:
This is a very subjective change and up for discussion. Does it look good? Should it be larger or smaller? Maybe larger for the title and smaller for the other menu screens?
More space to the page content?

image


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/336/

Trekmarvel:
I've made a bit of font alternatives:
https://imgur.com/a/XOI1w

1 Mr Pixel
2 Neris
3 Eurostile
4 Cooperplate

Cervator:
Not much of a preference from me. I like the old one, the new one, and the ones by @Trekmarvel :-)

If anything I'd say I like a slightly unusual font and/or all caps.

Is the new image included here based on a specific font or drawn freestyle? Should we include a title font (that could be more unique - no need to internationalize) just so it can be text and resize/reshape itself easier? Or just put the font in TeraMisc if we ever need to tweak it?

msteiger:
Thanks for the suggestions, @Trekmarvel 
The image I used was made by @skaldarnar and taken from the splash screen. I prefer that one.
In any case I believe that the two should be consistent.

PR #2054, opened 2015-12-12
Translated using Weblate (Portuguese)

GooeyHub:
Currently translated at 100.0% (134 of 134 strings)


GooeyHub:

Issue #1568, opened 2015-02-01
add a way to unbind keybindings


4rz0:
Most games allow it to unbind keys, for when you don't need their function and don't want them to use space on your keyboard.
I suggest using *Escape* for that, at the moment, *Escape* just gets mapped as the keybind.

4rz0:
Seeing that the ingame menu, which is usually hardmapped to escape in most games, is actually a mappable function, it'd be a bad idea to use it to unbind keys.
What about having a button somewhere in the UI that you have to click to unbind keys?
Either one that unbinds the key that's active for mapping, or a button next to every function.

Cervator:
Another possible thought: If the user left clicks a button to make a key, start the mapping process and accept *any* valid input with no exception for unmapping. If the user *right* clicks a button reset its mapping and don't enter the mapping process at all?

Sk0ut:
@Cervator 
My group has taken interest in this issue, so I would like to solve some doubts before starting.
I already thinkered with the keybinds menu and found that this issue is still on.
So, my question is, should we not allow the right click to be mapped and use it as a cancel for the mapping, add a button to cancel the mapping (this one would make the UI bigger, because there's no space for a reset button next to each keybing) or clicking anywhere outside the keybing box cancel the mapping?

Sk0ut:
I realized right now that using the right click to cancel it's not a good idea, since the right click is associated with the action "Use Held Item", so right click should also be mappable

msteiger:
I think that `ESC` would be appropriate to cancel the process.

In any case, I suggest creating a new modal dialog that opens when the key is clicked on to indicate that the game is waiting for a new button binding. Pressing `ESC` will close the dialog so that would be in line with cancelling the binding. Maybe the dialog could also have additional options such as "reset default" or just a text label that states the default binding.

Side-note: Maybe also #1511 is of interest to you.

sceptross:
I also thought about that alternative, and I think it's the best way to do it, through implementing an additional button, instead of using the Esc key (because of what 4rzo previously said).

We'll also check out issue https://github.com/MovingBlocks/Terasology/issues/1511 and decide if we'll solve it after working on this one.

skaldarnar:
Instead of thinking of "mapable" buttons you could _activate_ the binding process with left mouse, and _remove_ an existing binding with right mouse. The binding process could then accept any key/button. 

The modal dialog looks like another good option.

Cervator:
[...]

PR #2059, opened 2015-12-13
Keybindings in Modal Window

Sk0ut:
Implementation of feature #1568 
Allows for new binding preview, remove binding and set it to default.
The layout needs to be worked a bit but its functional.


GooeyHub:

Issue #2066, opened 2015-12-16
Various causes for chunk generator threads dying


Cervator:
We've seen these for various reasons for quite a while now, but not really addressed them in a very focused way yet. I figured we should start tracking individual events and try to see what we can do to harden the system.

Scenario: Something in chunk generation causes an NPE and kills the generator thread. It seems possible for a repeat (rather than rare/inconsistent) instance to kill all threads at which point world generation stops.

Is there anything we can do to better guard against these problems? *Do* the threads indeed die and should we attempt to recycle them to avoid a slowdown of world gen? Or could that just make it worse?

Example - something related to the custom cactus block shape? Was running JoshariasSurvival with Caves enabled in Perlin. Happened a moment after I started moving (I assume when a chunk with a cactus attempted to generate). Race condition of some sort in loading assets?

```
22:54:17.336 [Chunk-Generator-5] INFO  o.t.w.b.internal.BlockManagerImpl - Registered BlockFamily[core:DarkLeaf]
22:54:17.336 [Chunk-Generator-5] INFO  o.t.w.b.internal.BlockManagerImpl - Registered Block core:DarkLeaf with id 336
22:54:18.933 [main] INFO  o.t.r.world.RenderableWorldImpl - New Viewing Distance: Far
22:54:23.182 [Chunk-Generator-0] INFO  o.t.w.b.internal.BlockManagerImpl - Registered BlockFamily[core:RedLeaf]
22:54:23.182 [Chunk-Generator-0] INFO  o.t.w.b.internal.BlockManagerImpl - Registered Block core:RedLeaf with id 337
22:55:24.102 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Creating game snapshot
22:55:24.128 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Snapshot created: Writing phase starts
22:55:25.949 [Saving-0] INFO  o.t.p.internal.SaveTransaction - Save game finished
22:55:55.958 [Chunk-Generator-7] INFO  o.t.w.b.internal.BlockManagerImpl - Registered BlockFamily[core:Cactus]
22:55:55.960 [Chunk-Generator-7] INFO  o.t.w.b.internal.BlockManagerImpl - Registered Block core:Cactus with id 338
22:55:56.157 [Chunk-Generator-3] ERROR o.t.u.concurrency.TaskProcessor - Error in thread Chunk-Generator-3
java.lang.NullPointerException: null
[...]

PR #1760, opened 2015-06-05
Changed all (short-lived) tasks to execute from a single thread pool.

prestidigitator:
* ScheduledExecutorService thread pool can be modified and retrieved through (context-reachable) TaskManager object.
* TaskMaster objects each submit a limited number of concurrent tasks to the pool.
* TaskMaster.shutdown() no longer needs a special termination task, and has extra options for waiting.
* Added "getClientThreads", "setClientThreads N", "getServerThreads", and "setServerThreads N" commands

Issue #1737, opened 2015-05-23
Reminder for disposal strategies for new Context system when ready


Cervator:
As discussed in #1727 - a reminder for when we're ready to work more specifically on disposal with the new Context concept. Possibility of contexts going out of scope - how to dispose what they reference, if at all? Maybe make an AutoCloseable of some sort available if desired?

Sort of related: #1296 & @emanuele3d has expressed interest in learning more later
##

immortius:
In general Context can hold any sort of object, so there's no defined interface for it to hook off of for cleanup.

Possibly the Context could have special handling of anything registered that implements Closable (rather than Autoclosable which is a special case), disposing them when the context itself is disposed.

pollend:
I think going forward I would like to adopt this: https://github.com/MovingBlocks/gestalt/tree/develop/gestalt-di

Issue #1296, opened 2014-10-18
Improve game entry/exit with more hooks for disposal + docs


Cervator:
This is in response to recent disposal issues (mainly #1266) and the idea of including System.exit to be sure to not leaving ghost processes around (#1292)

I'm reluctant to just dive in with plain System.exit to avoid dead processes when we don't have a more solid way to catch disposal issues. We're in much better shape now than a couple weeks ago, but might still have some issues floating around with more that could pop up later and hide behind a System.exit

I'd like to make sure we catch all (few) entry and exit points and make them work consistently. For instance the recent activity with the PC facade hasn't been tested with the Applet facade to my knowledge (and the AWT facade which @mkienenb was close to fixing). Probably the browser is hard-killing the applet, but anyway.

To go with this I think we should document entry and exit points as an engine feature with its own forum thread and wiki page. Along with what the few exit codes indicate. Maybe we can somehow log the state of disposal immediately before System.exit (new utility method?) so if there is still something left we have a chance to find out. Keeping in mind the game may be unstable (out of memory, etc). Maybe there are other good metrics we can write on shutdown?

Not in a big rush to get this done (thus target beta or sooner) as there are several things I'd like to wait on / take into account:
- Headless stable and loading/saving/disposing the same way as normal (#1049)
- Straighten out Facades better (mainly on me to straighten out how they build and bundle - in case of any major changes)
- Auto saves functional for both normal game and server - #477 (note: this is in progress by @flo and may conflict if this issue is worked - please wait for it to finish first!)
- Maybe #1234 - external asset system
- Multi world / multi-node servers (probably not - just figuring either would relate should they come up)
- Environment-based automated testing solid including in Jenkins (has main methods and use of resources)
- Reintroduction of the advanced monitor - #692 (currently if enabled fails to dispose)
##

Cervator:
Just to make a note of it - a few disposal issues have been sneaking back in slowly. I don't have a lot of details I just notice occasionally when crashing the game during testing.

@msteiger also mentioned an invalid port scenario over in #1622

prestidigitator:
@Cervator - Yes, the javadoc format works fine.  I can restyle it if you like.  No big deal.  No checkstyle error in the files I touched; I'm fastidious about that.  However, what I just did find is that javadoc complains loudly now about missing "@param" tags, so simple comments like:

   /** Equivalent to {@link #foo(int, int) foo(1, 2)} **/

[...]

PR #2073, opened 2015-12-18
Terminate on dying task processors

msteiger:
Currently, uncaught errors in `TaskProcessor` are forwarded to the thread's default `UncaughtExceptionHandler`, which just prints the stack trace.

As a result, the threads die silently and game stalls while waiting for chunks.

The change in `TerasologyEngine` fixes the missing stacktrace/error report in the log file.
The change in `TaskProcessor` looks awful, but it seems to be the cleanest way to terminate.

Since the `LocalChunkProvider` is disposed only when leaving the game, it still does not get disposed in that situation leaving the other thread pools alive.

PR #1760 would actually fix this issue, because the `ThreadPoolSubsystem` is disposed properly.

Related/partly fixed: #2066 and https://github.com/MovingBlocks/CrashReporter/issues/23

### DIFF
--- a/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
+++ b/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
@@ -367,7 +367,7 @@ public class TerasologyEngine implements GameEngine {
             changeState(initialState);
 
             mainLoop(); // -THE- MAIN LOOP. Most of the application time and resources are spent here.
-        } catch (RuntimeException e) {
+        } catch (Throwable e) {
             logger.error("Uncaught exception, attempting clean game shutdown", e);
             throw e;
         } finally {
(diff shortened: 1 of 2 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/350/

Cervator:
Awesome! Does this help any of our existing disposal issues at all, you think? My ReallyCrashGameBlock seems to be caught well enough to not leave a dead process, at least when running from source.

Is there any block we can fashion that'll kill chunk generator threads on purpose to better examine that sort of situation? Maybe something that crashes a thread on chunk unload? Pathfinding / L&S mobs might offer some inspiration ;-)

Linking for reference to #1737 and #1296

msteiger:
I triggered the situation by throwing an Error (not an Exception) such as NoSuchMethodError.

Cervator:
Yep that's what the ReallyCrashGame block does too (by trying to use a class not allowed by the security sandboxing). But that doesn't help crash chunk threads in a clean convenient fashion :-)

PR #2082, opened 2015-12-21
Allows to use multiple @DefaultBinding.

### LINKED ISSUE #1511: Enhance @DefaultBinding to support setting secondary key [$5]
Currently you can map a key binding like so:

`@DefaultBinding(type = InputType.KEY, id = Keyboard.KEY_G)`

However we support two mappings per function. If we could set two default keys in the annotation that would allow an easy way to "fix" for instance the situation with the console key being on `grave` which isn't a good key for some localized keyboards. We could simply set its default to `F1` *as well*

User can still remap to their heart's content in the Input Settings. But at least this way a working and familiar key should be available across just about any keyboard.

<bountysource-plugin>
---
There is a **[$5 open bounty](https://www.bountysource.com/issues/7648386-enhance-defaultbinding-to-support-setting-secondary-key?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

MaloJaffre:
Fixes #1511.
Now the config is sourced in code and not in default.cfg for the first startup.
Resolves conflicts between annotations and default config file.

### DIFF
--- a/engine/src/main/java/org/terasology/config/BindsConfig.java
+++ b/engine/src/main/java/org/terasology/config/BindsConfig.java
@@ -182,13 +182,11 @@ public final class BindsConfig {
 
     private void addBind(Name moduleName, Class<?> buttonEvent, RegisterBindButton info) {
         List<Input> defaultInputs = Lists.newArrayList();
-        for (Annotation annotation : buttonEvent.getAnnotations()) {
-            if (annotation instanceof DefaultBinding) {
-                DefaultBinding defaultBinding = (DefaultBinding) annotation;
-                Input input = defaultBinding.type().getInput(defaultBinding.id());
-                if (!data.values().contains(input)) {
-                    defaultInputs.add(input);
-                }
+        for (Annotation annotation : buttonEvent.getAnnotationsByType(DefaultBinding.class)) {
+            DefaultBinding defaultBinding = (DefaultBinding) annotation;
+            Input input = defaultBinding.type().getInput(defaultBinding.id());
+            if (!data.values().contains(input)) {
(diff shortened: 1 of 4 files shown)


GooeyHub:

msteiger:
ok to test

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/356/

Cervator:
Thanks! I checked it out and did some extra testing, applied a few tweaks on top: #2083 

F1 was indeed a key we wanted to use as secondary for the console, but it was currently in use for a debug panel (hard coded, not a bindable function yet). So I moved that to F2.

With two bindings possible the nice input scraper @msteiger just submitted recently didn't know what to do with multi-bound functions, so I sorted that out real quick - or thought it would be real quick, then 2-3 hours passed :D

Also edited the readme some.

@MaloJaffre you also managed to hit a first for our project! #1511 had a grand $5 county on it. We've never tried to have a bounty claimed yet, want to try it? :moneybag: If you don't want to worry about cashing out such an amount you can just roll it over to some other issue of your choice :-)

https://www.bountysource.com/issues/7648386-enhance-defaultbinding-to-support-setting-secondary-key?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github

Cervator:
Bounty has been paid out! That was easy :-)

PR #2086, opened 2015-12-22
Changes to Java 8 allowed structures to cleanup code.

MaloJaffre:
Removed 4 TODOs and refactor some code.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/LoggingContext.java
+++ b/engine/src/main/java/org/terasology/engine/LoggingContext.java
@@ -16,6 +16,9 @@
 
 package org.terasology.engine;
 
+import org.slf4j.MDC;
+import org.terasology.engine.modes.GameState;
+
 import java.io.IOException;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
@@ -26,10 +29,10 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Date;
-
-import org.slf4j.MDC;
-import org.terasology.engine.modes.GameState;
 
 /**
  * Configures the underlying logback logging framework.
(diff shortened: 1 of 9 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/360/

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/361/

engine/src/main/java/org/terasology/logic/console/Console.java
```diff
@@ -25,6 +25,9 @@
 /**
  */
 public interface Console {
+
+    String NEW_LINE = "\n";
```
msteiger: I guess this could be moved to an even more general interface - maybe `TerasologyContants`? There are many other places where `\n` was hardcoded and could be replaced by a proper constant.

engine/src/main/java/org/terasology/engine/LoggingContext.java
```diff
@@ -74,7 +77,7 @@ public static void initialize(Path logFileFolder) {
         System.setProperty(LOG_FILE_FOLDER, pathString);
 
         try {
-            deleteLogFiles(logFileFolder, 5 * 24 * 60 * 60); // JAVA8: Duration.ofDays(5).getSeconds();
+            deleteLogFiles(logFileFolder, ((int) Duration.ofDays(5).getSeconds()));
```
msteiger: Maybe `deleteLogFiles` could take a `long` parameter? This could avoid the need to downcast without reason.

MaloJaffre: Ok, I will do that now.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/362/

MaloJaffre:
I think ```TerasologyConstants.NEW_LINE``` is too long for a very commonly used char.
See https://gist.github.com/MaloJaffre/153fdaa0009b94256b71 for an example of why we want use this in all places.

msteiger:
I guess it depends on the scenario. The GL shader spec states that it supports any combination of `\r` and `\n`. On the other hand, console commands need to be handled consistently with the text renderer.

Maybe some indirection could make sense: the text renderer and the console could reference a common newline constant.

Anyway - it's good as it is now.

PR #2088, opened 2015-12-24
Create a new overloaded method for WorldProvider.setBlock(), which ta…

MaloJaffre:
…kes a Vector3f.

See https://trello.com/c/WqgcMjaq.

### DIFF
--- a/engine/src/main/java/org/terasology/world/internal/WorldProviderCore.java
+++ b/engine/src/main/java/org/terasology/world/internal/WorldProviderCore.java
@@ -17,6 +17,7 @@ package org.terasology.world.internal;
 
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.math.Region3i;
+import org.terasology.math.geom.Vector3f;
 import org.terasology.math.geom.Vector3i;
 import org.terasology.world.WorldChangeListener;
 import org.terasology.world.biomes.Biome;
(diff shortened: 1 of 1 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/363/

msteiger:
Thank you for your contribution. I don't think that this method should be in there, though.

It doesn't provide significant functionality and it would be inconsistent with the other getters/setters. Maybe we might even want to remove the `getBlock(Vector3f)` method.

@Cervator Where does the Trello requirement card come from?

Cervator:
It was initially raised in the Suggestion forum, in the backlog thread I just cleared out and turned into Trello cards, where this is https://trello.com/c/WqgcMjaq/6-add-overload-to-setblock-to-take-a-vector3f-and-convert-it-to-a-vector3i-automatically-like-getblock

I didn't think much of it, seemed like fair utility, but I don't know the code in the area. If it doesn't make sense we can close it out :-)

I've got some more stuff to move into Trello, when we're more stable there I can repost in the forum for some initial review of backlog items then enable Slack integration so any new cards will be more visible for review.,

MaloJaffre:
If this PR is useless, we can close this and the trello card.

Cervator:
Closed Trello card and am closing this PR unmerged. Sorry about the trouble @MaloJaffre - the initial import of issues into the backlog were just old entries from the forum that haven't had much review. When Trello is more fully set up we should do some review to disqualify anything else we don't actually want.

PR #2098, opened 2015-12-31
Update external versions

msteiger:
This updates:

* Gradle to 2.10
* VersionEye-integration to 1.10
* PMD and CheckStyle
* slf4j
* treelayout (used for the behavior editor)

Guava and GSON were removed as they are already specified in gestalt.


GooeyHub:

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/368/

msteiger:
Requires https://github.com/MovingBlocks/gestalt/pull/48

Cervator:
I'll roll this in with my first round of Gradle + infrastructure upgrades as both include a swap to Gradle 2.10. Which causes an Artifactory issue first introduced in Gradle 2.4 (we are/were at 2.3) that I'm not sure how to fix properly but there's an easy workaround.

Got JaCoCo 0.7.5 working too and reports updating on codecov using XML instead of the exec files. Jenkins still wants exec though, but at least it works with the latest versions of everything.

On Guava and GSON - if we require them shouldn't we to be proper keep them as direct dependencies rather than rely on the transitive dependencies from Gestalt? I know that makes it easier to stay in sync and we'll always have the dependency on Gestalt - it just seems iffy from a dependency management perspective :-)

Cervator:
Got this merged in and checked on. There was a weird error seemingly only in the new setup (upgraded Jenkins/Artifactory/plugins) but not in the old Jenkins or locally related to the slight upgrade of reflections at the build script level. Probably (and likely coincidentally) a conflict between transitive dependencies of Guava from different things or something. Maybe with the Artifactory Gradle plugin which I finally got hooked up more properly.

msteiger:
I amended the commit and added Guava and GSON back in.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/372/

PR #2104, opened 2016-01-02
Changes for Jenkins+Artifactory+Gradle upgrades and related

Cervator:
Pile of logistics changes.

* Artifactory server upgraded (infrastructure change) - also doubled memory to avoid crashes
* Jenkins server upgraded (infrastructure change)
* All Jenkins plugins upgraded and a few new ones added
* Gradle upgraded to 2.10 along with other version changes by @msteiger in #2098 (this PR includes it all)
* Gradle Artifactory plugin in particular upgraded, led to some more tweaks to have it work
 * Of particular note: `generatePomFileForMavenJavaPublication` has to be explicitly called now in a Gradle build step to create the `.pom` files for publishing successfully to Artifactory. This used to be an automatic dependency but for some reason doesn't even work when set as an explicit dependency now (the long task is a dynamic one so maybe it isn't available early enough)
* JaCoCo upgraded to 0.7.5 which is more or less a new major version which held us back in the past as other plugins/services weren't up to date yet, including Jenkins. Works now! Including codecov.io
* Various Jenkins config tweaking has been done and more needs to be done. Anybody who previously authorized OAuth for login via GitHub will get a little prompt to change (actually lessen) permissions granted.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/374/

build.gradle
```diff
@@ -25,20 +25,18 @@ import org.ajoberstar.gradle.git.tasks.*
 buildscript {
     repositories {
         mavenCentral()
-        maven {
-            url 'http://dl.bintray.com/jfrog/jfrog-jars'
-        }
+        jcenter()
```
msteiger: I think that jcenter includes the Maven Central repo.

build.gradle
```diff
@@ -25,20 +25,18 @@ import org.ajoberstar.gradle.git.tasks.*
...
     }
 
     dependencies {
         // Artifactory plugin
-        classpath(group: 'org.jfrog.buildinfo', name: 'build-info-extractor-gradle', version: '2.2.3')
+        classpath(group: 'org.jfrog.buildinfo', name: 'build-info-extractor-gradle', version: '4.0.0')
```
msteiger: This change might be relevant for gestalt, too.

msteiger: https://github.com/MovingBlocks/gestalt/blob/3df57b943eacebf06466c290a6737d1b582aed82/build.gradle#L157

build.gradle
```diff
@@ -51,13 +49,13 @@ buildscript {
  * 4. Comment 'wrapper' task
  */
 //task wrapper(type: Wrapper) {
```
msteiger: I think that the comment is no longer valid (at least not for gradle 2 and above) and the wrapper does not need to be commented out. It if is, the task still exists, but the version is derived from the used gradle version (I think..).

config/gradle/artifactory.gradle
```diff
@@ -4,24 +4,9 @@ apply from: "$rootDir/config/gradle/common.gradle"
 
 // Artifactory publishing requires these (without maven we'd get no .pom with dependency info published)
 apply plugin: 'maven'
-apply plugin: 'artifactory-publish'
+apply plugin: "com.jfrog.artifactory"
```
msteiger: Wrong ticks?

Cervator:
Oh sorry, I didn't see you adding comments, got sidelined in RL with the merge ready to push :-)

I'll go over the comments in the morning and do another round of tweaks soon after

build.gradle
Cervator: Good point! I still need to update some of the other projects and do a Gestalt and TeraMath release :(

So far just the engine job is updated plus the engine PR and the module template

build.gradle
Cervator: Oh handy. I just noticed it replacing the old plain URL with the latest artifactory plugin, didn't test deeper :-)

Cervator: Dug a little deeper. There seems to be a fair amount of hostility between Maven Central and jCenter, probably because their attached business entities are competitors. jCenter is supposedly proxying Maven Central but the details of how well (new packages not being available for a while) are apparently up for debate.

I'll make our config consistent with jcenter placed first everywhere since it sounds like they might have better performance etc.

build.gradle
Cervator: Righto! Took it out entirely. Sounds like we can do future upgrades either via command line argument or simply tweaking `gradle-wrapper.properties`

config/gradle/artifactory.gradle
Cervator: Fixed!

PR #2119, opened 2016-01-10
Add @FunctionalInterface  to all functional interfaces.

MaloJaffre:
This is only for documentation purposes.

### DIFF
--- a/engine/src/main/java/org/terasology/audio/AudioEndListener.java
+++ b/engine/src/main/java/org/terasology/audio/AudioEndListener.java
@@ -20,6 +20,7 @@ package org.terasology.audio;
  * <br><br>
  * Notice: The code in onAudioEnd is run in the update() method of the AudioManager once the sound/music ends playing.
  */
+@FunctionalInterface
 public interface AudioEndListener {
     void onAudioEnd();
 }
--- a/engine/src/main/java/org/terasology/engine/StateChangeSubscriber.java
+++ b/engine/src/main/java/org/terasology/engine/StateChangeSubscriber.java
@@ -18,6 +18,7 @@ package org.terasology.engine;
 /**
  * An interface for subscribers to engine state changes
  */
+@FunctionalInterface
 public interface StateChangeSubscriber {
 
     void onStateChange();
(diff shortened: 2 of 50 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/386/

Cervator:
Interesting. I wonder how many of these little extras are out there and we just haven't caught on to using them. In this case I see how this helps encourage adherence to the Java language specifications of what a functional interface is *but* there's a piece missing - educating our users on what it is, when to apply it, and how we can take more advantage (lambdas!) :D

I think this (and the `@Generated`) are good examples of something we should put into our coding conventions with a good explanation as to why they're there. So we need a spot to write about that plus including a link to https://docs.oracle.com/javase/8/docs/api/java/lang/FunctionalInterface.html in this case

Closest thing we have at present is http://forum.terasology.org/threads/terasology-code-style-conventions.501. I think we should leave that for discussion but put some convention doc itself on GitHub somewhere. Probably if we started using our /doc dir we could put it there and update the doc at the same time as adding a tweak like `@FunctionalInterface`

What do you think? I'm sure @msteiger would love to have a reason to put some stuff under /docs and I've been playing with the idea of automated documentation and an enhanced GitHub Page setup showing the docs nicely.

Not expecting that we'll set up any new document infrastructure before merging this, just going to think about it a little first then merge :-)

MaloJaffre:
I think annotations is a good and fast way to document code (eg. ```@ThreadSafe```,```@SafeVarargs```) and it can help to avoid ```NullPointerException``` (eg. ```@NotNull```, ```@Null```).
We can also create new documentation-only annotations, and add ```@Documented``` to those which can benefit to be added in javadoc when used.
I think we could provide use cases of annotations in /doc.

[Predefined Annotations (Oracle Tutorial)](http://docs.oracle.com/javase/tutorial/java/annotations/predefined.html). This is a good start to the most-used annotations.

Cervator:
Thanks for the extra link. I added a reminder to the main v1 doc issue so we'll remember this later if not right away :-)

PR #2124, opened 2016-01-10
Add Russian translation

msteiger:
Thanks Trams!


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/393/

Cervator:
Thanks indeed! :-)

@msteiger I noticed something sort of funny. The two "meta" fields that expect to contain the language itself in English and its native tongue can be misunderstood or left out, leading to confusion. I fixed it up for now in https://github.com/MovingBlocks/Terasology/commit/bc773b6710e254677601787f1190ad45d74262af

In one case I thought something was corrupt in a file as it looked like a label name was showing or even broken with two values, and we also had two English entries.

terasology _ alpha_2016-01-11_00-19-49

But on closer inspection the label names had been translated *verbatim* - rather than the language name itself in English and Portuguese :-)

funwithlabels

I wonder if we can pick a more descriptive label name for those two, or maybe mark them as special? 

With me having gone in and editing those files manually will they populate alright back in Weblate? For Japanese it was listing English/English which was rather confusing, now it is Japanese/Japanese pending a translation of "Japanese" - but with a value it might end up showing as translated in Weblate? If we simply need to push the latest branch to Weblate we should probably mark that translation as fuzzy or wrong or something, but we do sort of need to show something in the UI :-)

PR #2133, opened 2016-01-13
Decouple camera from the character

Josharias:
This is a polished version of the camera part of #2115.

The main part of this is the Gaze system.  Gaze is specific to a character (whether AI controlled or player controlled) and not to be confused with the camera. The gaze of a character (eyeball location) is different from the location than the character location (foot location).  Rotating and moving the gaze entity separately from the character decreases coupling between camera and character mechanisms allowing room for 3rd person cameras, tank turret-like camera rotation, and arbitrary mounting of the camera to other entities other than the character.

A diagram to help (now with arms and legs):
image

This commit is different from #2115 by:
- Not including any first person rendering refactoring.
- Reducing complexity of the gaze system by linking the gaze entity directly to the character entity.

Modules in omega affected by this commit:
- IRLCorp
- Miniion
- Minimap
- Pathfinding


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/402/

Cervator:
Excellent :-)

Will aim to test around some in the weekend. Maybe we should catch a good time and do some multiplayer testing together with a few of us.

Cervator:
Grabbed the branch for testing, game runs fine, but I realized I have no idea how to validate it otherwise :D

How did you put together the ponkey? Aka pony-monkey hybrid ...

That mutant may not be worth making the default avatar (@glasz we need you!) but we could put a test prefab for it in the Sample module :-)

Josharias:
Ponkey is by adding/changing the mesh for the character and gaze entity.

Kind of like these from the other PR:
[gaze.prefab](https://github.com/Josharias/Terasology/blob/6b0a34e7da489525e979965e16422af8acc463ec/engine/src/main/resources/assets/prefabs/player/gaze.prefab)
[player.prefab](https://github.com/Josharias/Terasology/blob/6b0a34e7da489525e979965e16422af8acc463ec/engine/src/main/resources/assets/prefabs/player/player.prefab)

Josharias:
I had a short list of things to test... but it got lost when Windows decided it wanted to install updates.

Something like:
- Look around (left/right, up/down)
- Walk around in all directions (forward, backward, strafe, walk in a circle)
- Ensure attacking a block targets what the cursor is on
- Ensure placing a block places where the cursor is at
- Ghost/Fly (I dont actually remember doing this late in the game)
- Throw an item in various directions

Cervator:
Yep! That works for testing alright. Eerie! :D

Merging away, thanks!

So what's next? I figure you have some notes on pending stuff like putting it to use. I seem to recall some bits of discussion on IRC. After taking the Gaze mesh back out the good ole monkeyheads don't move their gaze up and down, which makes sense since they aren't attached to the gaze - right? So we we need a character model with two parts, the main body and then a movable head? Which probably doesn't quite go the full 180 degrees from looking up to down :arrow_up_down: 

And I guess tweaks to the 4 modules affected

@glasz - ping!

Josharias:
Will get the broken modules fixed up asap. Sorry, got distracted by the next bit with the first person view.

PR #2135, opened 2016-01-17
Introducing FrameBuffersManager

emanuele3d:
With the work I did around six months ago, the class LwjglRenderingProcess had been progressively stripped of much of its functionality and classes such as the PostProcessor and GraphicState made their appearance.

This PR finally addresses what is left in the LwjglRenderingProcess class (FBO-related functionality) and refactors it into a FrameBuffersManager class.

The first [commit](https://github.com/MovingBlocks/Terasology/commit/ca92f309968a099f24c7649693de3c377ae9eb1d) simply deals with the immediate consequences of the renaming.
The second [commit](https://github.com/MovingBlocks/Terasology/commit/59e8d892aebbe8f0993a3d5ed3e0d0946651f48e) is where the "meat" is, as the newly renamed class is considerably refactored.
The third [commit](https://github.com/MovingBlocks/Terasology/commit/5f8db8fa341e32ebfc1721aa551ccfc529c4889e) mostly add javadocs, comments and TODOs, with some very minor refactoring alongside.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
+++ b/engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
@@ -51,9 +51,12 @@ public class FrameBuffersManager {
 
     private static final Logger logger = LoggerFactory.getLogger(FrameBuffersManager.class);
 
-    private PBO readBackPBOFront;
-    private PBO readBackPBOBack;
-    private PBO readBackPBOCurrent;
+    private PBO frontReadbackPBO;
+    private PBO backReadbackPBO;
+    private PBO currentReadbackPBO;
+
+    private FBO sceneShadowMap;
+    private FBO sceneOpaque;
 
     // I could have named them fullResolution, halfResolution and so on. But halfScale is actually
     // -both- fullScale's dimensions halved, leading to -a quarter- of its resolution. Following
@@ -65,9 +68,6 @@ public class FrameBuffersManager {
     private Dimensions one8thScale;
     private Dimensions one16thScale;
     private Dimensions one32thScale;
-
-    private int overwriteRtWidth;
-    private int overwriteRtHeight;
 
     private String currentlyBoundFboName = "";
     private FBO currentlyBoundFbo;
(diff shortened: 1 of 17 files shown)


engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
```diff
@@ -41,20 +41,35 @@
...
+ * <br/>
+ * The existing set of public methods is primarily intended to allow communication between this manager and other parts
+ * of the rendering engine, most notably the PostProcessor and the GraphicState instances and the shaders system.
+ * <br/>
+ * An important exception is the takeScreenshot() method which prompts the renderer to eventually (not immediately)
+ * redirect its output to a file. This is the only public method that is intended to be used from outside the
+ * rendering engine.
  */
-public class LwjglRenderingProcess {
+// TODO: should most of the public methods really be as much as possible package-private?
```
emanuele3d: This is a question for debate. Some of the public methods are used by other (nearby) packages, but a number of methods are used only within the same package and perhaps it would be good to close them to external use?

engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
```diff
@@ -41,20 +41,35 @@
...
  *
+ * In most instances Frame Buffers can be thought of as 2D arrays of pixels in GPU memory: shaders write to them or
+ * read from them. Some buffers are static and never change for the lifetime of the manager. Some buffers are dynamic:
+ * they get disposed and regenerated, i.e. in case the display resolution changes. Some buffers hold intermediate
+ * steps of the rendering process and the content of one buffer, "sceneFinal", is eventually sent to the display.
+ * <br/>
+ * At this stage no buffer can be added or deleted: the list of buffers and their characteristics is hardcoded.
+ * <br/>
+ * The existing set of public methods is primarily intended to allow communication between this manager and other parts
+ * of the rendering engine, most notably the PostProcessor and the GraphicState instances and the shaders system.
```
emanuele3d: Theoretically in this javadoc I could go in some detail regarding all the FBOs managed. Apart from the source code itself, those details are undisclosed. In practice the system is not really designed (yet!) to be used by, say, modules. As such I'm thinking that providing per-FBO information here is probably not necessary.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/403/

engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
```diff
@@ -78,98 +89,171 @@
...
 
-        FBO scene = fboLookup.get("sceneOpaque");
-        final boolean recreate = scene == null || (scene.dimensions().areDifferentFrom(fullScale));
+        return refreshed;
+    }
 
-        if (!recreate) {
-            return;
-        }
+    // providing a rough guide of each FBO here, as the method that creates them (recreateDynamicFBOs) is quite dense already
```
emanuele3d: This is perhaps a bit unorthodox, providing information in the method that disposes of the FBOs...

engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
```diff
@@ -180,43 +264,50 @@ public void createOrUpdateFullscreenFbos() {
...
-            GL11.glDeleteTextures(fbo.depthStencilTextureId);
-            GL11.glDeleteTextures(fbo.colorBufferTextureId);
+    private void handleDisposedShadowMap(FBO fbo) {
+        if (fbo.getStatus() == FBO.Status.DISPOSED) {
+            logger.warn("Failed to generate ShadowMap FBO. Turning off shadows.");
+            renderingConfig.setDynamicShadows(false);
         }
     }
 
+    // TODO: (?) move to the post-processor?
```
emanuele3d: Most of the screenshot-related methods could perhaps be moved into the post-processor, leaving in this manager only a method that returns the data held in the sceneFinal FBO as a ByteBuffer.

Cervator:
Good to have you back in action @emanuele3d ! :-)

Probably @msteiger and @Josharias are your best handy reviewers for this, so - ping!

As for public methods in general they're still closed to modules if not marked by `@API` - not sure that impacts the decision but thought I'd highlight it.

engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
```diff
@@ -252,109 +344,123 @@ public void saveScreenshot() {
             }
         };
 
-        CoreRegistry.get(ThreadManager.class).submitTask("Write screenshot", task);
-
+        threadManager.submitTask("Write screenshot", task);
```
emanuele3d: This line originally fetched the threadManager from the CoreRegistry and used it inline. This generated a warning as the result of the CoreRegistry.get() method could return null. The warning is no longer there but the possibility of a null threadManager remains. How defensively should code here? Should I add a check and log an error if the threadManager is null after all?

engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
msteiger: Agreed!

engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
msteiger: I'd leave it as it is.

I think that the ThreadManager should be injected with `@In` in the future. In any case, I guess this code will never be executed if the thread manager creation failed before.

emanuele3d: I'd be happy to learn how to use injection. Can you show me how to do it?

Josharias:
Looks very sensible the direction you are heading.  (I realize you had some more direction stuff last year that I didn't get around to checking out).  The naming changes you have done significantly improves understandability for me as a noob 3d wizard.  Also, the comments make these bunches of code approachable.  Well done.

engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
msteiger: The idea is that some classes are automatically instantiated, for example those who are annotated with `@RegisterSystem`. One way is to call a (default) constructor and fill all fields that are annotated with `@In` through reflection with variables from `CoreRegistry` or  or pass them to a constructor that takes them as parameters.

For now, I would add a field `threadManager` and manually fill it in the constructor to facilitate later refactoring, but that's just my opinion - you can as well just leave it as-is.

engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
Cervator: I see some brief comments on some of them elsewhere in the class, that seems pretty good anyway :-) Adding some general Javadoc like this is then excellent and sure one day in the future maybe more when modules can use them. We either need a nice spot to put some doc that can be pulled into the modding api docs automatically or maybe some small intro article/page somewhere (risks going out of date tho)

I guess one could go overkill on the docs a little bit and make a method to create/dispose each FBO *just* so you have a place to put the javadoc ... but I don't know if it is useful at that level (or something modules could hook into anyway)

engine/src/main/java/org/terasology/rendering/opengl/FrameBuffersManager.java
Cervator: Oh, right, what I just commented on above, heh. We *could* do little shell methods just to have a place for a nice description in javadoc. Some day.

Cervator:
Merges with the latest in develop (including the camera changes) just fine and I can't find anything terrible while testing. I'm good with merging whenever :-)

Hooray for documentation!

emanuele3d:
I'll do one more commit to remove a few TODOs and take in account the feedback. Give me until the weekend please. =)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/408/

emanuele3d:
I moved the screenshot functionality to the post-processor, but didn't do anything about construction-time instantiation/injection for things like the ThreadManager. I'll be happy to eventually go through all the top-level classes of the rendering engine and refactor this aspect. 

Meanwhile, and save new problems emerging, I think this PR is good to go.

Cervator:
Works for me! Thanks as always :-)

Issue #2064, opened 2015-12-15
Follow-up tweaks for input mappings (PR #2060)


Cervator:
I noticed some items during testing and wanted to highlight them for later work :-)

- [x] Layout on the new screen is a little funky (buttons not limited to the usual size - seems more like they're expanding to take up max space). Already mentioned as known in the PR
- [x] If you open the secondary mapping for a key the label for the key name is empty (primary mapping displays the label just fine)
- [x] Doesn't appear like any fields under input settings or the new input binding screen are being translated if a different language is selected (probably pre-existing issue?)
- [x] If the primary mapping for a key is removed the value from the secondary mapping moves into the primary mapping's spot. That's probably not desirable as you might well want to group primary/secondary keysets and this would mix them up
- [x] Reset to defaults sets mouse sensitivity to 0. Which means you can no longer look around in-game :-) (probably also pre-existing?)

msteiger:
I'm working on the remaining issues now.

msteiger:
What do you guys think of this layout?
![image](https://cloud.githubusercontent.com/assets/1820007/12257049/33e4c1fa-b903-11e5-9561-598561204f72.png)

MaloJaffre:
I think you can add a cancel button but this layout is good for me.

Cervator:
Yeah cancel button makes sense :-)

PR #2136, opened 2016-01-17
Redesign input settings menu and popup

msteiger:
This fixes the remaining issues in #2064.

Some more testing would be great to ensure that it works under all circumstances.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/404/

Cervator:
Seems to work fine and cover more binds, huzzah :-)

One suggestion: Switch the location of the "Cancel" and "Ok" buttons on the Input Settings popup. They're reversed from the more typical setup (affirmative first, then negative) which is also how the player settings currently work ("Ok", then "Cancel").

Also, I'm seeing a warning:

`WARN  o.t.rendering.nui.skin.UISkinFormat - Failed to resolve UIWidget class UIInputBindButton, skipping style information`

The UIInputBindButton class was deleted but still has an entry in default.skin. 

BindingScraper still works, but isn't catching as many binds. Easy way to fix?

And yay for default controller buttons!

Works in any case so merging, thanks :-)

PR #2157, opened 2016-02-06
Remove the nesting of items inside dropped items

Josharias:
The goal of this set of changes is to simplify how items are dropped into the world.  There was complexity when dealing with dropped items because the original item was nested inside of a new entity.  Code could then operate on the nested entity which could cause implications that did not reflect on the dropped entity in the world.  Normally I would say this would not be a  problem because the nested entity still maintained a link to the dropped entity by ownership, but it got more confusing when comparing a remote client to a local client because of replication.

This new way of dealing with items dropped into the world will maintain the same entity whether held by the player,  or dropped into the world.  The systems are able to do this by adding a location component (so that there is a mesh/location combo which causes it to be rendered), and a RigidBodyComponent (which causes gravity and collision to be enabled for the item).  These components are then removed again when the item is picked up by the player.

# Summary of changes
- Dropped items are no longer nested inside of a pickup entity.
- PickupBuilder has been removed in favor of ```DropItemEvent``` which does the necessary things to make a normal item into physical item in the world. (pending that there is already a ```MeshComponent``` and ```BoxShapeComponent```)
- Fixed the issue of block meshes being invisible sometimes on a remote client. (this has to do with the block meshes not being a real asset and not able to be looked up by URI.  These block meshes are generated client side and then subsequent calls to the asset will resolve)
- ```ItemComponent.pickupPrefab``` is no longer the template for a new entity,  but a reference of components to add/remove to an item when it is dropped/picked up to the world.

# Affected Modules
- GrowingFlora
- ModularComputers
- SimpleFarming
- WoodandStone

# Todo before merging
- [x] Placed and picked up chests a few times, contents seem solid. Weirdly sometimes I'd place a chest then be unable to place a block (TNT in this case) on one or more of the sides until it finally worked (in one case on the 4th adjacent spot I tried)
- [x] the guest logged in with both hand + pickaxe showing until I cycled through the tool bar some


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/415/

Cervator:
Sounds like an improvement to me! :-)

I see hits for `PickupBuilder` in ModularComputers, WoodAndStone, GrowingFlora, and SimpleFarming

Testing locally seems pretty good after some trouble getting properly reset with the deleted class (tried cleaning reflections but had to forcibly recompile some stuff). Although weirdly I placed the starter chest and the first three times i `e` clicked it I got denied, then the 4th try worked. Got:

`[main] INFO  o.t.logic.characters.CharacterSystem - Denied activation attempt by Cervator since at the authority another entity would have been activated`

Seems like I can sometimes provoke that just trying to `e` click the ground.

Placed and picked up chests a few times, contents seem solid. Weirdly sometimes I'd place a chest then be unable to place a block (TNT in this case) on one or more of the sides until it finally worked (in one case on the 4th adjacent spot I tried)

Sometimes even just trying to place a torch on the ground gets the same log entry + denial to actually place.

Weirdest part is that I'm just playing in single player so I am the authority.

Tried it in multiplayer and both host and guest got the same issue inconsistently. For bonus points the guest logged in with both hand + pickaxe showing until I cycled through the tool bar some :-)

terasology-160207014231-1152x720

Josharias:
Thanks for the tests Cervator.  The "denied activation attempt" has to do with the physics colliding with the initiating character.  There is definitely some thought needed there, since the gaze system, because we are no longer raycasting from feet,  but from eyes.

I have added a todo list so we can track progress.

Josharias:
RE: unable to place a block
Cervator,  by chance were you right clicking on the side of the chest?  The chest specifically does not allow attachment,  so it prevents the normal right click block placing.  I wonder if between this and the raycasting hit on the character's body were making it seem impossible to place the TNT?

Josharias:
RE:  Hand + pickaxe showing
@Cervator: How did you get the hand and pick combo?  I have not been able to replicate on my local.  Any more tips on how to recreate?

Also,  Core's door is broken as of this PR.  I have it fixed in the move of inventory and health systems to Core.

Cervator:
Block placing: Actually later on it was intermittently *any* block placement. Didn't have to be near a chest, didn't have to be TNT blocks, etc. Even placing torches etc wouldn't work for a few tries, then it would work again. Maybe something you can do that triggers it, then it goes away after a bit.

Hand/pickaxe overlap: no idea :( i was just randomly switching items, picking up stuff, dropping stuff, and so on as the hosting player. Then I joined with a second player on a non-headless server and then it happened - maybe something from the hosting player's session/entity stuff "leaked" into the new joining client?

flo:
When a torch placment near a recently placed chest does not work then the following line gets logged each itme it gets tried:

> Denied activation attempt by flo since at the authority another entity would have been activated

Josharias:
Yeah,  that is the problem of raycasting from eye position (as opposed to the foot position like before) is still hitting the local player.  I wouldnt mind another brain to help figure out what is the correct way to fix that.  I had guessed that one could exclude some entities from being selected from the raycast.  But that delved way deeper into the physics system than I wanted to go.

flo:
In my case the issue is that the method activateTargetOrOwnedEntity of LocalPlayer determines that the character got hit instead of the ground next to the chest.

flo:
The method isPredictionOfEventCorrect of CharacterSystem determines that the ground got hit for the same ray cast. That is what the message is saying

Josharias:
Hrm.  Looking again.  Appears I took out the Character CollisionGroup from CharacterSystem,  but not LocalPlayer. 

Haha.  In LocalPlayer there is this ```// TODO use same as CharacterSystem?```
and in CharacterSystem there is this ```* TODO: Include the Character collision group```

Yup,  they just need to play nicer with each other.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/423/

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/426/

Cervator:
Seems better now :-)

The door is still a bit broken as you mentioned elsewhere. I found another couple quirks but probably unrelated to this, will retest / report more later.

PR #2164, opened 2016-02-10
Fix camera linking to a respawned character.

Josharias:
Rethought the way the events for this work.  This separates the mounting from the resetting of the camera a bit better.  Where a camera reset actually recreates the client side camera entity,  which naturally tiggers a mount.  Also,  force a camera reset when the character respawns.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
@@ -29,6 +29,7 @@ import org.terasology.logic.console.commandSystem.annotations.Command;
 import org.terasology.logic.location.Location;
 import org.terasology.logic.location.LocationComponent;
 import org.terasology.logic.permission.PermissionManager;
+import org.terasology.logic.players.event.OnPlayerSpawnedEvent;
 import org.terasology.logic.players.event.ResetCameraEvent;
 import org.terasology.math.geom.Quat4f;
 import org.terasology.math.geom.Vector3f;
@@ -43,19 +44,27 @@ public class CameraClientSystem extends BaseComponentSystem {
     EntityManager entityManager;
 
     @ReceiveEvent
-    public void onAutoMountCamera(OnChangedComponent event, EntityRef client, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
-        if (clientComponent.camera.exists()) {
-            client.send(new ResetCameraEvent());
+    public void ensureCameraEntityCreatedOnChangedClientComponent(OnChangedComponent event, EntityRef client, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
+        if (localPlayer.getClientEntity().equals(client)) {
+            ensureCameraEntityCreated();
         }
     }
 
     @ReceiveEvent
-    public void ensureCameraEntityCreated(OnActivatedComponent event, EntityRef entityRef, ClientComponent clientComponent) {
-        if (!clientComponent.camera.exists()) {
+    public void ensureCameraEntityCreatedOnActivateClientComponent(OnActivatedComponent event, EntityRef entityRef, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
+        if (localPlayer.getClientEntity().equals(entityRef)) {
+            ensureCameraEntityCreated();
+        }
+    }
+
+    void ensureCameraEntityCreated() {
+        if (!localPlayer.getCameraEntity().exists()) {
+            ClientComponent clientComponent = localPlayer.getClientEntity().getComponent(ClientComponent.class);
             EntityBuilder builder = entityManager.newBuilder("engine:camera");
             builder.setPersistent(false);
             clientComponent.camera = builder.build();
-            entityRef.saveComponent(clientComponent);
+            localPlayer.getClientEntity().saveComponent(clientComponent);
+            mountCamera();
         }
     }
 
@@ -65,7 +74,18 @@ public class CameraClientSystem extends BaseComponentSystem {
     }
 
     @ReceiveEvent
-    public void resetCamera(ResetCameraEvent resetCameraEvent, EntityRef client, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
+    public void resetCamera(ResetCameraEvent resetCameraEvent, EntityRef entityRef, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
+        if (localPlayer.getClientEntity().equals(entityRef)) {
+            clientComponent.camera.destroy();
+            clientComponent.camera = EntityRef.NULL;
+            // this will trigger a ClientComponent change which will in turn trigger a remount
+            localPlayer.getClientEntity().saveComponent(clientComponent);
+        }
+    }
+
+    @Command(shortDescription = "Mounts the camera to the character", requiredPermission = PermissionManager.NO_PERMISSION)
+    public void mountCamera() {
+        ClientComponent clientComponent = localPlayer.getClientEntity().getComponent(ClientComponent.class);
         EntityRef targetEntityForCamera = GazeAuthoritySystem.getGazeEntityForCharacter(clientComponent.character);
         LocationComponent cameraLocation = clientComponent.camera.getComponent(LocationComponent.class);
         Vector3f cameraPosition;
@@ -75,5 +95,12 @@ public class CameraClientSystem extends BaseComponentSystem {
             cameraPosition = cameraLocation.getLocalPosition();
         }
         Location.attachChild(targetEntityForCamera, clientComponent.camera, cameraPosition, new Quat4f(Quat4f.IDENTITY));
+    }
+
+    @ReceiveEvent
+    public void resetCameraOnCharacterSpawn(OnPlayerSpawnedEvent event, EntityRef character) {
+        if (localPlayer.getCharacterEntity().equals(character)) {
+            resetCamera();
+        }
     }
 }
--- a/engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
@@ -36,6 +36,23 @@ import org.terasology.math.geom.Vector3f;
 import org.terasology.network.ClientComponent;
 import org.terasology.registry.In;
 
+/**
+ * This is a system that creates and maintains a client side entity for the camera.
+ * <p>
+ * Entity Creation Flow:
+ * - ClientComponent for the local player added or changed
+ * - If Camera does not exist, create "engine:camera" entity
+ * - Save the ClientComponent with the new camera attached
+ * <p>
+ * Reset Camera Event Flow:
+ * - ResetCameraEvent
+ * - Camera entity on the ClientComponent is destroyed and set to EntityRef.NULL
+ * - ClientComponent is saved
+ * <p>
+ * Auto Mount Camera Flow:
+ * - ClientComponent or AutoMountCameraComponent for the local player added or changed
+ * - The local player's camera entity (created from above steps) is location linked to the Gaze Entity
+ */
 @RegisterSystem(RegisterMode.CLIENT)
 public class CameraClientSystem extends BaseComponentSystem {
     @In
@@ -44,27 +61,34 @@ public class CameraClientSystem extends BaseComponentSystem {
     EntityManager entityManager;
 
     @ReceiveEvent
-    public void ensureCameraEntityCreatedOnChangedClientComponent(OnChangedComponent event, EntityRef client, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
+    public void ensureCameraEntityCreatedOnChangedClientComponent(OnChangedComponent event, EntityRef client, ClientComponent clientComponent) {
         if (localPlayer.getClientEntity().equals(client)) {
             ensureCameraEntityCreated();
         }
     }
 
     @ReceiveEvent
-    public void ensureCameraEntityCreatedOnActivateClientComponent(OnActivatedComponent event, EntityRef entityRef, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
-        if (localPlayer.getClientEntity().equals(entityRef)) {
+    public void ensureCameraEntityCreatedOnActivateClientComponent(OnActivatedComponent event, EntityRef client, ClientComponent clientComponent) {
+        if (localPlayer.getClientEntity().equals(client)) {
             ensureCameraEntityCreated();
         }
     }
 
-    void ensureCameraEntityCreated() {
+    private void ensureCameraEntityCreated() {
         if (!localPlayer.getCameraEntity().exists()) {
             ClientComponent clientComponent = localPlayer.getClientEntity().getComponent(ClientComponent.class);
+            // create the camera from the prefab
             EntityBuilder builder = entityManager.newBuilder("engine:camera");
             builder.setPersistent(false);
             clientComponent.camera = builder.build();
             localPlayer.getClientEntity().saveComponent(clientComponent);
-            mountCamera();
+        }
+    }
+
+    @ReceiveEvent
+    public void resetCameraOnCharacterSpawn(OnPlayerSpawnedEvent event, EntityRef character) {
+        if (localPlayer.getCharacterEntity().equals(character)) {
+            resetCamera();
         }
     }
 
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
```diff
@@ -65,7 +74,18 @@ public void resetCamera() {
...
+    public void resetCamera(ResetCameraEvent resetCameraEvent, EntityRef entityRef, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
+        if (localPlayer.getClientEntity().equals(entityRef)) {
+            clientComponent.camera.destroy();
+            clientComponent.camera = EntityRef.NULL;
+            // this will trigger a ClientComponent change which will in turn trigger a remount
+            localPlayer.getClientEntity().saveComponent(clientComponent);
+        }
+    }
+
+    @Command(shortDescription = "Mounts the camera to the character", requiredPermission = PermissionManager.NO_PERMISSION)
```
emanuele3d: Premise: I know next to nothing on how cameras work in Terasology. I see where they plug into the rendering engine but that's about it.

That been said, what do you have in mind with the creation of this command? Is it some kind of debugging tool or should users and systems eventually be able to use it?

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
```diff
@@ -76,4 +96,11 @@ public void resetCamera(ResetCameraEvent resetCameraEvent, EntityRef client, Aut
         }
         Location.attachChild(targetEntityForCamera, clientComponent.camera, cameraPosition, new Quat4f(Quat4f.IDENTITY));
```
emanuele3d: I know you haven't touched this code, but I'm wondering if we can perhaps remove the Vector3f cameraPosition declaration and just place two Location.attachChild() calls within the IF statement. Also, is cameraLocation == null the expected typical case? If not I'd recommend switching the condition to cameraLocation != null.

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
```diff
@@ -43,19 +44,27 @@
...
 
     @ReceiveEvent
-    public void onAutoMountCamera(OnChangedComponent event, EntityRef client, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
-        if (clientComponent.camera.exists()) {
-            client.send(new ResetCameraEvent());
+    public void ensureCameraEntityCreatedOnChangedClientComponent(OnChangedComponent event, EntityRef client, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
+        if (localPlayer.getClientEntity().equals(client)) {
+            ensureCameraEntityCreated();
         }
     }
```
emanuele3d: I like your very verbose method names. Very clear! They are almost entirely self-documenting! 

I wonder if some javadocs for class and public methods would still be useful though. Perhaps offering some context to what's going on here?

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
```diff
@@ -43,19 +44,27 @@
...
     @ReceiveEvent
-    public void ensureCameraEntityCreated(OnActivatedComponent event, EntityRef entityRef, ClientComponent clientComponent) {
-        if (!clientComponent.camera.exists()) {
+    public void ensureCameraEntityCreatedOnActivateClientComponent(OnActivatedComponent event, EntityRef entityRef, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
+        if (localPlayer.getClientEntity().equals(entityRef)) {
+            ensureCameraEntityCreated();
+        }
+    }
+
+    void ensureCameraEntityCreated() {
```
emanuele3d: Is this method intentionally package-private?

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
Josharias: Correct, it is a debugging tool.  But not one that causes any bad effects as it will only touch client side entities.  I dont exactly know what cases one will need to re-mount the camera as opposed to resetting the camera.  Do you think I should avoid making it a command so that there is less command noise?

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
Josharias: I thought it might be more clear that the camera position is the only thing that could be different when there is no existing location.  That is why there is only one call to attachChild.   Let me know if you think it would be more readable/understandable as two different calls to attach child.
I am not sure what case would be more likely "cameraLocation == null" or "cameraLocation != null",  specially because there is no guarantee that the camera entity will come with a LocationComponent.

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
emanuele3d: Yes, that's what I was wondering about. But perhaps even just as a debug tool it might be worth keeping it. Perhaps we could put an expiration date on it, i.e. removing it after a number of releases so that we can use it for testing purposes while you are working on cameras and some time after.

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
emanuele3d: Regarding the 1 vs 2 attachChild calls: my thinking is that less code to read (in such a simple case) is less code to understand, hence my suggestion. But I'm not sure what the purists of code would suggest in this case. Another way to tackle this, if there was a condition more likely than the other, would be to initialize the cameraPosition with the typical value and then handle the more atypical case in an IF block without ELSE, leaving a single attachChild call. 

But you mention there is no guarantee that the camera entity will come with a LocationComponent. Why is that the case? I mean, what good would a camera be without LocationComponent? Or are you referring to cameras being in-transition between one entity and another?

emanuele3d:
@Cervator: I just noticed the 3D-Wizardry tag. I'm not sure that's accurate. This PR is pretty much all about the EntitySystem and its component. I don't think there is much 3D involved...

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/419/

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
Josharias: I do see your point of reducing the lines of code.  Thank you for your perspective.

The camera entity gets created from a prefab.  Theoretically a module could remove the location component from the prefab using a delta. That is how there is no guarantee.

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
Josharias: Agreed.  Ill add a class description.  I was hoping the method names would be enough.  :)

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
Josharias: Good catch.  C# habits coming through.

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
Josharias: <Chuckle>  expiration date!  I think I will take out the mount command as it will naturally happen on a reset camera.

Cervator:
@emanuele3d good point. I guess I just throw the label at anything that remotely seems related :D

Switching to Architecture.

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
Cervator: "Expiration date" could be simply adding `@deprecated` then we can just tolerate it while testing, but will be continuously reminded it shouldn't be a permanent thing :D

emanuele3d: I like the idea. I wonder if we could get one of the analisys tool to remind us how long each deprecated statement has been in the code.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/420/

Josharias:
Documentation and further refinement completed.  @Cervator would you mind sanity checking these changes in game?

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
```diff
@@ -29,12 +29,30 @@
...
+ * Reset Camera Event Flow:
+ * - ResetCameraEvent
+ * - Camera entity on the ClientComponent is destroyed and set to EntityRef.NULL
+ * - ClientComponent is saved
+ * <p>
+ * Auto Mount Camera Flow:
+ * - ClientComponent or AutoMountCameraComponent for the local player added or changed
+ * - The local player's camera entity (created from above steps) is location linked to the Gaze Entity
+ */
 @RegisterSystem(RegisterMode.CLIENT)
```
emanuele3d: I feel this class-level javadoc is like the New Horizon probe swinging by Pluto: it created more questions than it answered! =D

I guess one question mark buzzing over my head as I read this javadoc is: what is actually happening in this class/system and what is happening elsewhere? The idea of describing flows is good but the actions of this system and those of collaborating code are blurred together. This could be solved by qualifying some of the statements with a subject, i.e.:

- System receives an added/changed ClientComponent event
- If the ClientComponent has no camera this system creates and attaches an "engine:camera" entity
- Finally, the system triggers the saving of the ClientComponent.

Notice how in this example I try to clarify who is doing what. I.e. in the first line it's clear it is not the system that added or changed the ClientComponent - the system merely listens for such change. Then, in the third line by using the word "triggers" I imply that this system doesn't actually do the saving, it is collaborating code that does.

The other thing I'd like to mention is that by reading this javadoc I wonder: in what cases these flows arise? Perhaps this is detailed elsewhere in the codebase and nobody would ever see this class out of context like I am doing now. But as I am reading this out of context, I do wonder: in what circumstance the camera is created? And in what circumstances does it get reset?

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
```diff
@@ -43,19 +61,34 @@
...
+    public void ensureCameraEntityCreatedOnChangedClientComponent(OnChangedComponent event, EntityRef client, ClientComponent clientComponent) {
+        if (localPlayer.getClientEntity().equals(client)) {
+            ensureCameraEntityCreated();
         }
     }
 
     @ReceiveEvent
-    public void ensureCameraEntityCreated(OnActivatedComponent event, EntityRef entityRef, ClientComponent clientComponent) {
-        if (!clientComponent.camera.exists()) {
+    public void ensureCameraEntityCreatedOnActivateClientComponent(OnActivatedComponent event, EntityRef client, ClientComponent clientComponent) {
```
emanuele3d: Did you mean "...OnActivate**d**ClientComponent"?

Also, part of me would like "...OnClientComponentActivation" better but I can see you are being consistent with the OnActivatedComponent class, which is perhaps the source of the problem...

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
```diff
@@ -65,15 +98,38 @@ public void resetCamera() {
     }
```
emanuele3d: Something I simply don't know rather than a problem with the code: once a command is registered one has to bind it to a string or a key to use it respectively in a console or via keyboard and that is done elsewhere. Is that how it works?

emanuele3d: Also: why do you destroy the camera to reset its position, rather than, say, just change its position to some kind of default?

Josharias: The default is preferably specified in the camera.prefab.  Instead of grabbing the prefab definition, and then cloning a new component on to the existing entity.  I thought it would be cleaner to just wipe it and start over so that there is only one code path possibility for how the location default is retrieved instead of two.

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
Josharias: Correct. These descriptions are vague.  My intent was to generally outline what the code does so that there is a general idea of what to look for when perusing the code.
Possibly I have taken the summary too far in an attempt to improve skimmability by reducing word count and putting the target of the step at the front... Dunno.

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
```diff
@@ -65,15 +98,38 @@ public void resetCamera() {
...
+
+    @ReceiveEvent
+    public void mountCameraOnActivate(OnActivatedComponent event, EntityRef entityRef, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
+        if (localPlayer.getClientEntity().equals(entityRef) && clientComponent.camera.exists()) {
+            mountCamera();
+        }
+    }
+
+    @ReceiveEvent
+    public void mountCameraOnChange(OnChangedComponent event, EntityRef entityRef, AutoMountCameraComponent autoMountCameraComponent, ClientComponent clientComponent) {
```
emanuele3d: As I read the names of both _mountCameraOnActivate_ and _mountCameraOnChange_ methods I wonder: what is activated? What is changed? Can you clarify with verbose names or javadocs?

Also, the code inside both methods is identical. How about we shorten it by removing the if statement from both and by passing the condition in one or two arguments to mountCamera, i.e.:

mountCamera(localPlayer.getClientEntity().equals(entityRef)<sup>1</sup>, clientComponent.camera.exists());

We would then put the IF statement in mountCamera but in a more readable form. I.e. 

```java
private void mountCamera(boolean weAreDealingWithTheRightEntity, boolean aCameraAlreadyExist) {
    if(weAreDealingWithTheRightEntity && aCameraAlreadyExist) { // do the magic }
}
```

<sup>1</sup>I don't actually understand what this does.

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
Cervator: @emanuele3d: As far as the command/keyboard question goes the `@Command` annotation itself registers the command as available in the game console by its method name, no additional action needed. Binding it to a hotkey would have to be done separately, which is another annotation to define the hotkey then simply have its execution run the command's method :-)

Cervator:
Looking better! A simple respawn right near I started (just ghosted up in the air then fell to my death) worked just fine.

However flying away a fair distance from Spawn so the chunks would unload left me respawning in a void with a game crash following soon after:

```
01:19:27.782 [main] INFO  o.t.logic.console.ConsoleImpl - [CONSOLE] Ghost mode toggled
01:19:28.647 [main] INFO  o.t.logic.console.ConsoleImpl - [CONSOLE] High-speed mode activated
01:19:34.143 [main] INFO  o.t.logic.console.ConsoleImpl - [CONSOLE] Ghost mode toggled
01:19:56.700 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Creating game snapshot
01:19:56.702 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Snapshot created: Writing phase starts
01:19:59.132 [Saving-0] INFO  o.t.p.internal.SaveTransaction - Save game finished
01:20:17.344 [Timer-1] ERROR o.t.r.logic.NearestSortingList - Uncaught exception in sorting thread: java.lang.ArrayIndexOutOfBoundsException: 33
01:20:56.746 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Creating game snapshot
01:20:56.747 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Snapshot created: Writing phase starts
01:20:58.245 [Saving-0] INFO  o.t.p.internal.SaveTransaction - Save game finished
01:21:42.052 [main] INFO  o.t.logic.console.ConsoleImpl - [CONSOLE] High-speed mode activated
01:21:43.018 [main] INFO  o.t.logic.console.ConsoleImpl - [CONSOLE] Ghost mode toggled
01:21:53.114 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Creating game snapshot
01:21:53.117 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Snapshot created: Writing phase starts
01:21:57.056 [Saving-0] INFO  o.t.p.internal.SaveTransaction - Save game finished
01:21:59.519 [main] INFO  o.t.logic.console.ConsoleImpl - [CONSOLE] Ghost mode toggled
01:22:05.008 [main] ERROR o.terasology.engine.TerasologyEngine - Uncaught exception, attempting clean game shutdown
	at org.terasology.logic.players.FirstPersonClientSystem.update(FirstPersonClientSystem.java:191) ~[classes/:na]
	at org.terasology.engine.modes.StateIngame.update(StateIngame.java:176) ~[classes/:na]
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:414) ~[classes/:na]
	at org.terasology.engine.TerasologyEngine.run(TerasologyEngine.java:369) ~[classes/:na]
	at org.terasology.engine.Terasology.main(Terasology.java:154) [classes/:na]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_45]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_45]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_45]
	at java.lang.reflect.Method.invoke(Method.java:497) ~[na:1.8.0_45]
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:144) [idea_rt.jar:na]
01:22:05.008 [main] INFO  o.terasology.engine.TerasologyEngine - Shutting down Terasology...
01:22:05.511 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Creating game snapshot
01:22:05.512 [main] INFO  o.t.p.i.ReadWriteStorageManager - Saving - Snapshot created: Writing phase starts
01:22:05.516 [main] INFO  o.t.n.internal.NetworkSystemImpl - Client disconnected: Cervator
01:22:05.516 [main] INFO  o.t.logic.console.ConsoleImpl - [NOTIFICATION] Player "Cervator" has left the game
01:22:05.516 [main] INFO  o.t.n.internal.NetworkSystemImpl - Network shutdown
01:22:07.160 [Saving-0] INFO  o.t.p.internal.SaveTransaction - Save game finished
01:22:07.161 [main] WARN  org.reflections.Reflections - given scan urls are empty. set urls in the configuration
01:22:09.875 [main] WARN  o.t.rendering.nui.skin.UISkinFormat - Failed to resolve UIWidget class UIInputBindButton, skipping style information
01:22:09.896 [main] WARN  o.t.rendering.nui.asset.UIFormat - Field 'emptyIcon' not recognized for interface org.terasology.rendering.nui.UIWidget in {"type":"UIIconBar","id":"breathBar","icon":"engine:icons#bubble","family":"breathBar","emptyIcon":"engine:icons#burstBubble","halfIconMode":"shrink","spacing":2,"maxIcons":10,"layoutInfo":{"use-content-width":true,"use-content-height":true,"position-left":{"target":"CENTER"},"position-right":{"target":"RIGHT","widget":"toolbar"},"position-bottom":{"target":"TOP","widget":"toolbar","offset":1}}}
01:22:09.896 [main] WARN  o.t.rendering.nui.asset.UIFormat - Field 'crosshairIcon' not recognized for interface org.terasology.rendering.nui.UIWidget in {"type":"UICrosshair","id":"crosshair","crosshairIcon":"engine:gui#crosshair","chargeStages":["engine:gui#crosshairCharge1","engine:gui#crosshairCharge2","engine:gui#crosshairCharge3","engine:gui#crosshairCharge4","engine:gui#crosshairCharge5","engine:gui#crosshairCharge6","engine:gui#crosshairCharge7","engine:gui#crosshairCharge8"],"layoutInfo":{"use-content-width":true,"use-content-height":true,"position-horizontal-center":{},"position-vertical-center":{}}}
java.lang.NullPointerException
	at org.terasology.logic.players.FirstPersonClientSystem.update(FirstPersonClientSystem.java:191)
	at org.terasology.engine.modes.StateIngame.update(StateIngame.java:176)
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:414)
	at org.terasology.engine.TerasologyEngine.run(TerasologyEngine.java:369)
	at org.terasology.engine.Terasology.main(Terasology.java:154)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:144)
```

engine/src/main/java/org/terasology/logic/players/CameraClientSystem.java
emanuele3d: @Cervator: A-ha! It registers it under the method name! Crucial information I was missing. Thank you.

Cervator: Any time! Yep it is good to strike a balance between magic convenience and dark arcane wizardry :D

Josharias:
Hah.  I had this bug squashed in a different branch.  Serves me right for having more than one stream of development.

NPE fixed when respawning back into an unloaded chunk.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/427/

Cervator:
Tested again and all clear :-)

PR #2167, opened 2016-02-13
Fix crash when a entity with a behavior gets unloaded.

flo:
The behavior system did not clear the entity list properly. It tried
to work with unloaded entities. This resulted in a NullPointerException
when the behavior node tried to work with a component from the
unloaded entity.


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/422/

PR #2168, opened 2016-02-14
Add Character back in to ray tests to allow PvP combat

Josharias:
Add Character back in to ray tests to allow PvP combat.  Be wary of Mr Barsack and his axe!

Add the ability to exclude entities from a physics ray test so that a player cannot hit or activate their own character.

This also fixes the problem of being not able to place blocks close to your feet because of the mismatched ray trace filters.  This problem is 
mentioned in https://github.com/MovingBlocks/Terasology/pull/2157.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterSystem.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterSystem.java
@@ -16,6 +16,7 @@
 
 package org.terasology.logic.characters;
 
+import com.google.common.collect.Sets;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.entitySystem.entity.EntityManager;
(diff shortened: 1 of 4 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/424/

Cervator:
Aha I was wondering why I couldn't damage another player during some recent testing.

I tried to merge this on top of #2157 as testing that was better but still left me unable to place blocks close to / underneath me, and it sounded like this would fix it. But it conflicts - any chance you have a better working version locally? Maybe just add this on top of the other PR.

I went back to just using this branch solo without the other PR.

One oddity I spotted which may be unrelated - on my client session I saw a weird half-broken tree and some held items didn't display, like a red mushroom, or even a torch. Hosting player was fine and saw the tree normally.

Client player:

terasology-160214020644-1152x720

Hosting player:

terasology-160214020717-1152x720

Additional discovery that is certainly unrelated: tool bar doesn't display in screenshots anymore? Huh.

PR #2171, opened 2016-02-15
Move inventory and health to core

Josharias:
This is a large set of changes.  The intention of this is to reduce the code in the engine and pull it out to module systems.  This will provide a smaller testing surface in the engine and hopefully provide a more stable engine going forward.  The contained changes bring the inventory and health systems out of the engine and into the Core module.  The changes avoid namespace changes so that there is less impact on dependent modules.  I would suggest a renaming of namespaces and extraction into their own modules as a next step after this, although not urgent.

# Major Changes
- Inventory and related systems moved to Core
- Health and related systems moved to Core
- Added a "Builder Sample Gameplay" module which proves that the engine is viable without inventory and health.
- Annotations were removed from the inventory system as a means of tagging components with special functionality because of complications with module sandboxing.  Instead,  plain old interfaces are used in their place.
- The Core:Door item/block was reworked a bit to function a bit better.  It no longer copies the item wholeheartedly and instead uses its own separate entity prefab when placed in the world.  This removes the extra rendering artifact that existed since the First Person view revamp.

# Known Issues
- *Important* The keybinds have to be reset after moving to this code because the buttons have been moved and are unlinked.  You will not be able to manage your inventory well until you do this.
- There is still health and inventory linkage to fields in the ItemComponent.  There are associated TODO comments in there to mark this.
- There are still texture assets in the engine that relate to health and inventory.  Some used,  some not.

# Modules Affected
- ChangingBlocks
- CopperAndBronze
- Durability
- Fluid
- Genome
- IRLCorp
- Machines
- ManualLabor
- Portals
- SubstanceMatters
- WoodAndStone
- WorldlyTooltip

# Todo before merging
- [x] Splitting a stack causes the item no never stop being rendered in the first person view.
- [X] Denied activation on a remote client when opening a door
- [X] Breaking a block with DropBlockInventoryComponent does not work
- [X] PvP stopped working 
- [X] Direct pickup not working


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/428/

Cervator:
Looking good :-)

The new gameplay template is working and a nice demo. Do you intend for it to ship with the engine as an embedded module? If so we need to tweak a few more things like including a copy of Core's build.gradle, ignoring it in IntelliJ's config for Git, etc. I would be happy to do that.

Found one apparent bug with the inventory:

1. Select the stack of torches you start with
1. Open inventory
1. Right-click the torch stack to pick up half of it
1. Place it into the inventory
1. Switch toolbar slot - you now have a torch stuck to your hand :-)

terasology-160216235025-1152x720

Works with other items too, probably something simple.

And another one with the door: a connected client cannot `e` interact with any placed doors (the host is fine, including doors placed by player 2). You get the `o.t.logic.characters.CharacterSystem - Denied activation attempt by Cervator3 since at the authority another entity would have been activated` error in the log on the host.

And finally an observation: players can't hurt each other again. Or maybe that's just not in this branch?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/432/

Cervator:
Retested latest commit since I noticed it. No more item overlap, yay! But also no more hand when holding an empty toolbar slot, doh :-)

Still, everything else seems solid (save the noted issue with the door). Eager to merge, so close now!

Two remaining questions: Should the players be able to hurt each others? Not working either direction in this branch. If that's from just not merging (only checking out the branch) then that's fine. Also the question about the new module, I figure it is meant to be embedded in which case I'll apply the needed tweaks when it is time :-)

Josharias:
Yes,  the PvP breaking was a bug introduced in this PR. Good catch! I had over refactored something so that it was only pertaining to blocks.

Some other bugs I found also fixed (noted in the summary).

New module.  Yes, please!  Could you do the magic to embed it correctly?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/433/

Issue #663, opened 2013-08-02
Stereo SFX don't drop in volume as distance to player increases


Cervator:
Noted by @immortius while testing in multiplayer - stereo sfx won't drop in volume as distance from player increases. We should make sure everything is mono for now.

Music should still be stereo.

We should make this restriction clearer to users of this API. On the one hand, we need documentation that states this limitation in place discoverable by developers. On the other hand, we may want to make this even more obvious by throwing an exception in case the audio system fails to process the sound effect correctly, e.g.,

```java
throw new AudioSubsystemValueError("Tried to assign a direction vector to a multi-channel audio source", sound);
```

Cervator:
minnesotags from IRC was nice enough to look at this, and it looks like only the camera click noise and drink were in stereo.

In addition to mono versions of those two he added a few new sfx.

I put them here for now: https://github.com/MovingBlocks/TeraMisc/commit/1866899a2b2fb393b216381d7022831432980238

I realized I'm not sure we want to keep the sfx in engine long-term, but that it is probably too early to split them out (really need better ways to bundle more modules easily first). Maybe an IotaSfx module for base sounds?

Naturally we could still add more sfx :-)

Cervator:
The stereo/mono location issue may be gone now after #2177 - leaving this open though since the assets are unaffected and still need to be integrated sometime.

skaldarnar:
This is still an issue today :disappointed:

skaldarnar:
I think all existing issues were resolved and we added more documentation to avoid this in the future.

Issue #1064, opened 2014-04-14
Tile damage effect persists beyond an "Ace" multiblock being broken


Cervator:
Not entirely sure if this is a general issue or specific to the way the L&S aces are defined (W&S multiblock objects might work differently)

Found during testing for a stable build, with the standard L&S lineup active

![terasology-140413205956-1152x700](https://cloud.githubusercontent.com/assets/1063833/2691029/93127d8c-c370-11e3-9038-970c90409740.png)
##

Cervator:
Update: Still an issue but got a little quirkier. Maybe impacted by #2177 or something a little earlier, new issues didn't seem present in Omega 371 (Feb 6th)

New quirks:

- [ ] Overlay is not using the L&S specific (went back to cracked rock)
- [ ] When the block is fully damaged it doesn't actually break (nor does the overlay go away)
- [ ] The particle effects are fine for the L&S card but for a door it now looks like .. diamond ore?

Exhibit A:

![terasology-160222230656-1152x720](https://cloud.githubusercontent.com/assets/1063833/13241800/8aff6b78-d9bc-11e5-8d37-3f002275f848.jpg)

Exhibit B:

![terasology-160222233122-1152x720](https://cloud.githubusercontent.com/assets/1063833/13241804/97baab5c-d9bc-11e5-9d3b-85ea75a019df.jpg)

skaldarnar:
I noticed some weird particle effects when breaking blocks in TtA, looked like hitting the companion cube.

Cervator:
Closing this as complete / obsolete. Issues have since been resolved although the original block breaking effect still needs to be reactivated (submitted separate issue in a module repo for that)

PR #2177, opened 2016-02-23
Fix sound issues when in multiplayer

### LINKED ISSUE #715: No "place block" sfx played on client for own placed block (but OK for others)
Playing as server all sfx seem to play from everybody. Playing as client you seem to hear others placing blocks fine, but cannot hear yourself placing blocks. Footsteps play normally.

Might involve other sfx types too but this is the main one that stuck out and got confirmed :-)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1026729-no-place-block-sfx-played-on-client-for-own-placed-block-but-ok-for-others?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Josharias:
This solves hopefully all of the sound issues in multiplayer (fixes #715) and some issues with sound placement (#663).  There are two strategies to sound that are important to note:

1. Playing sounds on a common system (where the same code runs on remote client and authority)
  - Use the PlaySoundEvent event that takes in the EntityRef of the owner.  Events that trigger playing a sound need need to specify the instigator entity that caused the sound so that when the authority broadcasts the sound to all players,  it will skip repeating the sound on the owning client.
  - Accessing the audio system directly is generally a bad idea.
2. Playing sounds on the authority
  - Use the PlaySoundEvent without an owner EntityRef.  That way it will broadcast to everyone including the client that caused the sound.

Also, this PR functionally fixes the particle effects on remote clients by replicating the particle entities across the wire.  This is probably not the most efficient way to do this,  but it is functional.  Ideally in the future we may want to allow clients to scale back the particles so that lesser computers can get ok performance.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/behavior/tree/PlaySoundNode.java
+++ b/engine/src/main/java/org/terasology/logic/behavior/tree/PlaySoundNode.java
@@ -18,9 +18,8 @@ package org.terasology.logic.behavior.tree;
 import org.terasology.assets.ResourceUrn;
 import org.terasology.assets.management.AssetManager;
 import org.terasology.audio.AudioEndListener;
-import org.terasology.audio.AudioManager;
 import org.terasology.audio.StaticSound;
-import org.terasology.math.geom.Vector3f;
+import org.terasology.audio.events.PlaySoundEvent;
 import org.terasology.registry.In;
 import org.terasology.rendering.nui.properties.OneOf;
 import org.terasology.rendering.nui.properties.Range;
@@ -49,8 +48,6 @@ public class PlaySoundNode extends Node {
 
     public static class PlaySoundTask extends Task implements AudioEndListener {
         @In
-        private AudioManager audioManager;
-        @In
         private AssetManager assetManager;
         private boolean playing;
         private boolean finished;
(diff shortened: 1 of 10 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/437/

Cervator:
Seems to test out fine, sfx appears to be placed for both players now. Little hard to distinguish just trying two game instances on a single machine, heh :D

I fixed a breaking change with the `BlockParticleEffectComponent` to `String` in LightAndShadow's CardSystem. I figure it is is, the particles from the big cards are coming out correctly now anyway. I noticed another couple issues that may relate or predate this a little and updated #1064 with the details

Huzzah for sound fixes! @Josharias will you keep track of your notes somewhere for future use? Like in a tutorial that may briefly cover audio and sound? Could just add the notes to one of the tutorial issues so we won't forget. Or toss an example or two into the Sample module.

Issue #2179, opened 2016-02-23
UIDropdownScrollable shows scrollbar even when it's unnecessary


AlexanderGrooff:
The scrollbar shows up on that aren't longer than the given number of options. F.e., a menu with 3 options, and the setVisibleOptions is set to 5 will still show a scrollbar.

PR #2180, opened 2016-02-23
Fixed showing of unnecessary scrollbar UIDropdownScrollable

AlexanderGrooff:
Fix for #2179. The UIDropdownScrollable now only shows the scrollbar when there are *more* items than `visibleOptionsNum`. Also made the `onDraw` code more readable/maintainable.


GooeyHub:

AlexanderGrooff:
It seems that something went wrong with the merging of branches on my side. It doesn't seem to affect the changes I made to the code and/or the mergability, but it shows this long list of commits. My apologies for that.

Cervator:
Hi @AlexanderGrooff and yep, that's a common issue when getting used to Git :-)

It is what happens when you merge things downstream, like with https://github.com/delftswa2016/Terasology/pull/1 - due to the merge commit the order of commits now differs from upstream, so all of them show up on a next PR like this.

Generally you only *merge upstream* while you rebase or pull downstream. After a PR gets merged into upstream you should be able to do a simple pull via Git (not a PR via GitHub as that always results in a merge commit) which only adds the single merge commit on top.

If you happen to have any existing work you instead use rebase, which takes your older commits and rebases them on top of the upstream head commit. Again you end up in a nice clean state where a new PR to upstream would only include new commits :-)

Maybe you already solved that since I see the other PR nice and clean. For this one I suggest rebranching from upstream's develop branch and reapplying your changes then making a new PR. You can use something like `git cherrypick` but that can be a little tricky as well.

AlexanderGrooff:
I've duplicated the code and fixed the commits this time. See #2183.

PR #2184, opened 2016-02-24
Fix #2030 Add hostname & server address as log file name on server play

### LINKED ISSUE #2030: Game log files for server play lack a world name component
If you play a local game the log file for game/world stuff will be named something like `Terasology-world1.log`

If you join a server you'll get `Terasology-.log` - we probably should give that some alternative name (could just use a placeholder "world" unless we have a local world with a name)

Should be easy.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28680938-game-log-files-for-server-play-lack-a-world-name-component?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

arkka:
First, this fix check whether Game Name is present or not. If the game name is not present, then this fix look up for the Server hostname & address as log file name using existing SocketAddress.

The output forrmat log file name as follows:
Terasology-<hostname>-<IP>-<port>.log

Example:
Terasology-localhost-127.0.0.1-25777.log

### DIFF
--- a/engine/src/main/java/org/terasology/engine/modes/loadProcesses/JoinServer.java
+++ b/engine/src/main/java/org/terasology/engine/modes/loadProcesses/JoinServer.java
@@ -33,6 +33,7 @@ import org.terasology.module.ModuleEnvironment;
 import org.terasology.naming.NameVersion;
 import org.terasology.network.JoinStatus;
 import org.terasology.network.NetworkSystem;
+import org.terasology.network.Server;
 import org.terasology.network.ServerInfoMessage;
 import org.terasology.world.internal.WorldInfo;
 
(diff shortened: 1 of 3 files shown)


GooeyHub:

Cervator:
ok to test

Good to see you figured it out and thanks in advance! :-)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/441/

msteiger:
Great - thanks! I will look into this asap (Sunday?).

msteiger:
I'd also transmit the game name during the join process and combine it with the remote server address. The field should be already present in the Protobuf entry (maybe NetMessage?) and you probably just need to set it on the server side when a client connects.

arkka:
Yes, I have already take a look at NetMessage.ServerInfoMessage before. But, still couldn't understand why the game name is not present there in the first place (Protobuf entry in NetMessage). If i'm not mistaken, Game Name information is on protobuf NetData tag number 16. I've try to debug it on https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/protobuf/NetData.java#L14934, that never being executed. That's why I made the fix on the high level class without the gamename. Any other clue about this?

msteiger:
Yes, the game name should be set in [NetworkSystemImpl](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java#L813), but it isn't. You should be able to call `serverInfoMessageBuilder.setGameName()` in that method.

Then, it should be serialized and transferred as expected so you can read the game name from `ServerInfoMessage` on the other side of the wire.

Cervator:
Tested out and merged :-)

It satisfied #2030 as is (just needed *something* named), if we should clarify it even more and make the game name available remotely that's cool too. Can do in a follow-up if desired sometime :-)

Issue #2010, opened 2015-11-15
Java package structure review prior to v1.0.0 + maybe some package-level Javadoc


Cervator:
Came across http://forum.terasology.org/threads/alpha.1026/#post-10144 and wanted to make a reminder for the v1.0.0 milestone.

We're going "stable" so this in a way will be a last chance to restructure some deep stuff like whole packages. Everything may already be fine, but it seems like an idea to have somebody with some good architectural ideas go over our packages at a high level. Also ideal time to sprinkle some package-level Javadoc here and there.

Could relate to #1975 and *maybe* #1729 although that's more of a best practice topic.

Cervator:
Closing as complete for this round. Any future renames will likely have to wait for v2.0.0 (unless they don't break API or can be redirected cleanly). Doc we can of course add whenever.

PR #2187, opened 2016-02-28
Move Assets class from org.terasology.asset to org.terasology.utilities.

### LINKED ISSUE #2089: Asset clarification to avoid confusion
While playing with the new API scraper (#2079) I got myself a nice dose of confusion seeing all these org.terasology.assets package hits yet the engine project only has org.terasology.asset.Assets

The package-info in the engine at org.terasology.asset has what sounds like some comments relating to the time before gestalt-asset was split out, but now there is only one utility class left in the package

Suggestions:

* Move org.terasology.asset.assets in the engine to org.terasology.utilities instead, that way we only have org,terasology.assets in Gestalt
* Rename the package structure in Gestalt to include "gestalt" somewhere. I figure it was all split out with minimal changes to be less hassle, but in theory should probably be named more explicitly?
* Clean up the javadoc in package-info for engine's org.terasology.asset - or just delete it if we move the single remaining class to utilities

Admittedly there are a grand total of 432 usages of Assets to then fix (tons in modules), but I don't mind a mass refactor in my mega-workspace, that'll probably catch most of it in one go. And that's why this is a thing to get out of the way by v1.0,0 :-)

Related to #2010

@immortius what do you think? I'd like to move Assets and take that task on myself, that'll clear up most of the confusion. Should we change anything in Gestalt or not worth it?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29392536-asset-clarification-to-avoid-confusion?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Cervator:
Fixes #2089 and completes #2010 for now (any more breaking renames would have to wait to v2.0.0)

PR for doc purposes only, I've got fixes for some 30 modules coming right after merging :-)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/443/

Issue #1144, opened 2014-06-21
Additional graphic options in GUI


Cervator:
Need to add MOAR OPTIONS! In-game in the video settings. Some might go in the advanced section but really they just need to get crammed in there somewhere already. Putting config.cfg entry in ( ) if I happened to notice

- [x] Bloom (bloom)
- [x] Remove environment effects in favor of split-out animated grass/water
- [x] Animated grass (animateGrass)
- [x] Animated water (animateWater) - wavy sea / water blocks essentially
- [x] Particle effects (particleEffectLimit) - slider value, 0 disables, 10 default
- [x] Disable fog (the one at the limit of the players view range - might not be an option yet)
- [ ] Disable "tunnel vision" (the corners being darkened - also may not be an option yet)
- [ ] Add option to water reflection to not even reflect the sky (as currently on any level other than sea level). Maybe slightly related to #1371. And/or change refraction quality
- [x] Volumetric fog (may not actually work but anyway)

Any other obvious ones? There are a few more in config.cfg I'm not sure if we want in the main graphic settings or not (some are good candidates for the advanced screen, certainly). There are probably also some options we don't even have in config.cfg yet that would be useful to have toggles for.

Cervator:
Closing this as sufficiently fixed. More work could be done but we got the important ones and will start another round of improvements later. Leaving this as completed in v1.0.0.

PR #2211, opened 2016-03-07
Add fog, volumetric fog and particle effect limit options 

tdgunes:
## Contains

Yet another partial fix for #1144

## How to test

Additions can be seen in video settings menu. 

For testing, **particle effect limit**: picking block from ground generates multiple particles during animation. 

For testing: **fog**: blocks that are far from the player will be foggy or not according to setting. For future, there can be extra options for fog like density, colour, equation (linear, exp or etc.). 

For **volumetric fog**, I couldn't find a proper test case for trying out volumetric fog implementation in "combine" shader. But, my current addition can turn it on or off. It would be great if someone can provide instructions for me to test recent additions.


GooeyHub:

Cervator:
add to whitelist

Hey @tdgunes and thanks :-)

On volumetric fog I wonder if it is even enabled at present? I remembered an old video showcasing it nearly three years ago by @begla where he would enable / increase volumetric fog via sliders in TeraEd (an alternatively game facade launching with some extra graphics config)

https://www.youtube.com/watch?v=cXiX5ZJT_-k

I just realized that video is unlisted rather than published, so very few people have ever seen it. Maybe it was implemented but never fully finished up for actual use? That was back around the end of our 3d wizardry heavy period.

@emanuele3d might have a related clue or two

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/462/

engine/src/main/resources/assets/shaders/combine_frag.glsl
```diff
@@ -45,10 +45,10 @@ uniform sampler2D texEdges;
 uniform float outlineDepthThreshold;
 uniform float outlineThickness;
 
-# define OUTLINE_COLOR 0.0, 0.0, 0.0
+#define OUTLINE_COLOR 0.0, 0.0, 0.0
 #endif
 
-#if defined (VOLUMETRIC_FOG)
+#ifdef VOLUMETRIC_FOG
```
emanuele3d: I just read about the difference between #ifdef and #if defined. Good idea changing it.

engine/src/main/java/org/terasology/rendering/shader/ShaderParametersCombine.java
```diff
@@ -107,6 +107,12 @@ public void applyParameters(Material program) {
             program.setFloat("outlineThickness", outlineThickness, true);
         }
 
+        //TODO: Other parameters and volumetric fog test case is needed
+        Camera activeCamera = CoreRegistry.get(WorldRenderer.class).getActiveCamera();
```
emanuele3d: The activeCamera is potentially fetched twice - the first time earlier in the class. Do you mind changing this so that it is fetched at most once? 

Also, since you are working on this file, would you mind fetching the rendering config and the world renderer only once, somewhere at the beginning of applyParameters? Lines such as:

`CoreRegistry.get(Config.class).getRendering().isLocalReflections();`

Should really just be:

`renderingConfig.isLocalReflections();`

engine/src/main/java/org/terasology/rendering/shader/ShaderParametersCombine.java
```diff
@@ -107,6 +107,12 @@ public void applyParameters(Material program) {
...
+        if (activeCamera != null && CoreRegistry.get(Config.class).getRendering().isVolumetricFog()) {
+            program.setMatrix4("invViewProjMatrix", activeCamera.getInverseViewProjectionMatrix(), true);
```
emanuele3d: This is something I need to learn about: what does this line do? I mean, sure, I can see it sets a 4x4 matrix in the GPU to be the activeCamera's inverse view projection matrix. But why?

tdgunes: In https://github.com/tdgunes/Terasology/blob/new-options/engine/src/main/resources/assets/shaders/combine_frag.glsl line 54, I suppose if volumetric fog is enabled, the implementation requires invViewProjMatrix to be supplied by "activeCamera's inverse view projection matrix". But since, I couldn't figure out a way to test volumetric fog, for now this is useless. :(

engine/src/main/java/org/terasology/rendering/shader/ShaderParametersCombine.java
tdgunes: Throughout the implementation, I felt that lines with _"CoreRegistry.get(Config.class).getRendering()"_ are in a style of convention. After hearing this, I will definitely refactor it.

emanuele3d: Thank you. I'd love to change that convention everywhere, it makes for very cluttered and potentially slightly lower performance code. But I do not have the time for a systematic change. I just do it in the code I'm focusing on.

engine/src/main/java/org/terasology/rendering/shader/ShaderParametersCombine.java
emanuele3d: I see.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/465/

Cervator:
Hehe, probably best that volumetric fog is off by default :D

terasology-160308174329-1152x720

Otherwise yeah the switches work and that was the plan :-)

PR #2215, opened 2016-03-07
Handle zero division in InventoryGrid

rzats:
### Contains

Fixes #2212 - adds a catch block and warning-level log entry to `getPreferredContentSize()` & a precondition to `onDraw()` - logging here would be non-viable (think of the logspam!)

### How to test

Try playing TTA depending on a version of WoodAndStone that doesn't have Terasology/WoodAndStone#43 merged in?
This should be a fairly straightforward fix either way.

### Notes

These are the only methods that can cause a zero-division error within Core layouts.

Issue #2212, opened 2016-03-07
Harden UI code to avoid division by zero


Cervator:
Noted in #2188 as an issue in `InventoryGrid`

Need to add a zero check for `cellSize.getX()` in `getPreferredContentSize()` and in any similar places.

Edit: A warning-level log entry should also be added if a zero is detected so somebody can notice and fix it.

Should be easy! :-)

May be relevant to @oijazsh for a quick fix and @rzats as part of extracting NUI (to https://github.com/MovingBlocks/TeraNUI)

Issue #1402, opened 2014-11-27
Improve visibility of missing resource / asset scenarios


Cervator:
When the game launches it may be missing some references, like misconfigured icons . Usually this gets logged, but that's easy to miss. In-game missing icons in inventories get rendered with a ? so you know something's wrong. With #1400 we no longer crash on rendering a held object with a missing icon, but it is less visible.

I wonder what we can do to highlight both that situation (render an actual held `?` just like in the inventory?) and missing resources in general. Over time we've often had startup log entries reporting issues for weeks or months at a time, but they're so easy to ignore.

Maybe we can gather up such reports during runtime and offer the user a clear indication something isn't perfect. How about keeping a registry of unique errors and show a little red `!` somewhere in the main menu that the user can click to review and either report or dismiss the errors? IntelliJ does something similar itself (little event log button in the lower right corner). 

This sounds like it could be a perfect new feature for the CrashReporter. Could even hold some GSOC potential - getting close to that time of the year again :-)

Related:  #2577

Edit: This could perhaps also relate to the general concept of receiving automated or semi-automated reports and metrics from clients (opt-in to send specs etc to help the project)
##

Cervator:
Slight addendum / clarification: this is meant to be a sort of code analytic like style checking or even unit testing. At build time we'd want to run a validation that all JSON is cleanly formatted, for instance, and if possible scan the config and actually confirm that assets exist. I'm not sure if that's always possible but it would probably require the same kind of minimal environment established as unit testing does, quite possibly even reported as one or more unit tests (so it'll set builds unstable in Jenkins)

During runtime maybe when a critical missing asset like this is encountered it should be logged better and any crash from badly formatted JSON probably could be caught. If you try to create a new world and it crashes during initial loading the player could be dumped back at the "Create World" menu with an error about poorly configured assets in the chosen module set. Possibly a similar process could later help handle conflicts between modules.

With the added emphasis of unit testing and other analytics (see for instance the [TWOVANA thread in the forum](http://forum.terasology.org/threads/the-world-of-varied-and-noteworthy-achievements.1209/)) I'm also linking this to #135 (ancient request for better doc / tutorial to prepare unit tests and such)

This topic is pretty wide open for interpretation of how a proposal to address it for GSOC could be written, but also possibly a bit difficult to nail down unless you're already well-versed in unit testing, code analytics, JSON validation, and error handling.

Cervator:
[...]

PR #2216, opened 2016-03-08
Avoid catching of ArithmeticExceptions, by avoiding division through 0.

flo:
### Contains

Proper fix of #2212 . Replaces catch added by #2215 logic that avoid the division through 0.

### How to test

See  #2215

### DIFF
--- a/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryGrid.java
+++ b/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryGrid.java
@@ -18,8 +18,6 @@ package org.terasology.rendering.nui.layers.ingame.inventory;
 import com.google.common.base.Function;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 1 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/473/

Cervator:
Thanks @flo - I can confirm the fix and appreciate going the extra mile for this :-)

I know we all wish for the most perfect code we can get yet none of us have infinite time. Maybe my new day job is showing too much, talk about having to make compromises, heh.

In this case with WaS "broken" (set to inherit "default") you get a workstation screen with no slots in it, yet nothing gets logged hinting at what it might be. Do you think it Is better to keep the code clean and let a dev figure out what's wrong and never commit bad code or add in extra logging that in an ideal world will never trigger?

I thought the one-time logging was clever, with the simple if variant to avoid the try-catch. But I'll admit it then in theory would never trigger outside a developers workspace - you'd find the error and correct it before ever committing.

Is this something we'd be better off catching via the effort in #1402? Which coincidentally @rzats is showing an interest in as well, so that may be a perfect fit.

PR #2217, opened 2016-03-09
Health commands now working.

### LINKED ISSUE #1619: Some commands not recognized for clients connected to a server
In a somewhat odd twist after fixing "teleport" in #1618 it now works in single player, yay! But weirdly the command doesn't even appear to _exist_ when connected as a multiplayer client. Same for a few other commands like health and healthMax, yet certain other commands are fine.

A player hosting a server with the normal game client can also use those commands normally, even with another player connected being told they don't exist. 

Maybe it relates to where commands are defined - maybe some classes aren't processed the same when a client joins?

We probably should check to see if this is the case with other commands too.
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9054767-some-commands-not-recognized-for-clients-connected-to-a-server?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

HashCode55:
fixes #1619 
- The health commands are now working.
- The teleport command is now being recognized at the client side BUT with one major problem.

As @flo said the commands which are defined in the authority system cannot run on the client side because the systems are loaded which are in the same environment. But the problem with the teleport command is that it uses `onactivatedcomponent` event to activate the entity and throughout the code base there are various systems which are in AUTHORITY register mode and are used by the above method, so this means the teleport command won't work until and  unless we make AUTHORITY systems available on the client side.
I actually wrote some commands for the server like giving teleport-another-user capability but that won't work until and unless the above thing is solved.


GooeyHub:

Cervator:
add to whitelist

Interesting notes on the systems. I'm unsure what the ideal fix is, maybe it would make for a good thread in the dev portal forum (or just discussion on a new issue here)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/482/

Cervator:
Can confirm improvements. Health and teleport are seen on a client, although as stated teleport isn't quite right as client.

Health commands work with some oddities: the first time I do `damage 50` as a connected client the health bar doesn't update. Run `healthMax` and the next damage showed up as expected. But another `damage 50` at a point where the health was was higher than 50 killed the player. It seems like the displayed health and the actual health go out of sync?

At one point I could even use `showHealth` and get 100 (max) despite damage taken.

Want to take a look at the health issue and we can discuss teleport / general architecture elsewhere?

Josharias:
I made some comments in the #1619.   Short story, probably just want to move the ```runOnServer = true``` commands to a new common system that is registered as ```RegisterMode.ALWAYS```.

HashCode55:
Hey @Cervator 
Isn't it sort of a server lag? I tried doing it, it got updated after like 5-6 seconds. 
Before this I tried writing a command to check other player's health from the server, and the client's health got updated instantly on the server. I'm sort of confused. But most probably it is because of server lag.

Josharias:
@HashCode55,  I tried this out in a headless server multiplayer scenario and breaking blocks and damaging other players does not function.  I am guessing that in a local hosted game, everything works because the server is also acting as a client and so gets all the stuff that is on the CLIENT only system.  Give a headless server/remote client a shot and see if you can duplicate my findings.

flo:
The commands need to run on the server. It is strange that we let commands define in a client system that with runOnServer=true as they need then to call functionallity that is in the authority system.

We should probably change it and make it so that commands that get executed on the server can be (only) defined in systems with authority(or the always mode).

flo:
I suggest we move discussion to the issue ticket #1619

msteiger:
Any news on this one?

Cervator:
I guess this uncovered some deeper issues. Some discussion went for a while over in #1619 but seems to have quieted down. I'm not sure if it would be easy to finish this as-is. Unless conversation restarts maybe we close the PR but reference it in the issue as sample code to pick up from?

Cervator:
Superseded by #2600

Issue #2213, opened 2016-03-07
Recipe output not visible in TTA crafting


Cervator:
### What you were trying to do

Craft stuff in Throughout The Ages with the `g` crafting menu

### What actually happened

While the recipes were listed the product (right side) was invisible, making it more than a little tricky to actually know what you're crafting.

**Example:** 2 rough stones (drops from dirt blocks) produce 1 flint. However with 2 rough stones in your inventory you just get a recipe listing the 2 rough stones and no apparent product, yet if you craft you get some flint.

May require an adjustment in one of the TTA modules rather than engine, but probably came about from recent engine refactoring.

Cervator:
Closing as fixed, thanks @TobyShaw and then @Josharias for follow-up!

PR #2227, opened 2016-03-13
Fix bug #2213, result of crafting recipe is displayed in crafting window

TobyShaw:
### Contains

Fixes bug #2213 

ItemIcon always returned (0,0) for its preferred size, which is what the craft bench relied upon to render it. I chose to return the size of the texture instead, I'm not sure if this is the best approach to fixing it, but it fixes the recipe display perfectly.

An alternative fix would be to edit CraftRecipeWidget to not rely on calculatePreferredSize, and provide its own size.

### How to test

Obtain items for TTA crafting, eg destroy enough dirt to get stones. Press 'g' to open crafting window, you will see the result of the recipe displayed correctly.

ItemIcon is used all over the place, I am also unsure if it was set to return the 0 vector for a specific reason, or if it was a mistake.

### DIFF
--- a/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/ItemIcon.java
+++ b/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/ItemIcon.java
@@ -84,6 +84,12 @@ public class ItemIcon extends CoreWidget {
 
     @Override
     public Vector2i getPreferredContentSize(Canvas canvas, Vector2i sizeHint) {
+        if (icon != null) {
+            TextureRegion texture = icon.get();
+            if  (texture != null) {
+                return texture.size();
+            }
+        }
         return new Vector2i();
     }
 


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

This is probably then the same sort of issue that was fixed by #2216 (alternative fix from #2215). Something that got shook loose back when @Josharias was pulling stuff like inventory out of the engine. So may be relevant to him as well as @flo and @rzats for further review :-)

So I don't *think* we're talking about something done on purpose. One of those three might have some more ideas on the exact details and on whether this is a good fix.

In other news yep this is the perfect way to do PRs! :-)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/495/

Josharias:
I can confirm that this is a good fix to do.  Displaying something is much preferred to being hidden and having not good way to see and diagnose what is happening.  I will also fix up the skins in TTA so that it gets the intended sizing.

I checked it over with the JS modules, and it does not have any ill effects.

PR #2237, opened 2016-03-15
misc optimizations

automenta:
### Contains

misc trivial optimizations decided after profiling.  (ex: cache values, reduce memory allocation, etc)

### How to test

just running it on some empty worlds before and after with profiler


GooeyHub:

engine/src/main/java/org/terasology/rendering/opengl/PBO.java
```diff
@@ -30,7 +30,8 @@
  * Created by manu on 15.03.2015.
  */
 public class PBO {
-    private int pboId;
+    private ByteBuffer resultBuffer = null;
```
emanuele3d: Please, can you let me know your reasoning regarding this and the associated changes? What does this additional buffer do?

engine/src/main/java/org/terasology/world/propagation/light/InternalLightProcessor.java
```diff
@@ -64,12 +64,14 @@ private static void populateSunlight(LitChunk chunk) {
...
 
+        Vector3i pos = new Vector3i();
         for (int x = 0; x < ChunkConstants.SIZE_X; x++) {
             for (int z = 0; z < ChunkConstants.SIZE_Z; z++) {
                 for (int y = 0; y < ChunkConstants.MAX_SUNLIGHT; ++y) {
-                    Vector3i pos = new Vector3i(x, y, z);
-                    Block block = chunk.getBlock(x, y, z);
-                    byte light = sunlightRules.getFixedValue(block, pos);
+
+                    byte light = sunlightRules.getFixedValue(
```
emanuele3d: Good idea on extracting the Vector3i from the nested loops. Why did you decide to leave the byte allocation there? Could that be extracted too?

Also, is this part of the code actually looking for a light or is storing some kind of light intensity in this "light" variable? Can you have look? The latter option might call for a renaming of the variable.

engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
```diff
@@ -309,7 +312,7 @@ private void dumpCode(int type, Set<ShaderProgramFeature> features, String sourc
         }
 
         // Dump all final shader sources to the log directory
-        final String strippedTitle = getUrn().toString().replace(":", "-");
+        final String strippedTitle = COLON.matcher(getUrn().toString()).replaceAll(Matcher.quoteReplacement("-"));
```
emanuele3d: Hmmmm. I'm seeing a longer line of code for dubious performance gains here. Even if the way you just did it was considerably faster, I don't think it's worth the loss of readability.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
```diff
@@ -27,8 +27,6 @@
 import static org.lwjgl.opengl.EXTFramebufferObject.GL_FRAMEBUFFER_EXT;
 import static org.lwjgl.opengl.EXTFramebufferObject.glBindFramebufferEXT;
 import static org.lwjgl.opengl.GL11.*;
-import static org.lwjgl.opengl.GL11.GL_CULL_FACE;
-import static org.lwjgl.opengl.GL11.GL_LEQUAL;
```
emanuele3d: Thank you for removing this unused imports. Probably my bad leaving them there.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -248,11 +248,7 @@ private void preRenderUpdate(WorldRenderingStage renderingStage) {
         resetStats();
 
         currentRenderingStage = renderingStage;
-        if (currentRenderingStage == WorldRenderingStage.MONO || currentRenderingStage == WorldRenderingStage.LEFT_EYE) {
-            isFirstRenderingStageForCurrentFrame = true;
-        } else {
-            isFirstRenderingStageForCurrentFrame = false;
-        }
+        isFirstRenderingStageForCurrentFrame = currentRenderingStage == WorldRenderingStage.MONO || currentRenderingStage == WorldRenderingStage.LEFT_EYE;
```
emanuele3d: No. Please revert to previous code. For two reasons.
1. the new version is not as readable and this is a tricky part of the code (which one day I hope to handle very differently)
2. I'm working on this particular file and I will submit changes sometimes soon.

emanuele3d:
A general comment about this PR. In my opinion it is a bit too wide ranging. It implicitly involves lots of different people as it touches on many different portions of the codebase. 

If people like @Cervator, @flo or @msteiger think otherwise then it's fine: they are much more experienced then me, but if it was up to me I'd break it down in smaller mouthfuls, to ease the reviewing process.

Cervator:
Thanks for the initial commits @emanuele3d :-)

Hi @automenta ! Do we know you previously, like a GSOC student with a different GitHub username? Thanks for the optimizations!

But yeah it may be an idea to split this up a bit into grouped areas to make review easier. I'll trigger the bot for build review anyway just so we have it for this and future edits on this PR :-)

@GooeyHub okay to test

Cervator:
One of these days we'll update that bot command RegEx ...

@GooeyHub ok to test

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/499/

Cervator:
Oh, looks like that broke:

```
:engine:compileJava/home/jenkins/workspace/TerasologyPRs/engine/src/main/java/org/terasology/persistence/internal/SaveTransaction.java:22: error: package jdk.nio.zipfs does not exist
import jdk.nio.zipfs.ZipFileSystemProvider;
                    ^
/home/jenkins/workspace/TerasologyPRs/engine/src/main/java/org/terasology/utilities/tree/SpaceTree.java:29: error: package jdk.vm.ci.sparc does not exist
import static jdk.vm.ci.sparc.SPARC.o1;
                             ^
/home/jenkins/workspace/TerasologyPRs/engine/src/main/java/org/terasology/utilities/tree/SpaceTree.java:29: error: static import only from classes and interfaces
import static jdk.vm.ci.sparc.SPARC.o1;
^
/home/jenkins/workspace/TerasologyPRs/engine/src/main/java/org/terasology/persistence/internal/SaveTransaction.java:350: error: cannot find symbol
        FileSystemProvider zipProvider = new ZipFileSystemProvider();
                                             ^
  symbol:   class ZipFileSystemProvider
  location: class SaveTransaction
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
4 errors
 FAILED
```

engine/src/main/java/org/terasology/rendering/opengl/PBO.java
automenta: re-uses the buffer if it's the right size, avoiding:
``   ByteBuffer resultBuffer = BufferUtils.createByteBuffer(cachedBuffer.capacity());

engine/src/main/java/org/terasology/world/propagation/light/InternalLightProcessor.java
automenta: not sure, this change didnt seem as important as others might be

engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
automenta: dunno Intellij suggested that these pattern get factored out as compiled.  for those involved in importing like collada i figured they might somehow contribute to loading speedups

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
automenta: thank intellij =D

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
automenta: sure this one is unimportant

engine/src/main/java/org/terasology/rendering/primitives/BlockMeshGeneratorSingleShape.java
```diff
@@ -89,35 +89,39 @@ public void generateChunkMesh(ChunkView view, ChunkMesh chunkMesh, int x, int y,
...
-            drawDir[side.ordinal()] = blockAppearance.getPart(BlockPart.fromSide(side)) != null && isSideVisibleForBlockTypes(adjacentBlocks.get(side), selfBlock, side);
+            int sideOrd = side.ordinal();
+            drawDir[sideOrd] = blockAppearance.getPart(BlockPart.fromSide(side)) != null &&
+                    isSideVisibleForBlockTypes(adjacent[sideOrd], selfBlock, side);
         }
 
         // If the selfBlock is lowered, some more faces may have to be drawn
         if (selfBlock.isLiquid()) {
-            Block bottomBlock = adjacentBlocks.get(Side.BOTTOM);
+            Block bottomBlock = adjacent[1 /* BOTTOM */];
```
oniatus: Just browsing through this PR :) but I don't like the decreased readability of the array vs the enum map here.
Is the performance gain that high?

automenta: not sure but the enummap instantiation did appear non-negligible in the profiling results i got

engine/src/main/java/org/terasology/rendering/opengl/PBO.java
emanuele3d: As far as I know PBOs are always 1x1 pixels and are created only once. I appreciate the optimization effort but I feel this particular portion of code is getting more complicated for negligible gains. Unless we start to use PBOs more intensively, I'd recommend not to go ahead with these particular changes.

engine/src/main/java/org/terasology/world/propagation/light/InternalLightProcessor.java
emanuele3d: I'd suggest all memory allocation should occur outside the loops. So, let's get that light variable out of there too. =)

engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
emanuele3d: Nah. This seems to be acting on very small string, a title used to establish a file name. You'd start to notice speed-ups loading hundreds of thousands of shaders I'd imagine. Even including modules I doubt a user will ever find herself loading more than a hundred shaders.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: Thank you.

engine/src/main/java/org/terasology/rendering/primitives/BlockMeshGeneratorSingleShape.java
Cervator: Any way to get a better idea of how much it matters? Or at least use a local constant instead of a magic number + include a good comment explaining the details?

automenta: yeah that would probably make sense and there still might be a way to do it entirely with enums but without the enummap by some type of visitor pattern. either way not important

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/502/

automenta:
2 additional commits, one is a security disabling ability via environment variable, and the other fixes some of those bad imports (i was using a mix of jdk8/jdk9 in my IDE)

automenta:
cervator suggested that if this environment variable option is accepted, instructions for its use should be added to the modding wikipage

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/503/

Cervator:
Hey @automenta and thanks for the extra changes. The build works again but it looks like 16 more unit tests fail now. 22 knowingly broke with a PR a few days ago and are pending fixes, but it looks like more broke with this. Able to take a look and address some of the additional comments above?

http://jenkins.terasology.org/job/TerasologyPRs/503/testReport (the ones with age 1 are new)

Also, it might be good to split into multiple PRs, both with sizeable refactoring/optimization and something like the environment variable thing so they can be handled individually. Much easier than a big giant PR :-)

engine/src/main/java/org/terasology/persistence/internal/SaveTransaction.java
```diff
@@ -48,12 +49,7 @@
 import java.nio.file.StandardCopyOption;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.nio.file.spi.FileSystemProvider;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
msteiger: No wildcard imports, please.

engine/src/main/java/org/terasology/persistence/internal/SaveTransaction.java
```diff
@@ -230,13 +226,16 @@ private void preparePlayerStores(Set<EntityRef> unsavedEntities) {
 
 
     private void applyDeltaToPrivateEntityManager() {
+        EngineEntityManager privateEntityManager = this.privateEntityManager;
+
         deltaToSave.getEntityDeltas().forEachEntry((entityId, delta) -> {
             if (entityId >= privateEntityManager.getNextId()) {
-                privateEntityManager.setNextId(entityId + 1);
+                this.privateEntityManager.setNextId(entityId + 1);
```
msteiger: Is there a particular reason for doing this?

engine/src/main/java/org/terasology/rendering/primitives/BlockMeshGeneratorSingleShape.java
msteiger: As you can see in the implementation of EnumMap, it internally uses an array, since the number of elements is known. It boils more or less down to:
```EnumMap.get(Object key) { return ... vals[((Enum)key).ordinal()] }```
I seriously doubt that it has a measurable negative effect on performance. Agreeing with @oniatus I believe that using an enum is easier to read.

File: engine/src/main/java/org/terasology/utilities/collection/TypeMap.java [hunk 9]
```diff
@@ -70,7 +71,13 @@ public boolean containsValue(T value) {
         }
         return null;
     }
-
+    public <U extends T> U get(Class<U> key, Function<? super Class<? extends T>, ? extends T> ifAbsentPut) {
```
msteiger: I didn't check the use case here, so I might be wrong: have you thought about doing it the other way round? [Map.getOrDefault(Object key, V defaultValue)](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#getOrDefault-java.lang.Object-V-) could be useful.

engine/src/main/java/org/terasology/world/block/BlockUri.java
```diff
@@ -52,26 +57,36 @@ public BlockUri(String uri) throws BlockUriParseException {
                 throw new BlockUriParseException("Could not parse block uri: '" + uri + "'");
             }
             Name blockFamilyDefModule = new Name(split[0]);
+            ResourceUrn d;
+            Optional<ResourceUrn> s;
+            Name n;
```
msteiger: Please don't use one-letter abbreviations. This makes it harder to understand without any real benefit.

engine/src/main/java/org/terasology/world/block/family/AttachedToSurfaceFamily.java
```diff
@@ -68,7 +68,7 @@ public Block getArchetypeBlock() {
     public Block getBlockFor(BlockUri blockUri) {
         if (getURI().equals(blockUri.getFamilyUri())) {
             try {
-                Side side = Side.valueOf(blockUri.getIdentifier().toString().toUpperCase(Locale.ENGLISH));
+                Side side = Side.valueOf(blockUri.blockName.toString().toUpperCase(Locale.ENGLISH));
```
msteiger: Please don't use other classes members directly. This shouldn't even be possible. A legacy issue I assume.
Afaik, simple getters are compiled to the same bytecode as using the members directly, so their performance is identical.

engine/src/main/java/org/terasology/world/block/BlockUri.java
```diff
@@ -42,8 +42,13 @@
     public static final String IDENTIFIER_SEPARATOR_REGEX = "\\.";
 
     private final ResourceUrn blockFamilyDefinition;
-    private final Optional<ResourceUrn> shape;
-    private final Name blockName;
+    public final Optional<ResourceUrn> shape;
+    public final Name blockName;
```
msteiger: Don't make any fields public, please.

engine/src/main/java/org/terasology/world/chunks/blockdata/TeraDenseArray16Bit.java
```diff
@@ -68,6 +68,7 @@ public TeraArray deflate(TeraVisitingDeflator deflator) {
 
     @Override
     public int getEstimatedMemoryConsumptionInBytes() {
+        short[] data = this.data;
```
msteiger: I assume that you did this in the assumption that local variable access is faster than field instance access. As [indicated here on SO](http://stackoverflow.com/a/21614413/198397) there is no significant difference, so I'd assume that this is pointless.

engine/src/main/java/org/terasology/world/generation/RegionImpl.java
```diff
@@ -73,10 +75,6 @@ public Region3i getRegion() {
 
     @Override
     public Border3D getBorderForFacet(Class<? extends WorldFacet> type) {
-        if (borders.containsKey(type)) {
-            return borders.get(type);
-        } else {
-            return new Border3D(0, 0, 0);
-        }
+        return borders.getOrDefault(type,Border3D.ZERO);
```
msteiger: Good!

engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
emanuele3d: bump: this hasn't been addressed in the latest commit.

engine/src/main/java/org/terasology/rendering/opengl/PBO.java
emanuele3d: bump: also this hasn't been addressed in the latest commit. Please revert to original version.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: bump: please revert to original version.

engine/src/main/java/org/terasology/world/propagation/light/InternalLightProcessor.java
emanuele3d: Again, feedback not addressed.

automenta:
apologies for the vast amount of style violations. i didnt intend to get reviewed so carefully, just offered these changes (while exploring the source for the first time) in case any of them might be useful. definitely ignore the bad ones, ill be more careful in future PR's thanks!

engine/src/main/java/org/terasology/rendering/primitives/BlockMeshGeneratorSingleShape.java
```diff
@@ -173,14 +177,16 @@ private boolean isSideVisibleForBlockTypes(Block blockToCheck, Block currentBloc
 
     @Override
     public Mesh getStandaloneMesh() {
-        if (mesh == null || mesh.isDisposed()) {
-            generateMesh();
+        Mesh m = this.mesh;
```
Josharias: This kind of optimization is really only meaningful when you have the instance field accessed in a loop.  I would kindly suggest that micro-optimizing these kind of changes is bad for readability because it adds another variable to keep track of in your head as you read the code.

msteiger:
I suggest closing this rather broad PR and creating smaller and more precises ones where appropriate.

emanuele3d:
I agree with msteiger assessment: close this PR and restart with smaller, more focused ones.

Cervator:
Closing as scoped too wide / too many changes needed.

@automenta we'd still love to see some optimizations but it is better to do them in logical groups that can go through review and tweaking separately. I get that trying to readdress all the feedback here might be a bit overwhelming if you were just trying to point out some potential :-)

PR #2240, opened 2016-03-15
Added javadocs, todos, refactored WorldRenderer/WorldRendererImpl

emanuele3d:
* WorldRenderer interface: trimmed and fully documented
* WorldRendederImpl class: some refactoring, added javadocs for class and main public methods
* minor changes to most other files are a consequence of the interface changes

Next mini-goal:
* documenting of RenderableWorld/RenderableWorldImpl

### DIFF
--- a/engine/src/main/java/org/terasology/engine/modes/loadProcesses/AwaitCharacterSpawn.java
+++ b/engine/src/main/java/org/terasology/engine/modes/loadProcesses/AwaitCharacterSpawn.java
@@ -23,13 +23,14 @@ import org.terasology.entitySystem.systems.UpdateSubscriberSystem;
 import org.terasology.logic.players.LocalPlayer;
 import org.terasology.network.ClientComponent;
 import org.terasology.rendering.world.WorldRenderer;
+import org.terasology.world.chunks.ChunkProvider;
 
 /**
  */
 public class AwaitCharacterSpawn implements LoadProcess {
 
     private final Context context;
-    private WorldRenderer worldRenderer;
+    private ChunkProvider chunkProvider;
 
     public AwaitCharacterSpawn(Context context) {
         this.context = context;
@@ -49,18 +50,17 @@ public class AwaitCharacterSpawn implements LoadProcess {
         LocalPlayer localPlayer = context.get(LocalPlayer.class);
         ClientComponent client = localPlayer.getClientEntity().getComponent(ClientComponent.class);
         if (client != null && client.character.exists()) {
-            worldRenderer.setPlayer(context.get(LocalPlayer.class));
             return true;
         } else {
-            worldRenderer.getChunkProvider().completeUpdate();
-            worldRenderer.getChunkProvider().beginUpdate();
+            chunkProvider.completeUpdate();
+            chunkProvider.beginUpdate();
         }
         return false;
     }
 
     @Override
     public void begin() {
-        worldRenderer = context.get(WorldRenderer.class);
+        chunkProvider = context.get(ChunkProvider.class);
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
@@ -23,60 +23,227 @@ import org.terasology.rendering.world.viewDistance.ViewDistance;
 import org.terasology.world.WorldProvider;
 import org.terasology.world.chunks.ChunkProvider;
 
+/**
+ * Implementations of this class are responsible for rendering the whole 3D world,
+ * inclusive of background.
+ *
+ * While 2D GUIs are dealt with elsewhere, it is in the remit of implementations of this class
+ * to provide Augmented-Reality-Style user interfaces, i.e. boxes around selected objects
+ * 3d-positioned labels above player's heads or held-in-hand tools.
+ *
+ * If this is the first time you look into this interface, you should probably look at the
+ * following methods as they are the most important.
+ *
+ * <ul>
+ *     <li>update(float)</li>
+ *     <li>render(WorldRenderingStage)</li>
+ * </ul>
+ *
+ *
+ */
 public interface WorldRenderer {
     float BLOCK_LIGHT_POW = 0.96f;
     float BLOCK_LIGHT_SUN_POW = 0.96f;
     float BLOCK_INTENSITY_FACTOR = 1.25f;
 
+    /**
+     * This method is triggered when a chunk has been loaded.
+     *
+     * @param chunkPos a Vector3i providing the coordinates of the chunk that has just been loaded.
+     */
     void onChunkLoaded(Vector3i chunkPos);
 
+    /**
+     * This method is triggered when a chunk has been unloaded.
+     *
+     * @param chunkPos a Vector3i providing the coordinates of the chunk that has just been unloaded.
+     */
     void onChunkUnloaded(Vector3i chunkPos);
 
-    public enum WorldRenderingStage {
+    /**
+     * Lists the stages the rendering engine may go through on a given frame.
+     *
+     * At the time of the writing the rendering engine works in two conceptual modes: mono or stereo.
+     *
+     * In mono mode, used to render to a standard monitor, the rendering engine goes through
+     * a single rendering stage every frame, called MONO.
+     *
+     * Stereo mode is for stereoscopic displays: every frame the rendering engine first go through
+     * the LEFT_EYE rendering stage and then the RIGHT_EYE rendering stage. Each stage produces
+     * an image and the two are then combined at the end of the RIGHT_EYE stage.
+     *
+     * Notice that the renderer has no explicit notion of mono/stereo. It only knows which stage is the current one.
+     */
+    // TODO: candidate for renaming simply to RenderingStage
+    enum WorldRenderingStage {
         MONO,
         LEFT_EYE,
         RIGHT_EYE
     }
 
+    /**
+     * Retrieves the camera the world is being rendered from.
+     *
+     * @return a Camera object
+     */
     Camera getActiveCamera();
 
+    /**
+     * Retrieves the camera positioned and oriented to look down on the world from the point of view
+     * of the main light source. This camera is used to produce the shadow map.
+     *
+     * @return a Camera object
+     */
     Camera getLightCamera();
 
+    /**
+     * Retrieves the chunk provider currently in use by the renderer.
+     *
+     * @return a ChunkProvider object.
+     */
+    // TODO: candidate for removal - can be obtained through a context object.
     ChunkProvider getChunkProvider();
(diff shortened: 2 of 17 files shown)


engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -23,60 +23,227 @@
...
     void dispose();
 
+    /**
+     * Used only during the early part of the WorldRenderer lifecycle, this method triggers the loading of
+     * chunks around the camera and the generation of their meshes.
+     *
+     * @return Returns True if all necessary chunks have been loaded and their meshes have been generated, False otherwise.
+     */
+    // TODO: it might be desirable to separate the loading of chunks and mesh generation. The former may concern
+    // TODO: systems dealing with beyond-the-horizon simulations. The latter usually only concerns the renderer.
```
emanuele3d: These TODOs are more a long term discussion/goal. The basic thinking here is that the renderer shouldn't really be concerned with loading/unloading chunks, but only with generating and unloading their meshes. By splitting these two concerns the volume of loaded chunks could be larger than what is visible and entities and simulations could be enabled even if beyond the horizon. For example, a wild fire might initiate beyond the visible range and only eventually, as it spreads, become visible. Or a volcano might erupt beyond the visible range but lava flows might eventually reach the player.

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -23,60 +23,227 @@
...
-    float getBlockLightValueAt(Vector3f worldPos);
-
+    /**
+     * Returns the intensity of the light at a given location due to the combination of main light (sun or moon)
+     * and in-scene light sources, i.e. torches.
+     *
+     * @param coordinates a Vector3f providing the coordinates of the block to calculate the light intensity from.
+     * @return a float value representing the intensity of the main light and in-scene lights combined
+     */
+    float getRenderingLightValueAt(Vector3f coordinates);
```
emanuele3d: Somebody should check my interpretation of what the code in WorldRendererImpl does against this method's javadoc.

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -23,60 +23,227 @@
...
+     * Returns True if the head of the player is underwater. False otherwise.
+     *
+     * @return True if the head of the player is underwater. False otherwise.
+     */
     boolean isHeadUnderWater();
 
+    /**
+     * Returns the tint at the camera's current position.
+     *
+     * @return a Vector3f providing the ting at the camera's coordinates.
```
rzats: Should be "tint" instead of "ting".

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -23,60 +23,227 @@
...
-    public enum WorldRenderingStage {
+    /**
+     * Lists the stages the rendering engine may go through on a given frame.
+     *
+     * At the time of the writing the rendering engine works in two conceptual modes: mono or stereo.
+     *
+     * In mono mode, used to render to a standard monitor, the rendering engine goes through
+     * a single rendering stage every frame, called MONO.
+     *
+     * Stereo mode is for stereoscopic displays: every frame the rendering engine first go through
```
rzats: Should be "first goes through"

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -23,60 +23,227 @@
...
+    /**
+     * Returns the intensity of light at the current camera position due to the main light source (sun or moon).
+     *
+     * @return a float value
+     */
+    // TODO: candidate for removal - use getSunlightValueAt instead.
     float getSunlightValue();
 
+    /**
+     * Returns the intensity of the light at the current camera position due to sources in the scene, i.e. torches,
```
rzats: Odd trailing comma at the end here

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -23,60 +23,227 @@
...
+     * @return a Vector3f providing the ting at the camera's coordinates.
+     */
+    // TODO: candidate for removal: can be obtained through the worldProvider.getBlock(coords).getTint() methods
     Vector3f getTint();
 
+    /**
+     * Returns the current tick, an always progressing time value animations are based on.
+     *
+     * @return the number of milliseconds since the renderer starter rendering the first frame.
+     */
```
rzats: Should be "since the renderer started"

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -23,60 +23,227 @@
...
 
+/**
+ * Implementations of this class are responsible for rendering the whole 3D world,
+ * inclusive of background.
+ *
+ * While 2D GUIs are dealt with elsewhere, it is in the remit of implementations of this class
+ * to provide Augmented-Reality-Style user interfaces, i.e. boxes around selected objects
+ * 3d-positioned labels above player's heads or held-in-hand tools.
+ *
+ * If this is the first time you look into this interface, you should probably look at the
```
rzats: This paragraph sounds a bit odd because of the "if ths is the first time you're reading this..." false conditional - may need a rework.
I don't recall ever seeing documentation where methods are designated "important", anyway.

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
emanuele3d: Good catch!

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
emanuele3d: Thank you

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
emanuele3d: I'll clarify this.

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
emanuele3d: Hmmm. No. Those two methods are quite central to the WorldRenderer, with everything else sort of being in support. I'll try and rephrase but conceptually it's correct.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/510/

emanuele3d:
Just pushed a second commit on this, much meatier due to interface changes, added documentation and some other refactoring. 

In particular, WorldRendererImpl remains relatively complex. Most complex methods are private however and as such I am inclined not to add private javadocs for them. 

However, it would be good for other pairs of eyes to go through the WorldRendererImpl code and at least identify sections that could use futher clarification via comments or better self-commenting code.

Tagging @flo, @msteiger, @tdgunes, @pollend for feedback.

@flo: what would you say is a good strategy to introduce a context object in a class that is currently using the CoreRegistry? A number of shader-related classes take advantage of the global nature of the CoreRegistry to obtain it. Should I pass the context in via the constructor? Or via injection ideally?

engine/src/main/java/org/terasology/engine/modes/loadProcesses/AwaitCharacterSpawn.java
```diff
@@ -49,18 +50,17 @@ public boolean step() {
...
+            chunkProvider.completeUpdate();
+            chunkProvider.beginUpdate();
         }
         return false;
     }
 
     @Override
     public void begin() {
-        worldRenderer = context.get(WorldRenderer.class);
+        chunkProvider = context.get(ChunkProvider.class);
```
msteiger: Good!

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -23,60 +23,186 @@
...
+
+    /**
+     * Returns True if the head of the player is underwater. False otherwise.
+     *
+     * Implementations must take in account waves if present.
+     *
+     * @return True if the head of the player is underwater. False otherwise.
+     */
+    // TODO: while useful to the renderer, this should probably be moved somewhere closer
+    // TODO: to the camera-handling classes and perhaps renamed isCameraUnderWater()
```
msteiger: Agreed!

msteiger:
Looks good to me. Many useful corrections and additions.

In my opinion, `RenderingStage` could be something different. Maybe `depth-test`, `solidRendering`, `translucentRendering`, `SSAO`, `bloomEffect`, `underwaterEffect` and so on. Actual processing steps that are taken for every generated image (in accordance to the DAG idea). Whether this is for the left/right eye, or a single camera view should be a matter of the the camera matrix transformation only and thus be part of the camera. The renderer could even have a collection of cameras. Two cameras for the Oculus, one otherwise.

emanuele3d:
@msteiger: I hear you about the RenderingStage. From a semantic perspective you are perfectly right the term could be used in a much broader sense. Also, I agree that at some points the many rendering steps could be made more explicit. And I agree that mono/stereo camera handling leaves much to be desired and could be done differently, not so central to the working of the renderer.

My goal with most of the PRs I have submitted in the past year or so was not to radically change the functional architecture of the renderer but improve its readability and reduce its complexity without changing features, so that other people wouldn't be overly daunted by dabbing into it and bugfixing would be easier. I expect one more PR after this one, regarding the RenderableWorld class (holding acceleration structures). Once that's merged I'll consider this phase of my work complete: the top-level classes of the renderer will have been refactored and documented to a much higher readability standard then they were originally. 

Of course _possible_ work doesn't end there. If we do not want to make drastic architectural changes to the renderer the next logical step would be for me to dive into ShaderLand and see how I can refactor things again to improve readability and fix relatively obvious problems. However, there seem to be a convergence of opinions on the need for a brand new rendering architecture, with the DAG-based Rendering Pipeline proposal (#1741) cropping up at least as a discussion point. 

If that's the case I will concentrate my efforts in that direction. In that scenario what you refer to as RenderingStages will be nodes in the DAG and their precise will be up for debate as it will encompass a fairly heterogeneous set of processes.

emanuele3d:
@Cervator, @msteiger, what do I do regarding the "All checks have failed" message? The 22 failed tests seems to be from a part of the code base I have not touched.

msteiger:
The failed check seems to be caused by JaCoCo. Did anybody change the version or maybe it's because it's computed on a slave agent? @Cervator - should deactivate it for now? We're not really using the code statistics anyway (yet).

Sorry @emanuele3d - I meant to comment from a potential future work perspective.  Just like you said this would be a larger piece of work. I think that this PR is good for merging.

PR #2243, opened 2016-03-17
Split the language files and added more translation strings

Leonardoas26:
Split the menu_xx.lang files for readability.
Added more translation strings.
Translated more strings to portuguese.

Might affect weblate.


GooeyHub:

msteiger:
Thank you for your contribution, @Leonardoas26 - I need to think about the changes. It affects weblate in that it requires a subproject for each file.

Did you leave common terms like "OK" and "Close" in a general common language file? It is a bit tricky to review with 175 files touched.

Leonardoas26:
Yes, the menu_xx.lang still contains general strings.

Cervator:
Maybe we should target something like this for when we do the Weblate 2.5 upgrade? Chances are we'll need to configure some project stuff by then anyway, so we could prepare a new configuration approach in advance of that (but not apply it to the current instance)

msteiger:
Sounds good to me. Will dig into it as soon as 2.5 gets released.

Cervator:
Fun timing, just realized Weblate 2.5 actually came out a few days before this PR! :D

https://blog.cihar.com/archives/2016/03/10/weblate-25/

msteiger:
That's a surprise (to me at least). Sibbell didn't send a notification, I just checked again. OK - time to upgrade!

msteiger:
Oh noes - merge conflicts! The huge size of this PR makes is difficult to handle. We still need to upgrade, but maybe we can split this one into smaller chunks somehow?

msteiger:
Alright - Weblate 2.5 is up and running. What's the status on your end, @Leonardoas26?

msteiger:
I dug into this, trying to resolve merge conflicts and salvage as much as possible. Most of the changes are renames/file splits and only a few changes are related to new translation content. This is the trimmed version:

https://github.com/msteiger/Terasology/commit/ac8dc9ff350b89eebafb6aac6e6f1fc1810ac59d

It also contains some translation string for the input settings screen (see #2309).

Cervator:
@msteiger cool :-) How did it get down from 170+ files to 7? And do we need to change anything in Weblate like new sub-projects or anything? I figure not with the smaller commit.

@rzats ping - potentially relevant to your interests

msteiger:
I'd rather extract the strings to one or more translation constant classes. Creating a subproject in Weblate for each UI screen is overkill imho.

About 170 files resulted from the number of subprojects times the number of languages.

rzats:
@msteiger: the #2309-related fixes are more or less verbatim what I currently have in my fork: [rzats@`32efb26`](https://github.com/rzats/Terasology/commit/32efb26b894fd3936cf9eb64067ad720e74573d9). (That said, I've missed category labels as something that needs to be translated - nice catch!)

Are you planning to have the commits turned into a PR anytime soon? If so, I'll probably wait for that to be merged before changing language files, since merging them would probably involve more effort than just recreating after the PR.

msteiger:
No, I think the required effort does not justify the gain, in particular not if you have most of it covered in a separate commit. Thank you, nonetheless, @Leonardoas26.

Issue #2231, opened 2016-03-14
Fix broken behavior tree tests


Cervator:
Probably an easy one for a BT-interested GSOC student or so.

Link: http://jenkins.terasology.org/job/Terasology/1665/testReport/

Broken by the changes in #2199

Issue #2070, opened 2015-12-18
Behavior tree renovation / documentation [$100]


Cervator:
_Edit: Actually heard from @synopia and should know more at the end of December_

We've long had our behavior tree based AI thanks to @synopia but it hasn't moved some for quite a while. With the big v1.0.0 coming up I've been hoping to get some more tweaks applied along with a tutorial of some sort. Additionally Synopia actually worked on a v2 of behavior trees that lived mostly in a module. Ideally we'd get that dusted off and brought in as it seemingly came with a lot of improvements.
- _Original_ behavior tree video demo: https://www.youtube.com/watch?v=gPwBSwq4_mg
- Forum thread: http://forum.terasology.org/threads/behavior-trees.882/
- Example of a creature using the existing setup: https://github.com/Terasology/WildAnimals/blob/master/assets/prefabs/deer.prefab
- @synopia's _newer_ framework: https://github.com/synopia/jbt
- Module that went along with it (I think): https://github.com/Terasology/Behaviors
- The Maze module/world may help develop and test: https://github.com/Terasology/Maze/ (although Pathfinding also has a test world)

Two different scopes here. Naturally I'd prefer the bigger scope and for @synopia to come claim this, but it has been hard getting in touch :-)

**Fine tune the existing setup in engine**

I would consider this sufficient to claim the bounty I aim to put on here.
- [ ] Examine the system as-is and fix any note-worthy bugs, possibly involving the pathfinding or workboard system. Biggest obvious demonstration of this is spawning some L&S creatures and watching the logs go nuts with errors. Or then move out of range (let a chunk with creatures in it unload) and other errors may happen (used to crash the game, still might). I realize this may not directly be a consequence of the BT code, but the components are closely related. #1304
- [ ] Enhance the existing nodes with any missing utility. Main missing piece I can think of is pathfinding to a blocked location, assuming the presence of an ability to break through blocks. May put a separate issue/bounty in for that one.
- [ ] Add a bit of quality-of-life work to the BT editor. It works, but what could we add to make it more user friendly?
- [ ] Add (AI for) at least one more creature doing something more advanced than permanent "stray" or the order items in L&S
- [ ] Add javadoc to all involved classes that lack it, review any that have some
- [ ] Write a tutorial akin to the nice one @Josharias did for world gen: https://github.com/Terasology/TutorialWorldGeneration/ (probably including a utility module to go through each lesson)
- [ ] Record an updated demonstration video
[...]

PR #2245, opened 2016-03-19
resolved issue #2231

pollend:
### Contains

This pull request resolves issues relating to the behavior tree test failing.

### How to test

run the unit test for logic/behavior.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/behavior/tree/Task.java
+++ b/engine/src/main/java/org/terasology/logic/behavior/tree/Task.java
@@ -82,6 +82,7 @@ public abstract class Task {
             case NOT_INITIALIZED: {
                 onInitialize();
                 status = Status.RUNNING;
+                status = this.tick(deltaSeconds);
                 break;
             }
                 //running: update
@@ -91,6 +92,10 @@ public abstract class Task {
                     LOG.warn("update of Task {} returned invalid state {}", this.getClass(), newStatus);
                 }
                 status = newStatus;
+                if(newStatus == Status.FAILURE || newStatus == Status.SUCCESS) {
+                    onTerminate(status);
+                }
+
                 break;
             }
                 //suspended: do nothing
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
@GooeyHub: ok to test

Hey @pollend  - on IRC you mentioned thinking this looks like a hack, what do you mean?

Pinging @oniatus and @flo to review.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/504/

pollend:
Just seemed like a fairly roundabout way to fix a bug, but if it works then it's probably good. :D

engine/src/main/java/org/terasology/logic/behavior/tree/Task.java
```diff
@@ -91,6 +92,10 @@ public Status tick(float deltaSeconds) {
                     LOG.warn("update of Task {} returned invalid state {}", this.getClass(), newStatus);
                 }
                 status = newStatus;
+                if(newStatus == Status.FAILURE || newStatus == Status.SUCCESS) {
```
msteiger: Why is this better than calling `onTerminate()` in the corresponding switch cases?

engine/src/main/java/org/terasology/logic/behavior/tree/Task.java
```diff
@@ -82,6 +82,7 @@ public Status tick(float deltaSeconds) {
             case NOT_INITIALIZED: {
                 onInitialize();
                 status = Status.RUNNING;
```
msteiger: This would become obsolete, I guess.

engine/src/main/java/org/terasology/logic/behavior/tree/Task.java
pollend: It could still be in the running state.

Cervator:
Went ahead and merged this so PR checks will look less scary, even if maybe this isn't perfect yet. I added a note in #2070 to give it another pass in the future.

Thanks!

PR #2249, opened 2016-03-22
started working on gl state machine

pollend:
### Contains

This is some initial work on developing a GLStatemachine to avoid additional calls sent into Opengl.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/opengl/OpenGlStateMachine.java
+++ b/engine/src/main/java/org/terasology/rendering/opengl/OpenGlStateMachine.java
@@ -1,0 +1,150 @@
+package org.terasology.rendering.opengl;
+
+import org.lwjgl.opengl.GL11;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashSet;
+
+import static org.lwjgl.opengl.GL11.*;
+
+/**
+ * Created by michaelpollind on 3/22/16.
+ *
+ * This is an intermediary state machine to avoid additional calls to OpenGL
+ */
+public class OpenGlStateMachine {
+
+    private static final Logger logger = LoggerFactory.getLogger(OpenGlStateMachine.class);
+
+    //active states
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
@GooeyHub: ok to test

@pollend: thanks for the code :-) Shouldn't there be something more to then actually take advantage of the new class? How would that/this interact with PR #2240?

Somewhat reluctant to merge in an orphaned class before there's something ready to use it. Up to @emanuele3d tho if it makes sense to him!

pollend:
I wouldn't just merge it in. I might have forgoten to set the initial states correctly though.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/514/

msteiger:
Thanks for the contribution, @pollend. I'm somewhat skeptical that the equals checks make up for the state change it tries to avoid. Evaluating conditional statements could be more expensive that setting a simple `float` value, but that depends on the attribute, I guess. Most state changes are not that expensive, afaik. Maybe you can run a (micro-)benchmark to verify that the proposed changes are worth the effort?

pollend:
here are some benchmarks i did. It depends on what gl calls you are making. some are more expensive then others.

https://gist.github.com/pollend/da35bf1c183e6de49341

Hello LWJGL 3.0.0b SNAPSHOT!
GL Enable and Disable
Total execution time: 48ms
Total execution time: 83ms
GL Cull
Total execution time: 8ms
Total execution time: 23ms

msteiger:
Thanks for the effort and the information on this. I'd assume that Java (or rather the HotSpot VM) plays in your hands by (re-)compiling often-used instruction lists. But this is a different topic.

There's still a few refactorings open that could significantly boost the rendering performance. One would be to investigate the high memory throughput even if the 3D scene does not change. At least 500megs are allocated/deallocated per second. I'm sure that some of it can be reused directly avoid "slow" memory management and GCing.

PR #2251, opened 2016-03-23
Added more reasonable mass values

### LINKED ISSUE #2071: Review "weight" of base objects
Currently most our mass/weight settings are probably pretty arbitrary. This can result in *very* different results when throwing objects. A stone block may barely fly while if you try one of the starting tools like a pickaxe it may fly off like it has rockets attached.

We should check the basic items and make sure they have more sensible mass/weight so they won't fly quite as weirdly.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29229261-review-weight-of-base-objects?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tyuiwei:
### Contains
Fixes #2071 

### How to test

Throw objects around - namely, chest, wood blocks, soil blocks, pickaxe, axe and shovel. Make sure that you are using the same amount of force to throw!

### Notes

- The iconItem objects still behave differently from the usual blocks; this may have to do with something deeper in the physics engine. However, the masses in the objects are reasonable, and should continue to be like this even if there are later changes in the physics engine.


GooeyHub:

msteiger:
ok to test

msteiger:
Looks good to me - haven't tested the code in game though.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/515/

Cervator:
I think I spotted part of the likely issue :-)

In something like `axe.prefab` you have mass out on its own, likely doing nothing, changing the value has no effect:

```json
{
    "parent" : "engine:iconItem",
    "DisplayName" : {
        "name" : "Axe"
    },
    "Item" : {
        "icon" : "engine:items#axe",
        "damageType" : "axeDamage"
    },
    "mass": 42
}
```

But if you check `itemPickup.prefab` mass is one level deeper in, under `RigidBody` which affects the value on the entity's `RigidBodyComponent`

```json
{
    "Location" : {
        "scale" : 0.3
    },
    "Pickup": {
    },
    "Lifespan": {
        "lifespan": 30
    },
    "RigidBody" : {
        "mass" : 1,
        "collisionGroup" : "engine:debris",
        "collidesWith" : ["engine:world","engine:debris","engine:sensor"]
    }
}
```

So if I took that block and applied it to `iconItem.prefab` with a high mass suddenly all tools become super heavy, even if thrown with great force. 

```json
{
    "parent" : "engine:itemBase",
    "BoxShape": {
        "extents": [
            2,
            2,
            0.2
        ]
    },
    "RenderItemIconMesh": {
    },
    "RigidBody" : {
        "mass" : 100,
        "collisionGroup" : "engine:debris",
        "collidesWith" : ["engine:world","engine:debris","engine:sensor"]
    }
}
```

As an additional note changing the *parent* prefab with the game running doesn't seem to affect the children prefabs, probably because they don't know the parent has changed. If you change `axe.prefab` directly then it does take effect immediately.

With that little discovery it may be possible to fix this up a bit deeper :-)

I'm not sure if we should default `iconItem.prefab` to a specific weight, but probably? Since everything by default becomes too light. Maybe not super heavy.

I'm also not sure what, if anything, the plain `mass` out on its own connects to. It quite possibly could be a confusion between `.block` files where `mass` is a valid property and `.prefab` files where nearly everything has to be defined within a component block.

See what you can play around with that feels right :-)

Cervator:
@tyuiwei poke! Hoping you might find some time to take another look at this soon :-)

Cervator:
So I totally got myself turned around above. Adding RigidBody like above to `iconItem.prefab` (the object while in the toolbar) results in the butterfingers issue @SkySom ran into with Rails, which in turn came from (I believe) when @Josharias split apart the held item from the player so it is its own independent thing.

It was actually varying the weight in `itemPickup.prefab` that affected the weight of the thrown items without making them fall out of the toolbar. So just changing that default to 10 would probably complete this.

However, varying weight on an item-by-item basis may be trickier now. We used to be able to just have one object (the tool bar icon) and I guess it wasn't affected by some stuff as it was part of the player when held. Now it would be, so how would you selectively make a pickaxe heavier than an axe? May have to make another `.prefab` like @SkySom for Rails?

Mostly thinking out loud here, some of this is way beyond the scope of this PR :-)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/558/

Cervator:
Talked to @tyuiwei on Slack and sorted out remaining issues, threw around a bunch of stuff to test, much better now. Thanks! Added to credits etc :-)

PR #2252, opened 2016-03-23
Fixes #2174 - adds a useful error message in case of giveItem with an ambiguous item

### LINKED ISSUE #2174: Better error message when console-requesting ambiguous items
At present if you `giveitem axe` while running JoshariasSurvival you get:

`org.terasology.logic.console.commandSystem.exceptions.CommandExecutionException: java.lang.IllegalArgumentException: No block found for 'axe'`

Plus a big stacktrace in the game log. Works fine with plain Core. The issue is that both "Core:axe" and "ManualLabor:axe" exist, as easily visible with `search axe` - so really the problem isn't that `axe` doesn't exist, it is that the reference is too vague.

We should present a user a better error message and ideally offer the available suggestions ("It looks like you're trying to request an 'axe' - did you perhaps mean 'Core:axe' or 'ManualLabor:axe'?" - with Clippy mascot optional..)

The stacktrace in the log also isn't very useful since yes we know what the issue is. Probably a simpler warning should be logged instead (and if it is giveItem maybe it should say item instead of block)

@Limeth do you think this is similar to some of the auto-completion we've done in the past or maybe something the `search` command or some other method could achieve easily? I also wonder if there's a way to make `search axe` return only whole words or something .. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31107352-better-error-message-when-console-requesting-ambiguous-items?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

planetguy32:
### Contains

Fixes issue #2174 

### How to test

Using JoshariasSurvival, type into the console "giveItem axe". It will now list all items called "axe", instead of saying it couldn't find an item called "axe".

Type into the console "giveItem Stone" and it will still give you stone.

Type into the console "giveItem nothingWithThisNameExists" and it will tell you that no such item exists.

The wording on the details could be changed.

### DIFF
--- a/modules/Core/src/main/java/org/terasology/logic/inventory/ItemCommands.java
+++ b/modules/Core/src/main/java/org/terasology/logic/inventory/ItemCommands.java
@@ -16,6 +16,9 @@
 
 package org.terasology.logic.inventory;
 
+import com.google.common.base.Joiner;
+import org.terasology.assets.ResourceUrn;
+import org.terasology.assets.management.AssetManager;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.prefab.Prefab;
@@ -30,6 +33,8 @@ import org.terasology.network.ClientComponent;
 import org.terasology.registry.In;
 import org.terasology.world.block.entity.BlockCommands;
 
+import java.util.Set;
+
 /**
  */
 @RegisterSystem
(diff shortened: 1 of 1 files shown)


GooeyHub:

msteiger:
ok to test

msteiger:
Looks good to me - haven't tried it out though.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/516/

msteiger:
Thanks - works as expected!

PR #2255, opened 2016-03-24
Rearranged gitignore, minor fixes

LadySerenaKitty:
`.gitignore` - rearranged so ignored files stay ignored, added new Mac-specific ignores for things introduced in Mavericks, Yosemite, and El Capitan

`TerasologyEngine.java` - 1MB is `1048576` bytes, so we should just use that number instead.

`ProtobufPersistedData.java` - added a few things to (hopefully) aid in debugging

`facades/PC/build.gradle` - added "debug" task, same as "game" task but starts the JVM in debug mode (useful for IDEs)

It all passed in checkstyle.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
+++ b/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
@@ -20,7 +20,13 @@ import com.google.api.client.repackaged.com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.Queues;
 import com.google.common.collect.Sets;
-
+import java.util.Collection;
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.assets.AssetFactory;
@@ -77,14 +83,6 @@ import org.terasology.world.block.sounds.BlockSoundsData;
 import org.terasology.world.block.tiles.BlockTile;
 import org.terasology.world.block.tiles.TileData;
 
-import java.util.Collection;
-import java.util.Deque;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
-import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.TimeUnit;
-
 /**
  * <p>
  * This GameEngine implementation is the heart of Terasology.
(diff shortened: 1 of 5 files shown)


GooeyHub:

msteiger:
ok to test

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
```diff
@@ -114,7 +114,7 @@
 
     private static final Logger logger = LoggerFactory.getLogger(TerasologyEngine.class);
 
-    private static final int ONE_MEBIBYTE = 1024 * 1024;
```
msteiger: This point here is that 1024 * 1024 is easier to understand. The actual value is computed at "compile time" so there's no overhead involved.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/521/

facades/PC/build.gradle
```diff
@@ -81,11 +81,33 @@ task game(type:JavaExec) {
     classpath project(':engine').configurations.runtime
 }
 
+task debug(type:JavaExec) {
```
msteiger: Is this copy & paste from the `game` task? If so, maybe you can extend the existing task or derive both from a common ancestor to avoid code duplication.

I'm somewhat skeptical that this is a realistic use case, but I see potential for the `runServer` task. It would be good if JMX access could be exposed that easily.

I haven't checked, but I think that `server` has become obsolete, you might not need it. Also, the port `1044` should be communicated somehow, e.g. through logging? It would be nice if the `suspend` flag was (optionally) configurable to enable debugging the initialization process.

Apologies for the extra wishes :smile:

LadySerenaKitty: I don't know enough about gradle to be making those kinds of modifications, as I normally use maven.  The port number can be anything, and a decent IDE should pick it up.  I've changed it numerous times and my IDE picks up the change every time, I dunno about other IDEs, tho.  If anything, debuggers themselves are configurable.

I don't know if it's possible, but I'd like the debug task to take arguments and then go on to the next task, so there could be something like this:
`./gradlew debug port=1044,suspend=y game`
or this:
`./gradlew debug port=1033,suspend=y server`

Cervator:
Merged this in with the `ONE_MEBIBYTE` piece changed back to the original. I don't mind the Gradle bit left as is since it seems useful with Netbeans and the duplication is minor. Sadly the added parameters aren't as easy to wish for as Gradle does the whole `-Psome.Property=value` rather than allow for clean parameter submission. Which as actually how we got the `fetchModuleX` type tasks - the parameter is dynamically built right into the task name.

There might be a way to do that with run task variants and Gradle should be cleaned up more later. Some of that is still even outside my range and I love tweaking at Gradle. It is magic, but sadly sometimes near the dark arcane type. Gradle wizardry?

Issue #966, opened 2014-02-16
Changing graphic settings in-game results in odd behavior like models or blocks going all black or white


skaldarnar:
Starting the game on *Nice* or *Epic* will work fine, models are lit and you can see the textures.
![terasology-140216132849-1152x700](https://f.cloud.github.com/assets/1448874/2180221/51292c7c-9706-11e3-9058-f4c755480efd.png)
Changing to *Insane* or *Uber* will result in black models. 
![terasology-140216132909-1152x700](https://f.cloud.github.com/assets/1448874/2180222/54e238ea-9706-11e3-80b0-b5fba0a6df77.png)
Switching the seetings back does not solve the problem, models will stay black.

emanuele3d:
Just checking, is this still an issue?

Cervator:
Yep. I still get all-white dropped objects after having changed settings in-game.

emanuele3d:
Please reassign it to me for the time being. One day I might be able to fix this.

Translator5:
Terasology Alpha 64
Linux Mint 17-3
Java JDK 8
Installed using the launcher

![terasology-160320010345-1920x1080](https://cloud.githubusercontent.com/assets/9306861/13908572/8e86436c-ef05-11e5-9e95-6838b66be891.png)

Cervator:
Closing as complete, huzzah!

PR #2256, opened 2016-03-25
fix to resolve problems with entities appearing as either white or black

pollend:
### Contains

This resolves issue #966. This chunk of code just rebinds all the uniforms due to the fact that some uniforms may not update continuously at run-time. The shader is recompiled at this stage but, this ends up unbinding all the values bound to the last instance of the shader. This resolves the problem by just re-binds all the necessary uniforms. This problem may also exist on GLSL shader, but that issue has not shown up yet!

### How to test

run the game and change the settings.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/opengl/GLSLMaterial.java
+++ b/engine/src/main/java/org/terasology/rendering/opengl/GLSLMaterial.java
@@ -61,6 +61,7 @@ public class GLSLMaterial extends BaseMaterial {
 
     private TObjectIntMap<String> bindMap = new TObjectIntHashMap<>();
     private TIntObjectMap<Texture> textureMap = new TIntObjectHashMap<>();
+
     private GLSLShader shader;
     private boolean activeFeaturesChanged;
     private TObjectIntMap<UniformId> uniformLocationMap = new TObjectIntHashMap<>();
@@ -72,11 +73,13 @@ public class GLSLMaterial extends BaseMaterial {
     private ShaderParameters shaderParameters;
 
     private DisposalAction disposalAction;
+    private MaterialData materialData;
 
     public GLSLMaterial(ResourceUrn urn, AssetType<?, MaterialData> assetType, MaterialData data) {
         super(urn, assetType);
         disposalAction = new DisposalAction(urn);
         getDisposalHook().setDisposeAction(disposalAction);
+        this.materialData = data;
         shaderManager = CoreRegistry.get(ShaderManager.class);
         reload(data);
     }
@@ -143,6 +146,7 @@ public class GLSLMaterial extends BaseMaterial {
         }
         disposalAction.shaderPrograms.clear();
         uniformLocationMap.clear();
+        bindMap.clear();
 
         disposalAction.shaderPrograms.put(0, shader.linkShaderProgram(0));
         for (Set<ShaderProgramFeature> permutation : Sets.powerSet(shader.getAvailableFeatures())) {
@@ -152,6 +156,12 @@ public class GLSLMaterial extends BaseMaterial {
         if (shaderParameters != null) {
             shaderParameters.initialParameters(this);
         }
+
+        //resolves #966
+        //Some of the uniforms are not updated constantly between frames
+        //this function will rebind any uniforms that are not bound
+        rebindVaribles(materialData);
+
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

Huzzah! :-)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/523/

engine/src/main/java/org/terasology/rendering/opengl/GLSLMaterial.java
```diff
@@ -152,6 +156,43 @@ public void recompile() {
         if (shaderParameters != null) {
             shaderParameters.initialParameters(this);
         }
+
+        //solution due to problems with uniforms.
```
msteiger: It would be great if you could be a little more specific in this comment here. Maybe refer to issue #966 and/or wrap it in a method `rebindMaterial()`?
People will then know why this code part exists and what happens if they consider removing it.

engine/src/main/java/org/terasology/rendering/opengl/GLSLMaterial.java
```diff
@@ -152,6 +156,43 @@ public void recompile() {
...
+        for (Map.Entry<String, Float> entry : materialData.getFloatParams().entrySet()) {
+            setFloat(entry.getKey(), entry.getValue());
+        }
+
+        for (Map.Entry<String, Integer> entry : materialData.getIntegerParams().entrySet()) {
+            setInt(entry.getKey(), entry.getValue());
+        }
+
+        for (Map.Entry<String, float[]> entry : materialData.getFloatArrayParams().entrySet()) {
+            switch (entry.getValue().length) {
```
msteiger: This is nitpicking, but can you please extract a local variable for `getValue()`? It is accessed many times over the next couple of lines.

msteiger:
Great stuff - this is long overdue! Thanks, @pollend

engine/src/main/java/org/terasology/rendering/opengl/GLSLMaterial.java
Josharias: I am curious, is that a performance concern? Or a readability concern?

msteiger: It's just a readability thing. The generated bytecode should be identical, but I haven't checked.

pollend: This can't really hurt performance. This is only called when the shader is created and when it gets recompiled when you change options in settings.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/524/

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/525/

engine/src/main/java/org/terasology/rendering/opengl/GLSLMaterial.java
msteiger: Create a new local variable `value = entry.getValue()` and use that instead. I find that easier to read if used many times in a row. Again, it's a minor issue, really.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/526/

engine/src/main/java/org/terasology/rendering/opengl/GLSLMaterial.java
```diff
@@ -72,11 +73,13 @@
...
 
     private DisposalAction disposalAction;
+    private MaterialData materialData;
 
     public GLSLMaterial(ResourceUrn urn, AssetType<?, MaterialData> assetType, MaterialData data) {
         super(urn, assetType);
         disposalAction = new DisposalAction(urn);
         getDisposalHook().setDisposeAction(disposalAction);
+        this.materialData = data;
         shaderManager = CoreRegistry.get(ShaderManager.class);
```
emanuele3d: Pinging @flo for advice on future cases like this: here is a CoreRegistry usage. In this case, what would be the best way to get the Context instance? Or should we use the injection decorator to fill the shaderManager field?

flo: In that case I think the best way would be the constructor. The instancing class has already a Context field it could pass to it.

emanuele3d: Understood. Thank you Florian.

Issue #2218, opened 2016-03-09
Drag & dropping items with the mouse besides a dialog should drop them to the ground


flo:
Drag & dropping items with the mouse besides a dialog should drop them to the ground.

Currently the only way to clear an inventory is via hotkey which is not user friendly.

flo:
Ideal would be a solution that works for all modules.

If none can be found, it would be acceptable for now just to have the inventory dialog actually convering the whole screen with an invisible background area that accepts drops from the mouse and makes the item drop to the ground.

genuinelucifer:
@flo I was just looking for a place to start working on the issue and found that ```InventoryScreen.java``` already exists. What I couldn't find was how to activate/open it once in the game. Sorry, if it's silly. I opened the ```Builder Sample Gameplay``` and with right click I could drop only cubes and with left click erase them. What are the hotkeys for other shapes?
Is there some fundamental wiki page where I could learn about building gameplays?

genuinelucifer:
Figuring out this one is seeming to be more difficult than I accepted.
.
Firstly, when I did ```giveItem Core:TNT 5``` I recieved 5 TNT blocks which I could see in the slots when I pressed ```i``` to open the inventory. Now, I can only see sand in those boxes If I collect sand. If I give myself TNT or anything else it is NOT being shown in the inventory, rather If I scroll or use the numbers 0-9 then I am able to use TNT but can't see it in the inventory!!
What am I missing?
.
Also, I was looking for the codes where mouse drag drops of slot Items are handled. I went through ```InventoryCell.java``` which apparently has an ```InteractionListener``` but the drag drop is NOT handled there. 
Then I went ahead to check if ```InventoryUtils`` is doing this. I placed ```System.out.println``` in almost all the functions to figure out the order in which the functions are called once I dragged and dropped the Sand in inventory screen. But, NO output could be seen from the println statements!!!
Then I went ahead and did the same thing with some other classes related to inventory but couldn't figure it out.
.
Pleeease help..... atleast to find out where the drag drop on inventory screen is handled. And why can't I see anything else except sand/dust on my inventory screen?!

flo:
Builder Sample GAmplay is the wrong place to start as it is an example for a mode without inventory...

There is a hotkey for opening the inventory. The hotkey is I.

genuinelucifer:
[...]

PR #2262, opened 2016-03-27
Trying to add drag drop of items from inventory

genuinelucifer:
Fixing #2218 
.
I added an interaction region on all sides of the InventoryGrid, so the cells are put in the center of the screen (the offset is calculated by giving half of width left after drawing the cells).
There are 4 rectangular regions which need to be targetted. I have added all 4 with appropriate names.

I just tried to immitate the existing functionality of right click.. So the code is adapted version of [onUseItemButton](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java#L348) from ```LocalPlayerSystem```
.
I Expected that if I drag an item out of the inventory cells (on the inventory screen) and drop it on empty area around the grid then it would get dropped on the terrain. I saw that on right click I could put items on the terrain. So, I used its code.
.
But instead what happens is actually (i think) similar to if I press left click while inventory screen is not opened. i.e, the player tries to destroy the block which is just ahead (pointed to).
.
I tried to debug it checking for the function calls and yes all the lines are being executed as required. The problem that I think (as @flo mentioned) is that I couldn't understand the parameters which are passed to [onUseItemButton](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java#L348) and hence duplicated it in a wrong way.
.
.
I will now try to add the overlay as suggested.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/MenuControlSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/MenuControlSystem.java
@@ -46,6 +46,7 @@ public class MenuControlSystem extends BaseComponentSystem {
 
     @Override
     public void initialise() {
+        nuiManager.getHUD().addHUDElement("dropItemRegion");  //Ensure the drop region is behind the toolbar
         nuiManager.getHUD().addHUDElement("toolbar");
     }
 
--- a/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryGrid.java
+++ b/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryGrid.java
@@ -19,16 +19,20 @@ import com.google.common.base.Function;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.input.MouseInput;
 import org.terasology.logic.inventory.InventoryUtils;
 import org.terasology.math.geom.Rect2i;
 import org.terasology.math.geom.Vector2i;
+import org.terasology.rendering.nui.BaseInteractionListener;
 import org.terasology.rendering.nui.Canvas;
 import org.terasology.rendering.nui.CoreWidget;
+import org.terasology.rendering.nui.InteractionListener;
 import org.terasology.rendering.nui.LayoutConfig;
 import org.terasology.rendering.nui.UIWidget;
 import org.terasology.rendering.nui.databinding.Binding;
 import org.terasology.rendering.nui.databinding.DefaultBinding;
 import org.terasology.rendering.nui.databinding.ReadOnlyBinding;
+import org.terasology.rendering.nui.events.NUIMouseClickEvent;
 
 import java.util.Iterator;
 import java.util.List;
@@ -46,6 +50,17 @@ public class InventoryGrid extends CoreWidget {
 
     private List<InventoryCell> cells = Lists.newArrayList();
     private Binding<EntityRef> targetEntity = new DefaultBinding<>(EntityRef.NULL);
+
+    private InteractionListener interactionListener = new BaseInteractionListener() {
+        @Override
+        public boolean onMouseClick(NUIMouseClickEvent event) {
+            MouseInput mouseButton = event.getMouseButton();
+            if (mouseButton == MouseInput.MOUSE_LEFT) {
+                return true;
+            }
+            return false;
+        }
+    };
 
     @Override
     public void update(float delta) {
@@ -83,6 +98,8 @@ public class InventoryGrid extends CoreWidget {
         if (cellSize.getX() == 0 || cellSize.getY() == 0) {
             return;
         }
+        canvas.addInteractionRegion(interactionListener);
+
         int horizontalCells = Math.max(1, Math.min(maxHorizontalCells, canvas.size().getX() / cellSize.getX()));
         for (int i = 0; i < numSlots && i < cells.size(); ++i) {
             int horizPos = i % horizontalCells;
--- a/modules/Core/assets/skins/dropItemRegionDefault.skin
+++ b/modules/Core/assets/skins/dropItemRegionDefault.skin
@@ -1,0 +1,5 @@
+{
+    "inherit" : "default",
+    "elements" : {
+    }
+}
--- a/modules/Core/assets/ui/hud/dropItemRegion.ui
+++ b/modules/Core/assets/ui/hud/dropItemRegion.ui
@@ -1,0 +1,10 @@
+{
+  "type": "DropItemRegion",
+  "skin": "dropItemRegionDefault",
+  "contents": {
+        "type" : "relativeLayout",
+        "id" : "mainOverlay",
+        "contents" : [
+        ]
+  }
+}
--- a/modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
+++ b/modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
@@ -105,11 +105,12 @@ public class CharacterInventorySystem extends BaseComponentSystem {
             return;
         }
 
-        // remove a single item from the stack
+        int count = event.getCount();
+        // remove 'count' items from the stack
         EntityRef pickupItem = event.getItem();
         EntityRef owner = pickupItem.getOwner();
         if (owner.hasComponent(InventoryComponent.class)) {
-            final EntityRef removedItem = inventoryManager.removeItem(owner, EntityRef.NULL, pickupItem, false, 1);
+            final EntityRef removedItem = inventoryManager.removeItem(owner, EntityRef.NULL, pickupItem, false, count);
             if (removedItem != null) {
                 pickupItem = removedItem;
             }
--- a/modules/Core/src/main/java/org/terasology/logic/inventory/events/DropItemRequest.java
+++ b/modules/Core/src/main/java/org/terasology/logic/inventory/events/DropItemRequest.java
@@ -32,15 +32,22 @@ public class DropItemRequest implements Event {
     private EntityRef inventory = EntityRef.NULL;
     private Vector3f impulse;
     private Vector3f newPosition;
+    private int count;
 
     protected DropItemRequest() {
     }
 
-    public DropItemRequest(EntityRef usedItem, EntityRef inventoryEntity, Vector3f impulse, Vector3f newPosition) {
+    public DropItemRequest(EntityRef usedItem, EntityRef inventoryEntity, Vector3f impulse, Vector3f newPosition, int count) {
         this.item = usedItem;
         this.inventory = inventoryEntity;
         this.impulse = impulse;
         this.newPosition = newPosition;
+        this.count = count;
+    }
+
+    public DropItemRequest(EntityRef usedItem, EntityRef inventoryEntity, Vector3f impulse, Vector3f newPosition)
+    {
+        this(usedItem, inventoryEntity, impulse, newPosition, 1);
     }
 
     public EntityRef getItem() {
@@ -57,5 +64,9 @@ public class DropItemRequest implements Event {
 
     public Vector3f getImpulse() {
         return impulse;
+    }
+
+    public int getCount() {
+        return count;
     }
 }
--- a/modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/DropItemRegion.java
+++ b/modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/DropItemRegion.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2016 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.nui.layers.hud;
+
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.input.MouseInput;
+import org.terasology.logic.characters.CharacterComponent;
+import org.terasology.logic.inventory.InventoryUtils;
+import org.terasology.logic.inventory.events.DropItemRequest;
+import org.terasology.logic.players.LocalPlayer;
+import org.terasology.math.geom.Vector3f;
+import org.terasology.registry.In;
+import org.terasology.rendering.nui.BaseInteractionListener;
+import org.terasology.rendering.nui.Canvas;
+import org.terasology.rendering.nui.InteractionListener;
+import org.terasology.rendering.nui.events.NUIMouseClickEvent;
+
+/**
+ * A region/layer around the inventory grid to allow players to get rid of extra items
+ * and have free inventory slots by dropping them onto this layer.
+ *
+ */
+public class DropItemRegion extends CoreHudWidget {
+
+    @In
+    private LocalPlayer localPlayer;
+
+    private InteractionListener interactionListener = new BaseInteractionListener() {
+        @Override
+        public boolean onMouseClick(NUIMouseClickEvent event) {
+            MouseInput mouseButton = event.getMouseButton();
+            if (mouseButton == MouseInput.MOUSE_LEFT) {
+                EntityRef playerEntity = localPlayer.getCharacterEntity();
+                EntityRef movingItem = playerEntity.getComponent(CharacterComponent.class).movingItem;
+                EntityRef item  = InventoryUtils.getItemAt(movingItem, 0);
+                if (!item.exists()) {
+                    return true;
+                }
+                int count = InventoryUtils.getStackCount(item);
+
+                Vector3f position = localPlayer.getViewPosition();
+                Vector3f direction = localPlayer.getViewDirection();
+                Vector3f newPosition = new Vector3f(position.x + direction.x * 1.5f,
+                        position.y + direction.y * 1.5f,
+                        position.z + direction.z * 1.5f
+                );
+                //send DropItemRequest
+                Vector3f impulseVector = new Vector3f(direction);
+                playerEntity.send(new DropItemRequest(item, playerEntity,
+                        impulseVector,
+                        newPosition,
+                        count));
+                return true;
+            }
+            return false;
+        }
+    };
+
+    @Override
+    public void initialise() {
+    }
+
+    @Override
+    public void onDraw(Canvas canvas) {
+        super.onDraw(canvas);
+        canvas.addInteractionRegion(interactionListener);
+    }
+}


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/529/

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/530/

Cervator:
Heya @genuinelucifer and thanks for giving this a try. Renamed the title real quick to make it clear this isn't for merging yet :-)

Pinging @flo or @Josharias for review sometime

flo:
Just a "Currently doesn't work" isn't very helpful. You don't even write what is not working .Typically information for a problem description is that you write 
1. what you tried (steps to reproduce)
2. what you expected
3. what happened.
4. what you have tried already to solve it

When you create a pull request it is also good to write a short summary of what you changed and why.

I tried your branch out. When I open the inventory (via I hotkey) and click next to it, then a NullPointerException gets logged. The exception gets thrown in the interaction listener of InventoryGrid because the entity movingItem has no ItemComponent and saveComponent does not allow null values.

When you look at the usages (IDE feature!) of the movingItem field you will notice that it is actually a entity with a inventory and not a item (@Josharias: I think that name is quite missleading. Could you please rename the movingItem field to say "cursorInventory"? Also a javadoc would be nice (that explains what to expect and why it is an inventory and not directly the item *end of note to Josharias*)

e.g. when you use the "find usages" features you will find the following line in TransferItemCursor:
```
                        return InventoryUtils.getItemAt(charComp.movingItem, 0);
```
When you then look at getItemAt you will notice that it expects an entity with an inventory.

Also you will notice that the current location for the interaction catching is not optimal since it will also absorb all clicks that were intended for the toolbar at the bottom. So we probably need the "click => item drop" feature in an overlay that is below the overlay with the toolbar.

To find the location where the toolbar UI gets added to the screen I added a breakpoint to the initialize method of HudToolbar.  The toolbar gets added in MenuControlSystem to a instance of HUDScreenLayer. Sadly there is currently no way to define the order of initialization for systems. So it is not possible to have the toolbar added after a "item drop default handling" overlay.

So to me it looks like the simplest approach for now would be to add the new overlay in MenuControlSystem. 

Another possibility would to add a new addInteractionRegion to Canvas that adds the region to the bottom of the interaction regions. Changes to the NUI interfaces is however something we should clearify with @immortius  first.

@immortius  any thoughts on that topic?

genuinelucifer:
@Cervator @flo Thanks.
.
I've updated the PR with details. I'll try to implement the overlay and push the changes.

genuinelucifer:
Thanks a lot @flo.
It now works as desired. I only couldn't try it with the toolbar as you said.
I'll try that and push the changes as soon as it gets done.

modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryGrid.java
```diff
@@ -55,6 +55,23 @@
     private List<InventoryCell> cells = Lists.newArrayList();
     private Binding<EntityRef> targetEntity = new DefaultBinding<>(EntityRef.NULL);
 
+    private InteractionListener interactionListener = new BaseInteractionListener() {
+        @Override
+        public boolean onMouseClick(NUIMouseClickEvent event) {
+            MouseInput mouseButton = event.getMouseButton();
+            if (mouseButton == MouseInput.MOUSE_LEFT) {
+                LocalPlayer localPlayer = CoreRegistry.get(LocalPlayer.class);
+                EntityRef movingItem = CoreRegistry.get(LocalPlayer.class).getCharacterEntity().getComponent(CharacterComponent.class).movingItem;
```
flo: oh and new code should avoid core registry. you can get a instance of LocalPlayer by defining a field annotated with `@In` in UI screens and systems.

genuinelucifer:
@flo I tried but could not find the toolbar you are talking about. All I could find was open the chat and cheat consoles. Interaction with them is not blocked by current inventory grid.
How to display the HUDToolbar?

modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryGrid.java
genuinelucifer: I tried to add ```@RegisterSystem``` to ```InventoryGrid``` and then added a field with:
```
@In
private LocalPlayer localPlayer;
```
But it doesn't work. The variable is null at runtime.
Also, I tried to implement ```ComponentSystem``` in InventoryGrid but that didn't work too.
.
I did these things because all the classes with ```@In``` variable had atleast one of these 2 features.
How do I do that?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/564/

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/565/

genuinelucifer:
@flo Please Review. I have added the overlay as you said.
.
The only problem which I feel now is that if user hides the hud with ```h``` key then the drag drop also stops working. I don't know if should be considered as a bug or an intended feature.

Cervator:
Hey @genuinelucifer I tried to test this but either nothing is happening or I'm not sure what to test. I tried to pick up an item from an inventory (left click item) then move the cursor outside the inventory grid area to open space then left or right clicked and nothing happened. Tried dragging an item and releasing too. Tried player inventory and a chest.

As for the hud hiding with 'h' I'm remembering that doesn't hide the inventory screen which I would think it should. And if it did it would probably be intended for drag/drop to not work with everything hidden :-)

genuinelucifer:
@Cervator When you left click outside the inventory grid please make sure that hud is visible (that is, you can see hearts and 8-9  square blocks around the bottom of screen).
.
If the inventory would be hidden with 'h' that would be surely ok.

genuinelucifer:
Also if that doesn't work could you please place a break point on line number 45 ```if (mouseButton == MouseInput.MOUSE_LEFT) {``` in ```interactionListener``` in the file ```InventoryGridHud.java``` and run in debug mode to check if it is called when you left click outside the inventory grid.

Cervator:
Yep the break point triggers fine, even just left clicking without any inventory interaction at all. With the specific use case here it seems to validate that the item exists then does the `localPlayer.activateOwnedEntityAsClient(item) and keeps going for a while, I didn't trace too much further since I'm not sure what to look for :-) Item I picked up just stays on the cursor.

Also fun fact: same happens if you drop an item into the corners between inventory slots. That *may* not be desired, as a player could accidentally trigger this just moving items between slots inside the inventory grid area. I figure there must be some UI element that all the slots are nested inside that might be a better thing to look outside of.

genuinelucifer:
I actually tried it again. I took a torch from the inventory grid and when I click outside the grid then exit inventory I find it lying on the ground. (Maybe you forgot to point at a nearby ground block? (just a wild guess)).
.

> same happens if you drop an item into the corners between inventory slots.

This surely needs to be fixed. I just added the overlay and probably it gets all the clicks that are not handled by the cells and the toolbar.
I'll fix this in the next commit.

Cervator:
Oh, you have to be pointing at the ground? Had no clue - is that on purpose or just some consequence of needing a block's location to be dropping stuff?

genuinelucifer:
It is because of needing the location to drop the object to.

flo:
hmm, for dropping items with the Q hotkey you don't need a location I would assume.

modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/InventoryGridHud.java
```diff
@@ -0,0 +1,68 @@
...
+                localPlayer.activateOwnedEntityAsClient(item);
+            }
+            return true;
+        }
+    };
+
+    @Override
+    public void initialise() {
+        RelativeLayout overlay = find("mainOverlay", RelativeLayout.class);
+        overlay.onBindEvent(new BindButtonEvent());
```
flo: not sure what this line is for. Probably explaining it in comment would be good.

modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/InventoryGridHud.java
```diff
@@ -0,0 +1,68 @@
...
+        @Override
+        public boolean onMouseClick(NUIMouseClickEvent event) {
+            MouseInput mouseButton = event.getMouseButton();
+            if (mouseButton == MouseInput.MOUSE_LEFT) {
+                EntityRef movingItem = localPlayer.getCharacterEntity().getComponent(CharacterComponent.class).movingItem;
+                EntityRef item  = InventoryUtils.getItemAt(movingItem, 0);
+                if(!item.exists()) {
+                    return true;
+                }
+                localPlayer.activateOwnedEntityAsClient(item);
```
flo: Hmm activating the item? Don't we want to just drop it?

I would have expected something like in CharacterInventorySystem#onDropItemRequest
```
            entity.send(new DropItemRequest(selectedItemEntity, entity,
                    impulseVector,
                    newPosition));
```

genuinelucifer: It puts the item on the terrain. Like doing a right click.
I didn't call the event as i couldn't figure out exact parameters to pass.
But I am doing nearly the same thing as is done in the event.

genuinelucifer:
I think it's because when I press Q. The item does not appear on the
screen. It just vanishes!
With right click, it gets put on the terrain.

genuinelucifer:
@Cervator I couldn't yet figure out how to clear the issue for drop between 2 cells. I cannot access the ```cells``` in an interactionListener and hence am unable to find which is the region between cells.
.

> not sure what this line is for. Probably explaining it in comment would be good.

@flo That code was by mistake left there. I was trying something else and then forgot to remove those 2 lines. Thanks for pointing out

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/572/

modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/InventoryGridHud.java
flo: Not all items get placed on the ground when they get activated. Some items open a UI. e.g. execute the command "giveItem inspectionTool" to get an inspection tool item that shown an UI on activation.

genuinelucifer: > Not all items get placed on the ground when they get activated.
.
Yes I checked. Some are working as if a left click has been pressed.
Should I call the event then or is this behavior desired?

genuinelucifer:
@flo Is there a way to get ```InventoryGrid`` in my ```InventoryGridHud```?
I came to know that ```@In``` won't work. I also tried ```CoreRegistry``` and then ```find("inventory", InventoryGrid.class)``` But none worked...
.
Instead of accessing ```InventoryGrid``` in another class, I also tried to define and ```InteractionListener``` in the Grid itself but then I could not access the ```cells``` from the ```interactionListener```.
.
Any suggestions?

modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/InventoryGridHud.java
flo: actually usually a right click activates an item. The idea behind this whole drop item out of window feature is to provide the user a way to get rid of items that fill up the inventory. Since an activation does not necesarly does this, please change it to a item drop for all items.

flo:
@genuinelucifer There isn't just necessarly 1 inventory grid visible at a time. e.g. when you activate a placed chest by pressing E, then you get 3 inventory grids: 1. the toolbar inventory grid, 2. the players inventory grid, 3. the chests inventory grid.

So you can't get "THE" inventory grid and also I am not sure how that would help.

CoreScreenLayer already registeres a interaction listener. I wonder why it isn't catching the drop event yet. Maybe you oculd investigate into that direction. e.g. if it handles the mouse event that causes the drop then the other interaction listener in your new class won't trigger a drop.

modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/InventoryGridHud.java
genuinelucifer: Oh. I see.
Will change it :)

genuinelucifer:
> CoreScreenLayer already registeres a interaction listener

great

> Maybe you oculd investigate into that direction.

Sure. Thanks for the help.

genuinelucifer:
@flo I tried many things but could not activate the interactionListener of ```CoreScreenLayer```...
.
Always, the interactionListener from cells is called and when I click between cells then the interactionListener from ```InventoryGridHud``` is called.
.
Isn't there a way to call the interaction listener of all the cells from the interactionListener of ```InventoryGrid``` such that any interaction between the cells could be handled in the ```InventoryGrid``` itself? Can't we somehow make the cells static (to be accessed from the interactionListener), although I didn't see static variables being used in this project anywhere.... So probably some other similar solution?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/592/

genuinelucifer:
If it is acceptable, I would like to propose the following solution to the problem:
.
Lets assign a depth value to every interactionregion as it is added. Keep on increasing the depth value by 1 on each addition. Also, provide additional function to give a required depth.
Ex: If I give the interactionRegion in ```InventoryGrid``` depth value which is sum of current priority plus the total number of cells then I can make sure that the call to handle the mouse click will go to my required interactionRegion. After this the interactionRegions from cells will be added and will be given priority as per the original numbering that was continuing (before adding InventoryGrid)...
This seems to be a solution which solves the current problem.
.
This might require changing the core way of how interactionRegions are being handled, so I would wait for a confirmation before trying to implement this.
.
Pinging @Cervator @flo for providing confirmation/denial to implement this.

Cervator:
I can't help answer the architecture part, but I can ping @rzats who may be interested in chiming in at the NUI level :-)

rzats:
@genuinelucifer: the current canvas implementation already does something somewhat similar to what you're suggesting. To process mouse interaction, [`canvasImpl#processMousePosition`](https://github.com/MovingBlocks/Terasology/blob/cbab4c6978801d2285f6d51e950066582c2d1e11/engine/src/main/java/org/terasology/rendering/nui/internal/CanvasImpl.java#L160-L206) iterates through the `interactionRegions` deque, starting from the last element - that way, recently added regions are given priority. _(Region addition logic is located in [this `addInteractionRegion` overload](https://github.com/MovingBlocks/Terasology/blob/cbab4c6978801d2285f6d51e950066582c2d1e11/engine/src/main/java/org/terasology/rendering/nui/internal/CanvasImpl.java#L705-L717).)_

That said, custom region priorities can be quite helpful if you want to add a UI region that does **not** have priority over all the previously added ones - they may not be relevant to this particular issue, but would definitely work as a separate feature!

genuinelucifer:
@rzats Thanks for the references.
.
Actually, I thought about this issue because I wanted that the interactionListener of cells should fire and then of InventoryGrid and then that of InventoryGridHud... And it seemed the only way possible to me..
Although, there might be some other solution and I'll try to find them first.

genuinelucifer:
> CoreScreenLayer already registeres a interaction listener. I wonder why it isn't catching the drop event yet. 

@flo finally I found that without any other top layer ```CoreScreenLayer``` is indeed catching the drop event. But since ```InventoryGridHud``` is covering the whole screen on top of CoreScreen Layer hence it is not catching the events...
.
I am really stuck and since I cannot access InventoryGrid from InventoryGridHud.
So, I made ```cells``` as ```final``` in ```InventoryGrid``` and then accessed it from interactionListener. Also, I had to make many static variables in the InventoryGrid to keep track of number of cells in 1 row and size of one cell. I did this because the only other solution that came to my mind (custom priority for interactionListener) seemed to be huge task specially since there are sooo many layers on one screen and I would have to change priority of each one of them for the required listeners to work!
.
Now, I am stuck with one small hurdle. I am unable to send a mouse click event to ```InventoryCell```. I want that the cell's interactionListener be triggered. I tried ```onMouseEvent``` by changing a few parameters but it didn't work. Any help as to how to send this event to an ```InventoryCell```? @rzats @Cervator @flo 
.
I know that currently cells handle their own separate callbacks via their interactionListeners. But, since I have to add a layer on top I need to call their interactionListeners manually and then decide which clicks to handle and which to leave.

rzats:
@genuinelucifer: can you please push the relevant modifications to your fork? It would really help with debugging the code to see why exactly the `interactionListener` trigger doesn't work. If you don't want to modify the PR, use a separate branch and then link it a comment.

Also (emphasis mine):
>I had to make many **_static_** variables in the InventoryGrid to keep track of number of cells in 1 row and size of one cell.

Are you sure this is the right solution? Since the variables (and, presumably, related value checks) are static, they may cause unexpected behavior when several `InventoryGrid` elements are displayed (e.g. toolbar, extended inventory, chest inventory), especially if they have different row/cell sizes.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/594/

modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryGrid.java
```diff
@@ -44,9 +48,34 @@
...
+    private InteractionListener interactionListener = new BaseInteractionListener() {
+        @Override
+        public boolean onMouseClick(NUIMouseClickEvent event) {
+            MouseInput mouseButton = event.getMouseButton();
+            if (mouseButton == MouseInput.MOUSE_LEFT) {
+                Vector2i mousPos = event.getRelativeMousePosition();
+                int horizCells = (mousPos.getX()-1) / InventoryGrid.cSize.getX();
+                int vertCells = (mousPos.getY()-1) / InventoryGrid.cSize.getY();
+                InventoryCell cell = cells.get(vertCells * horizontalCells + horizCells);
+                cell.onMouseButtonEvent(new MouseButtonEvent(event.getMouseButton(), ButtonState.DOWN, 0.1f));
```
genuinelucifer: I need this to work!

genuinelucifer:
> can you please push the relevant modifications to your fork?

@rzats Done.

> when several InventoryGrid elements are displayed 

Sorry. I forgot about this. I just targetted one InventoryGrid. I thought all the grids are derived from separate classes. I guess then I'll have to incorporate the priorityQueue for interactionListeners?
Or, do you have another Idea to store size of cells and number of cells in one row (as ```canvas``` can't be accessed from ```interactionListener``` and hence the preferredSize can not be queried...

rzats:
The problem with the `cell.onMouseButtonEvent()` call is that it calls [`CoreWidget.onMouseButtonEvent()`](https://github.com/MovingBlocks/Terasology/blob/a388715a3f43167aa2952231082d66030beb84e1/engine/src/main/java/org/terasology/rendering/nui/CoreWidget.java#L43-L45) (the class hierarchy being `InventoryCell` → `ItemCell` → `CoreWidget`) - which is an empty method. It's actually called (easy to verify by setting a debugger breakpoint), but does nothing at all.

To send an event to `InventoryCell`, you need to create an `InventoryCell` method that either exposes the interaction listener or passes along the `NUIMouseClickEvent` to it - in my opinion, the latter solution is more optimal.

>Or, do you have another Idea to store size of cells and number of cells in one row (as canvas can't be accessed from interactionListener and hence the preferredSize can not be queried...

You can calculate these in [`onDraw()`](https://github.com/MovingBlocks/Terasology/blob/2cd5a03f813c4932649ddab9710b0eb8e93f4b77/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryGrid.java#L76-L92).

genuinelucifer:
> passes along the NUIMouseClickEvent to it - in my opinion, the latter
solution is more optimal.

Great. Will try ti implement it :) Thanks.

> You can calculate these in onDraw().

But then i can't access them in interactionListener. For that I will have
to make them 'final' but then i won't be able to update them in 'onDraw'.
Feels like a deadlock.

genuinelucifer:
> > You can calculate these in onDraw().
>
> But then i can't access them in interactionListener. For that I will have
to make them 'final' but then i won't be able to update them in 'onDraw'.
Feels like a deadlock.
.
.
My bad. I have been using C++ for so long that I never knew it was possible
in java. Finally got it working. Will try to polish everything and push the
final changes (working!  Hurray!) tomorrow.
Thanks a lot @rzats

flo:
> @flo finally I found that without any other top layer CoreScreenLayer is indeed catching the drop event. But since InventoryGridHud is covering the whole screen on top of CoreScreen Layer hence it is not catching the events...
Actually what you write is not the case: The interaction listener in CoreScreenLayer does not even get added when the inventory is open as the InventoryScreen#isModal method returns false. The CoreSCreenLayer method only adds the interaction listener in onDraw if it is a modal dialog
```
        if (isModal()) {
            canvas.addInteractionRegion(getScreenListener());
        }
```
I tried making the InventoryScreen not modal and then it does indeed catch the events, but sadly it does catch mouse events that cover the complete visible screen and not just the dialog. That probably means that we can't add (without big restructuring) a general solution. that makes all clicks in dialogs not result in an item drop. It needs to be solved for each screen separately.

I think the main issue are the gaps between the inventory cells, which you can click accidentally when moving items. So lets simply set the InventoryCell margin in InventoryDefault.skin to 0 to close that gaps and we should have a solution that at least works for the screens in the engine/core. Modules need then to add additional interaction regions to their screens if they want to prevent unwanted item dropping.

Static variables and getting updating various inventory grids/cells should not be necessary. As @rzats pointed correctly out static variables cause more issues than they solve in that case.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/596/

genuinelucifer:
@flo I dropped the interactionListener from InventoryGrid and changed the margin to zero instead, as you suggested.
I tested it and it seems to work fine for me. Maybe @Cervator could confirm...
.
I'll fix the checkstyle warning after a few hours. Till then, if you could please review this.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/597/

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/598/

modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/InventoryGridHud.java
```diff
@@ -0,0 +1,74 @@
...
+import org.terasology.logic.inventory.events.DropItemRequest;
+import org.terasology.logic.players.LocalPlayer;
+import org.terasology.math.geom.Vector3f;
+import org.terasology.registry.In;
+import org.terasology.rendering.nui.BaseInteractionListener;
+import org.terasology.rendering.nui.Canvas;
+import org.terasology.rendering.nui.InteractionListener;
+import org.terasology.rendering.nui.events.NUIMouseClickEvent;
+
+public class InventoryGridHud extends CoreHudWidget {
```
flo: Please change the name as it it is not a inventory grid.

e.g. how about calling it "DropItemRegion"?

modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/InventoryGridHud.java
```diff
@@ -0,0 +1,74 @@
...
+
+                Vector3f position = localPlayer.getViewPosition();
+                Vector3f direction = localPlayer.getViewDirection();
+                Vector3f newPosition = new Vector3f(position.x + direction.x * 1.5f,
+                        position.y + direction.y * 1.5f,
+                        position.z + direction.z * 1.5f
+                );
+                //send DropItemRequest
+                Vector3f impulseVector = new Vector3f(direction);
+                impulseVector.scale(100);
```
flo: That is a lot of impulse. The items get now thrown (try a torch). I think it would be enough for the items to just drop to the ground infront of the character.

modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/InventoryGridHud.java
flo: Also a javadoc header that explains this class would be good.

engine/src/main/java/org/terasology/logic/players/MenuControlSystem.java
```diff
@@ -46,6 +46,7 @@
 
     @Override
     public void initialise() {
+        nuiManager.getHUD().addHUDElement("inventoryGridHUD");
```
flo: A comment would be good that explains why we add it here. (e.g. we want it to be behind the toolbar, and the HUD is always in the back). There should also be a TODO to find a better way of doing item drops that do not depend on the visibility of the HUD.

modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/InventoryGridHud.java
```diff
@@ -0,0 +1,74 @@
...
+                playerEntity.send(new DropItemRequest(item, playerEntity,
```
flo: Currently only 1 item gets dropped, Ideally the complete stack should be dropped. I suggest that you either implement that or add a TODO. (CharacterInventorySystem#onDropItemRequest handles the event, the event would need to be extended to take for example an amount parameter that could then be processed in the event handling method)

flo:
@genuinelucifer 	I added a few comments.

Also I think it would be probably a good idea to smash the commits together into a single one as there are a lot of commits that don't really provide value with their commit message to the version history. e.g. one commit is called "More Useless Code". If you don't know how to do that I can help you with that.

flo:
btw. before squashing or rebasing the commits I would always suggest to create a "backup" branch at the current HEAD location of your branch. That way you can easily go back to it should something go wrong.

Cervator:
Tried testing it. Maybe did something this time that I didn't last, as I *can* see the functionality - but only after one play session, creating a world, exiting the game, then reloading the world. First play session on creating the world the drop behavior from inventory did nothing.

Went back to double check and saw it function the second time - nice!

The items look oddly large as icons though, as if there is no margin inside the cells anymore. I was also still able to drop an item through the very center of corners between cells, although it was a lot harder than last time (were margins just tweaked so there'd be less interaction space between cells?)

Not enough time tonight to test deeper, but it is encouraging to see it work. Keep it up @genuinelucifer and we'll finally be done and merging this soon! :-)

As for dropping one item vs a stack, what about using left click for one, right click for the other? Could even activate the mousewheel like possible in an inventory cell (although only on down-wheel since picking stuff back up makes no sense in this context)

genuinelucifer:
> were margins just tweaked so there'd be less interaction space between
cells?

Yes. Other solutions seemed too complex. And @flo suggested this.

> Keep it up @genuinelucifer and we'll finally be done and merging this
soon! :-)
>

Sure :)

> As for dropping one item vs a stack, what about using left click for one,
right click for the other? Could even activate the mousewheel like possible
in an inventory cell (although only on down-wheel since picking stuff back
up makes no sense in this context)
>

Although this could be done but IMHO this would be more confusing. Instead,
since use can select the required amount of items from the cells, he should
only select the amount that he wishes to drop.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/599/

genuinelucifer:
> were margins just tweaked so there'd be less interaction space between
> cells?

Yes. Other solutions seemed too complex. And @flo suggested this.

> Keep it up @genuinelucifer and we'll finally be done and merging this
> soon! :-)

Sure. Thanks for testing... :)

> As for dropping one item vs a stack, what about using left click for one,
> right click for the other? Could even activate the mousewheel like possible
> in an inventory cell (although only on down-wheel since picking stuff back
> up makes no sense in this context)

Although this could be done but IMHO this would be more confusing. Instead,
since user can select the required amount of items from the cells, he should
only select the amount that he wishes to drop.

genuinelucifer:
@flo I squashed the commits. Also, did the changes that you commented on. Please review again.
.
Also, what's your view on left click for dropping one item and right click for dropping the stack. (suggested by @Cervator)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/600/

flo:
A left click drops the complete item stack into a item slot, So I think it should be analog for dropping items to the ground. Since you can pick up just a single item (with mouse wheel) to drop it outside, I don't think we need the functionality to drop a single item. Instead I think we could use right click for the feature @genuinelucifer implemented originally. Right click could activate the item (right click activates items when you have no UI open, so I think it would be consistent).

@genuinelucifer About the larger icons: As a quick solution you could remove the marign size change again and add instead an interaction region in the draw method of InventoryGrid that is below the inventory cells (order of adding regions determines which comes first). That interaction region could then catch all mouse events.

genuinelucifer:
>  Right click could activate the item

And what about for multiple items? Loop through and activate each one?

flo:
Same as like when you activate the item stack via toolbar. Usually just 1 item gets placed. No special handling needs to be implemented.

flo:
but lets do the right click behavior in a separate pull request, so that this one gets done.

genuinelucifer:
@flo I did the changes and squashed the commits. But now this PR has conflicts. I'll rebase and push again in a few minutes.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/601/

genuinelucifer:
@flo Done. Please merge, if it's acceptable now.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/602/

Cervator:
Tested it again and all seems right! Worked first try, icons looked nicer (or my imagination is playing tricks on me), a full stack dropped as a single item, and I couldn't drop anything through the corners anymore.

Merging at last, huzzah! Thank you @genuinelucifer and also many thanks to @flo (and @rzats!) for the review and pointers :-)

Submitted new issue for reference on the right click situation, please comment if interested! #2295

Issue #693, opened 2013-09-27
We got a second sun and not the good kind ;-)


Cervator:
I've noticed this for a while but wrote it off as another one of the weird shader issues afflicting me. Turns out @immortius gets it too, yet not the shader issues.

A little dim sun hangs out right by the horizon line mostly visible at night (next to the moon-sun in the following screenie:)

![sneakysun](https://f.cloud.github.com/assets/1063833/1223292/62dbc49e-2716-11e3-9d87-c33c9a4aecca.png)

Investigated a bit and it is introduced by https://github.com/MovingBlocks/Terasology/commit/d872bf701f59bcc02411846a00fb1f93aa5aaacd

This wasn't the way I was hoping for multiple celestials :D
##

emanuele3d:
Ok, just had a look at this one. It is really just a texture issue: a bright orange blurry streak needs to be painted out of the texture. Unfortunately it is a humble but artistic task, which is not exactly my cup of tea. I just tried to delegate to pruteanu. If that doesn't work I'll ask in the Art and Media forum. Anybody proficient with a paint program won't need much time to fix it, I just do not have that kind of proficiency having spent most of my career in 3D programs such as Maya and Softimage. Paradoxically it might have been quicker for me to fix if it was a shader or programming problem. =)

Cervator:
Closing as fixed by #2871 at long last :-)

PR #2264, opened 2016-03-28
Added jpeg / bmp texture support, converted menuBackground to a jpeg for te…

### LINKED ISSUE #809: Support for JPG textures
I was quite surprised to find that .jpg was not yet supported. In particular for larger textures such as background images (the current one has 4,1MB) , it might be quite valueable.

Currently, TS uses Slick2D or more precisely, Slick-Util. It uses an internal PNGDecoder for png files and awt.ImageIO for all other formats. Is there a reason for *not* using AWT directly?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1612353-support-for-jpg-textures?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

viveret:
### Contains

Fixes #809, converted menuBackground to jpeg as imagined.

### How to test

Simply run Terasology and check the menu background.


GooeyHub:

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGDecoder.java
```diff
@@ -0,0 +1,143 @@
...
+import java.awt.image.DataBuffer;
+import java.awt.image.DataBufferByte;
+import java.awt.image.WritableRaster;
+
+/**
+ * @author Viveret (github.com/viveret)
+ */
+public class JPEGDecoder {
+    public static enum Format {
+        ERR, RGB, RGBA
```
msteiger: What is `ERR`? Are these target buffer formats? There are no alpha channels in JPG.

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGDecoder.java
```diff
@@ -0,0 +1,143 @@
...
+import java.nio.ByteBuffer;
+import java.io.InputStream;
+import javax.imageio.ImageIO;
+import java.awt.image.BufferedImage;
+import java.awt.image.DataBuffer;
+import java.awt.image.DataBufferByte;
+import java.awt.image.WritableRaster;
+
+/**
+ * @author Viveret (github.com/viveret)
```
msteiger: Please add the CheckStyle plugin to your IDE to identify code style issues. We're not using author tags.

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGDecoder.java
```diff
@@ -0,0 +1,143 @@
...
+    private Format myFmt;
+
+    public JPEGDecoder(InputStream theIs) {
+        if (theIs == null)
+            throw new IllegalArgumentException("InputStream must not be null");
+        myIs = theIs;
+        myFmt = Format.ERR;
+
+        try {
+            myBi = ImageIO.read(myIs);
```
msteiger: ImageIO has a fairly broad range of built-in en- and decoders. I believe that this class could be a general AWT-based texture decoder, but restricted by calling code according to some specifics. See #809 for some discussion (maybe only JPG and only for large images).

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGDecoder.java
```diff
@@ -0,0 +1,143 @@
...
+            if (myBi == null) {
+                System.out.println("bi error");
```
msteiger: Please use proper exception handling and useful messages - `throw new IOException()` probably

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGDecoder.java
```diff
@@ -0,0 +1,143 @@
...
+        } catch (java.io.IOException e) {
+            System.out.println("ImageIO error");
+        }
+        myWidth = myBi.getWidth();
+        myHeight = myBi.getHeight();
+    }
+
+    public void decode(ByteBuffer theBuf, Format fmt) {
+        WritableRaster raster = myBi.getRaster();
+        DataBufferByte dbf = (DataBufferByte) raster.getDataBuffer();
```
msteiger: I doubt that this works in general. Can you please verify?

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGDecoder.java
```diff
@@ -0,0 +1,143 @@
...
+    public int getHeight() {
+        return myHeight;
+    }
+
+    public boolean hasAlpha() {
+        return myFmt.equals(Format.RGBA);
+    }
+
+    public boolean hasAlphaChannel() {
+        return myFmt.equals(Format.RGBA);
```
msteiger: Where's the difference to `hasAlpha`?

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGDecoder.java
```diff
@@ -0,0 +1,143 @@
...
+        DataBufferByte dbf = (DataBufferByte) raster.getDataBuffer();
+        byte[] data = dbf.getData();
+        if (fmt.equals(Format.RGBA)) {
+            for (int i = 0; i < data.length; i += 3) {
+                theBuf.put(data, i, 3);
+                theBuf.put((byte) 0xff);
+            }
+        } else
+            theBuf.put(data);
+        // switch (dbf.getDataType()) {
```
msteiger: Please clean your code before submitting it.

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGDecoder.java
```diff
@@ -0,0 +1,143 @@
...
+        return this.myWidth == other.myWidth &&
+            this.myHeight == other.myHeight &&
+            this.myFmt.equals(other.myFmt);
+    }
+
+    @Override public String toString() {
+        final StringBuilder sb = new StringBuilder();
+        sb.append(getClass().getSimpleName());
+        sb.append("@");
+        sb.append(super.toString());
```
msteiger: Please create a self-explanatory text message for `toString`.

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGTextureFormat.java
```diff
@@ -0,0 +1,69 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.file.PathMatcher;
+import java.util.List;
+
+/**
+ * @author Viveret (github.com/viveret)
+ */
+public class JPEGTextureFormat extends AbstractAssetFileFormat<TextureData> {
```
msteiger: This looks like copy and paste from `PNG`. If so, please adapt so they can both use a common method or even class without duplicating the code.

engine/src/main/resources/assets/skins/mainmenu.skin
```diff
@@ -3,7 +3,7 @@
     "elements": {
         "UIScreenLayer": {
             "background-scale-mode": "scale_fill",
-            "background": "menuBackground"
+            "background": "menuBackgroundJPEG"
```
msteiger: I suggest removing the PNG - having the same image twice doesn't buy you anything.
Also, please convert all background images.

msteiger:
Thank you for your contribution, @viveret. I think that there's a little more work to do, but you're on the right track.

engine/src/main/java/org/terasology/rendering/assets/texture/JPEGDecoder.java
viveret: These are for output formats, if you wanted to read in a JPG with 4 components for some odd reason. If this should not be a case then it could be removed.

viveret:
@msteiger I've fixed up a few things and got checkstyle running. Everything checks out.

emanuele3d:
@viveret: I just noticed you converted the sky textures and the vignette textures to JPEG. Those are in-game textures, generally they shouldn't be lossy-compressed. The sky in particular is mapped onto the skysphere and the smallest jpeg artifact might show up the size of a constellation, especially once the color data has gone through the whole rendering pipeline, potentially amplifying unnoticeable issues into very noticeable problems.

Cervator:
@GooeyHub: ok to test

@viveret: thanks! Especially for the quick follow-up fixes :-)

And yeah I think it was just "static" sort of in-game stuff meant for jpg, like the background.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/544/

viveret:
@emanuele3d the vignette doesn't have an alpha channel and details aren't particularly noticeable as it is rendered very transparently, but the skybox could have potential side effects. I suppose using a higher quality factor could resolve those issues.

@Cervator awesome. Perhaps different distributions could include PNG over JPEG or if assets are downloaded on a "as-needed" basis then higher quality vs lower quality could be settings for which version to download.

LadySerenaKitty:
@viveret, I like the idea of having 2 sets of assets - JPG and PNG.  Purrhaps ...

Hey @Cervator, can there be 2 base distribs for us?  One with JPG and the other with PNG?  I know it'd take a bit of reworking the release system, but it might be worth while.

viveret:
@LadySerenaKitty It would probably work better to have a build script convert large PNG resources to JPEG for packaging, or download the type of resource (PNG / JPEG) based on user preferences (size vs graphics).

LadySerenaKitty:
@viveret I was thinking we could offer both as separate download packages.  One would give the user the option to play sooner, with a smaller download size.  Of course, there'd still be the option within the game to upgrade to PNG.

viveret:
@LadySerenaKitty Sounds purrfect :wink:

Cervator:
@LadySerenaKitty @viveret the idea of a smaller + full distro has come up on occasion, yep, although more so for the soundtrack (ChrisVolume1OST + 2) as those files blow everything else out of the water when it comes to shrinking the overall size for a faster download :-)

Some work has been done already to provide patch sets for incrementally upgrading the game through the launcher. Down the road when we have more assets (like tons of creature models and other things) and maybe a way to start with a lower quality soundtrack doing something similar to run the game with a partial download would be very cool to have.

That's a likely large effect at low priority though, so unless somebody gets really eager about making it work it'll probably not happen for a while. Which is fine since in the meantime we can do some of the prep work like supporting JPG in the first place :-)

For now we should probably just switch the couple sky textures back to PNG (how big is the file size difference anyway? Wish there was an easier way to tell here in the PR) so we can merge this and get to the next item of interest. One fix that could be bundled in would be fixing #693 as a mini sun got baked into one (both?) of the sky textures, whoops.

viveret:
@Cervator the sky180.png was crunched to 14% of its original size, and menuBackground.png to 11% of its original size without any noticeable artifacting. The textures I converted reduced the total repo zip size to 78% of its original size.

Cervator:
@viveret Nice! Didn't realize one was a full MB. I see we also have the loadingBackground.png at nearly 4 MB, maybe convert that one to JPG as well? :-)

I still have no idea how important it is to avoid minuscule changes in the sky images. I guess the sky looks normal in-game? I'd have to defer to @emanuele3d as the expert, but if it looks good maybe we're OK

The two water files are big as well, but again sort of central looking rendering assets rather than just backgrounds.

immortius:
I guess there are a few considerations in this space. Certainly nothing wrong with having a JPG format support, but some care in where it is used is necessary, and thought around what problem you are trying to fix.

Firstly, for source formats (formats during development) - jpg compression artifacts can become a problem if a file is repeatedly modified or updated. So if something is going to be jpg then you either want the source around in case you do further edits to it, or you just throw it away and replace it.

For distribution - keep in mind that jars can be compressed, or a module can be a zip file. While not as efficient as image specific compression this will save some space with little fuss. The idea of having a standard distribution and a high quality distribution is also possible - either by having a module providing HD overrides for assets or just having two versions of the same module differing only by quality of assets. The former formalises things, but does have issues in a net game - all clients will be made to use the same modules.

As an alternative to jpg, S3 compressed texture support would be nice, especially as hardware directly handles it so it doesn't have to be decompressed in memory at runtime. See: https://www.opengl.org/wiki/S3_Texture_Compression. Potentially conversion from png/jpg to an s3 compressed form could be done at built time based on settings in the .info files.

msteiger:
I think that runtime texture compression (through S3) and storage compression (jpg) could go hand in hand.

For now, I'd rather not change the in-game textures (vignette and sky) and convert only the menu background image and [loadingBackground.png](engine/src/main/resources/assets/textures/loadingBackground.png).

The code changes look good to me now. Thanks @viveret

Cervator:
@viveret Each pushed commit to the branch this PR is sourced from will gracefully update the PR so you should be fine if you want to just re-commit the PNGs and delete the JPGs :-)

If you feel daring try to use Git Rebase instead to entirely undo those changes so they won't appear in history. That can be tricky if you're not super comfortable with Git though. Another simple alternative is just re-apply the desired changes to a new branch and open a new PR :-)

viveret:
@Cervator I converted loadingBackround to a jpeg and reverted the in-game textures via checkout from the develop branch.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/548/

Cervator:
Thank you so much for hanging in there through all the review @viveret, it is really appreciated :-)

Everything looks correct now, yet oddly when I try to test it I get black backgrounds. I wonder if something is wrong in my workspace. Works OK with the old PNGs. It is like the engine isn't even recognizing JPEGs as something to scan for, no activity if I try to set breakpoints.

Can anybody else try it?

viveret:
@Cervator there's some assumptions I made when allocating data, which is that I assumed that texture data can be stored as just RGB. The PNG Texture Format allocates for RGBA, and I don't think anything allows for 3 channel texture data.

msteiger:
I believe we (whoever that is) should squash the commits to avoid an unnecessary increase of the git repository.
I can try it out tomorrow night if that's soon enough.

skaldarnar:
@viveret @Cervator I can confirm the black background (in the menu) and black artefacts in the loading screen.

Cervator:
> @viveret @Cervator I can confirm the black background (in the menu) and black artefacts in the loading screen.

:(

Did too much end up getting reverted after all? Or is there something wrong with the image (re: the RGB comment above) ? It just doesn't seem like the engine sees the files?

As for git rebase then yeah ideally we keep commits brief and to the point. It may be an idea to clean this up (after we're sure it works again). Not the end of the world if nobody gets it done, just 4 commits this time, seen more than that in the past to fix a typo, which is when it gets a bit silly :-)

viveret:
In a bit I'll try covertiny the rgb JPEG data to rgba when it passes it to TextureData

viveret:
@Cervator @skaldarnar
I confirmed, on line 57 of TextureData.java:
Texture data size incorrect, must be a set of RGBA values for each pixel (width * height)

I can fix this easily when I get home

skaldarnar:
For squashing commints, `git rebase -i HEAD~4` allows you to modify the last 4 commits. Simply use `pick` on the first one and `squash` the other ones. Afterwards you can also chose a new commit message (all commits msgs together will be the default).

Cervator:
> I can fix this easily when I get home

Awesome! :-)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/552/

viveret:
I think there is some issue with how textures are decided to be loaded in, which is easy to do as currently all textures are loaded through PNG. I'll continue working on this.

Cervator:
> EDIT: I say this as I don't believe JPEGTextureFormat is ever called.

Yep that's what looked puzzling to me as well. It is like the new format isn't completely hooked in somewhere, or we just previously assumed `.png` and didn't bother considering other options.

viveret:
I hooked it into lwjgl asset setup and it caused problems so I'll still try and figure out what's going on.

engine/src/main/java/org/terasology/rendering/assets/texture/AWTTextureFormat.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.InputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.file.PathMatcher;
+import java.util.List;
+import javax.imageio.ImageIO;
+
+/**
+ */
+public class AWTTextureFormat extends AbstractAssetFileFormat<TextureData> {
```
rzats: This needs to be annotated with `@RegisterAssetFileFormat` in order to actually hook up to the asset system - see [gestalt-asset-core documentation](https://github.com/MovingBlocks/gestalt/wiki/Gestalt%20Asset%20Core%20Quick%20Start#asset-formats)

viveret:
I got e-mailed about a link to [an annotation](https://github.com/MovingBlocks/gestalt/wiki/Gestalt%20Asset%20Core%20Quick%20Start#asset-formats) which says I should use `@RegisterAssetFileFormat`. I inserted it and now it fails to compile as it cannot find that symbol. PNGTextureFormat also does not have this annotation.

rzats:
@viveret: here's a method of hooking up the asset format that managed to trigger a breakpoint in `AWTTextureFormat.load()` for me.

```java
// ...
import org.terasology.assets.module.annotations.RegisterAssetFileFormat;

@RegisterAssetFileFormat
public class AWTTextureFormat extends AbstractAssetFileFormat<TextureData> {

    private Texture.FilterMode defaultFilterMode;
    private PathMatcher pathMatcher;

    public AWTTextureFormat() {
        super("jpeg", "jpg", "bmp", "gif");
        this.defaultFilterMode = Texture.FilterMode.NEAREST;
        this.pathMatcher = path -> path.getName(2).toString().equals("textures");
    }
    // ...
```

On the contrary, the one-liner `registerCoreFormat()` setup in HeadlessGraphics.java does not appear to work. This is really odd and should definitely be investigated as the two methods are effectively identical.

viveret:
I was following PNGTextureData so this might be a bigger issue.

viveret:
@Cervator @msteiger @emanuele3d @rzats @skaldarnar Finally fixed these issues, turns out BufferedImage's internal format is more than what I thought it was!
2016-03-31-200411_3280x1200_scrot

Cervator:
@viveret works! Well done, again thanks for hanging in there through the review process :D

It looks like you may only have squashed one or two commits though? There are still 9 total. Not a big deal, will merge sometime tomorrow in case you want to give that piece a try, but no worries if not :-)

viveret:
@Cervator bleh, I'll make try figuring it out on my own repos. Glad I got it through and contributed!

viveret:
@Cervator there we go, I squished them commits

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/556/

skaldarnar:
I can confirm, all looks good (tested on Linux) and only one commit. Merging.

Cervator:
@viveret - one last time thank you so much for your perseverance! You're now tagged and credited as appropriate all over :-)

Issue #2268, opened 2016-03-30
Support for disabled UI elements


Cervator:
Currently to the best of my knowledge we can't visually disable UI elements (please correct me if I'm wrong!), such as making a checkbox greyed out while being stuck in a particular setting.

This would help out #2246 as volumetric fog is broken but has a valid toggle in the graphic settings now. We should have the option to disable such an option so we can keep it, but leave it passive until the actual feature finally works again.

Probably not super hard for somebody familiar with UI work or interested in learning. May be relevant to @rzats' interests in doing NUI work.

msteiger:
Sounds reasonable to me. A `disabled` family mode already exists and has been implemented for buttons. Adding this to checkboxes and drop-down lists would probably cover most ground.

Cervator:
Closing as completed by @2274, could be further enhanced by #2277 :-)

PR #2274, opened 2016-04-05
[OK to merge] Support for disabled UI elements

rzats:
### Contains

Implementation for #2268. Adds a disabled state for UICheckbox, UIDropdown and UIDropdownScrollable - this should cover all of NUI's interactable widgets. _(UIButton can already be disabled via `"enabled": false`, while UIText can be set to readonly via `"readOnly": true`)_

### How to test

Interact with the Volumetric Fog checkbox (as per #2246).

### Outstanding before merging

No outstanding issues.

### Future improvements

- [ ] Add a skin for a disabled UIDropdown[Scrollable]. Should be fairly straightforward - just change the selected option's text color, as with the button.
- [ ] Add a skin for a disabled UICheckbox. Slightly harder to implement as it will require creating greyed-out variants of [checkbox.png](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/textures/ui/checkbox.png) and [checkboxChecked.png](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/textures/ui/checkboxChecked.png).
- [ ] Create a sample *.ui file containing a bunch of disabled items.

PR #2275, opened 2016-04-05
[Work in Progress] Yet another - Add support for disabled UI elements

tdgunes:
### Contains

Implementation for #2268 (Disables volumetric fog), adds support for disabled UILabel and UICheckbox. I didn't noticed that @rzats also working on the same issue. Here is my version of the implementation. It might be useful to merge these too, somehow. 

### How to test

Checking volumetric fog setting in video setting videos.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/AbstractWidget.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/AbstractWidget.java
@@ -43,6 +43,9 @@ public abstract class AbstractWidget implements UIWidget {
     private Binding<Boolean> visible = new DefaultBinding<>(true);
 
     @LayoutConfig
+    private Binding<Boolean> enabled = new DefaultBinding<>(true);
+
+    @LayoutConfig
     private Binding<UIWidget> tooltip = new DefaultBinding<>();
 
     @LayoutConfig
@@ -139,6 +142,14 @@ public abstract class AbstractWidget implements UIWidget {
 
     public void setVisible(boolean visible) {
         this.visible.set(visible);
+    }
+
+    public void setEnabled(boolean enabled) {
+        this.enabled.set(enabled);
+    }
+
+    public boolean isEnabled() {
+        return this.enabled.get();
     }
 
     public void bindVisible(Binding<Boolean> bind) {
--- a/engine/src/main/java/org/terasology/rendering/nui/Canvas.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/Canvas.java
@@ -145,11 +145,19 @@ public interface Canvas {
     void drawWidget(UIWidget element, Rect2i region);
 
     /**
-     * Draws test, using the current style.
+     * Draws text, using the current style.
      *
      * @param text
      */
     void drawText(String text);
+
+
+    /**
+     * Draws text, using the current style.
+     *
+     * @param text
+     */
+    void drawText(String text, boolean enabled);
 
     /**
      * Draws text to the given region, using the current style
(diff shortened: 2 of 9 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/568/

msteiger:
Is that related to #2274 ?

rzats:
The problem with this implementation is that the `drawText()` argument changes are very API-breaking. In fact, they **will** break [Journal](https://github.com/Terasology/Journal/blob/master/src/main/java/org/terasology/journal/part/TextJournalPart.java#L57), [WoodAndStone](https://github.com/Terasology/WoodAndStone/blob/master/src/main/java/org/terasology/crafting/ui/workstation/RecipeCategoryWidget.java#L63), [Tasks](https://github.com/Terasology/Tasks/blob/develop/src/main/java/org/terasology/tasks/gui/QuestRenderer.java#L76) and [NUI itself](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/itemRendering/StringTextIconRenderer.java#L72) (that last one is what causes the build to fail - not sure why you haven't spotted that locally).

They're also mostly unnecessary, since this is possible:

```java
// default.skin
"UILabel": {
  "modes": {
    "disabled": {
      "text-color": "FF808080"
    }
  }
}
```

The idea of `enabled` being an `AbstractWidget` property (to avoid repetition) is quite nice - although it could be applied to a more specific base class (e.g. "`InteractiveAbstractWidget`") to avoid using it in non-interactive widgets.

That being said, sorry for not explicitly expressing my interest in the issue sooner :(

Cervator:
Should we merge the other PR first then see about enhancing it further with the good parts from here? Like the disabled volumetric fog and abstract goodies :-)

tdgunes:
@ratzs, I was planning to put a default text color for disabled UILabel. In addition to that, I thought it would be nice to have every widget to be enabled or disabled. It's interesting that, drawText() caused tests to fail. I will look into more closely. I think I missed to restore default implementation.  

Also thanks for mentioning _InteractiveAbstractWidget_.

Cervator:
@tdgunes I went ahead and merged the other PR for the basic support, but there were several outstanding remaining pieces and some of the stuff in this could be useful there. Some copy pasta:

- [ ] Add a skin for a disabled UIDropdown[Scrollable]. Should be fairly straightforward - just change the selected option's text color, as with the button.
- [ ] Add a skin for a disabled UICheckbox. Slightly harder to implement as it will require creating greyed-out variants of [checkbox.png](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/textures/ui/checkbox.png) and [checkboxChecked.png](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/textures/ui/checkboxChecked.png).
- [ ] Create a sample *.ui file containing a bunch of disabled items (@rzats may be working on that in the Sample module and/or maybe `showScreen migTestScreen` would be useful)

I submitted #2277 for the visual distinction. Default text color tweak for various widgets and easy reusability would be a plus! :-)

You *may* want to close this PR and open a new one on a refreshed branch (or really you could re-submit an update in the same branch just force pushed / rebased, your pick)

rzats:
For the record, I'm indeed planning to look into UI samples - but everything else is OK to improve on! :)

Also @tdgunes: I've looked into the Jenkins logs and it looks like the reason that error didn't show up locally is that [your fork](https://github.com/tdgunes/Terasology/tree/new-options) doesn't contain the relevant file. Pulling from upstream and pushing to the fork every now and then is generally a good idea, especially if you're planning on doing some refactoring.

tdgunes:
Thanks for looking into that @ratzs. I will try to keep that in mind.

tdgunes:
Closing this one, new improvements will be in #2280.

Issue #2277, opened 2016-04-06
Visually distinct UI elements for enabled/disabled (and support for more disabled types)


Cervator:
*Edit: Started as a plain follow-up to #2274, after #2280 I'm hijacking it for a bit more detail, see new todo list below*

We can disable checkboxes now, huzzah! Yet they look the same. We should make that (and similar) situation appear more unique / grayed out :-)

A few more details in #2274 and maybe related to #2275 as well.

While in the main menu you can still access the game console and `showScreen migTestScreen` will show a screen with a bunch of test elements on it (close with `ESC`) - good place to test all the elements both for functionality, disable/enable, and visual distinction.

- [x] `UIButton` - done!
- [ ] `UISlider` - no disable support (so naturally no look either)
- [ ] `UIDoubleSlider` - no disable support (so naturally no look either)
- [ ] `UIText` - can be disabled, but doesn't have a unique look yet
- [ ] `UIDropdown` - can be disabled, but doesn't have a unique look yet
- [ ] `UIDropdownScrollable` - can be disabled, but doesn't have a unique look yet
- [x] `UICheckbox` - kinda done - very faint difference in appearance? Could be more distinct?
- [ ] `UIList` - no disable support, buttons have visual distinction so may just need the support
- [ ] `UIBox` - N/A ? Not sure why we would disable one of these or have it show differently. Maybe if you're grouping elements and want to show the entire group as disabled (like a part of a tech tree you haven't gotten to yet) ?
- [ ] `UIScrollbar` - can be disabled, but not sure if it shows well enough on the test screen to see a difference in appearance?
- [x] `UISpace` - included for completeness, but pretty sure this one is N/A
- [ ] `UILabel` - *MISSING* from the test screen, but otherwise supported and distinct (#2280) - just need to add it to the test screen for completeness

Any other elements? Please comment if so, we might as well catch em all!

Cervator:
Bump - rewrote the initial description some with a nice todo list based on `showScreen migTestScreen`

Since we surprisingly have 3-4 different contributors suddenly working on UI topics please post a note of interest here if wanting to do it so we don't end up with overlap :-)

tdgunes:
I can send PRs for this issue. Would be great to if you can assign this one to me.

Cervator:
Hehe done! Via bot noticing your submitted PR. I love it when automation works :D

(granted, had I been awake 5 hours ago I might have been able to beat the bot to the assignment step)

[...]

PR #2280, opened 2016-04-07
Add new checkbox textures, UILabel modes, enabled binding, make volumetric fog label gray

tdgunes:
### Contains

A partial fix for #2277 (previously #2274 and #2275)

### How to test

Check Volumetric Fog setting's checkbox and label, both their color should be more greyish.

screen shot 2016-04-07 at 23 25 53


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/575/

Cervator:
Checks out! 

Hah, hah, hah ;-)

I looked at it via `showScreen migTestScreen` as well, although in this case it is easier to compare on the Video Settings screen. Still pretty hard to see a difference between disabled and enabled - could it be clearer if the disabled background was slightly more light gray? Or even had subtle lighter gray marks of some sort, maybe even horizontal + vertical lines to put the checkbox in jail, so to say?

If you play with it and find an improvement feel free to just shove it into any more meaty PR that comes along sometime :-)

Thanks!

Cervator:
Slight addendum just for change notes: I took inspiration from this (read: shamelessly copy pasted) to also disable the Oculus Rift checkbox since that doesn't work at present either :-)

emanuele3d:
Odd about oculus. In one of my last PRs i could see the double image on
screen. It doesn't achieve stereoscopic vision or does it currently crash
the application?

Cervator:
@emanuele3d the current Oculus support was written against the DK1 with a waaay outdated SDK version. The option may work to where it lets a double image render to a normal screen, but I can't imagine it'll actually work with a Rift like you'd expect (turning your head, etc). Plus I believe the latest SDK version forces the use of direct HMD mode while our original version was probably for extended desktop (letting you see the double image in the first place without mirroring)

Like the volumetric fog option having it available in-game does enable some sort of feature for developing further, but it isn't end-user ready so I'd rather not get the hope of players up just yet :-) For testing 3d wizard PRs we might want to un-disable manually briefly if testing out the basics make sense.

emanuele3d:
@Cervator: ah, I understand now, thank you. I wonder if we should remove all the oculus-related code then. It would make maintenance and further development easier.

Cervator:
@emanuele3d maybe. I don't know enough about what we have and how many changes would be needed. Since some of the GSOC proposals involve VR I'd wait till after we know what topics get picked though.

Ultimately we should probably aim for an open VR standard, I don't know how Oculus specific what we have is. Maybe being able to render the double image is a worthwhile piece of code for future reuse. Unsure :-)

PR #2281, opened 2016-04-08
Reduce overbrightness in various parts of the system

Josharias:
### Contains

This fixes various parts of the lighting system.  Addressing issues with dropping multiple torches, oddness with specular light, and more usable attenuation values.  Screenshots to follow...

This does introduce something that needs addressing.  When a torch is held or dropped in the world,  it does not light itself.  This is probably an already existing issue present,  but made more obvious when I removed the light fading in and out when switching between torches and non torches.  Which brings me to something else that needs addressing,  items that are held need to be better shown in a multiplayer setting.  This will allow us to avoid putting lights on the character altogether.

### How to test

Drop lots of torches.  Hold torches.  Watch the sun/moon effect on the world.

- [x] ~~See if something easy can be done with dimming when walking in bright daylight.~~  I dont know how to fix this without causing more overbrightness.  Who lights a torch in daylight? ;)
- [x] Reduce overall specular light as everything is definitely too shiney

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
@@ -15,13 +15,14 @@
  */
 package org.terasology.rendering.logic;
 
-import org.terasology.reflection.metadata.FieldMetadata;
 import org.terasology.entitySystem.Component;
 import org.terasology.math.geom.Vector3f;
 import org.terasology.network.Replicate;
 import org.terasology.network.ReplicationCheck;
+import org.terasology.reflection.metadata.FieldMetadata;
 
 /**
+ * Add this component to an entity for it to transmit light from its location.  By default the component is configured to act similarly to a placed torch block.
  */
 // TODO: Split into multiple components? Point, Directional?
 public final class LightComponent implements Component, ReplicationCheck {
@@ -43,14 +44,29 @@ public final class LightComponent implements Component, ReplicationCheck {
     @Replicate
     public float lightAmbientIntensity = 1.0f;
 
+    /**
+     * This helps control how focused the specular light is. A smaller number will make a wider cone of light. A larger number will make a narrower cone of light.
+     */
     @Replicate
-    public float lightSpecularPower = 1.0f;
+    public float lightSpecularPower = 100.0f;
+    /**
+     * Light attenuation range used in the calculation of how light fades from the light source as it gets farther away.  It is use in the following calculation:
+     * <p>
+     * attenuation = 1 / (lightDist/lightAttenuationRange + 1)^2
+     * <p>
+     * Where lightDist is how far the point in the world is from the light source.
+     */
     @Replicate
-    public float lightAttenuationRange = 16.0f;
+    public float lightAttenuationRange = 10.0f;
+    /**
+     * After light travels the lightAttenuationRange, linearly fade to 0 light over this falloff distance.
+     */
     @Replicate
     public float lightAttenuationFalloff = 1.25f;
 
-    // The rendering distance for light components (0.0f == Always render the light)
+    /**
+     * The rendering distance for light components (0.0f == Always render the light)
+     */
     @Replicate
     public float lightRenderingDistance;
 
--- a/engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
@@ -40,7 +40,7 @@ public final class LightComponent implements Component, ReplicationCheck {
     @Replicate
     public float lightDiffuseIntensity = 1.0f;
     @Replicate
-    public float lightSpecularIntensity = 0.1f;
+    public float lightSpecularIntensity = 0.03f;
     @Replicate
     public float lightAmbientIntensity = 1.0f;
 
@@ -48,7 +48,7 @@ public final class LightComponent implements Component, ReplicationCheck {
      * This helps control how focused the specular light is. A smaller number will make a wider cone of light. A larger number will make a narrower cone of light.
      */
     @Replicate
-    public float lightSpecularPower = 100.0f;
+    public float lightSpecularPower = 80.0f;
     /**
      * Light attenuation range used in the calculation of how light fades from the light source as it gets farther away.  It is use in the following calculation:
      * <p>
--- a/engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
+++ b/engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
@@ -26,9 +26,36 @@ import java.nio.IntBuffer;
 import static org.lwjgl.opengl.EXTFramebufferObject.GL_COLOR_ATTACHMENT0_EXT;
 import static org.lwjgl.opengl.EXTFramebufferObject.GL_FRAMEBUFFER_EXT;
 import static org.lwjgl.opengl.EXTFramebufferObject.glBindFramebufferEXT;
-import static org.lwjgl.opengl.GL11.*;
+import static org.lwjgl.opengl.GL11.GL_ALWAYS;
+import static org.lwjgl.opengl.GL11.GL_BACK;
+import static org.lwjgl.opengl.GL11.GL_BLEND;
+import static org.lwjgl.opengl.GL11.GL_COLOR_BUFFER_BIT;
 import static org.lwjgl.opengl.GL11.GL_CULL_FACE;
+import static org.lwjgl.opengl.GL11.GL_DECR;
+import static org.lwjgl.opengl.GL11.GL_DEPTH_BUFFER_BIT;
+import static org.lwjgl.opengl.GL11.GL_DEPTH_TEST;
+import static org.lwjgl.opengl.GL11.GL_FILL;
+import static org.lwjgl.opengl.GL11.GL_FRONT;
+import static org.lwjgl.opengl.GL11.GL_FRONT_AND_BACK;
+import static org.lwjgl.opengl.GL11.GL_INCR;
+import static org.lwjgl.opengl.GL11.GL_KEEP;
 import static org.lwjgl.opengl.GL11.GL_LEQUAL;
+import static org.lwjgl.opengl.GL11.GL_LINE;
+import static org.lwjgl.opengl.GL11.GL_NOTEQUAL;
+import static org.lwjgl.opengl.GL11.GL_ONE;
+import static org.lwjgl.opengl.GL11.GL_ONE_MINUS_SRC_ALPHA;
+import static org.lwjgl.opengl.GL11.GL_ONE_MINUS_SRC_COLOR;
+import static org.lwjgl.opengl.GL11.GL_SRC_ALPHA;
+import static org.lwjgl.opengl.GL11.GL_STENCIL_BUFFER_BIT;
+import static org.lwjgl.opengl.GL11.GL_STENCIL_TEST;
+import static org.lwjgl.opengl.GL11.glBlendFunc;
+import static org.lwjgl.opengl.GL11.glClear;
+import static org.lwjgl.opengl.GL11.glCullFace;
+import static org.lwjgl.opengl.GL11.glDepthMask;
+import static org.lwjgl.opengl.GL11.glDisable;
+import static org.lwjgl.opengl.GL11.glEnable;
+import static org.lwjgl.opengl.GL11.glStencilFunc;
+import static org.lwjgl.opengl.GL11.glViewport;
 import static org.lwjgl.opengl.GL20.glStencilOpSeparate;
 
 /**
@@ -318,7 +345,7 @@ public class GraphicState {
         glDisable(GL_DEPTH_TEST);
 
         glEnable(GL_BLEND);
-        glBlendFunc(GL_ONE, GL_ONE);
+        glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_COLOR);
 
         glEnable(GL_CULL_FACE);
         glCullFace(GL_FRONT);
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -15,7 +15,6 @@
  */
 package org.terasology.rendering.world;
 
-import org.terasology.utilities.Assets;
 import org.terasology.config.Config;
 import org.terasology.config.RenderingConfig;
 import org.terasology.config.RenderingDebugConfig;
@@ -45,12 +44,13 @@ import org.terasology.rendering.cameras.OculusStereoCamera;
 import org.terasology.rendering.cameras.OrthographicCamera;
 import org.terasology.rendering.cameras.PerspectiveCamera;
 import org.terasology.rendering.logic.LightComponent;
-import org.terasology.rendering.opengl.GraphicState;
 import org.terasology.rendering.opengl.FrameBuffersManager;
+import org.terasology.rendering.opengl.GraphicState;
 import org.terasology.rendering.opengl.PostProcessor;
 import org.terasology.rendering.primitives.ChunkMesh;
 import org.terasology.rendering.primitives.LightGeometryHelper;
 import org.terasology.rendering.world.viewDistance.ViewDistance;
+import org.terasology.utilities.Assets;
 import org.terasology.world.WorldProvider;
 import org.terasology.world.chunks.ChunkConstants;
 import org.terasology.world.chunks.ChunkProvider;
@@ -169,11 +169,8 @@ public final class WorldRendererImpl implements WorldRenderer {
     // TODO: one day the main light (sun/moon) should be just another light in the scene.
     private void initMainDirectionalLight() {
         mainDirectionalLight.lightType = LightComponent.LightType.DIRECTIONAL;
-        mainDirectionalLight.lightColorAmbient = new Vector3f(1.0f, 1.0f, 1.0f);
-        mainDirectionalLight.lightColorDiffuse = new Vector3f(1.0f, 1.0f, 1.0f);
-        mainDirectionalLight.lightAmbientIntensity = 1.0f;
-        mainDirectionalLight.lightDiffuseIntensity = 2.0f;
-        mainDirectionalLight.lightSpecularIntensity = 0.0f;
+        mainDirectionalLight.lightAmbientIntensity = 0.75f;
+        mainDirectionalLight.lightDiffuseIntensity = 0.75f;
     }
 
     private void initRenderingSupport() {
@@ -576,7 +573,7 @@ public final class WorldRendererImpl implements WorldRenderer {
 
         if (lightComponent.lightType == LightComponent.LightType.POINT) {
             if (!geometryOnly) {
-                program.setFloat4("lightExtendedProperties", lightComponent.lightAttenuationRange * 0.975f, lightComponent.lightAttenuationFalloff, 0.0f, 0.0f, true);
+                program.setFloat4("lightExtendedProperties", lightComponent.lightAttenuationRange, lightComponent.lightAttenuationFalloff, 0.0f, 0.0f, true);
             }
 
             LightGeometryHelper.renderSphereGeometry();
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -171,6 +171,8 @@ public final class WorldRendererImpl implements WorldRenderer {
         mainDirectionalLight.lightType = LightComponent.LightType.DIRECTIONAL;
         mainDirectionalLight.lightAmbientIntensity = 0.75f;
         mainDirectionalLight.lightDiffuseIntensity = 0.75f;
+        mainDirectionalLight.lightSpecularIntensity = 0.02f;
+        mainDirectionalLight.lightSpecularPower = 100f;
     }
 
     private void initRenderingSupport() {
--- a/engine/src/main/resources/assets/shaders/lightGeometryPass_frag.glsl
+++ b/engine/src/main/resources/assets/shaders/lightGeometryPass_frag.glsl
@@ -89,8 +89,11 @@ void main() {
 
 #if defined (CLOUD_SHADOWS) && !defined (VOLUMETRIC_LIGHTING)
         // TODO: Add shader parameters for this...
-        float cloudOcclusion = clamp(texture2D(texSceneClouds, (worldPosition.xz + cameraPosition.xz) * 0.005 + timeToTick(time, 0.004)).r * 4,0,.4);
+        // Get the preconfigured value from the randomized texture, sampling a value from it to determine how much cloud shadow there will be.
+        // Clamp the value so that clouds do not turn the surface black.
+        float cloudOcclusion = clamp(texture2D(texSceneClouds, (worldPosition.xz + cameraPosition.xz) * 0.005 + timeToTick(time, 0.002)).r * 10,0.6,1);
 
+        // Combine the cloud shadow with the dynamic shadows
         shadowTerm *= rescaleRange(cloudOcclusion,0,1,0,shadowTerm);
 #endif
    }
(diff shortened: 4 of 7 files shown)


engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
```diff
@@ -43,14 +44,29 @@
...
+     * Where lightDist is how far the point in the world is from the light source.
+     */
     @Replicate
-    public float lightAttenuationRange = 16.0f;
+    public float lightAttenuationRange = 10.0f;
+    /**
+     * After light travels the lightAttenuationRange, linearly fade to 0 light over this falloff distance.
+     */
     @Replicate
     public float lightAttenuationFalloff = 1.25f;
```
Josharias: This now has new meaning.  Would anyone prefer a different name for this concept?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/578/

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -169,11 +169,8 @@ public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
     // TODO: one day the main light (sun/moon) should be just another light in the scene.
     private void initMainDirectionalLight() {
         mainDirectionalLight.lightType = LightComponent.LightType.DIRECTIONAL;
-        mainDirectionalLight.lightColorAmbient = new Vector3f(1.0f, 1.0f, 1.0f);
-        mainDirectionalLight.lightColorDiffuse = new Vector3f(1.0f, 1.0f, 1.0f);
-        mainDirectionalLight.lightAmbientIntensity = 1.0f;
-        mainDirectionalLight.lightDiffuseIntensity = 2.0f;
-        mainDirectionalLight.lightSpecularIntensity = 0.0f;
+        mainDirectionalLight.lightAmbientIntensity = 0.75f;
+        mainDirectionalLight.lightDiffuseIntensity = 0.75f;
```
Josharias: I believe these values were originally bumped up in order that the bug in the shader made sunlight a normal level with clouds.  This is the before and after without the clouds.

Before:
terasology-160406170158-1280x720

After:
terasology-160408115448-1280x720

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -576,7 +573,7 @@ private boolean renderLightComponent(LightComponent lightComponent, Vector3f lig
 
         if (lightComponent.lightType == LightComponent.LightType.POINT) {
             if (!geometryOnly) {
-                program.setFloat4("lightExtendedProperties", lightComponent.lightAttenuationRange * 0.975f, lightComponent.lightAttenuationFalloff, 0.0f, 0.0f, true);
```
Josharias: There is no reasoning to scale this back an arbitrary amount.

engine/src/main/resources/assets/shaders/lightGeometryPass_frag.glsl
```diff
@@ -89,8 +89,11 @@ void main() {
 
 #if defined (CLOUD_SHADOWS) && !defined (VOLUMETRIC_LIGHTING)
         // TODO: Add shader parameters for this...
-        float cloudOcclusion = clamp(texture2D(texSceneClouds, (worldPosition.xz + cameraPosition.xz) * 0.005 + timeToTick(time, 0.004)).r * 4,0,.4);
+        // Get the preconfigured value from the randomized texture, sampling a value from it to determine how much cloud shadow there will be.
+        // Clamp the value so that clouds do not turn the surface black.
+        float cloudOcclusion = clamp(texture2D(texSceneClouds, (worldPosition.xz + cameraPosition.xz) * 0.005 + timeToTick(time, 0.002)).r * 10,0.6,1);
```
Josharias: 2 things here, the first is to slow down the clouds so they dont zip past quite as fast (the 0.002 value).  The second is to fix a bug where the non coud areas were being shadowed because we were clamping the opposite way.  You will notice a stark difference in the sand brightness in the "before" screenshots.

Before with clouds:
terasology-160406170140-1280x720

Before without clouds:
terasology-160406170158-1280x720

After with clouds:
terasology-160408115515-1280x720

After without clouds:
terasology-160408115448-1280x720

engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
```diff
@@ -40,14 +40,8 @@ float calcLambLight(vec3 normal, vec3 lightVec) {
...
 
-float calcSpecLight(vec3 normal, vec3 lightVec, vec3 eyeVec, float exp) {
-    vec3 halfWay = normalize(eyeVec+lightVec);
-    return pow(clamp(dot(halfWay, normal), 0.0, 1.0), exp);
-}
-
 float calcSpecLightNormalized(vec3 normal, vec3 lightVec, vec3 eyeVec, float exp) {
-    vec3 halfWay = normalize(eyeVec+lightVec);
-    return ((exp + 8.0) / PI_TIMES_8) * pow(clamp(dot(halfWay, normal), 0.0, 1.0), exp);
+    return pow(max(0.0, dot(-eyeVec, reflect(lightVec, normal))), exp);
```
Josharias: This change makes a less spreading specular light.  Although the original method had some promise,  it just didnt look correct as the cone of light coming off did not match the reflection on the wall.    This formula is based off http://www.tomdalling.com/blog/modern-opengl/07-more-lighting-ambient-specular-attenuation-gamma/

Before:
terasology-160408105243-1280x720

After:
terasology-160408105228-1280x720

engine/src/main/resources/assets/shaders/lightGeometryPass_frag.glsl
```diff
@@ -143,21 +147,19 @@ void main() {
...
+    float denom = lightDist/lightAttenuationRange + 1;
     float attenuation = 1.0/(denom*denom);
-    //float attenuation = clamp (1.0 - ((lightDist * lightDist) / (lightAttenuationRange * lightAttenuationRange)), 0.0, 1.0);
-    //float attenuation = 1.0 - (pow (lightDist, lightAttenuationFalloff) / lightAttenuationRange);
 
-    attenuation = (attenuation - cutoff)/(1.0-cutoff);
+    // Force the light to gradually come to a complete stop at the attenuation range + falloff distance.
+    float lightDistPastRange = max(lightDist - lightAttenuationRange, 0.0);
+    float falloffTerm = 1.0 - min(lightDistPastRange / lightAttenuationFalloff, 1.0);
+    attenuation *= falloffTerm;
```
Josharias: This falloff mechanism allows simulating darkness game mechanics by cutting off the light after a certain distance.  One can still have light behave as if in a vacuum by setting the falloff value really high from the LightComponent.

Before:
terasology-160408120909-1280x720

After with a falloff value of 1.25:
terasology-160408120925-1280x720

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
```diff
@@ -318,7 +345,7 @@ public void preRenderSetupLightGeometry() {
         glDisable(GL_DEPTH_TEST);
 
         glEnable(GL_BLEND);
-        glBlendFunc(GL_ONE, GL_ONE);
+        glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_COLOR);
```
Josharias: This is what helps multiple dropped lights not to cause overbrightness.

Before (20 torches):
terasology-160408094050-1280x720

After:
terasology-160408114948-1280x720

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
```diff
@@ -43,14 +44,29 @@
     @Replicate
     public float lightAmbientIntensity = 1.0f;
 
+    /**
+     * This helps control how focused the specular light is. A smaller number will make a wider cone of light. A larger number will make a narrower cone of light.
+     */
     @Replicate
-    public float lightSpecularPower = 1.0f;
```
Josharias: Previously,  specular power of 1 made specular light do this:
terasology-160408105317-1280x720

With the change of the specular glsl shared function,  and with bumping up this value by default,  we now get specular light more like this:
terasology-160408104659-1280x720

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
```diff
@@ -43,14 +44,29 @@
...
+    /**
+     * Light attenuation range used in the calculation of how light fades from the light source as it gets farther away.  It is use in the following calculation:
+     * <p>
+     * attenuation = 1 / (lightDist/lightAttenuationRange + 1)^2
+     * <p>
+     * Where lightDist is how far the point in the world is from the light source.
+     */
     @Replicate
-    public float lightAttenuationRange = 16.0f;
+    public float lightAttenuationRange = 10.0f;
```
Josharias: This range (and with the a 1.25 falloff value), will make a point light source act in roughly the same radius as a placed light.

Before (placed torch):
terasology-160406170308-1280x720

Before (held torch):
terasology-160406170314-1280x720

After (placed torch):
terasology-160408093441-1280x720

After (held torch):
terasology-160408093448-1280x720

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -169,11 +169,8 @@ public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
     // TODO: one day the main light (sun/moon) should be just another light in the scene.
     private void initMainDirectionalLight() {
         mainDirectionalLight.lightType = LightComponent.LightType.DIRECTIONAL;
-        mainDirectionalLight.lightColorAmbient = new Vector3f(1.0f, 1.0f, 1.0f);
-        mainDirectionalLight.lightColorDiffuse = new Vector3f(1.0f, 1.0f, 1.0f);
-        mainDirectionalLight.lightAmbientIntensity = 1.0f;
-        mainDirectionalLight.lightDiffuseIntensity = 2.0f;
-        mainDirectionalLight.lightSpecularIntensity = 0.0f;
```
Josharias: With the default value of 0.1 and with the fixes,  we can now get specular light coming off the sun/moon.  It is neat because the specular light comes in line with the water specular light. :)

After (specular intensity of 0.1):
terasology-160408112602-1280x720

After (specular intensity of 0.5 so that you can see what to look for):
terasology-160408112615-1280x720

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
```diff
@@ -15,13 +15,14 @@
...
 import org.terasology.math.geom.Vector3f;
 import org.terasology.network.Replicate;
 import org.terasology.network.ReplicationCheck;
+import org.terasology.reflection.metadata.FieldMetadata;
 
 /**
+ * Add this component to an entity for it to transmit light from its location.  By default the component is configured to act similarly to a placed torch block.
  */
 // TODO: Split into multiple components? Point, Directional?
 public final class LightComponent implements Component, ReplicationCheck {
```
Josharias: Probably a good discussion we should all have is how "shiny" we want the world to be.  We do not have the option to set shinyness on a block level,  only at the light source.  The default value of 0.1 seems a little much for torch light when you drop it close to a wall.

emanuele3d: Indeed it looks almost mirror-like in the handling of the reflected light. Personally I feel that the light source shouldn't have any shininess parameter at all: it should have an (RGB) intensity and a way for that intensity to falloff with distance, but it should be up to a surface's shader to make the surface look shiny or not.

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
emanuele3d: I think both these images hint to the fact that more work needs to be done on the lighting model, as neither seems to be particularly realistic. I mean, wouldn't a torch a few meters from a wall light up a large section of the wall, similarly to what happens in the first image? On the other hand, wouldn't a torch near the floor light up a large portion of the floor too?

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
emanuele3d: Seems like an improvement to me, but only because I can see a little more! It's really hard to judge I guess because of the interplay between the lighting model per se and the perceptual system. If one lights up a torch in a dark place at the beginning it will blind you, something that could be achieved with exposure/contrast/bloom. But eventually, and especially if you shield the light source somehow, the rest of the environment would get quite visible. With this I do not want to detract from this PR. I'm merely pointing out that we'll probably have too look a little bit more holistically at the whole lighting situation, taking in account both the "simulated physics" or lights in Terasology and then how we want the "simulated eye" to respond to the various situation, something that is done in the post-processing steps of the rendering pipeline.

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
Josharias: Oh,  sorry I was not as clear as I could have been.  These images are with specular light intensity at 1.0 and the other intensities at 0.0.   It is only the specular component of the light.

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
Josharias: Indeed.  I am hoping that the real 3d wizards (or apprentices) will come and put in all those lovely extras.  The take away is the more consistent limits between the held torch and the placed torch.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
emanuele3d: Tricky, tricky, tricky. Again here there's a problem of simulated physics and perception. If I place 20 torches in a place where there was only one, the number of photons is multiplied by 20 and the increase in brightness is justified.  What is not justified is 
1. the sharply burned areas: if they get to be so white bright bloom should kick in and, at least initially, make everything or at least some areas one big bright blob. Eventually however the eye should adapt and everything should look pretty normal, in the general direction of the second image, OR we could decide that depending on the amount of light it stays bright and blinding: after all, if we look into the sun we -DO NOT- adapt to its brightness.
2. the other problem is the pitch black dark surfaces generating a strong sense of contrast with the burned surfaces. That of course doesn't happen in reality because photons bounce around and eventually would reach most surfaces in this kind of setup. This problem I feel it's harder to solve. Can thrown torches be plugged in the light propagation system? Or could they perhaps dimly light up also surfaces that are pointing away from them?

Again, all this is probably out of the scope of this particular PR but I mention it because we might one day have to undo these changes when addressing the whole simulated lighting/simulated perception situation as a whole.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: By eliminating the vector3 variables you seem to be eliminating the potential for a statically tinged (non-white) mainlight. Or could the tinge be added elsewhere, perhaps even at runtime?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: Is the idea here that eventually we'll get block/material-level specular control, so that grass does not have specular reflections?

engine/src/main/resources/assets/shaders/lightGeometryPass_frag.glsl
emanuele3d: Probably the best example of how this set of changes have impacted the look. Looks quite good to me.

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
Josharias: Yeah,  shininess should be configurable for each block type.  But that will need some significant architectural changes to accomplish.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
Josharias: They are still present as their default value.  That said, directional light color needs a major rethink because there are still hardcoded daylight/moonlight values in shaderland.  Additionally,  if we ever got per-world sunlight,  we would want the component stored on the world entity.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
Josharias: I would suggest that thrown torches cannot be plugged in to the light propagation system as it would likely cause a CPU bottleneck.  Mainly because the chunk meshes would have to be constantly updated as the physics updated.

More accurate presentation of light photons would be desireable.  I perused a bunch of different ways global illumination could be more realistic,  but all were:
1. above my paygrade.
2. significantly would change shaderland.

engine/src/main/resources/assets/shaders/lightGeometryPass_frag.glsl
emanuele3d: I guess I would suggest renaming falloff to "cutoff". Falloff and attenuation for me are almost synonyms in this context: how the light intensity attenuates/falls with distance. However, here you are operating a CUToff: beyond a certain distance the light intensity is set to zero. 

Again, this might be something we might want to eventually solve in post-processing: the lighting model might produce dimly-lit distant hills as in the "before" picture and as in reality, but the perception model might take those dimly lit pixels and turn them to black if a nearby light source or surface is bright enough.

engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
emanuele3d: So, do I understand correctly these are images only showing the specular contribution of a light, the diffuse contribution artificially turned off instead?

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
emanuele3d: Ah, thank you for clarifying this. It makes sense now. =) Disregard a similar comment later on then.

emanuele3d:
I think overall and in the short-medium term this PR significantly improves the look of things and solves or improves on issues that have plagued the renderer. In the long term we'll have to take the plunge and look at the interplay between the simulated lighting system and the simulated perceptual system.

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
Cervator: Having played EVE, I love the term "falloff" :D

engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
Josharias: Correct,  diffuse and ambient contributions are artificially turned off to make these images.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
Cervator: Waaaay out of my league but don't we have *two* types of light currently maintained per block: sunlight and "other" light (torches, but perhaps only placed ones, not held/dropped ones?)

If we do (my memory may be out of date) it may still not help since you are talking about dropped torches here. Unless they *should* factor into the "local" light based simply on what blockspace they're in, contributing to overall light locally, but not the direct stuff like the near reflection on immediately visible block faces. Rather, simply light up nearby blockspaces slightly, like sunlight being present a few blocks away, helping defeat the high contrast block faces that aren't in direct visibility from the torches?

I think I might as well be suggesting to a mechanic how to fix a car here (I'd be about as clueless) but just in case it helps... !

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
Cervator: Just makes me think @Josharias is working on turning the torches into flashlights, bringing us from the medieval age into modern times :-)

Cervator:
Very nicely done! This is @emanuele3d-level of well explained 3d wizardry IMHO :D

Here I thought you said you spent 10 hours on the whole PR but just all the screenies and comments sound like they could've taken that!

Anything pending before merging? Haven't tested yet since the screenies tell it so well (and it seems more a subjective topic than a clear cut science you can test)

Josharias:
I wouldn't mind a couple others taking it for a spin to ensure different graphics cards don't do anything very different.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: Ok, thank you for letting me know. I'll see what I can do about this, just nobody holds their breath...

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
emanuele3d: @Josharias: regarding light propagation generating a CPU bottleneck: mental note, it'd be nice to move light propagation to the GPU then, as it is probably parallelizable. That been said, I wonder if it would be horrible to wait until a torch has come to rest before light propagation kicks in. 

@Cervator: I think that's what I was hinting at. But I agree with Josharias this is out of this PR's scope as there would be significant changes.

Cervator:
Tests out fine for me, looks good :-)

Only thing that stuck out is slightly amusing: if you hold a torch in full sunlight (no clouds) and look down at sand it is actually slightly *darker* than without holding a torch. Same in an area with cloud shadows makes it slightly brighter holding a torch.

Full sunlight outside cloud shadows with torch:

terasology-160409233716-1152x720

Switch to shovel - makes it brighter! One powerful shovel

terasology-160409233715-1152x720

Hardly a particularly critical issue with the PR so I'm happy to merge whenever you are :-)

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -171,6 +171,8 @@ private void initMainDirectionalLight() {
         mainDirectionalLight.lightType = LightComponent.LightType.DIRECTIONAL;
         mainDirectionalLight.lightAmbientIntensity = 0.75f;
         mainDirectionalLight.lightDiffuseIntensity = 0.75f;
+        mainDirectionalLight.lightSpecularIntensity = 0.02f;
```
Josharias: Make sunlight much less shiny. (see previous screenshots with what 0.1 was like for sunlight)

terasology-160413203924-1280x720

engine/src/main/java/org/terasology/rendering/logic/LightComponent.java
```diff
@@ -40,15 +40,15 @@
     @Replicate
     public float lightDiffuseIntensity = 1.0f;
     @Replicate
-    public float lightSpecularIntensity = 0.1f;
+    public float lightSpecularIntensity = 0.03f;
```
Josharias: Make torches much less shiny,  yet a little more than sunlight.  You can definitely notice it when you drop a torch right in front of a block,  but it is much more subtle.

terasology-160413203905-1280x720

engine/src/main/resources/assets/shaders/lightGeometryPass_frag.glsl
```diff
@@ -126,7 +126,7 @@ void main() {
 
 #if defined (DYNAMIC_SHADOWS) && defined (FEATURE_LIGHT_DIRECTIONAL)
     // ensure that the shadow does not make the surface completely black
-    shadowTerm = clamp(shadowTerm, 0.25, 1.0);
+    shadowTerm = clamp(shadowTerm, 0.5, 1.0);
```
Josharias: I was having troubles seeing under trees.  Even with eye adaptation turned on.  This makes shadows come more in line with the cloud shadows.

Before:
terasology-160413204259-1280x720

After:
terasology-160413204235-1280x720

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/593/

Josharias:
This is good to merge pending that it passes subjective public opinion.  This should at least last until the rendering gets a overhaul that fixes some of these issues in an less duct tape way.

Cervator:
Checked it again and merging - the screenshots speak for themselves, I don't think anybody can argue with the improvements. I think the balance is really nice now. Tree shadows are a little less dark yet if you hit a really deep forest type spot it can get dark which I like :-)

Nights are decently dark within tree cover as well, seems a fair bit lighter on nice open sandy beaches.

PR #2282, opened 2016-04-09
Use the Biome interface instead of the CoreBiome class

Josharias:
### Contains

This allows modules to extend the biomes in external world generators and maintain compatibility with existing plugin world generation.  This was a problem with using JoshariasSurvival with ShatteredPlanes because the foragable food generation required Core's BiomeFacet.  

With this,  ShatteredPlanes can use the built in Facet mechanisms for biomes and add additional biomes of its own.

### How to test

Core perlin world generation still generates flora according to biome.  The world preview still colors correctly to biome.

- [ ] Do something with CoreBiomeColors


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/579/

Cervator:
Just to get used to thinking about SemVer, would you think this would technically break backwards compatibility with the modding API? As in, could a module previously have been developed in a way this PR would break?

I don't have a good feel yet for how to balance breaking changes vs. stuff that could be considered "Incubating" and not formally a concrete part of the modding API yet (Gradle does that - they insert features but declare them incubating and thus exempt from causing a breaking change)

Since we're still game-alpha I wonder if we should consider a system like that while there's still a fair amount of uncertainty in some of the engine fundamentals :-)

Josharias:
Yes, this will break anything that explicitly gets a CoreBiome out of this particular facet.  Luckily it is contained within the Core module, so there isn't any engine breaks. So I suppose a core module version bump would be in store.

That is a good idea having some sort of mechanism to tag stable (or unstable) mechanisms in the engine so that it is easier to keep sanity around iterative development.  However, I don't exactly know the best way to do it.

Cervator:
So far I've just made the little `API` label we can tag on issues we need to consider a little extra for picking version numbers and thinking about the impact on engine vs. Core modules vs Iota modules vs ... :D

Core technically got the v1.0.0 bump too just because I had to for releasing it but it isn't part of the modding API (yet). Eventually it'll go away / have its final pieces extracted into independent modules, but that's for another day. Maybe by the time we go "game beta" we'll have the Iota collection of modules bundled and declare them part of the modding API (since that would contain stuff like inventory and health)

For now I'll just tag this as a reminder.

engine/src/main/java/org/terasology/world/viewer/layers/NominalFacetLayer.java
```diff
@@ -56,8 +56,10 @@ public void render(BufferedImage img, Region region) {
         for (int z = 0; z < height; z++) {
             for (int x = 0; x < width; x++) {
                 Color src = getColor(facet, x, z);
-                int blend = blender.get(src.rgba());
-                dataBuffer.setElem(z * width + x, blend);
+                if (src != null) {
```
msteiger: Actually, my idea here was that `src` should never be `null`. Did you find a scenario where this is not the case?

modules/Core/src/main/java/org/terasology/core/world/generator/rasterizers/SolidRasterizer.java
```diff
@@ -93,52 +94,52 @@ public void generateChunk(CoreChunk chunk, Region chunkRegion) {
...
-                } else {
-                    return stone;
-                }
-            case BEACH:
-                if (depth < 3) {
-                    return sand;
-                } else {
-                    return stone;
-                }
+    private Block getSurfaceBlock(int depth, int height, Biome type, int seaLevel) {
```
msteiger: What is the difference to the block above?

msteiger:
I'm not sure what the motivation here is. Are you trying to find a way to combine the color mapping of `CoreBiome` with extensions? Maybe we can sort this out on Slack?

engine/src/main/java/org/terasology/world/viewer/layers/NominalFacetLayer.java
Josharias: If an external module of unknown quality does not play be the rules, it causes chaos.  I agree that it *should* never be null.   I found this when getting shatteredplanes to map to the Core biomes

Josharias:
@msteiger, how about this for simplifying things and moving farther away from exposing CoreBiome implementations?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/595/

flo:
I am not familiar with that code area but to me your proposed change looks good.

msteiger:
Will have a look tonight - sorry for the delay.

msteiger:
I'd rather keep the `Biome` interface as slim as possible and believe that color is not a property of a biome. Same goes for humidity and temperature, which don't fit in the PolyWorld setting, for example.

If the preview generator uses a different source for color, the method would quickly become obsolete, too. The AWT facade, for example, uses real block textures instead.

Instead, I suggest using a mapping function that assigns colors to biomes in ShatteredPlanes in analogy to the mapping in Core. A BiomeFacetLayer in ShatteredPlanes could work with the `Biome` interface: try to get a color from the ShatteredPlane color mapping and use `CoreBiomeColors` mapping as a fallback. This should cover all needs in a flexible way and without API break.

The `SolidRasterizer` problem could be approached in a similar manner: externalize the mapping `getSurfaceBlock(int depth, int height, CoreBiome type, int seaLevel)` to a new class and use it in both Core and ShatteredPlanes. The latter could again first to find a block for the new biome type and use the Core method as a fallback. This would avoid the ambiguous name comparisons as well.

I can sketch something, if desired.

Josharias:
Yup,  I believe I understand the direction you can see this going.  I think I will close this PR and submit another one fresh sometime soon(tm).  

> If the preview generator uses a different source for color

I am curious,  could you explain this point a bit more?  Is it because some module may want to re-colorize the core biomes? 

Maybe we should be asking the question, "Are Biomes actually doing something useful for us?"  Right now, I think the only usefulness after world gen they have is to tint grass.  Which itself is not a really reusable piece of magic either (I dont think there is a way to use that mechanism to tint anything but grass). 

Could/should biomes be a world gen only concept?

Cervator:
Not sure if this makes any sense or fits the situation (may be a different kind of "biome") but I like the idea of being able to use (and modify) biome later in a world's lifetime. My favorite example is using the biome field to mark a cave as "inside, cave" when it is created, but then allow for a future event that breaks the roof of the cave, exposing it to the outside. Then let the biome change to "outside" and gradually encroach deeper over time (as outside plants slowly grow a little inside the cave area)

That may be better left for a different concept though, especially since it may feel more like a tag-based concept. If a city is built in a tropical savanna are blocks deep in the city considered "city", "savannah", or both"? I just thought it up as a use case for the biome short available per-block.which has so much more potential than is tapped at present.

Josharias:
@Cervator, that sounds a lot like light propagation.   I wonder if we just need a more reusable block attribute propagation system that could "spread" things like this, each with its own ruleset.  Water is another one that could benefit from some system like this.

msteiger:
@Josharias I meant that the world preview generator could use the texture of the highest block to render a preview rather than the predefined Biome color. Now that I think of it, the Minimap module does that too to get trees in the picture. This is "real" rendering, not biome colors:

image


I agree that Biomes might not be very valuable after world generation, in particular not on a per-block basis. It could be a facet like any other, afaics. Maybe we should start a forum thread to discuss that?

@Cervator I think of Biome in a [traditional and large-scale sense](https://en.wikipedia.org/wiki/Biome#Walter_system). Cave or not, I'd say it's still a tropical or Mediterranean biome. That said, I agree that meta-information such as being in cave or settlement could be relevant at runtime. Cities stores that information in world entities and uses classic collision detection to derive that info.

PR #2286, opened 2016-04-09
Add auto complete for "showScreen" command

tdgunes:
### Contains

As title says, adds suggester for showScreen command. I couldn't resist to add this one. :)

### How to test

Open console, write showScreen and hit TAB, it should be able show available assets for selection. 

### Demonstration
screen shot 2016-04-09 at 17 14 44

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -15,6 +15,7 @@
  */
 package org.terasology.logic.console.commands;
 
+import org.terasology.logic.console.suggesters.ScreenSuggester;
 import org.terasology.utilities.Assets;
 import org.terasology.assets.management.AssetManager;
 import org.terasology.config.Config;
@@ -206,7 +207,7 @@ public class CoreCommands extends BaseComponentSystem {
     }
 
     @Command(shortDescription = "Shows a ui screen", helpText = "Can be used for debugging/testing, example: \"showScreen migTestScreen\"")
-    public String showScreen(@CommandParam("uri") String uri) {
+    public String showScreen(@CommandParam(value="uri", suggester = ScreenSuggester.class) String uri) {
         return nuiManager.pushScreen(uri) != null ? "Success" : "Not found";
     }
 
--- a/engine/src/main/java/org/terasology/logic/console/suggesters/ScreenSuggester.java
+++ b/engine/src/main/java/org/terasology/logic/console/suggesters/ScreenSuggester.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 MovingBlocks
+ * Copyright 2016 MovingBlocks
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/581/

msteiger:
Nice! Code looks good to me. Haven't tested it (yet). Does this work with qualified names as well?

tdgunes:
Thanks @msteiger, sorry but what do you actually mean by "qualified names"?

msteiger:
I meant to ask whether it (still) works, if you add the module name in front.
Example: `engine:migTestScreen`.

I just spotted that you're using `UIWidget`, which probably results in large amounts of false positives. Maybe you can narrow it down by using `ScreenLayer` or sth. similar?

tdgunes:
@msteiger it works, but auto completion only works with resource name. 
For narrowing down the results to UIScreenLayer, I didn't notice that definitely find a fix for it. 

@msteiger do you think auto completion should make suggestions with module name or without module name or both (with module name and resource name)?

The problem with both is that, there are lot's of duplicates:

screen shot 2016-04-10 at 00 45 19

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/586/

Cervator:
Tested it out real quick and I can confirm that it works, although now *only* with fully qualified command names :-)

For me the most intuitive would be without module/engine name unless there's an ambiguous screen name. But is that where the dupes come in?

tdgunes:
For instance, if there is `engine:migTestScreen`  and `newModule:migTestScreen`, both should be visible or else only `migTestScreen`, is that what you are trying say? In this way, there will be no duplicates, I guess.

Cervator:
@tdgunes exactly :-)

If you were to hit tab with `showScreen mig` present it would be nice if the logic was smart enough to pick either the engine or newModule option (starting at random or whatever) but leave both entries in the list so you can tab through them.

tdgunes:
@Cervator that would be great to have! :)

For testing, how can I create `newModule:migTestScreen`?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/588/

Cervator:
> For testing, how can I create a newModule:migTestScreen?

Not really my expertise but I'm hoping you could just make a copy of engine/src/main/resources/assets/ui/migTestScreen.ui and drop it straight into a similar path of a test module then run the game and see if it shows up? The file has `engine:migTestScreen` in it so just change that part to `someModule:migTestScreen`

Maybe that's naive of me to suggest or perhaps the engine really is that easy to use, unsure which :D

tdgunes:
Oh ok, :). I didn't think it would be that easy. By the way, is there a mechanism to state that "Terasology, please load this module in start"?

In theory, my latest commit generates suggestions just like I mentioned above. But I wanted to be sure whether it works or not.

Cervator:
> Oh ok, :). I didn't think it would be that easy. By the way, is there a mechanism to state that "Terasology, please load this module in start"?

As in, loading a module in the main menu so you can see things from the console there? I don't think so at present. We'll need that eventually for modules that reskin or otherwise change the menus.

Cervator:
I went ahead and tested this again just to be sure and all looks well :-)

Can confirm `migTestScreen` shows up unqualified if there is only one, and qualified if there are two (simply copied the `.ui` to Core) with the game running. Both work and open the UI.

With two options you can't auto-complete with `mig<tab>` since at that point it starts with either `engine` or `Core` - did you intend to find a way to make that work, @tdgunes? I'm happy merging this as-is since it is a nice addition and probably a rare use case to have identically named UIs in several places :-)

tdgunes:
Nice to hear that it works! :-)

I think making it work like that way, requires some changes in suggestion system. It seems that suggestion system only start auto completion from beginning. I will definitely look into that in future, when I have some free time. 

This PR will be extremely helpful for me while testing #2288 (hate typing same thing over and over again :-) ), it would be great if you can merge it.

Cervator:
Works for me! Thanks :-)

Issue #2283, opened 2016-04-09
Div by zero on min-length scrollbar (UI test screen)


Cervator:
Discovered while testing UI changes via `showScreen migTestScreen` 

If you click the "min" size UIScrollbar (or click-drag it) you get the following Exception - probably either easy to guard against or don't allow an entirely minimal scrollbar? Might be a unlikely use case in the first place (like other recent creative misuses of UI elements) but nothing crashes, you just get a quiet stacktrace.

```java
00:48:36.188 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
java.lang.ArithmeticException: / by zero
	at org.terasology.rendering.nui.widgets.UIScrollbar$2.onMouseClick(UIScrollbar.java:94) ~[classes/:na]
	at org.terasology.rendering.nui.internal.CanvasImpl.processMouseClick(CanvasImpl.java:224) ~[classes/:na]
	at org.terasology.rendering.nui.internal.NUIManagerInternal.mouseButtonEvent(NUIManagerInternal.java:474) ~[classes/:na]
	at org.terasology.rendering.nui.internal.NUIManagerInternalMethodAccess.invoke(Unknown Source) ~[reflectasm-1.11.0.jar:na]
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:506) ~[classes/:na]
	at org.terasology.entitySystem.event.internal.EventSystemImpl.sendConsumableEvent(EventSystemImpl.java:279) [classes/:na]
	at org.terasology.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:258) [classes/:na]
	at org.terasology.entitySystem.entity.internal.BaseEntityRef.send(BaseEntityRef.java:145) [classes/:na]
	at org.terasology.input.InputSystem.sendMouseEvent(InputSystem.java:503) [classes/:na]
	at org.terasology.input.InputSystem.processMouseInput(InputSystem.java:324) [classes/:na]
	at org.terasology.input.InputSystem.update(InputSystem.java:271) [classes/:na]
	at org.terasology.engine.modes.StateMainMenu.handleInput(StateMainMenu.java:138) [classes/:na]
	at org.terasology.engine.subsystem.lwjgl.LwjglInput.postUpdate(LwjglInput.java:54) [classes/:na]
[...]

PR #2287, opened 2016-04-09
UIScrollbar zero division hardening

rzats:
### Contains

Short and simple fix for #2283. I originally wanted to couple this with more zero division fixes, but after some investigation (i.e. searching for '/' in widget code) it seems that these are actually the only lines of code in the UI* widget family that don't handle zero division properly.

### How to test

Launch the game, type in `showScreen migTestScreen`, click on the `min`-sized UIScrollbar, check that it didn't throw a silent stacktrace.

No outstanding issues.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UIScrollbar.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UIScrollbar.java
@@ -91,11 +91,11 @@ public class UIScrollbar extends CoreWidget {
                 if (vertical) {
                     updatePosition(pos.y - mouseOffset);
 
-                    setValue(TeraMath.clamp(pos.y - mouseOffset, 0, sliderSize) * getRange() / sliderSize);
+                    setValue((sliderSize > 0) ? TeraMath.clamp(pos.y - mouseOffset, 0, sliderSize) * getRange() / sliderSize : 0);
                 } else {
                     updatePosition(pos.x - mouseOffset);
 
-                    setValue(TeraMath.clamp(pos.x - mouseOffset, 0, sliderSize) * getRange() / sliderSize);
+                    setValue((sliderSize > 0) ? TeraMath.clamp(pos.x - mouseOffset, 0, sliderSize) * getRange() / sliderSize : 0);
                 }
                 dragging = true;
                 return true;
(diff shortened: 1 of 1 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/582/

msteiger:
Thanks, @rzats. Does a `sliderSize` value `0` make sense?
If so (and this is the only spot where slider size is a divisor), then I would split the modified line in two to make it easier to read and understand. 
If not, then I'd clamp the minimum slider size to 1 (or maybe more?).

rzats:
@msteiger: `sliderSize` is checked for zero in the same file [(blob link)](https://github.com/rzats/Terasology/blob/d78b51bb619b9a9d9bb87334f2abf0ea3d530f93/engine/src/main/java/org/terasology/rendering/nui/widgets/UIScrollbar.java#L229), so I assume it's an acceptable value. 

Not quite sure what you mean regarding splitting the line in two. Should the control flow look like this?
```java
if (sliderSize > 0) {
    if (vertical) {
        updatePosition(pos.y - mouseOffset);
        setValue(TeraMath.clamp(pos.y - mouseOffset, 0, sliderSize) * getRange() / sliderSize);
    } else {
        updatePosition(pos.x - mouseOffset);
        setValue(TeraMath.clamp(pos.x - mouseOffset, 0, sliderSize) * getRange() / sliderSize);
    }
} else {
    setValue(0);
}
```

msteiger:
Thanks for the info. I find this easier to read and understand. You might need to add `updatePosition(..)` for the case where `sliderSize` is zero. How about this?
```java
int delta;

if (vertical) {
    delta = pos.y - mouseOffset;
} else {
    delta = pos.x - mouseOffset;
}
updatePosition(delta);
if (sliderSize > 0) {
    int clamped = TeraMath.clamp(delta, 0, sliderSize);
    setValue(clamped * getRange() / sliderSize);
} else {
    setValue( .. what is a reasonable value in that case and why? );
}
```

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/584/

rzats:
@msteiger: thanks for the feedback! Updated accordingly.
Fallback value should definitely be set to 0, IMO (since the slider is 0-sized, any higher values may potentially cause unexpected issues).

msteiger:
Great - thanks for the update.

PR #2289, opened 2016-04-09
Added item pickup time

portokaliu:
I have created a new movement mode ("Unhinged") that you can modify the "running" speed for with "setSpeedModifier" changes the speed of either current mode, or a specified one.

Items will now be picked up with a set amount of delay.


GooeyHub:

Cervator:
@GooeyHub: ok to test

Thanks @portokaliu ! Would you mind giving a bit more of an overall explanation of what was changed, how to test, and your goal with the changes? :-)

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/585/

portokaliu:
About the new movement mode:
I wanted to add a way to change the movement speed of the character without changing the overall "running speed". As I looked through the code it seemed necessary that a component that affected the movement speed would be necessary. That being said, I haven't separated the 2 axes yet, but it's a start for future improvements.

To test, just go type the "unhinged" command in console. It should be like the "ghost" one, without the fov change

About the debug command:
I found it necessary for movement modes to have easy to access speed modifiers, thus a debug command was needed.
You have the option of setting the speed of the current mode if you don't specify one. Of course, this just changes a speed parameter in the speed component so that logic elements may use it for further processing. In this case, the "unhinged" parameter is used in the KinematicCharacterMover by setting the maximum speed of the player 'x' times over the normal one while running. Otherwise, it keeps the "not running" speed, which is 1/1.45 * normal running speed.
For every mode that someone would want enhanced by this component, they would have to add a default value in the player prefab, a value in the component, and add the case scenarios in the ServerCharacterPredictionSystem and the SetSpeedModifierEvent.

A way to test it is to go in unhinged mode with the "unhinged" command,  then the "setSpeedModifier" command with any amount you wish to multiply the speed by. I don't think this speed is bound by server rules, but it is a player specific speed modifier, so I don't think it matters. I'll take any criticism tho :)

About the items change.
Items will now be dropped from the player with a 2 second pickup delay, and from items with the default value that exists in the ItemComponent, which I currently left 0. The condition for picking them up is the "pickableUp" bool to be true.
The current way I'm going about this is setting a delayed event as the item is dropped. This event will start after a time specified by the ItemComponent, called  "pickUpTime". This is the time in miliseconds after which any player can pick up the dropped items.
I plan on using this change in the future to make it so that "the owner will take longer to pick up the items than other players, but if the owner sticks around his dropped items, no other player can pick them up before him".

Way to test : throw items on the ground at your feet. They'll pop into your inventory as soon as their timer runs out. Blocks dropper from... blocks will still be 0 time pickup, unless you change the default value in the component.

Hope this was a bit more detailed @Cervator :)

flo:
As I explained on IRC, the movement speed modification possiblity seems redundant to GetMaxSpeedEvent. e.g.

https://github.com/Terasology/AlterationEffects/blob/master/src/main/java/org/terasology/alterationEffects/speed/SpeedAlterationSystem.java

engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java
```diff
@@ -59,20 +66,32 @@ public void onDropItemEvent(DropItemEvent event, EntityRef itemEntity, ItemCompo
...
+
+        itemEntity.getComponent(ItemComponent.class).pickableUp = false;
+        long timeToPickUp = itemEntity.getComponent(ItemComponent.class).pickUpTime;
+        delayManager.addDelayedAction(itemEntity, DELAYED_PICKUP_ACTION_ID, timeToPickUp);
     }
 
+    @ReceiveEvent
+    public void onDelayedPickup(DelayedActionTriggeredEvent event, EntityRef entityRef) {
+        if (event.getActionId().equals(DELAYED_PICKUP_ACTION_ID)) {
+            entityRef.getComponent(ItemComponent.class).pickableUp = true;
```
Josharias: Dont forget to save the component as well so that it will function in a remote client.

engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java
```diff
@@ -59,20 +66,32 @@ public void onDropItemEvent(DropItemEvent event, EntityRef itemEntity, ItemCompo
...
             itemEntity.addComponent(new LocationComponent(event.getPosition()));
         }
+
+        itemEntity.getComponent(ItemComponent.class).pickableUp = false;
+        long timeToPickUp = itemEntity.getComponent(ItemComponent.class).pickUpTime;
+        delayManager.addDelayedAction(itemEntity, DELAYED_PICKUP_ACTION_ID, timeToPickUp);
     }
 
+    @ReceiveEvent
+    public void onDelayedPickup(DelayedActionTriggeredEvent event, EntityRef entityRef) {
```
Josharias: You could possibly add the ItemComponent here as a parameter that way some optimizations can be done by only firing this code when a ItemComponent is present.  Also,  then you dont have to do a hasComponent or null check (which you dont presently have)

engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java
```diff
@@ -59,20 +66,32 @@ public void onDropItemEvent(DropItemEvent event, EntityRef itemEntity, ItemCompo
...
+        if (event.getActionId().equals(DELAYED_PICKUP_ACTION_ID)) {
+            entityRef.getComponent(ItemComponent.class).pickableUp = true;
+        }
+    }
 
     @ReceiveEvent
     public void onBumpGiveItemToEntity(CollideEvent event, EntityRef entity, PickupComponent pickupComponent) {
-        GiveItemEvent giveItemEvent = new GiveItemEvent(event.getOtherEntity());
-        entity.send(giveItemEvent);
+        if ( entity.getComponent(ItemComponent.class).pickableUp == true ) {
```
Josharias: You will probably want to ensure an ItemComponent exists on this first before doing this check.  This has bit me more times than I care to admit.

modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
```diff
@@ -115,6 +115,7 @@ public void onDropItemRequest(DropItemRequest event, EntityRef character) {
             }
         }
 
+        pickupItem.getComponent(ItemComponent.class).pickUpTime = 2000;
```
Josharias: Although very likely,  there is no guarantee that his will have an ItemComponent present.

Josharias: Oh,  something like this should probably live in the DropItemEvent?   Possibly the intention was to allow the character to provide some default value?  So that each character could customize how long the item stays unpickupable?  Otherwise,  I would expect setting a default amount of time to be set on an item's prefab (there is a common parent you could plunk the default value on to...  is it called "itemBase.prefab"?)

engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java
```diff
@@ -59,20 +66,32 @@ public void onDropItemEvent(DropItemEvent event, EntityRef itemEntity, ItemCompo
         if (!itemEntity.hasComponent(LocationComponent.class)) {
             itemEntity.addComponent(new LocationComponent(event.getPosition()));
         }
+
+        itemEntity.getComponent(ItemComponent.class).pickableUp = false;
+        long timeToPickUp = itemEntity.getComponent(ItemComponent.class).pickUpTime;
+        delayManager.addDelayedAction(itemEntity, DELAYED_PICKUP_ACTION_ID, timeToPickUp);
```
Josharias: You could also consider setting the game time that this component would be pick-up-able.  That way you could just test on collision if the game time is greater than the time saved and there would be less moving parts with setting the flag back and forth (you would not need to use the delayed action anymore)

engine/src/main/java/org/terasology/logic/inventory/ItemComponent.java
```diff
@@ -80,6 +80,10 @@
      */
     public int baseDamage = 1;
 
+    public boolean pickableUp = true;
```
Josharias: Could you create another component that has this delayed pickup stuff on it?  I have dreams to slim ItemComponent down until multiple components so that the item system becomes more modular and less monolithic.

engine/src/main/java/org/terasology/logic/characters/CharacterMovementComponent.java
```diff
@@ -71,6 +71,8 @@
     // Distance since last footstep
     public float footstepDelta;
 
+    // Distance desired from ground
+    public float groundDistance;
```
Josharias: I must have missed the conversation part about what this is about... what is this for?

Josharias:
I would also suggest sticking with the GetMaxSpeedEvent for adjusting movement speed.  It is presently complicated enough with the moving parts we have on character movement, I could see this muddying it up more so and making too many special conditions in the engine.

The speed modification component looked like an ok idea until I started thinking about what would happen long term.  Eventually would would need a field in the component for every movement mode.  And then additionally, you would also need a field for every type of modification that is possible with the GetMaxSpeedEvent where you can multiply before or after adding a static value.  It would start to get cumbersome quickly.

That said,  having the new Unhinged movement mode is a fine idea.  Basically mix and matching the movement settings together in a unique way.

msteiger:
Can you please update the title of this PR so it relates to the content?

portokaliu:
Sorry, didn't realize that was the name of the PR  ^^'

engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java
portokaliu: Couldn't I just do as above and only use onBumpGiveItemToEntity when the ItemComponent is present?

modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
portokaliu: About this, the idea was for the item to drop from the player's hand for that set amount of time. If I put that in the onDropItemEvent, I'd have to either have a way to check what dropped it to make sure I set the right amount of time. (whereas the DropItemRequest comes directly from the character, thus I control the time to pick up from it being thrown by the character instead of every block)

Having the pickUpTime dependant on the item entity is a bit different mechanic, but I could just do that if you think it's a better idea.

engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java
portokaliu: I wanted to use the delayer mostly for testing purposes. I wanted implement this step by step to get used to this part of the code. Now that I know what it looks like, what you said sound like the better idea :)

engine/src/main/java/org/terasology/logic/inventory/ItemComponent.java
portokaliu: Sure thing! But what would I call the new component? The only thing that comes to mind is "PickableComponent", but that doesn't sound right.

engine/src/main/java/org/terasology/logic/characters/CharacterMovementComponent.java
portokaliu: Ah, this is just a bit of code I didn't remove from rolling back HoverCam mode (which will keep the player a... player controlled distance from the ground). I'm guessing I'll put that in a hoverCam component?

portokaliu:
@Josharias About the speed modification, I talked to @flo on IRC about this, and my interpretation of what he said was that he doesn't completely agree with the current movement system, and that separating the movement modes into different classes would be a good idea; From an outsider perspective I found this a good idea, and having them separated would lean more towards the modular system that the project is trying to achieve.

I was thinking that we could separate the speed types as well (walking speed - shift pressed, and running speed - no shift) so that we can accurately modify the speed depending on the movement mode and movement speed modifiers ( for this I was thinking of having a component that keeps every type of speed modifier in one place, with a bool for the "active" status and a value for the speed impact. (and there should be some time element there as well, I'm not completely certain how to go about this) )

What is your take on this?

engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java
portokaliu: Good point (both about the optimization and the fact that I forgot checks... all over the place ^^' )

engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java
portokaliu: Is the saving of components relevant to the save system or the general flow of the game? I didn't know that component saving was necessary. Will do it from now on tho :D

engine/src/main/java/org/terasology/logic/inventory/ItemComponent.java
Cervator: IIRC "Pickable" actually has been a valid technical term for that sort of object :-)

If this is specific to delayed, uh, picks? I guess it could be DelayedPickableComponent. Or if it entirely splits out stuff related to picking up items then plain old PickableComponent sounds fine to me. @Josharias is the expert so get his opinion too :-)

flo:
What I meant is that MovementMode should be replaced in the long run with a something entity system based where new modes can be added easily. e.g. one approach could be to have one component class per movement mode. The movementMode field could then be a reference to a class. To get the current speed you would then do:
```
CharacterMovementComponent movementComponent = entiy.getComponent(CharacterMovementComponent.class)
Class<?> currentMovementModeClass = movementComponent.movementMode;
MovementModeComponent movementMode = entiy.getComponent(currentMovementModeClass);
float currentMaxSpeed = movementMode.maxSpeed;
```
MovementModeComponent would have the same field like MovementMode had and the concrete movement modes would inherit from it. Then you could specify the different movement modes like this in the prefab:
```
"WalkMovementMode": {
    "scaleGravity": 1.0,
     "scaleInertia": 1,
     "useCollision": true,
     "canBeGrounded": true,
     " respondToEnvironment": true,
     "maxSpeed": 3.0,
     "applyInertiaToVertical": false
},
"FlyMovementMode": {
    "scaleGravity": 0.0,
     "scaleInertia": 4,
     "useCollision": true,
     "canBeGrounded": false,
     " respondToEnvironment": false,
     "maxSpeed": 3.0,
     "applyInertiaToVertical": true
},
```
Another approach would be to have an entity per movement mode. That would be even more flexible, but I think we currently can't define sub entities in prefabs which would mean that code needs to add movement modes from prefabs.

Josharias:
Something like what @flo suggested sounds promising.  I would imagine only one of ```WalkMovementMode``` or ```FlyMovementMode``` being added to an entity at one time,  and then the engine would collect these variables like:

```java
@ReceiveEvent
public void setMovementDetails(SomeSpecialEvent event, EntityRef entity, WalkMovementMode walkComponent, CharacterMovementComponent characterMovementComponent) {
    event.setAllTheStuffHere(...);
    event.setTheMaxSpeedHere(...);

    // ensure only one system can set the movement details
    event.consume();
}
```

This way you would not have to store what mode you are in on the CharacterMovementComponent. And it would be easy to plug in to for external modules.

portokaliu:
@flo I'm trying to do that, but also want to add a "use collision damage" parameter, because sometimes you want to be stopped by the terrain but health not be an issue. Otherwise and "invulnerable" parameter would be ok, preventing any type of damage event to go through to the player (like it was described in the event system in general - by consuming the event if the param is true)

On the movement mode component idea, why can't we just get the component from the prefab and associate it to our current entity? Unless you meant that we could keep all movement modes under a single prefab with multiple defined entities, I'm not sure what you're saying.

@Josharias I stumbled upon AbstarctConsumableEvents while coding. Was wondering if I couldn't just use those for the same purpose?

Josharias:
Yes, the ```AbstractConsumableEvent``` is the one to use.  It has special logic in the engine to skip calling the rest of the event handlers if it is consumed.  By using an event to collect the movement variables,  then you do not have to limit movement modes to subclassing from a base component.

In regards to the damage bit, I would keep health and movement concepts isolated from each other so that we avoid flags and settings scattered around,  making it difficult to control.  That said,  there is already a few settings in the ```HealthComponent``` that relate to collision with the world: fallingDamageSpeedThreshold, horizontalDamageSpeedThreshold, excessSpeedDamageMultiplier.

flo:
@portokaliu  My first idea was just to have a single prefab for the character that has 1 component for each movement type. Thus character entities will also have 1 component for each movement type and it is possible to just  for example modfiy the walking speed while letting the flying speed in tact.

My second idea was to have 1 prefab per movement mode. The prefabs will be used to create 1 entity per movement mode and per character. Each character will have its own set of movement mode entities that represent possible movement modes. One of the movement mode entities of the character gets referenced by the movementMode field (the active one).

The prefab for the different movement modes will have the same components types, but different values. e.g. the walk prefab could look like this:
```
"CoreMovementModeProperties": {
     "scaleGravity": 1.0,
     "scaleInertia": 1,
     "useCollision": true,
     "canBeGrounded": true,
     "respondToEnvironment": true,
     "maxSpeed": 3.0,
     "applyInertiaToVertical": false
},
"MyComponent": {
   "useCollisionDamage": true
},
"ModeBasedAnimation": {
  "animationPool": ["humanWalk"]
}
```
The fly movement module would then look like this:
"FlyMovementMode": {
     "scaleGravity": 0.0,
     "scaleInertia": 4,
     "useCollision": true,
     "canBeGrounded": false,
     "respondToEnvironment": false,
     "maxSpeed": 3.0,
     "applyInertiaToVertical": true
},
"MyComponent": {
   "useCollisionDamage": false
},
"ModeBasedAnimation": {
  "animationPool": ["humanFly"]
}
```
@portokaliu  We could of course also only have those components at the character entity and copy then from prefabs. Then the question is however: When do we copy? When we copy the speed values from preabs it is course not possible to change the values of a movement mode at runtime.

 @Josharias I think the idea with the events sounds also very promising. e.g. there could be a MovementModeUpdated event (not consumeable). It could trigger a RecalculateMaxSpeedEvent. The RecalculateMaxSpeedEvent handler  could then send a GetMaxSpeedEvent and uses the result to update the speed value.A GetMaxSpeedEvent event handler with the condition WalkingComponent exists would then set the speed to a default value. A low priority GetMaxSpeedEvent event handler sets a default value but it will only be called if the event does not get consumed first. The WalkingComponent would just be a marker component without content. (The marker component concept could also be used later to have animals flee from the player when it is in running towards them (character has RunningComponent))

Thus GetMaxSpeedEvent would no longer have to be spamed and we could introduce events for other fields too.

portokaliu:
Hey guys

I read this whole thing over and I haven't really understood the need for separate entities for movement modes; I mean, I also don't see anything wrong with the current way of doing stuff. If you say it's "monolithic" to have all the movement modes in a single area, then I'd question the way KinematicCharacterMover is implemented;

Cervator:
It does seem like the discussion in this PR has gotten away from the original purpose into being more about redesigning the movement system to have better overall architecture :-)

Probably the current movement system *is* worth questioning, but maybe we should do that separately and just finish this up so it can be merged.

@portokaliu I noticed a couple odd Javadoc blocks and a star import, unsure if there might be a comment or two unaddressed. Maybe add another quick bit of tweaking and perhaps we can merge and worry more later :-)

GooeyHub:

GooeyHub:

PR #2296, opened 2016-04-17
Drop 1 item from inventory on right click in drop region.

### LINKED ISSUE #2295: Inventory related behavior on right-click
Follow-up from #2262 after which you can now left click with a held stack from an inventory to drop the stack outside the inventory where it drops to the ground.

Discussion in the PR touched on what to do with what kinds of clicks. At the same time playing with the inventory and comparing to MC I'm remembering that with MC right click with a held stack over an inventory cell drops a single item from the stack into that cell.

Currently we just have right click on a stack == pick up half. Maybe we should go ahead and add right click *with held stack* == drop one item into cell. With #2262 merged we could do the same to drop single items from an inventory. You could just pick up a single item and drop it that way but if we add it for inventory cells I figure we might as well add it for dropping :-)

Regular right click in the world will activate an item but I figure that is distinct from working an inventory and holding a stack on the cursor.

Slightly related / similar: #2101

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/33026102-inventory-related-behavior-on-right-click?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

genuinelucifer:
Fixes #2295 

Adds dropping of 1 item when user presses right click while holding a stack of item on the drop region around inventory grid.

### DIFF
--- a/modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/DropItemRegion.java
+++ b/modules/Core/src/main/java/org/terasology/rendering/nui/layers/hud/DropItemRegion.java
@@ -42,14 +42,17 @@ public class DropItemRegion extends CoreHudWidget {
         @Override
         public boolean onMouseClick(NUIMouseClickEvent event) {
             MouseInput mouseButton = event.getMouseButton();
-            if (mouseButton == MouseInput.MOUSE_LEFT) {
+            if (mouseButton == MouseInput.MOUSE_LEFT || mouseButton == MouseInput.MOUSE_RIGHT) {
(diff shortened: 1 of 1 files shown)


GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/603/

PR #2301, opened 2016-04-21
Fix normal maps when looking at a block on all sides.

Josharias:
### Contains

- Fix normal maps when looking at a block on all sides.
- Add "extra lighting" video option to turn on normal maps.
- Add gloss map block tiles that turn on specular light reflections for parts of a tile. This replaces the ```LightComponent.lightSpecularIntensity``` field.
- Add IronBar block to demonstrate the gloss effect.

[Video demo](https://youtu.be/54b7kA25Vno)

There is one ill side effect of this change in that the normal world atlas image that is spit out into the screenshots directory will look mostly white because by default it is fully transparent.

### How to test

Turn on "extra lighting" in the video options.  Ensure that specular lighting (shininess/glossiness) does not appear by default on blocks.  Place a IronBar block and hold up a torch to it.

- [x] Adjust the lightProperties to be a vec3 instead of a vec4 with an unused field.
- [x] Rename applyGrayscaleMaskToAlpha(...) to storeGreyscaleMapIntoAlpha(...) and make the parameter names more clear.


engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
```diff
@@ -250,9 +269,18 @@ private void createTextureAtlas(final Texture texture) {
...
 
     private ByteBuffer[] createAtlasMipmaps(int numMipMaps, Color initialColor, List<BlockTile> tileImages, String screenshotName) {
+        return createAtlasMipmaps(numMipMaps, initialColor, tileImages, screenshotName, Lists.newArrayList());
+    }
+
+    private ByteBuffer[] createAtlasMipmaps(int numMipMaps, Color initialColor, List<BlockTile> tileImages, String screenshotName, List<BlockTile> alphaMaskTiles) {
         ByteBuffer[] data = new ByteBuffer[numMipMaps];
         for (int i = 0; i < numMipMaps; ++i) {
             BufferedImage image = generateAtlas(i, tileImages, initialColor);
+            if (alphaMaskTiles.size() > 0) {
```
Josharias: I am not too enamoured with how this code is so embedded in amongst other things.  It was the only short term solution I could think of to get a shininess parameter into the existing shader textures.

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
```diff
@@ -205,7 +224,7 @@ private void buildAtlas() {
 
         int numMipMaps = getNumMipmaps();
         ByteBuffer[] data = createAtlasMipmaps(numMipMaps, TRANSPARENT_COLOR, tiles, "tiles.png");
-        ByteBuffer[] dataNormal = createAtlasMipmaps(numMipMaps, UNIT_Z_COLOR, tilesNormal, "tilesNormal.png");
+        ByteBuffer[] dataNormal = createAtlasMipmaps(numMipMaps, UNIT_Z_COLOR, tilesNormal, "tilesNormal.png", tilesGloss);
```
Josharias: Add the gloss tiles to the alpha channel of the normal map texture.  That way it can get sent in to the shaders without binding another texture and creating the infrastructure to support another bound texture through the layers of shaderland.

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
```diff
@@ -310,13 +355,14 @@ private BufferedImage generateAtlas(int mipMapLevel, List<BlockTile> tileImages,
...
         g.setColor(clearColor);
-        g.fillRect(0, 0, size, size);
         for (int index = 0; index < tileImages.size(); ++index) {
             int posX = (index) % tilesPerDim;
             int posY = (index) / tilesPerDim;
             BlockTile tile = tileImages.get(index);
             if (tile != null) {
                 g.drawImage(tile.getImage().getScaledInstance(textureSize, textureSize, Image.SCALE_SMOOTH), posX * textureSize, posY * textureSize, null);
+            } else {
+                g.fillRect(posX * textureSize, posY * textureSize, textureSize, textureSize);
```
Josharias: Fill each tile separately so that transparency on the original tile does not get blended with a base color.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -112,21 +112,17 @@ void main() {
...
+#if defined (PARALLAX_MAPPING)
     vec3 eyeTangentSpace = tbn * vertexViewPos.xyz;
 
     float height =  parallaxScale * texture2D(textureAtlasHeight, texCoord).r - parallaxBias;
 	texCoord += height * normalize(eyeTangentSpace).xy * TEXTURE_OFFSET;
 #endif
 
-    normalOpaque = (texture2D(textureAtlasNormal, texCoord).xyz * 2.0 - 1.0);
-
-    // Simplified tangent basis - because we can! Voxels and blocks are great
```
Josharias: If anyone knows how to actually get the simplified tangent basis to work,  please let me know and we can switch back to something less intensive.  I did not notice any framerate difference having to compute the tbn value.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -112,21 +112,17 @@ void main() {
...
-    normalOpaque.xyz = vec3(worldSpaceNormal.x, normalOpaque.x, normalOpaque.y) * abs(worldSpaceNormal.xxx)
-        + vec3(normalOpaque.y, worldSpaceNormal.y, normalOpaque.x) * abs(worldSpaceNormal.yyy)
-        + vec3(normalOpaque.x, normalOpaque.y, worldSpaceNormal.z) * abs(worldSpaceNormal.zzz);
+    normalOpaque = normalize(texture2D(textureAtlasNormal, texCoord).xyz * 2.0 - 1.0);
+    normalOpaque = normalize(tbn * normalOpaque);
```
Josharias: This is the magic that allowed the normal mapping to rotate correctly as you viewed a block from different sides.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -206,11 +202,12 @@ void main() {
     // ...and finally the occlusion value
     float occlusionValue = expOccValue(gl_TexCoord[1].z);
 
-    vec3 blocklightColorValue = calcBlocklightColor(blocklightValue
+    float blocklightColorBrightness = calcBlocklightColorBrightness(blocklightValue
```
Josharias: Capture just the brightness value because we can calculate the color in a different shader without using up 3 float spots.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -277,7 +274,7 @@ void main() {
     // Apply the final lighting mix
     color.xyz *= combinedLightValue * occlusionValue;
 #else
-    gl_FragData[2].rgba = vec4(blocklightColorValue.r, blocklightColorValue.g, blocklightColorValue.b, 0.0);
+    gl_FragData[2].rgba = vec4(blocklightColorBrightness, daylightValue, 0.0, 0.0);
```
Josharias: Store both the brightness (which can be later converted again into a color value) and the daylight value.  This way we can use the #2 frag texture for more things other than just the blocklight color.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -290,8 +287,8 @@ void main() {
     gl_FragData[0].rgb = color.rgb;
     // Encode occlusion value into the alpha channel
     gl_FragData[0].a = occlusionValue;
-    // Encode daylight value into the normal alpha channel
-    gl_FragData[1].a = daylightValue;
+    // Encode shininess value into the normal alpha channel
```
Josharias: Store shininess here because the normal texture gets sent to the geometry shader.

emanuele3d:
Thank you @Josharias for this new feature. I didn't go into the details because a) it works b) Shaderland is still to scary for me. =)

Won't be for long though: Shaderland is my next stop after I wrap up the work on the renderer's top level classes by adding some documentation to the last interface/class pair that still lack it.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/VideoSettingsScreen.java
```diff
@@ -229,6 +229,7 @@ public void set(Float value) {
         WidgetUtil.tryBindCheckbox(this, "volumetricFog", BindHelper.bindBeanProperty("volumetricFog", config.getRendering(), Boolean.TYPE));
         WidgetUtil.tryBindCheckbox(this, "inscattering", BindHelper.bindBeanProperty("inscattering", config.getRendering(), Boolean.TYPE));
         WidgetUtil.tryBindCheckbox(this, "cloudShadow", BindHelper.bindBeanProperty("cloudShadows", config.getRendering(), Boolean.TYPE));
+        WidgetUtil.tryBindCheckbox(this, "extraLighting", BindHelper.bindBeanProperty("normalMapping", config.getRendering(), Boolean.TYPE));
```
emanuele3d: I think Extra Lighting is a bit too generic. I'd prefer something like "Normals-based Glossyness". It's more technical but the user would be able to (eventually) find out through documentation or via tooltips.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -570,7 +569,7 @@ private boolean renderLightComponent(LightComponent lightComponent, Vector3f lig
             program.setFloat3("lightColorAmbient", lightComponent.lightColorAmbient.x, lightComponent.lightColorAmbient.y, lightComponent.lightColorAmbient.z, true);
 
             program.setFloat4("lightProperties", lightComponent.lightAmbientIntensity, lightComponent.lightDiffuseIntensity,
-                    lightComponent.lightSpecularIntensity, lightComponent.lightSpecularPower, true);
+                    0.0f, lightComponent.lightSpecularPower, true);
```
emanuele3d: mmmm...  please avoid magic numbers, especially when the setFloat4 function would give no chance of easily finding out what the fourth parameter passed in means. Perhaps make a constant at the beginning of the file?

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
emanuele3d: Hmmm. So here we are tying normals with glossiness... I wonder what this imply if one day we want to use normals for environmental reflections instead of simple glossyness. Just thinking out loud though, for the time being it's ok.

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
```diff
@@ -275,6 +303,23 @@ private void createTextureAtlas(final Texture texture) {
...
+
+        for (int i = 0; i < imagePixels.length; i++) {
+            int color = imagePixels[i] & 0x00ffffff; // Mask preexisting alpha
+            int alpha = maskPixels[i] << 24; // Shift blue to alpha
+            imagePixels[i] = color | alpha;
+        }
+
+        image.setRGB(0, 0, width, height, imagePixels, 0, width);
+    }
+
```
emanuele3d: I don't get this, mostly because I do not know what the Atlas does. Can you please let me know what is being combined here? There's a block that has color texture, normal texture and glossyness texture. What gets combined with what?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
Josharias: Good point.  This was just to remove the value.  Thinking about this now,  I should make it a vec3 in the shader and then just avoid passing a value alltogether. (tracked in the PR checklist)

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
Josharias: This is used to combine the gloss map on to the normal map as the alpha channel.  The gloss map is just a single per pixel value, and the normal texture was not using the alpha channel.

The atlas is a stitching together of all the block tiles into one texture so that all the block triangles in the world can use the same texture and we can minimize the number of texture switches to opengl. (my 3d wizardry newb-ness is showing)

Cervator:
Neat! Just a quick FYI the Jenkins job for this stalled. I killed it but it didn't seem to change the "pending" piece here.

Not testing yet, but am curious about the details and talk about normals/maps - the full block shininess I get, that's just going "Ok block, be shiny now!" - does this include or impact the potential for one day having blocks be shiny in particular spots, such as a mineral block where the stone bits are dull but exposed mineral/crystal pieces are shiny? Is that's what happening in that one screenshot?

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/VideoSettingsScreen.java
Josharias: Good idea about tooltips.

I chose a generic term because it enables a couple things.  Both Normal mapping and Glossiness.  I also thought it may also be more meaningful to the average Joe.  I would be happy to change to something that has a couple +1's from the team.

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
Josharias: Agreed,  there is definitely some reorganization needed with this stuff.  I hope that a better strategy to accommodate net new features like environmental reflections will come about when this gets an overhaul.  It seems extremely involved to add textures that are shared between shaders at the moment,  hopefully this will become easier in the future.

Josharias:
@Cervator, check out the video.  This implementation already allows defining glossiness for each individual pixel in the texture with the gloss map tile (look at the stone block in the video).

@emanuele3d, I look forward to the renewed strategy in shaders and rendering systems.  It seems one of the more important things to note is that scene textures (whose shader variables are labeled as such) are accessable between shaders, and non-scene textures are the actual representation of what is in java memory.

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/617/

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
emanuele3d: I understand, so the combination at this stage is purely for storage purposes prior to passing it to the GPU. 

I think perhaps the term applyGreyscaleMaskToAlpha is a bit misleading in the sense that it makes it sound like you have something in the alpha channel already and you are then applying/combining some further information onto it. In fact I seem to understand the alpha is empty and you are storing the glossyness map into that channel. I'd be more explicit and name the method "storeGreyscaleMapIntoAlpha()", but then use explicitly named parameters such as normalMap and glossinessMap when you call the method.

emanuele3d:
@Cervator: judging by the video texture-masked shininess is already available with this PR!

@Josharias: I must humbly submit that I really didn't understand your paragraph about scene/non-scene textures. Can you elaborate?

Josharias:
@emanuele3d, I admit that I barely know enough about it also.  I was attempting to mention the difference between a texture that is shared between shaders (where a shader can use the values set by a previous shader) and a texture that is only used in one particular shader.  I was confused for a time trying to figure out why the texture labeled "normal" did not have the data I was expecting.  That was because the "normal" texture that was available was the shared "scene" texture,  not the one that java-land had in memory. (hoping this makes more sense.  If not,  we may have to leave it till I can communicate this better)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/VideoSettingsScreen.java
Cervator: Would it be better with two checkboxes then? We used to have "Environment Effects" which was weirdly vague and covered two separate items that ultimately made more sense as individual checkboxes.

A really fancy and way overblown approach would be to start the config UI in a "Basic" mode with generic/vague option names for fewer grouped option, with an "Advanced" button that would then split out the exact details into individual checkboxes. But that's most likely waaay too much effort for its worth. Just nice and geeky.

Cervator:
Oh! That video is way longer. For some reason I thought that was the tiny 10-sec snippet or so you shared earlier. Yeah that demonstrates it really well :-)

It is unlisted though, should I not share it around some / use it in a news post after we merge this sometime?

Josharias:
Feel free to share the video link, I just don't want it on my public video list.

Cervator:
Ok!

Tested it out. Ooooohhhh shiny .... me likey!

Want crystal/gem overlays for mineral blocks now, and shiny maps for said crystals/gems not to mention metallic ores... woo!

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
Josharias: That is a fair rename.  Logged in the todo checklist.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/VideoSettingsScreen.java
Josharias: I hope to revisit the settings screen soon.  Could we leave "extra lighting" in there until then so that I can add a shader parameter to turn off specular (glossy) light entirely and tie it to a video setting?

GooeyHub:
http://jenkins.terasology.org/job/TerasologyPRs/619/

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/VideoSettingsScreen.java
emanuele3d: Ok for me.

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
emanuele3d: Thank you!

emanuele3d:
@Josharias: regarding shared/non-shared textures, my confusion was arising due to the terms used but I think I now understanding what you mean. 

For the sake of future discussions we might want to think to "textures" as images to apply to an object's surface, i.e the colors and insignia on a flag, the dirt on a wall or the bump/normal map of an orange. This is in line with how textures are thought of in 3D modelling/rendering programs.  

We might then want to use the term (gpu) "buffers" to talk about areas of GPU memory used by shaders to read from and write to. Somewhat confusingly, some of these buffers are of (opengl) type "texture" and do store the textures I defined above. Which is why I just prefer to call them all buffers. 

Also confusingly, buffers are not the same as Frame Buffers, which can have a number of buffers attached to them so that a shader can write to multiple buffers at once, as it is done in some steps of our rendering pipeline.

So, in this context I think in part you were talking about the buffers that shaders use to pass data to each other, and in part you were talking about the textures (such as the normal map) that usually start as a file and eventually get uploaded in a buffer on the GPU for shader to read from, such as the texture atlas. Am I close?

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
```diff
@@ -250,9 +269,18 @@ private void createTextureAtlas(final Texture texture) {
...
+        return createAtlasMipmaps(numMipMaps, initialColor, tileImages, screenshotName, Lists.newArrayList());
+    }
+
+    private ByteBuffer[] createAtlasMipmaps(int numMipMaps, Color initialColor, List<BlockTile> tileImages, String screenshotName, List<BlockTile> alphaMaskTiles) {
         ByteBuffer[] data = new ByteBuffer[numMipMaps];
         for (int i = 0; i < numMipMaps; ++i) {
             BufferedImage image = generateAtlas(i, tileImages, initialColor);
+            if (alphaMaskTiles.size() > 0) {
+                BufferedImage alphaMask = generateAtlas(i, alphaMaskTiles, Color.BLACK);
+                storeGreyscaleMapIntoAlpha(image, alphaMask);
```
emanuele3d: Is this used only to store the specular/glossyness map into the normals map's alpha or also for other things? 

Because I'd be tempted to write storeGreyscaleMapIntoAlpha(normalsMap, specularMap).

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
```diff
@@ -275,6 +303,23 @@ private void createTextureAtlas(final Texture texture) {
         return data;
     }
 
+    // Ref: http://stackoverflow.com/questions/221830/set-bufferedimage-alpha-mask-in-java/8058442#8058442
+    public void storeGreyscaleMapIntoAlpha(BufferedImage imageWithoutAlpha, BufferedImage greyscaleImage) {
+        int width = imageWithoutAlpha.getWidth();
+        int height = imageWithoutAlpha.getHeight();
+
+        int[] imagePixels = imageWithoutAlpha.getRGB(0, 0, width, height, null, 0, width);
```
emanuele3d: I'm wondering here if it'd be worth to use the BufferedImage.getData() method to obtain a Raster object and then iterate over pixels directly instead of going through integer arrays. Not quite a priority though.

Cervator:
Merge time or more tweaking first?

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
Josharias: I am hesitant to make the naming of these things too specific as it does perform a generic task.  Do you feel strongly about tweaking these variable names?

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
Josharias: There were a few different snippets of how to do this online.  The array method here seemed like the most efficient without getting too complicated.

Josharias:
@emanuele3d:  Yes,  frame buffers are the correct term for what I was trying to distinguish from textures.  Thank you for the clarification.  I am sure that writing some OpenGL from hello world would have taught me this distinction instead of just jumping in to a complicated rendering pipeline. :)

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
emanuele3d: I feel there are two reasons to keep the method call this generic:

1. it already deals with more than one specific texture type-pair.
2. you can _envision_ this method being used for more than one specific texture types-pair. And I'm not talking about hypotheticals, I'm talking about something you have in mind but can't quite implement yet or simply doesn't make sense adding to this PR.

If either one of the above statements is true then I'd be ok with the current version.

However, if the only texture type-pair going through this code is normalsMap/specularMap AND you can't think of any other type-pair soon going through this code... I'd then feel strongly about tweaking those variable names. =)

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
emanuele3d: Ok, let's leave it as it.

Cervator:
Looks like conflict time now - probably something minor.

engine/src/main/java/org/terasology/world/block/tiles/WorldAtlasImpl.java
Josharias: Hrm.  Going over this again...  I suppose it feels right to keep it generic is because of how the entry method is set up at the moment.  createAtlasMipmaps(...) is shared between the color, height, and normal atlas textures.  So it would be odd to have the variables named very specifically instide of createAtlasMipmaps when it has come in through more generic means.

emanuele3d: Hmmm... so you are saying that because the calling method is generic the callED method should be generic even though it's used only to specifically store normals data and specular data together? ;)

Anyway, I looked at the rest of the WorldAtlasImpl code and it generated enough question marks orbiting my head already: one more is not going to make a difference.

Josharias:
A couple things improved now.

1. Propagated sunlight (aka cave darkness)  now functions again... Oops.
2. The sunlight debug screen works again... also oops.
3. The hacks around illuminating held items are fixed using the new MeshComponent.selfLuminance field.
4. MeshRenderer now more maintainable.

Known issues:
1. Transparent meshes do not seem to respect lighting.  There has got to be some more opengl trickery needed to fix this as the same code renders opaque meshes fine.

GooeyHub:

Cervator:
Tested it one more time, still looking good - merged!

Thanks for the nice work :-)

Should we put in a follow-up issue for the note about transparent meshes, and maybe a bitesize issue for updating existing "should be shiny" blocks with new shininess? Like ice? I take it we could start enhancing other blocks too like mineral blocks.

Issue #2271, opened 2016-04-01
No UI Animation System (Required for Terasology/Breathing#4 and #1639)


viveret:
There is not any UI animation system (to my knowledge) to support these features / fixes, which would be required by Terasology/Breathing#4 and #1639. What can we do about it?
I'm thinking something along the lines of [Android's method](http://developer.android.com/guide/topics/graphics/view-animation.html) of animation using a formatted text file (like JSON) to specify animation details.

msteiger:
Interesting suggestion. In my opinion, such a formal specification would be overdoing it, though.

I'd rather define a set of Java classes that define basic animations and put that in `engine`. Modules can use those out-of-the-box or define their own animations by overriding the `update(float)` method.

Such animators exist already for in-game entities: https://www.youtube.com/watch?v=0sAIA1eaUPQ
Code: https://github.com/Terasology/ItemRendering

I might be able to help working on the specifics.

Cervator:
Trying to work this out in my head - the health bar currently shows changes as the player gains / loses health. Is that just a UI element that's updating or an animation?

Once upon a time in legacy @small-jeeper had added a decoration where the toolbar would roll out (with some cogs on the side and everything!) into the full inventory. Was pretty neat but pre-NUI so it didn't get converted. Is that more of an animation than a simple update in the UI?

IIRC you can bind UI elements to a backing data element so they get updated automatically. I don't actually know off-hand how the health bar updates, either in Core or the hunger/thirst in Josharias Survival or Throughout the Ages.

Would animation be more for taking updated data (say 1-10 for health) then mapping each value to a specific state of a custom UI element like a paperdoll representation of the player's health? Sort of like you say specifying stages in a JSON file?

viveret:
There would be default time and value interpolators for traversing animations, and the json file would describe starting and ending conditions. 

The health bar currently just renders whole or half icon images without transition.

viveret:
Based off the ideas on [Android's Animating Resources](http://developer.android.com/guide/topics/resources/animation-resource.html)

FrameComponent (basically a component that changes):
- [ ] References to values or Objects to transition from / to

Frame:
- [ ] List of frames components

Base Animation Mutator (Transitions between 2 or more frames):
- [ ] From and To frames
- [ ] Time delay and duration
- [ ] Type of interpolation (linear, quadratic, exponential, logarithmic, square inverse...)
- [ ] Repeat (set number (once, many), infinite)
[...]

PR #2273, opened 2016-04-03
[Don't Merge, IP] UI Animation System (Required for Terasology/Breathing#4 and #1639) #2271

viveret:
UI Animation System (Required for Terasology/Breathing#4 and #1639), full-fills discussion at #2271

PR #2306, opened 2016-05-01
An animation system for NUI

msteiger:
This PR adds an animation system to NUI, based on #2273. This is what the final result looks like: https://youtu.be/Pk2jjDyDpnA

Most of the changes were fairly straight-forward. One larger refactoring was made on `NUIMangerImpl` to separate the creation process from pushing a screen properly. This was required to trigger initialization on button click and to avoid being garbage collected during the animation phase.


GooeyHub:

GooeyHub:

GooeyHub:

msteiger:
There are two outstanding issues: in-game overlays are not initialized and the pause menu does not show up. Let me know if you find anything else.

Cervator:
Tests out good! Couple items:

* Maybe we should add a graphics option to suppress menu animations? They are nice but a little slower to use than no animations + to the right person the fast flying menu might be uncomfortable.
* The `ESC` key for me in-game isn't showing the `pauseMenu` any.... yeah what you just said @msteiger :D

Cervator:
On a related note so we don't forget: this would fix #2271

Should we perhaps throw together a quick wiki page or tutorial on this and/or animation in general? Not necessarily before merging this, but we could put some rough details in a separate issue to be completed later.

GooeyHub:

msteiger:
All done! The PR for gestalt must be merged before Jenkins can build this PR.

The gimmicks from the video are not present in the code.

It would be great if @Cervator could check in the mega-workspace for compilation problems, because the API for NUIManager and HUD elements has changed a little.

Cervator:
Sure I can run the usual round of testing. We'd need the Gestalt PR both merged and an updated lib published + adjusted in our build.gradle here tho, right? 

Looks like other recent merges caused conflicts somewhere but probably minor.

viveret:
There could be a few different options for menu transitions:
- Swipe left / right
- Fade in / out
- No animation

Otherwise everything looks great! Glad I could help!

GooeyHub:

msteiger:
There's a checkbox in the video settings menu now (default: off).

For testing: yes, that's how it should work. We can wait until this is sorted out though and test it afterwards.

GooeyHub:

msteiger:
Ok - this is ready for testing! You need to enable animated menus in the video settings screen to see a difference.

Cervator:
Merged at long last!

@viveret thank you so much for getting this started and thank you @msteiger for adding some additional tweaking on top :-)

I fixed what was probably a small merge quirk over getting some translation keys duplicated.

Did also detect an API violation - oh noes! `protected abstract void initialise();` from `CoreHudWidget` was removed / moved in favor of `void initialise();` in `ControlWidget` which is an interface. Spot the trouble? Method now became public meaning old implementers were trying to constrain the level of access.

Four modules in the lineup were affected: EventualSkills, InGameHelp, Tasks, and WorldlyTooltip - fix was simply changing `protected` to `public` in one class each.

Considering how trivial that fix is and how easily I could personally apply it to every known affected module I wonder if it would make sense to have a "trivial clause" while we're still in Alpha and have a harder time detecting API violations. We should probably aim to create a module that attempts to test *every single API method* so we can attach that module build to the engine PR builder. That should do the trick, so long as we keep that module accurate somehow. I'll submit a separate issue.

I let that slide for now since we're still learning and improving our tooling, plus I haven't even technically announced v1.0.0 formally yet ... :blush:

PR #2310, opened 2016-05-02
Item pickup delay

portokaliu:
This PR adds item pickup time capabilities for the game. This has been discussed in PR #2289, and this is what I reduced it to.

I added _timeToPickUp_ to represent the time needed to pass from when the item was dropped and _pickUpTime_ as the time from when the counter starts.

I added the value changes in onDropItemRequest because I only wanted the **player's** item dropping to be affected by this change. Of course, any other drop item event listeners will be able to change the time needed to pick up the item by just getting the component, changing the two values, and saving.

I found this necessary because dropping items relatively below you or close enough to you would result in them being picked up by the character. This way, a player can throw items at his feed with no issues.

To test, get an item and throw it near your character. You shouldn't be able to pick it up for 2 seconds. On the other hand, items dropped from breaking blocks should be pickable instantly.

This can be only a beginning for the way item ownership and sharing items would work. Players will be able to throw items that they din't want to throw and stand on top of them so that nobody could take them until the pickup timer runs out

The only issue I have is the way I changed onBumpGiveItemToEntity. I'm not exactly sure what exactly was going on before (it looked like the item was given to the player even if there would be no itemComponent in it, which would be weird (for me at least)

### DIFF
--- a/modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
+++ b/modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
@@ -119,7 +119,6 @@ public class CharacterInventorySystem extends BaseComponentSystem {
         if (pickupItem.hasComponent(PickupComponent.class))
         {
             PickupComponent pickupComponent = pickupItem.getComponent(PickupComponent.class);
-            pickupComponent.timeToPickUp = 2000;
             pickupComponent.timeDropped = time.getGameTimeInMs();
             pickupItem.saveComponent(pickupComponent);
         }
--- a/modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
+++ b/modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
@@ -114,12 +114,16 @@ public class CharacterInventorySystem extends BaseComponentSystem {
             }
         }
 
-        ItemComponent itemComponent = pickupItem.getComponent(ItemComponent.class);
-        itemComponent.timeToPickUp = 2000;
-        itemComponent.pickUpTime = time.getGameTimeInMs();
-        pickupItem.saveComponent(itemComponent);
-
         pickupItem.send(new DropItemEvent(event.getNewPosition()));
+
+        if (pickupItem.hasComponent(PickupComponent.class))
+        {
+            PickupComponent pickupComponent = pickupItem.getComponent(PickupComponent.class);
+            pickupComponent.timeToPickUp = 2000;
+            pickupComponent.timeDropped = time.getGameTimeInMs();
+            pickupItem.saveComponent(pickupComponent);
+        }
+
         pickupItem.send(new ImpulseEvent(event.getImpulse()));
     }
 
--- a/modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
+++ b/modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
@@ -114,6 +114,11 @@ public class CharacterInventorySystem extends BaseComponentSystem {
             }
         }
 
+        ItemComponent itemComponent = pickupItem.getComponent(ItemComponent.class);
+        itemComponent.timeToPickUp = 2000;
+        itemComponent.pickUpTime = time.getGameTimeInMs();
+        pickupItem.saveComponent(itemComponent);
+
         pickupItem.send(new DropItemEvent(event.getNewPosition()));
         pickupItem.send(new ImpulseEvent(event.getImpulse()));
     }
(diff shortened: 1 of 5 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

@portokaliu: thanks for the cleanup! Pinging @flo and @Josharias for mini-review

GooeyHub:

Josharias:
I will look more thoroughly later on tonight,  on first glance it seems like the two new fields could be a good fit on the ```PickupComponent```.   Possibly filling in the ```timeToPickup``` and ```pickupTime``` after sending the ```DropItemEvent```?

The ```GiveItemEvent``` is intentionally fired even if it does not have an ```ItemComponent``` so that, in theory, other inventory/item systems could be used that do not rely on ```ItemComponent```.  In hindsight,  I wonder if ```GiveEntityEvent``` would have been a better name.

Josharias:
...5 hours later...  To me,  it still seems like a good idea to reduce the amount of things that ```ItemComponent``` is apart of.  Right now there is damage,  and stacking,  and usage, and cooldowns all on the same component.   I wonder if even another new component would be a good choice to store this protected pickup stuff... like a ```ProtectedPickupComponent```? (no,  dont use that name,  but you know what I mean).   I suppose maybe fleshing out ownership and sharing a bit more may give a good idea of the direction to head.

That said,  I do like the mechanic of allowing dropped items to not get sucked up quite so fast.

portokaliu:
There's already a PickupComponent used to recognize the item as one that can be picked up, and is used via the itemBase.prefab. If the item doesn't have an PickupComponent is it at all possible that it doesn't have an ItemComponent? Since onBumpGiveItemToEntity only accepts items that have a PickupComponent, I'm not sure there should even be an item component check, unless it is possible that the parameters are null (I know they shouldn't, but I'm still new to this)

Josharias:
It is possible for an item to not have a ItemComponent.  Although not presently in practice in the Core module.  I am not sure exactly what form of alternative inventory management will be developed,  but keeping the system specific to only the ```PickupComponent``` is a good step to keeping options open.

Nice idea with the "timeDropped",  it would help clarify what that field is for.

Josharias:
More thinking happened on Slack with @portokaliu... For another inventory management system to make use of these pickup events,  I suppose some more refactoring would be needed to further separate out the differences between preparing the item to be dropped (adding components that let it act in the world), bumping into the entity,  and un-preparing the item from having been dropped (removing the components that let it act in the world).  Right now, bump and un-prepare are mixed in together.

@portokaliu, likely keeping the fields inside of ```ItemComponent``` is not going to be horrible.  We can deal with that more when ```ItemComponent``` is broken down into smaller components and the bigger picture is more clear.

portokaliu:
I added the changes we talked about. One thing I found weird is that the item **can** have a pickupComponent while not having an itemComponent. Note to self - don't assume anything... like, ever. (maybe it catches the item a second time after its components have been stripped from it?)

GooeyHub:

modules/Core/src/main/java/org/terasology/logic/inventory/CharacterInventorySystem.java
```diff
@@ -114,12 +114,16 @@ public void onDropItemRequest(DropItemRequest event, EntityRef character) {
...
-        itemComponent.timeToPickUp = 2000;
-        itemComponent.pickUpTime = time.getGameTimeInMs();
-        pickupItem.saveComponent(itemComponent);
-
         pickupItem.send(new DropItemEvent(event.getNewPosition()));
+
+        if (pickupItem.hasComponent(PickupComponent.class))
+        {
+            PickupComponent pickupComponent = pickupItem.getComponent(PickupComponent.class);
+            pickupComponent.timeToPickUp = 2000;
```
Josharias: As one final thought about the hardcoded delay amount.  Could we let the itemPickup.prefab define the timeToPickUp?  That way it comes configurable out of the box. :)

Josharias:
There are definitely some unexpected scenarios that happen with entities.   It gets even weirder when in a remote client situation and you also have to deal with the temporary entities that get overwritten by the entities sent over the wire from the server.

This looks good to me.   There is one tweak that could be done,  but not necessary.  @portokaliu,  let me know if you are also satisfied with this PR and Ill merge it in.

GooeyHub:

Cervator:
Tested it out, works as advertised, at least in a plain old single player game :-)

Am also happy to merge. Should we detail the more quirky issues elsewhere? I figure that's (mostly) pre-existing vs. caused by this PR? And what related tweak anyway?

portokaliu:
I haven't tried getting a server up or anything yet, so I'm not exactly sure how it would fare in that environment. Might try it as soon as I look up how to do it :D
I'm also curious about the tweak. Anything that makes it better should be worth implementing now rather than later. I'm not in a real hurry to submit; It's not a critical task or anything :P

Josharias:
I double checked on a headless server with 2 connected clients and it functions as advertised.

Sorry for the confusion,  the tweak was putting the hardcoded value in the prefab.  Which you did.

Issue #2309, opened 2016-05-01
No i18n support for Settings → Input


rzats:
Classes from the [org.terasology.rendering.nui.layers.mainMenu.inputSettings](https://github.com/MovingBlocks/Terasology/tree/develop/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings) package currently use hardcoded English strings to display UI elements. These need to be replaced with internationalizable keys and translated, if possible.
##

Cervator:
Cool, thanks for volunteering for this one :-)

Related: #2226, #2243,  #2285, #2207, #2028, maybe #2145 - if any of those are easy to throw in feel free, but no obligation!

rzats:
@Cervator: turning this into a Grand I18n Fix is fine by me (:
Although some of the issues you've mentioned have pending PRs open - what's the best way to handle this?

Cervator:
I included the two PRs for that reason - they're essentially issues with attached code at this point as it seems the original authors are out of touch or time at the moment :(

I would try to merge/rebase them on top of the latest develop before/after you make some of your desired changes, maybe in a separate PR. Then adjust as needed to finalize.

If you simply merge the commits the older PRs should auto-close as merged if you make a newer PR containing the merges + your adjustments. Alternatively if they are hitting too severe conflicts at this point (mainly meaning the one that splits the `.lang` files some) then limited copy-paste or redo + credit in the commit message should be alright.

flo:
@rzats thanks for this PR. I had a look at it although there isn't really much "logic" for review in it. @Cervator I will leave it up to you to merge it as you seem to have a good overview what should be merged in what order.

Cervator:
Bit of an update here since this seems the most relevant open issue :-)

PR #2404 has been merged and updated a few drop-downs/enums to respect the translation system, introducing a new utility class.

Based on a comment by @msteiger it could still be enhanced a bit more:

> I'd prefer to have two separate TextRenderer instances for the two use cases: as-is/translated. The ToStringTextRenderer could remain untouched and a new TranslatingTextRenderer could perform the translation work.

The object of this issue is very similar to that PR since it points out a bunch more UI elements currently hard coded despite translations being available (at least in some cases I think). @Margretor you had a bit more covered in your PR #2226 - want to grab latest code and add translation for a few more? Alternatively @indianajohn if you feel like a round two go for it, just comment here so we know who is working on it :-)

[...]

PR #2313, opened 2016-05-05
Input settings menu i18n; misc. i18n fixes

rzats:
### Contains

An implementation for #2309. Adds internationalization support to static strings, categories and binding names in the Input Settings menu - if a category/binding name is a URN to an internationalisable string, it will be translated, otherwise it will fall back to the default (presumably English) string value

Other miscellaneous additions:

- Sorted the language files by key (via https://www.npmjs.com/package/sort-json). Doesn't break anything but makes the diff look spookier than it actually is.
- Fixed the mess in the Galician file - it used some empty string values, causing awkward UI issues. Not that this particular translation was ever usable :confused:

### How to test

Go to Settings → Player, switch the language to Ukrainian, return to Settings → Input (the fourth button from the top :wink:) and check out the translated menu!

Translated menu screenshot

- [ ] The Inventory category has some strings specific to the Core module. If these are to be translated, they will require a new set of .lang files and a separate Weblate subproject.


GooeyHub:

Cervator:
Cool points for "Pyx" - that's a nice looking word! :-)

I see the `API` label there, is there anything changing the modding API potentially breaking modules or is it just a "seemed fitting" for the PR's content?

Also for whatever reason the PR is marked as having conflicts. Maybe missing a rebase on top of the latest develop branch?

GooeyHub:

rzats:
The `API` label refers to this, which is technically a binding API change:

> if a category/binding name is a URN to an internationalisable string, it will be translated, otherwise it will fall back to the default (presumably English) string value

Fixed the merge conflicts - these were caused by the extra #2301 string.

msteiger:
Modules can inject custom key bindings into the settings dialog. Like you say, the Core bindings should reside in new Weblate subproject.

The JSON keys in the files should be ordered in the same way that Weblate does to avoid confusion. I think that this is alphabetically, so this should be correct. You could add trailing new-line chars, but again, Weblate might override it.

Looks good to me. Thanks you, @rzats.

rzats:
@msteiger: is there any extra setup required to add new .lang files? I've created a Core/Assets/i18n directory and placed some .lang files with valid strings in there, but they aren't picked up by `TranslationSystemImpl`.
A module that has its own translation files might help as an example, although I'm not aware of any that exist.

msteiger:
The translation system iterates over all known translation assets: [TranslationSystemImpl#L69-L84](https://github.com/MovingBlocks/Terasology/blob/8358d38f2847e6faa3884a29eb276a0ab89d0628/engine/src/main/java/org/terasology/i18n/TranslationSystemImpl.java#L69-L84)
I would assume that the asset files are not loaded.

They should be named like this: ` <project-name>_<language-tag>.lang`.

Then, the `TranslationFormat` asset file reader should create a translation project from it: [TranslationFormat:L85-L107](https://github.com/MovingBlocks/Terasology/blob/8358d38f2847e6faa3884a29eb276a0ab89d0628/engine/src/main/java/org/terasology/i18n/assets/TranslationFormat.java#L85-L107). Does that help?

rzats:
@msteiger: well, yes and no - after some debugging it looks like the asset files are never loaded in either `TranslationSystem` or `TranslationFormat.load()`. Their names don't appear to matter: neither of _Core.lang_, _Core_en.lang_, _core.lang_ or _core_en.lang_ work.

This most likely has something to do with the files being in Core instead of engine, requiring some additional hooks. Pinging @immortius.

Cervator:
Weird, looks like another one of my messages never made it through. I must be writing posts then getting distracted by another browser tab and then never make it back :-)

@rzats thanks for the API description! That makes sense.

Should we go ahead and merge this then worry about Core later? That is the start to a very deep rabbit hole, with or without asset scanning quirks :-)

msteiger:
I agree - we should look into module i18n separately.

Issue #1773, opened 2015-06-13
Open up the DebugOverlay to add/remove MetricsModes


emanuele3d:
### Introduction ###
I was recently experimenting with a new MetricsMode for the DebugOverlay. I was aiming for a generic solution where objects for the retrieval, filtering, sorting and formatting of the data could be provided to the mode upon construction or even at runtime. From an architecture/coding perspective this wasn't difficult.

In practice, the setup of such objects was becoming an issue. Right now modes are quickly instantiated with a short "new MyMode()" instruction within a list initialization in the DebugOverlay class. Conversely, setting up a more generic mode in the worst case scenario would require something like:
```java
GenericMetricsMode myMode = new GenericMetricsMode();
myMode.setDataRetriever(new ExecutionTimeRunningMeansRetriever());
myMode.setDataFilter(new StartsWithFilter("MyString::"));
myMode.setDataSorter(new AscendingAlphabeticalOrder);
myMode.setDataFormatter(new CustomFormatter());
```
Data-processing objects could be reused by multiple modes, but would also require additional lines to instantiate them. Permutations in the configuration of modes would have to be dealt with by instantiating and adding new modes, i.e. the same code above with a DescendingAlphabeticalOrder instead. Eventually part of the DataOverlay class would be peppered with this stuff.

Incidentally, this also got me thinking that if a module wanted to add a custom mode at runtime, it can't.

### Proposal ###
I would therefore suggest for the DebugOverlay class to provide methods to dynamically register and unregister objects implementing MetricsMode, so they can be added to the list of modes to be cycled over. This would allow modules to add their own modes. Furthermore, mode-configuration code would stay out of the DebugOverlay class and remain a responsibility of whatever code registers the a mode. Specifically I would add the following methods:
```java
public Object register(MetricsMode mode);
public void unregister(Object receipt);
```
Notice the use of a simple receipt object, returned by the register() method, to unregister a mode. I'm not sure if this is the best way to go about it but the idea is that only the code that has registered a mode or code that has been given the receipt can unregister a mode. This is a pattern I've seen in the OSGI framework to register services, but perhaps there is a better way using the engine's security features?

### Runtime Re-configuration? ###
[...]

PR #2316, opened 2016-05-11
[Ready for Review] Open up the DebugOverlay to add/remove MetricsModes

tdgunes:
### Contains

A partial fix for #1773, can be thought as a start too. As also mentioned in #1773, current usage of MetricModes in [DebugOverlay](https://github.com/MovingBlocks/Terasology/blob/8358d38f2847e6faa3884a29eb276a0ab89d0628/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java) does not allow runtime modifications and possibility of implementing data-processing objects which could be reused in other modes. 

At the moment, PR adds new methods for _DebugOverlay_ class:

```java
public void register(MetricsMode mode);
public void unregister(MetricsMode mode); 
``` 

Moves default metrics to abstract _MetricsMode_ class:

```java
static void setDefaultMetrics(DebugOverlay debugOverlay) {
        debugOverlay.register(new RunningMeansMode());
        ...
        debugOverlay.register(new NetworkStatsMode());
}
``` 

In order to store _MetricsMode_ objects in _DebugOverlay_, PR includes _ListOrderedSet_ as a new collection which acts like a list and a unique set. Therefore, items are ordered and unique at the same time. It's based on Java's _HashMap_ and _ArrayList_ implementations. Name and some of the implementation aspects are inspired from Apache's [ListOrderedSet](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/set/ListOrderedSet.html). It's important to point out that this collection might be not the best choice in this scenario. It would be great to have some feedbacks especially on this.  

### How to test

- For testing the collection, PR includes _ListOrderedSetTest_. 
- For checking whether PR broke current functionality of _DebugOverlay_, during in game, press F3 and then toggle metrics by F4.


GooeyHub:

engine-tests/src/test/java/org/terasology/utilities/collection/ListOrderedSetTest.java
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.utilities.collection;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class ListOrderedSetTest {
```
emanuele3d: @msteiger: Martin, can you double-check all is well in this class? I have no reason to believe otherwise but I'm not up to speed with Unit testing goals and standards.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
```diff
@@ -168,6 +168,11 @@ public Boolean get() {
     @Override
     public void update(float delta) {
         if (metricsLabel != null) {
+            if (metricsModes.isEmpty()) {
+                metricsLabel.setText("No metric modes are available!");
+                return;
+            }
+
             metricsLabel.setText(metricsModes.get(currentMode).getMetrics());
```
emanuele3d: I think the typical case is that the metricsLabel is not empty. So, let's put that in the first part of the "if" block and let's handle the metricsMode.isEmpty() case with an "else" block.

Also, are there situations in which the metricsLabel is null?

tdgunes: 👍  for "else" block.

For metricsLabel, I need to double check. But I think that check is for:
```java
metricsLabel = find("metrics", UILabel.class);
```
If _find_ fails, most probably it returns null.

emanuele3d:
@tdgunes, thank you for looking into this. I only had the chance to have a brief look into this. I'll go more in-depth over the next 48 hours.

Something that is certainly missing is a healthy dose of javadocs. They are not necessary for the test class and the ListOrderedSet just needs a class-level javadocs, the methods are self-explanatory. 

However, as you probably had to look in-depth into what DebugOverlay and MetricsMode do, I'd encourage you to add class-level and method-level javadocs to them.

One thing I'm not convinced about is your somewhat special handling of the "DefaultMetrics". It seems to make things more complicated this way. I'd be happy for their instantiation to occur in the initialize() function, but I don't quite see the advantage in wrapping them in MetricModes.setDefaultMetrics(). I do like the use of the register() method to do it though.

By the way could you have used a LinkedHashSet instead of a custom class?

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
emanuele3d: I guess my concern here might be a life-cycle issue. Is a DebugOverlay instance and its labels alive at all times? Ideally it shouldn't be, only when the user presses F3 it should come alive. And are its labels instantiated only when there is content to put into them?

msteiger:
The code looks good to me in general. A few notes on the details:

I'd also go for a `LinkedHashSet`, if possible. Afaics, the index-based access  is not required, since the  modes are just cycled continuously. Maybe `Iterators.cycle()` can be useful here?

I don't like the idea of passing `this` from the `DebugOverlay` to the different metrics so they can register. How about this: A new `DebugRenderingSystem` that is annotated with `@Share` and the debug overlay just renders its content? Modules and engine can then use the `@In` annotation to inject this system and register with it. This would be very flexible, but you need to create the metric instances through other means - maybe `@RegisterSystem` and/or implement `RenderSystem`.

engine-tests/src/test/java/org/terasology/utilities/collection/ListOrderedSetTest.java
msteiger: Looks good to me, thanks!

emanuele3d:
@msteiger: Whoa. As soon as somebody pops decorators all over the place I start feeling dizzy. =)

So, `@Share`-ing a class is what allows other classes to get an instance of it via the `@In` decorator of a member variable?

And what do you think of the previous way of doing this, in which metrics were instantiated directly in DebugOverlay? That wouldn't require passing the DebugOverlay instance anywhere.

Cervator:
> So, @Share-ing a class is what allows other classes to get an instance of it via the @In decorator of a member variable?

Correct :-) `@RegisterSystem` tells the engine about it, `@Share` tells the engine to tell *others* about it, if they care to `@In` it - at least that's my understanding

GooeyHub:

tdgunes:
> @manu3d: One thing I'm not convinced about is your somewhat special handling of the "DefaultMetrics". It seems to make things more complicated this way.
> @msteiger: A new DebugRenderingSystem that is annotated with @Share and the debug overlay just renders its content? 

I had no experience on annotations so far in Java. It seems very handy. I created a DebugRenderingSystem class that registers and shares itself. It would be possible for any third party developer to modify debugging metric modules. 

> @manu3d: By the way could you have used a LinkedHashSet instead of a custom class?

I wasn't sure about adding `ListOrderedSet`. Yeah, seems like we don't need index-based access. My last commit replaces and removes `ListOrderedSet` with `LinkedHashSet`. 

One thing that I am not sure is the life-cycle of ui items and metric modules and totally adding a very custom ui item for displaying for instance a graph.

Cervator:
@msteiger & @emanuele3d are the two of you OK with the code now to where we can merge or do we need more tweaks? :-)

emanuele3d:
@Cervator:  no sorry, I need some more time to review it.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
```diff
@@ -37,7 +36,6 @@
 import org.terasology.world.biomes.Biome;
 import org.terasology.world.biomes.BiomeManager;
 
-import java.util.List;
 import java.util.Locale;
 
 /**
```
emanuele3d: Please add class-level javadoc. In a short paragraph somebody looking at this class should immediately understand:
* what is the purpose of the DebugOverlay
* what fixed information it provides and in what format
* how the dynamic information is generated and handled, i.e. what methods manipulate the dynamic portion of the overlay

I agree this information should have been there in the first place, written by the original author of the class.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugRenderingSystem.java
```diff
@@ -0,0 +1,78 @@
...
+        register(new RunningMeansMode());
+        register(new SpikesMode());
+        register(new AllocationsMode());
+        register(new RunningThreadsMode());
+        register(new WorldRendererMode());
+        register(new NetworkStatsMode());
+        register(new RenderingExecTimeMeansMode("Rendering - Execution Time: Running Means - Sorted Alphabetically"));
+    }
+
+    public MetricsMode register(MetricsMode mode) {
```
emanuele3d: I had a hunch that something wasn't quite right with this class but I couldn't quite point the finger to it. I now realize that issue #1773 recommends to add register() and unregister() methods to the DebugOverlay class. That's because DebugOverlay is not intended to be rendering-specific. Systems or even modules might wish to register their own MetricsMode with it, completely unrelated to rendering.

This is also in line with the NetworkStatsMode() above not being quite rendering-related. It is true that at this stage the DebugOverlay is mostly used for rendering but it's not intended to be exclusive for it.

So, I recommend to remove this class and move its functionality into DebugOverlay. I.e. I like the switch to a LinkedHashSet and the use of register() to register the default Metrics. 

Eventually we should move the registration of the default rendering metrics into the WorldRenderer initialization.

msteiger: I believe that this class has a right to exist. In order to register a `MetricsMode`, the system can be injected from any module. I think that the `DebugOverlay` should only display info. on user interface.

A similar approach can be found for the console. There's `ConsoleScreen` for rendering and `ConsoleSystem` that contains the actual content.

If the default metrics go into the world rendering init. code, we need to make sure that this is called afterwards.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
```diff
@@ -167,8 +163,10 @@ public Boolean get() {
 
     @Override
     public void update(float delta) {
-        if (metricsLabel != null) {
-            metricsLabel.setText(metricsModes.get(currentMode).getMetrics());
+        if (debugRendering.isMetricModesAvailable()) {
+            metricsLabel.setText(debugRendering.getCurrentMode().getMetrics());
+        } else {
+            metricsLabel.setText("No metric modes are available!");
```
msteiger: Instead of explicitly checking whether the list is empty of not, you could add a default dummy mode and guarantee that `getCurrentMode` is never `null`.

emanuele3d: In fact, isn't that what NullMetricsMode is for?

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugRenderingSystem.java
emanuele3d: Ok, sounds like a separation of presentation/content issue. You'd like this class take care of the content (avilable through the MetricModes instances) and then DebugOverlay takes care of presenting the data on screen. 

Ok, I understand the approach and I'm fine with it. I would then recommend to rename the DebugRenderingSystem class to DebugMetricsSystem, so that it is not rendering-specific and is clearly associated with the concept of Debug Metrics.

Regarding the default metrics, I'd like to be clear that only the rendering-related metrics would be instantiated and registered during the rendering engine initialization. The DebugMetricsSystem by then should be up and running already. And on shutdown the rendering engine should make sure to unregister all its metrics. But this is out of the scope of this PR. I'm happy to leave those things as they are for the time being.

Finally, as this class is bound to stay, it will need some javadocs then. Class-level AND non-overriding public methods pleaaaaaaase!!! =)

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+import org.terasology.entitySystem.systems.BaseComponentSystem;
+import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.registry.Share;
+
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+
+/**
+ * Provides MetricsMode objects to DebugOverlay. Default metrics nodes can be modified via register
+ * unregister and unregisterAll functions.
```
emanuele3d: Good start. I'd write:

/**
Manages an ordered set of MetricsMode instances. 

A number of default metrics modes is instantiated in the initialize() method and becomes immediately available by iterating through the set repeatedly invoking the toggle() method. The getCurrentMode() method can also be used to obtain the mode currently pointed at. Further modes can be added to the set via the register method, while registered modes can be removed via the unregister method. A convenience method unregisterAll() removes all registered modes at once.

MetricsModes are currently used by the DebugOverlay instance, displaying runtime statistics on screen.
*/

Notice that I'm being extra-verbose because eventually these methods will become part of the API, available to other parts of the engine and third-party modules. In fact, we might want to do it right away. How does this work @Cervator? How do we register these methods with the API? Any quirk we should know?

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+public class DebugMetricsSystem extends BaseComponentSystem {
+
+    private LinkedHashSet<MetricsMode> metricsModes;
+    private Iterator<MetricsMode> modeIterator;
+    private MetricsMode currentMode;
+
+    @Override
+    public void initialise() {
+        metricsModes = Sets.newLinkedHashSet();
+        modeIterator = Iterators.cycle(metricsModes);
```
emanuele3d: Please double check all methods that modify the LinkedHashSet. From what I read the iterator might need to be regenerated every time something is added or removed from the set. If that is the case it might be appropriate to have additional register(ArrayList) and unregister(ArrayList) methods to pool changes together and regenerate the iterator only once.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+        register(new RunningMeansMode());
+        register(new SpikesMode());
+        register(new AllocationsMode());
+        register(new RunningThreadsMode());
+        register(new WorldRendererMode());
+        register(new RenderingExecTimeMeansMode("Rendering - Execution Time: Running Means - Sorted Alphabetically"));
+    }
+
+    /**
+     * Registers a metrics mode, can be viewed via toggling {@link DebugMetricsSystem#toggle()}
```
emanuele3d: I'd write:

/**
Adds a MetricsMode instance to the set. Use the toggle() and getCurrentMode() methods to iterate over the set and obtain the MetricsMode instances.

@param a MetricsMode instance
*/

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @param mode
+     * @return
+     */
+    public MetricsMode register(MetricsMode mode) {
+        metricsModes.add(mode);
+        return mode;
```
emanuele3d: Can this method return void instead? I feel it's a bit funny to return the same thing we just provided as input.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /***
+     * Returns current mode, initializes a default {@link NullMetricsMode}, if currentMode is null
```
emanuele3d: I'd write:

/** (notice the 2 asterisks rather than 3)
* Obtains the MetricsMode instance currently pointed at by the internal iterator. If no current mode exists, i.e. because the set is empty, registers and returns an instance of NullMetricsMode.
*
* @return a MetricsMode instance
*/

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+            currentMode = register(new NullMetricsMode());
+        }
+        return currentMode;
+    }
+
+    /**
+     * Toggles to next mode
+     *
+     * @return
+     */
```
emanuele3d: In line with the code I suggest below, I'd write:

/** 
* Iterates through the set of registered MetricsMode and returns the first instance whose method isAvailable() returns true. Notice that this could be the MetricsMode that was current at the time this method was invoked.
*
* @returns a MetricsMode instance
*/

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+    public MetricsMode toggle() {
+        do {
+            currentMode = modeIterator.next();
+        } while (!getCurrentMode().isAvailable());
+        return this.getCurrentMode();
+    }
```
emanuele3d: I am concerned about this method. Isn't an exception generated if next() is called and the set is empty? Also, I see an unlikely but possible infinite loop if none of the MetricsMode.isAvailable() methods return true.

I wonder if a less compact but somewhat easier to read method could work:

```
public MetricsMode toggle() {
    MetricsMode startingMode = getCurrentMode();
    while(modeIterator.hasNext()) {
        currentMode = modeIterator.next();
        if(currentMode == startingMode || currentMode.isAvailable()) {
            break;
        }
    }
    return currentMode;
}
```

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Removes given metric mode, makes sure currentMode is updated
+     * @param mode
+     */
```
emanuele3d: I'd write:

/**
Removes from the set the MetricsMode instance provided as input. 

If the MetricsMode instance is the mode currently pointed at by the iterator, toggles the iterator forward.

@return True if the MetricsMode instance was in the set. False otherwise.
*/

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+    public void unregister(MetricsMode mode) {
+        if (currentMode == mode)
+            currentMode = toggle();
+
+        metricsModes.remove(mode);
+    }
```
emanuele3d: In line with the javadoc suggested above, I'd let this method return a boolean:

`return metricsModes.remove(mode);`

Also, in line with the considerations I made about the toggle() method, if toggle cycles back to the starting mode the variable currentMode will contain a MetricsMode instance that has been removed from the set. Perhaps we need to rethink toggle a bit.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Removes all registered metrics modes
+     */
+    public void unregisterAll() {
+        metricsModes.clear();
```
emanuele3d: As mentioned earlier, these calls might need to also regenerate the iterator. Please double-check.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
```diff
@@ -37,10 +36,12 @@
 import org.terasology.world.biomes.Biome;
 import org.terasology.world.biomes.BiomeManager;
 
-import java.util.List;
 import java.util.Locale;
 
 /**
+ * Displays and changes metrics modes, according to incoming actions from DebugControlSystem.
+ *
+ * DebugMetricsSystem provides available metrics modes to DebugOverlay.
```
emanuele3d: I'd write:

/**
Displays the content of the MetricsMode instances provided by the DebugMetricsSystem. 

See the toggleMetricsMode() method to iterate through the MetricsMode instances available for display.
*/

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
```diff
@@ -183,10 +180,9 @@ public boolean isEscapeToCloseAllowed() {
     }
```
emanuele3d: /**
Move forward through the MetricsMode instances and displays the content of the next available one.
*/

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
tdgunes: @emanuele3d thanks for your detailed feedback. I have modified java docs as you requested. I am a little skeptical about writing extra-verbose comments. But since this will be part of API, I am ok with it.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
tdgunes: Sorry for missing this. Wouldn't regenerating iterator object after each modification be costly? I tried some manual approaches, but I couldn't find a clean and simple solution. 

Also, what do you mean by pooling changes with `register(ArrayList) ...` ? It would be great if you could give an example.

msteiger: Regenerating a simple object such as the cycling iterator is almost free. So no need to worry about that. In general, this must be done as soon as the collection changes. An interesting alternative could be `CopyOnWriteArrayList`, which supports concurrent access (changing content while iterating).

You would need to manually check if an object already exists, since it is not a set.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
msteiger: Readability is the most important thing here, I agree. My preference is the proposed `do..while` loop, but I'm ok with both solutions.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
msteiger: Or maybe a boolean whether the mode was added or not? This would be inline with the Collection interface, but `void` is ok, too. It should be consistent with unregister, though.

GooeyHub:

tdgunes:
I might be doing some "premature optimisation" unintentionally, in this issue. Nevertheless, my latest commit ensures, that given metrics modes will be unique and insert-ordered with default metrics mode as NullMetricsMode(). I couldn't find a proper data structure to represent this particular problem. `LinkedHashSet` was a good candidate, however generating a new iterator and resetting toggle-order after every modification literally irritated me. I know that proposing a new data structure without bugs is not easy. Therefore, I wrote unit tests for both `DebugMetricsSystem` and `CircularToggleSet` and tested in various scenarios. It seems to be working quite alright. But if there is a missing corner case, it would be great if you can extend it.  

By the way, it's also important to mention that DebugMetricsSystem is not thread-safe. `register` and `unregister` operations are not done atomically. Since DebugMetricsSystem will be used for debugging purposes only, I believe it wouldn't hurt to leave this as is.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,120 @@
...
+     * Iterates through the set of registered MetricsMode and returns the first instance whose method isAvailable()
+     * returns true. Notice that this could be the MetricsMode that was current at the time this method was invoked.
+     * @returns a MetricsMode instance
+     */
+    public MetricsMode toggle() {
+        do {
+            currentMode = modes.toggle();
+        } while (!currentMode.isAvailable());
+
+        return currentMode;
```
emanuele3d: This method continues to concern me. The risk I see is an infinite loop. I do not know why isAvailable() should ever return false. The MetricsMode interface doesn't document it and the implementations I've seen all return true if I remember correctly. 

However, it is conceivable that a MetricsMode instance might change from available to unavailable at runtime. If this method is invoked when no MetricsMode are available, it will loop forever.

I can think of two solutions:
1. if the loop gets back to the beginning, stop it.
2. if the loops gets back to the beginning, inject a NullMetricsMode instance (isAvailable always returns true), make it the currentMode and return it.

tdgunes: While testing DebugMetricsSystem, I noticed that WorldRendererMode in test mode, isAvailable() returns false. But by providing NullMetricsMode instance every time as a default even if it's cleared (it's always returns isAvailable() true), I believe it will not loop forever.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Removes from the set the MetricsMode instance provided as input.
+     *
+     * If the MetricsMode instance is the mode currently pointed at by the iterator, toggles the iterator forward.
+     *
+     * @return True if the MetricsMode instance was in the set. False otherwise.
+     */
+    public boolean unregister(MetricsMode mode) {
+        if (currentMode == mode) {
+            toggle();
```
emanuele3d: As described above in my comments about the toggle() method, it is conceivable that toggle() might loop all the way to the mode it started with if all other modes are in the list but are not available. This method would then remove the mode from the list, but currentMode would still be set to the mode that has just been removed.

An easy solution I see is to start the method with

if(modeToBeRemoved == defaultMode) {
    <throw exception>
}

This way the NullMetricsMode instance is guaranteed to always be in the list of MetricsMode and toggle() will, in the worst case scenario, loop back onto that instance and stop. Incidentally this approach would make toggle() easier because it wouldn't have to inject a NullMetricsMode instance as suggested above.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        return modes.remove(mode);
+    }
+
+    /**
+     * Removes all registered metrics modes
+     */
```
emanuele3d: I'd add to the comment:

"except an instance of NullMetricsMode, which is guaranteed to be always available."

I guess this is a detail that would be good to add to the class-level javadoc: no matter what a NullMetricsMode instance is always available somewhere in the list.

engine/src/main/java/org/terasology/utilities/collection/CircularToggleSet.java
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * CircularToggleSet is a "toggleable" doubly-linked list, provides uniqueness with the help of HashMap.
+ * <p>
+ * CircularToggleSet's {@link #toggle()} provides iteration through out given elements. Whenever, {@link #toggle()} is called
+ * first-inserted element is returned. Therefore, iteration starts from beginning again.
+ *
+ * @param <T>
+ */
+public class CircularToggleSet<T> {
```
emanuele3d: I don't understand why you added this class. What was wrong with the LinkedHashSet? You needed to regenerate its iterator when modified, but as @msteiger mentioned that's not an expensive operation and the modifications do not occur in hot code anyway. Am I missing something?

emanuele3d: Sorry, I hadn't read your explanation above. I'll answer there.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
tdgunes: Can you given an example showing this scenario? I might have misunderstood something here.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
emanuele3d: Ah, thank you for letting me know about the one case in which isAvailable() return false. At some point it might reveal what isAvailable() is for. =)

Regarding NullMetricsMode I think we are on the same page. I just think that it can be removed by the unregister() method and if confirmed that's a hole that should be plugged.

engine/src/main/java/org/terasology/utilities/collection/CircularToggleSet.java
tdgunes: No problem, it became a little messy around here. Sorry for that.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
tdgunes: I think that is not possible. Assume that user wants to delete defaultMode with `unregister` method.  A possible scenario, I think:

Generating a new instance: 
``` java
NullMetricsMode mode = new NullMetricsMode();
system.unregister(mode);
```
This behaviour will not remove NullMetricsMode. Because mode's pointer != defaultMode. Also `defaultMode` is private for not enabling this behaviour. 

Do you have another scenario that I might be missing?

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
emanuele3d: Mmmm... now that I think about it, if a NullMetricsMode instance is always available (*) the only case would be if the NullMetricsMode instance is passed-in as input AND there are no other available MetricsMode instances.

In that case toggle() loops back onto the NullMetricsMode instance and this method removes it from the list AND leaves currentMode pointing to it.

(*) when I wrote the comment above I hadn't seen the unregisterAll() method puts NullMetricsMode back in after clearing - lesson learned for me: make sure to consider aaaaaall the code.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
emanuele3d: defaultMode can be obtained via getCurrentMode().

tdgunes: Oh, that's a really bad corner case. Thanks. I will be fixing this real quick.

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
emanuele3d: Yes, absolutely, I have no idea why anybody would like to remove the NullMetricsMode instance actually. By mistake though, it might happen.

emanuele3d:
I can't seem to find a way to reply immediately after your comment containing:

> LinkedHashSet was a good candidate, however generating a new iterator and resetting toggle-order after every modification seems not the best way to do it. I know that proposing a new data structure without bugs is not easy. Therefore, I wrote unit tests for both DebugMetricsSystem and CircularToggleSet and tested in various scenarios. 

So, I'll reply here. 

I'd like to ask @msteiger his opinion about this. Originally the structure holding the MetricsMode instances was a simple list. Indeed using a plain list did not prevent registering the same instance multiple times, but the register() method could easily enforce that, by iterating over the list before something is added to it. This assumes that the list will never have enough items to run into performance issues - a safe assumption I believe.

In this context at this stage I do not see the utility of the new CircularToggleSet class. In fact, if register() enforced uniqueness, I don't even see the utility of a LinkedHashSet: the likely list size is too small to benefit from a hashed structure. It's late though. I might be missing something obvious.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
tdgunes: How about null checks to `register` and `unregister`? Should we also consider the case where a user might try to register an instance, but mistakenly tries to register something null?

tdgunes:
`CircularToggleSet` class' advantage over 'LinkedHashSet' is preserving the position of the iterator, even if there is a modification during iteration. It provides same other properties as 'LinkedHashSet': uniqueness and insert-ordered. With `CircularToggleSet`, I tried to make data structure have all these previously mentioned properties and efficient at the same time.  It's also true that a simple List might suffice for resolving our dilemma here.

msteiger:
Using a `CopyOnWriteArrayList` should allow you to reuse the iterator, preserving the position.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
msteiger: Checking for `null` arguments makes sense, imho. You could split the metrics modes into two collections: core types that cannot be removed and addons. You can create a single view on both with `Iterables.concat(first, second)`. Register and unregister would work on the addons list only.

Alternatively, you could explicitly check in `unregister` if the passed mode is the `NullMetricsMode` or in a list of core types.

tdgunes:
> From java documentation: The "snapshot" style iterator method uses a reference to the state of the array at the point that the iterator was created. This array never changes during the lifetime of the iterator, so interference is impossible and the iterator is guaranteed not to throw ConcurrentModificationException. The iterator will not reflect **additions, removals, or changes to the list** since the iterator was created.

@msteiger: It seems ConcurrentModificationException will not thrown, but additions and removals after creation of iterator will not be visible or is there something that I am missing?

GooeyHub:

msteiger:
That is correct. The `Iterators.cycle()` will request a new iterator when a cycle is complete, though. Thus, the changes will be visible in the next cycle.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -95,8 +98,9 @@ public MetricsMode toggle() {
      * @return True if the MetricsMode instance was in the set. False otherwise.
      */
```
emanuele3d: Please update the javadoc with the exceptions this method might throw.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -95,8 +98,9 @@ public MetricsMode toggle() {
...
     public boolean unregister(MetricsMode mode) {
+        Preconditions.checkNotNull(mode, "Illegal argument passed, mode is null.");
```
emanuele3d: Please update the javadoc with the exceptions this method might throw.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -95,8 +98,9 @@ public MetricsMode toggle() {
...
         if (mode == defaultMode) {
-            throw new IllegalArgumentException("Removing defaultMode is not allowed!");
+            throw new IllegalArgumentException("Removing defaultMode is not allowed.");
```
emanuele3d: @Cervator: this is the exception I was asking you about, but I've come to the conclusion it is the right thing to do, so I won't ask confirmation after all.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -62,6 +64,7 @@ public void initialise() {
      * @param mode a MetricsMode instance
      */
     public boolean register(MetricsMode mode) {
+        Preconditions.checkNotNull(mode, "Illegal argument passed, mode is null.");
         return modes.add(mode);
     }
```
emanuele3d: From the getCurrentMode() javadoc:

> Returns current mode, initializes a default {@link NullMetricsMode}, if currentMode is null

The second part of the sentence is no longer true. I'd simply write:

`* @return the current MetricsMode instance`

nothing else.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -121,4 +125,35 @@ public void unregisterAll() {
     public int getNumberOfModes() {
         return modes.size();
     }
+
+
+    private class CircularToggleSet<T> {
```
emanuele3d: I still don't know why you didn't simply use an ArrayList straight into DebugMetricsSystem rather than wrap it into a private container class. It's not like the DMS is a particularly complicated piece of code. =) But I think we have argued enough over this PR and I'm ok leaving this as it is just because you like it this way. =)

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -95,8 +98,9 @@ public MetricsMode toggle() {
      * @return True if the MetricsMode instance was in the set. False otherwise.
```
emanuele3d: Above, in the toggle() method, I'd add the following comment line, directed to fellow developers, between the end of the while loop and the return statement:

// This loop will always find at least one available MetricsMode instance, the defaultMode, 
// hence never running indefinitely. 

It is possible of course to arrive to that conclusion by looking at the rest of the code, but I figure we can make things easier to the techno-archeologist who will dig up this piece of code 1378 yeas from now. =)

emanuele3d:
@tdgunes: I've added a number of comments but somehow they are listed as comments to an outdated diff. Please make sure to read them! =)

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,137 @@
...
+
+        if (!modes.contains(mode)) {
+            return modes.add(mode);
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Returns current mode, initializes a default {@link NullMetricsMode}, if currentMode is null
```
emanuele3d: This javadoc is no longer valid. This method does not initialize a default NullMetricsMode. Just leave the @return line. Also, you might want to squash this and the next commit.

tdgunes: Oh, how did I miss this? Definitely, seems like I need to focus more on java docs, in my future commits.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugMetricsSystem.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Removes from the set the MetricsMode instance provided as input.
+     *
+     * If the MetricsMode instance is the mode currently pointed at by the iterator, toggles the iterator forward. Unregistering
+     * defaultMode is not allowed, therefore throws {@link IllegalArgumentException}.
+     *
+     * @return True if the MetricsMode instance was in the set. False otherwise.
+     */
+    public boolean unregister(MetricsMode mode) {
```
emanuele3d: In the previous comment regarding this method I suggested to add to the javadoc the exceptions this method throws. There's also this form of method definition that includes the exceptions thrown:

`public boolean unregister(MetricsMode mode) throws except1, except2 {`

I'm not sure when to do it only in the javadoc, when to do it only in the method definition and when to do it both ways. @msteiger: Maaaaaaaaaartin?!?!?!?

By the way, the same applies to the register method.

GooeyHub:

msteiger:
Thanks, Taha!

Issue #2317, opened 2016-05-13
New API testing module


Cervator:
While reviewing and merging #2306 I detected a tiny API violation of SemVer that was able to slip past detection earlier since we don't yet have any ability to actually test the module API on engine changes.

While we're in Alpha and flying half-blind like this it is probably very hard to avoid trivial violations like that which can be so subtle. Maybe we can be accepting of that on a case-by-case basis until Beta or when we no longer can say we are aware of and able to fix all live modules. Thoughts?

In either case we need a better way to detect API violations. We have the `ApiScraper` already to produce *what* to test, how about creating a unit test module that'll exercise every single entry minimally, including a bit of logic to compare what's tested vs. what exists? Maybe there's some clever way to work code coverage into it, this is something we should hit 100% on.

Related: #2159, #1975, https://github.com/MovingBlocks/Terasology/wiki/Modding-API

GabrielXia:
Heya :-)

I implement an API test for interface `Context` and class `ComponenetFieldUri`,  check it out:
 [ApiTest](https://github.com/GabrielXia/ApiTest/tree/master). In short it compares the signatures of current public & protected methods (method name, modifiers, parameter, return type, exceptions) to the signatures registered. Is it what we want? Now the signature registering stuff is kind of manual. I think maybe it's better to gather some comments before going on testing the huge size of API ;-)

Thanks for your suggestions!

Cervator:
Bit of a belated response here but yep, that seems about right :-)

At first I was thinking simply *using* each method in some test, since in that case any method signature change would cause a failure. But if there are validators like that it may be even more correct!

We still would want actual unit tests that exercise stuff too, eventually. I wonder if having the method signature validators would cause code coverage analysis to think that code is covered? That might be something to look into. IMHO simply validating the method signatures shouldn't count as covering the code itself.

This reminds me a bit about #2851 which was using a generator that makes unit tests. Really bare ones but it might have some potential for API testing I wonder? Pinging @iaronaraujo for reference.

iaronaraujo:
I just made a pull request with a tutorial about how to use EvoSuite to create unit tests!
https://github.com/MovingBlocks/Terasology/pull/2906

Cervator:
Thanks @iaronaraujo for the nice write-up! And for popping up on this issue as I wanted to run something by both you and @GabrielXia :-)

[...]

PR #2325, opened 2016-05-25
Wrapping ShadowMap and WorldReflection into a "Node"

tdgunes:


GooeyHub:

GooeyHub:

Cervator:
Went ahead and tested this, since it was here, although I had no idea what to really test :D

Ran around a world a bit looking at the pretty flowers, then cycled through the various debug rendering modes, including some pretty trippy ones (more than I remember! Neato, almost have LSD-mode in there again)

Is this related to some discussion elsewhere, like on Slack/IRC/Forum/GSOC conference meetings? I figure it might be from the empty description. What are our plans with this PR - just for reference or merge it soon? :-)

emanuele3d:
The goal of this PR is to implement a barebone DAG class(es) and have -one- rendering step wrapped into a node. Currently the targeted step is the one generating the main light's shadow map.

That been said, @tdgunes had his thesis discussion today I think, the last hurdle before graduation. So, he might have needed a bit of time to prepare that in the past week or so and he might need a bit of time to get over the celebrations in the next few days. =D

emanuele3d:
Sorry, I didn't address one of your points @Cervator: indeed the goal of this PR has been set in one of our live meeting, first on appear.in and then via skype due to technical problems. @tdgunes also already has some feedback regarding this PR, discussed at our last meeting, and we are awaiting a new commit to incorporate that feedback.

In general Martin, Taha and I have agreed that it will be preferable to have small(ish) weekly or bi-weekly incremental changes to the rendering engine instead of developing on a separate branch for a long time to then spend a long time integrating everything. So, we don't expect any individual PR to stay open for too long. Perhaps this first one might take a bit longer but mostly because Taha is busy wrapping up four years of study. =)

Cervator:
I'd double-thumbs-up that follow-up post, but there isn't a reaction-emoticon for that so I'll post a full comment just to say thanks for all the details! :-)

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
```diff
@@ -75,14 +78,19 @@
...
+                               DirectedAcyclicClassGraph<RenderNode> dag) {
+
         this.worldProvider = worldProvider;
         this.chunkProvider = chunkProvider;
         chunkTessellator = new ChunkTessellator(bufferPool);
         chunkMeshUpdateManager = new ChunkMeshUpdateManager(chunkTessellator, worldProvider);
 
         this.playerCamera = playerCamera;
-        this.shadowMapCamera = shadowMapCamera;
+        this.dag = dag;
```
emanuele3d: Let's use the class and instance name to communicate as much information as possible. Let's call it RenderingPipeline.

engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
```diff
@@ -372,7 +381,7 @@ public boolean isChunkValidForRender(RenderableChunk chunk) {
     }
 
     public boolean isChunkVisibleLight(RenderableChunk chunk) {
-        return isChunkVisible(shadowMapCamera, chunk);
+        return isChunkVisible(dag.get(ShadowMapNode.class).getCamera(), chunk); //TODO: find an elegant way
```
emanuele3d: Hmmm. I have a number of concerns here.
1. dat.get(class) seems to imply there should be only once instance of a given node class. Without thinking too deeply about it I have the impression we might now want to limit it like that. I'd rather work with nodes declaring/registering their own name instead. I.e. in the case of the Blur nodes, BlurForDoF and BlurForBloom for example. That been said, we might want to keep this for a later stage.
2. I am not so fond on the idea of fetching the ShadowMapNode from a map and then fetching the camera from that node, potentially frequently. Perhaps the RenderableWorldImpl could have a setShadowMapCamera() method? Or a setShadowMapNode() which sets the camera?

That been said, let's rename this method to isChunkVisibleFromMainLight(),  which is what this method does. 

Nevermind that at some point we'll want multiple main lights. How do we get Tatooine in Terasology otherwise?? =)

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -125,6 +126,9 @@
     private GraphicState graphicState;
     private PostProcessor postProcessor;
 
+
+    private ShadowMapNode shadowMap;
+    private DirectedAcyclicClassGraph<RenderNode> dag = new DirectedAcyclicClassGraph<>();
```
emanuele3d: Similar to comment in previous file, let's rename this variable renderingPipeline. Also, we might want to add a comment in both cases saying that this is a work in progress and eventually the whole rendering process will go through this object, not just the ShadowMap. Just in case somebody browses the code in the next few weeks, before the whole thing actually happens. Otherwise it'd be a bit tricky to understand why something like a rendering pipeline is taking care only of a shadow map.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -165,7 +169,7 @@ public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
         LocalPlayerSystem localPlayerSystem = context.get(LocalPlayerSystem.class);
         localPlayerSystem.setPlayerCamera(playerCamera);
 
-        renderableWorld = new RenderableWorldImpl(worldProvider, context.get(ChunkProvider.class), bufferPool, playerCamera, shadowMapCamera);
+        renderableWorld = new RenderableWorldImpl(worldProvider, context.get(ChunkProvider.class), bufferPool, playerCamera, dag);
```
emanuele3d: Hmmm... in light of what I suggested above I'd not pass the dag nor the shadowMapCamera. Just create a new setShadowMapCamera() method which get called on ShadowMapNode insertion into or removal from the graph. However, this would bring up the question what to do if there is no ShadowMapCamera set in the RenderableWorld instance? I guess the isChunkVisibleFromMainLight() method might want to throw an exception if no shadowmap camera has been set. But what consequences would that have elsewhere in the code? At this stage I don't know. Let me know what you think about this.

By the way, the RenderableWorldImpl instance effectively holds a number of acceleration structures, I'm tempted to call them caches but I'm not entirely sure it would be correct. One of these is related to shadow mapping, which is why we are having this conversation. EVENTUALLY, the ShadowMapNode (or any other node), probably on insertion in the graph, would want to have the chance to register the acceleration structures it needs. So, EVENTUALLY, the RenderableWorld class might either become a more generic acceleration structures manager (and be renamed accordingly) or, even better, acceleration structures become nodes in the pipeline.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -195,13 +199,21 @@ private void initRenderingSupport() {
...
         postProcessor.initializeMaterials();
         initMaterials();
+
+        shadowMap = new ShadowMapNode(getMaterial("engine:prog.shadowMap"),
+                playerCamera,
+                this,
+                backdropProvider,
+                renderingConfig,
+                graphicState);
+
```
emanuele3d: Generally speaking a RenderNode should somehow find everything it needs. This way the DAG remains agnostic of what a node does or needs and can treat all nodes the same.

So:
1. let the node find the renderingConfig and the backdropProvider via a context object and let it obtain the playerCamera from the WorldRenderer instance.

2. GraphicState is a class I created to collect easily recognizable opengl directives in one place. Earlier they were "visually polluting" the renderer code making it hard to get an overview - like a picture with too many details and no focus to guide the eye. However, as we divide the code of the existing rendering engine into RenderNodes, it will make sense to move those gl directives into the nodes, near where it's needed. The code within a node will never be horribly big, probably. So, what you could do is to place the code currently in GraphicState.preRenderSetupSceneShadowMap() and GraphicState.postRenderCleanupSceneShadowMap() within the ShadowMapNode, eliminating the dependency on the graphicState instance.

Word of caution though: I remind you that at this stage generating the ShadowMap or executing any other step in the rendering engine might still rely on state changes occurred much earlier in the flow of execution. Until every state change is first done and then undone immediately after use, leaving to the DAG to optimize the process, you must ensure that RenderNodes are executed in the current order. Let me know if you need me to further clarify this.

engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
Cervator: get dat class!

(and yes, want multiple celestials, gimme!)

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -271,20 +252,16 @@ private void preRenderUpdate(RenderingStage renderingStage) {
         resetStats();
 
         currentRenderingStage = renderingStage;
-        if (currentRenderingStage == RenderingStage.MONO || currentRenderingStage == RenderingStage.LEFT_EYE) {
-            isFirstRenderingStageForCurrentFrame = true;
-        } else {
-            isFirstRenderingStageForCurrentFrame = false;
-        }
+        isFirstRenderingStageForCurrentFrame = currentRenderingStage == RenderingStage.MONO || currentRenderingStage == RenderingStage.LEFT_EYE;
```
emanuele3d: Hmmmm... of course here the code it's syntactically equivalent to the previous version. In fact, I think originally the code was similar to what you suggest. However, I remember changing it into an if/else block to make it more readable given how important ifFirstRenderingStageForCurrentFrame is. So, I'd prefer to keep it as it is for the time being. Eventually the whole thing will go away as we deal with a Stereoscopic setup through RenderNodes and the pipeline.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -195,13 +199,21 @@ private void initRenderingSupport() {
...
         initMaterials();
+
+        shadowMap = new ShadowMapNode(getMaterial("engine:prog.shadowMap"),
+                playerCamera,
+                this,
+                backdropProvider,
+                renderingConfig,
+                graphicState);
+
+        dag.add(shadowMap);
```
emanuele3d: Let's call the instance a shadowMapNode. This RenderNode is not the shadow map itself.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -271,20 +252,16 @@ private void preRenderUpdate(RenderingStage renderingStage) {
...
         // this is done to execute this code block only once per frame
         // instead of once per eye in a stereo setup.
         if (isFirstRenderingStageForCurrentFrame) {
             timeSmoothedMainLightIntensity = TeraMath.lerp(timeSmoothedMainLightIntensity, getMainLightIntensityAt(playerCamera.getPosition()), secondsSinceLastFrame);
 
             playerCamera.update(secondsSinceLastFrame);
-            positionShadowMapCamera();
-            shadowMapCamera.update(secondsSinceLastFrame);
+
+            shadowMap.preRender(secondsSinceLastFrame);
```
emanuele3d: Throughout Terasology's code the methods update() and render() are common occurrences.  In particular it is known that updates happen before the rendering and might occur multiple times per frame. I'd recommend in this case to stick to that and have shadowMapNode.update(secondsSinceLastFrame) instead.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -374,19 +351,7 @@ public void render(RenderingStage renderingStage) {
     private void renderShadowMap() {
         if (renderingConfig.isDynamicShadows() && isFirstRenderingStageForCurrentFrame) {
             PerformanceMonitor.startActivity("Render World (Shadow Map)");
```
emanuele3d: Let's put the PerformanceMonitor statements within the node please.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -374,19 +351,7 @@ public void render(RenderingStage renderingStage) {
     private void renderShadowMap() {
         if (renderingConfig.isDynamicShadows() && isFirstRenderingStageForCurrentFrame) {
```
emanuele3d: Ok, this is a tricky one. Ultimately, we want the isDynamicShadows() check to NOT occur. If dynamicShadows is activated, the ShadowMapNode is added in the graph and gets processed. If it isn't, there is no ShadowMapNode in the graph. Similarly, the whole renderShadowMap() method should cease to exist. Initially it could be replaced by the shadowMapNode.render() method, eventually even that will be called through something like "RenderPipeline.transverse() or .process() or .execute() or .render()".

What I don't know is if it's worth taking care of it now already. Perhaps not yet. Perhaps it can wait until the next PR. What do you think?

What might be worth is to move the second check, isFirstRenderingStageForCurrentFrame, within the shadowMapNode, until we deal with VR support.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -374,19 +351,7 @@ public void render(RenderingStage renderingStage) {
...
-            renderChunks(renderQueues.chunksOpaqueShadow, ChunkMesh.RenderPhase.OPAQUE, shadowMapCamera, ChunkRenderMode.SHADOW_MAP);
-
-            for (RenderSystem renderer : systemManager.iterateRenderSubscribers()) {
-                renderer.renderShadows();
-            }
-
-            playerCamera.lookThrough(); // not strictly needed: just defensive programming here.
-            graphicState.postRenderCleanupSceneShadowMap();
-
+            shadowMap.render();
```
emanuele3d: I am thinking it might not be appropriate to call this method .render(). The graph will eventually contain nodes that are not render nodes, i.e. cameras and acceleration structures. For those it wouldn't make sense to have a "render()" method. And the graph should be transversed by calling the same method on each node as the graph is not really supposed to know much about what nodes do. So, I'd recommend something like "node.execute()" or something similar.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -777,7 +739,8 @@ public Camera getActiveCamera() {
 
     @Override
     public Camera getLightCamera() {
-        return shadowMapCamera;
+        //FIXME: remove this method
```
emanuele3d: @Cervator: how do we know if a method is used in the API, the one that we shouldn't break because we fixed it for version 1?

engine/src/main/java/org/terasology/rendering/world/dag/RenderNode.java
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.world.dag;
+
+import org.terasology.rendering.cameras.Camera;
+
+public interface RenderNode {
```
emanuele3d: I'm thinking perhaps we should rename this to DAGNode to keep it generic and allow for things such as camera nodes and acceleration structures.

engine/src/main/java/org/terasology/rendering/world/dag/RenderNode.java
```diff
@@ -0,0 +1,37 @@
...
+
+    void initialise();
+
+    void preRender(float update);
```
emanuele3d: See comment above: I recommend changing this to update(float deltaInSeconds).

engine/src/main/java/org/terasology/rendering/world/dag/RenderNode.java
```diff
@@ -0,0 +1,37 @@
...
+
+    void render();
```
emanuele3d: As per comment above, I recommend changing this into execute(). Or perhaps process()?

engine/src/main/java/org/terasology/rendering/world/dag/RenderNode.java
```diff
@@ -0,0 +1,37 @@
...
+
+    void postRender();
```
emanuele3d: I'm not sure if this is needed now. I can see the aesthetics of having a pre/postRender method, but as I'm suggesting to rename preRender(), I wonder what would go in postRender() that can't go in process().

I guess we could have a preProcess() method executed before all nodes in the dag get processed and a postProcess() method executed after all nodes in the dag have been processed. At this stage I just don't know if we'd need them.

engine/src/main/java/org/terasology/rendering/world/dag/RenderNode.java
```diff
@@ -0,0 +1,37 @@
...
+
+    void dispose();
+
+    void insert();
```
emanuele3d: I'm thinking this is effectively an event-processing method, in the sense that from my understanding it should be executed -on- insertion in the graph. So, I'd recommend to call it that onInsert(). I'd also add a method onRemove().

engine/src/main/java/org/terasology/rendering/world/dag/RenderNode.java
```diff
@@ -0,0 +1,37 @@
...
+
+    // FIXME: might be deleted in the future, some of the RenderNode's may not need a camera
+    Camera getCamera();
```
emanuele3d: I'd definitely remove it. Some -nodes- might not have a camera at all and some nodes might be the camera!

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+
+import static org.lwjgl.opengl.EXTFramebufferObject.GL_FRAMEBUFFER_EXT;
+import static org.lwjgl.opengl.EXTFramebufferObject.glBindFramebufferEXT;
+import static org.lwjgl.opengl.GL11.*;
+
+
+/**
+ * Diagram of this node can be viewed from:
+ * - https://docs.google.com/drawings/d/13I0GM9jDFlZv1vNrUPlQuBbaF86RPRNpVfn5q8Wj2lc/edit?usp=sharing
+ */
```
emanuele3d: Let's add a TODO: move diagram to the wiki when this part of the code is stable.

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Override
+    public void initialise() {
+        // TODO: shader and fbo shall be initialised here
+        // TODO: using ShaderParametersShadowMap here, instead of storing in ShaderManagerLwjgl
+    }
+
+    @Override
+    public void preRender(float update) {
+        // positionShadowMapCamera()
```
emanuele3d: Is there a reason to have this code not in render()/execute()? In fact, if you agree with my suggestion to turn this into an update(float deltaInSeconds) method, it would be a mistake to place this code here as it would potentially get executed multiple times per frame, unnecessarily.

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+
+        camera.update(update);
+
+    }
+
+
+    @Override
+    public void render() {
+        // TODO: removing this assignment
+        fbo = state.buffers.sceneShadowMap; // added here since graphic settings can be modified and a new fbo can be generated.
```
emanuele3d: Ok, two things to do immediately:
1. Do not obtain the buffers from GraphicState. Get them from FrameBuffersManager, their actual home.
2. Do not hotlink the (originally private) buffer structure: use getFBO(String name)

Something for later: this fetches the FBO once per frame. Ideally:
1. the ShadowMapNode is responsible for generating the FBO and registering it with the FrameBuffersManager. This would require a "putFBO()" method in the manager - it's not there right now.
2. the ShadowMapNode should be responsible for monitoring the renderConfig and regenerate the FBOs on shadow map size changes. I think this should be handled by making the renderConfig observable, so that observer are notified of any changes or specific changes. 
3. the ShadowMapNode should also dispose of the FBO when the node itself gets disposed. 

More medium term, this makes me think that there is scope for a DAGNode.onRemoval() method and a separate DAGNode.onDeletion() method. This would allow for a node to be quickly re-inserted into the graph, it's structures and data not yet disposed. This way a user could quickly switch back&forth between having and not having the shadows, i.e. toggling back&forth the option in the config menu. Then, when the user leaves the menu, the node is either in the graph or gets deleted.

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * Diagram of this node can be viewed from:
+ * - https://docs.google.com/drawings/d/13I0GM9jDFlZv1vNrUPlQuBbaF86RPRNpVfn5q8Wj2lc/edit?usp=sharing
+ */
+public class ShadowMapNode implements RenderNode {
+    private static final int SHADOW_FRUSTUM_BOUNDS = 500;
+
+    private Material shader;
+    private FBO fbo;
```
emanuele3d: Let's call this shadowMap, as that's what it contains.

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+
+    // FIXME: unnecessary arguments must be eliminated
+    public ShadowMapNode(Material shader,
+                         Camera playerCamera,
+                         WorldRendererImpl worldRenderer,
+                         BackdropProvider backdropProvider,
+                         RenderingConfig renderingConfig,
+                         GraphicState state) {
+
+        this.shader = shader;
```
emanuele3d: Hmmmm... I don't like the generic name. But I do not like the asset name either ("shadowMap"). I'm thinking "shadowCaster" or perhaps "shadowCastingShader"?

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+        // preRenderSetupSceneShadowMap
+        fbo.bind();
+        glViewport(0, 0, fbo.width(), fbo.height());
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+        GL11.glDisable(GL_CULL_FACE);
+
+        // render
+        camera.lookThrough();
+
+        shader.enable();
```
emanuele3d: Hmmm... I don't like the generic "shader" name. We should probably call it something more specific, i.e. "shadowCastingShader". This reminds me I need to study the specifics of Material vs Shader within Terasology.

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+
+
+    @Override
+    public void render() {
+        // TODO: removing this assignment
+        fbo = state.buffers.sceneShadowMap; // added here since graphic settings can be modified and a new fbo can be generated.
+
+        // preRenderSetupSceneShadowMap
+        fbo.bind();
+        glViewport(0, 0, fbo.width(), fbo.height());
```
emanuele3d: Ok, good that you already did what I suggested earlier: you moved the content of GraphicState.preRenderSetupSceneShadowMap() and GraphicState.postRenderCleanupSceneShadowMap()  into the node.

Not as good that you unwrapped some of the gl calls I had wrapped into more meaningfully named method names. Notice that as those methods are very small the java compiler will place them inline. So, we are loosing readability and gaining no performance.

So, I'd prefer if you reverted this line back to setViewportToSizeOf(shadowMap).

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
```
emanuele3d: Please add a TODO here: verify the need to clear color buffer - if I remember correctly the shadowMap fbo has no color buffer at all!

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+
+        shader.enable();
+        // FIXME: storing chuncksOpaqueShadow or a mechanism for requesting a chunk queue for nodes which calls renderChunks method?
+        worldRenderer.renderChunks(worldRenderer.renderQueues.chunksOpaqueShadow, ChunkMesh.RenderPhase.OPAQUE, camera, WorldRendererImpl.ChunkRenderMode.SHADOW_MAP);
+        playerCamera.lookThrough(); //FIXME: not strictly needed: just defensive programming here.
+
+
+        // postRenderCleanupSceneShadowMap
+        GL11.glEnable(GL_CULL_FACE);
+        glBindFramebufferEXT(GL_FRAMEBUFFER_EXT, 0); // bindDisplay()
```
emanuele3d: Similarly to above, I'd rather get this back to bindDisplay(). Also, the postRenderCleanupSceneShadowMap() method had an additional line: setViewportToWholeDisplay(); It's probably redundant but for the time being we should leave it there. Perhaps the way to do this is to keep GraphicState until it is reduced to a skeleton (without all the preRenderSetup/postRenderCleanup methods) and then rename it to something more meaningful, i.e. OpenGLHelper class or OpenGLUtils or something along those lines.

engine/src/main/java/org/terasology/rendering/world/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Override
+    public void postRender() {
+
+    }
+
+    @Override
+    public void dispose() {
+        // TODO: disposing shader and fbo whenever node is removed from DAG
```
emanuele3d: In line with a previous comment I'd recommend to change this to onDeletion() and add an onRemoval() method.

engine/src/main/java/org/terasology/utilities/collection/DirectedAcyclicClassGraph.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.utilities.collection;
+
+import com.google.api.client.util.Maps;
+
+import java.util.Map;
+
+public class DirectedAcyclicClassGraph<T> extends DirectedAcyclicGraph<T> {
```
emanuele3d: I won't make comments about these two classes but I'd like to talk about them again before this PR gets merged. Primarily my question is, why -two- classes instead of a more typical interface/class combo? Time for me to go to bed!

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/opengl/OpenGLUtil.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 MovingBlocks
+ * Copyright 2016 MovingBlocks
```
emanuele3d: I saw this file's name and I started thinking: should be Util or Utils. And then it occurred to me that we really shouldn't abbreviate: 

http://programmers.stackexchange.com/questions/72593/how-to-abbreviate-variable-names

Let's rename it simply to "Utilities". As it's in the OpenGL package we can drop the OpenGL prefix.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -41,16 +41,18 @@
...
-import org.terasology.rendering.cameras.OrthographicCamera;
 import org.terasology.rendering.cameras.PerspectiveCamera;
 import org.terasology.rendering.logic.LightComponent;
 import org.terasology.rendering.opengl.FrameBuffersManager;
 import org.terasology.rendering.opengl.GraphicState;
 import org.terasology.rendering.opengl.PostProcessor;
 import org.terasology.rendering.primitives.ChunkMesh;
 import org.terasology.rendering.primitives.LightGeometryHelper;
+import org.terasology.rendering.world.dag.Node;
+import org.terasology.rendering.world.dag.ShadowMapNode;
```
emanuele3d: Didn't we say to move the dag package up a level? Directly under rendering?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
Cervator: Generally I simply do a full build of my mega-source workspace, with all the stable modules present. If something breaks then we broke API :-)

This tests out fine - no compile errors. You're also *generally* safe if there's no `@API` at the top, or in a package info file. It looks like there are just four usages of this method elsewhere in the rendering system.

There are better ways to check for API violations in a more sane and thorough way. I documented one suggestion in #2317

engine/src/main/java/org/terasology/rendering/opengl/OpenGLUtil.java
Cervator: I'd be hesitant to drop the `OpenGL` part just because we might end up in a situation of having multiple `Utilities` classes, having to preface each usage in a file with the full package path. That seems like it would be a little messy :-)

I dunno if there'd be a better more descriptive name that bakes in the exact purpose, somehow splicing together the rendering nature + utility?

Cervator:
Tested it out merged with latest develop in a local branch. All seems to be well :-)

Although we might have a shadow-related issue hiding somewhere. I started up JoshariasSurvival and was able to find a big cave system that got me half a kilometer underground. After a while Eye Adaption started wigging out, causing lag. Disabled it. Bit further and shadows did the same. Disabled those (and everything else) and could continue fine.

I'm not sure this PR actually changes anything like that. I figure it is pre-existing :-)

The feeling of being underground is great. The lighting doesn't reach very far, so I'd have to be careful with ledges edging an abyss, throwing torches down to see how deep they were before continuing. Torchlight from thrown torches does end somewhat abruptly, although it may be hard to tell when not in a huge underground cave where you can see the edge of thrown light very easily.

tdgunes:
@Cervator Thanks for testing it 👍.  What I try to be careful is to not break any existing rendering features, but carefully wrapping rendering tasks into nodes. The more visible effects of actual changes will appear in the feature (especially in 3rd party module development.)

engine/src/main/java/org/terasology/rendering/opengl/OpenGLUtil.java
emanuele3d: Ah. Fair enough @Cervator. Then I'd just rename it OpenGLUtilities, at least for the time being.

engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 MovingBlocks
+ * Copyright 2016 MovingBlocks
```
emanuele3d: @tdgunes, I just had a thought. If it's too difficult to split the content of RenderableWorldImpl into the nodes that use its acceleration structures, perhaps we can turn it into a node of its own, on which other nodes are dependent.

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
tdgunes: @emanuele3d, I think it's a really good idea! After this PR is merged, instead of giving WorldRenderer object within in a context to other nodes, a WorldRenderer instance can be turned into a node, as you mentioned.  For instance nodes that require `renderChunks()` method can use WorldRenderer node, maybe via `getDependentNodes().get(WorldRendererNode.class)` or also maybe a syntactic sugar achieved by annotations:
```ocaml
@In 
WorldRendererNode worldRenderer;
```
(Side note: As we discussed in the meeting, this approach has some issues but I am writing here just in case.)

emanuele3d: Watch out, I'm talking about the "RenderableWorld" instance, not the WorldRenderer. =)

But I'll have a look if by any chance also the WorldRenderer could be turned into a node...

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,146 @@
...
+
+
+    public Camera camera = new OrthographicCamera(-SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, -SHADOW_FRUSTUM_BOUNDS);
+    private RenderQueuesHelper renderQueues;
+    private Context context;
+    private Material shadowMapShader;
+    private FBO shadowMap;
+
+    // TODO: every node proposes its modifiable configuration?
+    private RenderingConfig renderingConfig;
```
emanuele3d: I guess there could be nodes that:
* have no modifiable configuration
* nodes whose configuration is not in a render config
* nodes that can be configured only via their API

Is this what you were aiming at? For the time being however, I'd say all nodes do.

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,146 @@
...
+    public ShadowMapNode(Context context, Material shadowMapShader, Camera playerCamera) {
+        this.context = context;
+        this.shadowMapShader = shadowMapShader;
+        this.playerCamera = playerCamera;
+        this.worldRenderer = context.get(WorldRenderer.class);
+        this.backdropProvider = context.get(BackdropProvider.class);
+        this.renderingConfig = context.get(Config.class).getRendering();
+        this.shadowMap = context.get(FrameBuffersManager.class).getFBO("sceneShadowMap");
+        this.renderQueues = context.get(RenderQueuesHelper.class);
+    }
```
emanuele3d: My only suggestion here would be that you can get the playerCamera via WorldRenderer.getActiveCamera().

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Diagram of this node can be viewed from:
+ * TODO: move diagram to the wiki when this part of the code is stable
+ * - https://docs.google.com/drawings/d/13I0GM9jDFlZv1vNrUPlQuBbaF86RPRNpVfn5q8Wj2lc/edit?usp=sharing
+ */
+public class ShadowMapNode implements Node {
+    private static final int SHADOW_FRUSTUM_BOUNDS = 500;
+
+
+    public Camera camera = new OrthographicCamera(-SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, -SHADOW_FRUSTUM_BOUNDS);
```
emanuele3d: Let's call this shadowMapCamera please.

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,146 @@
...
+        this.worldRenderer = context.get(WorldRenderer.class);
+        this.backdropProvider = context.get(BackdropProvider.class);
+        this.renderingConfig = context.get(Config.class).getRendering();
+        this.shadowMap = context.get(FrameBuffersManager.class).getFBO("sceneShadowMap");
+        this.renderQueues = context.get(RenderQueuesHelper.class);
+    }
+
+    @Override
+    public void update(float deltaInSeconds) {
+        this.shadowMap = context.get(FrameBuffersManager.class).getFBO("sceneShadowMap");
```
emanuele3d: Hmmmm... no. What is currently in update() here, should be be executed within process() instead. In Terasology update() methods are potentially executed multiple times per frame, which in this case I do not think we need. In fact at this stage I do not think we need an update method at all and I'd remove it.

emanuele3d: Of course this doesn't mean you should just cut and paste everything into process(). I.e. all camera positioning stuff should go in its private method.

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,146 @@
...
+
+        // positionShadowMapCamera()
```
emanuele3d: What is this doing here?

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
msteiger: I would use the InjectionHelper here. This enables the `@In` annotation for the fields that are known to the context.

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+
+    @Override
+    public void process() {
+        PerformanceMonitor.startActivity("Render World (Shadow Map)");
+
+        // preRenderSetupSceneShadowMap
+        shadowMap.bind();
+        setViewportToSizeOf(shadowMap); // TODO: how about shadowMap.setAsViewport() ?
```
emanuele3d: Hmmmm. Reading shadowMap.setAsViewport() I'd think that something is happening to the object shadowMap rather than the to viewport. Normally a method affects the instance is attached to or the instance is the "agent" that causes something. In this case neither make sense. We could theorethically go for something like shadowMap.setViewportToItsSize() but it's a bit convoluted. I prefer the immediacy of setViewportToSizeOf(anFBO), as it's almost plain English.

engine/src/main/java/org/terasology/rendering/dag/WorldReflectionNode.java
```diff
@@ -0,0 +1,105 @@
...
+    private Material chunkShader;
+
+    private Context context;
+    private BackdropRenderer backdropRenderer;
+    private Camera playerCamera;
+
+    private FBO sceneReflected;
+    private FBO sceneOpaque;
+
+    public WorldReflectionNode(Context context, Camera playerCamera, Material chunkShader) {
```
emanuele3d: Similar to above, you can get the playerCamera via WorldRenderer.getActiveCamera().

engine/src/main/java/org/terasology/rendering/dag/WorldReflectionNode.java
```diff
@@ -0,0 +1,105 @@
...
+        this.backdropRenderer = context.get(BackdropRenderer.class);
+        this.playerCamera = playerCamera;
+        this.renderingConfig = context.get(Config.class).getRendering();
+        this.chunkShader = chunkShader;
+        this.worldRenderer = context.get(WorldRenderer.class);
+        this.renderQueues = context.get(RenderQueuesHelper.class);
+    }
+
+    @Override
+    public void update(float deltaInSeconds) {
```
emanuele3d: Similar to above, the content of this method can be placed in process().

Also, get the FrameBuffersManager instance only once please, on construction I'd suggest.

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
emanuele3d: Also, can you obtain the FrameBuffersManager instance on construction instead?

engine/src/main/java/org/terasology/rendering/dag/WorldReflectionNode.java
```diff
@@ -0,0 +1,105 @@
...
+        this.sceneReflected = context.get(FrameBuffersManager.class).getFBO("sceneReflected");
+        this.sceneOpaque = context.get(FrameBuffersManager.class).getFBO("sceneOpaque");
+    }
+
+    @Override
+    public void process() {
+        PerformanceMonitor.startActivity("Render World (Reflection)");
+
+        sceneReflected.bind();
+        OpenGLUtils.setViewportToSizeOf(sceneReflected);
```
emanuele3d: Let's remove the "OpenGLUtils."

engine/src/main/java/org/terasology/rendering/dag/WorldReflectionNode.java
```diff
@@ -0,0 +1,105 @@
...
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+        GL11.glCullFace(GL11.GL_FRONT);
+        playerCamera.setReflected(true);
+
+
+        playerCamera.lookThroughNormalized(); // we don't want the reflected scene to be bobbing or moving with the player
+        backdropRenderer.render(playerCamera);
+        playerCamera.lookThrough();
+
+        if (renderingConfig.isReflectiveWater()) {
```
emanuele3d: Hmmmm.... let's put a TODO here:

    TODO: the isReflectiveWater() block should include content of the whole process() method. Eventually the check will be removed, when node insertion will signify that the feature is enabled.

engine/src/main/java/org/terasology/rendering/dag/WorldReflectionNode.java
```diff
@@ -0,0 +1,105 @@
...
+
+        sceneReflected.bind();
+        OpenGLUtils.setViewportToSizeOf(sceneReflected);
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+        GL11.glCullFace(GL11.GL_FRONT);
+        playerCamera.setReflected(true);
+
+
+        playerCamera.lookThroughNormalized(); // we don't want the reflected scene to be bobbing or moving with the player
+        backdropRenderer.render(playerCamera);
```
emanuele3d: Let's add:

    TODO: convert backdropRenderer into a node.

Note for @tdgunes: this will be an interesting use case for GL state management and order of execution of nodes: both the the BackdropNode and the WorldReflectionNode need similar/identical states, so the nodes should be processed one after another and the first node to be processed should not be allowed to reset to default the states the second node needs. Also interesting, the BackdropNode will need to be used twice, one here, to render the reflection, and then later, for the proper renderer. Something to put in the back of our minds.

engine/src/main/java/org/terasology/rendering/dag/WorldReflectionNode.java
```diff
@@ -0,0 +1,105 @@
...
+            chunkShader.activateFeature(ShaderProgramFeature.FEATURE_USE_FORWARD_LIGHTING);
+            worldRenderer.renderChunks(renderQueues.chunksOpaqueReflection, ChunkMesh.RenderPhase.OPAQUE, playerCamera, WorldRendererImpl.ChunkRenderMode.REFLECTION);
+            chunkShader.deactivateFeature(ShaderProgramFeature.FEATURE_USE_FORWARD_LIGHTING);
+        }
+
+        playerCamera.setReflected(false);
+
+        GL11.glCullFace(GL11.GL_BACK);
+        bindDisplay();
+        OpenGLUtils.setViewportToSizeOf(sceneOpaque); // toWholeDisplay
```
emanuele3d: Let's remove "OpenGLUtils." 

Regarding the //toWholeDisplay comment, indeed the original code had "setViewportToWholeDisplay();"  Why did you change it? 

Interestingly, this is one case in which a state change is used also later, outside of the render reflection step. In the following rendering steps, i.e. backdrop render and then the whole scene opaque, the viewport stays to what is set here. Again, state management will have to take care of this: each node should be independent of the others unless there is an explicit requirement (I'm thinking a node could have a setExecutionDependency(Node ancestor) method - or something like that), state management will then ensure that the GL state is changed only if necessary.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
```diff
@@ -81,9 +80,12 @@
     // eventually decided for GraphicState as it resides in the rendering.opengl
     // package anyway and rendering.opengl.OpenGLState felt cumbersome. --emanuele3d
 
+    // FIXME: for making nodes work without breaking anything
+    public Buffers buffers = new Buffers();
```
emanuele3d: If you move this line and the comment back where it used to be github will do a better job at doing the diff.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
```diff
@@ -515,28 +470,6 @@ public void setRenderBufferMask(FBO fbo, boolean color, boolean normal, boolean
...
     /**
-     * Sets the viewport of the currently bound FBO to the size of the Display.
-     * 
-     * Which might mean the full screen, the full size of the window or the
-     * currently set screenshot size if a screenshot is being taken.
-     */
-    // TODO: verify if this can become part of the bindDisplay() method.
-    public void setViewportToWholeDisplay() {
-        glViewport(0, 0, fullScale.width(), fullScale.height());
-    }
```
emanuele3d: Shouldn't this method and its comment also go in OpenGLUtils?

emanuele3d: Scrap that. I can imagine it's not straightforward to get the content of fullScale. We can leave with this for a while.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
```diff
@@ -515,28 +470,6 @@ public void setRenderBufferMask(FBO fbo, boolean color, boolean normal, boolean
...
-     */
-    // TODO: verify if this can become part of the FBO.bind() method.
-    public void setViewportToSizeOf(FBO fbo) {
-        glViewport(0, 0, fbo.width(), fbo.height());
-    }
-
-    /**
      * Unbinds any currently bound FBO and binds the default Frame Buffer,
      * which is usually the Display (be it the full screen or a window).
      */
```
emanuele3d: Please copy this javadoc to bindDisplay() in OpenGLUtils. Can bindDisplay() already be moved completely? Or do you prefer to leave here to avoid touching too many things?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -86,20 +90,18 @@
...
 
     private final Camera playerCamera;
-    private final Camera shadowMapCamera = new OrthographicCamera(-SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, -SHADOW_FRUSTUM_BOUNDS);
 
     // TODO: Review this? (What are we doing with a component not attached to an entity?)
     private LightComponent mainDirectionalLight = new LightComponent();
     private float timeSmoothedMainLightIntensity;
 
-    private final RenderQueuesHelper renderQueues;
+
```
emanuele3d: If you remove the content of a line you can also remove the line! =)

emanuele3d: Now that I think about it, I've noticed this happening a number of times. Please go through the code and eliminate lines you left empty. Of course you can leave empty lines that were already there or if you feel that it make sense to leave an additional empty line.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -195,13 +204,25 @@ private void initRenderingSupport() {
         shaderManager.initShaders();
         postProcessor.initializeMaterials();
         initMaterials();
+
+        context.put(WorldRenderer.class, this);
+        context.put(RenderQueuesHelper.class, renderQueues);
+
+        // FIXME: playerCamera to context?
```
emanuele3d: No, it can be obtained via the WorldRenderer for the time being.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -195,13 +204,25 @@ private void initRenderingSupport() {
...
+        shadowMapNode = new ShadowMapNode(context, getMaterial("engine:prog.shadowMap"), playerCamera);
```
emanuele3d: How about you obtain the Material in initMaterial()?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -195,13 +204,25 @@ private void initRenderingSupport() {
         shaderManager.initShaders();
         postProcessor.initializeMaterials();
         initMaterials();
+
+        context.put(WorldRenderer.class, this);
```
emanuele3d: Can this be placed at the very end of the constructor?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -195,13 +204,25 @@ private void initRenderingSupport() {
...
+
+        // FIXME: playerCamera to context?
+        shadowMapNode = new ShadowMapNode(context, getMaterial("engine:prog.shadowMap"), playerCamera);
+        worldReflectionNode = new WorldReflectionNode(context, playerCamera, chunkShader);
+
+        renderingPipeline = Lists.newArrayList();
+        renderingPipeline.add(shadowMapNode);
+        renderingPipeline.add(worldReflectionNode);
+
+        renderableWorld.setShadowMapCamera(shadowMapNode.camera); //TODO: verify this assignment at this stage does not cause any problems
```
emanuele3d: Can this be done within the shadowMapNode?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -271,20 +261,19 @@ private void preRenderUpdate(RenderingStage renderingStage) {
         resetStats();
 
         currentRenderingStage = renderingStage;
-        if (currentRenderingStage == RenderingStage.MONO || currentRenderingStage == RenderingStage.LEFT_EYE) {
-            isFirstRenderingStageForCurrentFrame = true;
-        } else {
-            isFirstRenderingStageForCurrentFrame = false;
-        }
+        isFirstRenderingStageForCurrentFrame = (currentRenderingStage == RenderingStage.MONO) || (currentRenderingStage == RenderingStage.LEFT_EYE);
```
emanuele3d: I mentioned this already: I prefer the IF block as it was before as it's a crucial piece of code and it needs its "visual space."

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -271,20 +261,19 @@ private void preRenderUpdate(RenderingStage renderingStage) {
...
             timeSmoothedMainLightIntensity = TeraMath.lerp(timeSmoothedMainLightIntensity, getMainLightIntensityAt(playerCamera.getPosition()), secondsSinceLastFrame);
 
             playerCamera.update(secondsSinceLastFrame);
-            positionShadowMapCamera();
-            shadowMapCamera.update(secondsSinceLastFrame);
+
+            for (Node node : renderingPipeline) {
+                node.update(secondsSinceLastFrame);
+            }
+
```
emanuele3d: Let's remove the update() method for the time being.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -320,8 +309,11 @@ private void preRenderUpdate(RenderingStage renderingStage) {
     public void render(RenderingStage renderingStage) {
         preRenderUpdate(renderingStage);
 
-        renderShadowMap();          // into shadowMap buffer
-        renderWorldReflection();    // into sceneReflect buffer
+        if (renderingConfig.isDynamicShadows() && isFirstRenderingStageForCurrentFrame) {
+            shadowMapNode.process(); // into shadowMap buffer
+        }
+
+        worldReflectionNode.process(); // into sceneReflect buffer
```
emanuele3d: In this section, let's move the IF block INTO the shadowMapNode.process() method. Also, rather than processing each node explicitly, you can now iterate the list.

emanuele3d:
Code reviewed and feedback provided up to commit 98687b4.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
tdgunes: I think `update()` is needed for updating shadowMapCamera (for only shadowMap node) and setting FBOs to nodes. Therefore, I think removing `update()` method would break existing functionality.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
tdgunes: At this stage, every node depends on `WorldRendererImpl`, wouldn't be better to not make constructors overly crowded?

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
tdgunes: I would prefer removing `bindDisplay()` from `GraphicState` completely.

engine/src/main/java/org/terasology/rendering/dag/WorldReflectionNode.java
tdgunes: I didn't find any elegant solution for storing dimensions (`fullScale`) without a class like `GraphicState`. Any ideas?

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
tdgunes: Oh sorry, forgot to remove it. My bad! 😄

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
tdgunes: I am curious about a node like ShadowMapNode proposes its configuration besides just on/off, maybe different selectable techniques. These proposed configurations from nodes can be directly displayed in `VideoSettings` screen.

engine/src/main/java/org/terasology/rendering/dag/ShadowMapNode.java
tdgunes: I was trying to avoid something like:

```java
    @Override
    public Camera getLightCamera() {
        return shadowMapNode.shadowMapCamera;
    }
```
But since in the future we would not access shadowMapNode in this way. I think it's ok.

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -39,6 +41,15 @@
     float BLOCK_LIGHT_SUN_POW = 0.96f;
     float BLOCK_INTENSITY_FACTOR = 0.7f;
 
+
+    // TODO: appropriate javadocs
```
emanuele3d: Indeed! =)

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: Ok, let's leave it there then.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: So, can you please revert to the IF block?

GooeyHub:

msteiger:
Thanks for updating the code base @tdgunes. I sent a PR to your `dag` branch about the NodeCreator.

I'm hesitant to add new methods to `WorldRenderer`, unless really needed:

* The time-dependent part (`getSecondsSinceLastFrame`) could/should be part of the update cycle rather than being inside the rendering.
* getMaterial could be replaced with injection of AssetManager. This also provides other convenience methods as well as other Asset types.
* isFirstRenderingStageForCurrentFrame is too specific. Dependencies and ordering to other nodes can be dealt later with. 
* renderChunks() looks like an internal(=private) method to me. What you're interested in is another render pass with different parameters, right?

GooeyHub:

tdgunes:
@msteiger: Thanks for your feedback! For `getMaterial()`, It would be great if you can give an example. It seems like static `Assets` helper class is used much extensively in the codebase rather then `AssetManager.getAsset()`.

GooeyHub:
Yes, I would already prefer using `Assets` over `WorldRenderer`, but getting the actual asset manager from the context and using that would be even better (more flexibility, better testing caps.)

Cervator:
I think that last @GooeyHub was secretly @msteiger in disguise, or simply having forgotten to change accounts back again :D

I'll aim to test and merge this shortly to produce a new Omega build we can use for some play testing in a couple hours, unless there are any objections :-)

msteiger:
Whoops! That's correct - it was me.

PR #2337, opened 2016-06-03
Reduce the amount of openGL noise in world rendering

Josharias:
Applying a shader uniforms to not just "currentOnly" causes lots of openGL state changes because it has to enable all the other shaders individually,  set the value across the board,  then reload the original shader.  This starts to be a more prominent problem when each chunk does this multiple times in a frame.

With these changes,  I boosted my average FPS by 10+ and "mega" view distance became playable (over 20fps).

In measuring the amount of openGL calls,  I used gDEBugger and observed the difference in "OGL calls/frame".  The amount of OGL calls reduced by over 50%.

### Contains

- Reduction of openGL calls by only using "currentOnly = true" when setting shader uniform values in world rendering.  (NUI rendering is another issue.  But it does not seem to affect FPS like world rendering)
- Reduction of openGL calls by only enabling the shader once per set of chunks being rendered with ```renderChunk(...)``` (it now takes in the whole queue instead of one at a time).

### How to test

Run game, make sure all rendered stuff still renders correctly.

- [x] Test on another graphics card other than my own.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -56,6 +56,8 @@ import org.terasology.world.chunks.ChunkConstants;
 import org.terasology.world.chunks.ChunkProvider;
 import org.terasology.world.chunks.RenderableChunk;
 
+import java.util.PriorityQueue;
+
 /**
  * Renders the 3D world, including background, overlays and first person/in hand objects. 2D UI elements are dealt with elsewhere.
  *
@@ -80,6 +82,8 @@ public final class WorldRendererImpl implements WorldRenderer {
     private final BackdropProvider backdropProvider;
     private final WorldProvider worldProvider;
     private final RenderableWorld renderableWorld;
+    private final ShaderManager shaderManager;
+    private final EntityManager entityManager;
 
     private final Camera playerCamera;
     private final Camera shadowMapCamera = new OrthographicCamera(-SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, -SHADOW_FRUSTUM_BOUNDS);
@@ -145,6 +149,8 @@ public final class WorldRendererImpl implements WorldRenderer {
         this.renderingConfig = context.get(Config.class).getRendering();
         this.renderingDebugConfig = renderingConfig.getDebug();
         this.systemManager = context.get(ComponentSystemManager.class);
+        this.shaderManager = context.get(ShaderManager.class);
+        this.entityManager = context.get(EntityManager.class);
 
         if (renderingConfig.isOculusVrSupport()) {
             playerCamera = new OculusStereoCamera();
@@ -186,7 +192,7 @@ public final class WorldRendererImpl implements WorldRenderer {
         buffersManager.setPostProcessor(postProcessor);
         buffersManager.initialize();
 
-        context.get(ShaderManager.class).initShaders();
+        shaderManager.initShaders();
         postProcessor.initializeMaterials();
         initMaterials();
     }
@@ -372,9 +378,7 @@ public final class WorldRendererImpl implements WorldRenderer {
             graphicState.preRenderSetupSceneShadowMap();
             shadowMapCamera.lookThrough();
 
-            while (renderQueues.chunksOpaqueShadow.size() > 0) {
-                renderChunk(renderQueues.chunksOpaqueShadow.poll(), ChunkMesh.RenderPhase.OPAQUE, shadowMapCamera, ChunkRenderMode.SHADOW_MAP);
-            }
+            renderChunks(renderQueues.chunksOpaqueShadow, ChunkMesh.RenderPhase.OPAQUE, shadowMapCamera, ChunkRenderMode.SHADOW_MAP);
 
             for (RenderSystem renderer : systemManager.iterateRenderSubscribers()) {
                 renderer.renderShadows();
@@ -399,11 +403,7 @@ public final class WorldRendererImpl implements WorldRenderer {
 
         if (renderingConfig.isReflectiveWater()) {
             chunkShader.activateFeature(ShaderProgramFeature.FEATURE_USE_FORWARD_LIGHTING);
-
-            while (renderQueues.chunksOpaqueReflection.size() > 0) {
-                renderChunk(renderQueues.chunksOpaqueReflection.poll(), ChunkMesh.RenderPhase.OPAQUE, playerCamera, ChunkRenderMode.REFLECTION);
-            }
-
+            renderChunks(renderQueues.chunksOpaqueReflection, ChunkMesh.RenderPhase.OPAQUE, playerCamera, ChunkRenderMode.REFLECTION);
             chunkShader.deactivateFeature(ShaderProgramFeature.FEATURE_USE_FORWARD_LIGHTING);
         }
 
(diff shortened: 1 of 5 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -618,45 +608,50 @@ private void renderSimpleBlendMaterials() {
...
-                            chunkPosition.y * ChunkConstants.SIZE_Y - cameraPosition.y,
-                            chunkPosition.z * ChunkConstants.SIZE_Z - cameraPosition.z);
-
-            if (mode == ChunkRenderMode.DEFAULT || mode == ChunkRenderMode.REFLECTION) {
-                if (phase == ChunkMesh.RenderPhase.REFRACTIVE) {
-                    chunkShader.activateFeature(ShaderProgramFeature.FEATURE_REFRACTIVE_PASS);
-                } else if (phase == ChunkMesh.RenderPhase.ALPHA_REJECT) {
-                    chunkShader.activateFeature(ShaderProgramFeature.FEATURE_ALPHA_REJECT);
-                }
+    private void renderChunk(PriorityQueue<RenderableChunk> chunks, ChunkMesh.RenderPhase phase, Camera camera, ChunkRenderMode mode) {
```
Josharias: The diff does not do these changes justice.  Basically I moved up all the stuff that is the same for all the chunks (camera position, logic about render phase, and common uniforms), and then made sure all uniform setting used the "currentOnly = true".

Most importantly moved up the ```chunkShader.activateFeature(...)``` because that causes the shader to become dirty and re-enable the shader with openGL.

Cervator:
Checked the box for works on another video card. Looks fine to me :-)

Harder to give an impression of performance improvements without doing some more solid metrics, or even preparing right for Mega. It did run but slowly got more crawly until it finally ran out of memory ;-)

Didn't have much time to take it for a better spin, but so far so good! Sounds like a great improvement.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -438,18 +434,14 @@ private void renderObjectsOpaque() {
 
     private void renderChunksOpaque() {
         PerformanceMonitor.startActivity("Render Chunks (Opaque)");
-        while (renderQueues.chunksOpaque.size() > 0) {
-            renderChunk(renderQueues.chunksOpaque.poll(), ChunkMesh.RenderPhase.OPAQUE, playerCamera, ChunkRenderMode.DEFAULT);
-        }
+        renderChunk(renderQueues.chunksOpaque, ChunkMesh.RenderPhase.OPAQUE, playerCamera, ChunkRenderMode.DEFAULT);
```
emanuele3d: I like this simplification: good idea. However, the semantic of the method name is no longer correct. The method now deals with chunkS and should be renamed to renderChunks.

Josharias: Good catch.  Yes, plural is needed there for sure!

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -618,45 +608,50 @@ private void renderSimpleBlendMaterials() {
...
+                        new Vector3f(chunkPosition.x * ChunkConstants.SIZE_X - cameraPosition.x,
+                                chunkPosition.y * ChunkConstants.SIZE_Y - cameraPosition.y,
+                                chunkPosition.z * ChunkConstants.SIZE_Z - cameraPosition.z);
+
+                chunkShader.setFloat3("chunkPositionWorld",
+                        chunkPosition.x * ChunkConstants.SIZE_X,
+                        chunkPosition.y * ChunkConstants.SIZE_Y,
+                        chunkPosition.z * ChunkConstants.SIZE_Z,
+                        true);
+                chunkShader.setFloat("animated", chunk.isAnimated() ? 1.0f : 0.0f, true);
```
emanuele3d: If I'm interpreting the original code correctly, these two setFloat3 and setFloat lines should only run when mode is either ChunkRenderMode.DEFAULT or ChunkRenderMode.REFLECTION.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -618,45 +608,50 @@ private void renderSimpleBlendMaterials() {
         PerformanceMonitor.endActivity();
     }
 
-    private void renderChunk(RenderableChunk chunk, ChunkMesh.RenderPhase phase, Camera camera, ChunkRenderMode mode) {
```
Josharias: (re-adding this comment,  because I changed the target line and force-pushed)
The diff does not do these changes justice. Basically I moved up all the stuff that is the same for all the chunks (camera position, logic about render phase, and common uniforms), and then made sure all uniform setting used the "currentOnly = true".

Most importantly moved up the chunkShader.activateFeature(...) because that causes the shader to become dirty and re-enable the shader with openGL.

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -618,45 +608,50 @@ private void renderSimpleBlendMaterials() {
...
+            chunkShader.enable();
 
-                chunkShader.enable();
+        } else if (mode == ChunkRenderMode.SHADOW_MAP) {
+            shadowMapShader.enable();
 
-            } else if (mode == ChunkRenderMode.SHADOW_MAP) {
-                shadowMapShader.enable();
+        } else if (mode == ChunkRenderMode.Z_PRE_PASS) {
+            context.get(ShaderManager.class).disableShader();
```
emanuele3d: Minor issue from the original code, so not your fault @Josharias: any chance you can change this to obtain the ShaderManager on class initialization? Look in the method initRenderingSupport() where the ShaderManager is already obtained once from the context. Let's make that fill a private variable so we don't have to fetch it from the context every time.

emanuele3d:
Thank you very much for the changes @Josharias: I like how the code has been simplified and of I am very glad you found a way to obtain a considerable performance gain. @Cervator, make sure to highlight the performance improvements in the next release notes!

I am hopeful the next commit will render this PR good to go!

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
Josharias: Will do.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
Josharias: Also a nice catch.  Thanks!  That is what late night coding will miss.

Josharias:
Addressed all the comments from @emanuele3d.  Thanks for the review!

GooeyHub:

Cervator:
Proper test success! This time on EXTREME with 8 GB memory allocated. Spent something like half an hour just waiting for chunks to generate till I was past 5000 (alas, not over 9000) and only an occasional trickle would be deemed worthy to go visible for rendering.

After the "main" round of CPU-grinding generation (5-10 minutes) I was still at 15-20 FPS with nearly 2k chunks showing. I figure the CPU grinding is chunk gen and slowly increasing rendering is chunks going visible.

A few screenies, in chronological order

terasology _ alpha_2016-06-03_14-11-53

terasology _ alpha_2016-06-03_14-19-36

terasology - j__dev_terasology_git_integrate_terasology - engine - _engin_2016-06-03_14-27-09

Included IntelliJ's log in the final one to show auto save time stamps. At that point it was taking about 35s to auto-save, running once per minute :-) Not sure how much longer it would have kept growing if left alone.

Cervator:
Also added the run config for IntelliJ that I used, although we should probably still increase the default across the board (to maybe 1.5GB or so, still allowing for lower end systems or 32-bit Javas to function)

Issue #2329, opened 2016-05-26
showScreen() autocompletion breaks screen initialization


rzats:
### What you were trying to do

Open a NUI screen with the help of `showScreen` autocompletion (introduced in #2286).

### What actually happened

The screen opened fine, but was not properly initialized - more specifically, the [`ControlWidget.initialise()`](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/ControlWidget.java#L25) override for the screen was never called. Any further attempts to re-initialize the screen, even without using autocompletion to trigger the console command or without using the console at all, have subsequently failed - this is caused by the NUI asset already being loaded into memory when [`NUIManager.createScreen()`](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/internal/NUIManagerInternal.java#L258) is called; therefore, [`initialiseScreen()`](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/internal/NUIManagerInternal.java#L325) is not called [later on in the same method](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/internal/NUIManagerInternal.java#L268).

### How to reproduce

1. In your IDE of choice, place a breakpoint at the first line of [`MigTestScreen.initialise()`](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/MigTestScreen.java#L29). (Also works with any other UI screen.)
2. Launch the game in debug mode from the IDE. In the main menu, open the console and type `showScreen migTestScreen` WITHOUT using autocompletion. Verify that the breakpoint is triggered.
3. Shutdown and re-launch the game in debug mode. Re-open the console and type `showScreen mig`, then press Tab so that it autocompletes to `migTestScreen`. Verify that the breakpoint is NOT triggered and that some values are therefore not loaded properly.

Might be relevant to @tdgunes's interests.

Cervator:
Can confirm this issue. While testing #2288 (merged to latest develop as of right now) if I used auto-completion the MigTestScreen didn't look quite right and hit an NPE on `ESC` - seemingly breaking on trying to automate the return to the main menu or something.

Issue #2266, opened 2016-03-29
Implement support for S3 texture compression


msteiger:
I think it would be great if we could support S3 texture compression. This would greatly reduce the amount of required graphics memory. The standard is a few years old already and most gfx card suppliers should have adopted it. Still, I'd prefer to make this an optional feature.

https://www.opengl.org/wiki/S3_Texture_Compression

pollend:
Texture's are not quite self contained. There are a lot of hardcoded OpenGL calls in the code. This would need to be address first.

emanuele3d:
@pollend: Can you please elaborate? How is a texture self-contained or not? And how are S3 textures hampered by the existing opengl calls?

emanuele3d:
@msteiger: what do you use to gather the amount of graphics memory used? And how do you define "greatly reduce"? =)

pollend:
I just think it would probably be better to rework textures first before implementing this feature, but I don't believe it would hurt the existing calls. This should reduce texture memory, but is it really that necessary at the moment. Mobile would probably be the point where reducing texture memory would be important.

emanuele3d:
Please allow me to ask again:
* How is a texture self-contained or not?
* how are S3 textures hampered by the existing opengl calls?

Or to put it more succintly: what do you mean? ;)

pollend:
I believe that PBO's and FBO's in the current engine still declare their own texture bindings. I've been slowly reworking FBO's and would like to include them into the asset system. I want texture to be a bit more flexible in terms of what they can be declared to. There are quite a lot of formats that are available to openGL that could be integrated better with textures. 

https://www.opengl.org/wiki/Image_Format

emanuele3d:
@pollend: First of all, are you aware that PBOs are used only as 1x1 pixel buffers for readback purpose, as documented in the FrameBufferManager class? Because to put them in the same class with FBOs is a bit like trying to find commonalities between cats and dolphins: there are but it'd be very unusual to put them in the same enclosure.

Second, you write: "I believe [...] FBO's in the current engine still declare their own texture bindings." which is true but only half of the story. The other half is how you think they should be different and why. 

Now, you hint that FBOs should be included in the asset system. I know very little about the asset system and I look forward to your explanation about this. However, are you aware that most of the FBO instances in the rendering engine have nothing to do with assets and are "just" used to store the result of renderings which in turn are used as inputs for further rendering steps, i.e. postprocessing?

[...]

PR #2339, opened 2016-06-04
Quick fix for #2329, disposes loaded assets for correct initialization

tdgunes:
### Contains

My auto-complete PR(#2266) seems like broke proper initialisation of UIScreenLayer. An example for reproducing as mentioned in  #2329, when `migTestScreen` is loaded with auto completion, exiting the screen with `ESC` throws an Exception.  

My previous PR used `AssetManager`'s `getAvailableAssets()` method for suggesting available UIScreenLayer's for `showScreen()`. What I missed at that time, ~~`getAvailableAssets()`~~ `getAsset(...)` method tries to  initialise the assets and when `showScreen()` method is invoked with auto completion and proper initialisation steps are not considered.  A quick fix in this PR disposes these loaded assets, therefore screens are correctly loaded. 

### How to test

For testing, loading `migTestScreen` with auto completion(writing `showScreen mig` and hitting `TAB` key) and exitting the screen by hitting `ESC` must suffice. 

- [ ] A method in AssetManager, that I might be missing for getting available assets without requiring disposal

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/suggesters/ScreenSuggester.java
+++ b/engine/src/main/java/org/terasology/logic/console/suggesters/ScreenSuggester.java
@@ -15,6 +15,7 @@
  */
 package org.terasology.logic.console.suggesters;
 
+import com.google.api.client.util.Maps;
 import com.google.common.collect.Sets;
 import org.terasology.assets.ResourceUrn;
 import org.terasology.assets.management.AssetManager;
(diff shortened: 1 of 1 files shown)


GooeyHub:

msteiger:
Thanks for looking into this. It looks like the showScreen method does not work properly for UIElement parameters. I will have a look so we can discuss this tomorrow morning.

PR #2340, opened 2016-06-04
Do not load assets when suggesting screens

### LINKED ISSUE #2329: showScreen() autocompletion breaks screen initialization

msteiger:
Currently, every screen that is suggested is loaded into memory and tested if the root widget implements `UIScreenLayer`. I removed the test to avoid unnecessary loading and to avoid creating duplicates. It will suggest other UIElements as well - I think that there are none though.

Fixes #2329, supersedes #2339

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/suggesters/ScreenSuggester.java
+++ b/engine/src/main/java/org/terasology/logic/console/suggesters/ScreenSuggester.java
@@ -21,13 +21,15 @@ import org.terasology.assets.ResourceUrn;
 import org.terasology.assets.management.AssetManager;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.logic.console.commandSystem.CommandParameterSuggester;
-import org.terasology.rendering.nui.UIScreenLayer;
 import org.terasology.rendering.nui.asset.UIElement;
 
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 
+/**
+ * Suggest screen names. When only one screen resource exists for the entire set of loaded modules,
+ * the name alone suffices. Otherwise, the module name must be prepended.
+ */
 public final class ScreenSuggester implements CommandParameterSuggester<String> {
     private final AssetManager assetManager;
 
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Tested and merged :-)

Thanks @msteiger & @tdgunes !

PR #2341, opened 2016-06-05
Remove italic fonts - impossible for CJK fonts

### LINKED ISSUE #2330: Remove support for italics
Italic font style is not sensible in CJK and other Asian fonts. I therefore suggest removing it entirely to be consistent across all languages. It also makes font rendering somewhat easier.

It is currently not used in any part of engine/Core. There are potential use cases through the HTML paragraph builder, though.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/34688686-remove-support-for-italics?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

msteiger:
It looks like there are no implementations of `HTMLFontResolver` except for one in MarcinScIncubator, which doesn't look like a real module. Am I missing something here? What modules are using the HTML paragraph renderer?

Fixes #2330


GooeyHub:

Cervator:
MarcinScIncubator was meant to be @MarcinSc's new "misc" module rather than keep pouring extras into WoodAndStone. I suspect it has some unfinished stuff in it.

Likely this HTML stuff was aimed at the computer setup, such as the browser-based help system for commands and such. Although MarcinScIncubator at present only seems to be a dependency for BlockNetwork, which is *not* a dependency for the computer modules. Hmm. Maybe just not finished.

I tried running the game with Signalling and thus BlockNetwork and MarcinScIncubator active but it actually crashed on startup. So it would be a little hard testing the `htmlTest` command from `RegisterHTMLTester` that uses `HTMLFontResolver` 

Since it'll need fixing in either case we might just want to move past it for now.

Cervator:
Found and fixed the issue in Signalling. Correctly setup with a dependency on ModularComputers it runs and the `htmlTest` opens a UI - but I'm not sure it works particularly well. Brings a somewhat interesting thought to mind: we probably would want to explicitly exclude support for italics in in-game HTML-holding widgets (like the help system for computers). Assuming said help system would itself be translated sometime in the future.

msteiger:
Thanks! I reviewed all modules that come into question (Journal, ModularComputers, InGameHelp, Dialogs). None of them use the `HTMLFontResolver`, so I guess it's unused in general. As a result, I believe that this is save for merging.

Cervator:
Tested real quick with an added fix to the usage in MarcinScIncubator even though it may not functionally matter / be in use. I guess technically this would be an API change so I'll add that label, but cover it under the "trivial api change with fix" clause I have made up until we are more solid (and/or have more modules beyond our control)

Issue #2158, opened 2016-02-07
Tutorial module: Events, interactions, and inventory


Cervator:
Akin to #2153 and #2155 - a dedicated tutorial module for taking a step-by-step approach to learning how events work (beyond the super basics) and how to interact with something (inventory, interaction screen, etc)

Maybe #2153 will minimally touch on how to catch an event of interest, but just to show how you can get an entity and use a component attached to it. This one should go into details about you consume/cancel events, do more fancy things with them, and interact with some core systems like inventory.

We have a forum thread on interaction screens: http://forum.terasology.org/threads/interaction-screens.1106/ - it linked to a tutorial n the wiki which I removed in the great wiki purge intending to get something updated elsewhere, namely this dedicated tutorial!

The original tutorial is still available at https://github.com/MovingBlocks/TempDoc/blob/master/Tutorial:-How-to-make-blocks-with-custom-UI.md

Bonus: probably should wait and see what magic new inventory magic @Josharias is sprinkling on the engine before doing this :-)

We should also be sure to update / replace the forum thread when this is done.

I marked this with the multiplayer label thinking this might get advanced enough to touch on specific things to consider there. Unsure. Might be needed in the others too (like marking variables in Components for multiplayer)
##

Cervator:
Addition: With #2175 merged we now have some fun explosions to play with. Tutorials are always better if they contain explosions, right? 

Probably some good examples in there of configuration, the "Action" setup, and interactions between things.

Cervator:
Another angle: Show how a *layered* set of related systems might flow through a series of events differently. Thinking something like damage being modified due to equipment (both armor and weaponry), magic (shielding etc), and special modes like ghost/god mode.

Got to thinking about it while reviewing https://github.com/MovingBlocks/Terasology/pull/2323 which makes ghost/god mode not hurt the player, but by explicitly checking for that in the health system. The really old issue #652 that the PR fixes gets a little into event cancellation, which should probably be used instead, but I don't think we have any great examples or best practices for how to do so.

I did find a water breathing setup in the AlterationEffects module that uses event cancellation to prevent drowning damage, but I'm not sure how completed all that was. I think it was one of the things @MarcinSc started near the end of sorts of the active period for Throughout The Ages development (alchemy/potions) - unsure if usable, pinging in case of knowledge :-)

Pinging @portokaliu and @xrtariq2594 for general interest.

Cervator:
[...]

PR #2345, opened 2016-06-07
Allow modification of jump speed using events

Josharias:
### Contains

An event that allows modifying the character's jump speed based off some other factors.  This could be used for things like:
1. Training a jumping skill (this is what I personally plan on doing)
2. Wearing springy boots
3. Drinking a super jumping potion

### How to test

This change should not affect current jumping behavior but allow extension.  After this is merged,  I have a jumping skill that will be added to ManualLaborEventualSkills.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/AffectJumpForceEvent.java
+++ b/engine/src/main/java/org/terasology/logic/characters/AffectJumpForceEvent.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2016 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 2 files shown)


GooeyHub:

msteiger:
Looks good to me. It took me some time to understand what's going on. Maybe "speed" could be "force" and "get" could be "alter" or "adjust"? The class name would then be `AlterJumpForceEvent` - does that make sense?

Cervator:
Yeah the name throws me off a bit, especially the `Get` part. Either "alter" or "adjust" sound better to me as well. Maybe `AffectJumpForceEvent` ? Which can be modified by `Effect` type interactions.

Alter does goes well with AlterationEffects the module though. Which come to think of it has both interesting words.

This is a great example of how to code something the event-based way. I brought up #2158 as related on Slack as we really need to develop and emphasize some good practices in this area :-)

Also:

jumpgood

Cervator:
As for the subtly different feel in English (not that I'm a native speaker) 

* `Alter` suggests to me something like "This is totally getting changed somehow" 
* `Affect` instead is more vague, sort of a "this *may* get changed"

That may be completely subjective/arbitrary/wrong on my account though.

msteiger:
Both sound good to me.

GooeyHub:

Josharias:
Better event naming complete.

xrtariq2594:
This sounds like a great idea! Personally, more events to hook into and modify with other modules will open up the game more in terms of possibilities. This will also be useful for adding a JumpHeightPotion or JumpSpeedPotion into WaS.

Cervator:
Tested out, much jumping was performed!

@Josharias & @xrtariq2594 start your implementation engines!

PR #2347, opened 2016-06-08
Reduce chunk generation time by ~50%

Josharias:
### Contains

Cache the Air block so that looking up air by ```BlockUri``` is reduced. This is important during chunk generation when changing from unknown to a block because the unknown block is given back as Air.  Each time air is given back, it has to then do a lookup of the air block by ```BlockUri``` which eventually is looked up based on ResourceUrn hash code which is expensive when you do it for every single block in the known world. 

### How to test

Load a world, see how long it takes for all the chunk generation to cease.  This is different than having all chunks visible,  you must keep an eye on the chunk threads in the debug menu.  In my case I hooked up Yourkit profiler so that I could measure more accurately.

For my test I used the Builders World with a seed of "0" with Ultra view distance in a single player game. 

Before (look at the total duration selected) - 24sec:
cachedair before

After - 14sec:
cachedair after

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/internal/BlockManagerImpl.java
+++ b/engine/src/main/java/org/terasology/world/block/internal/BlockManagerImpl.java
@@ -73,6 +73,10 @@ public class BlockManagerImpl extends BlockManager {
     private boolean generateNewIds;
     private int nextId = 1;
 
+    // Cache this for performance reasons because a lookup by BlockURI happens the first time a block is set when getting the previous block.
+    // This causes performance problems eventually down the line when it then uses the ResourceUrn's hashcode to do a lookup into the block map.
+    private Block airBlock;
+
     public BlockManagerImpl(WorldAtlas atlas, AssetManager assetManager) {
         this(atlas, assetManager, true);
     }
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Awesome :-)

By the way, would it be worth a quick wiki page to detail how exactly to set up with profiling and such to find and test these performance improvements? If we make it easier to approach maybe more can help optimize.

Josharias:
Chuckle.  YourKit makes it pretty easy with Intellij.  Basically install YourKit and then press the button.

image

msteiger:
Great stuff! That is really valuable.

Would you mind adding a few line breaks in the comment to make it more readable? I usually wrap it at 120 so it fits nicely into GitHubs window. Our "official" code style limit is higher (175 or so) but imho this is too much.

Josharias:
Right.  I really must do code style analysis before submitting PRs.  Thanks for the reminder. Ill fix that.

GooeyHub:

Josharias:
Fixed comment line length.

Cervator:
Tested out - definitely feels better. At least 42 feels better. You can quote me on that, it is a very scientific term!

PR #2338, opened 2016-06-03
UITreeView (& test method conventions)

rzats:
### Contains

- A core UITreeView widget as described in [rzats/gsoc-nui-editor#1](https://github.com/rzats/gsoc-nui-editor/issues/1)!

image

- A `Tree` data structure that contains the widget's elements and a `TreeModel` helper class.

- `TreeViewTestScreen` - fairly self-explanatory :stuck_out_tongue_winking_eye:

- A codestyle fix renaming all the test methods to `testMethodName` as opposed to `methodName` - this is a mostly unrelated change as suggested in a Slack discussion.

_This PR was previously located in my Terasology fork. In retrospect, this wasn't a great idea: local PRs have no @GooeyHub testing, no mergeability checks with the upstream repository and little to none community exposure: IMO, going for a base repository fork is a better idea for subsequent deliverables._

### How to test

- Launch the game and type `showScreen treeViewTestScreen` in the in-game console. Make sure you do so manually without using autocompletion - otherwise #2329 triggers, causing initialization issues.
- Play around with the trees!

### Controls

- Mouse wheel: scroll up/down.
- LMB: select/deselect a node.
- RMB: expand/collapse a node.
- Delete: remove a node (and all its' children).
- Up/Down: change a node's position within the parent node.
- Ctrl+C: copy a selected node.
- Ctrl+V: paste the copied node as a child of the currently selected node.
- Ctrl+A / Insert: add a new child with a placeholder value to the currently selected node.

PR #2348, opened 2016-06-08
Fix silent UITreeView error when no item is selected

rzats:
### Contains

A quick fix for an issue with #2338: when a `UITreeView` item is expanded/contracted with no item having been selected, a silent NPE is thrown. This doesn't cause any issues with the component (apart from the performance loss associated with throwing/logging exceptions?) but is still unintended.


GooeyHub:

PR #2349, opened 2016-06-09
JSON/TreeView adapter + UITreeView expand/contract buttons

rzats:
### Contains

- Small buttons are now drawn within a `UITreeView` that expand/contract a node (unless it is a leaf, in which case no button is drawn). This supersedes the previous method of expanding/contracting nodes (by clicking on an individual items) that didn't make it clear whether an item is _expanded_ or _expandable_ at all.

- A `JsonTreeAdapter` utility class has been created that can serialize a GSON `JsonElement` into a `Tree` (and vice versa). `treeViewTestScreen` now displays its' own interface definition file instead of a hardcoded tree.

expand-contract-adapter

### How to test

Run `showScreen treeViewTestScreen` to launch the test screen, then click on the buttons!

- [ ] Better button textures - the current ones are salvaged from `UIDropdown` and don't look that great. A +/- sign instead of an arrow might also make more sense.


GooeyHub:

Cervator:
I have clicked the buttons with great prejudice! No issues to report :-)

Will give this PR some time in case @flo wants to check out the code in more detail, I can just report that it seems to test out fine.

GooeyHub:

Cervator:
And this is when I remember @flo mentioned he'd be away this week and I spaced and let this wait anyway :D

As per Slack there's another piece coming tomorrow, I'll do another round of review then likely merge if all is well, we can always improve later after all :-)

GooeyHub:

flo:
The code looks good to me. The only thing I noticed is the missing attribute name for attributes of type object or array.

Like address or phoneNumbers in:
https://github.com/rzats/gsoc-nui-editor/blob/master/Proposal/media/image_1.png

flo:
I noticed that the screenshot above has more (attribute names) than I get when I open it:
http://postimg.org/image/dpzdksv2j/

GooeyHub:

flo:
@rzats thanks for the quick fix :)

I retested with your latest patch: The collapse buttons are now only visible when you hover them. Looks to me like a regression, but it is kind of unexpected to me based on your latest change oO

Btw. I would also like to discuss the current way of representing json data with you. I am not sure how you would want to add json (and later UI json) specific logic to a node. e.g. currently I can copy a object node from an array node and paste it in an object node as attribute (which would result in invalid json).

Cervator:
Tested it out again and ran the unit tests too. Merging as discussed on IRC earlier, so we can continue in a new PR next :-) I think the discussion there also covered the weirdly visible or not issue mentioned above? I didn't encounter the issue, looked fine.

Observations:
* The disabled side still actually allows interaction (is that just not added yet?)
* Would it make sense to have a little icon or graphic for leaf nodes, so you can see the tree via icons? Some other tree views elsewhere in software land do that (yet others do not)
* Concur with the + / - stuff.

rzats:
@Cervator:

- I've personally interpreted the `disabled` state as the widget being view-only, so expanding/contracting nodes works but everything else doesn't. This is easy to tweak, though.
- Not entirely sure what you mean re: tree icons - is there a specific TreeView that uses this? Icons is general is something I'm planning on implementing later: https://github.com/rzats/gsoc-nui-editor/issues/7

Cervator:
> I've personally interpreted the disabled state as the widget being view-only

That's a good point. I missed that. Although maybe somebody would want to "lock" a tree view with a specific state? Unlocking it if needed. Maybe there is potential for more than one state out of enabled/disabled, locked/unlocked, read/write? I don't want to over complicate things before we actually have use cases though.

My other note was more thinking out loud. One of your classes has a little ASCII tree showing the structure, and some trees have something like `\- entry` in addition to the + or - by nodes. I'm not sure if either approach is more or less ideal. There isn't anything wrong with the current approach :-)

PR #2355, opened 2016-06-13
Tree architecture improvements

rzats:
### Contains

Several improvements to the `org.terasology.rendering.nui.widgets.models.Tree` class and related utility classes, as discussed on Slack.

- `JsonTreeAdapter` has been renamed to `JsonTreeConverter`.
- The existing `Tree<T>` class has been split into a `Tree<T>` abstract class (includes some default methods and the `DepthFirstTreeIterator`) and a simple `GenericTree<T>` implementation.
- A `JsonTree` class has been added. It's a `Tree<JsonTreeNode>` implementation that generally copies `GenericTree<T>` but also includes some initial JSON-specific validation (i.e. child elements cannot be added to `PRIMITIVE` or `NULL` nodes).
- `TreeViewTestScreen` and `JsonTreeAdapter` have been modified to use `JsonTree`, while `TreeTest` now utilises `GenericTree<Integer>`.

### How to test

`showScreen treeViewTestScreen`, then verify that
- children cannot be added (Ctrl+A or Ctrl+C/Ctrl+V) to non-object/array nodes;
- no regressions relative to #2349 are present.


GooeyHub:

flo:
A few things I noticed:
* It is possible to add attribute nodes to arrays, which doesn't make json wise sense too.
* Copying collapsed elements in an array does not seem to work.
* Copying an attribute even if it does not exist at the target does not seem to work (it's okay not to allow it for now, I just noticed it).

For drag and drop of items via mouse it will probably be later useful to have a non expensive way of checking if a node can be added/moved to another node. Exception throwing and catching is expensive in java. Instead you could for example add a method `boolean canChildBeAdded(int index)` or something like that. Although I don't see that as high prio, compared to staying on time plan.

rzats:
@flo: re: attribute node children in arrays / `Preconditions` usage - this is a fairly low-effort fix, so I went ahead and implemented it :)

re: copying collapsed elements in arrays - I've tried copying/pasting collapsed array nodes (i.e. _contents → {}_) in the demo and they are pasted just fine. Is there a way to replicate this?

GooeyHub:

rzats:
Turns out that the existing check also invalidates [arrays of primitive values](https://github.com/MovingBlocks/Terasology/blob/develop/engine-tests/src/test/resources/example.json#L8-L16) (as opposed to arrays of key-value pairs), causing the test to fail. I've rolled back this particular check for now - going to investigate tomorrow.

One thing to consider is that GSON doesn't seem to have a way to distinguish between a key-value pair and a simple value - both of them are marked as `JsonPrimitive`.

GooeyHub:

flo:
@rzats ok, copying of collapsed nodes worked. I assumed wrongly that a copy & paste after each other would result in a copied object, but I need to switch to parent to do so

Pasting objects within objects works, although it is not valid json as far as I know(even with patch).

I would put `acceptsChild` in the parent class and document it in the javadoc of the `Tree#addChild` methods that `acceptsChild` needs to be true for added elements and what happens when it is not true(throwing an exception for invalid method usage is okay).

About the current split of the logic between Tree, GenricTree and JsonTree I am also not totally convinced but if you like it it's okay from my side. Questions I had in my mind when I reviewed it: 
* Is there a reason for still having a getValue method in Tree or could it be a GenericTree only method? e.g. could JosnNode and JsonTree be merged in a single class for simplicity? It is btw. possible to have a class that has methods that are about itself. See Comparator as an example
* Why are some Tree methods abstract when the implementations are the same? Maybe GenericTree and JsonTree could be greartly simplified by having not to override all methods.

engine/src/main/java/org/terasology/rendering/nui/widgets/models/JsonTreeNode.java
```diff
@@ -21,8 +21,20 @@
  * The value type for the tree representation of a {@link JsonElement}.
  */
 public class JsonTreeNode {
+    /**
+     * The type of the JSON node.
+     */
     public enum ElementType {
-        PRIMITIVE, ARRAY, OBJECT, NULL
+        // Primitive data type (string, boolean, array).
```
flo: you oculd use a javadoc doc here. It is also possible as one liner if you wanted that e.g.`/** Some javadoc */`

GooeyHub:

GooeyHub:

flo:
thanks for the pull request

PR #2356, opened 2016-06-15
Major change to DelayedActionSystem.

xrtariq2594:
### Contains

Minor yet important alteration to how the delayedActionComponent and PeriodicActionComponent are checked in the addDelayedAction and addPeriodicAction methods. 

In short, even if the old earliest operation wakeup time of the "...OperationsSortedByTime" is not less than the new earliest operation wakeup time, the new action should still be added to delayedOperationsSortedByTime/periodicOperationsSortedByTime. This will allow the event to be fired off when it's duration expires.

### How to test

Call addDelayedAction with various delay values, and see if the approraite events are sent out at the correct time. 

- [x] Need to consider all uses of addDelayedAction/addPeriodicAction in the engine.
- [ ] Need to consider uses of addDelayedAction/addPeriodicAction in other modules..
- [x] Need to perform unit testing.

### DIFF
--- a/engine-tests/src/test/java/org/terasology/logic/delay/DelayedActionSystemTest.java
+++ b/engine-tests/src/test/java/org/terasology/logic/delay/DelayedActionSystemTest.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2016 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.logic.delay;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+import org.terasology.TerasologyTestingEnvironment;
+import org.terasology.engine.Time;
+import org.terasology.entitySystem.entity.EntityManager;
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.event.ReceiveEvent;
+import org.terasology.logic.location.LocationComponent;
+import org.terasology.registry.CoreRegistry;
+import org.terasology.registry.CoreRegistryTest;
+import org.terasology.registry.In;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class DelayedActionSystemTest extends TerasologyTestingEnvironment {
+    private DelayedActionSystem delayedActionSystem;
+    private List<Integer> vals; // Use this for ordering the expected values.
+
+    @In
+    private Time time;
+
+    long nextFakeEntityId = 1;
+    int lookingForId = 0; // Use this for ordering the time events. 0 is earliest.
+
+    @Before
+    @Override
+    public void setup() throws Exception{
+        super.setup();
+
+        delayedActionSystem = new DelayedActionSystem();
+        nextFakeEntityId = 1;
+        lookingForId = 0;
+
+        time = mockTime;
+        delayedActionSystem.setTime(mockTime);
+    }
+
+    private EntityRef createFakeEntityWith(ArbritaryDelayActionComponent arbritaryDelayActionComp) {
+        EntityRef entRef = mock(EntityRef.class);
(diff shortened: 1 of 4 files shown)


GooeyHub:

Cervator:
Tested out with the basic setup in TTA brewing potions (along with latest Workstation) and it seemed fine there. No time tonight to dig deeper with custom set timers and such, although I suspect the unit tests would ultimately be a better way to validate anyway. Looking forward to those very much :-)

GooeyHub:

xrtariq2594:
Alright, I added five unit tests in my last push. Though, I needed to apply some extra changes in order for it to work.

engine-tests/src/test/java/org/terasology/logic/delay/DelayedActionSystemTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class DelayedActionSystemTest extends TerasologyTestingEnvironment {
+    private DelayedActionSystem delayedActionSystem;
+    private List<Integer> vals; // Use this for ordering the expected values.
+
+    @In
```
Cervator: You probably don't need the `@In` here since you set it in the constructor?

engine/src/main/java/org/terasology/logic/delay/DelayedActionSystem.java
```diff
@@ -46,6 +46,12 @@
     private SortedSetMultimap<Long, EntityRef> delayedOperationsSortedByTime = TreeMultimap.create(Ordering.natural(), Ordering.arbitrary());
     private SortedSetMultimap<Long, EntityRef> periodicOperationsSortedByTime = TreeMultimap.create(Ordering.natural(), Ordering.arbitrary());
 
+    // ONLY use this for testing. DO NOT use this during regular usage.
+    public void setTime(Time t)
```
Cervator: Might this be a valid spot to use package private level access instead of public? Since the test is in the same package just under engine-tests? Can't entirely recall if it works that way.

engine/src/main/java/org/terasology/logic/delay/DelayedActionSystem.java
```diff
@@ -77,7 +83,7 @@ private void invokeDelayedOperations(long currentWorldTime) {
             }
 
             for (String actionId : actionIds) {
-                delayedEntity.send(new DelayedActionTriggeredEvent(actionId));
+                 delayedEntity.send(new DelayedActionTriggeredEvent(actionId));
```
Cervator: Weird 5 instead of 4 spaces here

engine-tests/src/test/java/org/terasology/logic/delay/DelayedActionSystemTest.java
```diff
@@ -0,0 +1,178 @@
...
+    public void test1DelayedAction() {
+        ArbritaryDelayActionComponent a1 = new ArbritaryDelayActionComponent();
+        a1.value = 3;
+
+        vals = new ArrayList<Integer>(Arrays.asList(3));
+
+        long currentTime = time.getGameTimeInMs();
+
+
+        delayedActionSystem.addDelayedAction(createFakeEntityWith(a1),
```
Cervator: I don't think you need a line break here. 120ish chars is OK.

Cervator:
Thanks a lot for adding tests :-) I made a few line-level comments.

This is a little outside my realm of familiarity, so hoping @immortius @flo or maybe @Josharias could take a look.

Specifically on the unit tests - I see you have the `finishWaiting` method and use assertion there when scheduled actions trigger, but how does that get tied to the individual methods marked with `@Test` ? Wouldn't they just always pass assuming the actions succeed being scheduled? Maybe that's intentional. I dunno! :)

engine/src/main/java/org/terasology/logic/delay/DelayedActionSystem.java
xrtariq2594: It works :).

GooeyHub:

Josharias:
Seems good to me. It is worth noting that there is possibly another solution to this with catching the component changed event.  Possibly it is an even worse solution,  but mentioning it.

The test does not actually wait long enough for the delayed actions to trigger.  Presently it is only just testing to see if one can add delayed actions to the system,  not actually waiting for the results.

That said,  thank you for the inspiration to get my module code unit tested.

xrtariq2594:
Darn. How much time does the test wait until it auto-completes?

Cervator:
> Darn. How much time does the test wait until it auto-completes?

It doesn't :-) I was mentioning that on Slack earlier. Unit tests don't tend to wait for anything, they just execute and if something asserts you have a result. I'm not sure how to best validate a timed execution or something. You have a method that captures when the timers trigger but that doesn't tie back to the initial `@Test` method so the unit testing results won't necessarily look right.

Cervator:
Counting this as tested about as well as it can get now after today's multiplayer test session. No seeming issues noted then or earlier :-)

I'm sure we can improve the area more later especially on the unit test angle, but no need to delay (get it? "delay" ? hah, hah ...) this PR further for it.

Thanks!

Issue #97, opened 2012-02-03
Skysphere: Special effects


Cervator:
So this one is one of the rare times where I'll let a personal indulgence slip in - I finally started playing Skyrim and those northern lights... dang ;-)

I wonder how much work it would be for us to allow special events to appear in the sky, such as northern lights (maybe only in cold biomes?), meteor showers, eclipses, etc? Maybe somebody could find out!

They should of course eventually be rare within the game, not constant. Could be a hook here to allow events in the world to be impacted by Skysphere state - such as primitive beasts flipping out during a meteor shower, or a particular race reacting dramatically to an eclipse / full moon / odd temporary discoloration of a celestial
##

soloturn:
> **AI-assisted update.** Filed by agent driven by @soloturn via [GDD](https://github.com/soloturn/yggdrasil).

Implemented the meteor shower option (picked over aurora/eclipse - most tractable with the existing particle system, no new rendering-pipeline work needed): a rare, per-night chance of a brief burst of streaking particles above each connected player.

- Engine (`MeteorShowerSystem`, the trigger/spawn logic): #5388
- CoreAssets (the particle prefab + texture + test): Terasology/CoreAssets#12

Aurora and eclipse remain unimplemented - both would need real rendering-pipeline changes (an animated shader for aurora, sun/moon alignment detection for eclipse) that need live visual verification I couldn't reliably do in this environment. The "creatures reacting to it" hook from the original request is also not built - nothing currently listens for a shower starting/ending; that'd be a natural follow-up once the base effect is confirmed to look right in-game.

Not closing this - the PRs aren't merged yet, and the aurora/eclipse/creature-reaction parts of the original ask are still open regardless.

PR #2362, opened 2016-06-18
[RFR] Backdrop and Skybands Nodes for DAG

tdgunes:
### Contains

Two new nodes for DAG (Directed Acyclic Graph), wrapping tasks that are related with rendering Sky. 

### How to test

Check if there are any weirdness related to rendering of Sky.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/BackdropNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/BackdropNode.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2016 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.dag;
+
+import org.terasology.config.Config;
+import org.terasology.config.RenderingDebugConfig;
+import org.terasology.monitoring.PerformanceMonitor;
+import org.terasology.registry.In;
+import org.terasology.rendering.backdrop.BackdropRenderer;
+import org.terasology.rendering.cameras.Camera;
+import org.terasology.rendering.opengl.FBO;
+import org.terasology.rendering.opengl.FrameBuffersManager;
+import org.terasology.rendering.world.WorldRenderer;
+
+import static org.lwjgl.opengl.GL11.*;
+import static org.terasology.rendering.opengl.OpenGLUtils.bindDisplay;
+import static org.terasology.rendering.opengl.OpenGLUtils.enableWireframeIf;
+import static org.terasology.rendering.opengl.OpenGLUtils.setRenderBufferMask;
+
+/**
+ * TODO: Diagram of this node
+ */
+public class BackdropNode implements Node {
+
+    @In
+    private Config config;
+
+    @In
+    private BackdropRenderer backdropRenderer;
+
+    @In
+    private WorldRenderer worldRenderer;
+
+    @In
+    private FrameBuffersManager frameBuffersManager;
+
+    private RenderingDebugConfig renderingDebugConfig;
+    private Camera playerCamera;
+    private FBO sceneOpaque;
+
+    @Override
+    public void initialise() {
+        renderingDebugConfig = config.getRendering().getDebug();
+        playerCamera = worldRenderer.getActiveCamera();
+    }
+
+    @Override
+    public void process() {
+        PerformanceMonitor.startActivity("Render Sky");
+        sceneOpaque = frameBuffersManager.getFBO("sceneOpaque");
+
+        enableWireframeIf(renderingDebugConfig.isWireframe());
+        initialClearing();
+
+        sceneOpaque.bind();
+        setRenderBufferMask(sceneOpaque, true, true, true);
+
+        playerCamera.lookThroughNormalized();
+        /**
+         * Sets the state to render the Backdrop. At this stage the backdrop is the SkySphere
+         * plus the SkyBands passes.
+         *
+         * The backdrop is the only rendering that has three state-changing methods.
+         * This is due to the SkySphere requiring a state and the SkyBands requiring a slightly
+         * different one.
+         */
+        setRenderBufferMask(sceneOpaque, true, false, false);
+        backdropRenderer.render(playerCamera);
+    }
+
+    /**
+     * Initial clearing of a couple of important Frame Buffers. Then binds back the Display.
+     */
+    // It's unclear why these buffers need to be cleared while all the others don't...
+    private void initialClearing() {
+        FBO sceneReflectiveRefractive = frameBuffersManager.getFBO("sceneReflectiveRefractive");
+        sceneOpaque.bind();
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+        sceneReflectiveRefractive.bind();
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+        bindDisplay();
+    }
+}
--- a/engine/src/main/java/org/terasology/rendering/dag/BackdropNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/BackdropNode.java
@@ -26,9 +26,7 @@ import org.terasology.rendering.opengl.FrameBuffersManager;
 import org.terasology.rendering.world.WorldRenderer;
 
 import static org.lwjgl.opengl.GL11.*;
-import static org.terasology.rendering.opengl.OpenGLUtils.bindDisplay;
-import static org.terasology.rendering.opengl.OpenGLUtils.enableWireframeIf;
-import static org.terasology.rendering.opengl.OpenGLUtils.setRenderBufferMask;
+import static org.terasology.rendering.opengl.OpenGLUtils.*;
 
 /**
  * TODO: Diagram of this node
@@ -59,10 +57,11 @@ public class BackdropNode implements Node {
 
     @Override
     public void process() {
-        PerformanceMonitor.startActivity("Render Sky");
+        PerformanceMonitor.startActivity("rendering/backdrop");
+        enableWireframeIf(renderingDebugConfig.isWireframe());
+
         sceneOpaque = frameBuffersManager.getFBO("sceneOpaque");
 
-        enableWireframeIf(renderingDebugConfig.isWireframe());
         initialClearing();
 
         sceneOpaque.bind();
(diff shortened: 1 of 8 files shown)


GooeyHub:

Cervator:
I can already feel #97 moving ever so slightly! Aurora Borealis here we come! :fireworks: 

Tested it out just a bit to look at the sky in varying conditions (changing time of day, video settings) and it still looks like good ole sky.

engine/src/main/java/org/terasology/rendering/dag/SkyBandsNode.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        sceneOpaque.bind();
+
+        playerCamera.lookThrough();
+        endRenderSkyActivity();
+    }
+
+    private void endRenderSkyActivity() {
+        PerformanceMonitor.endActivity();
```
emanuele3d: Good coding choice to clarify what would otherwise be hard to understand. However I'd solve the problem by having Performance Monitor statements at the beginning and end of each node. This is a change from what we have now, grouping together a number of rendering steps within the same activity. My vision is to group together activities by using hierarchical activity names instead. I.e. you could name the activity in the BackdropNode "rendering/backdrop" and the other "rendering/backdrop/skybands". It will be up to DebugOverlay to group these together if the user wishes it. Let me know if this is clear enough.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
```diff
@@ -15,45 +15,13 @@
...
-import static org.lwjgl.opengl.GL11.GL_STENCIL_BUFFER_BIT;
-import static org.lwjgl.opengl.GL11.GL_STENCIL_TEST;
-import static org.lwjgl.opengl.GL11.glBlendFunc;
-import static org.lwjgl.opengl.GL11.glClear;
-import static org.lwjgl.opengl.GL11.glCullFace;
-import static org.lwjgl.opengl.GL11.glDepthMask;
-import static org.lwjgl.opengl.GL11.glDisable;
-import static org.lwjgl.opengl.GL11.glEnable;
-import static org.lwjgl.opengl.GL11.glStencilFunc;
+import static org.lwjgl.opengl.GL11.*;
```
emanuele3d: I think the project's coding standards are against this. Did you use Checkstyle?

engine/src/main/java/org/terasology/rendering/dag/SkyBandsNode.java
tdgunes: You mean like:
``` java
PerformanceMonitor.startActivity("rendering/backdrop") // Executed inside Backdrop
...
PerformanceMonitor.startActivity("rendering/backdrop/skybands") // Executed inside SkyBand node
... 
PerformanceMonitor.endActivity("rendering/backdrop/skybands") // Executed inside SkyBand 

PerformanceMonitor.endActivity("rendering/backdrop") // Executed not inside SkyBand, but inside Backdrop
```

But the activity backdrop (`rendering/backdrop`) in total is going to include measurements from `rendering/backdrop/skybands` right?

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
tdgunes: Just rechecked. Seems like `import static` statements with wild cards are not against CheckStyle.

engine/src/main/java/org/terasology/rendering/dag/SkyBandsNode.java
```diff
@@ -0,0 +1,112 @@
...
+    public void initialise() {
+        renderingConfig = config.getRendering();
+        renderingDebugConfig = renderingConfig.getDebug();
+        blurShader = worldRenderer.getMaterial("engine:prog.blur");
+        playerCamera = worldRenderer.getActiveCamera();
+    }
+
+    @Override
+    public void process() {
+        enableWireframeIf(renderingDebugConfig.isWireframe());
```
emanuele3d: Please enable AND disable wireframe in all nodes that were originally between the enableWireframeIf/disableWireframeIf statements in the WorldRendererImpl.render method. So, each of those nodes will need its own enable/disable pair. This of course has a small performance hit but we'll eventually solve it with state management.

tdgunes: So you mean these two nodes also need a `disableWireframeIf` statement at the end of `process()` method? Should changed states for a certain node in OpenGL be reverted and left for our state management?

engine/src/main/java/org/terasology/rendering/opengl/PostProcessor.java
```diff
@@ -317,9 +308,9 @@ private void generateSkyBand(FBO skyBand) {
         }
 
         skyBand.bind();
-        graphicState.setRenderBufferMask(skyBand, true, false, false);
+        setRenderBufferMask(skyBand, true, false, false);
 
-        setViewportTo(skyBand.dimensions());
+        setViewportToSizeOf(skyBand.dimensions());
```
emanuele3d: In the previous PR we used a version of this method that had in input the FBO, not its dimension. So, this line would read: setViewportToSideOf(skyBand). Can we use that version instead?

engine/src/main/java/org/terasology/rendering/dag/SkyBandsNode.java
emanuele3d: - The answer to the first question is yes.
- OpenGL state changes shouldn't be touched at this stage. I was very, very tempted of touching some in the SkyBands node but I decided not to suggest it so that we can continue focusing on wrapping things.
- PerformanceMonitor is not an opengl thing so I don't mind touching it. The content of every process() method of every node should really be bracketed by startActivity()/endActivity() as a rule.

engine/src/main/java/org/terasology/rendering/opengl/PostProcessor.java
tdgunes: By the way, @emanuele3d important to note that I left some of the wrapped and unused methods for removal in the future, after all nodes are implemented. My motive is [making it work/clean/efficient ](http://c2.com/cgi/wiki?MakeItWorkMakeItRightMakeItFast)(in order).

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/SkyBandsNode.java
emanuele3d: Regarding the activity statements: correct. Just remember endActivity() takes no parameter. Which won't be a problem now that start/end are closer to each other.

Regarding your question: ideally, but only in a DebugOverlay. And initially I don't mind if it isn't. If the DebugOverlay is ordered alphabetically for example, related nodes will be listed near each other anyway.

There could be an argument for naming rendering/backdrop more specifically. Or perhaps naming rendering/backdrop/skybands to rendering/skybands instead. In fact, I would prefer it that way. I am not convinced that the skybands should be grouped with the backdrop. If I remember correctly the generate some distant haze that gets overimposed on the world rendering. So, they are atmospheric effects on the world, not on the sky. 

So, I'd suggest rendering/backdrop and rendering/skybands for the time being.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
emanuele3d: Let's ask @msteiger or @Cervator, they know better.

engine/src/main/java/org/terasology/rendering/opengl/PostProcessor.java
emanuele3d: As a philosophy sounds good. Which unused methods are you referring to?

tdgunes: Such as `public void generateSkyBands()` method in `PostProcessor`. But all of these will be removed once all nodes are in place. Best to begin the huge refactor then.

emanuele3d: Hmmmm... I do not see the rationale of leaving unused methods. There might be an argument for methods used elsewhere, but if you are moving a method from one place to another I do not see an argument in keeping the old version, especially if it's not used anywhere else.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
Cervator: Static imports are considered "different" somehow, but I can't actually remember why. A quick Google search and a page find for "static" gets you a little bit of reasoning: http://stackoverflow.com/questions/2067158/what-is-the-proper-style-for-listing-imports-in-java

That thread also finally made me realize why regular import star is bad - not just for some optimization reason, but if new classes are added that would suddenly qualify due to the wildcard bad things can suddenly happen. Which differs with static imports... somehow.

emanuele3d: So..... import star for static imports is ok? Can you confirm?

engine/src/main/java/org/terasology/rendering/opengl/PostProcessor.java
tdgunes: I see, my next changes after @msteiger's review will not include those. 👍

emanuele3d:
@msteiger! We will be waiting for your words of wisdom. 
@tdgunes: do you already have a branch for the next few nodes? To familiarize myself with what's coming?

tdgunes:
@emanuele3d Yes! I have _prepost-composition_ nodes in [this branch](https://github.com/tdgunes/Terasology/commits/dag-prepostcomposite).

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
Cervator: Can't confirm completely, no. I wouldn't go out of my way to star import statics if they're split out already, but it may make sense in general. @immortius could probably call it, or @flo / @msteiger

engine/src/main/java/org/terasology/rendering/dag/BackdropNode.java
```diff
@@ -0,0 +1,96 @@
...
+import org.terasology.config.RenderingDebugConfig;
+import org.terasology.monitoring.PerformanceMonitor;
+import org.terasology.registry.In;
+import org.terasology.rendering.backdrop.BackdropRenderer;
+import org.terasology.rendering.cameras.Camera;
+import org.terasology.rendering.opengl.FBO;
+import org.terasology.rendering.opengl.FrameBuffersManager;
+import org.terasology.rendering.world.WorldRenderer;
+
+import static org.lwjgl.opengl.GL11.*;
```
msteiger: No wildcard imports, please. See [checkstyle.xml](https://github.com/MovingBlocks/TeraConfig/blob/master/checkstyle/checkstyle.xml#L96-L101) for the check.

engine/src/main/java/org/terasology/rendering/opengl/GraphicState.java
msteiger: I don't see any advantage of using wildcard imports. Any reasonable IDE would fold imports, so you don't even see them. Most notable drawback imho is that you don't immediately see which class is used if multiple wildcards are used. You can also run into conflicts if two classes with the same name are imported.

msteiger:
Looks good to me in general. Anything specific (apart from the wildcard imports) that I can contribute to?

GooeyHub:

msteiger:
Thanks!

emanuele3d:
Yay!!!

Issue #2043, opened 2015-12-06
Clear out remaining (old) deprecations


Cervator:
_Edit: Moved to v2.0.0 as v1.0.0 was released. The engine and Core entries may not be doable early_

Before we go v1.0.0 with engine we should clean out the remaining spots where code uses deprecated methods from elsewhere in engine, and finally delete those methods. That way we have a clean slate to start the major release with where any future deprecations will stand out better.

_Deprecation warnings in engine + standard module lineup_
- [ ] engine: 17
- [ ] Core: 1
- [ ] SimpleFarming: 1
- [ ] Maze: 3
- [ ] MarkovChains: 1
- [ ] WeatherManager: 3 (okay, technically not in the lineup but I have it in my workspace)
- [ ] ItemRendering: 1
- [ ] CustomOreGen: 2
- [ ] Pathfinding: 5
- [ ] JoshariasSurvival: 1
- [ ] ClimateConditions: 2
- [ ] LightAndShadow: 10
- [ ] Miniion: 4 (but we might as well retire that one entirely already)
- [ ] BlockPicker: 1
- [ ] AnotherWorld: 14
- [ ] GrowingFlora: 12
- [ ] WoodAndStone: 3

Pinging @msteiger @Josharias @MarcinSc @synopia 
##

Cervator:
Engine cleared up some by #2062

skaldarnar:
Clearing out deprecations is something we should do with an upcoming revive, but the information in this ticket is likely outdated, thus, closing it.

PR #2365, opened 2016-06-20
removing obsolete interfaces and classed

adam83112:
### Contains

Removing obsolete Noise2D and Noise3D interfaces, and some onsolete Noise classes.


GooeyHub:

Cervator:
Hi @adam83112 and thanks for the PR :-)

Was this inspired by #2043? If so I must apologize for not having moved it to engine v2.0.0 yet. We didn't get it finished in time for the v1.0.0 release and there are still usages in the module ecosystem (like WeatherManager), so we can't remove them yet.

Since we're still in Alpha we're considering API violations on a case-by-case basic, if they're relatively trivial. If each current module using these classes was updated first we might be able to merge it early, but I'm not sure how much effort that would be.

adam83112:
No, I just wanted to remove some warnings and unused code. But if you have a different approach, these changes can wait.

Cervator:
We definitely want to remove warnings and unused code, yep. We are just also trying to hold ourselves to the principles of [Semantic Versioning](http://semver.org) :-)

It is generally easier to fix Checkstyle, Findbugs, and related warnings, as those generally do not impact API. The thing to be careful with there is not changing too many files at once as then that sort of work could impact functional changes other people are working on, causing conflicts.

I'll go ahead and close this, unless you are up for hunting down and adjusting usages across all [game modules](https://github.com/Terasology) as well - but it might be tricky even finding exactly which modules are using these classes. If you're interested let me know and I can help make a list and we can re-open this.

In any case thanks for your interest! I hope you might like to do more code, will be happy to come up with ideas if needed :-)

adam83112:
I think you are right. It is too early for cleanup code. I will take some of existing task, but it will be quite easy because I am a C# programmer :-)

Cervator:
Excellent! Looking forward to it :)

PR #2367, opened 2016-06-23
[RFR] Render World Nodes for DAG

tdgunes:
### Contains

Eight new nodes for DAG (Directed Acyclic Graph), wrapping tasks that are related with rendering world.

- _ChunksAlphaRejectNode_
- _ChunksOpaqueNode_
- _ChunksRefractiveReflectiveNode_
- _DirectionalLightsNode_
- _FirstPersonViewNode_
- _LightGeometryNode_
- _ObjectsOpaqueNode_
- _OverlaysNode_

### How to test

Check if there are any weirdness related to rendering of world. (chunks and other in-game entities)

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/ChunksAlphaRejectNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/ChunksAlphaRejectNode.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2016 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.dag;
+
+import org.terasology.config.Config;
+import org.terasology.config.RenderingDebugConfig;
+import org.terasology.engine.ComponentSystemManager;
+import org.terasology.monitoring.PerformanceMonitor;
+import org.terasology.registry.In;
+import org.terasology.rendering.cameras.Camera;
+import org.terasology.rendering.primitives.ChunkMesh;
+import org.terasology.rendering.world.RenderQueuesHelper;
+import org.terasology.rendering.world.WorldRenderer;
+import org.terasology.rendering.world.WorldRendererImpl;
+
+import static org.terasology.rendering.opengl.OpenGLUtils.enableWireframeIf;
+
+
+/**
+ * TODO: Diagram of this node
+ * Alpha reject is used for semi-transparent billboards, which in turn are used for ground plants.
+ */
+public class ChunksAlphaRejectNode implements Node {
+
+    @In
+    private Config config;
+
+    @In
+    private ComponentSystemManager componentSystemManager;
+
+    @In
+    private WorldRenderer worldRenderer;
+
+    @In
+    private RenderQueuesHelper renderQueues;
+
+    private Camera playerCamera;
+    private RenderingDebugConfig renderingDebugConfig;
+
+    @Override
+    public void initialise() {
+        renderingDebugConfig = config.getRendering().getDebug();
+        playerCamera = worldRenderer.getActiveCamera();
+    }
+
+    @Override
+    public void process() {
+        PerformanceMonitor.startActivity("Render Chunks (Alpha Reject)");
+        enableWireframeIf(renderingDebugConfig.isWireframe());
+
+        worldRenderer.renderChunks(renderQueues.chunksAlphaReject,
+                ChunkMesh.RenderPhase.ALPHA_REJECT, playerCamera,
+                WorldRendererImpl.ChunkRenderMode.DEFAULT);
+        PerformanceMonitor.endActivity();
+    }
+}
--- a/engine/src/main/java/org/terasology/rendering/dag/ChunksAlphaRejectNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/ChunksAlphaRejectNode.java
@@ -25,9 +25,8 @@ import org.terasology.rendering.primitives.ChunkMesh;
 import org.terasology.rendering.world.RenderQueuesHelper;
 import org.terasology.rendering.world.WorldRenderer;
 import org.terasology.rendering.world.WorldRendererImpl;
-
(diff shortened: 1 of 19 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/ChunksAlphaRejectNode.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public void initialise() {
+        renderingDebugConfig = config.getRendering().getDebug();
+        playerCamera = worldRenderer.getActiveCamera();
+    }
+
+    @Override
+    public void process() {
+        PerformanceMonitor.startActivity("rendering/chunksalphareject");
```
emanuele3d: Note for later: I just realised that the _actual_ rendering pipeline (the list of commands generated out of the DAG) should also contain PerformanceMonitor statements, so that they do not just wrap the process() method of a node but also the state changes required by that node.

engine/src/main/java/org/terasology/rendering/dag/ChunksOpaqueNode.java
```diff
@@ -0,0 +1,66 @@
...
+    private RenderQueuesHelper renderQueues;
+
+    private RenderingDebugConfig renderingDebugConfig;
+    private Camera playerCamera;
+
+    @Override
+    public void initialise() {
+        renderingDebugConfig = config.getRendering().getDebug();
+        playerCamera = worldRenderer.getActiveCamera();
+    }
```
emanuele3d: Scrap previous comment. Let's leave it as it is for now, some initialize methods are not quite the same.

engine/src/main/java/org/terasology/rendering/dag/ChunksRefractiveReflectiveNode.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public void initialise() {
+        playerCamera = worldRenderer.getActiveCamera();
+    }
+
+    @Override
+    public void process() {
+        PerformanceMonitor.startActivity("rendering/chunksrefractivereflective");
+
+        boolean isHeadUnderWater = worldRenderer.isHeadUnderWater();
```
emanuele3d: Let's make this boolean a private variable please.

engine/src/main/java/org/terasology/rendering/dag/ChunksRefractiveReflectiveNode.java
```diff
@@ -0,0 +1,101 @@
...
+     * Sets the state for the rendering of the reflective/refractive features of the scene.
+     * <p>
+     * At this stage this is the surface of water bodies, reflecting the sky and (if enabled)
+     * the surrounding landscape, and refracting the underwater scenery.
+     * <p>
+     * If the isHeadUnderWater argument is set to True, the state is further modified to
+     * accommodate the rendering of the water surface from an underwater point of view.
+     *
+     * @param isHeadUnderWater Set to True if the point of view is underwater, to render the water surface correctly.
+     */
```
emanuele3d: Once we switch to properties in a map we'll have to pay special attention to how we retain the content of these javadocs.

engine/src/main/java/org/terasology/rendering/dag/DirectionalLightsNode.java
```diff
@@ -0,0 +1,149 @@
...
+     * calculations to the previously flat-lit world rendering stored in the primary FBO.   // TODO: rename sceneOpaque* FBOs to primaryA/B
+     * <p>
+     * See http://en.wikipedia.org/wiki/Deferred_shading as a starting point.
+     */
+    private void applyLightBufferPass() {
+        int texId = 0;
+
+        GL13.glActiveTexture(GL13.GL_TEXTURE0 + texId);
+        sceneOpaque.bindTexture();
+        lightBufferPass.setInt("texSceneOpaque", texId++, true);
```
emanuele3d: These three lines blocks are used quite often, either here or in the ShaderParameters classes, we might want to invent methods that wrap these into more compact code. Not needed yet though.

engine/src/main/java/org/terasology/rendering/dag/Node.java
```diff
@@ -15,7 +15,6 @@
  */
 package org.terasology.rendering.dag;
 
-
 public interface Node {
 
     void initialise();
```
emanuele3d: I just checked and this is... correct! I thought it was wrong and it should have been initialiZe, but it turns out initialise is British so, let's stick to the highest linguistic standard. =)

engine/src/main/java/org/terasology/rendering/opengl/OpenGLUtils.java
```diff
@@ -167,6 +167,20 @@ public static void renderFullscreenQuad() {
...
+    /**
+     * First sets a viewport and then renders a quad filling it.
+     *
+     * @param x              an integer representing the x coordinate (in pixels) of the origin of the viewport.
+     * @param y              an integer representing the y coordinate (in pixels) of the origin of the viewport.
+     * @param viewportWidth  an integer representing the width (in pixels) the viewport.
+     * @param viewportHeight an integer representing the height (in pixels) the viewport.
+     */
+    // TODO: perhaps remove this method and make sure the viewport is set explicitely.
+    public static void renderFullscreenQuad(int x, int y, int viewportWidth, int viewportHeight) {
```
emanuele3d: Hmmm... where did this come from? This method is not actually _necessarily_ rendering a fullscreen quad. It renders a full viewport quad, which might or might not be full screen. In fact, even if it rendered to the whole display, it might still not be full screen as the display might be windowed. Can you find a way to clean this up? You might have to rename this and the other renderFullscreenQuad to a combination of setViewportToSizeOfDisplay() and renderViewportFillingQuad() or just renderQuad().

engine/src/main/java/org/terasology/rendering/dag/Node.java
Cervator: This is one area where we may actually have no standard, I'm pretty sure we have both variants scattered through the code base, depending on who was writing it ;-)

msteiger:
Looks good to me.

Cervator:
Weirdness checked for, none observed! (Radeon on Win7)

Pinging @indianajohn for introducing to rendering changes :-)

engine/src/main/java/org/terasology/rendering/opengl/OpenGLUtils.java
tdgunes: This is from PostProcessor. I will add a TODO on top of this method for the time being, since it might be hard to merge other branches with nodes. But definitely we need to find a suitable name for this method.

GooeyHub:

tdgunes:
@Cervator (thanks for checking it out!) and @msteiger, I think this is ready for merging! Could you please merge this PR, so that we can focus on other nodes?

Cervator:
Done! :-)

Cool to see some of the new rendering categories in the monitoring system slowly sneak in.

PR #2368, opened 2016-06-25
Reduce the log level for opengl notifications.

flo:
### Contains
Reduce the log level for opengl notifications.
 
Reasoning: It used to log a line about buffer usage whenever a buffer got used to draw a line of a UI widget.

The logged message looked like this:
```
    11:21:05.913 [main] INFO  OpenGL - [api] [other] Buffer detailed info:
    Buffer object 212 (bound to GL_ARRAY_BUFFER_ARB,
    usage hint is GL_STATIC_DRAW) will use VIDEO memory as the source for
    buffer object operations.
```

### How to test

Start the game, check the log file for the notifications above. May only occur on certain systems. I use a Geforce on a Linux system with Nvidia drivers.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/subsystem/lwjgl/DebugCallback.java
+++ b/engine/src/main/java/org/terasology/engine/subsystem/lwjgl/DebugCallback.java
@@ -63,7 +63,7 @@ class DebugCallback implements KHRDebugCallback.Handler {
                 break;
             default:
             case GL_DEBUG_SEVERITY_NOTIFICATION:
-                logger.info(logFormat, args);
+                logger.trace(logFormat, args);
                 break;
         }
     }


GooeyHub:

Cervator:
I think `trace` is generally not used with Logback? Something about it usually just linking back to `debug`

I don't recall getting any log messages like that, I only remember the big blurb on startup documenting capabilities and such. So yeah probably card/driver specific. Is there an actual improvement to be made to avoid it in the first place?

This made me think about the "shaders" directory we made a while ago under /logs for some excessive graphics logging, but oddly now that I check it appears to be empty. I wonder what has changed with graphics-related logging lately. Pinging @emanuele3d and @tdgunes :-)

emanuele3d:
Not quite an expert on logging but I don't think we touched any of the logging functionality in recent work and I do not recall having touched it in previous work. 

Also, I do not recall that particular message on my platform (Win7/860M Nvidia).

It's a good reminder for us though. I don't know how much logging we do in the rendering code and it might be good for me to review this aspect.

Cervator:
Merged this and submitted #2387 as a potential follow-up to review later :-)

Issue #1120, opened 2014-05-30
Add button to show CrashReporter from main menu


Cervator:
To ease the reporting of non-crash issues it would be an idea to put a button in the main menu that'll open the Crash Reporter (which perhaps should be called "Issue Reporter", really) under normal working conditions. 

While the bottom of the log file wouldn't necessarily be useful in that case (unless we make it detect the most recent error logging entries or stacktrace?) it would provide super easy access to the game log file, uploading it, filing an issue, etc

Closing #670 and #713 in favor of this
##

Cervator:
Hey @mtjvankuik sending this one your way since you expressed interest on IRC. Sorry for being AFK without being marked as such :-) Feel free to ask questions here, there (or elsewhere) anyway so long as it is somewhere logged we can catch later.

mtjvankuik:
Hey @Cervator, I have reported an issue that may be worthwhile to review in relation to this issue. [https://github.com/MovingBlocks/CrashReporter/issues/29](url)

rzats:
Closing in favour of https://github.com/MovingBlocks/Terasology/issues/2577.

PR #2386, opened 2016-06-29
Move the CrashReporter to the PC facade and make it possible to embed it into a Terasology workspace

Cervator:
Okay I'm officially on vacation/sabbatical woohoo! First stage is *only* till mid-July, but hoping by then to have another couple months lined up :-)

Dug into the idea of having the CrashReporter embedded in a Terasology workspace as a library entry (same approach could be applied to other stuff like Gestalt) and got it working after a bit, sure has been a while but I missed it!

This was in part for @mtjvankuik to make it easier to make changes to the CR then immediately test them via Terasology. Reference: #1120 + https://github.com/MovingBlocks/CrashReporter/issues/29

Pinging @msteiger and @rzats for some review as well, especially for how this might handle in Eclipse rather than IntelliJ. I believe you can already link to the CR in a big Eclipse workspace, but it takes some setup, while this hopefully should make it as simple as running a couple Gradle commands.

### Contains

Most the changes are specific to `.gradle` files, but I renamed the three sub-components in the CrashReporter to be less ambiguous (particularly "core" and "terasology") which means almost everything will show up as changed. So I thought I'd highlight the actual changes here to make it easier. Paths are relative to how they'll appear in a Terasology workspace with the CR embedded, but in reality naturally anything `libs/CrashReporter/...` is in the actual CR repo.

* `engine/build.gradle` - The dependency on CR has finally been moved from here to the PC facade. Apparently a transitive dependency on some Google stuff in the CR led to misdirected imports in a few places (like `Preconditions` being present with two different import paths) - fixed
* `facades/PC/build.gradle` - The CR dependency was moved here and made conditional to check for source vs binary. Can be done with a cleaner Gradle dependency substitution trick later, but that's for another PR
* `libs/subprojects.gradle` - Enhanced to consider the presence of a `settings.gradle` in a library, such as the one that for the CR hooks up *its* three sub-components (which this way works both embedded and standalone)
* `libs/CrashReporter/settings.gradle` - simplified sub-component naming and made the `include` paths conditional on whether the CR is embedded or standalone
* `libs/CrashReporter/build.gradle` - added `idea` plugin and tweaked the setting of `group` to work the same both embedded and standalone
* `libs/CrashReporter/cr-core/build.gradle` - no actual changes other than the different path (was "core" is now "cr-core")
* `libs/CrashReporter/cr-destsol/build.gradle` + `cr-terasology` version - renamed and updated to work at different levels in the Gradle project tree

The imports in various Terasology engine classes have been changed but nothing else. This shouldn't happen again as now engine no longer has access to the transitive dependencies of the CR with its alternative Google stuff (something about guava vs. an HTTP gadget). Sure confused the heck out of me for a little while!

### How to test

* Grab the embeddedCR branch for Terasology
* `gradlew fetchLibCrashReporter`
* `gradlew clean cleanIdea idea`
* Restart IntelliJ (should prompt you)
* Have a way to crash the game, I usually just use the Sample module's crash block (console `giveblock reallyCrashGameBlock` then place and `e` click it)
* Change something in the source version of the CR like the "editBeforeUpload" value in `MessagesBundle.properties` so you can verify you actually are using the embedded CR rather than the binary dependency
* Be sure to run with the CR enabled (non-default utility run config provided for IntelliJ - "Terasology PC (CR enabled)")
* Crash! Confirm that the CR pops up and that your change is present
* Close IntelliJ
* Delete `libs/CrashReporter`
* Run `gradlew clean cleanIdea idea` again
* Retest - you should now again be using the binary dependency

Since this messes with project structure some it may need some extra cleaning. Plain `gradlew idea` isn't necessarily enough.

- [x] Check on any misdirected imports across modules (I'm working in a mini workspace for Gradle without my usual full module setup)
- [x] Increment the CR version to 4.0.0, release it, and make sure its new Artifactory name matches the dependency in Terasology
- [x] Remove the hard link to the CR PR branch in the Terasology `utility.gradle` (just included for easy testing)
- [ ] Also check on Destination Sol - bump the version of the CR it uses (past v1.4) and make sure it has the updated name if changed.
- [x] Double check on import order (I edited some imports in-place and re-added others, so the order differs a bit) and other small stuff, maybe update the license header year etc

### Screenie!

crash reporter_2016-06-28_23-08-23

### DIFF
--- a/config/gradle/utility.gradle
+++ b/config/gradle/utility.gradle
@@ -354,6 +354,11 @@ tasks.addRule("Pattern: fetchLib<ID>") { String taskName ->
                 case 'Index' :
                     githubHome = 'Terasology'
                     break;
+
+                case 'CrashReporter' :
+                    // Temporary for testing only
+                    branch = 'embeddedCR'
+                    break;
             }
 
             // Allow user to override the GitHub account to pull from. Supply with -PgithubAccount="TargetAccountName"
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -127,9 +127,6 @@ dependencies {
     // Wildcard dependency to catch any libs provided with the project (remote repo preferred instead)
     compile fileTree(dir: 'libs', include: '*.jar')
 
-    // TODO: These could be moved into facade
-    compile group: 'org.terasology', name: 'CrashReporter-terasology', version: '3.0.0'
-
     runtime group: 'ch.qos.logback', name: 'logback-classic', version: '1.1.7'
     runtime group: 'org.slf4j', name: 'jul-to-slf4j', version: '1.7.21'
 
--- a/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
+++ b/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
@@ -16,7 +16,7 @@
 
 package org.terasology.engine;
 
-import com.google.api.client.repackaged.com.google.common.base.Preconditions;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.Queues;
 import com.google.common.collect.Sets;
--- a/engine/src/main/java/org/terasology/engine/TerasologyEngineBuilder.java
+++ b/engine/src/main/java/org/terasology/engine/TerasologyEngineBuilder.java
@@ -15,7 +15,7 @@
  */
 package org.terasology.engine;
 
-import com.google.api.client.util.Lists;
+import com.google.common.collect.Lists;
 import com.google.common.base.Preconditions;
 import org.terasology.engine.subsystem.EngineSubsystem;
 import org.terasology.engine.subsystem.common.TimeSubsystem;
--- a/engine/src/main/java/org/terasology/entitySystem/prefab/internal/PojoPrefabManager.java
+++ b/engine/src/main/java/org/terasology/entitySystem/prefab/internal/PojoPrefabManager.java
@@ -15,7 +15,7 @@
  */
 package org.terasology.entitySystem.prefab.internal;
 
-import com.google.api.client.repackaged.com.google.common.base.Strings;
+import com.google.common.base.Strings;
 import org.terasology.assets.management.AssetManager;
 import org.terasology.context.Context;
 import org.terasology.entitySystem.Component;
--- a/engine/src/main/java/org/terasology/logic/console/suggesters/ScreenSuggester.java
+++ b/engine/src/main/java/org/terasology/logic/console/suggesters/ScreenSuggester.java
@@ -15,7 +15,7 @@
  */
 package org.terasology.logic.console.suggesters;
 
-import com.google.api.client.util.Maps;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import org.terasology.assets.ResourceUrn;
 import org.terasology.assets.management.AssetManager;
--- a/engine/src/main/java/org/terasology/rendering/animation/Rect2fAnimator.java
+++ b/engine/src/main/java/org/terasology/rendering/animation/Rect2fAnimator.java
@@ -17,9 +17,9 @@ package org.terasology.rendering.animation;
 
 import java.util.function.Consumer;
 
+import com.google.common.base.Preconditions;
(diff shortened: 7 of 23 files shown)


GooeyHub:

msteiger:
Will have a look shortly - sorry for the delay

rzats:
Can confirm that the entire testing sequence works as intended :smiley:

oh no bad things happened

Some more stuff for the outstanding checklist:

- The input order is definitely off (not alphabetically sorted) in `BlockManagerImpl.java` and `TerasologyEngineBuilder.java`. I think it's fine everywhere else, though.
- `cr-destsol` needs to be removed since it doesn't serve any purpose - ditto for `cr-terasology` in Destination Sol.

Cervator:
Thanks for the review @rzats :-)

> The input order is definitely off (not alphabetically sorted) in BlockManagerImpl.java and TerasologyEngineBuilder.java. I think it's fine everywhere else, though.
> cr-destsol needs to be removed since it doesn't serve any purpose - ditto for cr-terasology in Destination Sol.

Could you expand on these? The first one ... *import* order? Second one removed from where? The Gradle dependency declaration on the CR? In existing code preceding this PR in Terasology that's `CrashReporter-terasology` which I'll update during merging to whatever name it ends up with in Artifactory. Maybe I'm missing something? :-)

rzats:
@Cervator: sorry, that should have been import order. D'oh!

As for the second issue: I've expected `CrashReporter-terasology` to only include the `cr-core` and `cr-terasology` submodules, but it includes `cr-destsol` as well. This needs to be fixed within `CrashReporter`'s Gradle configs - ideally it should be split into two Artifactory repos (one with core&terasology, the other with core&destsol submodules).

Cervator:
@rzats I'm still not entirely sure where exactly you're referring too, can you link to any involved lines? Sorry for me being so dense :-)

If you mean how https://github.com/MovingBlocks/CrashReporter/blob/b3245089719b00fe2dc8c59f386b9b39686a46d8/settings.gradle is referencing all three sub-projects then that's purely for the Gradle project tree, not ultimately something that gets used in a published Terasology or DS flavored jar in Artifactory.

For instance in the https://github.com/MovingBlocks/CrashReporter/blob/b3245089719b00fe2dc8c59f386b9b39686a46d8/cr-terasology/build.gradle the only actual library style dependency is the cr-core - no sign of DS anywhere. Same for the Terasology PC facade Gradle.

I'm probably still missing something!

As for a multi-repo setup well you *could* split out CrashReporter "plugins" or "adapters" or whatever we might call the app specific stubs. But they should be published to Artifactory as individual jars as-is even from a single repo. If we get to a point where the CR suddenly picks up in 3rd party activity then yeah at that point it could make sense to not have *any* app stubs in a central CR repo :-)

rzats:
@Cervator: sorry, that was some confusion on my part - `gradlew fetchLibCrashReporter` downloads the entire repo including the DS adapter so I originally thought it's included as a Gradle dependency, but looks like the actual dependency is on `CrashReporter-terasology` - no problems here :sweat_smile: 

Come to think of it I wouldn't split the game-specific facades (`CrashReporter-terasology`/`CrashReporter-destsol`) into separate repos just yet. Since the configurable variables are prone to change refactoring them will require going through three separate repos to modify the core code, then the respective files.

Cervator:
Yep any split would happen way later :-)

Thanks again for the review!

msteiger:
Looks good to me. I think that we can merge this one.

Cervator:
After just about exactly a month I've finally merged my own PR woo! That should highlight the impartiality of merge durations, hehe :D

@mtjvankuik: You should be able to dev on the CR in an embedded Terasology workspace now in pure source, woo!

@tdgunes @emanuele3d @rzats @Josharias @xrtariq2594: Be advised of minor import changes / sorting in some stuff affecting you (EventualSkills was the only affected module in the stable lineup). May cause conflicts in said imports only, should be pretty minimal.

I only hit files covered by this PR but added minor commits on top not visible here since I sent them directly to `develop` as part of the merge. https://github.com/MovingBlocks/Terasology/commits/develop should show all related commits with one or two including import sorting and such.

After pulling latest from `develop` a fresh `gradlew idea` (or comparable) should be done to pick up the updated lib dependency / adjust project structure (engine no longer depends on the CR)

PR #1591, opened 2015-02-15
Brought alpha-blended mesh rendering in line with opaque mesh rendering.

immortius:
Fix for invisible translucent mesh bug as reported by @skaldarnar 

There are still some rendering issues related the interaction of translucent mesh and depth blur.
zesphere

PR #2390, opened 2016-07-02
Support transparent FBOs

msteiger:
Setting the background to full alpha for FBOs is not necessary. This avoids state changes is also less code. 

Can be tested with [MiniMap](https://github.com/Terasology/Minimap).

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/opengl/LwjglFrameBufferObject.java
+++ b/engine/src/main/java/org/terasology/rendering/opengl/LwjglFrameBufferObject.java
@@ -63,13 +63,6 @@ public class LwjglFrameBufferObject implements FrameBufferObject {
             throw new IllegalStateException("Something went wrong with framebuffer! " + result);
         }
 
-        // clear and fill with full alpha
-        FloatBuffer buffer = BufferUtils.createFloatBuffer(16);
-        GL11.glGetFloat(GL11.GL_COLOR_CLEAR_VALUE, buffer);
-        GL11.glClearColor(0f, 0f, 0f, 1f);
-        GL11.glClear(GL11.GL_COLOR_BUFFER_BIT);
-        GL11.glClearColor(buffer.get(), buffer.get(), buffer.get(), buffer.get());  // reset
-
         GL30.glBindFramebuffer(GL30.GL_FRAMEBUFFER, 0);
     }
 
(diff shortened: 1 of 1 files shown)


GooeyHub:

msteiger:
I'm pretty sure that this didn't work half a year ago. I wonder what changed so that grass is no longer semi-transparent in minimaps.

Cervator:
I have no idea but pinging @emanuele3d and @tdgunes just in case they'd have any clues :-)

@msteiger can you find any old related issues/forum threads? Anything that might remind us of past details? If actually needed/useful, unsure if.

Main thing that came to mind for me was http://forum.terasology.org/threads/bug-with-the-transparent-mesh.1058/ which in turn led to the closed PR #1591 but that could be entirely unrelated.

engine/src/main/java/org/terasology/rendering/opengl/LwjglFrameBufferObject.java
```diff
@@ -63,13 +63,6 @@ public LwjglFrameBufferObject(ResourceUrn urn, BaseVector2i size) {
...
         }
 
-        // clear and fill with full alpha
-        FloatBuffer buffer = BufferUtils.createFloatBuffer(16);
-        GL11.glGetFloat(GL11.GL_COLOR_CLEAR_VALUE, buffer);
-        GL11.glClearColor(0f, 0f, 0f, 1f);
-        GL11.glClear(GL11.GL_COLOR_BUFFER_BIT);
-        GL11.glClearColor(buffer.get(), buffer.get(), buffer.get(), buffer.get());  // reset
-
         GL30.glBindFramebuffer(GL30.GL_FRAMEBUFFER, 0);
```
emanuele3d: We shouldn't be using OpenGL 3.0 directives if we only check for 2.1 support on initialization. What is achieved with this line (binding the display) can be achieved with:

glBindFramebufferEXT(GL_FRAMEBUFFER_EXT, 0);

which is more in line with the capability checks we do on initialization.

Having said that, of course I can see that's not code you introduced @msteiger.

emanuele3d: P.S. I'm just noticing there are other lines using OpenGL 3.0. I guess to convert everything to 2.1-compatible code is beyond the scope of this PR.

emanuele3d:
Sorry, as far as I know the modified class is used only in the context of the UI and I know nothing of that side of things! =(

So, I trust msteiger that he knows what he's doing... =)

Cervator:
Went ahead and did a TeraMath release for v1.3.0, updated engine to use that, then merged latest Minimap which uses it, and finally tested this with all that in place and everything seemed well :-)

Editing title here real quick to have it also count as bumping the TeraMath dependency.

PR #2394, opened 2016-07-03
[NFM] StateManager for DAG 

tdgunes:
### Contains

Add StateManager, StateManagerTest, State, AbstractNode, Node, RenderingGraph and so on.

#### Structure:

In _dag-statemanager_ branch, structure of _dag_ package is:

* dag
	* nodes**
		* AmbientOcclusionPassesNode
		* ...
		* BackdropNode*
		* LightGeometryNode*
		* SkyBandsNode*
		* ChunksRefractiveReflectiveNode*
		* ChunksAlphaRejectNode*
		* ChunksOpaqueNode*
		* FirstPersonViewNode*
		* ObjectsOpaqueNode*
		* OverlaysNode*
		* ...
		* WorldReflectionNode
	* states
		* AbstractState
		* State
		* StateType
		* StateTypeImpl
		* StateValue
		* Wireframe
	* AbstractNode
	* Node
	* NotAnalyzedException
	* RenderingGraph
	* StateManager

\*: _node desires Wireframe state by_ `addDesiredState(StateTypeImpl.WIREFRAME, StateValue.ENABLED)`.
\**: _important to note that these nodes are very outdated when it's compared to ones that are already merged and accepted to `develop` branch._

#### Definitions:

**RenderingGraph:** Stores all nodes and a _StateManager_ instance, generates (will be) topological order of the graph. RenderingGraph can not be "processed", if it is not analyzed by the _StateManager_ instance. If `process` is invoked before invocation of `analyze()`, RenderingGraph throws a custom *NotAnalyzedException*.

**StateManager:** Gets topologically sorted rendering pipeline(not sorted now, since a basic insert-ordered list is used), prepares state changes by `findStateChanges(Collection<Node> nodes)` and executes these changes by `prepareFor(Node node)` method.

For demonstrating the need of **State**, **StateType**, **StateTypeImpl** and **StateValue** please check **StateManagerTest** class.

A snippet for quick reviewing:

```java
@Test
public void testOneStateSixNodes2() {
        StateManager stateManager = new StateManager();
        DummyState dummyStateA = new DummyState();
        stateManager.addState(DummyStateType.A, dummyStateA);
        DummyNode node1 = new DummyNode("node1");
        DummyNode node2 = new DummyNode("node2");
        DummyNode node3 = instantiateDummyNodeWithState("node3", DummyStateType.A, StateValue.ENABLED);
        DummyNode node4 = instantiateDummyNodeWithState("node4", DummyStateType.A, StateValue.ENABLED);
        DummyNode node5 = new DummyNode("node5");
        DummyNode node6 = new DummyNode("node6");

        Collection<Node> dummyPipeline = createDummyPipeline(node1, node2, node3, node4, node5, node6);
        stateManager.findStateChanges(dummyPipeline);

        // run pipeline
        for (Node node : dummyPipeline) {
            stateManager.prepareFor(node); 
            node.process() // doesn't have any effect on the test
        }

        assertEquals("10", dummyStateA.getHistory());
    }

```
 

#### _(Unofficial)_ Abbreviations:

**NFM**: _Not for Merge_
**WIP**: _Work in Progress_
**RFR**: _Ready for Review_
**RFM**: _Ready for Merge_

_Would be great to have @cervator's opinion about these, plus making them official perhaps?_

### How to test

* Run _StateManagerTest_ for testing _StateManager_.
* Play the game with different video settings and observe any irregularities.  

- [ ] Naming, patterns and architecture?, from @manu3d:

> 
* [ ] a **RenderPipelineGenerator** that takes the current render graph in input and generates the render pipeline 
* [ ] once generated the render pipeline should be printed out in the log, so we can inspect it
a **RenderPipelineOptimizer** that takes an existing pipeline and optimizes it by removing redundant state changes
* [ ] once again, we should get a print out to inspect the outcome of the optimization process 
* [ ] the rendering occurring by processing the rendering pipeline and generating images that are identical to pre-dag changes.


GooeyHub:

engine-tests/src/test/java/org/terasology/rendering/dag/StateManagerTest.java
```diff
@@ -0,0 +1,229 @@
...
+import static org.junit.Assert.assertNull;
+
+public class StateManagerTest {
+    @Test
+    public void testOneStateTwoNodes() {
+        StateManager stateManager = new StateManager();
+        DummyState dummyStateA = new DummyState();
+        stateManager.addState(DummyStateType.A, dummyStateA);
+        DummyNode node1 = instantiateDummyNodeWithState("node1", DummyStateType.A, StateValue.ENABLED);
+        DummyNode node2 = instantiateDummyNodeWithState("node2", DummyStateType.A, StateValue.ENABLED);
```
emanuele3d: As a general comment on testing, tests are supposed to be part of the documentation on how to use the tested code. Not quite as far as a tutorial, but representative of production code nevertheless. At least that's the theory.

In this case I'd therefore keep render graph instantiation, node instantiation and wiring between the two in very explicit form rather than wrapping them within private methods such as instantiateDummyNodeWithState and createDummyPipeline. Unless you can create a single "setup(void)" method that does all this for many test cases. On that base you can then customize nodes and graph depending on need, i.e.:

```
commonTestSetup()
nodeA.customizeColor(blue)
nodeB.customizeColor(red)
...etc, etc...
```

I'd also consider what is representative in terms of instantiation timeline. For example, the render graph is among the first classes to be instantiated in this context. I'd therefore instantiate that first. I'd then instantiate and add nodes to it. (I remind you those instantiations might be in a setup method. Finally I'd deal with the focus of the testing, that is the StateManager instantiation and its interactions with the other objects. This would lead to all "focus" code, include its testing asserts, all close to each other and without distractions due to setting up other instances. This way you get a macrostructure of a test such as:

- common setup
- setup customization for specific test
- focus code
- actual testing

Having said that, don't rush to change this specific test class as I'm not sure the class it tests will survive scrutiny in its current form. So, consider what I wrote only in general terms, for future test classes.

engine/src/main/java/org/terasology/rendering/dag/states/StateValue.java
```diff
@@ -0,0 +1,25 @@
...
+package org.terasology.rendering.dag.states;
+
+/**
+ *
+ */
+public enum StateValue {
+    ENABLED, // state change must happen
+    OPTIONAL,
+    DISABLED,
+}
```
emanuele3d: I imagine this is just the start, to deal with binary state changes?

engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
```diff
@@ -0,0 +1,45 @@
...
+
+
+import com.google.api.client.util.Maps;
+import java.util.Map;
+import org.terasology.rendering.dag.states.StateType;
+import org.terasology.rendering.dag.states.StateValue;
+
+
+public abstract class AbstractNode implements Node {
+    protected Map<StateType, StateValue> desiredStates;
```
emanuele3d: Ok, first of all, as we delve into the details of this, it should be noted that OpenGL is a big giant state characterized by many attributes. Pretty much all opengl functions change the (big giant) OpenGL state by changing one or more of its attributes. So, we should stop thinking and communicating in terms of "states" and think in terms of "state changes", because there is really only one state and that changes through time.

Regarding this particular line I initially thought of a <string, object> map for this. But perhaps even that is too complicated. Perhaps what we need is a _set_ (or a list perhaps) of desired StateChange instances. 

Now, let's remind ourselves what a Node might want:

1. A node might require an opengl attribute set to its default. 
1.1. In the node, the desiredStateChanges set has no entry for this.
1.2. Before Node.process() is executed there is nothing to do because state management should always reset attributes to their default unless this node wanted something different
1.3 After Node.process() is executed there is nothing to do unless the next node wants the attribute set to a non-default value

2. A node might require an opengl attribute set to a non-default value.
2.1. In the node the desiredStateChanges set has an entry for this. This entry contains the non default value.
2.2. Before Node.process() is executed the state management system verifies if that particular attribute is already set to that value, if not the state change is added to the pipeline, to be executed before the process() method.
2.3. After Node.process is executed the state management system verifies if that particular attribute+value is needed by the next node. If it is needed, it doesn't do anything. If it isn't it adds a state change to the pipeline to revert the attribute to its default.

3. A node might be capable of handling any value of a particular opengl attribute.
3.1. In the node the desiredStateChanges set has an entry for this. This entry contains a null value.
3.2. Before Node.process is executed there is nothing to do because this node can handle whatever value the attribute is set to.
3.3 After Node.process is executed the attribute needs to be reset to its default unless the next node is also capable of handling any value OR it requires a different value, be it the default or otherwise.

In this context:

The RenderPipelineGenerator could add all the explicitly listed state changes to the pipeline, disregarding of them being redundant or not. For example it could generate:

```
set attribute "color" to (non default) "blue" - explicitely requested
process node A
set attribute "color" to (default) "red" - default behavior but redundant: next node needs "blue"
set attribute "color" to "blue". - explicitely requested but redundant: "blue" is the current value already
process Node B
set attribute "color" to (default) "red"
```

Then, it would be up to the RenderPipelineOptimizer to turn that pipeline into:

```
set attribute "color" to (non default) "blue"
process Node A
process Node B
set attribute "color" to (default) "red"
```

And this is the pipeline that would actually get executed. 

Notice I haven't said anything about what the pipelines actually look like. I imagine them as lists, but the first one, generated by the RenderPipelineGenerator could list StateChanges and Nodes directly as it is geared for analyses. The second one, generated by the RenderPipelineOptimizer might contain PipelineTask instances, geared instead for fast execution. I.e. a class implementing the interface PipelineTask might look like this:

```
class SetColorTask implements PipelineTask {
 
    private final Color color;

    // called by the RenderPipelineOptmizer, before it adds the task to the pipeline to execute.
    SetColorTask(StateChange colorStateChange) {
        color = colorStateChange.getValue();
    }

   // called by whatever is iterating the pipeline being executed
   public void execute() {
       GL99.setColorFunction(color);
   }
}
```

Sorry I have been very verbose, but I strongly believe this is what we should be aiming for.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
```diff
@@ -0,0 +1,140 @@
...
+import org.terasology.rendering.dag.nodes.ToneMappedSceneNode;
+import org.terasology.rendering.dag.nodes.WorldReflectionNode;
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * TODO: Add javadocs
+ */
+public class RenderingGraph {
```
emanuele3d: Let's call it RenderGraph and RenderPipeline. Doesn't change the meaning and saves some characters.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
```diff
@@ -0,0 +1,140 @@
...
+    private final Context context;
+    private Map<String, Node> nodeMap;
+    private StateManager stateManager;
+    private Collection<Node> topologicalOrder;
+    private Config config;
+    private boolean analyzed;
```
emanuele3d: In line with the architecture I suggested earlier I do not see this as necessary: the render graph is used to generate an unoptimized render pipeline and this in turn is used to generate the pipeline to be executed.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.Map;
+
+/**
+ * TODO: Add javadocs
+ */
+public class RenderingGraph {
+    private final Context context;
+    private Map<String, Node> nodeMap;
+    private StateManager stateManager;
+    private Collection<Node> topologicalOrder;
```
emanuele3d: As we are still dealing with a simple list I'd remove any functionality regarding ordering in future implementations, where it would actually be useful. The less code there is to review, the faster the PR goes through.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
```diff
@@ -0,0 +1,140 @@
...
+        pipeline.add("lightGeometry", LightGeometryNode.class);
+        pipeline.add("directionalLights", DirectionalLightsNode.class);
+        pipeline.add("chunksRefractiveReflective", ChunksRefractiveReflectiveNode.class);
+        pipeline.add("outline", OutlineNode.class);
+        pipeline.add("ambientOcclusionPasses", AmbientOcclusionPassesNode.class);
+        pipeline.add("prePostComposite", PrePostCompositeNode.class);
+        pipeline.add("simpleBlendMaterials", SimpleBlendMaterialsNode.class);
+        pipeline.add("lightShafts", LightShaftsNode.class);
+        pipeline.add("initialPostprocessing", InitialPostProcessingNode.class);
+        pipeline.add("downSampleSceneAndUpdateExposure", DownSampleSceneAndUpdateExposureNode.class); // TODO: a rename needed :)
```
emanuele3d: In line with comments in the other PR currently open, I'd recommend "UpdateExposureNode"). The fact that this is done by first downsampling an existing rendering of the scene is not something we need to worry about at this level.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
```diff
@@ -0,0 +1,140 @@
...
+        pipeline.add("toneMapping", ToneMappedSceneNode.class);
+        pipeline.add("bloomPasses", BloomPassesNode.class);
+        pipeline.add("blurPasses", BlurPassesNode.class);
+        pipeline.add("finalPostProcessing", FinalPostProcessingNode.class);
+        pipeline.analyze();
+        return pipeline;
+    }
+
+    public void analyze() {
```
emanuele3d: Conceptually, this should go in the RenderPipelineOptimizer, that's its role and the implementation will obviously different.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
```diff
@@ -0,0 +1,140 @@
...
+        T node = InjectionHelper.createWithConstructorInjection(type, context);
+        // Then fill @In fields
+        InjectionHelper.inject(node, context);
+        node.initialise();
+        this.add(type.cast(node));
+    }
+
+    public Node get(String identifier) {
+        return nodeMap.get(identifier);
+    }
```
emanuele3d: Note for myself: what do we need identifiers for, at this stage? 

This comment will be removed if I find a good reason. That been said, I know they will be needed at some point.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
```diff
@@ -0,0 +1,140 @@
...
+        topologicalOrder = getTopologicalOrder();
+        stateManager.findStateChanges(topologicalOrder);
+        analyzed = true;
+    }
+
+    public void add(AbstractNode node) {
+        nodeMap.put(node.getIdentifier(), node);
+    }
+
+    public <T extends AbstractNode> void add(String identifier, Class<T> type) {
```
emanuele3d: I'm not convinced identifiers should be set on adding a node to the graph. Perhaps the node itself might come with its own identifier. 

By the way, identifiers open a considerable can of worms: what happens if a module wants to add a node with an id identical to another node? There are a number of strategies to deal with this but we'll have to deal with the problem.

emanuele3d: Scrap previous comment. I'm quite convinced that Nodes shouldn't be instantiated by the RenderGraph. Let's keep important classes nice and tight, doing only a few necessary things. Instantiation of nodes should be somebody's else problem.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
```diff
@@ -0,0 +1,140 @@
...
+        InjectionHelper.inject(node, context);
+        node.initialise();
+        this.add(type.cast(node));
+    }
+
+    public Node get(String identifier) {
+        return nodeMap.get(identifier);
+    }
+
+    public void process() throws NotAnalyzedException {
```
emanuele3d: In line with earlier considerations the RenderGraph should not have a process() method as it should get processed by the RenderPipelineGenerator instead.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
```diff
@@ -0,0 +1,140 @@
...
+        for (Node node : this.topologicalOrder) {
+            PerformanceMonitor.startActivity("rendering/" + node.getIdentifier());
+            stateManager.prepareFor(node);
+            node.process();
+            PerformanceMonitor.endActivity();
+        }
+    }
+
+    // TODO: getTopologicalOrder will be a complex task, therefore calling  it multiple times will be expensive
+    private Collection<Node> getTopologicalOrder() {
```
emanuele3d: 1. I think you can narrow its output type to a _list_ of Nodes rather than a generic collection.
2. I'm tempted of calling it "getNodes()", but I'd be happy with getNodesInTopologicalOrder().

engine/src/main/java/org/terasology/rendering/dag/StateManager.java
```diff
@@ -0,0 +1,121 @@
...
+import org.terasology.rendering.dag.states.State;
+import org.terasology.rendering.dag.states.StateType;
+import org.terasology.rendering.dag.states.StateTypeImpl;
+import org.terasology.rendering.dag.states.StateValue;
+import org.terasology.rendering.dag.states.Wireframe;
+
+/**
+ * TODO: add javadocs
+ */
+public class StateManager {
```
emanuele3d: In line with previous comments this architecture doesn't work for me, sorry. We need a RenderGraph, a RenderPipelineGenerator and a RenderPipelineOptimizer. To start with it would be good to focus only on the first two. We can think about the last one once we have nailed the first two.

engine/src/main/java/org/terasology/rendering/dag/StateManager.java
```diff
@@ -0,0 +1,121 @@
...
+        stateManager.addAvailableStates();
+        return stateManager;
+    }
+
+
+    public void addState(StateType stateType, State state) {
+        availableStates.put(stateType, state);
+    }
+
+    public void findStateChanges(Collection<Node> nodes) {
```
emanuele3d: The term "find" is a bit limited compared to the goal of this method. This method intends to prevent redundant state changes from occurring and should be renamed accordingly, i.e. "eliminateRedundantStateChanges".

That been said, I doubt it will take this form in the RenderPipelineOptimizer as an instance of that class will work on a list of StateChanges and Nodes, working sequentially and eliminating state changes while it iterates through them. In this method you are working with chunks of state changes (desired state changes vs planned state changes for a given node), making the method unnecessarily complex. 

We might re-introduce some complexity eventually, when we'll have to track changing node parameters and their impact on the pipeline. But at this stage I'd keep it simple.

engine/src/main/java/org/terasology/rendering/dag/states/State.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.dag.states;
+
+/**
+ *
+ */
+public interface State {
```
emanuele3d: In line with previous comment, this should be renamed to "StateChange".

engine/src/main/java/org/terasology/rendering/dag/states/State.java
```diff
@@ -0,0 +1,25 @@
...
+    void set(StateValue stateValue);
```
emanuele3d: State changes don't need to be mutable. Let's set the value only once, in the constructor instead.

engine/src/main/java/org/terasology/rendering/dag/states/AbstractState.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ *
+ */
+public class AbstractState implements State {
+    private Context context;
+    private StateValue value;
+
+    public AbstractState(Context context) {
+        this.context = context;
```
emanuele3d: What do we need the context for? Also, read below regarding the constructor.

engine/src/main/java/org/terasology/rendering/dag/states/State.java
```diff
@@ -0,0 +1,25 @@
...
+
+    StateValue getValue();
```
emanuele3d: Due to previous comments and architectural changes I think the type returned by getValue() should be dependent on the actual opengl state attribute being affected. So, a state change inheriting from AbstractStageChange could return a float, another an integer, another something else. I'm not sure how to implement this the best. Perhaps the method could return a generic object to be cast appropriately?

engine/src/main/java/org/terasology/rendering/dag/states/StateType.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.dag.states;
+
+
+public interface StateType {
```
emanuele3d: No longer necessary: classes inheriting from StateChange will define their own type.

engine/src/main/java/org/terasology/rendering/dag/states/StateValue.java
emanuele3d: Scrap previous comment. In line with earlier comments this is no longer relevant as I wouldn't encapsulate values in their own instance but directly in an instance of a class implementing StateChange.

engine/src/main/java/org/terasology/rendering/dag/states/Wireframe.java
```diff
@@ -0,0 +1,60 @@
...
+    public void set(StateValue stateValue) {
+        super.set(stateValue);
+
+        if (!renderingDebugConfig.isWireframe()) {
+            return;
+        }
+
+        switch (stateValue) {
+            case ENABLED:
+                GL11.glPolygonMode(GL_FRONT_AND_BACK, GL_LINE);
```
emanuele3d: In line with earlier comments regarding the whole architecture this line would be the only line contained in the execute() method of an instance implementing the RenderPipelineTask interface. The pipeline executed for rendering shouldn't go through branching execution path. It should be entirely linear. At least in theory. Perhaps we'll find cases where it's necessary to have IFs or switches. But I'd rather avoid them as much as possible.

emanuele3d:
Just a quick comment to say that my review of the currently available commits is finished.

I am sorry there is a lot of work to undo. I think the key to minimize this in the future is to present to us interfaces, abstract classes and semi-empty classes before you go too deep into implementations and test cases. In fact I'd love to see only new interfaces, abstract classes and semi-empty classes in the next commit, even if the commit is not functional - this PR is not for merging. No tests, no deep implementations, perhaps a few examples of how the new classes get used.

Similarly to what happened with nodes though, I believe it's normal we have some teething pains as we go through the first major piece of the new architecture. It's an important piece, let's get it right.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
tdgunes: Identifiers were just used at this branch in `initPipeline()` method of `WorldRendererImpl.java`:

```java
    private void initPipeline() {
        // FIXME: init pipeline without specifying them as a field in this class
        renderingGraph = RenderingGraph.createDefault(context);

        shadowMapNode = (ShadowMapNode) renderingGraph.get("shadowMap"); // TODO: remove shadow map node from world renderer
    }
```

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
tdgunes: `NodeFactory` class for instantiating nodes, maybe?

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
tdgunes: 1. I decided to make it generic collection to avoid a `(List)` since  .getValues() of a LinkedHashSet returns a Collection. I wouldn't mind casting it to a `List` to emphasise that it's an ordered _Collection_.  
2. Adding topological to the method name is more meaningful when compared with solely _getNodes()_. In order to find middle ground in naming, let's say `getNodesInTopologicalOrder()`.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
tdgunes: `process()` might be not the great name for the operation that its doing. What this method actually does is iterating _analyzed_ topologically ordered node list and before calling node.process(), stateManager prepares environment according to node's desires. `processAll()` might be perhaps a better name.

RenderPipelineGenerator, as from the name I assume, generates a render pipeline by topologically sorting the instance RenderGraph. I will edit this comment once, I find out in the later comments that I might be misunderstanding something.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
tdgunes: RenderGraph or RenderPipeline? 

Separating RenderingGraph into two different classes?

tdgunes:
@emanuele3d Thank you for your review, I really took me a while to digest all of comments. Even still I am a little confused and skeptical about some of them. 

My plan while approaching a problem is to select a simple particular issue and try to solve it and make it work. Currently, this branch eliminates redundant state changes of enabling/disabling Wireframe. As for testing, I also included several test cases to verify correctness by validating history of state variable changes. (only binary at this point by StateValue enumeration .ENABLED, .DISABLED) I selected Wireframe because we used as an example many times during our meetings.  

It seems like it's best to first construct the architecture and later move on important aspects. My prior perspective is that architecture is last thing that is important for me, because codebase can be refactored later on even if code is a mess, but it will stay working. Therefore, I am ready to undo and construct the architecture, as you mentioned.

engine-tests/src/test/java/org/terasology/rendering/dag/StateManagerTest.java
tdgunes: Noted. Thanks for your advices, especially on the macrostructure.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
emanuele3d: The current RenderingGraph is mostly what I'd call RenderGraph, to save some typing. Some of its code should go in RenderPipelineGenerator. RenderPipeline is the output of the generator, but at this stage it can just be a List<Object>, as it should include StateChange and Node instances.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
emanuele3d: Let's do it this way: let's keep this functionality, but let's think a little bit more in-depth about what object should store the id. 

Initially I thought all the information related to a node should be stored into a node. But if the purpose of an id is to retrieve a node from the render graph than perhaps the id could just be stored in the map held by the render graph - the node itself would have no id.

If we go along this line I propose the RenderGraph class manipulates nodes -mostly- only by id. Notable exception would be:

`public string add(Node node, string suggestedId)`

This method would return a string because it is possible that the suggested id might be already in use. In this case the returned string would be a slightly modified version of the suggested id. I.e. if a "blur" node already exist and a module wishes to register another "blur" node, the actual id returned could be "blur2".

One potentially trivial issue I'm wondering about is how to name the accessor methods. Should we use methods such as 

add(), get(), remove() or addNode(), getNode(), removeNode()? At this stage, with the render graph only handling nodes, this is perhaps an unnecessary distinction. But as we add edge-related functionality perhaps addNode/getNode/removeNode() would make more sense?

Thoughts?

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
emanuele3d: Sure! Would that wrap all the injection stuff? How would the code to instantiate a node look like? Something like:

`Node myNode = NodeFactory.createInstance(MyNode.class)`

perhaps?

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
emanuele3d: No, what I'm suggesting is that the RenderGraph class only returns a list of nodes through the getNodesInTopologicalOrder() method. The RenderPipelineGenerator takes that list and generates a new list that also includes StateChange instances interleaved with Nodes. The RenderPipelineOptimizer takes that new list and generates the list of commands to execute during rendering.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
emanuele3d: 1. Yes, I'd go for an explicit list because the output of this method must be considered ordered.
2. Agreed.

emanuele3d:
First of all, I'm sorry about the confusion and no doubt the frustration. 

I appreciate that you tried to solve the specific problem of wireframe state change and have working code addressing that particular issue. I also understand your focus on working code rather than architecture. Regarding the things you are skeptical about, I look forward hearing about them.

Personally, I do not see the necessity to choose between architecture and working code. Personally I try to figure an architecture that would work best for the intended users (in this case developers and mod developers) and then I work out the details. Of course one could write horrible code that works and then eventually refactor it. However, first of all other people have to review it in its horrible state. Then there is the remote but non-zero chance that you might have to leave the project prematurely and without the chance to do the refactoring, leaving a mess that in an open source context might lie untouched for months or years as nobody dares coming near it. (With this paragraph I do not mean that your code was horrible - not at all).

So, no, if the aim is to merge it in develop the architecture should be reasonably well defined AND the code must be working. That been said, let's not focus (too much) on any given philosophical approach: we are human and we can cope with some degree of ambiguity and contradictions.

Specifically on your code, I didn't perceive your approach as scalable enough nor simple and elegant enough from the perspective of the intended users. I have no doubt your approach might work, but I think we can do better by more clearly defining responsibilities in separate classes and using simpler objects encapsulating a state change.

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
tdgunes: Yeah, but context must be included as an argument for injection too. Like:

```ruby
Node myNode = NodeFactory.createInstance(MyNode.class, context)
```

engine/src/main/java/org/terasology/rendering/dag/states/AbstractState.java
tdgunes: Context was included for Wireframe task's RenderingDebugConfig dependency.

tdgunes:
Thanks for your very informative reply. My skepticism was just focusing perhaps "too much" on a philosophical approach while tackling on this issue. I might have tempted to move in this way, probably because of my prior commercial dev. experience where you need to deliver as fast as you can (make it a mess) and do the cleaning whenever a new delivery needed.  

Nevertheless, I started applying your comments and clearly defining architecture and responsibilities in separate classes with no actual implementation (+ short and sweet). I will be sharing that in another PR if that is ok and leave this PR as closed. (Please let us not remove this PR of course, an important knowledge source for the next PR.)

engine/src/main/java/org/terasology/rendering/dag/RenderingGraph.java
tdgunes: `add(Node node)`, `get()`, `remove(Node node)` seems much better. For dependencies between nodes, may be just `addEdge(Node node1, Node node2)` and `removeEdge(Node1 node1, Node1 node1)`?

Cervator:
I'm not even sure we *can* delete a PR :D Closing it is the only option I know of, so that's fine :-)

Feel free to close this any time if a new one will replace it. Closing it won't make it go anywhere, it'll be right here for reference pretty much forever :watch:

PR #2398, opened 2016-07-05
Fix to infinite multijumping if there are multiple characters

xrtariq2594:
### Contains

Fixes issue where if there are multiple entities with KinematicCharacterMovers in the world, a character could potentially jump infinitely.

### How to test

Spawn into a world. Spawn Gooey from the Gooey's Quests modules. Attempt to multi-jump. If it fails, the test is successful.

### Credits

Thanks to @Cervator for catching this issue.


GooeyHub:

xrtariq2594:
On a similar note, should the following two variables be moved onto character-specific components. After all, they aren't necessarily shared between character entities.

```
    private boolean stepped;
    // Processing state variables
    private float steppedUpDist;
```

Cervator:
Confirmed fix and merged :-)

Those other two variables - quite possibly. May just have been hard for that to cause issues so far. Probably other stuff like it hiding too.

PR #2403, opened 2016-07-10
Initial inline editor implementation

rzats:
### Contains

An implementation for rzats/gsoc-nui-editor#6.
- An option is added to replace a `UITreeView`'s node with an arbitrary widget that will be rendered in place of the node.
- It is used within `NUIEditorScreen` to replace a selected node with an inline `UITextEntry` editor using custom `Formatter`/`Parser` overrides to convert a `JsonTreeNode` to a JSON string and vice versa.

editor-before
editor-after

### How to test

Right-click on a NUI editor node, select "Edit", make some changes!

- [x] Currently testing for different validation quirks (other than the basic `JsonSyntaxException` that can be thrown by `JsonParser`) and implementing validation accordingly.
- [x] Should handle `OBJECT`/`ARRAY` nodes (with slightly different `Parser`/`Formatter` logic).
- [x] Editor should also open on doubleclick/F2.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorScreen.java
@@ -17,9 +17,12 @@ package org.terasology.rendering.nui.editor;
 
 import com.google.common.base.Charsets;
 import com.google.common.collect.Lists;
+import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import com.google.gson.JsonPrimitive;
 import com.google.gson.JsonSyntaxException;
 import com.google.gson.stream.JsonReader;
 import org.codehaus.plexus.util.ExceptionUtils;
@@ -38,11 +41,13 @@ import org.terasology.rendering.nui.WidgetUtil;
 import org.terasology.rendering.nui.asset.UIElement;
 import org.terasology.rendering.nui.asset.UIFormat;
 import org.terasology.rendering.nui.contextMenu.ContextMenuBuilder;
+import org.terasology.rendering.nui.contextMenu.ContextMenuScreen;
 import org.terasology.rendering.nui.databinding.Binding;
 import org.terasology.rendering.nui.events.NUIKeyEvent;
 import org.terasology.rendering.nui.widgets.UIBox;
 import org.terasology.rendering.nui.widgets.UIDropdownScrollable;
 import org.terasology.rendering.nui.widgets.UILabel;
+import org.terasology.rendering.nui.widgets.UITextEntry;
 import org.terasology.rendering.nui.widgets.UITreeView;
 import org.terasology.rendering.nui.widgets.models.JsonTree;
 import org.terasology.rendering.nui.widgets.models.JsonTreeConverter;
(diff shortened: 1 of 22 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorScreen.java
```diff
@@ -124,8 +131,11 @@ public void initialise() {
                 .stream()
                 .collect(Collectors.toList()));
 
-        // Exclude the NUI editor itself, then sort the list.
-        availableAssetList.removeIf(asset -> asset.getRootUrn().equals(ASSET_URI));
+        // Exclude screens used by the NUI editor, then sort the list.
+        availableAssetList.removeIf(asset -> asset.getRootUrn().equals(this.ASSET_URI));
+        availableAssetList.removeIf(asset -> asset.getRootUrn().equals(ContextMenuScreen.ASSET_URI));
+        availableAssetList.removeIf(asset -> asset.getRootUrn().equals(NUIEditorSettingsScreen.ASSET_URI));
+        availableAssetList.removeIf(asset -> asset.getRootUrn().equals(WidgetSelectionScreen.ASSET_URI));
```
flo: Is there a reason for excluding them?

rzats: Attempting to edit `ContextMenuScreen` and then opening a context menu causes a stack overflow - the other screens are excluded just in case they break in a similar way.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorScreen.java
```diff
@@ -281,22 +295,89 @@ private void paste(JsonTree item) {
...
+                    }
+                }
+                return new JsonTree(node);
+            } catch (JsonSyntaxException e) {
+                return null;
+            }
+        });
+
+        editorTreeView.clearAlternativeWidgets();
+        editorTreeView.addAlternativeWidget(editorTreeView.getModel().indexOf(item), inlineEditorEntry);
```
flo: I can't think of a use casd where we want multiple inline editors active at the same time.
So maybe a `showEditorWidgetForActiveItem(EditorWidget)` would be enough.
Or TreeView could offer a registerEditorForNodeType method.(Since usually double click and F2 activa the renaming/editing of an item and would thrn not to be implemented for each treeview useage again

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorScreen.java
flo: Hmm isn't the editorvworking with copies of the screen?

flo:
Very cool that you vot it working so nicely I even could use cooy and paste within the widget.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorScreen.java
rzats: The copies still get registered within `NUIManager` as the instance of a specific screen class, which means that `ContextMenuBuilder` instead of pushing a new context menu at a specific location [attempts to fetch](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/contextMenu/ContextMenuBuilder.java#L72) the existing (deliberately uninitialized) `ContextMenuScreen` and causes a NPE. This isn't normally a problem unless meta-editing like this is attempted :smile:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

flo:
good work, thanks for the pull request :)

GooeyHub:

PR #2226, opened 2016-03-12
Fix for the drop-down translation system.

Margretor:
Contains:

Fixes #2207

Testing:
Just make -> run -> change language -> observe drop-down video settings.

- [ ] Needs refactoring as  [skaldarnar](https://github.com/skaldarnar) suggested via IRC but there are some issues such as Blur.

PR #2404, opened 2016-07-11
Made the graphics presets respect translation keys (requires user reset of config.cfg)

### LINKED ISSUE #2207: Translation keys for graphic presets seem to not be in use

See the graphic presets like "Uber" and "Insane" get translated into other languages - we have the values in the language files.

### What actually happened

I always see English values in the graphics present drop-down, regardless of what language is selected.

Somewhat related to #2028 as we probably should clarify the actual names then re-translate where needed.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31633469-translation-keys-for-graphic-presets-seem-to-not-be-in-use?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

indianajohn:
### Contains

Fixes #2207. The presets menu now is translated according to the keys in menu_XX.lang.

I looked at this issue, and it appeared that the asignee had gotten busy and forgotten about it. I hope I'm not stepping on any toes.

### How to test

Before:
Go to Settings->Player and switch your language to "Deutsch (German)". Note that the first two entries remain "Custom" and "Minimal."

After:
Do the same thing. Note that the first two entries change to "Benutzerdefiniert" und "Einfach."

- [ ] Translations appear to be somewhat incomplete. That probably shouldn't stay in the way of this getting merged, though.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/itemRendering/ToStringTextRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/itemRendering/ToStringTextRenderer.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 package org.terasology.rendering.nui.itemRendering;
-
+import org.terasology.i18n.TranslationSystem;
 import java.util.Objects;
 
 /**
@@ -23,8 +23,22 @@ import java.util.Objects;
  */
 public class ToStringTextRenderer<T> extends StringTextRenderer<T> {
 
+    private TranslationSystem translationSystem;
+
+    public ToStringTextRenderer() { }
+
+    public ToStringTextRenderer(TranslationSystem translationSystemInput) {
+        translationSystem = translationSystemInput;
+    }
+
+
     @Override
     public String getString(T value) {
-        return Objects.toString(value);
+        if (translationSystem == null) {
+            return Objects.toString(value);
+        }
+        else {
+            return translationSystem.translate(Objects.toString(value));
+        }
     }
 }
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/CameraSetting.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/CameraSetting.java
@@ -18,9 +18,9 @@ package org.terasology.rendering.nui.layers.mainMenu.videoSettings;
 /**
  */
 public enum CameraSetting {
-    NORMAL("Normal", 1),
-    SMOOTH("Smooth", 5),
-    CINEMATIC("Cinematic", 60);
+    NORMAL("${engine:menu#camera-setting-normal}", 1),
+    SMOOTH("${engine:menu#camera-setting-smooth}", 5),
+    CINEMATIC("${engine:menu#camera-setting-cinematic}", 60);
 
     private String displayName;
     private int smoothingFrames;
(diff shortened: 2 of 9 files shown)


indianajohn:
Realized that https://github.com/MovingBlocks/Terasology/pull/2226/files contains this and more; feel free to close this if you'd like. I'm just poking around the code base and getting familiar with things.

GooeyHub:

Cervator:
I just commented over in the older PR, figuring we could merge this as it is "fresh" and pretty minimal, then with @Margretor back he could take what remains useful from his PR and submit it separately when ready?

No real stepping on toes, just a funny coincidence that he popped up right as this was submitted :-)

@Margretor feel free to grab the latest code, test this out, and see if any of your stuff should be added in immediately, or if we should go ahead and merge then you can look at adding anything remaining?

msteiger:
Thanks for the PR, @indianajohn. I would prefer to have a more general solution that works with other drop down lists as well. Also, it should support changing the language during the game, if possible.

One possibility would be to provide a label text function to the UIDropdown control that maps a Preset to a translated string. This could then be done by the translation system. What do you think?

GooeyHub:

indianajohn:
When starting on a new code base it's sometimes a bit hard to see the forest for the trees. :) I looked at the problem again and got what seems to be a fairly general solution.

In my revision to this branch, I grouped the translation tasks with the `ToStringTextRenderer`. To avoid having that affect other parts of the game that contain `ToStringTextRenderer` instances, A new constructor added to the class accepts a `TranslationSystem` instance, and another that does not is also added. The first causes the `ToStringTextRenderer` to run everything through the input `TranslationSystem` instance, and the second just passes the string through without any change. In the case that a `TranslationSystem` instance is provided and no translation key is given by the object's `toString()` method, it just renders the string.

I also adjusted every instance in the `VideoSettingsScreen` class to utilize the change and it seems to work fine. More importantly, it gives an easy one-line change for any further usages of `ToStringTextRenderer` to begin using `TranslationSystem`.

Let me know if there is anything you would like me to change, and, as always, thanks for the feedback.

Cervator:
Haven't forgotten about this one! Been focusing a little on poor neglected Destination Sol and made some solid progress there tonight :-)

@msteiger: ping! In case you get a chance to provide some feedback on the recent changes.

msteiger:
This looks a lot better - good work, @indianajohn! One last thing: I'd prefer to have two separate TextRenderer instances for the two use cases: as-is/translated. The `ToStringTextRenderer` could remain untouched and a new `TranslatingTextRenderer` could perform the translation work.

Is it possible that the reported NPE is independent from the changes in the PR?

Cervator:
> Is it possible that the reported NPE is independent from the changes in the PR?

Anything is possible, sure. I did try deleting config.cfg a couple times, then stepped through code to narrow down the exact issue. Do you not get the error? If not then it could very well be some dirty workspace issue I'm missing :-)

indianajohn:
I was out of town on business and couldn't get to this until now, sorry.

I tested this and found that the game didn't launch properly on my workspace either. I tracked the problem down to the following: the default config values seem to be keyed to the string that the `toString()` method provides. I changed `defaults.cfg` to be keyed to the new value, deleted my `config.cfg` file, and rebuilt my workspace, and everything worked fine upon next launch.

I should mention that `default.cfg` copies itself to several different places when the game is built, so you should probably search for and delete files while rebuilding your workspace.

GooeyHub:

Cervator:
Tested out again and all seems well! After I deleted `config.cfg` that is :-) I renamed the title slightly to highlight that.

As usual no problem on delays, just as long as you keep coming back eventually :D

Adding some related comments to #2309 for further work in the area. Thanks for the PR!

PR #2399, opened 2016-07-08
[RFR] Architecture for Redundant State Elimination

tdgunes:
### Contains

Initial architecture for redundant state elimination, render task creation and so on.

### How to test

Play the game with different video settings and observe any irregularities.

PR #2405, opened 2016-07-13
[NFM] Debugging Nodes with F11

tdgunes:
### Contains

An addition for [#8](https://github.com/tdgunes/Terasology/issues/8) (_Rethink debug strategy in light of the DAG-based architecture_). Adds F11 key to stop the execution of the pipeline at a certain point. As shown below:

out

Please just focus on these files while reviewing:

* rendering/nui/layers/ingame/metrics/DebugOverlay.java
* resources/assets/ui/ingame/debugOverlay.ui
* rendering/dag/RenderTaskSystem.java
* logic/players/DebugControlSystem.java
* rendering/world/WorldRendererImpl.java (Especially usage of `RenderTaskSystem`)

### How to test

Start the game and after enabling debug mode, use F11 key. It's good to enable all of the available effects while testing.

It's based on #2399, so it includes changes in that branch. I will rebased this PR based on that.


Cervator:
Cooooool! I guess in a way this is the first super visible benefit to the pipeline setup? Looks like it would be hugely useful in 3d wizardry land :-)

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
```diff
@@ -142,6 +146,17 @@ public String get() {
                 }
             });
         }
+
+        UILabel debugLine5 = find("debugLine5", UILabel.class);
+        if (debugLine5 != null) {
+            debugLine5.bindText(new ReadOnlyBinding<String>() {
+                @Override
+                public String get() {
+                    return String.format("Executed final node task: %s", ((NodeTask) renderTaskSystem.getCursorTask()).getNode().getClass().getSimpleName());
```
emanuele3d: I'd suggest the string can just be

"Last node processed: %s"

because from a debugging perspective it's the nodes that will matter, not the tasks.

emanuele3d: Actually, let me refine it:

"Last RenderNode processed: %s". 

And somehow this debugging mechanism will probably have to be restricted specifically only to render nodes (nodes writing to buffers), as I imagine that if one day we have additional nodes, i.e. cameras, buffers, acceleration structures, it wouldn't make sense to stop execution at one of those.

Also, we'll have to think about nodes that generate outputs that are not immediately visualizable or that are in an attachment of the FBO that is not visualized, I'm thinking the depth map generated by the shadow node and things like the normal buffer in the FBO sceneOpaque.

But for the time being of course this helps!

engine/src/main/java/org/terasology/rendering/dag/RenderTaskSystem.java
```diff
@@ -0,0 +1,84 @@
...
+    public void toggle() {
+        do {
+            cursorTask = taskIterator.next();
+        } while (!(cursorTask instanceof NodeTask));
+    }
+
+    public void executeAll() {
+        for (RenderPipelineTask task : taskList) {
+            task.execute();
+            if (task == cursorTask) {
```
emanuele3d: Hmmmm... this is very problematic: for every executed task we now have an if statement to go through effectively doubling the number of things to do during rendering. 

I could forgive this approach if this method became "debuggingExecuteAll()" and the "executeAll()" didn't have the IF block in it, each method executed depending if debugging was enabled or not.

But in general I think it's a very dirty approach to visual debugging. For the time being I'd prefer you first find where the tasklist should be cut short, and then I'd either iterate over a range ending there or I'd make a shortened copy of the list before iterating over it.

Eventually visual debugging will probably have to be handled at the RenderGraph level, perhaps by having an "OutputNode" that gets attached to another node and displays its output.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskSystem.java
```diff
@@ -0,0 +1,84 @@
...
+import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.registry.Share;
+import static org.terasology.rendering.opengl.OpenGLUtils.bindDisplay;
+import static org.terasology.rendering.opengl.OpenGLUtils.renderFullscreenQuad;
+
+/**
+ * TODO: Add javadocs
+ */
+@RegisterSystem
+@Share(RenderTaskSystem.class)
```
emanuele3d: I think I find the creation of an entire system for this task not recommendable and the interaction it has with the WorldRenderer quite confusing. 

- I think the DebugControlSystem should inform the DebugOverlay (if enabled) that F11 has been pressed. 
- The DebugOverlay could then be an observable, the WorldRenderer being the observer.
- The DebugOverlay could then relay to the WorldRenderer the number of times F11 has been pressed.
- WorldRenderer then uses that number to process only a part of the task list.

This way you don't need to add a new BaseComponentSystem implementation.

Cervator:
Bump - this one hasn't moved for a little while, is it special somehow, waiting for something, ready for merge minus any problematic conflicts or just close in favor of a later redo of some sort? :-)

tdgunes:
I think it's better to close this since it's not in the roadmap right now. I'd like to revisit this after other issues are dealt with.

Cervator:
Ok! It looks like some of the code/comments/TODOs have made it in via the other PRs anyway, now that I check, also likely the reason for it being flagged for conflicts :-)

PR #2411, opened 2016-07-20
[Merge it! Merge it!] Rendering and the RenderingDebug configs now "subscribable", on a per-property basis.

emanuele3d:
- Added a number of "Subscribable" interfaces and abstract classes, allowing implementations to be monitored for general changes, specific changes or both.
- RenderingConfig and RenderingDebugConfig are now extending AbstractSubscribable. Should they extend AbstractSpecificSubscribable instead?
- Bonus Feature: added 4K UHD resolution among the possible Screenshot sizes.
- Eliminated 3 warnings related to private enum constructors - the "private" keyword is redundant in those circumstances.

**EDIT**: It should be noted that while individual parameters are, with this PR, monitorable, they are not yet monitored. This will come as @tdgunes work on the rendering engine proceeds.

### DIFF
--- a/engine/src/main/java/org/terasology/config/RenderingConfig.java
+++ b/engine/src/main/java/org/terasology/config/RenderingConfig.java
@@ -24,11 +24,57 @@ import org.terasology.rendering.world.viewDistance.ViewDistance;
 import org.terasology.utilities.subscribables.AbstractSubscribable;
 
 import java.beans.PropertyChangeListener;
-import java.beans.PropertyChangeSupport;
 
 /**
  */
 public class RenderingConfig extends AbstractSubscribable {
+
+    public static final String PIXEL_FORMAT = "PixelFormat";
+    public static final String WINDOW_POS_X = "WindowPosX";
+    public static final String WINDOW_POS_Y = "WindowPosY";
+    public static final String WINDOW_WIDTH = "WindowWidth";
+    public static final String WINDOW_HEIGHT = "WindowHeight";
+    public static final String FULLSCREEN = "FullScreen";
+    public static final String ANIMATED_MENU = "AnimatedMenu";
+    public static final String VIEW_DISTANCE = "viewDistance";
+    public static final String FLICKERING_LIGHT = "FlickeringLight";
+    public static final String ANIMATE_GRASS = "AnimateGrass";
+    public static final String ANIMATE_WATER = "AnimateWater";
+    public static final String DYNAMIC_SHADOWS = "DynamicShadows";
+    public static final String FIELD_OF_VIEW = "FieldOfView";
+    public static final String CAMERA_BOBBING = "CameraBobbing";
+    public static final String RENDER_PLACING_BOX = "RenderPlacingBox";
+    public static final String BLUR_INTENSITY = "BlurIntensity";
+    public static final String REFLECTIVE_WATER = "ReflectiveWater";
+    public static final String VIGNETTE = "Vignette";
+    public static final String MOTION_BLUR = "MotionBlur";
+    public static final String SSAO = "Ssao";
+    public static final String FILM_GRAIN = "FilmGrain";
+    public static final String OUTLINE = "Outline";
+    public static final String LIGHT_SHAFTS = "LightShafts";
+    public static final String EYE_ADAPTATION = "EyeAdaptation";
+    public static final String BLOOM = "Bloom";
+    public static final String OCULUS_VR_SUPPORT = "OculusVrSupport";
+    public static final String MAX_TEXTURE_ATLAS_RESOLUTION = "MaxTextureAtlasResolution";
+    public static final String MAX_CHUNKS_USED_FOR_SHADOW_MAPPING = "MaxChunksUsedForShadowMapping";
+    public static final String SHADOW_MAP_RESOLUTION = "ShadowMapResolution";
+    public static final String NORMAL_MAPPING = "NormalMapping";
+    public static final String PARALLAX_MAPPING = "ParallaxMapping";
+    public static final String DYNAMIC_SHADOWS_PCF_FILTERING = "DynamicShadowsPcfFiltering";
+    public static final String CLOUD_SHADOWS = "CloudShadows";
+    public static final String LOCAL_REFLECTIONS = "LocalReflections";
+    public static final String INSCATTERING = "Inscattering";
+    public static final String RENDER_NEAREST = "RenderNearest";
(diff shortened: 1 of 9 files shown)


emanuele3d:
To see the previous comments in their context:

https://github.com/MovingBlocks/Terasology/commit/0339630c9050223f81da06b86aeb26cbfa1e29e7#diff-36f2e62ff3c830150ab363d678aedcadR31

GooeyHub:

Cervator:
Can confirm this merges fine with latest develop (after merging in PR #2309) and runs the game, although I'm unsure what specifically might exercise new functionality. I just looked at graphic settings, tweaked stuff, loaded up in-game, took a sweet new 4k screenie, edited settings while in-world, and so on :-)

As for architecture / naming then I'm not sure but am pinging @tdgunes and @msteiger to see if they have interesting feedback, maybe others as well (@flo?)

emanuele3d:
@Cervator, you wrote:
> Can confirm this merges fine with latest develop (after merging in PR #2309) and runs the game, although I'm unsure what specifically might exercise new functionality.

At this point nothing is using the new functionality. Only the monitoring of the ViewDistance parameter is being used (I think - didn't double-check) but that was pre-existing. 

However, the new rendering engine architecture will take advantage of the monitorable config parameters. I just removed one item from tdgunes' todolist. =)

engine/src/main/java/org/terasology/config/RenderingConfig.java
```diff
@@ -491,10 +528,10 @@ public boolean isVSync() {
         return this.vSync;
     }
 
-    public static final String V_SYNC = "VSync";
-    public void setVSync(boolean vSync) {
-        this.vSync = vSync;
-        propertyChangeSupport.firePropertyChange(V_SYNC, !this.vSync, this.vSync);
+    public void setVSync(boolean vsync) {
```
emanuele3d: I'm not really sure here why vsync was flagged up as hiding a field while the method above (and others), using the input parameter "meshLimit", doesn't.

emanuele3d:
Unless @msteiger and @flo have architectural issues with this PR, I think it's ready to be merged.

GooeyHub:

emanuele3d:
After careful scientific studies, I have come to the conclusion that adding the suffix [Merge it! Merge it!] to the PR title does not increases the chances of said PR being properly reviewed nor merged.

Cervator:
Haha yes indeed. It also takes some more time availability :D

Will try to get to it later tonight my time :-)

Cervator:
Merged it! Merged it!

:-)

engine/src/main/java/org/terasology/config/RenderingConfig.java
```diff
@@ -423,65 +543,79 @@ public int getFrameLimit() {
...
     }
 
     public String getScreenshotFormat() {
         return screenshotFormat;
     }
 
     public void setScreenshotFormat(String screenshotFormat) {
-        this.screenshotFormat = screenshotFormat;
+
+        // propertyChangeSupport fires if oldFormat != newFormat which in the case of strings is always true.
```
msteiger: The JVM can throw them in a pool (in particular constants) and reuse them. For example, `"a" == "a"` should be `true`. Maybe we should not use the ancient property change support at all ..

engine/src/main/java/org/terasology/utilities/subscribables/AbstractGeneralSubscribable.java
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.utilities.subscribables;
+
+import java.beans.PropertyChangeListener;
+import java.beans.PropertyChangeSupport;
+
+/**
+ * {@inheritDoc}
```
msteiger: I think that this is not needed.

engine/src/main/java/org/terasology/utilities/subscribables/SpecificSubscribable.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.terasology.utilities.subscribables;
+
+import java.beans.PropertyChangeListener;
+
+/**
+ * Originally created to provide Observable-like functionality to config classes,
+ * it can be used by any class that wishes to be monitored for change in specific properties.
+ */
+public interface SpecificSubscribable {
```
msteiger: I would try to keep the number of interfaces and abstract base classes as small as needed. How about renaming `SpecificSubscribable` to `Subscribable` interface and `AbstractSpecificSubscribable` to `SubscribableConfig` (the `transient` makes it somewhat unique) and removing the other ones?

Cervator:
Oh hey, some 11th hour code review, thanks for the comments @msteiger :-)

engine/src/main/java/org/terasology/config/RenderingConfig.java
emanuele3d: @msteiger: Dear Martin, as Immortius is no longer as involved in the project as he used to, I'd be happy to refactor this again, in accordance with your preferences. However, I need an example pattern of what you'd like for a property. 

Also, @tdgunes will hopefully start using this functionality soon, to dynamically add and remove nodes from the RenderGraph as the config properties are modified. So, I'd like to deal with this quickly.

engine/src/main/java/org/terasology/utilities/subscribables/AbstractGeneralSubscribable.java
emanuele3d: You mean if there is no javadoc it will inherit it anyway?

engine/src/main/java/org/terasology/utilities/subscribables/SpecificSubscribable.java
emanuele3d: Short answer: if you really want me to. 0:)
Longer answer: what I was aiming for is to create a flexible solution, as some objects might need to be monitored on a per-property basis, some objects might need to be monitored as a whole and some objects might need to be monitored both ways. As there is no multiple inheritance in java I decided to provide all three possible combinations of abstract base class plus the two interfaces that can be composed together. Why is this not a good approach?

engine/src/main/java/org/terasology/utilities/subscribables/AbstractGeneralSubscribable.java
Cervator: Yeah would have to double check the docs but it is something like automatic inheritance unchanged, but if you want the original *and* something new use that tag to place the original somewhere. I think.

PR #2413, opened 2016-07-23
Merge weblate

Cervator:
The `weblate` branch had fallen quite a ways behind and it had been hard keeping it updated. I dug into the details and did a pile of merging and cleanup. Also grabbed the very latest pending stuff from Weblate.

I updated the plain `menu.lang` which had lagged behind substantially as new keys were added to the English file or maybe multiple language files, but not the template. That in turn likely means a bunch of potential translations weren't being served in Weblate.

### Conflict  Resolution

Italian
* just had a key with no value added by Weblate. Ignored it

Japanese
* both sides seemed to have valid translations and a few differed (can't tell which were better)
* accepted the GitHub version as it had more overall translations done
* Weblate did manage to add a new translation for "default-binding" but not the other 3 new keys - added this back in

Polish
* Identical both on GitHub and Weblate but Weblate had added a few valid keys not yet on GitHub's lang file (but in the template)
* Spliced in the few Weblate-side keys that were added, everything else seemed identical with GitHub adding newer keys as well

Russian
* Weblate just had values for the four missing keys so grabbed them

### Other

Missed keys caught by Weblate (added to the template so Weblate spotted them but manual GitHub translators didn't)
* "default-binding"
* "new-binding"
* "remove-binding"
* "reset-default"

Did an overhaul of translation keys in general, deleting outdated ones and adding new ones from the English file to the template

Thought is that only Weblate actually uses the template, meaning it might keep encouraging translations for keys no longer in use. It also likely means a lot of keys haven't shown up in Weblate at all since the template was missing them

Some other (especially rarely used) language files still may have excess or unsorted keys, but it is better than before.

Other stuff done: Deleted Esperanto and American English. Had no actual translation work applied, not sure why they were created/requested.

@XroMyla: You were listed on the Esperanto entry, did you want to actually apply translations to it? As a language it seems somewhat obscure.

Similar situation with Galician. We had just a few entries translated for it long ago yet it seems otherwise abandoned and fairly niche. Maybe we should delete that too.

PR just for documentation purposes, merging momentarily then updating the `weblate` branch.


GooeyHub:

Cervator:
Merged and added @XroMyla, @Translator5, and Koga Masato to the credits based on Weblate stats. Thanks all :-)

Let me know if anybody has forum accounts needing contributor badges and such.

Weblate is fully up to date again with the percentage having dropped some thanks to all the new keys in the template.

Issue #2412, opened 2016-07-22
Rendering Relevant Entities


skaldarnar:
While working on the magic dome for LaS I came around a specific problem. The dome basically is a sphere, placed at (0,0,0). I want this sphere to be always visible right away from the start. However, when the player spawns far outside so that the chunk around (0,0,0) is not loaded, no sphere is visible. 

I'm adding a `RelevanceRegionComponent` to the entity, and once the chunk is loaded (e.g., when the player gets near enough) it stays active.
https://github.com/Terasology/LightAndShadow/blob/master/src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java#L58-L69

I'm trying to get the sphere rendered without the necessity to have the player come nearby. Is the current behavior a bug as the entity should also be rendered being relevant? Is it possible to request a specfic chunk to be loaded? 

@immortius seems to be responsible for the implementation, so here's a ping for you. Not sure who else might have a clue about this, pingin @msteiger @Josharias @flo

flo:
Adding a RelevanceRegionComponent shoulds like a very bad idea to me as you force the server to have chunks loaded permanently that are not relevant to the user except for the sphere.

I would guess that your issue is that you create a always relevant entity during the generation of the chunk (0,0,0). So you need to the chunk to get generated to have the entity be created.

To avoid this you could create the entity in the `postBegin` of a system when the entity isn't there yet. Propably (since it seems to be a graphic that is only relevant to clients) it doesn't even need to persistent. So you could just always create a non persistent entity in the system.

Cervator:
Yep that's already the approach: https://github.com/Terasology/LightAndShadow/blob/feature-magic_dome/src/main/java/org/terasology/ligthandshadow/dome/MagicDomeSystem.java#L67

I'm not sure but I think it goes something like this:

* If the player spawns at 0,0,0 the dome is present and being rendered, even if you then teleport far away (you can see the dome from way further away than the chunk rendering distance - thanks to the relevance thing?)
* If the player spawns far away the dome is *created* still and its effects are felt (the player is pulled toward it) but it isn't being rendered until the player gets to a distance where the "owning" chunk would load normally. Which could be further in than the edge of the dome itself.

[...]

PR #2414, opened 2016-07-23
Check entity relevance for rendering

skaldarnar:
Solves #2412.

As stated in the linked issue, I was looking for a way to always render an entity's mesh, given that the entity is always relevant. Therefore, I changed the render check to include the relevance of the entity.

With this change, an entity that holds `MeshComponent` and `LocationComponent` with `alwaysRelevant` being `false` is handled like before, i.e., the mesh is only renderd when the chunk around it is loaded. 
Setting `alwaysRelevant` to `true` will cause the mesh to be rendered at the given location, regardless whether the chunk around is loaded or not. 

Pinging @immortius @flo @msteiger @Josharias @emanuele3d for quick review.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
@@ -18,6 +18,9 @@ package org.terasology.rendering.logic;
 import com.bulletphysics.linearmath.Transform;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.SetMultimap;
+import java.nio.FloatBuffer;
+import java.util.Arrays;
+import java.util.Set;
 import org.lwjgl.BufferUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -46,10 +49,6 @@ import org.terasology.rendering.assets.material.Material;
 import org.terasology.rendering.opengl.OpenGLMesh;
 import org.terasology.rendering.world.WorldRenderer;
 import org.terasology.world.WorldProvider;
-
-import java.nio.FloatBuffer;
-import java.util.Arrays;
-import java.util.Set;
 
 /**
  * TODO: This should be made generic (no explicit shader or mesh) and ported directly into WorldRenderer? Later note: some GelCube functionality moved to a module
@@ -192,8 +191,7 @@ public class MeshRenderer extends BaseComponentSystem implements RenderSystem {
                     MeshComponent meshComp = entity.getComponent(MeshComponent.class);
                     LocationComponent location = entity.getComponent(LocationComponent.class);
 
-                    if (isHidden(entity, meshComp) || location == null || meshComp.mesh == null
-                            || !worldProvider.isBlockRelevant(location.getWorldPosition())) {
+                    if (isHidden(entity, meshComp) || location == null || meshComp.mesh == null || !isRelevant(entity, location.getWorldPosition())) {
                         continue;
                     }
                     if (meshComp.mesh.isDisposed()) {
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
```diff
@@ -244,6 +242,21 @@ private void renderEntitiesByMaterial(SetMultimap<Material, EntityRef> meshByMat
         }
     }
 
+    /**
+     * Checks whether the entity at the given position is relevant.
+     * <p>
+     * The entity is at the given position is relevant if
```
emanuele3d: The first "is" should be removed I think. Should read "The entity at the given position...", right?

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
```diff
@@ -244,6 +242,21 @@ private void renderEntitiesByMaterial(SetMultimap<Material, EntityRef> meshByMat
...
+     * a) the entity itself is always relevant, or
+     * b) if the block at the position is relevant.
```
emanuele3d: the "if" should be removed as it is present in the line just below the bulleted list.

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
```diff
@@ -244,6 +242,21 @@ private void renderEntitiesByMaterial(SetMultimap<Material, EntityRef> meshByMat
...
+     *
+     * @param entity   the entity to check for relevance
+     * @param position the position the entity is located
```
emanuele3d: Would it be good to specify "worldPosition" instead?

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
```diff
@@ -244,6 +242,21 @@ private void renderEntitiesByMaterial(SetMultimap<Material, EntityRef> meshByMat
...
+     * @return true if the entity itself or the block at the given position are relevant, false otherwise.
+     */
+    private boolean isRelevant(EntityRef entity, Vector3f position) {
+        return entity.isAlwaysRelevant() || worldProvider.isBlockRelevant(position);
```
emanuele3d: Have you considered what is the most likely case if any? Is it more likely that an entity is always relevant or is it more likely that the block at the given position is relevant? If the latter is more likely you'd gain a smidgen of performance by inverting the two expressions.

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
emanuele3d: I'd probably also remove the ", or" one line above, but I now understand better how you tried to let it read like a whole sentence. It's up to you.

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
skaldarnar: Good point! As there should not be too many entities that have to be rendered all the time, the latter is indeed more likely I think.

GooeyHub:

Cervator:
@flo any remaining comments before we merge this? Referencing #2412 

@emanuele3d & @tdgunes - also a last reference ping, just in case :-)

flo:
The change is fine by me

flo:
It's good to have it fixed

Cervator:
Merged then :-)

Tested in Light & Shadow, although not thoroughly (kept spawning inside or near the dome this time).

While testing I discovered had to bump the ItemRendering dependency across a few modules, as it recently went from v.0.1 to v0.2. IIRC despite only being a minor version bump on IR since it is still below v1.0.0 the minor effectively counts as major for in-game compatibility checking. Actual compilation was fine, the issue only appeared at runtime.

skaldarnar:
Is the strict version checking intended? The patch to ItemRendering just added some constructor variants and convenience methods, thus it was backwards-compatible.

skaldarnar:
> Now that modules other than Core are far more active with more version bumps like this it might either be time to readdress again or just push for releasing more modules at v1.0.0 since then regular SemVer will apply :-)

Alright, sounds very reasonable. Let's try to push some modules to a proper 1.0.0 release then :)

PR #2415, opened 2016-07-23
[RFR] Smart property addition

rzats:
### Contains

An implementation for rzats/gsoc-nui-editor#8.

- The context menu screen and related systems are reworked to support multiple context menu levels.
- Property scaffolding has been added for `ARRAY` and `OBJECT` nodes. The exact list of functionality can be viewed in [NUI editor scaffolding @ Google Docs](https://docs.google.com/document/d/1WNOejL9ytLKrE3hTsH6-w8k35BxdbTaFTNkf0HfU5Js/edit?usp=sharing).

### How to test

Open the NUI editor (F10), use the "Add..." context menu option for `ARRAY` and `OBJECT` nodes.

- [x]  Context menu levels should be drawn right next to each other, but sometimes the deeper level is drawn slightly off to the right - off-by-margin error?
- [x] Add a "Create new screen" option to make the editor fully usable (will also improve testing for all of the below items as I can attempt to recreate a screen or two from the game and fix any outstanding issues that show up).
- [x] "Generic node" scaffolding needs some work as it doesn't do a great job picking up fields when e.g. a child of `layoutInfo`.
- [x] Better handling for object fields (should be expanded into separate children, perhaps recursively).
- [x] Better handling for `Binding` fields (should be initialized with their default values).


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+        while (layoutHintType == null) {
+            if (currentNode.hasSiblingWithKey("type")) {
+                type = (String) currentNode.getSiblingWithKey("type").getValue().getValue();
+                layoutMetadata = nuiManager.getWidgetMetadataLibrary().resolve(type, ModuleContext.getContext());
+
+                // If one is found, attempt to get the layout hint type.
+                try {
+                    layoutHintType = (Class) ReflectionUtil
+                        .getTypeParameter(layoutMetadata.getType().getGenericSuperclass(), 0);
+                } catch (NullPointerException e) {
```
flo: Catching a NullPointerException is really bad design. There should be a null check at the proper place.

Reasoning: 
1. it is then clearer what is null. 
2. Exceptions are very costly in java, 
3. when nullpointerexceptions are only used for errors then it is possible to set a breakpoint on NullPointerExceptions and have the program only stop there and not at hunderts of other places

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+        }
+    }
+
+    private void populateLayoutContextMenu(JsonTree node, ContextMenuLevel addLevel) {
+        String type = "";
+        JsonTree currentNode = node;
+        ClassMetadata layoutMetadata = null;
+        Class layoutHintType = null;
+
+        // Iterate down the tree and attempt to find a UILayout node.
```
flo: when you can comment a block like this it is a typical case for something that can be extracted into a method. Good methods are about 7 lines long.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+                        .getTypeParameter(layoutMetadata.getType().getGenericSuperclass(), 0);
+                } catch (NullPointerException e) {
+                    layoutHintType = null;
+                }
+            }
+            if (layoutHintType == null) {
+                if (currentNode.isRoot()) {
+                    break;
+                }
+                currentNode = (JsonTree) currentNode.getParent();
```
flo: That suprises me a bit, why is a recursive search upwards necessary?

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+        } else {
+            logger.warn("Unknown layout type: {}", type);
+        }
+    }
+
+    private void populateGenericContextMenu(JsonTree node, ContextMenuLevel addLevel) {
+        if (node.hasSiblingWithKey("type")) {
+            // Get the ClassMetadata of the widget with the given type.
+            String type = (String) node.getSiblingWithKey("type").getValue().getValue();
+            ClassMetadata elementMetadata = nuiManager.getWidgetMetadataLibrary().resolve(type, ModuleContext.getContext());
```
flo: Same gets done in the method populateLayoutContextMenu above. Could be extracted into a method.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+            if (layoutHintType == null) {
+                if (currentNode.isRoot()) {
+                    break;
+                }
+                currentNode = (JsonTree) currentNode.getParent();
+            }
+        }
+
+        if (layoutHintType != null) {
+            for (Field field : layoutHintType.getDeclaredFields()) {
```
flo: I wonder if that is the right way of doing it, as it is not the terasology way of getting field meta data.

That the field are annotated with FieldMetadata is one of the few classes that evaluates the SerializedName annotation. So I would assume that during widget creation the FieldMetadata gets also used as source of information about widget fields.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+        externalConsumers.put(key, value);
+    }
+
+    public void subscribeAddContextMenu(UpdateListener listener) {
+        addContextMenuListeners.add(listener);
+    }
+
+    public ContextMenuBuilder createPrimaryContextMenu(JsonTree node) {
+        ContextMenuBuilder contextMenu = new ContextMenuBuilder();
+        ContextMenuLevel primaryLevel = contextMenu.addLevel(LEVEL_PRIMARY);
```
flo: I am not really sure why this extra complexity with named levels is necessary. The constant LEVEL_PRIMARY seems to get only used here.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+        primaryLevel.setVisible(true);
+
+        JsonTreeValue.Type type = node.getValue().getType();
+
+        // Create the ADD_EXTENDED level.
+        if ((type == JsonTreeValue.Type.ARRAY && !node.getValue().getKey().equals("contents"))
+            || type == JsonTreeValue.Type.OBJECT) {
+            ContextMenuLevel addLevel = contextMenu.addLevel(LEVEL_ADD_EXTENDED);
+            primaryLevel.addOption(OPTION_ADD_EXTENDED, n ->
+                contextMenu.getLevel(LEVEL_ADD_EXTENDED).setVisible(!addLevel.isVisible()), null, false);
```
flo: Hiding a context menu should be responsibility of the context menu framework. Users of it should not have to deal with it. This makes it possible to improve later on the framework on when the context menu gets shown and have all users profit from it. e.g. it would be nice if in future the context menu would show on mouse over and would close when you hover a different item.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
rzats: Fixed!

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
rzats: This particular block can't be extracted into a method, since all of the variables declared above the comment are also used later. They could be wrapped within an inner class, but IMO that is unnecessarily complex.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
rzats: Agreed, but IMO the method doesn't need any changes since the recursive search will always stop at the second child anyway.
In the future sufficiently complex widgets (i.e. with widget fields of widget fields) might actually require recursion.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
rzats: As per above comment, both `type` and `elementMetadata` will need to be returned by the helper method.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
rzats: Agreed - even though `@SerializedName` isn't applied until after the field is retrieved, I think the approach you've mentioned is better.
Related question: how would I create a `ClassMetadata` based on `layoutHintType`? Using the [class constructor](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/reflection/metadata/ClassMetadata.java) is obviously not an option.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
rzats: The constants are temporary either way since I'm planning to apply i18n to the editor after all the strings are finalized. Using constants instead of inline values would make the refactor easier.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
rzats: Agreed! One way to fix this would be to have navigational-only options responsible for displaying and/or hiding further levels - the method to show/hide the level would be contained within the framework.

Not quite sure how to improve the approach w/ string maps, though - only one map per sub-menu is used.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
flo: That is fine. The constant above however gets never shown to the user as far as I know

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
flo: Yes navigation only options are exactly what I meant. When you add e navigation/submenu to the context menu builder you could just get a sub context menu builder that the context menu builder knows about. So the context menu builder can directly link nav option and sub context menu

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
rzats: I see! Fixed.

engine/src/main/java/org/terasology/rendering/nui/editor/NUIEditorContextMenuBuilder.java
rzats: Fixed (and removed level naming altogether)!

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

flo:
Thanks for the pull request! The editor really feels usable now.

PR #2419, opened 2016-07-26
[RFR] Adds SetViewportSizeOf StateChange 

tdgunes:
### Contains

As title mentions, adds a state change for [_All obvious State Changes implemented_](https://github.com/tdgunes/Terasology/issues/10)

### How to test

Check the log for any difference with given one below and any weirdness on shadow mapping and reflections. (Testing window resizing would be great!)

```
19:15:32.747 [main] INFO  o.t.r.dag.RenderTaskListGenerator - -- Task List --
19:15:32.748 [main] INFO  o.t.r.dag.RenderTaskListGenerator - SetViewportSizeOfTask: sceneShadowMap(1024x1024)
19:15:32.748 [main] INFO  o.t.r.dag.RenderTaskListGenerator -           BindFBOTask: sceneShadowMap(26)
19:15:32.748 [main] INFO  o.t.r.dag.RenderTaskListGenerator - NodeTask: ShadowMapNode
19:15:32.748 [main] INFO  o.t.r.dag.RenderTaskListGenerator - SetViewportSizeOfTask: sceneReflected(576x360)
19:15:32.748 [main] INFO  o.t.r.dag.RenderTaskListGenerator -           BindFBOTask: sceneReflected(11)
19:15:32.748 [main] INFO  o.t.r.dag.RenderTaskListGenerator - NodeTask: WorldReflectionNode
19:15:32.748 [main] INFO  o.t.r.dag.RenderTaskListGenerator - SetViewportSizeOfTask: sceneOpaque(1152x720)
19:15:32.748 [main] INFO  o.t.r.dag.RenderTaskListGenerator -           BindFBOTask: display(0)
19:15:32.748 [main] INFO  o.t.r.dag.RenderTaskListGenerator - NodeTask: BackdropNode
```

- [x] Any suggestions on resolving the dirty approach in stateChanges/SetViewportSizeOf.java, line 51? 

More state changes are coming!

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
@@ -28,6 +28,7 @@ import org.terasology.rendering.cameras.OrthographicCamera;
 import org.terasology.rendering.dag.AbstractNode;
 import org.terasology.rendering.dag.stateChanges.BindFBO;
 import org.terasology.rendering.dag.stateChanges.SetViewportSizeOf;
+import org.terasology.rendering.opengl.FrameBuffersManager;
 import org.terasology.rendering.primitives.ChunkMesh;
 import org.terasology.rendering.world.RenderQueuesHelper;
 import org.terasology.rendering.world.RenderableWorld;
@@ -49,6 +50,9 @@ public class ShadowMapNode extends AbstractNode {
     private static final String SCENE_SHADOW_MAP = "sceneShadowMap";
     private static final int SHADOW_FRUSTUM_BOUNDS = 500;
     public Camera shadowMapCamera = new OrthographicCamera(-SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, -SHADOW_FRUSTUM_BOUNDS);
+
+    @In
+    private FrameBuffersManager frameBuffersManager;
 
     @In
     private RenderableWorld renderableWorld;
@@ -79,8 +83,8 @@ public class ShadowMapNode extends AbstractNode {
         this.renderingConfig = config.getRendering();
         renderableWorld.setShadowMapCamera(shadowMapCamera);
 
-        addDesiredStateChange(new BindFBO(SCENE_SHADOW_MAP));
-        addDesiredStateChange(new SetViewportSizeOf(SCENE_SHADOW_MAP));
+        addDesiredStateChange(new BindFBO(SCENE_SHADOW_MAP, frameBuffersManager));
+        addDesiredStateChange(new SetViewportSizeOf(SCENE_SHADOW_MAP, frameBuffersManager));
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
@@ -27,7 +27,7 @@ import org.terasology.rendering.cameras.Camera;
 import org.terasology.rendering.cameras.OrthographicCamera;
 import org.terasology.rendering.dag.AbstractNode;
 import org.terasology.rendering.dag.stateChanges.BindFBO;
-import org.terasology.rendering.opengl.FBO;
+import org.terasology.rendering.dag.stateChanges.SetViewportSizeOf;
 import org.terasology.rendering.opengl.FrameBuffersManager;
 import org.terasology.rendering.primitives.ChunkMesh;
 import org.terasology.rendering.world.RenderQueuesHelper;
@@ -39,7 +39,6 @@ import static org.lwjgl.opengl.GL11.GL_COLOR_BUFFER_BIT;
 import static org.lwjgl.opengl.GL11.GL_CULL_FACE;
 import static org.lwjgl.opengl.GL11.GL_DEPTH_BUFFER_BIT;
 import static org.lwjgl.opengl.GL11.glClear;
-import static org.terasology.rendering.opengl.OpenGLUtils.setViewportToSizeOf;
 
 
 /**
@@ -48,6 +47,7 @@ import static org.terasology.rendering.opengl.OpenGLUtils.setViewportToSizeOf;
  * - https://docs.google.com/drawings/d/13I0GM9jDFlZv1vNrUPlQuBbaF86RPRNpVfn5q8Wj2lc/edit?usp=sharing
  */
 public class ShadowMapNode extends AbstractNode {
+    private static final String SHADOW_MAP_FBO_NAME = "sceneShadowMap";
     private static final int SHADOW_FRUSTUM_BOUNDS = 500;
     public Camera shadowMapCamera = new OrthographicCamera(-SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, SHADOW_FRUSTUM_BOUNDS, -SHADOW_FRUSTUM_BOUNDS);
 
@@ -70,11 +70,11 @@ public class ShadowMapNode extends AbstractNode {
     private BackdropProvider backdropProvider;
 
     private Material shadowMapShader;
-    private FBO shadowMap;
     private RenderingConfig renderingConfig;
     private Camera playerCamera;
     private float texelSize;
     private float stepSize;
+
 
     @Override
     public void initialise() {
@@ -82,7 +82,9 @@ public class ShadowMapNode extends AbstractNode {
         this.shadowMapShader = worldRenderer.getMaterial("engine:prog.shadowMap");
         this.renderingConfig = config.getRendering();
         renderableWorld.setShadowMapCamera(shadowMapCamera);
-        addDesiredStateChange(new BindFBO("sceneShadowMap", frameBuffersManager));
+
+        addDesiredStateChange(new BindFBO(SHADOW_MAP_FBO_NAME, frameBuffersManager));
+        addDesiredStateChange(new SetViewportSizeOf(SHADOW_MAP_FBO_NAME, frameBuffersManager));
     }
 
     @Override
@@ -93,9 +95,8 @@ public class ShadowMapNode extends AbstractNode {
         // TODO: check these conditions before adding this node inside the DAG, removing this condition completely from process()
         if (renderingConfig.isDynamicShadows() && worldRenderer.isFirstRenderingStageForCurrentFrame()) {
             PerformanceMonitor.startActivity("rendering/shadowMap");
-            shadowMap = frameBuffersManager.getFBO("sceneShadowMap");
+
             // preRenderSetupSceneShadowMap
-            setViewportToSizeOf(shadowMap);
             // TODO: verify the need to clear color buffer
             glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
             GL11.glDisable(GL_CULL_FACE);
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
@@ -28,7 +28,6 @@ import org.terasology.rendering.cameras.OrthographicCamera;
 import org.terasology.rendering.dag.AbstractNode;
 import org.terasology.rendering.dag.stateChanges.BindFBO;
 import org.terasology.rendering.dag.stateChanges.SetViewportSizeOf;
-import org.terasology.rendering.opengl.FrameBuffersManager;
 import org.terasology.rendering.primitives.ChunkMesh;
 import org.terasology.rendering.world.RenderQueuesHelper;
 import org.terasology.rendering.world.RenderableWorld;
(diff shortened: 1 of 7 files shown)


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
```diff
@@ -48,6 +47,7 @@
  * - https://docs.google.com/drawings/d/13I0GM9jDFlZv1vNrUPlQuBbaF86RPRNpVfn5q8Wj2lc/edit?usp=sharing
  */
 public class ShadowMapNode extends AbstractNode {
+    private static final String SHADOW_MAP_FBO_NAME = "sceneShadowMap";
```
emanuele3d: I'd remove _NAME, it feels distracting.

engine/src/main/java/org/terasology/rendering/dag/nodes/WorldReflectionNode.java
```diff
@@ -35,15 +35,14 @@
...
-import static org.terasology.rendering.opengl.OpenGLUtils.bindDisplay;
-import static org.terasology.rendering.opengl.OpenGLUtils.setViewportToSizeOf;
 
 /**
  * Diagram of this node can be viewed from:
  * TODO: move diagram to the wiki when this part of the code is stable
  * - https://docs.google.com/drawings/d/1Iz7MA8Y5q7yjxxcgZW-0antv5kgx6NYkvoInielbwGU/edit?usp=sharing
  */
 public class WorldReflectionNode extends AbstractNode {
+    private static final String REFLECTED_FBO_NAME = "sceneReflected";
```
emanuele3d: Same as above, I'd remove "_NAME".

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -0,0 +1,92 @@
...
+import org.terasology.rendering.dag.StateChange;
+import org.terasology.rendering.dag.tasks.SetViewportSizeOfTask;
+import org.terasology.rendering.opengl.FBO;
+import org.terasology.rendering.opengl.FrameBuffersManager;
+
+/**
+ * TODO: Add javadocs
+ */
+public final class SetViewportSizeOf implements FBOManagerSubscriber, StateChange {
+    private static final String DEFAULT_FBO_NAME = "sceneOpaque";
```
emanuele3d: I definitely have something visceral against _NAMEs.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
tdgunes: Hmmm, my thinking behind this is making distinction between a real `FBO` and a name of an `FBO`.  But it seems like you are strongly against `NAME` 😄, let me remove them! :tada:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -0,0 +1,92 @@
...
+        this.frameBuffersManager = frameBuffersManager;
+        fbo = frameBuffersManager.getFBO(fboName);
+    }
+
+    public SetViewportSizeOf() {
+        this.fboName = DEFAULT_FBO_NAME;
+    }
+
+    @Override
+    public StateChange getDefaultInstance() {
```
emanuele3d: Hmmm. Trickytrickytricky. I have a couple of thoughts.

1. Theoretically, the default instance could reset the viewport to the size of the window rather than the size of a buffer, and this would free the default instance from having to deal with the FBM. This would work if we accept that we need to query the size of the display every frame and monitor for changes. Perhaps a subscribable object could do the monitoring work, notifying subscribers such as the default instance if display changes.

2. In practice, how about you convert the frameBuffersManager variable into a static one, and you initialize it in the WorldRendererImpl, just after the frameBuffersManager is instantiated? This way the class will always have the instance of the FBM available and you can even shorten the constructors.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -0,0 +1,92 @@
...
+        return defaultInstance;
+    }
+
+    @Override
+    public RenderPipelineTask generateTask() {
+        if (task == null) {
+            task = new SetViewportSizeOfTask(fboName, fbo.width(), fbo.height());
+            frameBuffersManager.subscribe(this);
+        } else {
+            update();
```
emanuele3d: I'm ok with leaving this as it is, for safety, but I do wonder if the update() is really necessary here. I mean: the update heregets executed only if the task is not null. But that means the state change is subscribed to the fbm. And that means that whenever the buffer change the update gets called anyway, but only if needed.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        return task;
+    }
+
+    @Override
+    public boolean isEqualTo(StateChange stateChange) {
+        if (stateChange instanceof SetViewportSizeOf) {
+            return this.fboName.equals(((SetViewportSizeOf) stateChange).fboName);
```
emanuele3d: Doesn't this generate an error? fboName is private. Maybe this is a quirk of java I don't know.  In any case I'd implement a getter so that we can write:

`return this.fboName.equals(((SetViewportSizeOf) stateChange).getFboName());`

tdgunes: No, not at all. Just searched the situation and found a similar case in [StackOverflow](http://stackoverflow.com/questions/17027139/access-private-field-of-another-object-in-same-class). It seems that `private` keyword in Java hides away from other classes to access the variable. But since the access is happening inside the same class, it compiles & works. From design perspective, it would be good to have a getter for this.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
emanuele3d: Thank you. In any case the type of the variable gives it away that it cannot be the FBO itself. =)

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
emanuele3d: Oh, I see. Thank you for the link to the SO answer, that was informative. Given that it works I guess we do not strictly need to change it. But it might distract somebody reading through the code as it is, I feel, unintuitive. I therefore continue to recommend a getter.

emanuele3d:
Reviewed first two commits. Just minor issues apart perhaps the the slightly more delicate choice regarding how to get the FBM into the default instance. 

I'd also recommend to add a few more StateChange implementations and uses before we request to merge it in.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
```diff
@@ -23,7 +23,9 @@
  * TODO: Add javadocs
  */
 public final class SetWireframe implements StateChange {
-    private static SetWireframe defaultInstance = new SetWireframe(false);
+    private static final boolean DEFAULT_VALUE = false;
```
emanuele3d: I don't understand what was wrong with the simple approach for this class. The other state changes are more debatable because they need the FBM, but this class was good as it was.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -250,11 +252,20 @@ private void initRenderGraph() {
...
+
         List<Node> orderedNodes = renderGraph.getNodesInTopologicalOrder();
         RenderTaskListGenerator generator = new RenderTaskListGenerator();
         renderPipelineTaskList = generator.generateFrom(orderedNodes);
     }
 
+    private void initDefaultStateChanges() {
+        SetViewportSizeOf.setDefaultInstance(new SetViewportSizeOf(frameBuffersManager));
+        BindFBO.setDefaultInstance(new BindFBO());
+        SetWireframe.setDefaultInstance(new SetWireframe());
```
emanuele3d: I am sorry but this way of doing things didn't quite "click" with me. 

The SetWireframe state change is quite simple due to its boolean nature and shouldn't need any pre-initialization: initialization on instantiation is enough.

For the other two state changes, here in WorldRendererImpl.initDefaultStateChanges() I'd just write:
```
SetViewportSizeOf.setFrameBuffersManager(fbm);
BindFBO.setFrameBuffersManager(fbm);
```

In both StateChanges there should be:

`private static FrameBuffersManager fbm;`

and

public setFrameBuffersManager(FrameBuffersManager fbm) {
    this.fbm = fbm;
}

In so doing, the constructors won't need the FBM parameter in them:

```
SetViewportSizeOf(String fboName);
BindFBO(String fboName);
```

and we won't need the setDefaultInstance() method.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -0,0 +1,95 @@
...
+import org.terasology.rendering.dag.RenderPipelineTask;
+import org.terasology.rendering.dag.StateChange;
+import org.terasology.rendering.dag.tasks.SetViewportSizeOfTask;
+import org.terasology.rendering.opengl.FBO;
+import org.terasology.rendering.opengl.FrameBuffersManager;
+
+/**
+ * TODO: Add javadocs
+ */
+public final class SetViewportSizeOf implements FBOManagerSubscriber, StateChange {
```
emanuele3d: Where is this class subscribing to FBM changes?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
```diff
@@ -33,11 +35,21 @@ public SetWireframe(boolean enabled) {
...
+    }
+
+
     @Override
     public StateChange getDefaultInstance() {
         return defaultInstance;
     }
 
+
+    public static void setDefaultInstance(SetWireframe defaultInstance) {
```
emanuele3d: To be removed in my opinion.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
```diff
@@ -33,11 +35,21 @@ public SetWireframe(boolean enabled) {
...
+        SetWireframe.defaultInstance = defaultInstance;
+    }
+
     @Override
     public RenderPipelineTask generateTask() {
         if (enabled) {
```
emanuele3d: Cut&paste error on the second return: shouldn't it be "disablingTask"?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
```diff
@@ -33,11 +35,21 @@ public SetWireframe(boolean enabled) {
         this.enabled = enabled;
     }
 
+    public SetWireframe() {
```
emanuele3d: I'd remove this: it's not necessary for what I can see. Just do:

private static SetWireframe defaultInstance = new SetWireframe(DEFAULT_VALUE);

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
```diff
@@ -30,17 +30,17 @@
...
         this.fboName = fboName;
         this.frameBuffersManager = frameBuffersManager;
         fboId = frameBuffersManager.getFBO(fboName).fboId;
     }
 
-    private BindFBO() {
+    public BindFBO() {
         this.fboName = DEFAULT_FRAME_BUFFER_NAME;
         fboId = DEFAULT_FRAME_BUFFER_ID;
     }
```
emanuele3d: I'd like to remove this method. Just write

`private static BindFBO defaultInstance = new BindFBO(DEFAULT_FRAME_BUFFER_NAME);`

and also

```
public BindFBO(String fboName) {
    this.fboName = fblName;
    if (fboName != DEFAULT_FRAME_BUFFER_NAME) {
        fboId = frameBuffersManager.getFBO(fboName).fboId;
        // also add here functionality to subscribe to the FBM!
    } else {
        fboId = DEFAULT_FRAME_BUFFER_ID;
    }
}
```

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
```diff
@@ -50,6 +50,11 @@ public StateChange getDefaultInstance() {
         return defaultInstance;
     }
 
+
+    public static void setDefaultInstance(BindFBO defaultInstance) {
```
emanuele3d: In line with what written below, I don't think we need a setDefaultInstance method.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -0,0 +1,95 @@
...
+    public SetViewportSizeOf(FrameBuffersManager frameBuffersManager) {
+        this(DEFAULT_FBO, frameBuffersManager);
+    }
+
+    @Override
+    public StateChange getDefaultInstance() {
+        return defaultInstance;
+    }
+
+    public static void setDefaultInstance(SetViewportSizeOf defaultInstance) {
```
emanuele3d: In line with what written below this should be removed.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
tdgunes: Please consider #2421 on this file. Changes in this PR will be overridden by that.

emanuele3d: Thank you!!

Cervator:
FYI: Merged as part of #2421.

PR #2425, opened 2016-07-31
[RFR] A Minor Refactor in StateChanges

tdgunes:
### Contains

As title mentions, adds some necessary refactors according to given feedback on #2419 for [_All obvious State Changes implemented_](https://github.com/tdgunes/Terasology/issues/10)

### How to test

Check any weirdness on rendering, especially during the startup.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
@@ -31,17 +31,10 @@ public final class BindFBO implements FBOManagerSubscriber, StateChange {
     private static FrameBuffersManager frameBuffersManager;
     private static BindFBO defaultInstance = new BindFBO(DEFAULT_FRAME_BUFFER_NAME);
     private String fboName;
-
-    private int fboId;
     private BindFBOTask task;
 
     public BindFBO(String fboName) {
         this.fboName = fboName;
-        if (!fboName.equals(DEFAULT_FRAME_BUFFER_NAME)) {
-            fboId = frameBuffersManager.getFBO(fboName).fboId;
-        } else {
-            fboId = DEFAULT_FRAME_BUFFER_ID;
-        }
     }
 
     public String getFboName() {
@@ -65,13 +58,15 @@ public final class BindFBO implements FBOManagerSubscriber, StateChange {
     @Override
     public RenderPipelineTask generateTask() {
         if (task == null) {
-            task = new BindFBOTask(fboId, fboName);
             // Subscription is only needed if fboID is different than default frame buffer id.
-            if (fboId != DEFAULT_FRAME_BUFFER_ID) {
+            if (!fboName.equals(DEFAULT_FRAME_BUFFER_NAME)) {
+                task = new BindFBOTask(frameBuffersManager.getFBO(fboName).fboId, fboName);
                 frameBuffersManager.subscribe(this);
+            } else {
+                task = new BindFBOTask(DEFAULT_FRAME_BUFFER_ID, DEFAULT_FRAME_BUFFER_NAME);
             }
         } else {
-            if (fboId != DEFAULT_FRAME_BUFFER_ID) {
+            if (!fboName.equals(DEFAULT_FRAME_BUFFER_NAME)) {
                 update();
             }
         }
@@ -88,12 +83,11 @@ public final class BindFBO implements FBOManagerSubscriber, StateChange {
 
     @Override
     public void update() {
-        fboId = frameBuffersManager.getFBO(fboName).fboId;
-        task.setFboId(fboId);
+        task.setFboId(frameBuffersManager.getFBO(fboName).fboId);
     }
 
     @Override
     public String toString() { // TODO: used for logging purposes at the moment, investigate different methods
-        return String.format("%21s: %s(%s)", this.getClass().getSimpleName(), fboName, fboId);
+        return String.format("%21s: %s", this.getClass().getSimpleName(), fboName);
     }
 }
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
@@ -37,10 +37,10 @@ public final class BindFBO implements FBOManagerSubscriber, StateChange {
 
     public BindFBO(String fboName) {
         this.fboName = fboName;
-        if (fboName.equals(DEFAULT_FRAME_BUFFER_NAME)) {
-            fboId = DEFAULT_FRAME_BUFFER_ID;
-        } else {
+        if (!fboName.equals(DEFAULT_FRAME_BUFFER_NAME)) {
             fboId = frameBuffersManager.getFBO(fboName).fboId;
+        } else {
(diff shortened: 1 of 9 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
```diff
@@ -28,33 +28,35 @@
...
+
     private int fboId;
     private BindFBOTask task;
 
-    public BindFBO(String fboName, FrameBuffersManager frameBuffersManager) {
+    public BindFBO(String fboName) {
         this.fboName = fboName;
-        this.frameBuffersManager = frameBuffersManager;
-        fboId = frameBuffersManager.getFBO(fboName).fboId;
+        if (fboName.equals(DEFAULT_FRAME_BUFFER_NAME)) {
```
emanuele3d: I think the most likely case is that fboName is not the default frame buffer. But it's a very minor issue, so let's leave it as it is.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -27,35 +27,33 @@
...
-    private FrameBuffersManager frameBuffersManager;
 
-    public SetViewportSizeOf(String fboName, FrameBuffersManager frameBuffersManager) {
+    public SetViewportSizeOf(String fboName) {
         this.fboName = fboName;
-        this.frameBuffersManager = frameBuffersManager;
-        fbo = frameBuffersManager.getFBO(fboName);
+        if (!fboName.equals(DEFAULT_FBO)) {
+            fbo = frameBuffersManager.getFBO(fboName);
+        }
```
emanuele3d: I'm not sure about this IF block. Why do we need it? Can we just store only the fboName and then get the fbo itself only when we need it?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -86,6 +84,9 @@ public void update() {
 
     @Override
     public String toString() { // TODO: used for logging purposes at the moment, investigate different methods
+        if (isTheDefaultInstance()) { // necessary since default instance is generated before setFrameBuffersManager
+            fbo = frameBuffersManager.getFBO(fboName);
+        }
```
emanuele3d: I can see what you are doing here, attempting to avoid an NPE if fbo is still null. However, an instance of this class doesn't need knowledge of the width and height of the fbo except when it generates the task or updates it. 

So, I'd remove this IF block and I'd remove the printout of fbo.width() and fbo.height(). Those can stay in the task.

Also, I'd remove the private fbo variable: we can instantiate one briefly when we need to gets its width and height, in generateTask() and update().

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
emanuele3d: (...some time later...) 

I'm pretty convinced that you can remove the whole IF block, in line with the comment regarding the toString() method. It's important that the instance of this state change holds the fboName. But the fbo itself is only needed briefly when the task is generated or when the FBM signals a change through the update() method. So, I don't think we need to store the fbo permanently nor keep it up to date.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
emanuele3d: On second thought, given that there are other changes to do to this PR, let's change it to

`if (!fboName...`

emanuele3d:
A few quick issues to deal with and also this PR will be good to go.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
tdgunes: The reason why I've added these if-blocks because of logging fob's width and height. If it's ok to eliminate them from the log, we can definitely move on with your suggestion. Let me make these changes in my next commits.

emanuele3d: Yes, I confirm it's ok to eliminate width and height from the log. We anyway get width and height from the task log, right?

tdgunes: Sure, of course. It was just staying there for validating whether correct viewport state change is selected or removed.

emanuele3d: Just to be clear, the viewport state change should also stay in the log, when the intermediate list is printed out. It just doesn't need to printout the width and height information, only the fboName. I'm not sure if I was clear on this.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
```diff
@@ -37,10 +37,10 @@
 
     public BindFBO(String fboName) {
         this.fboName = fboName;
-        if (fboName.equals(DEFAULT_FRAME_BUFFER_NAME)) {
-            fboId = DEFAULT_FRAME_BUFFER_ID;
-        } else {
+        if (!fboName.equals(DEFAULT_FRAME_BUFFER_NAME)) {
             fboId = frameBuffersManager.getFBO(fboName).fboId;
+        } else {
+            fboId = DEFAULT_FRAME_BUFFER_ID;
```
emanuele3d: Sorry. I was hoping this PR could get through but I still it has some issues to deal with. 

In line with what we discussed regarding SetViewportSizeOf here it makes sense to just store the fboName and not worry about the fboId. The fboId is not needed yet AND if it was needed we'd have to subscribe to the FBM right away.

So, I'd remove the IF/ELSE block completely. The fboId can be fetched within the update() method and the generateTask() method. I'd also remove the fboId variable completely: let the BindFBOTask class print it out in its toString() method.

Finally, please review the generateTask method. it can check if `fboName != DEFAULT_FRAME_BUFFER_NAME` instead of checking the id. There's a small performance loss (integer vs string comparison) but it's not hot code so I don't think we shouldn't worry about it. You might also want to add an uninitialized static variable defaultTask. In generateTask() it would get filled by an unchangeable default task.

emanuele3d: That been said, I just reviewed the rest of this PR and apart from these issues I couldn't find any other problem. So, we are almost there with this one too. :smiley:

GooeyHub:

GooeyHub:

emanuele3d:
Ok, for some reason github didn't notify me that the latest commits have been submitted after my last visit. That's why I got confused and didn't realise that my feedback has been addressed. But addressed it has been and I think this PR can be merged at your convenience @Cervator.

Good job @tdgunes!

Cervator:
Merged! :-)

emanuele3d:
Thank you @Cervator! Good job @tdgunes!

PR #2426, opened 2016-07-31
[RFR] Added TODOs for future work on ShaderParameter classes

emanuele3d:
The code of ShaderParameter implementations will have to find its way into GLSLMaterial implementations or within the rendering engine's Node implementations. As the ShaderParameter implementations become empty, they will be removed.

This PR:
- provides no functional changes
- adds a large number of TODOs for future work
- in ShaderParametersChunk small code changes have been made to obtain the renderingConfig only once, add some empty lines and move a program.setFloat4 line to remove an otherwise unnecessary IF block.

I've ignored a large number of pre-existing CheckStyle warnings. We will deal with them as we move the ShaderParameter code into materials and nodes.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/shader/ShaderParametersBase.java
+++ b/engine/src/main/java/org/terasology/rendering/shader/ShaderParametersBase.java
@@ -43,9 +43,11 @@ public class ShaderParametersBase implements ShaderParameters {
 
         program.setFloat("viewingDistance", CoreRegistry.get(Config.class).getRendering().getViewDistance().getChunkDistance().x * 8.0f, true);
 
+        // TODO: obtain once in superclass?
         WorldRenderer worldRenderer = CoreRegistry.get(WorldRenderer.class);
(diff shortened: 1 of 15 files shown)


emanuele3d:
@tdgunes: this is work that at some point will need to be done, but only after the GSOC project ends.

ShaderParameters implementations however establish hidden dependencies between nodes, i.e. reading the content of a buffer attached to an FBO. So, we'll have to be careful when the functionality to add edges (intended as execution dependencies) is added to the engine: a node might be dependent on a different set nodes if the shader parameters classes are or aren't taken into account.

Cervator:
So do we just merge this right away or still wait for review of the TODOs?

emanuele3d:
Hehe, if it was up to me I'd merge it right away, but there is a reason PR are subject to the review process! =)

Let's leave it open for a few days, just in case. Perhaps @tdgunes might want to review the few actual code changes, to have them go through an additional pair of eyes' scrutiny.

Issue #2418, opened 2016-07-26
Graphic artifacts possible after changing video settings in-game


Cervator:
Found this while testing #2399 but confirmed as pre-existing in Omega build 510 (July 19th 2016)

Appears to only activate with animated water _and something else_ enabled. I didn't narrow down the other factor but if I set graphic detail to Ultra and enable everything else it showed up. Disabling animated water made it go away again. Leaving graphic settings on default + animated water however did _not_ provoke the issue.

Considering this low priority as the animated (waving) water is somewhat tricky on its own, see  #1477

Screenie demonstrating the artifacts - curious if anybody else seems them as well (Radeon line of GPU here):

![terasology-160725235702-1152x720](https://cloud.githubusercontent.com/assets/1063833/17125863/632bcda8-52c5-11e6-837b-4902f734a9a5.jpg)
##

tdgunes:
Thanks for reporting this one @Cervator. I'll look into this.

Cervator:
Cool, keep an eye out if you like but don't worry about it vs. more important stuff for now :-)

Cervator:
Another observation: parallax mapping is likely related (another immature option that may have never been fully finished or broke after it was). Found a combination where I could disable animated water yet still get the artifacts, disabling parallax mapping after that then removed the artifacts.

emanuele3d:
Managed to reproduce it on an nVidia GPU (860M). It's quite erratic though: it's not always straightforward to reproduce. It seems to be a more general problem though. Many settings when changed seem to display problems. For example, disabling motion blur made the screen go black at one point. I managed to reproduce it multiple times and then I couldn't anymore. Also, the animated grass settings seems to be ineffective: the grass remains always animated.

So, perhaps we should rename this issue to something more general, like Graphic Artifacts and video-settings problems.

Cervator:
Renamed!

4Denthusiast:
It looks like this is caused by the default bump-map texture being all 0, which when interpreted with an offset of -0.05 leads to everything having an offset, which with parallax-mapping enabled leads to it attempting to fetch textures from beyond the edge of the correct one, hence the bands of different colours near the edges of blocks. For some reason the minimal settings I can reproduce this on are Parallax mapping + Extra lighting, which makes me suspect that parallax mapping doesn't work without extra lighting.

[...]

PR #2427, opened 2016-07-31
[Good To Merge] New StateChanges for DAG

tdgunes:
~~**Important note for reviewers:** This PR is based on #2425, please not merge after that one is merged~~.

### Contains

As title mentions, adds five new state changes for [_All obvious State Changes implemented_](https://github.com/tdgunes/Terasology/issues/10)

### How to test

Please check any weirdness in the log and rendering of reflections, hdr and shadows.


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -120,7 +119,11 @@ private void generateIntermediateList(List<Node> orderedNodes) {
             intermediateList.add(node);
             // Add state changes to reset all desired state changes back to default.
             for (StateChange stateChange : node.getDesiredStateChanges()) {
-                intermediateList.add(stateChange.getDefaultInstance());
+                // TODO: consider using optional here?
+                StateChange defaultInstance = stateChange.getDefaultInstance();
+                if (defaultInstance != null) {
```
emanuele3d: I'm assuming later in this PR there is a _GOOD_ reason for a default instance to _EVER_ be null. :smiling_imp:

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
```diff
@@ -81,6 +80,8 @@ public void initialise() {
 
         addDesiredStateChange(new BindFBO(SCENE_SHADOW_MAP));
         addDesiredStateChange(new SetViewportSizeOf(SCENE_SHADOW_MAP));
+        addDesiredStateChange(new Clear(true, true, false));
```
emanuele3d: I don't think clearing operations are state changes. It's more of a degenerate type of rendering: you fill the buffer(s) with one color rather than a pretty picture. So, I'd leave clearing statements within the Node.process() methods and eventually verify and if they are actually useful. I feel that some are unnecessary rather than redundant.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
```diff
@@ -81,6 +80,8 @@ public void initialise() {
...
+        addDesiredStateChange(new DisableCapability(GL_CULL_FACE));
```
emanuele3d: I'd rename DisableCapability to simply Disable. But first let me see how you implemented this as right now I'm not sure if there should be a generic Disable state change or if there should be a DisableCullFace state change instead.

engine/src/main/java/org/terasology/rendering/dag/nodes/ToneMappingNode.java
```diff
@@ -15,41 +15,25 @@
...
-    private Config config;
-
-    private Material toneMapping;
-    private FBO sceneToneMapped;
-    private FBO sceneOpaque;
 
     @Override
     public void initialise() {
-        toneMapping = worldRenderer.getMaterial("engine:prog.hdr"); // TODO: rename shader to toneMapping)
+        addDesiredStateChange(new EnableShader("engine:prog.hdr"));
```
emanuele3d: Let's rename this to EnableMaterial. A material is the shader + all its configuration parameters.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableShader.java
```diff
@@ -0,0 +1,75 @@
...
+        this.shaderName = shaderName;
+    }
+
+    public String getShaderName() {
+        return shaderName;
+    }
+
+    @Override
+    public StateChange getDefaultInstance() {
+        return null;
```
emanuele3d: The default instance of this state change should generate a task calling `glUseProgram(0);` to disable the shader/material.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableShader.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public RenderPipelineTask generateTask() {
+        if (task == null) {
+            Material shader = getMaterial(shaderName);
```
emanuele3d: getMaterial really returns a material, not a shader. Shader objects are wrapped by the GLSLShader class.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * CullFace mode defined as an integer value in lwjgl, however in OpenGL documentation mode can be only three
+     * symbolic constants, therefore Mode enumeration ensures just these values are given to this state change, apart
+     * from having an integer with many different values.
+     */
+    public enum Mode {
+        GL_BACK,
+        GL_FRONT,
+        GL_FRONT_AND_BACK
+    }
```
emanuele3d: Hmmmmm.... nah. =)

That's overcomplicating I think. I like the immediacy of SetCullFace(GL_FRONT_AND_BACK). Just add a check in the constructor to make sure you are getting one of the three values and thrown an exception otherwise.

engine/src/main/java/org/terasology/rendering/dag/tasks/SetCullFaceTask.java
```diff
@@ -0,0 +1,53 @@
...
+import org.terasology.rendering.dag.stateChanges.SetCullFace;
+
+/**
+ * TODO: Add javadocs
+ */
+public class SetCullFaceTask implements RenderPipelineTask {
+    private int modeValue;
+    private SetCullFace.Mode mode;
+
+    public SetCullFaceTask(SetCullFace.Mode mode) {
```
emanuele3d: As you switch back to OpenGL integer constants rather than enums, this whole method will be drastically simplified as you'll be able to rely on one of the three correct integers coming through. However, you will need some kind of integer to meaningful string translation for the toString() method.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -539,6 +539,8 @@ private boolean isLightVisible(Vector3f positionViewSpace, LightComponent compon
 
     @Override
     public boolean isHeadUnderWater() {
+        // TODO: Making this as a subscribable value especially for node "ChunksRefractiveReflectiveNode",
+        // TODO: glDisable and glEnable state changes on that node will be dynamically added/removed based on this value.
```
emanuele3d: Hmmmm... careful here. Adding/removing nodes will probably be a relatively expensive operation. The head going under/above water might happen multiple times in a short period. So, we'll have to think carefully about this. Perhaps rephrase this as a question: // TODO: review - what happens RenderGraph-wise when the head goes under water?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableCapability.java
```diff
@@ -0,0 +1,62 @@
...
+package org.terasology.rendering.dag.stateChanges;
+
+import org.terasology.rendering.dag.RenderPipelineTask;
+import org.terasology.rendering.dag.StateChange;
+import org.terasology.rendering.dag.tasks.DisableCapabilityTask;
+
+/**
+ * TODO: Add javadocs
+ */
+public class DisableCapability implements StateChange {
```
emanuele3d: I am not quite convinced about this DisableCapability/EnableCapability pattern. I think each capability that can be enabled/disabled should have a single state change, like we did for wireframe. If the capability is enabled by default we can call the class "Disable<Something>". If the capability is disabled by default we can call the class "Enable<Something>". Which makes me think by the way SetWireframe should probably be called "EnableWireframe". But we can leave it as it is for now.

emanuele3d: By the way, this and other previous comments should remove the necessity to handle the case where the default instance or the generate task return null.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
emanuele3d: In light of what I wrote yesterday I think this should specifically become something like 

`new SetCullFace(enabled)`

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
tdgunes: Hmm, well what I had in mind was not adding/removing nodes whenever player dives into to the water. I thought what if for instance ChunksRefractiveReflectiveNode is informed about this event and state changes added or removed according to this event.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
emanuele3d: Reviewing this again. I'm getting confused between enabling culling and the command that sets _what_ should be culled. I'm thinking if a StateChange _enables_ or _disables_ something we are dealing with a boolean parameter and we can call that state change `EnableSomething` or `DisableSomething`, depending on what's the default. The constructor parameters than can be just true or false, i.e.:

`new EnableFaceCulling(true)`

On the other hand a state change that sets a more complex type of parameter, i.e. what to cull, can be named SetSomething, i.e.:

`new SetFacesToCull(GL_BACK)`

I hope this clarifies things a little.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: Sorry, I misread the comment. I now understand you were referring to adding/removing StateChange instances. We can try. Perhaps it won't be much of a performance hit. Certainly we should try to have no instantiation occurring during the switches. I.e. have everything ready at initialization and then only change the content of desiredStateChanges set. 

That been said changes in the set at this stage will always trigger a refresh of the whole intermediate and task lists. I'd be surprised if this doesn't have any impact. For this reason I suspect that whatever happens when the head is underwater might have to be handled with an if block within a Node.process() method, just to keep everything very responsive.

emanuele3d:
@tdgunes: what's happening with this PR? I'd really like to see this one closed before the FBO one proceeds. Furthermore, I suspect we need to make a pause here and assess what's happening with the issue #2418. Perhaps if this PR is completed and all nodes take advantage of the new state changes that issue might get solved automagically.  But I wouldn't bet on it.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableMaterial.java
```diff
@@ -25,7 +25,7 @@
 /**
  * TODO: Add javadocs
  */
-public class EnableMaterial implements StateChange {
+public final class EnableMaterial implements StateChange {
```
emanuele3d: Interesting. What's the reasoning for this?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
```diff
@@ -15,48 +15,55 @@
...
+import org.terasology.rendering.dag.tasks.DisableCapabilityTask;
 import org.terasology.rendering.dag.tasks.EnableCapabilityTask;
 
 /**
  * TODO: Add javadocs
  */
-public class EnableCapability implements StateChange {
-    private int capability;
-    private EnableCapabilityTask task;
+public final class SetBlend extends SetCapability {
```
emanuele3d: The term "capability" conflicts with the OpenGL concept with the same name, detailing the capabilities of a particular opengl implementation. For this reason I'd recommend to extend a class named SetStateParameter instead, to remove ambiguities.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableMaterial.java
tdgunes: As far as I know, it's for denoting that this state change implementation can not used as a subclass in another state change implementation. There were already `final` keywords in other state changes. So for sake of consistency, I added that.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
emanuele3d: Also, this is a boolean parameter. For these circumstances I suggested using something such as EnableBlend(true) and EnableBlend(false).

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
```diff
@@ -15,48 +15,55 @@
...
+    private static final int CAPABILITY = GL_BLEND;
+    private static StateChange defaultInstance = new SetCullFace(false);
```
emanuele3d: Err... does this have anything to do with setting the blend parameter?

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
```diff
@@ -15,48 +15,55 @@
...
+        return defaultInstance;
     }
 
     @Override
-    public RenderPipelineTask generateTask() {
-        if (task == null) {
-            task = new EnableCapabilityTask(capability);
+    protected RenderPipelineTask getDisablingTask() {
+        if (disablingTask == null) {
+            disablingTask = new DisableCapabilityTask(CAPABILITY);
```
emanuele3d: I think you should create the Disable/Enable tasks on construction. This would eliminate these two methods and keep this class and the SetStateParameter (currently SetCapability) leaner.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCapability.java
```diff
@@ -17,46 +17,50 @@
...
     }
 
     @Override
     public RenderPipelineTask generateTask() {
-        if (task == null) {
-            task = new DisableCapabilityTask(capability);
+        if (isEnabled()) {
+            return getEnablingTask();
+        } else {
+            return getDisablingTask();
```
emanuele3d: In line with the comment above you can remove the getEnabling/DisablingTask calls and simply replace them with return enablingTask and return disablingTask. I'd also just check for if(enabled) and eliminate isEnabled() unless there is another reason to keep it.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
```diff
@@ -15,38 +15,23 @@
...
 import org.terasology.rendering.dag.StateChange;
-import org.terasology.rendering.dag.tasks.SetCullFaceTask;
+import org.terasology.rendering.dag.tasks.DisableCapabilityTask;
+import org.terasology.rendering.dag.tasks.EnableCapabilityTask;
 
 /**
  * TODO: Add javadocs
  */
-public class SetCullFace implements StateChange {
+public final class SetCullFace extends SetCapability {
```
emanuele3d: In line with comments above, as this is a boolean state, I'd rather see EnableCullFace(true|false).

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
```diff
@@ -55,45 +40,30 @@ public StateChange getDefaultInstance() {
...
 
-        return task;
+        return disablingTask;
     }
 
     @Override
-    public boolean isEqualTo(StateChange stateChange) {
-        if (stateChange instanceof SetCullFace) {
-            return mode == ((SetCullFace) stateChange).getMode();
+    protected RenderPipelineTask getEnablingTask() {
```
emanuele3d: Previous comments regarding eliminating get[Enabling|Disabling]Task apply here too.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
```diff
@@ -55,45 +40,30 @@ public StateChange getDefaultInstance() {
...
-                break;
-            case GL11.GL_FRONT:
-                modeName = "GL_FRONT";
-                break;
-            case GL11.GL_FRONT_AND_BACK:
-                modeName = "GL_FRONT_AND_BACK";
-                break;
-        }
-        return modeName;
+        return this == defaultInstance;
```
emanuele3d: Can this method be moved into a parent class?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFaceMode.java
```diff
@@ -0,0 +1,99 @@
...
+import static org.lwjgl.opengl.GL11.GL_FRONT;
+import static org.lwjgl.opengl.GL11.GL_FRONT_AND_BACK;
+import org.terasology.rendering.dag.RenderPipelineTask;
+import org.terasology.rendering.dag.StateChange;
+import org.terasology.rendering.dag.tasks.SetCullFaceModeTask;
+
+/**
+ * TODO: Add javadocs
+ */
+public final class SetCullFaceMode implements StateChange {
```
emanuele3d: In a comment to a previous commit I recommended for this to be called `SetFacesToCull`.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFaceMode.java
```diff
@@ -0,0 +1,99 @@
...
+    private static SetCullFaceMode defaultInstance = new SetCullFaceMode(GL_BACK); // also specified in OpenGL documentation
+    private SetCullFaceModeTask task;
+    private int mode;
+
+    public SetCullFaceMode(int mode) {
+        this.mode = mode;
+        validate();
+    }
+
+    private void validate() {
```
emanuele3d: Feels a bit overkill to have a whole method for such a small check. I'd put everything in the constructor unless
1. the validation code becomes too complex
2. the validation code needs to be called in more than one place

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFaceMode.java
```diff
@@ -0,0 +1,99 @@
...
+        if (mode != GL_BACK
+                && mode != GL_FRONT
+                && mode != GL_FRONT_AND_BACK) {
+            throw new IllegalArgumentException("Mode must be GL_BACK, GL_FRONT or GL_FRONT_AND_BACK.");
+        }
+    }
+
+    public int getMode() {
+        return mode;
+    }
```
emanuele3d: Have you used these type of getters in previous StateChanges too? Somehow they standout. Do we really need them? I'd leave them out unless we have a specific, class-external use for them.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
```diff
@@ -56,12 +56,16 @@ public RenderPipelineTask generateTask() {
     @Override
     public boolean isEqualTo(StateChange stateChange) {
         if (stateChange instanceof SetWireframe) {
-            return this.enabled == ((SetWireframe) stateChange).enabled;
+            return this.enabled == ((SetWireframe) stateChange).isEnabled();
```
emanuele3d: A-ha. Now I see why we need the isEnabled() method.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFaceMode.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    public boolean isTheDefaultInstance() {
+        return defaultInstance == this;
+    }
+
+    public static String getModeName(int mode) {
```
emanuele3d: Does this needs to be public? It seems to be used only in the toString() method. If that's the case I would:

1. make this method private
2. remove the input parameter - read `this.mode` instead.
3. remove the switch block and use a static final int,String map instead.

In fact, you could probably remove this whole whole method and in toString write instead:

`return String.format("%21s(%s)", this.getClass().getSimpleName(), modeToStringMap[mode];`

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
emanuele3d: On second thought: can't you use that quirk of java that allows instances of the same class to access each other private members? Or did I perhaps explicitly suggest to not use it? I think I did, didn't I? :smile:

engine/src/main/java/org/terasology/rendering/dag/tasks/DisableCapabilityTask.java
```diff
@@ -21,7 +21,7 @@
 /**
  * TODO: Add javadocs
  */
-public class DisableCapabilityTask implements RenderPipelineTask {
+public final class DisableCapabilityTask implements RenderPipelineTask {
```
emanuele3d: I recommend to rename it to DisableStateParameterTask.

engine/src/main/java/org/terasology/rendering/dag/tasks/EnableCapabilityTask.java
```diff
@@ -21,7 +21,7 @@
 /**
  * TODO: Add javadocs
  */
-public class EnableCapabilityTask implements RenderPipelineTask {
+public final class EnableCapabilityTask implements RenderPipelineTask {
```
emanuele3d: In line with the comment above this should be renamed EnableStateParameterTask

engine/src/main/java/org/terasology/rendering/dag/tasks/SetCullFaceModeTask.java
```diff
@@ -17,15 +17,15 @@
...
 import org.lwjgl.opengl.GL11;
 import org.terasology.rendering.dag.RenderPipelineTask;
-import org.terasology.rendering.dag.stateChanges.SetCullFace;
+import org.terasology.rendering.dag.stateChanges.SetCullFaceMode;
 
 /**
  * TODO: Add javadocs
  */
-public class SetCullFaceTask implements RenderPipelineTask {
+public final class SetCullFaceModeTask implements RenderPipelineTask {
```
emanuele3d: In line with a comment above I recommend to rename this to `SetWhatFacesToCullTask`.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableMaterial.java
emanuele3d: But what was the reasoning behind the first final classes then? I didn't notice them before and I wonder why we should close the door to further subclassing. I can hardly imagine why anybody _would_ subclass these as they are pretty specialized already. But I guess one never knows?

I've done a quick search on the subject and some people advocate that a class should either be designed to be extensible OR declared final, mostly as a form of defensive programming. I don't think we use this rule in Terasology's codebase though. So I'm not sure this is necessary but it's ok to leave it like this for the time being. I'm just curious if there was a good reason for this.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
```diff
@@ -82,6 +82,7 @@ public void initialise() {
 
         addDesiredStateChange(new BindFBO(SCENE_SHADOW_MAP));
         addDesiredStateChange(new SetViewportSizeOf(SCENE_SHADOW_MAP));
+        addDesiredStateChange(new SetCullFace(false));
```
emanuele3d: Really, for readability this should be `addDesiredStateChange(new EnableFaceCulling());`

engine/src/main/java/org/terasology/rendering/dag/nodes/SimpleBlendMaterialsNode.java
```diff
@@ -44,7 +45,7 @@
 
     @Override
     public void initialise() {
-
+        addDesiredStateChange(new SetBlend(true));
```
emanuele3d: In line with previous comment, this should be: ` addDesiredStateChange(new EnableBlending());`

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCapability.java
```diff
@@ -63,4 +64,23 @@ public String toString() { // TODO: used for logging purposes at the moment, inv
...
+                break;
+            case GL11.GL_STENCIL_TEST:
+                capabilityName = "GL_STENCIL_TEST";
+                break;
+            case GL11.GL_CULL_FACE:
+                capabilityName = "GL_CULL_FACE";
+                break;
+        }
+        return capabilityName;
+    }
```
emanuele3d: In line with previous comment on the same method, this could be handled through a static final [int,String] map, making this method unnecessary.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetDepthTest.java
```diff
@@ -0,0 +1,70 @@
...
+import static org.lwjgl.opengl.GL11.GL_DEPTH_TEST;
+import org.terasology.rendering.dag.RenderPipelineTask;
+import org.terasology.rendering.dag.StateChange;
+import org.terasology.rendering.dag.tasks.DisableCapabilityTask;
+import org.terasology.rendering.dag.tasks.EnableCapabilityTask;
+
+/**
+ * TODO: Add javadocs
+ */
+public final class SetDepthTest extends SetCapability {
```
emanuele3d: Please rename to EnableDepthTest and create a parameterless constructor so that we can simply write:

`addDesiredStateChange(new EnableDepthTest());`

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetDepthTest.java
```diff
@@ -0,0 +1,70 @@
...
+        super(enabled);
+    }
+
+    @Override
+    public StateChange getDefaultInstance() {
+        return defaultInstance;
+    }
+
+    @Override
+    protected RenderPipelineTask getDisablingTask() {
```
emanuele3d: In line with comments in the previous commit, this and getEnablingTask() should be removed.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetStencilTest.java
```diff
@@ -0,0 +1,70 @@
...
+import static org.lwjgl.opengl.GL11.GL_STENCIL_TEST;
+import org.terasology.rendering.dag.RenderPipelineTask;
+import org.terasology.rendering.dag.StateChange;
+import org.terasology.rendering.dag.tasks.DisableCapabilityTask;
+import org.terasology.rendering.dag.tasks.EnableCapabilityTask;
+
+/**
+ * TODO: Add javadocs
+ */
+public final class SetStencilTest extends SetCapability {
```
emanuele3d: In line with previous comments this should be renamed to EnableStencilTest and have a parameterless constructor.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetStencilTest.java
```diff
@@ -0,0 +1,70 @@
...
+        super(enabled);
+    }
+
+    @Override
+    public StateChange getDefaultInstance() {
+        return defaultInstance;
+    }
+
+    @Override
+    protected RenderPipelineTask getDisablingTask() {
```
emanuele3d: getDisablingTask/getEnablingTask to be removed.

engine/src/main/java/org/terasology/rendering/dag/tasks/DisableCapabilityTask.java
```diff
@@ -32,4 +33,9 @@ public DisableCapabilityTask(int capability) {
     public void execute() {
         glDisable(this.capability);
     }
+
+    @Override
+    public String toString() {
+        return String.format("%21s: disabled", SetCapability.getCapabilityName(capability));
```
emanuele3d: I just focused for the first time on what is being done in this and other similar method. I'm concerned about this way of doing things because it's hardly extensible. But let's leave it as it is for now. We'll get back to it at a different stage.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFaceMode.java
tdgunes: I see, but I think it's much cleaner even though it's just several lines. Of course, if you are strongly against it, I'll move them into the constructor. It's a tendency, mostly because of [this course ](http://aktemur.github.io/cs222/resources.html) and [Uncle Bob's famous book](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882).

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
tdgunes: This was for particularly for `glEnable(GL_BLEND)`.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
tdgunes: This is not possible, because if Abstract class proposes static tasks, all other subclases of that class would have access to same static variables. They will not be duplicated. So, this was a necessity to eliminate redundancy in subclasses' `generateTask()` method, especially for having static `Enable` and `Disable` task for each StateChange that extends SetCapability.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCapability.java
tdgunes: Please check my previous comment above.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
tdgunes: Hmmm, I didn't want to lose the verbosity of `SetWireframe(true)`. If you think it's better, how about a new `final` StateChange class that extends SetCullFace?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
tdgunes: Please refer my previous comment above.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
tdgunes: Not possible, because of Java. Declaring a static variable in super will not generate copies of the same static variable inside every subclass.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
tdgunes: Hmmm, I really don't see the conflict. Can you explain it more, please? My rationale was, from OpenGL Documentation for `glEnable`:

screen shot 2016-08-09 at 22 30 56

For your second comment, I really don't see any good reason why we need someting like that with boolean parameter. But I am willing to create a subclasses for SetBlend:

* EnableBlend extends SetBlend(true)
* DisableBlend extends SetBlend(false)

which would be more better, I think.

tdgunes:
I've pushed my local changes, while awaiting replies to my comments. Perhaps maybe the latest commit makes this PR merge-worthy! 😄 🎊 🎉

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
emanuele3d: These are the capabilities I am referring to:

http://feedback.wildfiregames.com/report/opengl/

A prettier but more Apple-centric table is available here:

https://developer.apple.com/opengl/capabilities/

Finally, for use of the term "capabilities" in LWJGL and Terasology see:

`org.terasology.engine.subsystems.lwjgl.LwjglGraphics.checkOpenGL()`

In line with comments made later I recommend EnableBlend() instead of EnableBlend(true) and then using EnableBlend(false) as the default instance, which should be used only internally rather than in nodes. Alternatively you could use DisableBlend() for the default instance but I'm not sure if this would make things unnecessarily complicated when the checks are made to verify redundancy. 

I don't see how a SetBlend class used as a base would be beneficial. But I do see the benefit of using the term Enable/Disable for boolean parameters and Set for non-boolean parameters.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
emanuele3d: Perhaps I'm not understanding or perhaps I wasn't clear enough:

why is the default instance of SetBlend an instance of SetCullFace?

tdgunes: Oh, sorry. This is a horrible copy-paste error. Let me fix it.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlend.java
emanuele3d: I see what you mean:  SetCapability.generateTask() cannot have access to the static tasks variables declared in its subclasses unless it uses overridable methods. Ok fair enough. What about instantiating the Disable/Enable tasks in the constructor so that these methods become just accessors, without task instantiation and null checking?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCapability.java
emanuele3d: I confirm we can keep the methods. What about the `if(enabled)` instead of `if(isEnabled())` ?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
emanuele3d: Think about the node developer. He needs to work on the desired state changes. He has opengl knowledge and knows he needs to enable face culling. What's going to be most intuitive for him to use: EnableFaceCulling() and SetWhatFacesToCull(mode) or SetCullFace(boolean) and SetCullFaceMode(int)?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
emanuele3d: Agreed on slack to keep the methods but move null check and task instantiation in the constructor.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFace.java
emanuele3d: Ok, let's keep it as it is!

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCullFaceMode.java
emanuele3d: Nah. The intent of the code it's clear: constructor input validation. Because it's so simple it can stay in the constructor without a method wrapping it. I have not read those to books, in my library I have Code Complete for those kind of issues. My understanding of it is that we need to strike a balance: too large and complex methods are hard to read, but segmenting the code into very small methods is just as bad. So, I'd still remove validate() as this is a very trivial validation.

engine/src/main/java/org/terasology/rendering/dag/tasks/SetCullFaceModeTask.java
emanuele3d: Wow. That's verbose even for my standards... =) SetFacesToCullTask would be fine too. =)

engine/src/main/java/org/terasology/rendering/dag/tasks/SetWhatFacesToCullTask.java
```diff
@@ -17,15 +17,15 @@
...
 import org.lwjgl.opengl.GL11;
 import org.terasology.rendering.dag.RenderPipelineTask;
-import org.terasology.rendering.dag.stateChanges.SetCullFaceMode;
+import org.terasology.rendering.dag.stateChanges.SetFacesToCull;
 
 /**
  * TODO: Add javadocs
  */
-public final class SetCullFaceModeTask implements RenderPipelineTask {
+public final class SetWhatFacesToCullTask implements RenderPipelineTask {
```
emanuele3d: Really, this should be consistent with the associated state change. So, SetFacesToCull should be associated with SetFacesToCullTask. Sorry for suggesting this inconsistency, my fault.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlending.java
```diff
@@ -24,13 +24,13 @@
 /**
  * TODO: Add javadocs
  */
-public final class SetBlend extends SetCapability {
+public class SetBlending extends SetCapability {
```
emanuele3d: Still recommending to rename SetCapability to SetStateParameter or perhaps an even more specific SetBooleanStateParameter.

Cervator:
Checked this out for testing, but the conflicts are a bit beyond my quick merging skills :-) They don't look major but probably better to have one of you do the updates to make sure all is well. Affected files:

* `ShadowMapNode`
* `SimpleBlendMaterialsNode`
* `WorldReflectionNode`
* `BindFBO`
* `SetViewportSizeOf`

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFacesToCull.java
```diff
@@ -15,34 +15,34 @@
...
 /**
  * TODO: Add javadocs
  */
 public final class SetFacesToCull implements StateChange {
-
     private static SetFacesToCull defaultInstance = new SetFacesToCull(GL_BACK); // also specified in OpenGL documentation
-    private SetWhatFacesToCullTask task;
+    private static Map<Integer, String> modeNameMap = ImmutableMap.of(GL11.GL_BACK, "GL_BACK",
+            GL11.GL_FRONT, "GL_FRONT",
+            GL11.GL_FRONT_AND_BACK, "GL_FRONT_AND_BACK");
```
emanuele3d: In the constructor you use GL_BACK and the other constants without the GL11 qualifier. I recommend to remove it from here too, if possible. Or, if you prefer, to use GL11 also below.

engine/src/main/java/org/terasology/rendering/dag/tasks/DisableStateParameterTask.java
```diff
@@ -17,15 +17,15 @@
...
 
 /**
  * TODO: Add javadocs
  */
-public final class DisableCapabilityTask implements RenderPipelineTask {
+public final class DisableStateParameterTask implements RenderPipelineTask {
     private int capability;
 
-    public DisableCapabilityTask(int capability) {
+    public DisableStateParameterTask(int capability) {
```
emanuele3d: I'd rename `capability` to `parameter`.

engine/src/main/java/org/terasology/rendering/dag/tasks/EnableStateParameterTask.java
```diff
@@ -17,15 +17,15 @@
...
 
 /**
  * TODO: Add javadocs
  */
-public final class EnableCapabilityTask implements RenderPipelineTask {
+public final class EnableStateParameterTask implements RenderPipelineTask {
     private int capability;
 
-    public EnableCapabilityTask(int capability) {
+    public EnableStateParameterTask(int capability) {
```
emanuele3d: capability -> parameter

engine/src/main/java/org/terasology/rendering/dag/tasks/SetFacesToCullTask.java
```diff
@@ -22,10 +22,10 @@
 /**
  * TODO: Add javadocs
  */
-public final class SetWhatFacesToCullTask implements RenderPipelineTask {
+public final class SetFacesToCullTask implements RenderPipelineTask {
```
emanuele3d: Let's add:

// TODO: review the possibility of making task classes private, contained in the StateChange implementation they are associated with. Rationale: module developers shouldn't be able to reach or instantiate these classes directly: only the StateChange classes.

What do you think?

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -120,9 +119,9 @@ private void generateIntermediateList(List<Node> orderedNodes) {
             intermediateList.add(node);
             // Add state changes to reset all desired state changes back to default.
             for (StateChange stateChange : node.getDesiredStateChanges()) {
-                intermediateList.add(stateChange.getDefaultInstance());
+                StateChange defaultInstance = stateChange.getDefaultInstance();
+                intermediateList.add(defaultInstance);
```
emanuele3d: Why this change? Did I ask for it?

engine/src/main/java/org/terasology/rendering/dag/nodes/ToneMappingNode.java
```diff
@@ -62,20 +48,10 @@ public void initialise() {
...
-        sceneToneMapped = frameBuffersManager.getFBO("sceneToneMapped");
-        sceneOpaque = frameBuffersManager.getFBO("sceneOpaque");
-
-        toneMapping.enable();
-
-        sceneToneMapped.bind();
-        setViewportToSizeOf(sceneToneMapped);
-        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);     // TODO: verify this is necessary
 
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
```
emanuele3d: re-instate TODO please!

engine/src/main/java/org/terasology/rendering/dag/nodes/WorldReflectionNode.java
```diff
@@ -69,17 +66,16 @@ public void initialise() {
         this.renderingConfig = config.getRendering();
         this.chunkShader = worldRenderer.getMaterial("engine:prog.chunk");
         this.playerCamera = worldRenderer.getActiveCamera();
-        addDesiredStateChange(new BindFBO(SCENE_REFLECTED_FBO, frameBuffersManager));
-        addDesiredStateChange(new SetViewportSizeOf(SCENE_REFLECTED_FBO, frameBuffersManager));
-
+        addDesiredStateChange(new BindFBO(SCENE_REFLECTED_FBO));
+        addDesiredStateChange(new SetViewportSizeOf(SCENE_REFLECTED_FBO));
+        addDesiredStateChange(new SetFacesToCull(GL_FRONT));
```
emanuele3d: Careful. This imply face culling must be enabled. We need an additional state change in this list.

engine/src/main/java/org/terasology/rendering/dag/nodes/ToneMappingNode.java
```diff
@@ -15,41 +15,27 @@
...
-    private Config config;
-
-    private Material toneMapping;
-    private FBO sceneToneMapped;
-    private FBO sceneOpaque;
 
     @Override
     public void initialise() {
-        toneMapping = worldRenderer.getMaterial("engine:prog.hdr"); // TODO: rename shader to toneMapping)
+        addDesiredStateChange(new EnableMaterial("engine:prog.hdr"));
```
emanuele3d: please reintroduce TODO.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetStateParameter.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * TODO: Add javadocs
+ * Indented for capabilities that are enabled/disabled via glEnable and glDisable.
+ */
+public abstract class SetStateParameter implements StateChange {
+    private static Map<Integer, String> paramaterMap = ImmutableMap.of(GL11.GL_BLEND, "GL_BLEND",
+            GL11.GL_DEPTH_TEST, "GL_DEPTH_TEST",
+            GL11.GL_STENCIL_TEST, "GL_STENCIL_TEST",
+            GL11.GL_CULL_FACE, "GL_CULL_FACE");
```
emanuele3d: I've been thinking. What are we gaining in having this map here? It's a lot of "knowledge" aggregated into a class. Ideally that knowledge is placed close to where it's needed. And it's needed for the log output of tasks. 

So, how about:

`disablingTask = new DisableStateParameterTask(PARAMETER);`

becomes

`disablingTask = new DisableStateParameterTask(int parameter, String parameterName);`

This way we do not need a central repository for this information and individual state changes can provide the appropriate name to the disabling/enabling task.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -86,6 +84,9 @@ public void update() {
 
     @Override
     public String toString() { // TODO: used for logging purposes at the moment, investigate different methods
+        if (isTheDefaultInstance()) { // necessary since default instance is generated before setFrameBuffersManager
```
emanuele3d: Let's change this to:

`if (this.isTheDefaultInstance()) { // necessary since default instance is generated before setFrameBuffersManager`

as it reads like plain English.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -27,35 +27,33 @@
...
-    private FrameBuffersManager frameBuffersManager;
 
-    public SetViewportSizeOf(String fboName, FrameBuffersManager frameBuffersManager) {
+    public SetViewportSizeOf(String fboName) {
         this.fboName = fboName;
-        this.frameBuffersManager = frameBuffersManager;
-        fbo = frameBuffersManager.getFBO(fboName);
+        if (!fboName.equals(DEFAULT_FBO)) {
+            fbo = frameBuffersManager.getFBO(fboName);
+        }
```
emanuele3d: Didn't we discuss eliminating this if block, with everything happening only in the generateTask() method instead?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -86,6 +84,9 @@ public void update() {
 
     @Override
     public String toString() { // TODO: used for logging purposes at the moment, investigate different methods
+        if (isTheDefaultInstance()) { // necessary since default instance is generated before setFrameBuffersManager
+            fbo = frameBuffersManager.getFBO(fboName);
+        }
         return String.format("%21s: %s(%sx%s)", this.getClass().getSimpleName(), fboName, fbo.width(), fbo.height());
```
emanuele3d: Didn't we discuss eliminating the fbo.width()/fbo.height() output from the state change log, keeping it only in the task's log instead?

engine/src/main/java/org/terasology/rendering/dag/tasks/DisableStateParameterTask.java
```diff
@@ -0,0 +1,41 @@
...
+import org.terasology.rendering.dag.RenderPipelineTask;
+import org.terasology.rendering.dag.stateChanges.SetStateParameter;
+
+/**
+ * TODO: Add javadocs
+ */
+public final class DisableStateParameterTask implements RenderPipelineTask {
+    private int capability;
+
+    public DisableStateParameterTask(int capability) {
```
emanuele3d: In line with a previous comment, the constructor should get an int and a string as input, the string being the name of the parameter being disabled. The string is then used for logging purposes instead of the map lookup.

engine/src/main/java/org/terasology/rendering/dag/tasks/EnableStateParameterTask.java
```diff
@@ -0,0 +1,41 @@
...
+import org.terasology.rendering.dag.RenderPipelineTask;
+import org.terasology.rendering.dag.stateChanges.SetStateParameter;
+
+/**
+ * TODO: Add javadocs
+ */
+public final class EnableStateParameterTask implements RenderPipelineTask {
+    private int capability;
+
+    public EnableStateParameterTask(int capability) {
```
emanuele3d: Same comments as for the DisableStateParameterTask's constructor and toString() methods.

emanuele3d:
Found a number of very small issues plus a couple of slightly larger ones. I think we are very close to proceeding to the next phase though, where we start using the state changes in all nodes. 

Please check the comments added to the last commit AND the comments added to the whole set of changes.

emanuele3d:
Also, before we get to rolling all the state changes into all the nodes, can you resolve the conflicts?

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -90,7 +90,7 @@ private void generateTaskList() {
                             taskList.add(stateChange.generateTask());
                             persistentStateChanges.remove(stateChange.getClass());
 
-                        } else if (!stateChange.isEqualTo(persistentStateChange)) { // another new method, just for readability
+                        } else if (!stateChange.equals(persistentStateChange)) { // another new method, just for readability
```
emanuele3d: We don't need the comment anymore here! 

And good job on implementing the equals() method!

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableBlending.java
```diff
@@ -26,6 +26,7 @@
  */
 public class EnableBlending extends SetStateParameter {
     private static final int PARAMETER = GL_BLEND;
+    private static final String PARAMETER_NAME = "GL_BLEND";
```
emanuele3d: I wonder what the utility of these fields are versus simply having:

`disablingTask = new DisableStateParameterTask(GL_BLEND, "GL_BLEND");`

but I'm ok with leaving them as they are to get through this stage of the PR and progress to the next stage.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFacesToCull.java
```diff
@@ -15,31 +15,29 @@
...
+            GL_FRONT, "GL_FRONT",
+            GL_FRONT_AND_BACK, "GL_FRONT_AND_BACK");
     private SetFacesToCullTask task;
     private int mode;
 
     public SetFacesToCull(int mode) {
-        if (mode != GL_BACK
-                && mode != GL_FRONT
-                && mode != GL_FRONT_AND_BACK) {
+        if (!modeNameMap.containsKey(mode)) {
```
emanuele3d: The most likely case of this if statement is the opposite I think.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFacesToCull.java
```diff
@@ -37,7 +37,9 @@
     private int mode;
 
     public SetFacesToCull(int mode) {
-        if (!modeNameMap.containsKey(mode)) {
+        if (mode != GL_BACK
```
emanuele3d: What happened here? Why did you revert to individual checks instead of checking against the map? 

In any case you should reverse the handling of this IF block. The current true case is the least likely to happen.

emanuele3d:
The latest changes are good. A couple of very minor issues but those can wait until the next commit. In a normal PR I'd say this is ready for merge, but we have decided to go ahead and change all nodes so that they use the StateChange implementations. So, let's move to that phase.

Good job @tdgunes!!

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFacesToCull.java
tdgunes: I was getting NPE, because of uninitialised static variable during initialisation. I didn't have the time to properly check why this tiny change resulted as NPE, so I reverted it back. For if-block, let me correct that.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFacesToCull.java
```diff
@@ -37,12 +37,11 @@
     private int mode;
 
     public SetFacesToCull(int mode) {
-        if (mode != GL_BACK
-                && mode != GL_FRONT
-                && mode != GL_FRONT_AND_BACK) {
+        if (mode == GL_BACK || mode == GL_FRONT || mode == GL_FRONT_AND_BACK) {
```
emanuele3d: :+1:

GooeyHub:

Cervator:
FYI: Took this for a quick spin and couldn't find any obvious issues in-game or in the log :-) Happy to merge whenever appropriate!

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Tested it a bit locally - nothing obvious in the overworld I could spot :-)

One interesting thing I did find is if I clip through the terrain with ghost the "underworld" looks different than it used to. However, I think it kinda looks way cooler this way!

These screenies are all with the "Uber" graphic preset. Interesting difference between holding dynamite vs the torch - almost like the torch attempts to light up the back of block faces "visible" when clipping? 

terasology-160820140039-1152x720

terasology-160820140041-1152x720

Old school screenie without this PR just for reference:

terasology-160820140510-1152x720

Again just to be sure: I personally *like* this new look vs. the old look. Especially the un-lit version where you can see distant "terrain" :D Looks way less broken than the old approach

engine/src/main/java/org/terasology/rendering/dag/tasks/SetFacesToCullTask.java
emanuele3d: Doh. Of course they can't be private otherwise they can't get into the tasklist.

emanuele3d:
Starting review of the latest commits. The new rendering style of the underworld is interesting. I'd cautiously take it as a good sign that we might be doing something good. :smile:

engine/src/main/java/org/terasology/rendering/dag/nodes/SimpleBlendMaterialsNode.java
```diff
@@ -74,23 +76,11 @@ public void process() {
      */
     private void preRenderSetupSimpleBlendMaterials() {
         READ_ONLY_GBUFFER.setRenderBufferMask(true, true, true);
-        GL11.glEnable(GL_BLEND);
         GL11.glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA); // (*)
```
emanuele3d: I'd recommend to remove this method and place its content in process. These preRenderSetup and postRenderCleanup methods made sense when they were full with tricky opengl stuff. Now that they are becoming basically empty they can be removed.

emanuele3d: By the way, this line should become a StateChange. If you'd like to leave it to after the end of the project, please add a TODO.

emanuele3d: Finally, make sure to move the javadoc for this method somewhere else rather than delete it, as it has useful information.

engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
```diff
@@ -65,13 +66,8 @@ public void initialise() {
     @Override
     public void process() {
         PerformanceMonitor.startActivity("rendering/prePostComposite");
-        prePostComposite.enable();
         sceneReflectiveRefractive = displayResolutionDependentFBOs.get(REFLECTIVE_REFRACTIVE);
```
emanuele3d: Can we move this line to just above READ_ONLY_GBUFFER.attachDepthBufferTo?

engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
```diff
@@ -65,13 +66,8 @@ public void initialise() {
...
 
-        // TODO: verify if there should be bound textures here.
-        WRITE_ONLY_GBUFFER.bind();
-
-        setViewportToSizeOf(WRITE_ONLY_GBUFFER);
         glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT); // TODO: verify this is necessary
 
         renderFullscreenQuad();
```
emanuele3d: I think the bindDisplay() and setViewportToSizeOf() lines can be safely eliminated. Just make sure to test this specific change on its own! =)

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -24,30 +24,27 @@
 import org.terasology.rendering.dag.StateChange;
 import org.terasology.rendering.dag.tasks.SetViewportSizeOfTask;
 import org.terasology.rendering.opengl.FBO;
-import org.terasology.rendering.opengl.fbms.DisplayResolutionDependentFBOs;
-
+import static org.terasology.rendering.opengl.DefaultDynamicFBOs.READ_ONLY_GBUFFER;
 /**
  * TODO: Add javadocs
  */
 public final class SetViewportSizeOf implements FBOManagerSubscriber, StateChange {
```
emanuele3d: To be close to plain English this should really be `SetViewportToSizeOf`. Notice the "To" in the middle. Right now it's a bit grammatically incorrect...

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportSizeOf.java
```diff
@@ -89,8 +83,12 @@ public boolean isTheDefaultInstance() {
...
     @Override
     public void update() {
-        FBO fbo = frameBuffersManager.get(fboName);
-        task.setDimensions(fbo.width(), fbo.height());
+        if (defaultDynamicFBO == null) {
+            FBO fbo = frameBuffersManager.get(fboName);
+            task.setDimensions(fbo.width(), fbo.height());
+        } else {
+            task.setDimensions(defaultDynamicFBO.width(), defaultDynamicFBO.height());
+        }
```
emanuele3d: Hmmm. I'm really not liking how the default FBOs need to have special handling. For me they should be pretty normal FBOs that nodes can rely on are available right from the start. Them, _ideally_, they should be easily reachable, i.e. through static variables. Can you write a TODO in the default FBOs class on this regard?

engine/src/main/java/org/terasology/rendering/dag/nodes/OutlineNode.java
```diff
@@ -75,23 +75,13 @@ public void initialise() {
...
-
-            // TODO: verify inputs: shouldn't there be a texture binding here?
-            outlineFBO.bind();
+        PerformanceMonitor.startActivity("rendering/outline");
 
-            setViewportToSizeOf(outlineFBO);
-            glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT); // TODO: verify this is necessary
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT); // TODO: verify this is necessary
+        renderFullscreenQuad();
+        bindDisplay();  // TODO: verify this is necessary
```
emanuele3d: I'm pretty sure we can remove this now. If necessary the default state change task will do it. Just verify enabling/disabling the outline still works ok.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -114,8 +114,8 @@ private void generateTaskList() {
     private void generateIntermediateList(List<Node> orderedNodes) {
         intermediateList.clear();
         for (Node node : orderedNodes) {
+            node.setTaskListGenerator(this);
```
emanuele3d: Hmmm... I don't think nodes should have access to the task list. StateChanges are the interface between nodes and the task list. Why are we doing this?

engine/src/main/java/org/terasology/rendering/dag/nodes/LightGeometryNode.java
```diff
@@ -99,43 +98,16 @@ public void process() {
...
-    private void postRenderCleanupLightGeometry() {
-        glDisable(GL_STENCIL_TEST);
-        glCullFace(GL_BACK);
 
-        bindDisplay();
+        PerformanceMonitor.endActivity();
     }
 
     // TODO: figure how lighting works and what this does
     private void preRenderSetupLightGeometry() {
```
emanuele3d: Let's place the much more simplified content of this method into process(). And of course ideally the glStencil|BlendFunc calls would become StateChanges. If you'd rather do it after the GSOC deadline you can just add a TODO.

engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
```diff
@@ -0,0 +1,48 @@
...
+import com.google.api.client.util.Lists;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+import java.util.List;
+import java.util.function.Supplier;
+
+/**
+ * TODO: Add javadocs
+ */
+public abstract class ConditionDependentNode extends AbstractNode implements PropertyChangeListener {
```
emanuele3d: This is a tricky class I find. I just understood that it is written this way to allow for multiple conditions. I'm not sure if we have a use case for that already but... ok. I guess the one problem I have is with the term "Condition". It's quite generic while the class really only deals with boolean values. Not only that: all conditions have to be true. So, a node that, say, needs to be inserted into the graph if the condition is false (for whatever reason) can't use this.

So, it seems to me that the usability of this class is quite narrow. I'd remove it in favor of nodes subscribing to the rendering config on initialization and implementing the PropertyChangeListener interface, to leave maximum freedom to the node developers.

engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
```diff
@@ -0,0 +1,48 @@
...
+        for (Supplier<Boolean> condition : conditions) {
+            areSatisfied = areSatisfied && condition.get();
+        }
+        setEnabled(areSatisfied);
+    }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent evt) {
+        checkConditions();
+        refreshTaskList();
```
emanuele3d: I now see why the nodes currently need the taskListGenerator. 

I really don't think nodes should have a reference to it. If we wanted nodes to have access to the task list we could have skipped the whole StateChanges layer and given them the freedom to insert/remove tasks into the list. Which strikes me as potentially very chaotic!!

But I'd be ok if they can call a method to refresh the task list, say, on WorldRenderer or another object known to all nodes. Ideally though, the wouldn't trigger the refresh directly but put it in some kind of "pending" mode, so that if N nodes ask for the list to be refreshed it gets refreshed only once.

emanuele3d: About the "pending" part, I actually don't know if that would be easy to implement and we might want to leave it aside for the time being. Perhaps a TODO somewhere?

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -131,5 +132,6 @@ public void swapReadWriteBuffers() {
         WRITE_ONLY_GBUFFER.setFbo(fbo);
         fboLookup.put(READ_ONLY_GBUFFER.getName(), READ_ONLY_GBUFFER.getFbo());
         fboLookup.put(WRITE_ONLY_GBUFFER.getName(), WRITE_ONLY_GBUFFER.getFbo());
+        notifySubscribers();
```
emanuele3d: Looking at this line I noticed that in update() subscribers are notified in createFBOs and _before_ updateDefaultFBOs() is called. If I am correct you might want to move the notifySubscribers() line from createFBOs to the end of the IF block in update(), just after the line updateDefaultFBOs().

emanuele3d:
@tdgunes: I finished the review. Looks good! Relatively few issues to deal with, maybe a couple of issues a little more involved but nothing horribly wrong. And I like how the Node.process() methods are now very, very lean in most cases.

engine/src/main/java/org/terasology/rendering/dag/nodes/OutlineNode.java
tdgunes: This is already removed in latests commits, I think.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
tdgunes: This was already there, especially needed for WireframeCapableNode.

engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
tdgunes: Module developers can already do what you are saying. Freedom is preserved! 😄 This class actually introduces an helper method `requiresCondition` which accepts a closure(lambda function). This function can be anything, a very good use case for this one is in `BlurPassesNode`.

```java
    private FBO sceneToneMapped;

    @Override
    public void initialise() {
        renderingConfig = config.getRendering();
        blur = worldRenderer.getMaterial("engine:prog.blur");

        renderingConfig.subscribe(RenderingConfig.BLUR_INTENSITY, this);
        requiresCondition(() -> renderingConfig.getBlurIntensity() != 0);
        requiresFBO(new FBOConfig(BLUR_0, HALF_SCALE, FBO.Type.DEFAULT), displayResolutionDependentFBOs);
        requiresFBO(new FBOConfig(BLUR_1, HALF_SCALE, FBO.Type.DEFAULT), displayResolutionDependentFBOs);
        requiresFBO(new FBOConfig(TONE_MAPPED, FULL_SCALE, FBO.Type.HDR), displayResolutionDependentFBOs);
    }
```

I was between the term `ConditionDependentNode` and `ConfigDependentNode`. `Condition` seemed more generic, as you mentioned. Another benefit of having this class is reducing huge amount of redundant code. (Same lines in many different nodes.) Therefore, for the time being, I would not remove this class. In long term goals, I would investigate better ways to connect current configurations with nodes.

engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
tdgunes: I think TODO is better for the time being. Let me add that.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
tdgunes: Yes, true. But since DefaultFBOs internal FBO reference is up to date just before task list is executed, this didn't cause any harm. On the other hand, there can be some unexpected bugs that are not currently found because of this. So I think it's better to move `notifySubscribers` after `updateDefaultFBOs()`.

GooeyHub:

GooeyHub:

emanuele3d:
Declaring this PR officially **GOOD TO MERGE**!!! Yieppieeee!!

GooeyHub:

Cervator:
Tested again and no obvious issues when just varying shadows and such under normal conditions, water reflections on global still look awesome *but* - I think there may be an issue (at least for me) with lighting, specifically when the player equips a luminous object like a torch.

[Last time I tested](https://github.com/MovingBlocks/Terasology/pull/2427#issuecomment-241214903) the "underside" view of the world changed dramatically when I equipped a torch which I wrote off as being an oddity from being under the world, but I just realized it happens when just walking around on the right side of the world and switching to a torch. The sky goes black, water disappears, and some weird skysphere-related pattern becomes visible.

While I have nothing to base this on I suspect this is probably something simple for a 3d wizard (similar effects have happened before). It appears active across all settings I tried, from minimal to ultra, before or after changing settings in-game. But it would be good to know if anybody else sees it :-)

Black sky:

blacksky

No Water:

nowater

Weird skysphere pattern or something:

skyspherepattern

Double checked that it seems to be caused by this PR, at least when merged on top of latest develop.

GooeyHub:

Cervator:
Merged! Discussed briefly with @tdgunes on Slack and the one problematic Node was reverted for the time being, seems there's something quirky about the original. No surprise there - it'll take time to unravel all the spiderwebs that built up over the years :-)

Thanks!

PR #2434, opened 2016-08-07
Added categories to leaf block template to make it breakable

xrtariq2594:
### Contains

Changed the leaf block template to add two default categories ("wood" and "leaf") to it. Also altered the shovel damage prefabs so that shovels can now damage leaves. Thus, leaves can now be broken using fists, axes, or shovels.

### How to test

Create a new world, and try destroying the leaves off of trees using unarmed combat, axes, and shovels.

### Credits

@Cervator for finding this bug in the first place.

### DIFF
--- a/modules/Core/assets/blocks/templates/leaf.block
+++ b/modules/Core/assets/blocks/templates/leaf.block
@@ -3,6 +3,7 @@
  */
 {
     "template" : true,
+    "categories" : ["wood", "leaf"],
     "attachmentAllowed" : false,
     "hardness" : 1,
     "colorSource" : "foliage_lut",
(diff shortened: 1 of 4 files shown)


GooeyHub:

Cervator:
Merged after some testing of JS and the sort of "neo TTA" WorkstationCrafting (not in the lineup yet). Base game (Core) is unaffected, JS seems fine, and WC needs more work in any case but this way there's a easier foundation to build on.

skaldarnar:
Are block categories just strings or do they correspond to some component? Would be good if the categories in use could be noted down in the wiki (if not done yet).

Cervator:
Just strings.  The support for categories is there in the engine but it is barely utilized, typical story :-)

Probably would be ideal if we did some thinking about conventions to apply to the feature and maybe throw some examples or a small tutorial at it. Will make a separate issue for it.

Wiki doc would go out of date unless automated - but if the system becomes well used with good docs on how to apply it we could probably do something automagic.

PR #2437, opened 2016-08-13
Override config

oniatus:
### Contains

See [this post](http://forum.terasology.org/threads/multiplayer-server-management.1583/#post-13835) for a detailed description.

In short, it adds a run parameter `-overrideDefaultConfig=` to the PC facade.
The override config goes between the default config and the user config and gives the option to predefine settings for a headless server.

### How to test

Create a file `override.cfg` with the following content:
```
{
  "defaultModSelection": {
    "modules": [
      "MyModule"
    ],
    "defaultGameplayModuleName": "MyModule"
  },
  "worldGeneration": {
    "worldTitle": "World Title",
    "defaultSeed": "custom seed for the server",
    "defaultGenerator": "MyModule:MyWorldgen"
  },
  "moduleConfigs": {},
  "network": {
    "servers": [
      {
        "name": "localhost",
        "address": "localhost",
        "port": 25777,
        "active": true
      }
    ],
    "upstreamBandwidth": 1024,
    "serverPort": 25777,
    "masterServer": "meta.terasology.org"
  }
}
```

Start the pc facade with `-overrideDefaultConfig=override.cfg -headless`. Verfiy that the generated config.cfg contains the modifications from the override.cfg.
Further changes to the config.cfg should not be overriden by the override.cfg (user changes have always priority) 

Secondly, the change wires the splashEnabled and soundEnabled to the headless option.
Disabling the splash when already passed the headless option is very unintuitive. Not sure if the sound was active (mybe implicit) but now the flag is set properly.

Maybe checkstyle? :smile:  Good old Eclipse formatting.

### DIFF
--- a/facades/PC/src/main/java/org/terasology/engine/Terasology.java
+++ b/facades/PC/src/main/java/org/terasology/engine/Terasology.java
@@ -347,9 +347,7 @@ public final class Terasology {
             System.out.println((recognized ? "Recognized" : "Invalid") + " argument: " + arg);
         }
 
-        try
-
-        {
+        try {
             if (homePath != null) {
                 PathManager.getInstance().useOverrideHomePath(homePath);
             } else {
--- a/facades/PC/src/main/java/org/terasology/engine/Terasology.java
+++ b/facades/PC/src/main/java/org/terasology/engine/Terasology.java
@@ -17,6 +17,7 @@ package org.terasology.engine;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
+import org.terasology.config.Config;
 import org.terasology.config.SystemConfig;
 import org.terasology.crashreporter.CrashReporter;
 import org.terasology.engine.modes.StateLoading;
@@ -101,6 +102,7 @@ public final class Terasology {
     private static final String NO_SOUND = "-noSound";
     private static final String NO_SPLASH = "-noSplash";
     private static final String SERVER_PORT = "-serverPort=";
+    private static final String OVERRIDE_DEFAULT_CONFIG = "-overrideDefaultConfig=";
 
     private static boolean isHeadless;
     private static boolean crashReportEnabled = true;
@@ -170,7 +172,7 @@ public final class Terasology {
 
         SplashScreenBuilder builder = new SplashScreenBuilder();
 
-        String[] imgFiles = new String[] {
+        String[] imgFiles = new String[]{
                 "splash_1.png",
                 "splash_2.png",
                 "splash_3.png",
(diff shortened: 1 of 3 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

@oniatus: thanks! :-)

GooeyHub:

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -332,14 +338,18 @@ private static void handleLaunchArguments(String[] args) {
...
                 recognized = false;
             }
 
             System.out.println((recognized ? "Recognized" : "Invalid") + " argument: " + arg);
         }
 
-        try {
+        try
+
+        {
```
Cervator: Dangling bracket still sticking out :-)

Cervator:
Alright, tested it out some. Comments :)

* Should it be made more obvious how the override flag interacts with -homedir? Namely (as far as I can tell) the -homedir is ignored as far as the override is concerned, which makes sense, but isn't pointed out anywhere
* Should the `override.cfg` be the default name for the file, respected even without the flag active? That might make it easier to use, and we could add that file to the `.gitignore` for completeness for developers who might use it. Then the flag could just be if you wanted to vary the filename for some reason
* Could you update https://github.com/MovingBlocks/Terasology/blob/develop/docs/Playing.md#multiplayer or something with a note about the flag? Or maybe that should be in the wiki, unsure. We might be lacking docs there / actually have too much in the `Playing.md` for the basics
* Should the extra couple tweaks disabling the splash screen and sound be included? They make sense on the surface but I don't know if the sound disabling in particular could cause trouble? I tried running with the soundtrack module enabled and that was fine for a connected client, so maybe I'm just paranoid

Cervator:
@oniatus - bump! :D

GooeyHub:

oniatus:
Busy week :smile: 

> Should it be made more obvious how the override flag interacts with -homedir? Namely (as far as I can tell) the -homedir is ignored as far as the override is concerned, which makes sense, but isn't pointed out anywhere

The override should not interact with the homedir. Am I missing something? I was always using a custom homedir for testing, never the default one.

> Should the override.cfg be the default name for the file, respected even without the flag active?
For me it would be okay to specify the path as i keep the file separate from the server.

I also updated the documentation. Maybe it is too long now but at least documented :)
I will revert the sound change in the next commit, just to be sure it does not break anything (hard to test).
Removing the splashscreen is okay IMHO as headless means "no ui actions" for me and using something like "no ui actions" and "and no specific ui action" seems to be very counterintuitive :wink:

GooeyHub:

Cervator:
Thanks :-)

Yeah the homedir shouldn't interact, I was more wondering if somebody could end up thinking it might (without explicitly pointing it out). But I may lean toward over-documenting specifics, bloated docs can be bad :-)

Thanks for adding the usage in /docs, even if it is a bit long it is easier to move/tweak existing doc later than adding it from scratch, especially since we'll often forget about something after a PR is merged.

Cervator:
Tested again real quick and merged! :-)

PR #2438, opened 2016-08-13
NUI editor: additional settings and finishing touches

rzats:
### Contains

Various improvements to the NUI screen/skin editors; see rzats/gsoc-nui-editor#10, rzats/gsoc-nui-editor#13 and rzats/gsoc-nui-editor#14.

- Added a 'Preview screen locale' setting - this will change the language of the editor's preview widget (but not the NUI editor or the game itself). Useful for reviewing things such as long UI strings that may break interface rendering!
- Added a "Close" button to the editors (as an alternative to ESC & F10/F11) and a "Delete" option to tree view node context menues (as an alternative to DELETE).

### How to test

Press F10/F11 in the main menu to run the NUI screen/skin editors, then verify that all of the above work as intended :smile:

### Outstanding before merging

- [ ] Implement the features from rzats/gsoc-nui-editor#13 (tree view to widget binding).
- [x] Review the source code's Javadoc, modify and fill stubs where appropriate.
  - [x] Also fix Checkstyle warnings where appropriate.
- [x] Autosave the asset changes to a temporary JSON file, then add an option to apply the changes to the actual asset file.
- [x] Add a popup to warn the user that changes will be lost on asset change.
- [x] Automatically edit newly added attributes.
  - [x] Automatically edit the IDs of newly added widgets.
- [x] A "Debug Tools" menu with options to show different dev tools (console, UI editor, skin editor, BT editor).
- [x] When the editor's strings are somewhat finalized, internationalize them all!

PR #2448, opened 2016-08-21
NUI editor autosave

rzats:
### Contains

Implements autosaving for the NUI editor, as per https://github.com/MovingBlocks/Terasology/pull/2438#issuecomment-241214770.

### How to test

Open the NUI (skin/screen) editor, edit, Alt+F4 without saving changes, restart the game and open the same editor again.

No outstanding issues.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/editor/layers/AbstractEditorScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/editor/layers/AbstractEditorScreen.java
@@ -25,6 +25,7 @@ import org.slf4j.LoggerFactory;
 import org.terasology.assets.ResourceUrn;
 import org.terasology.input.Keyboard;
 import org.terasology.input.device.KeyboardDevice;
+import org.terasology.rendering.nui.Canvas;
 import org.terasology.rendering.nui.CoreScreenLayer;
 import org.terasology.rendering.nui.editor.systems.AbstractEditorSystem;
 import org.terasology.rendering.nui.events.NUIKeyEvent;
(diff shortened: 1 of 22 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/editor/layers/AbstractEditorScreen.java
```diff
@@ -113,6 +118,15 @@
     protected abstract Path getAutosaveFile();
 
     @Override
+    public void onDraw(Canvas canvas) {
+        super.onDraw(canvas);
+        if (!autosaveLoaded) {
+            loadAutosave();
+            autosaveLoaded = true;
+        }
```
flo: That looks like a hack to me (no one would expect a autosave loading in a draw method). A comment why it is done there should at least be added.

flo:
The autosave gets created and loaded, but it is neitehr possible to use the "Save" nor the "Save As" button. "Save" is grayed out and "Save As" results in an exception. This problem is propably related to the edited asset no longer being selected after you restarted the game

There is btw. also the special case that someone opens the editor in a situation where the loaded asset is no longer available. I leave it up to you if you want to handle it properly. It is not that important.

GooeyHub:

GooeyHub:

flo:
@rzats The "Save" button still isn't enabled (but "Save As") when you restart the game after a change.

I also noticed a second bug: When you select a different screen, but then choose to cancel the operation (since there are changes) then the save button will override the wrong widget. It will override the widget of which the selection got aborted and not the widget with the modification.

GooeyHub:

GooeyHub:

flo:
nice roundup, merged :)

PR #2453, opened 2016-08-23
[Review/Discussion] Set multiple blocks at once using WorldProvider

skaldarnar:
- added `setBlocks(Map<Vec3i, Block>)` to WorldProviderCore
- chunks are dirtied after all blocks are placed
- finally notifications for each changed block are send
- adjusted BlockPlacingSystem to take advantage of `setBlocks`

- - -

I tested this with structure templates (generating a PlaceBlocks event from the regions to fill) and the BlockBufferSystem of DynamicCities. I did not notice any crashes, and the overall time of placing blocks is reduced as each affected chunk is only re-calculated once after all blocks are set.

Does anybody see a problem in this implementation. Anything I forgot to take care of? I was wondering if the implemenation should make sure that the map is not too big (either restrict by API/contract or paritioning the blocksToPlace and run several iterations). Would it make sense to order by chunks and finish one chunk after the other?

Pinging @immortius @flo @Josharias @msteiger for feedback.

### DIFF
--- a/engine-tests/src/main/java/org/terasology/testUtil/WorldProviderCoreStub.java
+++ b/engine-tests/src/main/java/org/terasology/testUtil/WorldProviderCoreStub.java
@@ -17,6 +17,10 @@
 package org.terasology.testUtil;
 
 import com.google.common.collect.Maps;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.math.Region3i;
 import org.terasology.math.geom.Vector3i;
@@ -29,10 +33,6 @@ import org.terasology.world.internal.WorldProviderCore;
 import org.terasology.world.liquid.LiquidData;
 import org.terasology.world.time.WorldTime;
 import org.terasology.world.time.WorldTimeImpl;
-
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Map;
 
 /**
  */
@@ -109,6 +109,16 @@ public class WorldProviderCoreStub implements WorldProviderCore {
             return air;
         }
         return old;
+    }
+
+    @Override
+    public Map<Vector3i, Block> setBlocks(Map<Vector3i, Block> blocksToPlace) {
+        Map<Vector3i, Block> result = new HashMap<>(blocks.size());
+        for (Map.Entry<Vector3i, Block> entry : blocksToPlace.entrySet()) {
+            Block b = setBlock(entry.getKey(), entry.getValue());
+            result.put(entry.getKey(), b);
+        }
+        return result;
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/world/block/entity/placement/BlockPlacingSystem.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/placement/BlockPlacingSystem.java
@@ -21,13 +21,9 @@ import org.terasology.entitySystem.event.ReceiveEvent;
 import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.entitySystem.systems.RegisterMode;
 import org.terasology.entitySystem.systems.RegisterSystem;
-import org.terasology.math.geom.Vector3i;
 import org.terasology.registry.In;
 import org.terasology.world.WorldComponent;
 import org.terasology.world.WorldProvider;
-import org.terasology.world.block.Block;
-
-import java.util.Map;
 
 /**
  */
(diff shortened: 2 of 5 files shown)


GooeyHub:

flo:
It us very coil that you work on improving the block placement speed. 

I don't think the placment method needs a return value. I can't imagine many/any cases where you need it afer the block placment. If you do need it you can just get the block before the replacement.

The aspect you method addresses is the dirty marking of chunks.  Was it just a guess that this is the largest slowness factor or did you made some measurements? Other potentially performance bottlenecks I could think of would be the per block locking and change event sending.

I wonder if we really need to introduce an additional interface (I mean the method) for setting blocks.
Option 1) The single block placement method could buffer the changes till either end of frame or till chunk switch or block count limit.
Option 2) we could just remember the dirtied chunks in a set till the end of the frame and mark them then as dirty.
Option 3) If locking is a problem we could replace the locking with a kind of optimistic locking where for example setting a block would just result in the increment of a version field. An algorithm that needs to update somthing (graphic/pathfinding graph/light) could remember at the start the value of the version field. At the end the algorithm would then check if it changed. If it changed it would simply run again.

skaldarnar:
Thanks for the feedback @flo. I gave the new methods a return value to correspond to the existing `setBlock` method. I will check if it is ever used. I agree with you that the return value (the previous block) is not needed, but what about the information whether the process was successful or not?

I did no serious profiling on finding the bottleneck, the dirtying of chunks was my first guess and it turned out fine. For a medium-sized structure template, the spawning time dropped from ~90ms to under 20ms. 
I will get my hands on YourKit to check for other factors.

re Option 1) The buffering sounds reasonable (with block count or frame as trigger to build). I'd look into that first and see how that turns out. I don't think that chunk switch would be good trigger as you might have cases where a multi-block structure crosses chunk borders often. For buffering, the method signature would have to be changed to `void` I guess.

re Option 2) This suggestion seems very similar to the buffering in (1) to me. What would the difference be? Setting blocks and sending change notifications right away, but updating the renderable chunks later on?

re Option 3) I'm not absolutely sure I understand what you are saying. I need to have a closer look at the locking you are speaking about...

flo:
yes, that is basically the difference. Of course the advantage of option 2 is that the set (Map) of changed chunks will be much smaller than the set of changed blocks (Map) of option 1. Also it might be implementation wise simpler as you don't need to redirect requests for blocks at a certain position to that map. Implementation wise option 2 would propably have to provide some kind of ChunkDirtyEvent that gets send of the end of the frame and can be registered to also by modules. (Would be usefull for the Pathfinding module that currently schedules a mesh graph calculation for a complete chunk whenever a single block changes).

About option 3) Without locking, the following could happen: Say a chunk has a state A, now you make a change b to it and it has the state B. In terasology the light calculation gets done in a separte thread. So that light calculation might read the chunk in state B and do some calculations on it. While that is in progress, a user makes a change c and the chunk has now state C. Now the light calculation finishes and stores the result that is based on state B in a separate part of the chunk instance. The light calculation algorithm must now realize that it needs to calculate the light again, as the chunk has now state C and not B like when it started.
Currently the light calculation algorithm locks all chunks while it performs the light calculation and would can potentially cause the main thread to wait in a "set block operation" on the light calculation thread. 

When you place just a single building this locking is probably not noticeable, as the light calculation is very unlikely to work with chunks that are near the player. However if a building just got placed, the light calculation might be running when you place a second building.

flo:
I created a way to record ingame performance data about block placement.

It is available in the following StructureTemplates branch performance-measurements: https://github.com/Terasology/StructureTemplates/tree/performance-measurements

Once you have the self build StructureTemplates module enabled you can do the performance measurements simply use it ingame via the following command `showScreen BlockPlacementPerformanceTestScreen`. It replaces the blocks near the character so you sould not use it in worlds you need 

I got the following results after 200 iterations of replacing all 65536 blocks in a complete chunk region:
* original: 721 ms
* pull request: 206 ms
* original + no locking: 615 ms
* pull request + no locking: 95 ms


Through disabling of codes and hot code replacement I figured out that about 40% of the remaing time (95 ms) gets used for the `notifyBlockChanged` calls. 

So you really found and fixed the biggest performance issue.

skaldarnar:
Nice (and easy) way to measure the performance. :+1: So the topics remaining are whether to introdcue a new method (I'd vote for yes for convenience) and how to implement the buffering. You were talking about marking chunks as dirty at the end of frame -- how does the world provider get notified about that? For component systems its via the `update` method, but I don't see anything like that for world provider.

flo:
Merged it. That way we have something to work with in for example a commit that adds a benchmark (coming soon).

PR #2454, opened 2016-08-25
Add a screen for doing benchmarks

flo:
### Contains

This pull request adds a screen that can be used to select and start a benchmark.

Currently the screen contains 2 benchmarks: 
1. A benchmark for the WorldProvider method setBlock 
2. A benchmark for the new WorldProvider method setBlocks (Recently added via #2453 )

### How to test

Run the following command on the command line: `showScreen benchmarkScreen`

The window shows up. You can select a bench mark and start it. It will do 200 iterations and then stop automatically. The results can by selecting the text and then pressing ctrl+c.

### DIFF
--- a/engine/src/main/java/org/terasology/world/internal/WorldProviderCoreImpl.java
+++ b/engine/src/main/java/org/terasology/world/internal/WorldProviderCoreImpl.java
@@ -21,13 +21,6 @@ import com.google.common.base.Predicate;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
 import org.terasology.context.Context;
 import org.terasology.engine.SimpleUri;
 import org.terasology.entitySystem.entity.EntityManager;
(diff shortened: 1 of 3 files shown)


GooeyHub:

flo:
Oh and @rzats, I used the NUI editor to make the screen. I am really glad we have that tool now

rzats:
@flo: great! Feel free to send any further feedback my way :)

Cervator:
Merged it although I couldn't truly finish a test - but it looked to be working alright. For me after 5-15 iterations (of `setBlock`, not `setBlocks`) my CPU would be near 100% and the game window would stop refreshing. Waited 20+ minutes for one test and it was unchanged although it should've been done by then :(

Tried with 8 GB max memory and that helped, got to about 40 cycles but by then the auto-save had triggered and the window was beginning to struggle with 4-5 GB max used. Maybe running benchmarks should temporarily disable auto-save? Or we can do something funny with threading to keep the UI able to still respond? Since I couldn't work it to hit the button to stop benchmarking.

Would also be nice to be able to pick the number of cycles for simple validation testing like this :-)

I wonder, could we bundle the benchmark setups as actual unit tests so they could also be run via Jenkins? Just have the in-game option go scan for available benchmark tests in a given package and fill the benchmark drop-down with those entries?

Probably doable at some point later, I figure :)

flo:
#2460 should mostly fix your memory issue.

flo:
The current location of the benchmark offers the following advantages:
* It is easy for a developer to try out when he reads the comment
* Since it runs ingame it is as close as possible to ingame performance. When we would have a unit test with a test environment we might get completly different results as there might be then for example no parallel threads etc.
* Since the benchmark screen can be opend ingame, it can be used to try out what impact modules have on performance

Cervator:
Thanks for the memory fix!

Yep I don't want to remove the current benchmark screen at all. Am more wondering if we could package the actual benchmark tests in a way so they could be hit both from the in-game screen as well as during automated testing in Jenkins. The results will surely vary between automated vs manual tests but the *incremental* comparisons on the automated tests could be hugely valuable, especially if we run them on PRs :-)

PR #2456, opened 2016-08-25
Improve performance by removing unnecessary locking.

flo:
### Contains
A performance optimization, archived by removing uncessary locking.

In my test this made the setBlocks method of WorldProvider more than twice
as fast.

The locking used to be done mainly by the main thread. Besides the main thread the light merging thread and the chunk mesh building thread were also locking. One reason why it probably got locked was the disposal of the chunk. When a chunk was no logner needed, it got made unusable which could have potentially be a problem for the background threads that were still using it. This patch makes the chunk remain readable even when the chunk is no longer needed. For the garbage collector it makes no difference and we can that way dispose chunks without having to let the main thread wait for background threads which were still reading the chunk(which should improve performance at that point too).

A possible small disadvantage of the removed locking is that if you do a mass block placement, that you could theoretically see very shortly a state where some of the new block placements are done while other aren't. However this is mainly theoretically and should not be a problem during normal gameplay. Even if it is visible in some mass block replacement scenarios, I think most would prefer that glitch over a lag.

The performance benchmarks have been done via the benchmark screen from PR #2454:

Without this patch:

benchmark: WorldProvider.setBlocks
iteration:  200 / 200
last duration:  113.7 ms
min duration:  64.2 ms
median duration:  121.5 ms
avg duration:  122.1 ms
max duration:  216.3 ms

With this patch:

benchmark: WorldProvider.setBlocks
iteration:  200 / 200
last duration:  59.0 ms
min duration:  28.9 ms
median duration:  56.2 ms
avg duration:  56.7 ms
max duration:  112.1 ms

### How to test

Merge #2454 first. Then run the benchmarks once with and without with patch. The setBlocks benchmarks can be run by first calling "showScreen benchmarkScreen"  however it is by default not selected so you need to change the benchmark selection first before you start it.

I would like to give other developers a chance to comment on this and to give feedback. So I think we should wait a few days before we merge this. @immortius and @skaldarnar might be interested in this PR.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/ChunkMeshUpdateManager.java
+++ b/engine/src/main/java/org/terasology/rendering/world/ChunkMeshUpdateManager.java
@@ -164,17 +164,16 @@ public final class ChunkMeshUpdateManager {
             ChunkMesh newMesh;
             ChunkView chunkView = worldProvider.getLocalView(c.getPosition());
             if (chunkView != null) {
+                /*
+                 * Important set dirty flag first, so that a concurrent modification of the chunk in the mean time we
+                 * will end up with a dirty chunk.
+                 */
                 c.setDirty(false);
-                chunkView.readLock();
-                try {
-                    if (chunkView.isValidView()) {
-                        newMesh = tessellator.generateMesh(chunkView, ChunkConstants.SIZE_Y, 0);
+                if (chunkView.isValidView()) {
+                    newMesh = tessellator.generateMesh(chunkView, ChunkConstants.SIZE_Y, 0);
 
-                        c.setPendingMesh(newMesh);
-                        ChunkMonitor.fireChunkTessellated(c.getPosition(), newMesh);
-                    }
-                } finally {
-                    chunkView.readUnlock();
+                    c.setPendingMesh(newMesh);
+                    ChunkMonitor.fireChunkTessellated(c.getPosition(), newMesh);
                 }
 
             }
--- a/engine/src/main/java/org/terasology/world/ChunkView.java
+++ b/engine/src/main/java/org/terasology/world/ChunkView.java
@@ -220,30 +220,6 @@ public interface ChunkView {
      */
     void setDirtyAround(Region3i blockRegion);
 
-    /**
-     * Locks the chunk view, enabling write operations
-     */
-    void writeLock();
-
-    /**
-     * Unlocks the chunk view, disabling write operations
-     */
-    void writeUnlock();
-
-    /**
-     * Locks the chunk view, enabling write operations
-     */
-    void readLock();
-
-    /**
-     * Unlocks the chunk view, disabling write operations
-     */
-    void readUnlock();
-
-    /**
-     * @return Whether the chunk view is locked and hence whether write operations are allowed.
-     */
-    boolean isLocked();
 
     /**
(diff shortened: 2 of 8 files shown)


GooeyHub:

skaldarnar:
I'm on vacation until Sunday, will have a look when I'm back home 😉

Cervator:
I tested out a merged version of this just to make sure the game works, a world looks right, and to get some numbers. While the `setBlock` option for the new benchmarks doesn't play along nicely on my system it looks like `setBlocks` does pretty well - way faster than `setBlock` and it manages to finish the 200 iterations.

Before this PR:

```
benchmark: WorldProvider.setBlocks
iteration:  200 / 200
last duration:  215.9 ms
min duration:  111.3 ms
median duration:  210.9 ms
avg duration:  215.3 ms
max duration:  371.8 ms
```

With this PR:

```
benchmark: WorldProvider.setBlocks
iteration:  200 / 200
last duration:  46.9 ms
min duration:  34.1 ms
median duration:  61.4 ms
avg duration:  63.0 ms
max duration:  208.5 ms
```

Both runs were with the standard 1 GB memory and various other things active on my system.

I don't think there's an API impact since the affected interfaces and such aren't exposed via `@API` - everything is internal. Just architecture. Nothing fails to compile :-)

Cervator:
Wait ... `@API` in `package-info.java` :(

Back the label goes. In this case I take it we're changing API that just doesn't happen to have any external implementation. Why *is* `ChunkView` API anyway?

flo:
@Cervator Wow, for you it is even more than 3 times faster.

skaldarnar:
I can confirm the performance improvements for `setBlocks` as reported by you two. I'm fine with removing the locking alltogether from `ChunkView` - should we wait for a final word by @immortius? The performance improvements are worth some minor visual glitches from time to time imho. 

before patch:

	benchmark: WorldProvider.setBlocks
	iteration:  200 / 200
	last duration:  169,0 ms
	min duration:  67,5 ms
	median duration:  140,7 ms
	avg duration:  142,1 ms
	max duration:  207,3 ms

after patch: 

	benchmark: WorldProvider.setBlocks
	iteration:  200 / 200
	last duration:  56,5 ms
	min duration:  37,7 ms
	median duration:  49,6 ms
	avg duration:  51,2 ms
	max duration:  141,8 ms

Cervator:
Maybe we can aim to merge it right after a soonish Alpha 4 release (couple days hopefully), then as one of the multiplayer tests on Saturday use the newest dev build with this merged to make sure we don't see anything weird on a server?

Cervator:
Merging along with #2460 - while this is technically an API violation I'm not sure the affected code should be exposed in the first place. Might be more of an "undocumented" thing in classic API considerations (meant to be internal with only internal implementations known, but exposed anyway for whatever reason) rather than the occasional trivial violation we hop over while in Alpha.

Cervator:
Oh by the way - @flo should we go ahead and add the benchmark screen to the shiny new developer tools menu?

flo:
@Cervator theoretically yes, but it's not that trival to add properly as it use the Stone block from Core, which is why the screen is currently defined in the Core module which should probably not be accessed from a engine menu.

Cervator:
Got it. Maybe another annotation-based thing to scan for, build the list of available benchmarks that way? But yeah not really as trivial as just connecting a button to a command.

PR #2457, opened 2016-08-25
Add a default implementation for setBlocks.

flo:
### Contains

Makes the setBlocks addition added by #2453 not break the Pathfinding module.

### How to test

Build this PR with Pathfinding module. There should be no compile error.

### DIFF
--- a/engine/src/main/java/org/terasology/world/internal/WorldProviderCore.java
+++ b/engine/src/main/java/org/terasology/world/internal/WorldProviderCore.java
@@ -15,8 +15,7 @@
  */
 package org.terasology.world.internal;
 
-import java.util.Collection;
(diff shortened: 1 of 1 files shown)


GooeyHub:

skaldarnar:
X-reference: https://github.com/Terasology/Pathfinding/pull/25

Cervator:
Compile error here was just from importing `Maps` from a wrong package, the one that previously came from a PC Facade dependency now out of the way. @flo: you might want to run a `gradlew cleanIdea idea` to make sure the dependencies are fully updated and cleaned up :-)

Fixed up and merged, thanks! This may be our first example of an engine API break then quickly resolved via `default` method, woot!

PR #2460, opened 2016-08-28
Fix a memory leak by properly clearing recorded data about temporary block entities

flo:
### Contains

A memory leak fix, that is especially noticable in the setBlock benchmark available via the `showScreen benchmarkScreen` command.

The `EntitySetDeltaRecorder` class records changes to persistent entities
in order to forward them later on to the background saving thread.

It got informed about changes to temporary block entities, but did not
get informed about the silent destruction of those entities when the
entities were no longer needed to represent the block.

This patch marks the entity manager inform the `EntitySetDeltaRecorder`
indirectly about the destruction of the temporary block entities,
so that it knows that the entities do no longer need to be stored.

### How to test
1. Run in game (with 2GB memory)
2. run the command `showScreen benchmarkScreen`
3. Use the dialog to start the "setBlock"-Benchmark

There should be an issue from to much memory consumption even when an auto save happens.

Please that this isn't the only memory leak. I also found one that causes the game to need more memory each frame due to an to large growing `chunksOpaqueReflection` field in `RenderQueuesHelper`. For the latter I will create a ticket.

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityManager.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityManager.java
@@ -380,16 +380,29 @@ public class PojoEntityManager implements EngineEntityManager {
         return EntityRef.NULL;
     }
 
+    /**
+     * Creates the entity without sending any events. The entity life cycle subscriber will however be informed.
+     */
     @Override
     public EntityRef createEntityWithoutLifecycleEvents(Iterable<Component> components) {
-        return createEntity(components);
+        EntityRef entity = createEntity(components);
+        for (Component component: components) {
+            notifyComponentAdded(entity, component.getClass());
+        }
+        return entity;
     }
 
+    /**
+     * Creates the entity without sending any events. The entity life cycle subscriber will however be informed.
+     */
     @Override
     public EntityRef createEntityWithoutLifecycleEvents(String prefabName) {
         return createEntityWithoutLifecycleEvents(getPrefabManager().getPrefab(prefabName));
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


GooeyHub:

skaldarnar:
I can confirm this allows to run the benchmark of `setBlock` with 2GB memory whereas the benchmark crashes with OutOfMemory without the patch. Here are my statistics:

after patch:

	benchmark: WorldProvider.setBlock
	iteration:  200 / 200
	last duration:  959,9 ms
	min duration:  798,1 ms
	median duration:  836,9 ms
	avg duration:  881,7 ms
	max duration:  2856,5 ms

Cervator:
Can also confirm the improvement, even with just 1 GB memory although it is a bit more sluggish. Didn't go through the whole set but was able to gracefully stop the benchmark nearly half way there which is waaay more than I managed last time through :-)

```
benchmark: WorldProvider.setBlock
iteration:  91 / 200
last duration:  1860.5 ms
min duration:  1820.3 ms
median duration:  1920.2 ms
avg duration:  2007.4 ms
max duration:  2914.6 ms
```

One interesting error did get logged but may more hint at an issue with sorting / timing?

```
01:42:14.490 [main] WARN  o.t.engine.internal.TimeBase - Delta too great (3344), capping to 1000
01:42:15.024 [Saving-0] INFO  o.t.p.internal.SaveTransaction - Save game finished
01:42:15.850 [Timer-1] ERROR o.t.r.logic.NearestSortingList - Uncaught exception in sorting thread: java.lang.ArrayIndexOutOfBoundsException: 87718
01:42:17.458 [main] WARN  o.t.engine.internal.TimeBase - Delta too great (2968), capping to 1000
```

PR #2465, opened 2016-09-02
Fix an issue with drawLine() connecting wrong edges of a rectangle

rzats:
### How to test

Open the behaviour tree editor and select one of the nodes in the default behaviour tree. Drag it in a circle around the other and check that the line is properly drawn at all angles.

No outstanding issues!


GooeyHub:

skaldarnar:
Tested before and after - looks all good.

Issue #1694, opened 2015-04-26
Any block setting code should be using PlaceBlocks to allow vetoing


MarcinSc:
Some places in the code use setBlock(...) method of WorldProvider. This should not be allowed and possibly setBlock() should be a private method. Instead the PlaceBlocks event (or a separate new interface/class) should be used to ensure all the events are correctly sent.

msteiger:
Currently, `WorldProviderCoreImpl` sends out notifications to `WorldChangeListener`s whenever `setBlock()` is called. This could become obsolete then. The only implementation of that listener (which would need adjusting) seems to be `NetClient`.

One corner case could be WeatherManager's `CloudUpdateManager`: it puts/removes large numbers of cloud blocks rather frequently, but no one needs to care about it. There's no point in sending events for every single one of them.

One option could be aggregating multiple block changes into a single notification event.

PR #2467, opened 2016-09-03
[Discussion] Block placement renovation

skaldarnar:
### Contains

A renovation of the block placement process. 

Based on a discussion with @flo and @MarcinSc this PR should be used to discuss possible changes to the block placement process. The main questions/criteria addressed are:
- Block placement should be vetoable by systems (prevent placement/destruction of blocks). Therefore, the CheckPlacementPermissionEvent is added. The event is sent whenever a block placement via setBlocks is requested. 
- Keep the performance impact low. Per default, the bounding box for multi block placement is used for the permission check. TODO: implement placementRegions

Related topics: 
- https://github.com/MovingBlocks/Terasology/issues/1694 : cosumable PlaceBlocks event to allow vetoing of block placements.

A lot of discussion and testing. I'm shifting things around in my head an may come up with some alternative suggestions how to handle this. 

I appreciate any feedback! Probably also interested in the topic: @Cervator @immortius @msteiger 
- [ ] implement placement regions
- [ ] add boolean flag to permission event to allow early consumption with positive result (e.g., an authority system might allow all placement instigated by some server entity)
- [ ] deprecate PlaceBlocks and adapt usages to new process.
- [ ] update/add documentation to wiki
- [ ] discuss API and behavior of `setBlocks`

### DIFF
--- a/engine-tests/src/main/java/org/terasology/testUtil/WorldProviderCoreStub.java
+++ b/engine-tests/src/main/java/org/terasology/testUtil/WorldProviderCoreStub.java
@@ -19,7 +19,6 @@ package org.terasology.testUtil;
 import com.google.common.collect.Maps;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Map;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.math.Region3i;
@@ -103,22 +102,18 @@ public class WorldProviderCoreStub implements WorldProviderCore {
     }
 
     @Override
-    public Block setBlock(Vector3i pos, Block type) {
-        Block old = blocks.put(pos, type);
-        if (old == null) {
-            return air;
-        }
-        return old;
+    public boolean setBlock(Vector3i pos, Block type, EntityRef instigator) {
+        //TODO: send check permission event?
+        blocks.put(pos, type);
+        return true;
     }
 
     @Override
-    public Map<Vector3i, Block> setBlocks(Map<Vector3i, Block> blocksToPlace) {
-        Map<Vector3i, Block> result = new HashMap<>(blocks.size());
+    public boolean setBlocks(Map<Vector3i, Block> blocksToPlace) {
         for (Map.Entry<Vector3i, Block> entry : blocksToPlace.entrySet()) {
-            Block b = setBlock(entry.getKey(), entry.getValue());
-            result.put(entry.getKey(), b);
+            blocks.put(entry.getKey(), entry.getValue());
         }
-        return result;
+        return true;
     }
 
     @Override
--- a/engine-tests/src/test/java/org/terasology/math/Region3iTest.java
+++ b/engine-tests/src/test/java/org/terasology/math/Region3iTest.java
@@ -17,12 +17,13 @@
 package org.terasology.math;
 
 import com.google.common.collect.Sets;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
 import org.junit.Test;
 import org.terasology.math.geom.Vector3i;
 
-import java.util.Arrays;
-import java.util.List;
-import java.util.Set;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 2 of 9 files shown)


GooeyHub:

flo:
I wonder if we should be more precise what the entity of the event is meant to be. If we could for example specify (maybe even with field name) that it must be set to the client info entity of the player that that triggered the action directly or indirectly.

Otherwise one module developer might set it to the character entity, the other sets it to the client entity, while the third sets it to the used item. Same goes for the event subscribers which might assume because 1 module sends a client entity that all modules send a client entity when it is about a player.

I think the "client info entity" is the right choice because:
- It stores the permissions the player has
- It is available after the player disconnected. e.g. When a player places a TNT block and then disconnects before it explodes, then the event handler must be able to tell if the player had the permission to modify the world with the TNT. 

Of course that means that every block playing module needs to be adjusted to provide a proper entity. E.g. in my Gooey's quest module the dungeon generation process needs to know who asked gooey to generate the dungeon. Also gooey should probably check before offering a dungeon quest that the player has the permission to modify the nearby area. Same would go for the computer moulde which would need to remember who started the computer, so that the placements done by the computer can be verified.

skaldarnar:
Hmm, I'm not sure about the "client info entity" contract. I think it should be encouraged (through documentation) that, if the player is the instigator, a specific entity should be used (general documentation on the entity system). There could be other use cases where it does not matter which player triggered the placement. For instance, instead of cancelling a dungeon generation because the player has no permissions, he should not be allowed to talk to Gooey in the first place. Having a system enttiy as instigator (as @MarcinSc suggested) would imho be the better option.

skaldarnar:
I tried to put my thoughts into two diagrams for better understanding and as base for discussions. I'm not sure if the world provider should be responsible for setting up the placement permission check, handling the instigator, etc. Therefore, I want to present two approaches.

- - -

blockplacementversiona
This first approach is what is implemented in this PR. A system directly calls `setBlock` on the world provider, who is in turn responsible for setting up the permission check, send it out, wait for the event to be processed, and, if not consumed/canceled, performs the block placement. Finally, the status of the operation is returned to the calling system. Although this process is very simple I'm not sure it's the most intuitive way to do it. Moreover, there is no way to bypass the permission check (is this needed, maybe for internal operations of engine on the world provider?).

- - -

blockplacementversionb

The second approach aims to encapsulate the world provider and prevents modules from using `setBlock` directly. Any system that wants to place blocks would create a `PlaceBlocksEvent`. A basic `BlockPlacementSystem` in the engine would create the permission check, i.e., handle placement restrictions and manage the instigator, etc. The `setBlocks` method of world provider itself would just be responsible for placing the blocks, dirtying the chunks, and so forth. An event would then inform all system about the placement, containing the list/map of placed blocks.

The second approach obviously has more overhead, but I'm not sure about the performance impact. We would need to test that with the new benchmark to get a comparison. Also, I wondered what the cost of attaching the `PlaceBlocksEvent` to the `CheckPermissionEvent` would be? That way all the information on block placement would be available for permission checks, although in most cases it would not be used...

- - -

A final question on the API of `setBlocks` and `setBlock`: I changed the signature to return a boolean as discussed previously. Although this is very reasonable for the singular `setBlock` (either is successful or not), the behavior of `setBlocks` has to be defined in the case where _some_ blocks cannot be placed. Should they be skipped, and what should the result of the operation be? Should the whole process be aborted with `false` as return (need to reset already placed blocks)? Any thoughts?

flo:
1. On the approach to capsulate WorldProvider by a BlockPlacementSystem. WorldProivder is already a capsulation. The engine has other ways to access  the data (e.g. by doing manually what WorldProivder does). There is also a WorldProviderCore interface that is in the internal package, maybe that was an intended way to provide the engine with additional methods, athough I am not sure if we need that.

2. on the return value of setBlocks: I think the setting should be atomic (nothing or all) and thus the return value could be a boolean. If we just do a bounding region check then we don't know which blocks prevented the placement anyway.

3. on the entity contract: What shall a module that checks permissions do, when it receives a unknown entity? Should it deny the placement or allow it?
Option a.) It denies the placment when the entity is unkown. As a result it would mean that all other systems that provide a different entity type won't work, and thus we basically still enfoce a single entity type. So that is probably an option module developers shouldn't go for.
Option b.) It allows the placement when the entity is unknown. As a result systems can bypass any placement check by simply setting it to Entityref.NULL or some other empty entity. Thus the permission check becomes optional again. So the "enforcing" advantage of doing it in setBlocks is gone, although doing the check in setBlock(s) will sill encourage developers to provide an argument. Especially when they know what they need to provide.

By specifying on the other hand that a client info entity should be provided, we could introduce a system that allows only players with the modifyWorld permission to place blocks. This simple permission would be then enough to cover all cases where players modify the world as all systems/modules would have to go through the permission check with a proper player in order to place blocks at all. (For an observer only feature)

On the other side: Are there really good use cases were we would want to provide a non player entity as permission source?

Cervator:
One big question: is this permission setup something we *have* to build into the engine, or can we just support it but leave it up to modules to define permission types and how to grant & test them? Which I guess the initial post is suggesting via `CheckPlacementPermissionEvent` .. (support only)

A classic use case (simply thinking as a modder or player, not an engine dev) would be a player-based ownership system where Player A couldn't interact with blocks "owned" by Player B unless explicitly granted permission. But that could be completely separate/different than something tied more to whether a client player has "cheat" rights to affect blocks in unusual ways.

(Please forgive me if I'm being dense, only just digging out from under recent RL craziness and not quite back on top of things yet)

flo:
@skaldarnar I think we talked about this pull request and decided to close it as it is not so important right now.

GooeyHub:

bpas247:
So, are we going to close it?

jellysnake:
Closing as per discussion above. Thanks bpas247

PR #2479, opened 2016-09-12
Vive support

indianajohn:
### Contains

Adds OpenVR support to Terasology.

When making my changes, my objective was to make this change completely invisible to people without VR headsets. There are many design decisions that will need to come downstream of this change, such as in-world overlays. This commit does not include any of that. If Terasology is launched with the `vrSupport` flag in `config.cfg` set equal to true, the world will be rendered in the headset (as well as on the screen). If not, everything acts exactly as it did before.

The designers of the game should draw their attention to the following aspects of this commit.
- I have removed any references to Oculus and TeraOVR.
- I renamed the flag `oculusVrSupport` to `vrSupport` for consistency.
- Since this uses OpenVR, it should work with both Oculus headsets and HTC Vive.
- OpenVR native libraries are included in `project_root/openvr_natives`. This directory must be present when the game is launched.
- My wrapper has been added directly to the Terasology source code / repository. I did this rather than keep it in a separate repository because the wrapper uses lwjgl, and compiling the wrapper to a jar would add a certain version dependency for lwjgl that could conflict with Terasology.
- A jar for `jopenvr` has replaced TeraOVR. This jar depends on JNA version 4.2.2, so as to follow Terasology. If the JNA dependency in Terasology ever changes, we will need to rebuild this jar. I felt this was cleaner than having a ton of machine-generated source code tagging along in the main repository, and the JNA version is unlikely to change in the future.
- Motion controller support is implemented through a listener. This listener is not used at the moment; this patch is already big enough, and that should be done separately and in a focused manner.
- As far as I know, this patch adds no work and actually subtracts some work to the directed acrylic graph effort. It's not programmed in an ideal way (My render frame buffers exist outside of their managing system), but it works as it is, mostly follows their design patterns (VR support is isolated to a single, new DAG node), and can be refactored later. The reason I left it like that is that it's clean/tidy that `openvrprovider` has no dependencies inside of Terasology and is completely self-contained.
- Ideally, down the road, we probably want to add some way that modules can add DAGs to the rendering pipeline and refactor VR support into an external module, but that is impossible with the DAG stuff in its current state. However, by merging this patch now, it's no more difficult to refactor later, and I can begin to write some of the support classes for VR interaction/gameplay.

Anyway, this has already gotten long enough. I'll add more comments later if I think of them.
### How to test
- If you have a VR headset, edit config.cfg such that `vrSupport` is set to true. Launch SteamVR. Launch the game. Edit your rendering settings to disable anything that could be incompatible with VR; for example "film grain" or "blur." Launch the game and put on your headset.
- If you have no VR headset, this patch should have no visible effect on your gameplay. If it does, I did something wrong and we should address it.


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/cameras/OpenVRStereoCamera.java
```diff
@@ -15,23 +15,27 @@
...
 import static org.lwjgl.opengl.GL11.glMatrixMode;
 
 /**
- * Camera which can be used to render stereoscopic images of the scene for the Oculus Rift.
+ * Camera which can be used to render stereoscopic images of the scene for VR.
  *
  */
-public class OculusStereoCamera extends Camera {
+public class OpenVRStereoCamera extends Camera {
+    private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(OpenVRStereoCamera.class);
```
Cervator: You could shorten this a bit with a couple imports, dunno why the extra qualification snuck in :-)

`private static final Logger logger = LoggerFactory.getLogger(OpenVRStereoCamera.class);`

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
```diff
@@ -0,0 +1,131 @@
...
+    public void initialise() {
+        renderingConfig = config.getRendering();
+        renderingDebugConfig = renderingConfig.getDebug();
+
+        finalPost = worldRenderer.getMaterial("engine:prog.post");
+        debug = worldRenderer.getMaterial("engine:prog.debug");
+        addDesiredStateChange(new BindFBO(FINAL));
+    }
+
+    /**
```
Cervator: May want to remove this or fill it in - the javadoc from whatever the `@Override` comes from should get used by default in some cases, maybe unless there's a replacement bit of javadoc like this. Not totally sure, but it stuck out :-)

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,332 @@
+package org.terasology.rendering.openvrprovider;
+
+import com.sun.jna.Memory;
+import com.sun.jna.NativeLibrary;
+import com.sun.jna.Pointer;
+import jopenvr.*;
```
Cervator: Code style generally favors writing out imports instead of allowing wildcards. Also, the `jopenvr` package doesn't include a domain? That's sort of odd.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,332 @@
...
+                controllerStateReference[c] = inputStateRefernceArray[c];
+                vrState.updateControllerButtonState(controllerStateReference);
+            }
+        }
+    }
+
+    private boolean initializeOpenVRLibrary() throws Exception {
+        if (initialized)
+            return true;
+        System.out.println("Adding OpenVR search path: " + OSValidator.getLibPath());
```
Cervator: Might want to convert these to `logger` style, if possible :-)

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
```diff
@@ -0,0 +1,99 @@
...
+    private Matrix4f[] projectionMatrices = new Matrix4f[2];
+
+    // In the tracking system intertial frame
+    private Matrix4f headPose = new Matrix4f(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1);
+
+    // Controllers
+    private static Matrix4f[] controllerPose = new Matrix4f[2];
+    private static VRControllerState_t[] lastControllerState = new VRControllerState_t[2];
+
+    OpenVRState() {
```
Cervator: Package private constructor, rather than just public?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -33,36 +34,14 @@
...
-import org.terasology.rendering.dag.nodes.ObjectsOpaqueNode;
-import org.terasology.rendering.dag.nodes.OutlineNode;
-import org.terasology.rendering.dag.nodes.OverlaysNode;
-import org.terasology.rendering.dag.nodes.PrePostCompositeNode;
-import org.terasology.rendering.dag.nodes.ShadowMapNode;
-import org.terasology.rendering.dag.nodes.SimpleBlendMaterialsNode;
-import org.terasology.rendering.dag.nodes.SkyBandsNode;
-import org.terasology.rendering.dag.nodes.ToneMappingNode;
-import org.terasology.rendering.dag.nodes.WorldReflectionNode;
+import org.terasology.rendering.dag.nodes.*;
```
Cervator: This probably got "optimized" by IDE settings thinking wildcards are good after a certain point. If using IntelliJ In Settings -> Editor -> Code Style -> Java -> Imports tab you should be able to set "Class count to use import with '*' to a really high number to avoid that.

Cervator:
Very cool to see this! Looking forward to trying a full test with my DK2 sometime soon. Right now I just went over the code looking for style issues and getting an idea of the logistics behind the wrapper, natives and such. @emanuele3d and @tdgunes will be much more useful with the actual (scary) 3d wizardry code :-)

I see some clues to more neat stuff like motion control and look forward to all the deeper work on actually having VR make sense. Also see some style issues and probably initial draft stuff (bit of copy paste / shoving stuff in from somewhere else?) that probably was on the todo list for later anyway, added some guidance for some of that. May want to set up and try Checkstyle (ideally as an IDE plugin) sometime later when clearing up style issues, makes it really easy.

Is the `jopenvr` jar coming from https://github.com/indianajohn/jopenvrwrapper or was that the wrapper bits that were just added as source instead and the jar came from OpenVR itself?

As for the natives we have LWJGL set up with a neat little `gradlew extractNatives` that grabs the relevant files out of actual LWJGL dependency jar files we fetch via Jenkins, then drops said files into a `/natives`. Maybe we could do something similar? I don't mind helping out with that part.

GooeyHub:

indianajohn:
I addressed your first round of comments.

Regarding the above, I compiled jopenvr and added it as a jar - to me, adding a ton of machine-generated source code is a bit ugly. If you would prefer, I can just stick the source in there.

In response to your comment, jopenvr, as harvested from Vivecraft, does not have a domain. If that is too untidy then we can crack open JNA and wrap it ourselves, with a domain name. But that seemed to be a lot of work for something that was mainly aesthetic.

While for this PR I favored not compiling jopenvrwrapper into an artifact, I'm perfectly open to doing it that way. In that case, feel free to fork jopenvrwrapper under MovingBlocks.

* Pro: jopenvrwrapper and all its contents remain a separate, independent package.
* Con: If we were to ever change the LWJGL version, we would have to recompile/add the jar. It also adds a little bit of inflexibility in the case that we ever want to edit the wrapper to gain access to something else. In such a case, we would have to recompile and add the jar again.
* Complication: Similar to lwjgl, packaging the native OpenVR libraries is kind of a pain in the ass, so it's not really possible to make a stand-alone jar. It has to drag its native libraries along with it.

I leave that decision up to you.

Cervator:
Thanks :-)

We have neat ways to avoid (some) machine generated source, TeraMath uses something cool to write a method type once then generate a bunch with specific argument / return types or something like that. But that's low priority and something we could always just look into later.

It is more just weird than a problem that the jopenvr package doesn't have a domain. Not worried.

Appreciate the extra details on wrapper etc, also something that be looked at more later, is good just to have something working right now :-)

indianajohn:
I just hit everything with CheckStyle. I left some stuff behind, such as:

`public static int k_EAxis_Trigger = 1;`

That variable name matches the original C API and it would be confusing to rename it just to fit Java style guidelines. Most of the examples I've left behind are similar.

If there's something you see that doesn't rub you the right way I'll change it.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/cameras/OpenVRStereoCamera.java
```diff
@@ -15,51 +15,56 @@
...
 import static org.lwjgl.opengl.GL11.GL_PROJECTION;
 import static org.lwjgl.opengl.GL11.glMatrixMode;
 
 /**
- * Camera which can be used to render stereoscopic images of the scene for the Oculus Rift.
+ * Camera which can be used to render stereoscopic images of the scene for VR.
  *
  */
-public class OculusStereoCamera extends Camera {
+public class OpenVRStereoCamera extends Camera {
```
emanuele3d: Some info for the future: camera should eventually become components in the Entity System. As the ES is all about reusing and composing components, I see VR cameras as pairs of standard cameras rather than specialized classes. Obviously this will work for the time being. But is there anything in this class that might cause trouble in that vision?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        PerformanceMonitor.endActivity();
+    }
+
+    private void renderFinalStereoImage(RenderingStage renderingStage) {
+        if (this.vrRenderer == null) {
+            this.vrProvider.init();
+            this.vrRenderer = new OpenVRStereoRenderer(this.vrProvider, 1280, 720);
+        }
```
emanuele3d: Why is this done here rather than in initialization? Or perhaps in setOpenVRProvider()? Bottom line is we shouldn't have to go through this IF every frame.

Also, please no magic numbers (1280, 720). And in this context I'd like to understand what's happening here. They seem to set the renderer resolution in an hardcoded way? It's not possible to render at different resolutions?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
```diff
@@ -0,0 +1,121 @@
...
+        switch (renderingStage) {
+            case LEFT_EYE:
+                glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+                vrProvider.updateState();
+                org.lwjgl.opengl.EXTFramebufferObject.glBindFramebufferEXT(
+                        org.lwjgl.opengl.EXTFramebufferObject.GL_FRAMEBUFFER_EXT, vrRenderer.getTextureHandleForEyeFramebuffer(0));
+                glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
```
emanuele3d: Why are you clearing the color and depth buffer twice here?

emanuele3d: Actually, let me rephrase. The first glClear seems to clear the Final FBO which is bound before process() is executed. But in this node nothing seems to get rendered to it? So, why bind it and clear it?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
```diff
@@ -0,0 +1,121 @@
...
+                renderFullscreenQuad(0, 0, 1280, 720);
+
+                break;
+
+            case RIGHT_EYE:
+                glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+                org.lwjgl.opengl.EXTFramebufferObject.glBindFramebufferEXT(
+                        org.lwjgl.opengl.EXTFramebufferObject.GL_FRAMEBUFFER_EXT, vrRenderer.getTextureHandleForEyeFramebuffer(1));
```
emanuele3d: This OpenVR thing requires you to write to its own FBOs rather than the default OpenGL FBO? Is it not compatible with the the OpenGL GL_LEFT/GL_RIGHT FBOs?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
```diff
@@ -0,0 +1,121 @@
...
+                renderFullscreenQuad(0, 0, 1280, 720);
+                vrProvider.submitFrame();
+                GL11.glFinish();
```
emanuele3d: Is glFinish() this something required by the VR setup?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
```diff
@@ -0,0 +1,121 @@
...
+                break;
+            case MONO:
```
emanuele3d: Hmmmm... I understand this is some defensive programming but is this case ever reachable?

engine/src/main/java/org/terasology/rendering/openvrprovider/OSValidator.java
```diff
@@ -0,0 +1,61 @@
+package org.terasology.rendering.openvrprovider;
+
+/* Mainly used to locate the OpenVR library on different platforms. */
+public class OSValidator {
```
emanuele3d: This looks like the kind of class that shouldn't be deep down in the render. @msteiger or @Cervator can perhaps provide a better home package for it?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,374 @@
...
+import static org.terasology.rendering.openvrprovider.ControllerListener.RIGHT_CONTROLLER;
+
+/* This class is designed to make all API calls to OpenVR, thereby insulating it from the user. If you're looking to get
+ * some information from the headset/controllers you should probably look at OpenVRStereoRenderer, ControllerListener,
+ * or OpenVRState
+  * */
+public class OpenVRProvider {
+    private static boolean initialized;
+    private static final Logger logger = LoggerFactory.getLogger(OpenVRProvider.class);
+    private static VR_IVRSystem_FnTable vrsystem;
```
emanuele3d: Should this be vrSystem instead?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRStereoRenderer.java
```diff
@@ -0,0 +1,73 @@
...
+        createRenderTexture(lwidth, lheight);
+    }
+
+    public void deleteRenderTextures() {
+        if (eyeTextureIDs[0] > 0) {
+            GL11.glDeleteTextures(eyeTextureIDs[0]);
+        }
+    }
+
+    public void createRenderTexture(int lwidth, int lheight) {
```
emanuele3d: Buffer management and (eventually) binding should be done through the existing facilities and directly in the nodes. 

See for example:

`requiresFBO(new FBOConfig(REFLECTED, HALF_SCALE, FBO.Type.DEFAULT).useDepthBuffer(), displayResolutionDependentFBOs);`

in the WorldReflectionNode. Are there incompatibilities with this way of doing things?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -256,6 +258,11 @@ private void initRenderGraph() {
         renderGraph.addNode(toneMappingNode, "toneMappingNode");
         renderGraph.addNode(bloomPassesNode, "bloomPassesNode");
         renderGraph.addNode(blurPassesNode, "blurPassesNode");
+        if (renderingConfig.isVrSupport()) {
```
emanuele3d: At this stage nodes are enabled or disabled via the ConditionDependentNode.requireCondition() method. So, no need to have an IF block here. And there are no performance consequences as a disabled node has no footprint in the rendering tasklist, which is what gets executed to render every frame. See the LightShaftsNode for a simple example on how to disable a node monitoring a rendering config property.

Eventually nodes will need to add or remove themselves from the node list (eventually node graph) depending on config parameters and any other criteria they wish to establish. Something for you to keep in mind for later.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -256,6 +258,11 @@ private void initRenderGraph() {
...
+            Node copyToVRFrameBufferNode = nodeFactory.createInstance(CopyToVRFrameBuffersNode.class);
+            ((CopyToVRFrameBuffersNode) copyToVRFrameBufferNode).setOpenVRProvider(this.vrProvider);
```
emanuele3d: Can the vrProvider be available through the Context and then via injection?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
```diff
@@ -0,0 +1,121 @@
...
+        renderingConfig = config.getRendering();
+        renderingDebugConfig = renderingConfig.getDebug();
+
+        finalPost = worldRenderer.getMaterial("engine:prog.post");
+        debug = worldRenderer.getMaterial("engine:prog.debug");
+        addDesiredStateChange(new BindFBO(FINAL));
+    }
+
+    @Override
+    public void process() {
```
emanuele3d: I find this method problematic. 

Ultimately the debug mode will probably be in its own node. This will remove the first IF block and each material will be managed through the appropriate EnableMaterial state change in each node.

Similarly, this whole VR-related node will be added to the render graph only if VR is enabled and removed from the graph when it gets disabled.

Given the current "in-flux" state of the renderer those things will take some time.

Meanwhile I think you could already use addDesiredStateChange(new EnableMaterial(....)) and its opposite removeDesiredStateChange() to manage which material should be enabled. Just remember to use Node.refresthTaskList() when you the swap of state changes has taken place.

emanuele3d:
Overall comment about this PR. 

By now you might have read my exchange with Cervator in Slack: I would have preferred you helped me and tdgunes with what's needed to complete the new architecture of the rendering engine rather than just plough ahead with a way of supporting VR that will, at least in part, need refactoring.

Furthermore, I think you can try harder in fitting the new code into the way the rendering engine works. I recommend you go through all the pre-existing nodes for inspiration.

That been said, I like your coding standards and the way you structured your commits. I could use some more comments and javadocs but they seem to be fallen out of fashion these days.

Cervator seems keen on having this PR merged at some point. So, let's work on it. I look forward to your replies on the various comments.

engine/src/main/java/org/terasology/rendering/cameras/OpenVRStereoCamera.java
indianajohn: In the end, the VR camera should be part of the entity; as a person using a room-scale VR system moves through their play space, I believe that should move the player location (for the purposes of lighting, world interaction, etc...). It should be easy to make VR work with standard cameras, provided it's possible to replace them at some later point if the VR API dictates it.

Typically, the projection matrices will not change, and the tracked headset pose is all that would need to be updated (and that would be associated with the user, not the camera). An example of where that might we relevant would be if the user adjusted the IPD (there is a knob on the Vive for that). That will signal the SDK to change the projection matrix and headset->left eye / headset -> right eye pose matrices.

In summary, "there shouldn't be any problems if we're careful."

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
indianajohn: For some reason, when I tried to include this in the initialization the frame buffers weren't being properly created.  In hindsight, I think that could have been due to an earlier issue I had where `updatePoses()` (an OpenVR API method) was being called more often than `submitFrame()` in a separate thread (my initial design was to have a separate poller for controllers/poses, but OpenVR didn't seem to like that, and the documentation suggests it was a a flawed approach; OpenVR wants `updatePoses()` to be called on the same thread as `'submitFrame()`). At any rate, I will try to work it back into the initialization.

The 1280 x 720 frame buffer size just happens to be the size that the SDK example uses.
https://github.com/ValveSoftware/openvr/blob/master/samples/hellovr_opengl/hellovr_opengl_main.cpp

I will change make the frame buffer size adjustable and take out the hard coding.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
indianajohn: That was just an artifact from earlier, when I had the final copy to the screen buffer and the copy to the headset frame buffers in the same node. I'll take it out.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
indianajohn: Actually, it's possible to assign a FBO to the headset, as in

```
vrProvider.texType[nEye].handle = eyeTextureIDs[nEye];
             vrProvider.texType[nEye].eColorSpace = JOpenVRLibrary.EColorSpace.EColorSpace_ColorSpace_Gamma;
             vrProvider.texType[nEye].eType = JOpenVRLibrary.EGraphicsAPIConvention.EGraphicsAPIConvention_API_OpenGL;
             vrProvider.texType[nEye].write();
```

My frame buffer wrapper class (OpenVRStereoRenderer, poorly named) made more sense from the perspective of an independent wrapper, but I guess it's not entirely out of line to expect a third-party application to have some sort of FBO wrapper class or registry, so I guess it was just short-sighted design.

As mentioned above, I'll try to make my handling of FBOs more consistent with your work.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
indianajohn: Refer to this example once again: https://github.com/ValveSoftware/openvr/blob/master/samples/hellovr_opengl/hellovr_opengl_main.cpp

Without `glFinish()` after `submitFrame()` (and also `glClear()` in some cases), the headset can flicker and act unpredictably. The good news is that, once I get it right on one computer, it seems to work on all other computers (that I have access to), so it's not a platform-specific thing.

If you see something strange with '`glClear()` or `glFinish()` in my commits down the road, it could be that I just be experimentally the thing that ended up working the best.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
indianajohn: That's actually something I meant to take out. I'll get rid of it.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
indianajohn: Yup, I think I fat-fingered that right before a commit.

engine/src/main/java/org/terasology/rendering/openvrprovider/OSValidator.java
Cervator: Good catch. Yep there is indeed a `org.terasology.utilities` package which even has an existing `NativeHelper`, it might be an idea to look at enhancing that :-)

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRStereoRenderer.java
indianajohn: As I mentioned earlier, I think I can just create the FBOs through Terasology's system and send their ID to OpenVR's compositor. I'll give it a shot anyway.

Cervator:
Thanks for the thorough review @emanuele3d ! And for already responding to some of it @indianajohn :-)

I like getting stuff merged so code doesn't get dusty or drifts away from `develop` too long in its own branch or so. Most notably that happened with the original combat system which kept improving in a branch till it looked awesome (positional damage from physics-driven arrows!) ... then the author disappeared leaving nobody to help get it up to speed with the changes that had happened in `develop` in the meantime and it was never merged :(

So smaller chunks of stuff we can keep merging in incrementally and improve further is good, although that still doesn't mean we need to rush, especially with something like VR which isn't exactly critical, even if it is fun to play with :-)

As for comments we need to make Checkstyle complain more about javadoc sometime soon. More doc is good!

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
indianajohn: Excuse my Java inexperience here, but you mean that I should add it to the context and then pass the context into the node?

Cervator: Not so much Java related as about our architecture with `Context` serving as a sort of object store (in a less global way than the `CoreRegistry that it is replacing). 

I suspect what Manu is getting at here is that if it was added to a Context it could be made available in other places later on, for instance if a module wanted to add some VR related effect it could eventually `@In OpenVRProvider vrProvider;` then use it for stuff.

That's a guess though :-)

indianajohn:
@emanuele3d I'll try to correct the points you noted ASAP. The branch will likely see a few commits soon, but I'll comment again when I reach a point where I think everything has been properly addressed.

I appreciate your time working through this with me. I have no problem pitching in on the refactoring effort (and will likely continue to do so in the future), but I'll ask you to forgive some selfishness on my part in terms of my own priorities/time. Without a VR headset staring you in the face, my motivations might be a bit hard to understand, but I just had to see it in VR. I'll stick more to the party line in the future.

Cervator:
/me pretends to totally not own Elite Dangerous because of wanting to try it in VR ...

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
indianajohn: For the time being, I just took the conditional out and bound the final post processing material. I hope that's fine.

GooeyHub:

indianajohn:
Okay. So with my most recent changes, all of the comments above have been addressed (aside from Javadocs :/).

One thing that we might want to address later is moving `vrProvider` into a the initialization sequence of the game and controlling when/if it gets initialized. I don't think it's time to do that now, however. Since we have no menus or anything to support the VR version of the game, it's best to leave it so that the VR version starts after the user has navigated all of the menus, and that's the current behavior with the object being initialized/instantiated when `WorldRendererImpl` is instantiated.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
emanuele3d: I vaguely remember reading about Oculus Rift and John Carmak explaining he had to circumvent a number of standard practices to extract as much performance as he could from the display. I wonder if this is somehow related: buffering issues that are not quite yet well known in the world of monitor-based graphics. Ultimately: ok.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: Ah, I just noticed you made changes to use injection already: great!

engine/src/main/java/org/terasology/config/RenderingConfig.java
```diff
@@ -328,7 +328,7 @@ public void setVignette(boolean vignette) {
     }
 
     public boolean isMotionBlur() {
-        return motionBlur && !isOculusVrSupport();
+        return motionBlur && !isVrSupport();
```
emanuele3d: @indianajohn : this is obviously a legacy issue, but did you have a look at why this is? Why is motion blur incompatible with VR? This leaves us vulnerable to code that just relies on the property change event alone rather than, say, calling isMotionBlur or keeping track of both motion blur AND vr support property.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
```diff
@@ -0,0 +1,118 @@
...
+        finalPost.enable();
+        if (!renderingConfig.isVrSupport()) {
+            logger.warn("CopyToVRFrameBufferNode processed in non-VR mode.");
+        } else {
+            renderFinalStereoImage(worldRenderer.getCurrentRenderStage());
+        }
+        PerformanceMonitor.endActivity();
+    }
+
+    private void renderFinalStereoImage(RenderingStage renderingStage) {
```
emanuele3d: Not quite there yet but moving in the right direction. ;)

The goal with FBOs is that a node requires one on initialization and that FBO is automagically bound before the process() method is executed. This is achieved just by calling requireFBO() like you did, no need for further .bind() calls on the FBO.

The problem is this node requires two FBOs, one for each eye. In the future much of the current nodelist will be duplicated into parallel sets of nodes when VR is enabled, so this class would be instantiated twice and each instance would just require one FBO. For the time being however you need to have the right FBO bound at the right time.

(....after a 10 mins pensive pause...)
(...started writing something... stopped again...)

Ok, I think we need to take a step back.

What does this node do? For every frame it first uses the finalPost material to render the image for the left eye into the device's FBO. Then the renderer does a second pass through all nodes, reaches this node and the image for the right eye is generated, again into the device's FBO. Then the frame is submitted, which I seem to understand is when the hardware/driver for the HMD takes over the content of the two FBOs and actually displays the image in the headset.

In both rendering stages however, the FinalPostProcessingNode is processed too. So if I'm reading everything correctly, this last node renders a mono image to display, twice per frame.

It seems to me these aspects are related and should be tackled together. It's all about rendering and displaying the right thing on the right device (HMD vs monitor). This is different from how we used to do things for Oculus: OpenVR seems to require two separate (undistorted) color buffers when Oculus needed a single pre-composed and pre-distorted image.

So. I think we need to refactor the two last nodes in the list, CopyToVRFrameBuffersNode and FinalPostProcessingNode, and make three nodes out of them, with narrower, more precise responsibilities.

1. First we need a FinalPostProcessing node whose _sole_ responsibility is to render the final image into the FINAL FBO, whatever the circumstances (mono/left/right).
2. Then we need a CopyImageToHMD node which (if VR is enabled) takes the content of the final FBO and _copies it_ (not renders it) into the appropriate device-bound FBO. When the second eye is processed this node is also responsible for copying the content of both left/right FBOs back onto FINAL, so that it contains a split screen with the images for both eyes.
3. Finally, a CopyImageToScreen node _copies_ the content of the FINAL FBO to the default FBO, the display, but only when in MONO mode or when the second eye is processed. And this node should also include the functionality to handle screenshots, which is currently in FinalPostProcessingNode.

Notice the emphasis on the _copies_ verbs: the last rendering to a quad happens in the FinalPostProcessing node. The following nodes use inexpensive copy operations instead. 

How does this sound? Once we have these three nodes up and running we can revisit the issue of binding the right FBO for the appropriate eye at the right time.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -349,7 +355,7 @@ private void preRenderUpdate(RenderingStage renderingStage) {
      * to provide statistics regarding the ongoing rendering and its individual steps (i.e. rendering shadows,
      * reflections, 2D filters...).
      *
-     * @param renderingStage "MONO" for standard rendering and "LEFT_EYE" or "RIGHT_EYE" for stereoscopic displays.
+     * @param renderingStage "MONO" for standard rendering and "leftEye" or "rightEye" for stereoscopic displays.
```
emanuele3d: LEFT/RIGHT_EYE are consistent with the constants the code use. Why then rename to left/rightEye? And why not rename MONO too?

engine/src/main/java/org/terasology/config/RenderingConfig.java
indianajohn: I would say that motion blur is incompatible with VR, as are several other effects. However, I think the best way to handle such cases is to take the extra conditional out and add a VR preset in a future PR.

I'll go ahead and take this out.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
indianajohn: I guess I got a little overzealous with CheckStyle; removing.

indianajohn: To be more clear, I was swatting CheckStyle violations in openvrprovider and that was caught by a pattern match.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
indianajohn: I'm on board with that, except for one detail. Do you really think that we need a split screen rendering in the default frame buffer? I personally think it would be better to just copy the left eye to the screen frame buffer. It's just as cheap (maybe cheaper), so at this point it's entirely a stylistic decision. It's a decent visual indication that VR rendering is in progress, allowing the user to be less confused if they start things up without the headset on, but for an outside observer it's easier to see what's going on in game if only a single frame buffer is copied to the screen.

The SDK also has an example of querying the distortion pattern and performing distortion on the image that is sent to the screen. Note that it's done after the frame is submitted to the compositor; it's entirely for visual effect on the rendering window. I'm personally not a fan, because, once again, there is some utility in somebody outside of VR being able to clearly observe what the player is seeing.

Anyway, I will get to this anywhere between the next couple hours to couple days.

emanuele3d: I did think about that actually and I suggested the split-screen mostly to keep in line with the visual behavior we had with the Oculus support. 

If @Cervator has nothing against it I'd be happy with the screen showing only one eye and I'd ask you to open an issue requesting the functionality to _optionally_ see on screen what the HMD displays, including distortion, for demo and debug purposes.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: Understood.

engine/src/main/java/org/terasology/config/RenderingConfig.java
```diff
@@ -328,7 +328,7 @@ public void setVignette(boolean vignette) {
     }
 
     public boolean isMotionBlur() {
-        return motionBlur && !isVrSupport();
+        return motionBlur;
```
emanuele3d: So, motion blur works fine also in stereo?

emanuele3d: Sorry, I just read a comment from you on this topic via email. Disregard the comment above. Indeed if some effects are incompatible with VR enabling VR should disable them automatically and perhaps leave the choice to the user to re-enable them to see what happens - this would also enable us developers to assess the problems and potentially solve them.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyToVRFrameBuffersNode.java
Cervator: I'm fine with whichever option. I like the idea of adding more utility/config later to be able to watch in various ways.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/config/RenderingConfig.java
indianajohn: I actually ended putting this back in. The reason being, it seems that motionBlur is not being read properly or something - in my settings it's unticked, but blur is active all of the time, and blur makes VR unusable. We can take this out when we fix the blur property reading / whatever the bug is.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/config/RenderingConfig.java
Cervator: There are two distinct categories of blur for what it is worth: motion blur (distorts the screen when you turn quickly, more visible if the system is laggy) and general blur (just uses distances to apply an overall blur to stuff, like when your vision intersects a goofy billboard plant in the foreground and everything behind it gets blurry)

indianajohn:
Alright, in the current branch, the rendering sequence in the case VR is not active looks like:
(all rendering nodes) -> FinalPostProcessingNode (writing to FINAL fbo) -> CopyImageToScreen

In the case that VR is active, it looks like:
LEFT_EYE : (all rendering nodes) -> FinalPostProcessingNode (...) -> CopyImageToHMD -> CopyImageToScreen ->
RIGHT_EYE:  (all rendering nodes) -> CopyImageToHMD

There are still steps that can be taken I think, but I think they'd be best addressed later:

* We can move the VR provider further up the stack and make it a subsystem that more components can access. (At the moment, the fact that OpenVR doesn't actually get bound/started until the user navigates the menu is actually preferable).
* We could just re-use the rendering logic for a single eye and input some flag for an eye upon node creation. We would then create two nodes of the same type, with different eye indices stored on them. (I think this should be part of a future commit that removes the LEFT_EYE/RIGHT_EYE logic entirely).
* We could create a pre-set for rendering settings that should be used with VR, but allow users a bit finer control over what they actually use. (waiting on the motion blur property reading being fixed).
* We can come up with a setting, which, when enabled, shows what the VR user is seeing on the main display.

I've tested that the following things still work:
* Game support when VR is disabled on a machine that does not support it (the game functions as it did before). Note that this is the default state of the config file.
* Game support when VR is enabled on a machine that supports it - the game renders in VR.
* When a computer that does not support VR tries to enable VR support, the game fails to start properly and log messages are generated.
* Screenshot support seems to function both with and without VR support active.

Moreover, I think that the rendering system is actually a bit wider than when I picked it up.

So anyway, I think the patch is ready to be merged. Let me know what you think.

GooeyHub:

indianajohn:
P.S. I can squash this if you want.

GooeyHub:

GooeyHub:

emanuele3d:
@indianajohn: I am aware of your latest commits. I will review you them as soon as possible. 

If you think it'd be good to squash some commits this would be a good time as it might take me a few days before I can go through them. I'm working on a PR too.

indianajohn:
Proving that I can't leave things alone, I pushed another commit today that handles the case:

* When a computer that does not support VR tries to enable VR support, the game fails to start properly and log messages are generated.

Now, when this happens, the game starts as normal with the log message:

```
12:27:07.404 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - VR Headset not detected.
```

In the future, we should plan for the game to initialize the VR library upon launch and present the user with VR menus, while completely disabling the in-window menus. The main window will only mirror what the VR user sees. Since we don't have VR menus yet, this gives us a way to get the game to the state we want before VR begins.

engine/src/main/java/org/terasology/config/RenderingConfig.java
emanuele3d: That would be called a DoF effect: https://en.wikipedia.org/wiki/Depth_of_field

GooeyHub:

GooeyHub:

indianajohn:
Alright, I took out the merge commit that got sloppily added to this branch, as well as a lot of small 'oops' commits. I think the commit history reflects the history of the pull request. The other option is that I can just squash everything into a single commit. (I'm in favor of leaving in the history but everybody is different).

GooeyHub:

emanuele3d:
Nononono! We like separate commits for separate tasks! Squash commits as much as reasonable but no more! :smiley:

Cervator:
Hooray for selectively squashed but useful commits :-)

> Now, when this happens, the game starts as normal with the log message:

That's probably better than leaving the game failing to start with no obvious indication as of why unless the user checks the log. Although with the CrashReporter enabled that may be one way to highlight it. I wish we had a better way. #1402 covers that a bit but certainly is outside the scope of this PR. I guess the user will notice when VR doesn't work :-)

GooeyHub:

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
```diff
@@ -0,0 +1,100 @@
...
+            lastControllerState[c] = new VRControllerState_t();
+            controllerPose[c] = new Matrix4f(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1);
+            eyePoses[c] = new Matrix4f(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1);
+            projectionMatrices[c] = new Matrix4f(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1);
+
+            for (int i = 0; i < 5; i++) {
+                lastControllerState[c].rAxis[i] = new VRControllerAxis_t();
+            }
+        }
+
```
emanuele3d: Unnecessary empty line.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
```diff
@@ -0,0 +1,100 @@
...
+    private List<ControllerListener> controllerListeners = new ArrayList<>();
+
+    // In the head frame
+    private Matrix4f[] eyePoses = new Matrix4f[2];
+    private Matrix4f[] projectionMatrices = new Matrix4f[2];
+
+    // In the tracking system intertial frame
+    private Matrix4f headPose = new Matrix4f(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1);
+
+
```
emanuele3d: Unnecessary empty line.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
```diff
@@ -0,0 +1,100 @@
...
+    private static VRControllerState_t[] lastControllerState = new VRControllerState_t[2];
+
+    private List<ControllerListener> controllerListeners = new ArrayList<>();
+
+    // In the head frame
+    private Matrix4f[] eyePoses = new Matrix4f[2];
+    private Matrix4f[] projectionMatrices = new Matrix4f[2];
+
+    // In the tracking system intertial frame
+    private Matrix4f headPose = new Matrix4f(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1);
```
emanuele3d: My eyes are crossing with all these 1s and 0s. How about you add a method to MatrixUtils:

`public static Matrix4f createIdentityMatrix4f() {return new Matrix4f(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1)}`

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -158,11 +161,17 @@ public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
...
-            currentRenderingStage = RenderingStage.LEFT_EYE;
-
+        if (renderingConfig.isVrSupport()) {
+            this.vrProvider = new OpenVRProvider();
+            if (this.vrProvider.init()) {
+                playerCamera = new OpenVRStereoCamera(this.vrProvider);
+                currentRenderingStage = RenderingStage.LEFT_EYE;
+                context.put(OpenVRProvider.class, vrProvider);
+            }
+            else {
```
emanuele3d: This should be `} else {`.

Also, in line with previous comments, if isVrSupport() returns true and this.vrProvider.init() doesn't work, returning false, some kind of logger error/warning would help, i.e. stating that MONO mode is enabled despite vrSupport being ON. Or is this logged elsewhere?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
```diff
@@ -0,0 +1,62 @@
...
+        requiresCondition(() -> worldRenderer.getCurrentRenderStage() == MONO || worldRenderer.getCurrentRenderStage() == LEFT_EYE);
+        addDesiredStateChange(new BindFBO(DEFAULT_FRAME_BUFFER_URN,displayResolutionDependentFBOs));
+        addDesiredStateChange(new EnableMaterial("engine:prog.defaultTextured"));
+    }
+
+    @Override
+    public void process() {
+        PerformanceMonitor.startActivity("rendering/copyImageToScreen");
+        FINAL.bindTexture();
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
```
emanuele3d: Most nodes currently have a glClear call but it is my intention to verify if they are strictly necessary or if they could clear perhaps only the color or only the depth buffer. My reasoning is that in some cases all the pixels are getting rewritten anyway and clearing the output buffers to a default value is unnecessary.

I will be able to verify this on all standard nodes, but would you mind checking what impact would have removing the call or perhaps only clearing one of the two buffers in the VR-related nodes? Right now I'm not even sure the default frame buffer has a depth buffer.

emanuele3d:
I just realized that one of the issues that kept cropping up while reviewing this PR is the error handling overall strategy. Now that most of the functional code is in place you might want to focus on it from a broader perspective. One way to think about it is this:

* If something fails on initialization, make sure the user is aware of it (potentially just through logger output), give up and proceed in mono mode.
* If initialization worked, assume that all required objects are there - do not nullcheck them unless strictly necessary.

engine/src/main/java/org/terasology/rendering/cameras/OpenVRStereoCamera.java
```diff
@@ -206,49 +211,68 @@ public void updateMatrices() {
...
     public void updateMatrices(float fov) {
         prevViewProjectionMatrix.set(viewProjectionMatrix);
 
-        // Nothing to do...
-        if (cachedPosition.equals(getPosition()) && cachedViewigDirection.equals(getViewingDirection())
-                && cachedZFar == zFar && cachedZNear == zNear) {
+        org.joml.Matrix4f leftEyeProjection = vrProvider.vrState.getEyeProjectionMatrix(0);
+        org.joml.Matrix4f rightEyeProjection = vrProvider.vrState.getEyeProjectionMatrix(1);
+        org.joml.Matrix4f leftEyePose = vrProvider.vrState.getEyePose(0);
+        org.joml.Matrix4f rightEyePose = vrProvider.vrState.getEyePose(1);
```
Cervator: Looks like we've got a `Matrix4f` in TeraMath as well. Maybe just an ambiguous import and the IDE picked `joml` by chance?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,407 @@
...
+
+        controllerDeviceIndex[RIGHT_CONTROLLER] =
+                vrSystem.GetTrackedDeviceIndexForControllerRole.apply(
+                        JOpenVRLibrary.ETrackedControllerRole.ETrackedControllerRole_TrackedControllerRole_LeftHand);
+        controllerDeviceIndex[LEFT_CONTROLLER] =
+                vrSystem.GetTrackedDeviceIndexForControllerRole.apply(
+                        JOpenVRLibrary.ETrackedControllerRole.ETrackedControllerRole_TrackedControllerRole_RightHand);
+    }
+
+    //jrbuda:: oh hello there you are.
```
Cervator: The wisdom of Buda Jr speaks ages! :D

Cervator:
Bump! Went to finally try testing this due to a convenient eyeglasses malfunction leading to me wearing contacts making using a headset somewhat easier.

Could grab the code, redo IntelliJ, see the VR enabled setting, and run the game normally as well as with SteamVR running with the headset on. *But* I had a somewhat older (but still past the v1 release) Oculus Runtime, and the SteamVR thing was complaining about not being able to run its vrCompositor or something. It *did* detect the Oculus just fine after I made it reboot the headset once.

On attempting to run the game in that scenario I got the following during startup - no luck, as somewhat expected - but at least it ran!

```
...
23:20:17.693 [main] INFO  o.t.w.b.internal.BlockManagerImpl - Registered Block engine:unloaded with id 9
23:20:17.883 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - Adding OpenVR search path: J:\Dev\Terasology\Git\integrate\Terasology\openvr_natives\win32-x86-64
23:20:18.121 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - VR Headset detected.
23:20:18.271 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - OpenVR initialized & VR connected.
23:20:22.051 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Invalid Connect Response (307)
23:20:22.354 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:22.657 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:22.960 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:23.263 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:23.566 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:23.871 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:25.218 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:25.521 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:27.034 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrCompositor initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:27.123 [main] INFO  o.t.r.openvrprovider.OpenVRProvider - vrOverlay initialization failed:Shared IPC Compositor Connect Failed (306)
23:20:27.900 [main] INFO  o.t.r.dag.RenderTaskListGenerator - -- Intermediate List --
23:20:27.900 [main] INFO  o.t.r.dag.RenderTaskListGenerator -               BindFBO: engine:sceneShadowMap
...
```

I then went and grabbed the very latest Oculus software and ... holy aggravation, batman! I am definitely not buying any newer Oculus stuff. I guess the big FB-fication triggered since last I updated, so this time I got the whole "Connect errything to FB! Is awesomesauce!" prompt along with the wildly open privacy settings and other "optionals". Couldn't even pick the installer path, had to use a command line argument to at least get it on the right drive.

Made it a few steps into the installation progress before it refused to continue without first updating my video card drivers. Which with two older mid-range Radeon cards on an old system ... ugh :( That tends to be a whole day exercise and I already did that last time I updated. Exited the installer and had nothing. SteamVR couldn't connect to anything anymore.

So maybe I'll have to try again later, but it may be less hassle just getting a whole new system already (probably happening within a month or two). Will see.

@indianajohn hi! Hope all is well :-)

indianajohn:
Work got a bit busy over the past few weeks and kept me from putting more time into this. I think I'll be able to get to it tonight and this weekend.

engine/src/main/java/org/terasology/rendering/cameras/OpenVRStereoCamera.java
indianajohn: JOML is needed here because VRProvider uses JOML matrices as outputs to this function. I initially wrote VRProvider to be a separate module and pulled it in. The upside is that VRProvider can easily be plucked out of Terasology and put elsewhere (and changes to VRProvider in Terasology can be brought out to the general repository from which it came, and be useful to others). I'm willing to compromise here if you guys insist.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToHMDNode.java
```diff
@@ -0,0 +1,120 @@
...
+    public void initialise() {
+        renderingConfig = config.getRendering();
+        requiresCondition(() -> (renderingConfig.isVrSupport()
+                && vrProvider != null
+                && vrProvider.isInitialized() ));
+        leftEye = requiresFBO(new FBOConfig(LEFT_EYE_FBO, FULL_SCALE,
+                FBO.Type.DEFAULT).useDepthBuffer(),displayResolutionDependentFBOs);
+        rightEye = requiresFBO(new FBOConfig(RIGHT_EYE_FBO,FULL_SCALE,
+                FBO.Type.DEFAULT).useDepthBuffer(),displayResolutionDependentFBOs);
+        if (this.vrProvider != null) {
```
indianajohn: vrProvider can only be null if isVrSupport() turns out false, I guess. These checks ended up in here because, although the name would suggest that we'll never end up hitting this code in certain conditions, the system was a bit unfamiliar to me so I wrote some extra checks. I can respect the desire for brevity so I'll remove them in the upcoming commit.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToHMDNode.java
```diff
@@ -0,0 +1,120 @@
...
+        PerformanceMonitor.endActivity();
+    }
+
+    private void renderFinalStereoImage(RenderingStage renderingStage) {
+        switch (renderingStage) {
+            case LEFT_EYE:
+                vrProvider.updateState();
+                leftEye.bind();
+                glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+                renderFullscreenQuad();
```
indianajohn: I'm going to hold off on this for the moment because of the following:

http://stackoverflow.com/questions/3073796/how-to-use-glcopyimage2d

I read this some time back and got the impression that rendering through a passthrough shader to the screen was actually faster than glCopyTexImage2D. I don't know if that's true, but the more concerning part is this:

"Second, as you’ve already observed, glCopyTexImage2D requires its width and height arguments to be power-of-two as well."

So that seems like a non-starter.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
```diff
@@ -0,0 +1,62 @@
...
+        renderFullscreenQuad();
```
indianajohn: See above. :)

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,407 @@
...
+    private final VRTextureBounds_t texBounds = new VRTextureBounds_t();
+    private float nearClip = 0.5f;
+    private float farClip = 500.0f;
+
+    public static Texture_t texType[] = new Texture_t[2];
+
+    public OpenVRState vrState = new OpenVRState();
+
+    public OpenVRProvider() {
+        for (int c = 0; c < 2; c++) {
```
indianajohn: Done.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,407 @@
...
+            texType[c] = new Texture_t();
+        }
+    }
+
+    public boolean init() {
+        if (!initializeOpenVRLibrary()) {
+            return false;
+        }
+        if (!initializeJOpenVR())
+            return false;
```
indianajohn: Done.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,407 @@
...
+        int nAttempts = 0;
+        boolean bSuccess = false;
```
indianajohn: Done. My other project uses Hungarian notation and that habit dies hard.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,407 @@
...
+        initialized = true;
+        return true;
+    }
+
+    public boolean isInitialized() {
+        return initialized;
+    }
+
+    public void shutdown() {
+        initialized = false;
```
indianajohn: So, this has never actually caused a problem with program flow, but I guess I should explicitly plan for VRProvider to get deleted and re-initialized later. I'm going add VR_Shutdown() here and clean things up, and test that there are no problems with it.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,407 @@
...
+            }
+            vrState.setHeadPose(hmdTrackedDevicePoses[JOpenVRLibrary.k_unTrackedDeviceIndex_Hmd].mDeviceToAbsoluteTracking);
+            headIsTracking = true;
+        } else {
+            headIsTracking = false;
+        }
+
+        findControllerDevices();
+
+        for (int c = 0; c < 2; c++) {
```
indianajohn: Changed to handIndex.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
indianajohn: Done, but in OpenVRUtil (due to it being a JOML type).

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
indianajohn: Fixed the formatting error.

if vrProvider fails to init, there is no scenario at this point where the log isn't blasted with errors, so I think we're safe to not add something here.

GooeyHub:

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
indianajohn: Removed.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
indianajohn: Removed.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -0,0 +1,407 @@
...
+        vrSystem.TriggerHapticPulse.apply(controllerDeviceIndex[controller], 0, (short) strength);
+    }
+
+    public void submitFrame() {
+        if (vrCompositor == null) {
+            return;
+        }
+        if (vrCompositor.Submit == null) {
+            return;
+        }
```
indianajohn: Removed these.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
indianajohn: I took this out, and it doesn't seem to have any ill effect.

GooeyHub:

GooeyHub:

GooeyHub:

indianajohn:
Okay, so I've addressed all of the comments above, and I've also tested with/without VR. All of the test cases I did before seem to work (system with headset, VR support turned on / system without headset, VR support turned off / system with headset, VR support turned of / system without headset, VR support turned off). I'm ready for whatever comes next.

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/cameras/OpenVRStereoCamera.java
emanuele3d: I'm not sure I understood your comment here @indianajohn. Are you saying you can move the VRProvider in a module and consequentially remove JOML from Terasology? How much of the code you have written would relocate into the module alongside the VRProvider?

In any case I feel this can be left to a separate PR and we can leave things as they are for the time being, so that this PR just doesn't stay open for too long. But if my understanding is correct I'd really like for VR code to exist in a module rather than directly in the engine.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
emanuele3d: Sorry. Got my eastern european countries mixed up, my apologies. :stuck_out_tongue_closed_eyes:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: So, do you expect the application to crash in those circumstances?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -106,6 +113,12 @@ public boolean isInitialized() {
     }
 
     public void shutdown() {
+        JOpenVRLibrary.VR_ShutdownInternal();
```
emanuele3d: I forgot: who is managing the FBOs VR uses? Who disposes them? One of the existing FBO managers or the VR code? If so, should FBO disposal happen here?

emanuele3d:
By the way, since my previous feedback one of my PR has gone through and it has changed the code in WorldRendererImpl.initRenderGraph() generating a conflict with your changes. You might want to grab the latest develop and make the required updates to your branch.

indianajohn:
I just added the TODO that was @emanuele3d requested and merged in the develop branch. It works on my Macbook, but I have to test it on my headset when I get home. I'll update this PR when it's all tested.

GooeyHub:

indianajohn:
Alright, I tested this on the Vive and it seems to work. I think we're ready to roll.

Something we'll definitely want to consider soonish is to refactor such that the playerCamera is part of the player. Then we can have OpenVRStereoCamera override the getPosition() method to include whatever has been stored on the camera compounded with the transform that OpenVR provides, and the player will be seen to walk around in a multiplayer game as he would be seen walking around his room. It would also ensure that held torches emit light from the position of the VR player, and that tools, if they have some sort of usable radius from the current player position, would be evaluated against the player position apparent to the HMD wearer.

GooeyHub:

indianajohn:
Ah, I should also mention that, due to the player camera position not being updated per-eye and upon player movement, shadows are a bit strange per-eye and as the player moves through the volume. I considered overriding playerCamera.position() to follow the position of the player as they move through the volume, but I was afraid of unintended consequences so I left off for now. I think I should leave that for the camera refactor I mentioned earlier.

emanuele3d:
@Cervator: I confirm this PR is good to go from my perspective.

@indianajohn: First of all, thank you for your work. You have done a lot. I look forward enjoying these changes and those that will be derived from them when I eventually buy a VR set. 

Second, cameras are meant to become components that can be easily attached/detached to an entity. The component itself would hold very few, simple pieces of data, i.e. focal length, far/near planes distances and so on. Things like the camera position would be stored in a location component instead (which already exists). A CameraSystem would probably hold at least part of current cameras functional code. This is hard work because the player camera is involved in some many areas of the codebase. I can't find right now an issue on the topic or a discussion about this in the forum but I thought there were some. @Cervator? Long story short, if you feel like looking into this you'd be more than welcome.

Finally, I'm not sure I understand this:
> (...) due to the player camera position not being updated per-eye and upon player movement, shadows are a bit strange per-eye and as the player moves through the volume.

The shadow map is updated when the player moves. In fact it is updated once every frame whether the player moves or not. It is not updated per eye because shadows are dependent only on the position of the light source, the objects in the world and the player position as the shadowmap is centered around the player. Are you perhaps talking about light shafts?

indianajohn:
> The shadow map is updated when the player moves. In fact it is updated once every frame whether the player moves or not. It is not updated per eye because shadows are dependent only on the position of the light source, the objects in the world and the player position as the shadowmap is centered around the player. Are you perhaps talking about light shafts?

Sorry, what I meant by 'player' in this case was the person wearing the HMD.

As I understand it, playerCamera.getPosition() returns the position to which the player has moved using the controls. Then, in ShadowMapNode.java we have this line of code, which determines the light direction:

```
 Vector3f lightPosition = new Vector3f(playerCamera.getPosition().x, 0.0f, playerCamera.getPosition().z);
```

When rendering a client who isn't using room-scale VR, this matches the camera position. However, when somebody is using room-scale VR, the user head position is actually the sum of this position and the translation component of the head pose.

I haven't fully dug into the implications of this (as I said, I think it belongs in another pull request that addresses camera refactoring), but I have noticed that, if I fly really high (and thus increase the distance between my player and whatever shadow is being cast), rotation and head movement cause distant shadows to swim a bit. I guess it has something to do with this.

engine/src/main/java/org/terasology/rendering/cameras/OpenVRStereoCamera.java
indianajohn: Yeah, I think it's best left off until we have mod support for rendering code, but peeling some of this off into another module would be the best plan for the future.

indianajohn:
P.S. thanks for the feedback. I'm glad we managed to do a proper integration of the code with the rendering engine rather than pushing it in half-baked.

Cervator:
Thanks again both :-)

I'm trying to work up the courage and time to survive the graphic card driver update I'd need to be able to use my Oculus DK2 again. With only two days left until I leave for the GSOC Mentor Summit (Wed) it might be unlikely to happen until after I get back (late Nov 1st). I'm not sure who else might both have an HMD and be able to test.

Still, so long as we can validate the base game works fine without VR really that's all we need. PR is approved and even if it doesn't work for Oculus or OSVR (@tdgunes - ping! Did you get it yet?) that doesn't really matter right this moment since that's new functionality anyway.

I'd aim to release Alpha 6 by the next multiplayer session on Nov 5th, probably we should aim to have this included by then if not sooner. Hope that sounds reasonable!

emanuele3d:
@indianajohn: thank you for letting us know of the issue you are experiencing with the shadows. My recommendation is that you write an issue about it unless you want to start digging into it in the short term.

@Cervator: that's ok for me to wait for the merge. Meanwhile I'll post a new node-refactoring PR that has been waiting in the wings. There will be some conflicts in the WorldRendererImpl.java class as we both modified initRenderGraph() but they'll be easy to solve.

tdgunes:
Sorry @Cervator, it's in Turkey right now. 😞 I hoped to received it before going to UK. Fortunately  my friend who is going to visit me is also going to bring it to me at the end of this month. 🎉 Then, I'll be looking into this PR (whether it works with OSVR or it requires tweaking).

GooeyHub:

emanuele3d:
Wow. I didn't realize this PR is still active: I thought it went through!

I recommend merging #2595 first so that only @indianajohn will have to solve the conflicts in WorldRendererImpl. Otherwise he will have to solve his conflicts and then I'll have to solve mine.

indianajohn:
I'm fine with that. I'll update this branch when that one gets merged, and signal that I'm ready.

Cervator:
#2595 merged so ready for quick conflict resolution here then merge :-)

Small update from me: Probably will be ordering a new system in about two weeks (first full system in near 10 years). Might wait for that until I try any VR testing. Although probably I should still update my current system so it can be another testing platform (Radeon + Win7 vs Nvidia + Win10)

emanuele3d:
Speaking of VR testing, I had the chance to try VR for the very first time -ever- in a local computer shop - incidentally on a Vive. I must say it was awesomer than I thought it would be. The demo application put me on top of a tower and I used the controllers to shoot arrows at an "horde" of cute little characters wanting to escape after having raided my castle. The image was rock-steady as I turned my head around and I was surprised by the field of view as I expected a bit more of a tunnel vision.

The box in which I was allowed to move (limited only by a neat visual cue when getting too close to the limit) coincided with the extent of the tower and it was quite neat to move around freely. On the other hand I am curious to try applications such as driving or flying as apparently they can make you dizzy. Did you have the chance to try that @indianajohn?

Bottom line: really cool.

GooeyHub:

indianajohn:
I merged and tested this again; seems to check out. It's ready to merge as far as I'm concerned.

Re: VR sickness, it's not just dizziness, it's motion sickness. I have a pretty strong stomach - I have never been carsick or seasick in my life - and it's a mixed bag for me. As long as you have a point of reference that is fixed in the real world and VR (such as the interior of a car or airplane), the problem is mitigated a lot. But I'm playing through Obduction now, and after a half hour or so with free movement turned on I have to take a break. There's a reason that most games use teleportation-based locomotion, despite the mechanic being extremely immersion-breaking.

The Vive is pretty amazing, but there's not a lot to do with it quite yet. Hopefully we can do our part to help alleviate that problem. :) Our first task will be to develop some sort of interesting mechanic, such as a bow and arrow or gun of some sort.

Cervator:
Gave it one more round of non-VR tested and finally merged! :D :tada: :tada: :tada: :tada: :tada: :tada: 

Thank you again so much for all this effort @indianajohn and @emanuele3d for all the review! I especially appreciate sticking with this process so long (active review and tweaking for just past two months!)

I ran it without the VR flag and everything was normal as expected. I then tried to put my broken setup to use by seeing what would happen if I *did* enable VR in `config.cfg` and again nothing happened, even with Steam VR running and my headset on (which Steam VR then promptly fails to find due to driver stuff). Just got some simple logging about no headset being detected so not loading VR mode anyway.

Very happy to finally have this merged, looking forward to more tweaking at VR in the future. I'm submitting a small follow-up item mostly for longer term reminders and possible extraction of the provider one day, maybe after we switch to JOML, upgrade to LWJGL 3, LibGDX, or who knows. No rush items in there but probably good to check out just to make sure I didn't forget anything (since this PR getting closed will make any follow-ups less visible)

bebra777261 (approved):

PR #2480, opened 2016-09-12
Fix null field deserialization crash + further possible crashes

flo:
### Contains

* Fixes a remote client only NullPoinerException crash that happens when you clear a "wrong placement hint" of a structure template. The crash was caused by wrong deserialization of a `Region3i` field that got set to null.
* Fixes further potential crashes caused by the deserialization of `Rect2f`, `Rect2i` and `BorderTypeHandler` fields.

### How to test
1. Start a server with the Gooey's Quests module enabled
2. Connect to the server as client
3. Get yourself a structure template with placement conditions. e.g. via the command `giveItem dungeonEntranceViaMountain`.
4. Try to place it at an invalid location (where the placement preview shows a red region)
5. You should see an error message dialog. Confirm the dialog
6. A red region should still be visible. Right click once to clear it. There should be no crash.


GooeyHub:

Cervator:
Fix confirmed and merged. Thanks!

Issue #2474, opened 2016-09-11
Automatically spawned chests start empty in multiplayer, can fix by breaking then placing again


Cervator:
Scenario to replicate:

* Start GooeysQuests with a local headless server
* Connect with a client
* Get Gooey to spawn
* `e` click Gooey then select spawn a dungeon
* Find a chest, it'll be empty. 
* Same thing in single player works fine.

Can fix it by breaking the empty chest then placing it again. Suddenly it'll have its intended contents.

The following highly suspicious log statement appears during this scenario:

`22:20:49.833 [main] ERROR o.t.e.e.internal.PojoEntityManager - Adding a component (class org.terasology.logic.inventory.InventoryComponent) over an existing component for entity 154`

Similar / related issues to perhaps also look into at the same time:

* #1693
* #1690

flo:
This was a bug in the StructureTemplate module. I  fixed it. The problem was indeed the addition of the `InventoryComponent`. I added a forum post about the fix.

Cervator:
Thanks @flo - renamed the issue accordingly. Appreciated :-)

PR #2481, opened 2016-09-12
Discussion: Make setBlocks update entities like setBlock - at huge performance cost !!!

flo:
### Contains

This fixes an NullPointerEException when the following steps are done:
1. A chest gets placed
2. The chest gets overridden with air by a setBlocks call
3. A stone block gets placed at the location of the chest.

The exception had the following stacktrace:

```
java.lang.NullPointerException
    at org.terasology.world.internal.EntityAwareWorldProvider.updateBlockEntityComponents(EntityAwareWorldProvider.java:235)
    at org.terasology.world.internal.EntityAwareWorldProvider.updateBlockEntity(EntityAwareWorldProvider.java:151)
    at org.terasology.world.internal.EntityAwareWorldProvider.setBlock(EntityAwareWorldProvider.java:122)
    at org.terasology.world.internal.WorldProviderWrapper.setBlock(WorldProviderWrapper.java:52)
    at org.terasology.network.internal.ServerImpl.processBlockChanges(ServerImpl.java:304)
    at org.terasology.network.internal.ServerImpl.processMessages(ServerImpl.java:269)
    at org.terasology.network.internal.NetworkSystemImpl.update(NetworkSystemImpl.java:317)
    at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:406)
    at org.terasology.engine.Terasology.main(Terasology.java:152)
```

The crash happened for the clients only, as they knew temporary block entities without any component. I haven't fully investigated how they get to such an entity, but making setBlocks do the same block entity updating like setBlock fixed the issue.

However it comes at a huge performance cost, now setBlocks is even
slightly worse in performance then setBlock:

```
benchmark: WorldProvider.setBlock
iteration:  200 / 200
last duration:  945.6 ms
min duration:  822.1 ms
median duration:  912.6 ms
avg duration:  1073.4 ms
max duration:  4839.3 ms
```

```
benchmark: WorldProvider.setBlocks
iteration:  200 / 200
last duration:  5471.2 ms
min duration:  627.1 ms
median duration:  950.6 ms
avg duration:  1089.6 ms
max duration:  5471.2 ms
```

Without this PR, the placement performance is much better:

```
benchmark: WorldProvider.setBlocks
iteration:  200 / 200
last duration:  66.3 ms
min duration:  36.9 ms
median duration:  60.8 ms
avg duration:  73.7 ms
max duration:  1883.9 ms
```
### How to test
1. Start a headless server with the Gooey's Quest module.
2. Join the server as a client with cheat permissions (or get those permissions via permission key)
3. Cheat yourself a dungeonTownInn item
4. place it on a block
5. place it on a second block that is 1 block closer to you

Alternativlty you can also create a structure template that replaces a block with air and then use that template on a chest and then place afterwards a solid block where the chest has been.

I think we should discuss how much of the "a block must behave like an entity" behavior we actually need. Currently the block entities have special rulss in some places where they don't belong (e.g. entity manager) and block placement results in a lot of overhead.

Currently basically placing a block means that the entity that represents the block will be updated to have the new components. if there is no such entity, an entity will be silently be created. Then the entity will be updated, and when it is not an "force active" block then the entity will be silently be destroyed at  the next update cycle.

Changes to be discussed:
1.) Reintroduce life cycle events for block entities, to make them just normal entities without special rules. Reasoning: Remove special handling currently present in entity manager where it does not belong in my opinion. It also fixes a potential memory leak: Systems that watch for component activation/deactivation could then properly track block entities. Currently a system might detect the addition of a component to an entity and remembers the entity for further processing. When that entity gets silently removed the system will still remember that entity.
2.) OnChangedBlock event could be sent to the world entity, instead of the temporary block entities. That way we don't need to create block entities just for the events. When a block change happens at the location of an entity that entity could of course still be updated.
3.) Maybe we could turn  EntityAwareWorldProvider into a normal system, that just uses activation/deactivation of BlockComponents to track block entities. It could store the entities in a map from location to entity and when a OnchangedBlock event gets received on the world entity it could notify the entity at the given location.
4.) Remove the automatic creation, update and deletion of block entities. Block entities could be created/updated manually from whatever system needs it. e.g. the system of a item entity could create the block entity when a "force active" block gets placed. A low prio event handler could delete entities with BlockComponents when they receives a OnChangedBlock event. 

Pinging @immortius, @skaldarnar , @MarcinSc, @Josharias  as they might be interested in joining the discussion.

### DIFF
--- a/engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
+++ b/engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
@@ -127,6 +127,23 @@ public class EntityAwareWorldProvider extends AbstractWorldProviderDecorator imp
     }
 
     @Override
+    public Map<Vector3i, Block> setBlocks(Map<Vector3i, Block> blocks) {
+        if (GameThread.isCurrentThread()) {
+            Map<Vector3i, Block> oldTypes = super.setBlocks(blocks);
+            for (Map.Entry<Vector3i, Block> entry: oldTypes.entrySet()) {
+                Vector3i pos = entry.getKey();
+                Block oldType = entry.getValue();
+                Block newType = blocks.get(pos);
+                EntityRef blockEntity = getBlockEntityAt(pos);
+                if (oldType != null) {
+                    updateBlockEntity(blockEntity, pos, oldType, newType, false, Collections.<Class<? extends Component>>emptySet());
+                }
+            }
+        }
+        return null;
+    }
+
+    @Override
     @SafeVarargs
     public final Block setBlockRetainComponent(Vector3i pos, Block type, Class<? extends Component>... components) {
         if (GameThread.isCurrentThread()) {


GooeyHub:

engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
```diff
@@ -127,6 +127,23 @@ public Block setBlock(Vector3i pos, Block type) {
     }
 
     @Override
+    public Map<Vector3i, Block> setBlocks(Map<Vector3i, Block> blocks) {
+        if (GameThread.isCurrentThread()) {
+            Map<Vector3i, Block> oldTypes = super.setBlocks(blocks);
```
skaldarnar: Here, the old/replaced blocks are required to update the entites. With #2467 we argued that that `setBlock` should return info on success/failure, but not more. Turning `EntityAwareWorldProvider` into a normal system listening for de-/activation of block components would be compliant to the new interface.

skaldarnar:
I suppose the life cycle events for blocks were removed/not added due to performance reasons - would be worth another benchmark to estimate the impact. I like the idea of an `OnChangedBlockEvent`, it's quite common to act on block changes, having a proper event can simplify working with it. 
Point (3) sounds like a reasonable change as it clearifies the purpose of `EntityAwareWorldProvider` imho.

I will test the changes as soon as I can find the time for it. :wink:

Cervator:
How exactly does this relate to #2474 (empty chest - but even with a single GQ dungeon, no overlapping needed) or #2361 (same exact error message, but at least a couple different scenarios) ?

Although I noticed https://github.com/Terasology/StructureTemplates/commit/24e806062b900fae3f0950d6799e4863bc3d309e - does that on its own fix #2474 ? :-) If it is/was a StructureTemplates issue instead of engine that makes sense.

flo:
About 3.) I found out that it is actually already keeping track of the block entities via component activation/deactivation. The temporary block entities are however separate. 

I really wonder if there is a nice minimalistic approach to get the performance we want.

Maybe we could simply skip the creation of the temporary block entities when a block gets placed. This would mean that of course there are no events that fake the conversion of a block entity from one type to another. It destroys a bit of the illusion that all blocks are entities. However that illusion isn't currently complete either, as you can't for example find a block via component when it has currently no temporary block entity.

Cervator:
Bump - @flo and @skaldarnar what do you think we should do with this PR and #2467?

flo:
I would suggest we keep this PR for a while. I still want to fix it at some point.

In case someone else wants to do it in the meantime:

It might be enough to limit the entity creation to entities that stay active.

GooeyHub:

flo:
Closing it now as the bug no longer seems reproducable.

Cervator:
Huh, it went away on its own? Am thrilled to have those closed but puzzled on what happened :-)

Any risk it may take part in other odd issues like #2516? I wouldn't want us to walk away from a potential bug/fix from it just having sneakily hid its role in other problems.

Issue #2473, opened 2016-09-10
Hardening of some sort needed in DelayedActionSystem (plant growth can cause NPE crashes?)


Cervator:
During multiplayer event the server crashed a minute after @rzats had himself crashed out via https://github.com/Terasology/JoshariasSurvival/issues/14 shortly after I crashed out via #2361

I suspect something about the blueberry bush growth stages from JoshariasSurvival, which is based off SimpleFarming which in turn uses the DelayManager system from the engine. If I were to wager a guess there was an immature blueberry bush in a chunk one of us had been in that had its growth *tick* in a way that was no longer valid with no players left on the server.

Crash happened in `DelayedActionSystem` and while the log doesn't identify exactly what the scheduled action was I think a fair bet would be a blueberry bush. That in turn is defined in `BlueberrySeed.prefab` from JoshariasSurvival, which has a `PlantDefinitionComponent` that still has a bit of logic and other fanciness despite being a Component (old code I suspect). The related logic system is `FarmingAuthoritySystem` which is what in turn uses the DelayManager.

So some things to consider looking at:

- [ ] Find out what causes the (likely) blueberry growth to result in an invalid scheduled action (something goes stale as player(s) disappear?)
- [ ] Harden `DelayedActionSystem` to where it has error handling for encountering invalid scheduled actions (expect that invalid tasks could be created somehow)
- [ ] Log some better details when a bad action is encounter (which one is it? Where did it come from?)
- [ ] Look at SimpleFarming's use of delayed actions (probably predates the refactoring @xrtariq2594 applied to the delay system and wouldn't have been an obvious thing to test)
- [ ] Simplify `PlantDefinitionComponent` also from SimpleFarming to follow code guidelines (no logic in Components, keep them basic)
- [ ] Bonus: Add some more documentation to the scheduled task system, maybe some basic examples or even a little tutorial somewhere?

Error log below:

```
14:01:59.002 [main] WARN  o.t.l.c.ServerCharacterPredictionSystem - Received too much input from EntityRef{id = 9050, netId = 20052, prefab = 'engine:player'}, dropping input.
14:01:59.007 [main] WARN  o.t.l.c.ServerCharacterPredictionSystem - Received too much input from EntityRef{id = 9050, netId = 20052, prefab = 'engine:player'}, dropping input.
[...]

PR #2484, opened 2016-09-12
Added null checks to reduce NPEs.

xrtariq2594:
### Contains

Fixes item 2 of issue #2473.

### How to test

Use systems that utilize the delay system. See if they still function as expected.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/delay/DelayedActionSystem.java
+++ b/engine/src/main/java/org/terasology/logic/delay/DelayedActionSystem.java
@@ -75,15 +75,18 @@ public class DelayedActionSystem extends BaseComponentSystem implements UpdateSu
         operationsToInvoke.stream().filter(EntityRef::exists).forEach(delayedEntity -> {
             final DelayedActionComponent delayedActions = delayedEntity.getComponent(DelayedActionComponent.class);
 
-            final Set<String> actionIds = delayedActions.removeActionsUpTo(currentWorldTime);
-            saveOrRemoveComponent(delayedEntity, delayedActions);
+            // If there is a DelayedActionComponent, proceed.
+            if (delayedActions != null) {
+                final Set<String> actionIds = delayedActions.removeActionsUpTo(currentWorldTime);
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Looks good, but maybe add some else branches to help detect when something is wrong? Hopefully there'll be something available to log (if appropriate), but if not even just knowing that something was null or empty might help figure out "Hey, with this combination of modules and some action something isn't quite right"

xrtariq2594:
Gotcha.

Do you want me to add a logger to the DelayActionSystem?

engine/src/main/java/org/terasology/logic/delay/DelayedActionSystem.java
```diff
@@ -104,15 +107,18 @@ private void invokePeriodicOperations(long currentWorldTime) {
         operationsToInvoke.stream().filter(EntityRef::exists).forEach(periodicEntity -> {
             final PeriodicActionComponent periodicActionComponent = periodicEntity.getComponent(PeriodicActionComponent.class);
 
-            final Set<String> actionIds = periodicActionComponent.getTriggeredActionsAndReschedule(currentWorldTime);
-            saveOrRemoveComponent(periodicEntity, periodicActionComponent);
+            // If there is a PeriodicActionComponent, proceed.
```
flo: I would go for a guard statement instead of indenting everything by 1. (e.g. 
if (isError) {
    return;
}
normalHandling();

I am also not familiar with this code, but I think there should at least be a comment why null is a common case and not an error. If it is an error that is worth investigating I suggest you change the code to log an error.

xrtariq2594: On it. Thanks flo.

GooeyHub:

Cervator:
Tested it again and ended up reliably reproducing #2361 which I suspect is unrelated to this PR. Merged this and pushed latest SimpleFarming develop to its master branch. Thanks for the updates :-)

It looks easier to attack the crash bug with blueberries now. Instructions in the linked issue if you're curious.

Issue #2387, opened 2016-06-30
General review of OpenGL related logging


Cervator:
Follow-up to #2368 - thought I'd leave this as a reminder for @emanuele3d or @tdgunes to check out sometime later when time permits :-)

It might be that no change is needed, but as noted in the PR it seems like the shader logging directory (`logs/shaders`) is now empty, while it used to get a bunch of debug info per shader once upon a time in the past.

I don't know what an ideal future setup is, but I bet our 3d wizards do! And if present state turns out to be fine, then that's fine too. If the shader logging isn't useful (and gone) we probably shouldn't create its log dir though.

May also be worth considering log levels, and double checking on whether `trace` makes sense with Logback / can actually be enabled or not.

PR #2487, opened 2016-09-20
Render Task List Generator refactor / Improved Logger Output

emanuele3d:
@tdgunes: you might want to have a look at this one.
@indianajohn: it might be interesting for you too, to increase your knowledge of the system.

### Contains
* Considerably refactored the RenderTaskListGenerator class, responsible for taking the nodes in the render graph (currently just a list) and generating the list of tasks that are executed every frame to generate the image shown on screen. Also added some javadocs.
* Changed toString() methods in a number of classes to generate a better logger output, with aligned text fields separated by colons.
* Added some statistics printout at the end of the tasklist generation process.

### How to test
* Load up a game and then check the logger output for two lists: the Intermediate Renderer List first and the Renderer Task List next. The first lists all potential opengl state changes caused by each node. The second lists tasks reflecting the actual state changes taking place every frame, some state changes having been eliminated because redundant with previous or following state changes. 
* Changing the video settings in some cases remove nodes from the lists - a feature, not a bug. However there is no simple relationship between video settings and nodes as some settings enable shader features that have no corresponding node. So, you might sometimes turn something off and the lists do not change.

### Known issues
This PR doesn't solve a number of existing issues in the rendering engine but improves one of the main diagnostic tool, the logger output describing the current task list. Now I expect to be able to begin addressing issues.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
@@ -19,117 +19,204 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.terasology.rendering.dag.tasks.MarkerTask;
+
 import java.util.List;
 import java.util.Map;
 
 /**
- * TODO: Add javadocs
+ * Instances of this class are responsible for generating the the list of tasks
+ * the renderer executes to (eventually) generate the image shown to the user.
+ *
+ * Tasks are generated out of an ordered list of nodes in the render graph. Each node provides the set
+ * of state changes it needs. Tasks are generated for each state change unless they are redundant.
+ * State changes from a node are redundant if the previous node want the exact same state change.
+ *
+ * After tasks for the non-redundant state changes desired by a node have been generated a task
+ * is appended to the list, to execute the node's process() method, which is where any rendering
+ * actually happens.
+ *
+ * Finally, each node also provides a list of state change resets: state changes bringing a property
+ * back to its default. Tasks are generated also these unless they'd be redundant. A state change
+ * reset is redundant if the next node needs to set that property to a non-default value.
+ *
+ * It should be noted that nodes are skipped if they are disabled.
+ *
  */
 public final class RenderTaskListGenerator {
     private static final Logger logger = LoggerFactory.getLogger(RenderTaskListGenerator.class);
-    private List<Object> intermediateList;
     private List<RenderPipelineTask> taskList;
     private List<Node> nodeList;
 
     public RenderTaskListGenerator() {
         taskList = Lists.newArrayList();
-        intermediateList = Lists.newArrayList();
     }
 
-    public List<RenderPipelineTask> generateFrom(List<Node> orderedNodes) {
-        nodeList = orderedNodes;
-        generateIntermediateList(orderedNodes);
-        generateTaskList();
-        return taskList;
-    }
+    private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
+        logger.info("----- INTERMEDIATE RENDERER LIST --------------------------------");
 
-    private void logList(String title, List<?> list) {
-        logger.info(title);
-        for (Object o : list) {
-            logger.info(o.toString());
+        for (Node node : orderedNodes) {
+            if (node.isEnabled()) {
+
+                // printing out node name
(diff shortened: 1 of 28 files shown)


GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Thanks @emanuele3d :-)

Will assign this one for review to @tdgunes and/or @indianajohn, can test just for reference myself sometime too if needed.

Would you say this completes #2387?

emanuele3d:
Sorry, no. By my understanding #2387 is about the logging the shaders and related classes do. This PR is just about the individual tasks the renderer gets to do every frame, only some of which involve (indirectly) shaders.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -19,117 +19,212 @@
...
-                taskList.add(((Node) object).generateTask());
+        long endTimeInNanoSeconds = System.nanoTime();
+
+        if (logger.isInfoEnabled()) {
+            logger.info("===== INTERMEDIATE RENDERER LIST =========================");
+            logIntermediateRendererListForDebugging(orderedNodes);
+            logger.info("===== RENDERER TASK LIST =================================");
+            logList(taskList);
+            logger.info("----------------------------------------------------------");
+            logger.info(String.format("Task list generated in %.3f ms", endTimeInNanoSeconds - startTimeInNanoSeconds / 1000000f));
```
indianajohn: As this line is written, it's performing the calculation:

```
endTimeInNanoSeconds - (startTimeInNanoSeconds / 1000000f)
```

Which, on my machine, is producing something like:

```
23:01:14.514 [main] INFO  o.t.r.dag.RenderTaskListGenerator - Task list generated in 505639151337472.000 ms
```

A simple fix is to change this line to:

```
logger.info(String.format("Task list generated in %.3f ms", (endTimeInNanoSeconds - startTimeInNanoSeconds) / 1000000f));
```

which appears to solve the issue:

```
23:13:11.863 [main] INFO  o.t.r.dag.RenderTaskListGenerator - Task list generated in 3.043 ms
```

indianajohn:
This was a good opportunity to put my eyes on some of the DAG code with a purpose. The logging output should be helpful in future debugging/optimization.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthMask.java
```diff
@@ -61,10 +61,7 @@ public boolean isTheDefaultInstance() {
 
     @Override
     public boolean equals(Object obj) {
-        if (obj instanceof DisableDepthMask) {
-            return this.enabled == ((DisableDepthMask) obj).isEnabled();
-        }
-        return false;
+        return (obj instanceof DisableDepthMask) && this.enabled == ((DisableDepthMask) obj).isEnabled();
```
tdgunes: Hmmm, I tried to avoid making these lines into one liner, do you think it's better in this way? _I suppose Intellij recommended this._

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
emanuele3d: Aaah! Thanks @indianajohn. I will change it ASAP.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthMask.java
emanuele3d: That's correct @tdgunes, it got flagged by one of the inspections before the commit takes place. I don't particularly like it either, the if block from my perspective was more readable, but it's a small thing and I just went along with the recommendation.

emanuele3d:
@tdgunes: I just submitted a new commit to fix the problem reported by @indianajohn but I did not fix the one-liners. Let me know if you are particularly adamant about them. If not, please give the greenlight to cervator.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthMask.java
Cervator: Eh, agreed on the if maybe being nicer, but it not being a big deal. I guess maybe it is ever so slightly more efficient :-)

emanuele3d:
@Cervator: I'm thinking this is good to go if you have some time to merge it.

Cervator:
Deal!

PR #2491, opened 2016-09-28
Rendering engine: refactored 2 nodes + new node type

emanuele3d:
**[EDIT] the latest commit fixes the bug and makes this PR, further feedback notwithstanding, ready for merge.**

### Contains
* Refactored ShadowMapNode to fully take advantage of the new architecture
* Refactored WorldReflectionNode by extracting a new BackdropReflectionNode class from it, both taking advantage of the new architecture.
* [EDIT] Refactored BackdropNode to partially take advantage of the new architecture.
* Added and using a new BufferClearingNode class
* Added a new NodeFactory.createInstance(Class<T>, Object) method to allow passing initial data to the new Node.initialise(Object) method.
* [EDIT] <s>Changed FBOConfig class to hold information about the fbo manager to be associated with a given FBO.</s> - removed: will be reintroduced in a separate PR. 

### How to test

* Start Terasology
* Create/Load a game
* try changing Video Settings / Shadows
* try changing Video Settings / Water Reflections
* everything should look unchanged

### Outstanding before merging - [EDIT] OBSOLETE: the bug has been fixed.
At this stage this PR and in particular commit e737750e introduce a bug: when the Water Reflections are set to "Sky Only" the sky goes grey. This is probably due to the interactions between the old rendering engine architecture and the new one, as some nodes are still using the old architecture.

I intend to fix this bug before we merge this. However, most if not all the changes already visible in this PR are likely to be left untouched by the debugging process and I thought we could start the review process already.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
@@ -40,6 +40,8 @@ public abstract class AbstractNode implements Node {
     private RenderTaskListGenerator taskListGenerator; // TODO: investigate ways to remove nodes influence on taskList
     private boolean enabled = true;
 
+    public void initialise(Object initialData) { } // added here to avoid adding it to all other nodes.
+
     protected FBO requiresFBO(FBOConfig fboConfig, BaseFBOsManager frameBuffersManager) {
         ResourceUrn fboName = fboConfig.getName();
 
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/BackdropNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/BackdropNode.java
@@ -15,30 +15,33 @@
  */
 package org.terasology.rendering.dag.nodes;
 
-import org.terasology.assets.ResourceUrn;
+import org.lwjgl.opengl.GL11;
+import org.lwjgl.util.glu.Sphere;
 import org.terasology.monitoring.PerformanceMonitor;
 import org.terasology.registry.In;
 import org.terasology.rendering.backdrop.BackdropRenderer;
 import org.terasology.rendering.cameras.Camera;
 import org.terasology.rendering.dag.WireframeCapableNode;
+
+import static org.lwjgl.opengl.GL11.*;
 import static org.terasology.rendering.opengl.DefaultDynamicFBOs.READ_ONLY_GBUFFER;
-import org.terasology.rendering.opengl.FBO;
-import org.terasology.rendering.opengl.FBOConfig;
-import static org.terasology.rendering.opengl.ScalingFactors.FULL_SCALE;
+
+import org.terasology.rendering.dag.stateChanges.DisableDepthMask;
+import org.terasology.rendering.dag.stateChanges.EnableFaceCulling;
+import org.terasology.rendering.dag.stateChanges.EnableMaterial;
+import org.terasology.rendering.dag.stateChanges.SetFacesToCull;
 import org.terasology.rendering.opengl.fbms.DisplayResolutionDependentFBOs;
 import org.terasology.rendering.world.WorldRenderer;
-import static org.lwjgl.opengl.GL11.GL_COLOR_BUFFER_BIT;
-import static org.lwjgl.opengl.GL11.GL_DEPTH_BUFFER_BIT;
-import static org.lwjgl.opengl.GL11.GL_STENCIL_BUFFER_BIT;
-import static org.lwjgl.opengl.GL11.glClear;
-import static org.terasology.rendering.opengl.OpenGLUtils.bindDisplay;
+
 
 /**
  * TODO: Diagram of this node
  */
 public class BackdropNode extends WireframeCapableNode {
-    public static final ResourceUrn REFRACTIVE_REFLECTIVE = new ResourceUrn("engine:sceneReflectiveRefractive");
 
+    public static final int RADIUS = 1024;
+    public static final int SLICES = 16;
+    public static final int STACKS = 128;
 
     @In
     private BackdropRenderer backdropRenderer;
@@ -50,49 +53,52 @@ public class BackdropNode extends WireframeCapableNode {
     private DisplayResolutionDependentFBOs displayResolutionDependentFBOs;
 
     private Camera playerCamera;
-    private FBO sceneReflectiveRefractive;
+    private int skySphere = -1;
 
     @Override
     public void initialise() {
         super.initialise();
         playerCamera = worldRenderer.getActiveCamera();
-        requiresFBO(new FBOConfig(REFRACTIVE_REFLECTIVE, FULL_SCALE, FBO.Type.HDR).useNormalBuffer(), displayResolutionDependentFBOs);
+        initSkysphere();
+
+        // We do not call requireFBO as we can count this default buffer is there.
+        //addDesiredStateChange(new BindFBO(READ_ONLY_GBUFFER)); // TODO: enable FBO this way when default FBOs are standard FBOs.
+        addDesiredStateChange(new EnableMaterial("engine:prog.sky"));
+
+        // By disabling the writing to the depth buffer the sky will always have a depth value
+        // set by the latest glClear statement.
+        addDesiredStateChange(new DisableDepthMask());
+
+        // Note: culling GL_FRONT polygons is necessary as we are inside the sphere and
+        //       due to vertex ordering the polygons we do see are the GL_BACK ones.
+        addDesiredStateChange(new EnableFaceCulling());
+        addDesiredStateChange(new SetFacesToCull(GL_FRONT));
     }
 
     @Override
     public void process() {
         PerformanceMonitor.startActivity("rendering/backdrop");
-
-        initialClearing();
-
-        READ_ONLY_GBUFFER.bind();
-        READ_ONLY_GBUFFER.setRenderBufferMask(true, true, true);
+        READ_ONLY_GBUFFER.bind(); // remove this when default FBOs become standard FBOs.
 
         playerCamera.lookThroughNormalized();
-        /**
-         * Sets the state to render the Backdrop. At this stage the backdrop is the SkySphere
-         * plus the SkyBands passes.
-         *
-         * The backdrop is the only rendering that has three state-changing methods.
-         * This is due to the SkySphere requiring a state and the SkyBands requiring a slightly
-         * different one.
-         */
         READ_ONLY_GBUFFER.setRenderBufferMask(true, false, false);
-        backdropRenderer.render(playerCamera);
+
+        glCallList(skySphere); // Draws the skysphere
+
+        READ_ONLY_GBUFFER.setRenderBufferMask(true, true, true); // TODO: handle these via new StateChange to be created
+        playerCamera.lookThrough();
 
         PerformanceMonitor.endActivity();
     }
 
-    /**
-     * Initial clearing of a couple of important Frame Buffers. Then binds back the Display.
-     */
-    // It's unclear why these buffers need to be cleared while all the others don't...
-    private void initialClearing() {
-        sceneReflectiveRefractive = displayResolutionDependentFBOs.get(REFRACTIVE_REFLECTIVE);
-        READ_ONLY_GBUFFER.bind();
-        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
-        sceneReflectiveRefractive.bind();
-        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
-        bindDisplay();
+    private void initSkysphere() {
+        Sphere sphere = new Sphere();
+        sphere.setTextureFlag(true);
+
+        skySphere = glGenLists(1);
+
+        glNewList(skySphere, GL11.GL_COMPILE);
+        sphere.draw(RADIUS, SLICES, STACKS);
+        glEndList();
     }
 }
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/BufferClearingNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/BufferClearingNode.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2016 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.dag.nodes;
+
+import org.terasology.rendering.dag.AbstractNode;
+import org.terasology.rendering.dag.stateChanges.BindFBO;
+import org.terasology.rendering.opengl.BaseFBOsManager;
+import org.terasology.rendering.opengl.FBOConfig;
+
+import static org.lwjgl.opengl.GL11.glClear;
+
+/**
+ * Instances of this node clear specific buffers attached to an FBOs, in accordance to a clearing mask.
+ * Normally this means that all the pixels in the buffers selected by the mask are reset to a default value.
+ *
+ * Notice that the node is fully initialised and ready to use only after calling the initialise(Object object) method.
+ *
+ * This class could be inherited by a more specific class that sets the default values, via (yet to be written)
+ * state changes.
+ */
+public class BufferClearingNode extends AbstractNode {
+
+    private int clearingMask;
+
+    /**
+     * Throws a RuntimeException if invoked. Use initialise(Object data) instead.
+     */
+    public void initialise() {
+        throw new RuntimeException("Please do not use initialise(). For this class use initialise(Object initialData) instead.");
+    }
+
+    /**
+     * Initialises the node by requesting the creation (if necessary) of the FBO implied in the RequiredData object
+     * passed as argument and by requesting for this FBO to be bound by the time process() gets executed. Also
+     * stores the clearing mask contained in the RequiredData object, for later use.
+     *
+     * @param initializationData an instance of BufferClearingNode.RequiredData, containing an FBOConfig, an FBO manager
+     *                           and a clearing mask needed by this type of nodes to operate. See contained RequireData class.
+     * @throws IllegalArgumentException if the object is not an instance of BufferClearingNode.RequiredData.
+     */
+    @Override
+    public void initialise(Object initializationData) {
+
+        if (initializationData instanceof RequiredData) {
+            RequiredData requiredData = (RequiredData) initializationData;
+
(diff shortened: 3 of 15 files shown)


engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
```diff
@@ -40,17 +40,20 @@
     private RenderTaskListGenerator taskListGenerator; // TODO: investigate ways to remove nodes influence on taskList
     private boolean enabled = true;
 
-    protected FBO requiresFBO(FBOConfig fboConfig, BaseFBOsManager frameBuffersManager) {
+    public void initialise(Object initialData) { } // added here to avoid adding it to all other nodes.
```
emanuele3d: This could be controversial. The alternative would have been to add the same empty method to many node classes.

engine/src/main/java/org/terasology/rendering/dag/nodes/BackdropNode.java
```diff
@@ -56,27 +56,27 @@
...
-        playerCamera.lookThroughNormalized();
-        /**
+        /*
          * Sets the state to render the Backdrop. At this stage the backdrop is the SkySphere
          * plus the SkyBands passes.
          *
          * The backdrop is the only rendering that has three state-changing methods.
          * This is due to the SkySphere requiring a state and the SkyBands requiring a slightly
          * different one.
          */
```
emanuele3d: This was classified as a dangling comment, hence the move by a few lines and the demotion from javadoc to simple block comment.

engine/src/main/java/org/terasology/rendering/dag/nodes/BufferClearingNode.java
```diff
@@ -0,0 +1,140 @@
...
+ */
+public class BufferClearingNode extends AbstractNode {
+
+    private int clearingMask;
+
+    /**
+     * Throws a RuntimeException if invoked. Use initialise(Object data) instead.
+     */
+    public void initialise() {
+        throw new RuntimeException("Please do not use initialise(). For this class use initialise(Object initialData) instead.");
```
emanuele3d: Is this a good way to do this?

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
```diff
@@ -37,21 +37,26 @@
...
+ * - are enabled and disabled depending on the shadow setting in the rendering config.
+ * - in VR mode regenerate the shadow map only once per frame rather than once per-eye.
+ *
  * Diagram of this node can be viewed from:
  * TODO: move diagram to the wiki when this part of the code is stable
  * - https://docs.google.com/drawings/d/13I0GM9jDFlZv1vNrUPlQuBbaF86RPRNpVfn5q8Wj2lc/edit?usp=sharing
  */
-public class ShadowMapNode extends AbstractNode {
+public class ShadowMapNode extends ConditionDependentNode {
     public static final ResourceUrn SHADOW_MAP = new ResourceUrn("engine:sceneShadowMap");
```
emanuele3d: Checkstyle is flagging a lot of these public static final ResourceUrn variables, inside nodes, as they could be made private or package private. The things is, these could be potentially used by nodes provided by a module. How to avoid the warning then? Or should I simply ignore it?

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
```diff
@@ -72,86 +77,123 @@
...
+     * only once per frame. I.e. in VR mode they are executed only when the left eye is processed. This is
+     * done in the assumption that we do not need to generate and use a shadow map for each eye as it wouldn't
+     * be noticeable.
+     */
+    @Override
+    public void process() {
+        // TODO: remove this IF statement when VR is handled via parallel nodes, one per eye.
+        if (worldRenderer.isFirstRenderingStageForCurrentFrame()) {
             PerformanceMonitor.startActivity("rendering/shadowMap");
+            positionShadowMapCamera();
```
emanuele3d: @indianajohn: this move makes sure that the shadow camera is positioned only once per frame instead of once per eye in VR mode. Does this look ok to you? Or do we need to render a different shadow map for each eye? (which would require some modification to the ShadowMapNode I imagine).

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
```diff
@@ -72,86 +77,123 @@
...
-            // postRenderCleanupSceneShadowMap
-            GL11.glEnable(GL_CULL_FACE);
 
+            playerCamera.lookThrough(); //TODO: camera setting need to go into a state change: enable camera, back to default camera
 
             PerformanceMonitor.endActivity();
         }
     }
 
     private void positionShadowMapCamera() {
```
emanuele3d: Are the new comments in this method clear enough?

engine/src/main/java/org/terasology/rendering/opengl/FBOConfig.java
```diff
@@ -20,117 +20,125 @@
...
- * <p>
- * The new FBO is automatically registered with the LwjglRenderingProcess, overwriting any
- * existing FBO with the same title.
+ * An instance of this class uniquely describes an FBO.
+ *
+ * The description can then be used to create an FBO or to retrieve an FBO from one of the FBO managers.
+ *
+ * The constructors alone generate the description for two possible FBOs variety: one with no attached
+ * buffers and one with a single color buffer attached to it. To describe an FBO with additional buffers
+ * see the use*Buffer() methods.
```
emanuele3d: I updated all the javadocs here as the class had been cut&pasted from an FBOBuilder class and the javadocs where no longer up to date.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -207,8 +212,24 @@ private void initMaterials() {
     private void initRenderGraph() {
         // FIXME: init pipeline without specifying them as a field in this class
         NodeFactory nodeFactory = new NodeFactory(context);
+
+        // ShadowMap generation
+        FBOConfig shadowMapConfig =
+                new FBOConfig(ShadowMapNode.SHADOW_MAP, FBO.Type.NO_COLOR, shadowMapResolutionDependentFBOs).useDepthBuffer();
+        BufferClearingNode.RequiredData shadowMapClearingData =
+                new BufferClearingNode.RequiredData(shadowMapConfig, GL_DEPTH_BUFFER_BIT);
+        Node shadowMapClearingNode = nodeFactory.createInstance(BufferClearingNode.class, shadowMapClearingData);
```
emanuele3d: Can't think of a way to do this more neatly.

engine/src/main/java/org/terasology/rendering/dag/nodes/BufferClearingNode.java
Cervator: Certainly looks effective! "Trapping" methods like this is valid when they shouldn't be used IIRC, although there is a balance vs. architecture that attempts to avoid invalid methods entirely instead.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
Cervator: Constants that just happen to not be in use yet are fine to ignore warnings like that on. In theory we could probably add some sort of `@SuppressWarnings` thing when we know something is left available for use but doesn't have a guaranteed consumer.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
Cervator: They look reasonable (and helpful!) to me :-)

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
Cervator: Would it help to extract RequiredData into its own class (even if different flavors might be needed)? I don't know if there's a particular reason to keep it an internal class but it ends up looking kinda awkward (and it all looks public anyway?)

engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
Cervator: It looks kinda odd (along with the trapped version later) but I haven't looked at the code enough yet to understand it well enough to advise better :/

emanuele3d: I will be happy to receive advice.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
emanuele3d: In theory? It wouldn't work in this case?

engine/src/main/java/org/terasology/rendering/dag/nodes/BufferClearingNode.java
emanuele3d: I am open to suggestions to change the architecture if it would be better to do it in some other way.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: It felt neater to put RequireData as a class within the Node as it is very node-specific.  But I can certainly put all the RequireData classes into dag/nodes/requiredData, it would certainly shorten things in this part of the code.

What do you mean with "it all looks public anyway"?

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
Cervator: Sorry, just funny wording. I haven't gone and outright tried/confirmed it, so I didn't want to say this was an absolutely clear case where we should do that. I just *think* so :-)

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
Cervator: This might just be me (and I'm hardly an architect) but when I think about inner classes it is something exclusively used by the wrapping class - meaning it might as well be private within the wrapping class. Here you've got a public static inner class with mostly public methods, so functionally there'd be little to no change with it extracted.

Would different nodes have different flavors of required data? Would they fit well extending an abstract NodeRequiredData class, that has any utility value? Could Nodes with required data have an instance variable of type NodeRequiredData, initialized to the right flavor?

engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
Cervator: Will try to test out the branch here over the weekend then, if @tdgunes (moving IRL right now?) or @indianajohn don't beat me to it :-)

flo: I am alsi not familiar with that code but introducing a class type parameter T taht initualize takes could potentially allow you to keep typesavety. Another idea that might be applyable or not would be to do the initialization in a node specific factory class that implements a node factory interface.

emanuele3d: Thank you @flo for the comment. I'm not 100% sure I understood it. Are you suggesting to make a `NodeFactory<T, S>`, type S being the type of data required by the initialize method? Or are you suggesting to simply constrain the type passed to initialise() with some kind of interface?

My problem right now is that I haven't gone through enough Node refactoring to know if there will be a pattern in the usage of the initialize(Object) method. If there was of course I'd require it to accept only instances implementing an interface or extending a base class. But if I have to look just at the BufferClearingNode, for example, its initial data is very type-specific.

In this context I could perhaps add a TODO and review the situation when most or all nodes have been refactored.

Is there any reason to _not_ use Object as input type? Of course a node could potentially get the wrong object, but isn't that easily solved using `instanceof`?

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
emanuele3d: I'll try it then.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: As mentioned in a new, earlier comment I'm just starting with the refactoring of the nodes to make them fully take advantage of the new architecture. So, I don't know just yet how many node types will be reusable to the point of requiring initial data to customize each instance. 

A number of node types will be instanced only once. But some others will be certainly instanced more than once.  For example right now the "blur" shader is applied five times (as far as I remember), in different circumstances (DoF, Bloom). Those will most likely become instances of the same BlurNode class, each instance initialized with slightly different parameters.

I can imagine every Node type requiring initial data will probably require some kind of FBO to work on. Perhaps the BaseInitialData class could just bring that information, other nodes extending it with their own additional data. 

I'm really thinking this should be tackled when I have more nodes "under the belt", so that I can establish a usage pattern. But I'd be happy to eventually have some kind of base class or an interface to constrain the parameter type.

All that been said, in the meantime I'll extract the inner class.

Cervator: Yep that's fair, as are TODO notes to catch later :-)

Cervator:
Tested this out properly now. It seemed to work entirely fine the first time, I didn't even get the gray sky with the "Sky Only" reflection setting. Restarted the game and tried again, that time I did get it. Part of the oddly inconsistent video settings update issue?

Otherwise everything seems well.

As for the architecture - yeah probably just let it evolve a bit further without worrying too much until the nodes become more clear. Maybe if a subset need that particular required data set there could be a `DataBearingNode` variant of `Node` ? I really don't know and it is probably better to wait.

I'm good with this one right now, other than the gray sky caveat. Some oddity in the order things happen when updating/recompiling video settings stuff?

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
```diff
@@ -72,86 +81,113 @@
...
         renderableWorld.setShadowMapCamera(shadowMapCamera);
 
         requiresFBO(new FBOConfig(SHADOW_MAP, FBO.Type.NO_COLOR).useDepthBuffer(), shadowMapResolutionDependentFBOs);
 
-        // TODO: fix them
+        texelSize = 1.0f / renderingConfig.getShadowMapResolution() * 2.0f;
+        renderingConfig.subscribe(RenderingConfig.SHADOW_MAP_RESOLUTION, this);
+
+        requiresCondition(() -> renderingConfig.isDynamicShadows());
+        renderingConfig.subscribe(RenderingConfig.DYNAMIC_SHADOWS, this);
```
tdgunes: We might probably need an helper function for these lines, I think it started to get a little redundant especially for the nodes that are `ConditionDependentNode`. But I am not sure about how to implement them.

emanuele3d:
@Cervator : the bug is driving me nuts. I spent about 20 hours on it over the past week or so and just ended up banging my head against the keyboard - metaphorically, but if it continues this way... :rage: 

I'm now looking into proper OpenGL debugging, including evaluating some tools. I need to find what the GPU is doing with the commands our application sends to it.

By the way, reading on these issues I also found that what we are doing with the PerformanceMonitor is probably -not- measuring how long each step in the rendering pipeline takes but only how long it takes to submit the command to the GPU. I'll probably learn more as I go through more general debugging knowledge.

emanuele3d:
By the way, all that means that this PR will stay open a bit longer than I'd like. :disappointed:

Cervator:
No problem at all @emanuele3d, take your time and sorry to hear it is causing so much of a headache :-) Effort very much appreciated!

Renderland seems to be taking a well-deserved break anyway, while @tdgunes is moving and @indianajohn is getting lost in the matrix :D

emanuele3d:
Just wanted to document some small progress for those following this PR and occasionally wondering why it's taking so long.

The bug _seems_ to be caused by the sky shader failing silently when it's time to render the skysphere and when SkyOnly reflections are enabled in the video setting. This was confirmed in two ways:
* I changed the final instruction in the fragment shader so that it would set every pixel to read. This works when Global reflections are on but doesn't when Sky Only reflections and SSR reflections are chosen.
* A couple of OpenGL debuggers I have installed allow me to inspect the opengl calls made each frame. Both confirm that when SkyOnly reflections are enabled the shader program doesn't seem to get enabled: the default shader program is active instead and the buffer is painted white.

Notice that the same shader program was successfully enabled in an earlier step of the pipeline (the sky reflected in the water) and did its job. So, this behavior is very peculiar. Will keep you posted.

Cervator:
Thanks for the update @emanuele3d - at least you're in the right place! Messing with the skysphere was the very first thing you wanted to do, IIRC. Took a little while to get there and maybe we aren't adding new celestials just yet but ... so close! :D :fireworks: :star: :star2:

GooeyHub:

emanuele3d:
I seem to have finally managed to fix the bug. 

This PR had its last commit _replaced_ by ba84900, the one I just pushed. But I will eventually re-introduce those changes in another PR. 

During the past 10 days of intense and often frustrating debugging I have seen a lot of rendering code and I had made a number of changes that I was hoping to push with this fix. Unfortunately a couple of different branches ended up with broken rendering for reasons that are hard to discern. The rendering engine remains fragile because OpenGL state changes done in one node affect subsequent nodes in ways that are hard to follow. Even this fix makes some sense but some of the symptoms of the bug remain mysterious to me.

The new architecture developed by @tdgunes and I during summer deals with these problems by bracketing the work done by a node with self-canceling state changes, i.e.:

* set OpenGL state required by node
* node does the work - i.e. rendering an effect such as light shafts
* reset OpenGL state back to default - if appropriate

This will eventually make all nodes mostly independent from each other: a state change required by a node won't affect the subsequent ones unless it is appropriate.

But many nodes are not yet taking advantage of the new architecture, using straight gl* directives to set the OpenGL state and relying on subsequent nodes to reset the OpenGL state.

Furthermore other parts of Terasology might leave the OpenGL state in a non-default configuration before the rendering engine starts its work, further complicating the issue.  

For these reasons the rendering engine will remain fragile for a while, until all nodes have been painstakingly converted to the new architecture. But slowly slowly will get there.

Meanwhile, and feedback notwithstanding, this PR could be good to go.

Cervator:
Appears to test out fine! Thanks again for sticking with it, obscure bugs can really suck :-)

Any chance this optimized anything coincidentally? I have nothing to base that on but a feeling and am probably wrong, but running the game for testing so often seems to at times trigger little "Hmm, something is off...." moments.

I didn't set out to test performance and had some mishmash of medium graphic settings, yet a perfectly smooth 60fps despite spinning around like a madman. Bumped to ultra with smoothed shadows and global reflections and still felt almost as great with 50+ fps

Probably just my mind playing tricks on me, but wanted to mention it just in case. Merging!

emanuele3d:
Regarding optimization: the new architecture does some optimization by preventing redundant OpenGL state changes. But as it has been deployed only on a few nodes I do not expect significant boosts in performance. Also, most of the work is still done in shaders, which haven't been touched.

I'm not quite sure what you noticed though. Do you feel that something is wrong with the rendering or do you feel that it's faster than usual?

Cervator:
Faster, nothing was wrong. Just didn't feel right that it was *so smooth* :D So that's not a bad thing!

Chances are it was just me, was mostly just curious :-)

Issue #1703, opened 2015-04-28
Support for per module permissions


MarcinSc:
Add a permission system per module, querying upon startup of a specified module (first time only) if user allows the specified operation to be performed by the module. The popup message should also display the reason why this module asks for this permission. This setting should be saved and used whenever this module is used (in that version).

As an example, if user starts a game with ModularComputers module, a window should pop-up saying "Modular Computers module is requesting a read/write access to clipboard. The reason specified is: copying/pasting the code in computer terminal." with two buttons "[Allow] [Deny]".

Modules should be able to check if they were granted the specified permission and react accordingly, rather than borking out with a security error.

There should also be a button on the UI that allows user to modify the permissions they have specified earlier.

Cervator:
With https://github.com/MovingBlocks/gestalt/pull/13 / https://github.com/MovingBlocks/gestalt/pull/14 merged to Gestalt we just need to have some functionality in the engine using it to have the security setup a bit more user friendly. It won't achieve all this and it might take a while to get there, but we've talked about it in the past (Android style permissions) and I'm sure we'll get there :-)

SoniEx2:
I can make a TODO list for this.

- [ ] Remove CoreRegistry.

PR #2494, opened 2016-10-03
Whitelist many stream classes

SoniEx2:
### Contains

Adds many InputStream/OutputStream-related classes to the whitelist.

The classes whitelisted in this PR are safe and cannot open files.

This makes way for a potential Virtual FS API, and adds support for many libraries which use InputStream and OutputStream.

### How to test

No idea. Custom asset formats maybe?

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
+++ b/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
@@ -90,6 +90,25 @@ public final class ExternalApiWhitelist {
             .add(java.awt.datatransfer.UnsupportedFlavorException.class)
             .add(java.nio.ByteBuffer.class)
             .add(java.nio.IntBuffer.class)
+            .add(java.io.DataInput.class)
+            .add(java.io.DataOutput.class)
+            .add(java.io.EOFException.class)
+            .add(java.io.UTFDataFormatException.class)
+            /* All sorts of input streams */
+            .add(java.io.InputStream.class)
+            .add(java.io.ByteArrayInputStream.class)
+            .add(java.io.FilterInputStream.class)
+            .add(java.io.PipedInputStream.class)
+            .add(java.io.BufferedInputStream.class)
+            .add(java.io.DataInputStream.class)
+            .add(java.io.PushbackInputStream.class)
+            /* All sorts of output streams */
+            .add(java.io.OutputStream.class)
+            .add(java.io.ByteArrayOutputStream.class)
+            .add(java.io.FilterOutputStream.class)
+            .add(java.io.PipedOutputStream.class)
+            .add(java.io.BufferedOutputStream.class)
+            .add(java.io.DataOutputStream.class)
             .build();
 
     private ExternalApiWhitelist() {


GooeyHub:

Cervator:
@GooeyHub ok to test

I could use some additional opinions on this, if there are any creative ways to somehow use the affected classes in a naughty fashion. Pinging @immortius @flo @rzats @skaldarnar @msteiger .. anybody else?

GooeyHub:

SoniEx2:
These stream classes are safe. However, I don't trust `java.awt` with this change.

It might be a good idea to find out what's actually *required* from `java.awt` and manually whitelist *only* those classes.

flo:
It could be that there are ways for modules already to get access to for example output streams that they are not intended to write to. Allowing the usage of outputstreams could thus potentially allow for abuse that wasn't possible previously. 

So it should be checked carefully if there is a way already for modules to get those streams. 

While exposing output streams introduces the danger of someone discovering in future a way to get an output stream (e.g. via a new method added to an existing whitelisted method in a future JRE version) I also see the potential of allowing modules to read/write binary data from/to verified sources

msteiger:
Why exactly do you need I/O operations? I believe that the idea was to have assets, which are entirely managed by engine to provide all kinds of data.

/offtopic: AWT is currently used to render 2D graphics, read image data (need to verify), write screenshots in different file formats and some copy&paste functionality, iirc. That said, it should be possible to truncate the amount of allowed classes somewhat.

SoniEx2:
Not I/O operations in general. Just I/O streams. Just about any API dealing with file formats will use I/O streams.

If you wanna go and add a virtual filesystem to [ModularComputers](https://github.com/Terasology/ModularComputers), and pass those virtual files to existing java libraries, then you'll want to be able to use I/O streams, because that's the java way to do it.

/offtopic: I also think the API whitelist system should be extended to support method-level and constructor-level whitelisting. It'd help filter out `java.awt`.

Cervator:
On the off-topic note, yeah permissions could be improved substantially, see for instance #1703. Main reason they haven't yet is just because nobody has been interested enough to do it, in part from not having any functionality that would really take advantage of it.

A related on-topic note would be separately providing some sample code to do just that :-) For many things activity just remains low since nobody has pushed the topic to where some functionality can be consumed in its final end-user state. For instance we don't have a lot of customizable character support because we don't have any customizable player characters, because we don't have any customizable character support ...

SoniEx2:
Laziest way to go about this is to just merge and wait for the 0days, really.

Cervator:
> Laziest way to go about this is to just merge and wait for the 0days, really.

100% correct statement! *Really* not what I would prefer doing :D

Mainly I think a simple even tutorial-level example of a Stream being used in a module (assuming this PR merged) would be good to examine this closer, especially since it might also reveal other things that need whitelisting.

SoniEx2:
I couldn't find any existing APIs that give me byte[]s... If I could find one, I'd be able to convert it into a ByteArrayInputStream, then wrap it in a DataInputStream, and read strings, ints, etc off it instead of raw bytes.

And with DataOutputStream and ByteArrayOutputStream I could produce byte[]s from ints, strings and stuff, but I couldn't find any APIs that take byte[]s either...

Then again, I can't use "[]" in github search, so it's kinda hard to search for "byte[]"...

Cervator:
> Then again, I can't use "[]" in github search, so it's kinda hard to search for "byte[]"...

You *reeeaaally* should be using something like IntelliJ search to run through code locally, not off GitHub. Then finding `byte[]` would be trivial.

I don't know if there could be localization issues to using raw bytes somehow. Anyway, that's probably a separate discussion than this PR.

flo:
>and pass those virtual files to existing java libraries, then you'll want to be able to use I/O streams, because that's the java way to do it.
Oh when you want to use an external libary that wirks with streams thrn that library will probably also use file related streams that can't be whitelisted for obvious reasons. So you can't use those external libraries in normal terasology modules.

So you could:
* start eorking on introducing a way of installing "unchecked modules" that can use any/external APi. Of course such modules would thenhave to be installed in a way that makes the user realize that the module can do anything without restriction.
* If you don't mind if the user has to do some manual installation steos it may already possible to add unchecked modules via classpath. The engine itself is implemented as a classpath module. So maybe it is possible to add other classpath modules.
* If you want a simpler solution you could simply fork Terasology and put the problematic code directly in the engine.

In each case you would't need this PR

SoniEx2:
Or, I do this PR and we don't have to worry about literally breaking the sandbox just to support fake files. All 3 options require manual install on the user's end, while this PR doesn't.

Also, if we ever get built-in fake file support, you'll thank me for this PR... Well I mean the stuff in this PR is kinda required for built-in fake file, assuming you want any sort of interoperability with existing java libs (which you should!).

Cervator:
> Also, if we ever get built-in fake file support, you'll thank me for this PR...

Kinda. This PR just whitelists a few classes. What we'd be really thankful for is more related sample code, or alternative engine implementations that get you where you are wanting to go while enhancing the engine further :-)

I'm not opposed to merging this, it just helps to have a clearer goal. The bit about maybe finding more things needed on the whitelist is central there - if they turn out to be insurmountable obstacles we have to find an alternative. If they're just benign extras we can add them to this PR before merging.

SoniEx2:
@Cervator This will allow a bunch of existing java libs to work inside the sandbox.

flo:
If it is a lubrary that we expect a lot of modules to use we could include it in the engine and expose of it what is needed. Which may not be output streams.

If it is a specific library that is only useful to a module or two then that library would have to meet currently the restrictions.If a library uses output streams then it very likely will also will use other file related API.So why schould we allow the use of output streams when the typical library needs more to work anyway?

SoniEx2:
The way Java works is that libraries are more likely to use streams than file APIs. E.g. https://github.com/eNByeX/NBX-API doesn't use any file APIs but it does use streams.

The Java file APIs are not for reading files. They're only for opening files. File I/O is done through FileInputStream/FileOutputStream but last I checked most libs don't use those directly.

msteiger:
What exactly is it, you're trying to build there, @SoniEx2? I'm curious..

SoniEx2:
@msteiger Streams would be useful for image-to-world importers, song-to-world importers (NBS song importers), virtual filesystems, bytecode-to-file exporters, Lua's load() function, etc.

It also opens way for a bunch of new APIs, such as a file transfer API (anything involving files), an audio streaming API (in-game synths, anyone?), custom asset formats (XML instead of JSON (assuming you depend on an XML module)), maybe even a built-in virtual filesystem API backed by actual on-disk files (instead of being backed by RAM and using unnecessary amounts of RAM).

msteiger:
Well, the idea is to *not* allow modules to access the file system. New file types - be it lua scripts, nbs songs, etc) should be managed by engine, so that they're available everywhere in a managed environment.

The listed I/O streams look OK to me in general, but I'd still close this until we have a real use case.

immortius:
The stream classes themselves look safe to me. I would say a simple use case for adding them is to support introduction of new asset file formats by streams. The important thing is they cannot open or save files, or access the network, in of themselves. If you want to be very careful, the only immediately needed stream is BufferedInputStream, since that is what the asset system makes available to formats.

On awt... yeah, could do with a review. Perhaps we should just drop support for awt, although I know this probably affects some world gen code. Alternatively, wrap it, throw in permission checks.

If someone is willing to do some work on prompting the user for permission to use a module with non-default api reqs, then Terasology can do the android style "This module requires File permissions" thing.

On virtual file systems, gestalt-module actually presents an active module environment as a virtual filesystem (via NIO) - Terasology doesn't make this available to modules though.

On doing method and constructor level whitelisting... this is sort of possible, but.. ugh. The sandbox system at the moment works through custom classloaders that do not make non-whitelisted classes available to code. This is performant and straight forward. Method level whitelisting... I can think of a couple of approaches, but they are all fairly involved. Probably the least icky is dynamically reviewing module code to ensure it doesn't call disallowed methods, but you'ld have to deal with how this works with inheritance in the mix. It is generally simpler and better to create and expose custom api that wraps the functionality instead. Then you can also use the standard java permissions system to limit method calls.

SoniEx2:
Use cases include Lua and Note Block Song support. Those shouldn't go in the engine, nor should they run unsandboxed.

Cervator:
Thanks for the reviews everybody.

I am going to *conditionally* merge this PR with the expectation that @SoniEx2 will actually soon provide some module code exercising one or more of the whitelisted classes, yes :-) If not we can always roll this back if the classes go unused.

I submitted separate issue #2531 so somebody can take a closer look at AWT.

Separately: sounds like method / constructor level whitelisting isn't worth it so lets forget about that and stick with nice little wrapper classes instead :-)

Cervator:
Hmm, this didn't merge since it was targeted at the `master` branch, not `develop` as is our usual practice. And trying the new feature to change the target branch for a PR didn't work after I already merged the commit into `develop`

So this is merged even though I am technically closing the PR.

PR #2496, opened 2016-10-05
add Hungarian flag

Avalancs:
Add hungarian flag for the settings screen. Also corrected some translations in weblate, how does it appear in git?


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

@Avalancs: thanks again! Huh, we have a Hungarian translation file, but it isn't showing in the player settings list (so easy to miss it also didn't have the flag). That's weird. 

Translations from Weblate we pull over from its branch to develop here occasionally, sometimes right around when we're preparing a release. 

I see pending changes there, will try to merge over tomorrow then figure out why the entry doesn't show up. When it does we should be able to validate the flag from here and merge :-)

GooeyHub:

Avalancs:
I could see the language in the develop build, that's where I noticed it's missing the flag.

msteiger:
Thanks @Avalancs - where does the icon come from? If I remember correctly, we're using [famfam flags](http://www.famfamfam.com/lab/icons/flags/). It would be nice, if the flag icons were consistent (licensing issues aside).

Avalancs:
Ohh sorry, I didn't know. I took the image from the hungarian wikipedia https://hu.wikipedia.org/wiki/Magyarország_zászlaja and edited it with paint so it would look decent in this size.

Cervator:
> Should I replace it with the one from famfam?

Yes please, so we can be consistent with the icons and licensing :-)

Avalancs:
Okay, let me start another pull request, so the one I uploaded won't appear in the history.

PR #2497, opened 2016-10-05
Fixes #2489 - ESC now recompiles shaders while exiting video settings

### LINKED ISSUE #2489: VideoSettings Bug: "Back" button != Esc key
Switch between disabled shadows, On shadows and On (pcr) shadows.

### What actually happened
Starting with shadows "On (pcr)", depending on the order of operations shadows would stay soft (pcr) even if "On" shadows are selected OR the shadows move with the camera when shadows are supposed to be disabled.

### How to reproduce
* Load a game with shadows set to "On (pcr)". 
* make sure to have some shadows in sight, if necessary use the arrow up/down keys to move the sun.
* While in-game, call the Video Settings menu and switch the shadows to "On"
* IMPORTANT: do not press the "back" button in the UI. Simply press the "Esc" key to go back to the game. Result: the shadows are still soft.
* call the Video Settings menu again and this time set the shadows to "Off". 
* IMPORTANT: again press the "Esc" key instead of the "back" button in the UI. Result: the expected shadows are gone but some shadows are still visible, moving with the point of view and orientation of the player's camera.

### Cause
At the end of VideoSettings.initialize() the line

`CoreRegistry.get(ShaderManager.class).recompileAllShaders();`

guarantees that shaders are recompiled according to the current video settings, but _ONLY_ if the "back" button is pressed. If the button is pressed, all is well.

On the other hand, pressing the Esc key does not undo the changes that have taken place in the VideoSettings windows. But the shaders are not recompiled. So, parts of the rendering engine, monitoring the video settings directly correctly take advantage of any new setting. But the shaders are unaware of the changes, resulting in visual artifacts.

### Some possible solutions
1. The "Esc" key is made to have the same behavior as the "back" button, forcing the shaders to recompile.
2. The "Esc" key is disabled while the Video Settings menu is active.
3. The "Esc" key undoes all the changes before it closes the Video Settings window.
4. Other?

### Game Version
At least commit a6ea19277 - at the time of the writing it's the latest in develop

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38003814-videosettings-bug-back-button-esc-key?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nihal111:
### Contains
Fix for #2489: The ESC key now does exactly what the "back" button does i.e recompiles shaders etc.

### How to test
- Enter Setting -> Video, (either in-game or from outside)
- Hit ESC
- Check the logger
It should now show something like `VideoSettingsScreen - Video Settings: {"pixelFormat":24,"wind...`, which would earlier only arise after the back button was clicked.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/VideoSettingsScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/VideoSettingsScreen.java
@@ -23,6 +23,7 @@ import org.terasology.config.Config;
 import org.terasology.engine.GameEngine;
 import org.terasology.engine.subsystem.DisplayDevice;
 import org.terasology.i18n.TranslationSystem;
+import org.terasology.input.Keyboard;
 import org.terasology.logic.players.LocalPlayer;
 import org.terasology.registry.CoreRegistry;
 import org.terasology.registry.In;
@@ -32,6 +33,7 @@ import org.terasology.rendering.nui.WidgetUtil;
 import org.terasology.rendering.nui.animation.MenuAnimationSystems;
 import org.terasology.rendering.nui.databinding.BindHelper;
 import org.terasology.rendering.nui.databinding.Binding;
+import org.terasology.rendering.nui.events.NUIKeyEvent;
 import org.terasology.rendering.nui.itemRendering.StringTextRenderer;
 import org.terasology.rendering.nui.itemRendering.ToStringTextRenderer;
 import org.terasology.rendering.nui.widgets.UIDropdown;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Using the instructions from #2489 I seemingly get the correct settings even using ESC now, huzzah!

@emanuele3d does this approach look OK to you? I suggested we just add the shader compilation step on using ESC. Less complicated and less potential for confusion than leaving ESC to roll back changes made while the user is on the settings screen.

msteiger:
Thanks for the PR, @nihal111 - I'd prefer to extract the logic to a new method (maybe call it "save" or something like that) rather than duplicating the code.

GooeyHub:

nihal111:
@msteiger @Cervator Did the said changes. Have a look :)

Cervator:
Thanks for the update! Re-tested and merged :-)

emanuele3d:
Just went through this PR: thank you @nihal111, this must have been one of the fastest issue-to-merge transition. I appreciate your contribution. I hope you'll stick around for a while.

nihal111:
@emanuele3d Thanks a lot! Yes, I plan to stick around! :)

PR #2502, opened 2016-10-06
add Hungarian flag

Avalancs:
add Hungarian flag from famfam.


GooeyHub:

Cervator:
Oh! I see why I missed Hungarian in-game now. Before merging translations (#2510) the Hungarian language name in English was "Angol" - I was wondering what that entry was :D

This round of translation fixes straightened that out so it makes sense now. Most typical translation field that goes funny since that field needs to be in English in *any* language :-)

So now I see language entry as well as the flag! Huzzah!

Avalancs:
Yes, I realized that only when I checked it out in game, sorry for the confusion :D Since all the languages came up as translations, I thought that was supposed to be "English" in the target language :)

PR #2503, opened 2016-10-06
JSON formatting overhaul

rzats:
### Contains

Reformats all of the game's JSON-based asset types
- *.atlas
- *.behavior
- *.block
- *.blocksounds
- *.info (shader metadata)
- *.lang
- *.mat
- *.prefab
- *.shape
- *.skin
- *.texinfo (texture metadata)
- *.ui

using a common JSON code style template:

- *Use tab character:* ✗
- *Tab size:* 4
- *Indent:* 4
- *Continuation indent:* 8
- *Keep indents on empty lines:* ✓
- *Spaces*
  - *Within*
    - *Braces:* ✗
    - *Brackets:* ✗
  - *Other*
    - *Before comma:* ✗
    - *After comma:* ✓
    - *Before ':':* ✗
    - *After ':':* ✓
- *Keep Maximum Blank Lines*
  - *In code:* 0
- *Right margin (columns):* Default
- *Wrap on typing:* Default
- *Keep when reformatting*
  - *Line breaks:* ✓
- *Ensure right margin is not exceeded:* ✗
- *Arrays:* Do not wrap
- *Objects:* Wrap always
- *Align object properties:* Do not align

### How to test

The actual changes were made using IDEA's autoformatter tool, so it's unlikely that any JSON got corrupted in the process. Still, a quick look at the affected files would be nice.

- [x] Any modifications to the codestyle being used are welcome!
- [ ] default.cfg is not modified as changing it might lead to unexpected "config values changed" triggers. Should it be reformatted anyway? If so, non-default JSON configs will need to be modified to produce properly indented (4-space) output for consistency.


GooeyHub:

Cervator:
Mashed forward buttons a lot to glance at all modified file, caught one where the order of two attributes was also inconsistent (unrelated to formatting) and swapped that too, plus removed piles of license headers. Started a Core world while checking the log closely for parsing errors and all seems well! Thanks :-)

The shader assets in the engine still have the license header, but they're in a weird limbo between code and asset. Dunno what's appropriate there, but maybe we should eventually update those headers (they probably count enough as code). Not going to do that now though, leaving the 3d wizardry files to the 3d wizards.

Cervator:
Oh, as for `default.cfg` I'm not too worried. Maybe sometime when we change that bit of auto-generating JSON code for some other reason.

@rzats think there's anything we can do to help customize expected JSON code conventions in IntelliJ/Eclipse config?

emanuele3d:
The content of shader files are concatenated together with other files and strings to produce viable shader code, compiled at runtime. Without further information/discussion I'd recommend to leave some kind of license information in the asset files. What would be nice is to somehow mark license information in a way that makes it easy to remove it programmatically. Right now license information from three files is concatenated in the string passed to the shader compiler. =P

Cervator:
>  Right now license information from three files is concatenated in the string passed to the shader compiler. =P

Doh!

I wonder if we could just use a policy listed elsewhere to state "All files under /asset directories are covered by so and so" then remove all the individual headers.

emanuele3d:
I don't know. Somehow I feel somewhat reassuring the presence of the license information in every file. This way nobody can't say: "Oh! I didn't realize these files are under some kind of license."

Cervator:
In that case it is quite possible we can write something in Gradle to strip out the headers at compile time. Heck I think there are entire libraries meant to minify the actual compiled source (even to the level of obfuscation for simply shrinking the files)

emanuele3d:
Sorry, the word "compilation" in this context is ambiguous. Shaders are not compiled during the build. They are compiled at runtime. And the code that is compiled at runtime is itself a string assembled at runtime from a number of pieces. And they can be recompiled multiple times per session. So, it's something I'll deal with at some point. 

I'm just suggesting the license information should stay and it should be written in a way that should make it easy to remove it at runtime, so that eventually I'll purge the duplicate information before the shader is compiled.

rzats:
@Cervator: 
>think there's anything we can do to help customize expected JSON code conventions in IntelliJ/Eclipse config?

I've done some research on this. Apparently JSON auto-indenting in Eclipse actually requires a [plugin](https://github.com/boothen/Json-Eclipse-Plugin), and I couldn't find a setting in the vanilla `formatter.xml` that would modify IDEA's JSON indentation settings.

Pinging @msteiger since he's the current maintainer for the plugin! Great screenshot in the README 😜

Issue #1145, opened 2014-06-21
Allow player customization of eye (camera) height, maybe collision height


Cervator:
Another fairly simple configuration change for more player flexibility!

Currently the player's eye height and character (collision) height is set to 1.4 and 1.6 (in block units) which some player's find a tad low. It would be nice to offer a narrow range on the player configuration screen

This needs a bit of experimentation to see what good values are and if they should just be for eye height (within the collision/character height so it doesn't feel like your eyes are scraping against the ceiling) or for both.

Alternatively if it doesn't break the world maybe wider values would be fun to support. We'll get shrink/growth potions eventually!

havanagrawal:
Hi, I'd like to contribute to this project, and this seems like a nice place to start :grin: . Is this issue still active?

Cervator:
Hi @havanagrawal and yes! It absolutely is :-)

It has been a little while since this issue was submitted. We have a settings menu for the player now and I just submitted #2476 which goes into way deeper detail on a bigger topic (player selection of avatar) but some of those details might be useful to consider for this topic. Or they could be overwhelming. Try out just making sure you can make and validate changes first of all! :D

@xrtariq2594 has been working on the Potions module which could eventually result in grow/shrink potions that would also affect the eye/camera height. I think when this issue was originally submitted it was for a bit more of a debug tool to get a better feel for how the height would make the world appear, but really in some less focused gameplay settings (like a creative mode) personal preference on such things would be very valid.

You might want to check out how the current player options (name, color..) are taken from the menu settings and applied to in-game configuration. That might help figure out how to create the new setting.

Are you up and running from source yet? You could post over in the forum as well to say hello: http://forum.terasology.org/forum/contributor-introductions.7/ 

Welcome! :-)

havanagrawal:
Hi @Cervator , I'm up and running from source :grin: I've just started exploring the source though.. Perhaps you could suggest an issue which might not be too overwhelming for me.. Otherwise, I could try working on this one.. I'm pretty comfortable in Java, but haven't worked on any Game dev of any kind before, so it might take me some time..

Cervator:
@havanagrawal great!

This one shouldn't be *too* bad since you should be able to go look at how the existing player settings apply their config to actual in-game objects (giving the player a name and a color). Should be doable to use the same approach to apply an updated camera height.

Other easy ones that stick out: #2383, #2382, #2321

[...]

PR #2511, opened 2016-10-08
Fix for #1145 adjustable playerHeight

nihal111:
### Contains

Fix for #1145

### How to test

- Open Player Settings to now see two sliders for height change
- `playerHeight 2.8` in console would change height to 2.8. Only takes effect after restart of game.

- [x] Need to change height instantaneously without restart.
- [x] Need to find a way for saved config values in PlayerConfig.java to take effect atleast on player spawn.
- [x] Setting eye-height


GooeyHub:

nihal111:
@Cervator @msteiger @rzats Need help on this. I've created a console command for changing height. This seems to work only after restart. Also, setting height to a large value like 10, results in slow increase of height gradually, as though blocks are being piled up to create our character. I think something is happening around [here](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java#L133)

GooeyHub:

Cervator:
Cool!

Sliders look good :-)

The height setting command did its trick and it appears it ends up dropping the player into the ground to match the new height to some degree. So you have to slowly jump out until your feet are finally on top of solid ground again.

I wonder if the increased height should also affect jump height. That'd be a cool extra. Would need to then also be taken into account for fall damage (a 30m giant able to jump 15 blocks in one go probably shouldn't then die from a 15 block fall)

It may well be that you need a UI binding to something that'll trigger an event that can be caught by the ES to adjust the player's height dynamically ... somehow. Like we went over on chat there may be some odd interactions with the physics system.

May need some more tinkering and discussion with those who'd likely know more, like @Josharias @portokaliu, @Flo or @immortius if you can catch him on IRC sometime (the others Slack may work best with)

For general FYI on justifying this odd feature: I consider it an experimental stepping stone to get into working on grow/shrink potions (hi @xrtariq2594!), or varying the player's height when crouching or even crawling (allow passing through a 1-block height gap, don't allow a sniper to hit a player crouching behind blocks). Some of those may be better handled via temporary modifiers rather than changing height and such directly.

Maybe some gameplay would allow the setting (creative building) while others should be able to block it somehow.

I've got more thoughts on the topic but not enough time, up way too late as it is. But keep up the good work! We need more experts in stuff like this :D

GooeyHub:

engine/src/main/java/org/terasology/config/PlayerConfig.java
```diff
@@ -64,4 +80,12 @@ private Color defaultPlayerColor() {
         List<Color> colors = CieCamColors.L65C65;
         return colors.get(rng.nextInt(colors.size()));
     }
+
+    private Float defaultPlayerHeight() {
```
skaldarnar: default height could be constants in this case

```java
pivate static final float DEFAULT_PLAYER_HEIGHT = 1.8f;
pivate static final float DEFAULT_PLAYER_EYE_HEIGHT = 1.6f;
```

engine/src/main/java/org/terasology/config/PlayerConfig.java
```diff
@@ -46,6 +50,18 @@ public void setColor(Color color) {
...
+    public Float getHeight() { return height; }
+
+    public void setHeight(Float height) {
+        this.height = height;
+    }
+
+    public Float getEyeHeight() { return eyeHeight; }
+
+    public void setEyeHeight(Float eyeHeight) {
+        this.eyeHeight = eyeHeight;
```
skaldarnar: would it make sense to restrict eye height to be lower than player height? should we check this?

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -38,6 +41,10 @@
 @RegisterSystem
 public class MovementDebugCommands extends BaseComponentSystem {
 
+    @In
+    private PhysicsEngine physics;
+    private static final Logger logger = LoggerFactory.getLogger(MovementDebugCommands.class);
```
skaldarnar: I think the logger should be the first member (style violation)

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -203,4 +217,23 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+    public String playerHeight(@Sender EntityRef client, @CommandParam("height") float amount) {
+        ClientComponent clientComp = client.getComponent(ClientComponent.class);
+        CharacterMovementComponent move = clientComp.character.getComponent(CharacterMovementComponent.class);
+        if (move != null) {
+            float prevHeight = move.height;
+            move.height = amount;
+            clientComp.character.saveComponent(move);
+            LocationComponent loc = client.getComponent(LocationComponent.class);
+            Vector3f currentPosition = loc.getWorldPosition();
+            teleportCommand(client, currentPosition.getX(), currentPosition.getY() + (amount - prevHeight)/2, currentPosition.getZ());
```
skaldarnar: instead of using the other command go for `CharacterTeleportEvent`(as in the teleport command itself). Two reasons for this: 
(a) the visible effect of sorts should be a change of player height, I don't bother whether the character needed to be teleported, and
(b) it should be easier to refactor this later on when you want to integrate it with potion effects and such.

engine/src/main/java/org/terasology/config/PlayerConfig.java
nihal111: p**r**ivate you mean :stuck_out_tongue:

portokaliu:
I think the point in putting the static default values was to get rid of the methods that returned them altogether (since they were private anyway). There are ways to get the current height though.
Could we check that once we set a new height we reset the eye height to be <= than the new set player height?

GooeyHub:

GooeyHub:

nihal111:
@Cervator Huzzah! :smile: 
@portokaliu @skaldarnar Height changes are being made onPlayerSpawn in `LocalPlayerSystem` by calling the console method I created in `MovementDebugCommands`.
This was ad hoc for testing. What changes should I make?
Also, if somebody could help me with where to start with changing the eyeHeight.. Deals with "gaze" etc @Josharias

GooeyHub:

GooeyHub:

xrtariq2594:
Good work!

As @Cervator mentioned, once this is more stable, I'll look into adding grow/shrink potions.

Cervator:
Did you get the eyeHeight working? Commits indicate yes, checkboxes in OP indicate no :-)

Two PRs merged, two more reviewed tonight, but I'm out of time. At least I got one of your other ones, the march to a nice shirt is making progress! :D

portokaliu:
I might've found an issue with this while testing out the eye height. When the player is taller (Tried above 3 height) he cannot collect items off the ground. When changing it back, he can collect them once again.

The eye height itself didn't seem to work though. Is there a certain way to set it? Changed it in the slider went to the main menu and back, player remained the same (with height set to 1.5 from default and eye height set tot 0.5 from default)

nihal111:
@portokaliu The eyeHeight is reset only onSpawn, so if you are in the same game, the quickest way to test it would be a `teleport 0 100 0` and Respawn. Otherwise, create a new game so that the onSpawnEvent is triggered.

nihal111:
As for the picking up items off the ground, any way that might be fixed?

portokaliu:
Look into what is needed for picking them up. Colliding with the capsule should've been enough, but maybe it's something else? I can't check right now but you can have a look see what's needed. 

From what I recall the event sent to pick up an item is the CollideEvent. See what you can do with that.

nihal111:
@Cervator I'll deal with the extras and other problems in a different PR, creating issue for the same.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Merged after a last round of tweaks - thanks! :-)

Issue #2492, opened 2016-10-01
Quirks related to swimming in shallow water (with or without jump/crouch)


Cervator:
We seem to have arrived at some odd behavior when swimming, possibly distinct but related issues.

* You can jump in some cases when "grounded" under water (use `ctrl` to descend and force yourself to make contact with the ground underwater) - the resulting jump is substantially higher than a regular jump, maybe somehow connected to the "helper jump" that allows the player to more easily get out of water
* In shallow water you can get stuck in a cycle of sinking, then auto-jumping as if you're getting out, then sinking again, even when staying completely still. Been able to inconsistently reproduce this in either 1 or 2 block deep water
* You can "walk on water" oddly long with minimal sinking + the ability to still execute jumps, even the "super" jump that is higher than regular jumps (despite pushing off from a water surface)

I suspect this is caused by poor usage of the grounded state in movement, plus unbalanced config on exactly how fast the player should sink and how powerful a "getting out of water" jump should really be. We've tweaked at this from time to time in the past. I think it is fair enough that a player can tread water so to say (like if you've swum back up to the surface), but that shouldn't be triggered while the player's feet are barely getting wet yet.

Maybe an ideal goal would be:

* No jumping while in water what so ever (even if "grounded") - unless I guess some module wants to add "water jumping". By default the jump key (`space`) during swimming should purely ascend, like the crouch key (`ctrl`) descends. No special handling while swimming.
* If a player is actively pushing into a non-liquid block with air above it (beach/land) give the player a small momentum boost to go up and in (the little "helper jump" to get onto land) - beware of messing with underwater behavior (swimming down while pushing against a wall, or weird conditions where an air block exists away from a regular beach)
* Let the player sink *fast* while primarily above water, then slow down radically when 80-90% of height is underwater (let the player swim downwards or use `ctrl` to descend further)

Unsure if recent changes caused some of these quirks, like the ability to add multi-jump potions or change swim speed, etc (pinging @xrtariq2594 for reference)

AWebb3701:
> Let the player sink fast while primarily above water, then slow down radically when 80-90% of height is underwater (let the player swim downwards or use ctrl to descend further)

Tap ctrl to descend normally, Hold to do a "dive" that quickly reaches normal descent speed?

Cervator:
> Tap ctrl to descend normally, Hold to do a "dive" that quickly reaches normal descent speed?

[...]

PR #2515, opened 2016-10-08
Swimming improvements

coty91:
### Contains

Improvements to swimming and diving for #2492. Damage was not added to diving but is planned for a later PR.
### How to test

Jump while surface swimming, fully submerged, and walking on a solid block with one block of water on top. Also, jump from several different heights into the water to test the new diving improvements.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterStateEvent.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterStateEvent.java
@@ -31,6 +31,7 @@ public class CharacterStateEvent extends NetworkEvent {
     private MovementMode mode = MovementMode.WALKING;
     private boolean grounded;
     private Vector3f velocity = new Vector3f();
+    private float scaleGravity = 0f;
     private float yaw;
     private float pitch;
     private float footstepDelta;
@@ -46,6 +47,7 @@ public class CharacterStateEvent extends NetworkEvent {
         this.mode = previous.mode;
         this.grounded = previous.grounded;
         this.velocity.set(previous.velocity);
+        this.scaleGravity = previous.mode.scaleGravity;
         this.sequenceNumber = previous.sequenceNumber + 1;
         this.pitch = previous.pitch;
         this.yaw = previous.yaw;
@@ -68,6 +70,7 @@ public class CharacterStateEvent extends NetworkEvent {
         this.rotation.set(rotation);
         this.velocity.set(velocity);
         this.mode = mode;
+        this.scaleGravity = mode.scaleGravity;
         this.grounded = grounded;
         this.sequenceNumber = sequenceNumber;
         this.pitch = pitch;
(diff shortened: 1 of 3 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

Looking good! Feels way better. Well done!

Some quirks /problems noted:

* In one block deep water I could end in up in two states, more rarely I could hear footsteps, typically water squishy noises (expected). I *think* the footstep state comes from swimming up and over to the 1 block deep area, but only sometimes. The footsteps state could get disrupted very easily, like just pushing up against a block could reset it to the expected squishy state, and jumping out then back in definitely "fixed" it
* In either case I could jump from the one block deep water which IMHO is fine, especially since it makes it easier to get out. I wonder if there is a way to limit the strength some of a jump from that sort of starting point?
* I think you mentioned this one, but now I see it better: if you make a column of blocks in deeper water so the top block is even with the water you still can't get out (as you aren't grounded so you can't jump out). We need some sort of fix for that. Ladder blocks use a good solution there you could probably steal *but* - read on!
* It seems to be messing with ladder blocks. Try `giveblock ladder` and place them against a wall. Before this code you can move up and down the ladder easily simply by pointing up or down, even including strafing into/away from the wall to climb up/down. If the ladder is flush with the top block you can simply move up and you'll move gracefully on top of the block the top ladder is against. *With* this code it feels like you are swimming when in a ladder and you can't get up and over the top block.
** Also, hah, the Breathing module triggers when you're on a ladder, as if you're in water :D Before or after this PR, so no change there. But if you spot any obvious issue causing that feel free to fix (note that the icon and placement of the breathing bar is off, that's a known issue)
* It breaks the unit tests in `KinematicCharacterMoverTest` :(
* It has some minor code style violations (usually whitespace) - easy to fix with Checkstyle hooked up
* Fine-tuning: Is there any way the player max "poking out of water" distance could be lowered just a bit? It feels like I'm too high when I am holding down `space` to "tread water" at maximum height
* Fine-tuning: If the resting height when "treading water" is lowered a bit would it make sense to let the player sink when not actively treading water with `space` ?

I love the little bounce when you come out of water, or are swimming right by the surface. Cool when it pulls you back down under water a little so it takes a bit of effort to stabilize above water.

Also neat to see the player dive deep from higher jumps (if that's actually changed, unsure ..)

GooeyHub:

coty91:
Thanks for the input. I totally forgot to check ladders before doing the PR. oy. Anyway, I'm going to work down your bullet points and fine tune it some more.

Cervator:
Bump - any progress? Lost track of this one since it got closed. You can actually keep PRs open and just apply more commits to the same branch then the PR here will auto-update. Easier to keep track over time and work on review etc :-)

coty91:
I haven't had time to make progress lately. I'll work on it this weekend, though. Didn't realize I could keep them open until recently.

GooeyHub:

Cervator:
Bump - @coty91 I'm curious about your thoughts on my comment in #2586 (and how you're coming along on this PR). I figure changing around the player's (or creature's?) shape automatically (if appropriate for the active model) on entering/leaving water might relate to your interests :-)

flo:
@coty91  do you still plan to work on this further?

GooeyHub:

jellysnake:
Pinging @Cervator and @coty91
Any updates on this? Perhaps close it since last ping didn't result in a response?

coty91:
Close it. I haven't had time to work on it.

jellysnake:
Okay. Remember you can re-open pull requests so if you have time in the future then feel free to reopen it

Issue #2517, opened 2016-10-08
Windowed / borderless fullscreen mode


Cervator:
Feature request that came in from today's play test. At present we only support regular windowed mode and regular fullscreen. Plenty of games have a nice hybrid where you can run a borderless windowed mode you can still move the mouse out of. Similar to maximizing the windowed game but that leaves in place the menu bar + task bar.

*Probably* this just involves switching the fullscreen checkbox under graphic options to a drop-down with three options:

* Normal window (default)
* Borderless maximized
* Fullscreen

Or something like that (maybe check the naming other games use)

Then affect some toggle in LWJGL/OpenGL to set the correct mode as you confirm the setting. I don't know where it is but am hoping looking at the existing fullscreen code would make this easy enough to be bite-sized.

kartikey0303:
I can try this. In involves adding an additional dropdown for borderless fullscreen and implementing the same in LWJGL/OpenGL. Right?

coty91:
Already took care of it.

PR #2519, opened 2016-10-09
2517 - Added ability to set display mode to Windowed Fullscreen.

coty91:
### Contains

Solves #2517 

### How to test

Go to Settings -> Video and change Display Mode to Window Fullscreen and then back to Fullscreen or Windowed.


GooeyHub:

Cervator:
Cool! Can verify it works as advertised :-)

@emanuele3d anything here you want to check out real quick, since it kinda-sorta relates to your field?

@coty91 one potential issue: with this PR I can no longer resize the game window or use the maximize button in the window frame (at least after cycling through the options a few time). Subtly different window mode setting that disallows that?

coty91:
@Cervator Sounds like a bug I missed.  I'll patch it up asap.

coty91:
So, from the little bit of googling I've done on this issue, it appears to be related to the version of lwjgl we're using. I'll continue to look into it and create a ticket with my findings.

engine/src/main/resources/default.cfg
```diff
@@ -24,6 +24,7 @@
     "windowWidth": 1152,
     "windowHeight": 720,
     "fullscreen": false,
+    "windowedFullscreen": false,
```
Cervator: Should this be merged with `fullscreen` and turned into enum values instead of booleans?

Cervator:
Submitted #2532 to cover the separate bug :-)

Agreed with @emanuele3d that maybe we need a better name. "Borderless windowed" or "Borderless windowed fullscreen" seem common. But it seems like there's actually *another* option in having border less applied to *either* a regular window or a maximized one (I was thinking "Borderless maximized window" to differentiate it from true full screen.

I noted that there is a new boolean being added to the config. Maybe we *could* use two booleans but just one is "borderless" and applies to either maximized or regular windowed mode?

Cervator:
Catching back up with PRs and testing this again - @coty91 I'm hitting an odd issue with this now seemingly defaulting to fullscreen despite the config (and even default config after deleting `config.cfg`) being set to false.

On first opening the game it'll jump straight into fullscreen. Even if you change to windowed in-game, exit, reload game, starts up in fullscreen again.

It looks like the `displayModeSetting` starts out as "${engine:menu#video-fullscreen}"

While looking around I noticed `menu.lang` was missing the new `video-windowed-fullscreen` key and the value for it in `menu_en.lang` was "Window Fullscreen" instead of "Windowed Fullscreen" - the latter maybe because the full value makes the line break. We might need to widen that UI element just a little bit.

Could you confirm?

coty91:
I did in fact set the menu language string to Window Fullscreen prevent it from breaking the line. I'll have time today to look into this.

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/DisplayModeSetting.java
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.nui.layers.mainMenu.videoSettings;
+
+import static java.lang.System.in;
+
+/**
+ */
+public enum DisplayModeSetting {
```
skaldarnar: This could use some javadoc imho. You should shortly mention what the display modes are (take it from the discussions about borderless/fullscreen/...).

GooeyHub:

GooeyHub:

GooeyHub:

PR #2522, opened 2016-10-10
Fix for #2518

### LINKED ISSUE #2518: Bulk version of giveBlock
We have a `search` console command that is great at finding stuff related to a particular topic, like "cake" with the CakeLie module enabled. However, you then have to type out or copy paste each block if you want to eat *all* the cake!

Probably it wouldn't be hard to combine the two, create a `bulkGiveBlock` command that takes a search term, feeds that to `search`, filters out only the block hits, then executes `giveBlock` once for each search result, offering each output to the player's inventory (until inventory is full if there are too many hits)

Potential extra would be also having a version for `giveItem` or maybe find a way to complete #2293 first then add a single bulk variant. Also potentially related: #2430, #2293

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38231872-bulk-version-of-giveblock?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nihal111:
### Contains

Fixes #2518

### How to test

- Open Console
- type `bulkGiveBlock cake` to get all cake blocks :smile: 
- type `bulkGiveBlock core` to get a larger list and see what happens when inventory gets full.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
@@ -234,7 +234,7 @@ public class BlockCommands extends BaseComponentSystem {
 
     }
 
-    @Command(shortDescription = "Add multiple blocks to inventory",
+    @Command(shortDescription = "Gives multiple stacks of blocks matching a search",
             helpText = "Adds all blocks that match the search parameter into your inventory",
             runOnServer = true, requiredPermission = PermissionManager.CHEAT_PERMISSION)
     public String bulkGiveBlock(
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
@@ -17,6 +17,7 @@
 package org.terasology.world.block.entity;
 
 import com.google.common.base.Joiner;
+import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import org.terasology.assets.ResourceUrn;
 import org.terasology.assets.management.AssetManager;
@@ -60,7 +61,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-
+import java.util.stream.Collectors;
 
 @RegisterSystem
 @Share(BlockCommands.class)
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
```diff
@@ -275,18 +307,18 @@ public String giveBlock(
                 throw new IllegalArgumentException("No block found for '" + uri + "'");
             }
         } else if (matchingUris.isEmpty()) {
-        	return suggestItemIfAvailable(uri);
+            return suggestItemIfAvailable(uri);
```
Cervator: Nice catch

nihal111: Autoformatting ftw :smile:

GooeyHub:

Cervator:
:cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake:

Cervator:
Just for reference I added `bulkGiveBlocks` to the docs for better exposure. Best to be open about how to get more cake,faster!

engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
```diff
@@ -233,6 +234,37 @@ public void replaceBlock(
...
+            @Sender EntityRef sender,
+            @CommandParam("searched") String searched) {
+        String searchLowercase = searched.toLowerCase();
+        List<String> blocks = findBlockMatches(searchLowercase);
+        String result = "Found " + blocks.size() + " block matches when searching for '" + searched + "'.";
+        if (blocks.size() > 0) {
+            result += "\nBlocks:";
+            for (String block : blocks) {
+                result += "\n" + block;
+                giveBlock(sender, block, null, null);
```
nihal111: @Cervator, I could easily change this to allow a numberOfBlocks user input and give those many per block. Do we want that?

Cervator: I would want *both* just like with giveBlocks. Just have a variant that takes a second parameter indicating count :-)

You could even add support for the shape option, but that's going a little far. Bulk amounts of staircase-shaped cake!

nihal111: I don't know how the shape option works. Could you help me with that?

Cervator: I think it is probably overkill, but that's fourth parameter right there, the second null. You can check out how it works by simply playing with giveBlock :-)

`giveblock stone 99 stair`

Should give you 99 stone stair-shaped blocks and is an example from the Playing doc ... although after fixing it as the doc has the quantity and shape switched ;-) Fixing next time I push something.

Issue #1690, opened 2015-04-25
Bug - PlayerStartingInventorySystem for some reason doesn't give items (just blocks)


MarcinSc:

Cervator:
More intel:

org.terasology.ligthandshadow.componentsystem.controllers.LASSystem#onPlayerSpawn

The three items from Pathfinding work

The three items from LASR do not (and are not qualified with the module name)

They're given in different ways and one set is unstackable while the other stackable, for what it is worth

Cervator:
Bump - just for reference this issue is still valid. I tried the Light & Shadow scenario after merging #1988 since that one dealt with module prefixing of assets too, but no change (expected, since that change was specific to "auto" blocks). But maybe there is something similar at work here.

nihal111:
@Cervator Will look into this. Could you write down the steps to reproduce this?

Cervator:
@nihal111 easiest is simply try Light & Shadow

1. Start a game client, host a multiplayer game with the Light & Shadow gameplay template. Use seed "donthurtme" (some seeds in L&S can lead to the player spawning outside the dome, usually resulting in an infinite death loop)
1. Connect to the server with a different game client 

You may want to also activate CheatsForAll so you can experiment with something like `giveItem clubsAce` and `giveItem lightandshadowresources:clubsAce` for the second game client

* The host player from step 1 will have a near full toolbar (slot 1 is empty)
* The client player will be missing the "suit" items in slots 7-10

Oddly, the client player when picking up items will first fill toolbar slot 1 (truly empty) then further items will skip slots 7-10 and go straight to the main inventory. Yet you can place items on top of slots 7-10 and you won't get anything back like you would if there were items there.

It seems like those four items are not fully created somehow, when given to a client player (without the authority (server) present)

The `LASSystem` class is adding the items differently. three different ways no less. They all use `GiveItemAction` which is deprecated, probably there's a better/newer way to do it.

* First three are fully qualified items (not blocks) and work
* Next two are blocks looked up via the BlockManager, and work
* Next four are items that are *not* fully qualified - just plain "clubsAce"

I wonder if that matches entirely with @MarcinSc's use cases. Blocks are fine, items are not if they're not qualified (although maybe they work even unqualified if they come from Core?)

Might also relate to #2430 and/or #1619 although it isn't exclusively happening from console commands (since the starting items are affected as well)

Cervator:
[...]

PR #2525, opened 2016-10-10
Fixed #1988 auto blocks in-game console

### LINKED ISSUE #1988: "auto" blocks unavailable via in-game console, blocks also missing from listBlocks [$5]
I'm not sure if this is an asset or console bug. Maybe console as auto blocks *do* work if used from `PlayerStartingInventorySystem` to grant said auto blocks to a player. You start with a stack of companion cubes (in a chest), which is an auto block, but you can't do `giveBlock companion` in the console

Console command "listBlocks" also doesn't seem to be catching all blocks, auto or not. For instance with Minerals enabled (which has a large amount of regular blocks + auto blocks) not a single one gets listed by listBlocks. While that might be because none have been assigned a block id yet (lazy assigning?) we should probably still list their generic name.

Pinging @Limeth if around and interested :-)

<bountysource-plugin>
---
There is a **[$5 open bounty](https://www.bountysource.com/issues/27657441-auto-blocks-unavailable-via-in-game-console-blocks-also-missing-from-listblocks?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nihal111:
### Contains

Fixes #1988 

### How to test
- Start a game
- Console command `giveBlock companion` (or any other auto block like dirt [BuilderSampleGameplay:Dirt] etc) should now give you the block! :smile:

### What I gathered
- with ` forEach(tile -> result.add(tile.getUrn().getResourceName()));` the result was "**companion**". However when this was passed back to the giveBlock method inside `BlockCommands.java` (Refer: [this](https://github.com/MovingBlocks/Terasology/blob/3252f9ec05241ac370430abdf2118b771a6d590e/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java#L277)), the matchingUris set from `Set<ResourceUrn> matchingUris = Assets.resolveAssetUri(uri, BlockFamilyDefinition.class);` now contained "Companion:companion". As pointed out by @Waterpicker.
- with ` forEach(tile -> result.add(tile.getUrn().getModuleName()));` instead, the result was "**core**". However when this was passed back to the giveBlock method inside `BlockCommands.java`, the matchingUris set from `Set<ResourceUrn> matchingUris = Assets.resolveAssetUri(uri, BlockFamilyDefinition.class);` now contained "core:companion".

Don't really know what's happening inside resolveAssetUri, but this one word change (which took a lot of detective work, phew!) fixes it! :laughing:

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/loader/AutoBlockProvider.java
+++ b/engine/src/main/java/org/terasology/world/block/loader/AutoBlockProvider.java
@@ -60,7 +60,7 @@ public class AutoBlockProvider implements AssetDataProducer<BlockFamilyDefinitio
         assetManager.resolve(resourceName.toString(), BlockTile.class).stream()
                 .map(urn -> assetManager.getAsset(urn, BlockTile.class).get())
                 .filter(BlockTile::isAutoBlock)
-                .forEach(tile -> result.add(tile.getUrn().getResourceName()));
+                .forEach(tile -> result.add(tile.getUrn().getModuleName()));
         return result;
     }
 


nihal111:
Have a look @Cervator!

GooeyHub:

Cervator:
Thanks so much for figuring this out! I'm out of time early tonight so I'll have to sneak out without doing any serious PR review, but this one in particular is a big won. Much appreciated :-)

Cervator:
Tested out and merged! Glad this was an easy one, and really good to have it fixed. 

Don't forget there is a grand $5 bounty on this one, if you want to claim it :D Can always just roll it onto some other issue if it is too much hassle to cash it in (no fees unless cashing out)

https://www.bountysource.com/issues/27657441-auto-blocks-unavailable-via-in-game-console-blocks-also-missing-from-listblocks?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github

Cervator:
Oh and @nihal111 if you want a similar issue that might be more of a challenge check out #1690 - it *might* be caused by the same sort of problems as some items fail to work (or at least fail to produce a visible item, sometimes slots are taken up by something invisible) specifically depending on whether the module prefix is present or not.

nihal111:
@Cervator Bountysource does not seem to update very often. Still shows this issue as open. I did check in as a "BountyHunter" and refrenced this PR, though.

Cervator:
Yeah I think there are only two devs and they don't make enough money off it to make it their job, so their availability can be spotty. Methinks they need more automation! Everybody does :-)

Issue #1697, opened 2015-04-27
Items are not being destroyed when dropped into lava


ghost:
<edit: original description lost or left empty>

sergiomieic:
@Cervator I'm one of the Porto students and at this point (speaking on behalf of my group) I would like to contribute (at least try) correcting some kind of bug, this one seems to be feasible, but I would like if posible to get some more information about the packages related to this bug in order to easily find the target area of the bug issue

Cervator:
Heya @sergiomieic - sure :-)

So this would be a new feature, and actually an optional one you could choose to enable by activating a module containing the functionality. It shouldn't be an engine feature at this point.

A similar existing module is Breathing: https://github.com/Terasology/Breathing

Take a look at that module (use gradlew fetchModuleBreathing to add it to an existing workspace) and see how it determines how the player enters and exits water to trigger "holding your breath" or eventually suffocating. You'd likely be looking at doing something similar for players/items that enter lava.

Very similar would be something like having a player or creature take damage when you get too close to a cactus. You probably would want an ability to mark blocks as causing damage in some fashion. The breathing system was designed to eventually do that as well, but may be hard coded to water at present. 

Really, a human entering water might be unable to breathe, but an amphibian would be okay. Both would be burned in lava *and* suffocate at the same time. Imagine having a potion making you immune to fire damage - now you wouldn't burn in lava, but you'd still be unable to breathe.

A good first step: Post a suggestion thread at http://forum.terasology.org/forum/suggestions.21/ and we'll talk more design there first. Talk among yourselves on how you would enable a many-to-many system where players/creatures/items could take damage from any special blocks defined as being hurtful to them or a category in general.

Start simple, then build on it. And while you're at it maybe see if you can enhance Breathing in the same way as they're conceptually similar. Different players/creatures would potentially have to hold their breath in different kinds of situations. Water, vacuum, a tar pit, etc. Maybe a tar monster can breathe fine in a tar pit while a player (or fish) cannot :-)

I am also noticing the bubble icons in Breathing are broken. You hold your breath and the breath meter goes down by the icons are messed up. That could be the first quick bug to fix right there!

I can get you access to the Breathing module if you would like to work on it, and/or create a new one for blocks that'll hurt you, like lava. Maybe we can call it "HurtfulBlocks" :D

sergiomieic:
Thanks for the great reply @Cervator :D 
[...]

Issue #1443, opened 2014-12-16
"Sticky" thrown objects (especially torches)


Cervator:
On entering one of the large churches in a City world I realized how handy it would be to be able to throw some sticky torches up high on the walls to get some proper lighting. Not that we'd want the normal torches to work that way, but it seems to be a useful option to support. I know it can work as the arrows in the combat branch worked that way - one type you'd shoot into a wall and they'd stick there.

Nice little separate item that can be done any time, probably without too much difficulty for somebody familiar with game physics a bit.
##

portokaliu:
On to porting the behavior to a module... right after some Rust ^^

Cervator:
We have sticky arrows in CombatSystem now - does that count as this being complete, or do you still have sticky torches hiding somewhere @portokaliu ? :D

jdrueckert:
It's not entirely clear here, whether the requested feature is a way to persistently place objects like torches by throwing them against a surface or a way to temporarily place them by having them stick in the target object using physics before they despawn later-on (like the "sticky" arrows in CombatSystem).
In the org meeting it was agreed-upon, that this neither is a "Good First Issue" for new contributors, nor has enough detail to keep around until gameplay content feature requests can actually be prioritized again.

Issue #1852, opened 2015-07-26
Dropped items from blocks oddly large and staying upright?


Cervator:
At least in the case of dirt blocks dropping stones in TTA - they used to be smaller and/or fall over, now they just sort of stand there:

![terasology-150726163143-1152x720](https://cloud.githubusercontent.com/assets/1063833/8895786/edb7e13c-33b3-11e5-953b-4fbd93973cb6.jpg)

Sticks dropping from trees also seem oddly specific on how they drop/land on a flat surface:

![terasology-150726163347-1152x720](https://cloud.githubusercontent.com/assets/1063833/8895794/2afae030-33b4-11e5-961b-b9cba8f0a7a0.jpg)

I'm not sure if they're actually larger or just looking that way from landing funny. If there is more than one item then drops happen more normally and random (like two stones in one dirt block causes them to fall over, a stick falling on top of another stick also makes sense)

@Josharias maybe something for you?
##

Cervator:
Sending this @portokaliu's way since it still seems active yet closely relates to some of the physics stuff around dropping/throwing things. I figure as a default anything dropped should have a slight (slight!) bit of random impulse added to it.

As a good example compare grass drops in Core to JS. In Core the actual billboard plant drops but has a random impulse added to it so it shoots off into a random direction (maybe even a little too forcefully, but maybe that's more about them being too low mass). In JS instead grass drops organic matter items, which just plop into existence standing upright with no random motion added at all.

Issue #206, opened 2012-03-10
Stone dust


t3hk0d3:
They basic idea what stone should leave something after it has been broken.

After breaking rock block there is 1/7 of stone dust created, which have very high viscosity liquid dynamics.

Rock dust is penetrable unti 4/7, but reduce player movement speed slightly.

Player could collect stone dust using shovel and bag, and unload it into minecart.

Cervator:
Closing in favor of a bullet point in the backlog thread: http://forum.terasology.org/threads/backlog.1101/

PR #2533, opened 2016-10-12
#2139 Added item impact events when hitting world

### LINKED ISSUE #2139: Able to throw some items through blocks at just the right angle
Not a particularly high priority as it is hard to reproduce and easy to work around if you for some reason didn't mean to stick some item a couple blocks into the ground :-)

Grab a small *item* like the dynamite or railgun or blueberries from JS or so on. Throw it with `q` with some force into the horizontal split between two blocks where one is ground level and the other sticks up one block. With enough attempts/practice you can make the item squeeze between the blocks somehow and get stuck one or even two blocks deep before it loses momentum and does the little vibrating thing from being stuck between solid objects (can see by ghosting through the world)

Probably has something to do with items more or less being 2D sprites, but not quite since they do have some minimal depth? Maybe they only look like they have depth in the world and are treated like flat 2D objects by the physics, so if hitting edge on between blocks they can squeeze through? Thus the need to throw it into a corner between blocks so the item will bounce and arrange itself in just the right direction as it hits.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30033939-able-to-throw-some-items-through-blocks-at-just-the-right-angle?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

portokaliu:
Contains:

This PR fixes #2139 by making any item that hits the ground (hard enough to register) send an event that is then caught, processed and gives the item a set speed upwards which prevents it from tunneling. 

By doing so, I've seen improvements in framerate when using the railgun and had the comical effect of some blocks bouncing around before they settle. 

This also puts the building blocks down for creating the StickySystem for #1443 and _could_ help with #1697

How to test:

Throw any item from any height onto any piece of land. They will most likely not go through them.

Possible improvements I need feedback on:
- [ ] Consider giving the item a physical reaction by giving it a reduced speed on the vector opposite of the collision normal. I do still like how they bounce though :)
- [ ] Adjust the space items are thrown from by the player (when throwing items directly into a wall in front of the player, they are thrown from the same relative point to the player, which in this case is already inside the wall. This can be fixed by changing said position withing a max range of the direction the player is looking, but I'm not sure if we want that).
- [ ] Consider sending more information throught the event. (probably the other entity?)

### DIFF
--- a/engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
+++ b/engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
@@ -29,7 +29,6 @@ import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.entitySystem.systems.RegisterMode;
 import org.terasology.entitySystem.systems.RegisterSystem;
 import org.terasology.entitySystem.systems.UpdateSubscriberSystem;
-import org.terasology.logic.location.Location;
 import org.terasology.logic.location.LocationComponent;
 import org.terasology.logic.location.LocationResynchEvent;
 import org.terasology.math.geom.Quat4f;
@@ -41,14 +40,18 @@ import org.terasology.physics.HitResult;
 import org.terasology.physics.StandardCollisionGroup;
 import org.terasology.physics.components.RigidBodyComponent;
 import org.terasology.physics.components.TriggerComponent;
-import org.terasology.physics.events.*;
+import org.terasology.physics.events.ChangeVelocityEvent;
+import org.terasology.physics.events.CollideEvent;
+import org.terasology.physics.events.ForceEvent;
+import org.terasology.physics.events.ImpulseEvent;
+import org.terasology.physics.events.PhysicsResynchEvent;
+import org.terasology.physics.events.ImpactEvent;
 import org.terasology.registry.In;
 import org.terasology.world.OnChangedBlock;
 import org.terasology.world.block.BlockComponent;
 
 import java.util.Iterator;
 import java.util.List;
-import java.util.Vector;
 
 /**
  * The PhysicsSystem is a bridging class between the event system and the
@@ -167,7 +170,7 @@ public class PhysicsSystem extends BaseComponentSystem implements UpdateSubscrib
                 float fDistance = vDirection.length();
                 vDirection.normalize();
 
-                HitResult hitInfo = physics.rayTrace(vLocation, vDirection, fDistance * delta, StandardCollisionGroup.WORLD);
+                HitResult hitInfo = physics.rayTrace(vLocation, vDirection, fDistance * delta, StandardCollisionGroup.WORLD, StandardCollisionGroup.CHARACTER );
                 if (hitInfo.isHit()) {
                     entity.send(new ImpactEvent());
                 }
--- a/engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
+++ b/engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
@@ -29,6 +29,7 @@ import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.entitySystem.systems.RegisterMode;
 import org.terasology.entitySystem.systems.RegisterSystem;
 import org.terasology.entitySystem.systems.UpdateSubscriberSystem;
+import org.terasology.logic.location.Location;
 import org.terasology.logic.location.LocationComponent;
 import org.terasology.logic.location.LocationResynchEvent;
 import org.terasology.math.geom.Quat4f;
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
```diff
@@ -29,26 +29,26 @@
...
+import org.terasology.physics.HitResult;
+import org.terasology.physics.StandardCollisionGroup;
 import org.terasology.physics.components.RigidBodyComponent;
 import org.terasology.physics.components.TriggerComponent;
-import org.terasology.physics.events.ChangeVelocityEvent;
-import org.terasology.physics.events.CollideEvent;
-import org.terasology.physics.events.ForceEvent;
-import org.terasology.physics.events.ImpulseEvent;
-import org.terasology.physics.events.PhysicsResynchEvent;
+import org.terasology.physics.events.*;
```
portokaliu: Ah, I didn't know there was a setting for that. Will be looking into it :)

engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
```diff
@@ -29,26 +29,26 @@
...
 import org.terasology.registry.In;
 import org.terasology.world.OnChangedBlock;
 import org.terasology.world.block.BlockComponent;
 
 import java.util.Iterator;
 import java.util.List;
+import java.util.Vector;
```
portokaliu: Ah yes, I thought my ctrl+z did the job for me. I was apparently wrong.

portokaliu:
The bounciness was more of a proof of concept (reason there's a random 4.0f as a reaction speed there :D )

I think the reaction to blocks should be set straightforward. When we will implement materials on blocks we can add the custom reactions by listening to the event (unless we have that already and I'm completely oblivious). 

I'm not sure how a module could influence the main engine yet ( since I'm just starting to dabble into those ), but we should indeed have a way to modify the bounce speed and maybe even direction depending on several factors ( like if we have bounces disabled; In that case we just stop the object before it hits the block at the distance of its bounding circle.... which I'm not sure if we have that yet but I'll be searching for options )

I will change the item throwing position to the closes possible point in to the player that's free. The only downside would be in the case where the player is inside a block when throwing an item, the item will just be placed where the player's eye position is.

Since the raytrace has a specific collider to WORLD, that would mean that only world blocks will be returned on impact. Would we want to extend this behaviour to any type of entity? We could use the current collision to damage said block ( if we fire a couple of arrows at a block of wood, we'd break it [if it doesn't regen - then maybe we make a survival where you have to repair your house blocks and enemies can break them] )

I'm not sure on #1852, since it doesn't seem to have anything to do with going through blocks, and I haven't encountered that specific issue. What we can do in turn is use rng to give the item that comes out of a broken block a random rotation that's not 0 ( just so it doesn't think it's a perfect little princess and can stay upright )

Cervator:
On the world collision bit - as in, you mean adding support for throwing events when thrown objects impact other stuff like deer? That sounds like it would slide into possible combat uses or other fun stuff (throw an axe at a deer instead of whacking at it with said axe). Accidental combat when throwing companion cubes at Gooey? :-)

I would *love* to see arrows and such damage soft enough materials, and in some cases let damage persist so you need repair. Which seems easy enough to enable, only the point there in part was to "heal" blocks to where they no longer needed a temp entity to track damage. But really, you could just spend a few more block ids to get cracked wood/stone blocks with lower HP, swapping into those rather than fully heal (blocks dropped due to explosions could also be cracked variants). I've got an [old design](http://forum.terasology.org/threads/the-night-is-dark-and-full-of-terrors.1067) that includes having your shelter slowly get broken down at night :D

Good point on #1852, I didn't think about the collision case varying, so to say. Last I tried it you would get it in single player on breaking any leaf blocks in TTA or JS, with sticks and such always landing perfectly the same way, and yup just adding a little random impulse addition would fix it. I haven't verified it very recently, it might have been disrupted already.

portokaliu:
Ah yes, completely forgot about them having to hit entities as well. I'm going to have to check for more than just world entities then (since currently only world blocks are hit). Just tried it with the Deer from WildAnimals and with the change it can hit their (rather blocky) hitbox. 

I've chatted a bit with mannu3d about what approach we can do to basic shapes for entities and there are some things to look into regarding that. Still a lot to research though.

I totally understand what you're trying to say with throwing items into Gooeys and such, and that might need some tuning in its own module (or maybe used in the projectile Module that I had in mind a while ago? )

On the note of using other modules, is there a more direct way to fetch a module together with all its dependencies? I've had a few issues when trying to get WildAnimals, mostly having to gradlew fetchModule all the dependencies and their dependencies manually before getting this one, otherwise I'd get errors that modules are missing without actually fetching them. A bit odd...

We could do the damaging of blocks, but this would have to go into a PersistentDamage module (unless there is one) where all the assets are stored and blocks are replaced with inferior ones when reaching a certain amount of health as you said. I did look around the idea mentioned there and we could do weather as well (I also looked through the general issues and there was something related to snow that seemed pretty interesting; Penetrable blocks and all)

I not sure what you mean by it being disrupted, but the fix would be easy and on engine side nontheless.

Cervator:
Good bits of discussion on Slack :-)

On disruption I just meant some other physics-related changes or something similar may already have made item drops actually fall over more randomly, I haven't checked lately.

And yeah likely several modules could come out of this kind of brainstorming :-) There isn't a module anywhere that does replacement of "weaker" blocks yet. The WeatherManager module *does* have hooks for different kinds of weather, so including snow and having snow blocks appear would be a great addition. And yeah that's an interesting block to travel through.

Mildly related to ancient (closed/backlogged) issues #206 and #204. Where does one really draw the line between a solid and a liquid? :D

GooeyHub:

GooeyHub:

Cervator:
Checked on this again. I guess I have to manually approve review changes despite them all having now been hidden as obsolete (and fixed)

Am going to merge this so we can move forward with one caveat: the bouncy factor is still a bit goofy (especially when you drop/throw torches) but really that could just be a fun little thing to leave in for a bit. I take it you have more tweaks coming anyway @portokaliu :-)

Thanks! Go forth and do great things!

portokaliu:
I wanted to change this on some more by the directions you said ( reflections off walls and stuff ) but I've had a few health issues the last couple of days and couldn't even concentrate on anything ( also not a fun way to take days off from work ). I'll change this tonight most likely to what it should be :)

Cervator:
Doh! Sorry to hear that.

I just saw more commits and the review notes now obsolete, finally went for it so our pile of PRs can shrink. Everything works fine so could just continue next PR :-)

PR #2540, opened 2016-10-14
giveBlock return instead of exception

nihal111:
### Contains

Extended fix for #2518 

### How to test

Try `bulkGiveBlock grass 10 sphere` with and without applying this patch

giveBlock now returns string instead of throwing a new exception. So if in a list of 10 blocks that contain "grass" in their name, one in the middle not having a sphere shape available will not prevent the player from getting the rest.
The red text on console looked good though :stuck_out_tongue:, won't get that any more.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
@@ -249,8 +249,8 @@ public class BlockCommands extends BaseComponentSystem {
         if (blocks.size() > 0) {
             result += "\nBlocks:";
             for (String block : blocks) {
-                result += "\n" + block;
-                giveBlock(sender, block, quantityParam, shapeUriParam);
+                result += "\n" + block + "\n";
+                result += giveBlock(sender, block, quantityParam, shapeUriParam);
             }
         }
         return result;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Tests out fine! Merging :-)

Feel free to tweak at the console logging some if you end up in that area again  - I think it could be clarified a bit. For instance the billboard grass blocks successfully were given, yet they are shapeless so it isn't like they were successfully requested in there sphere shape. The grass block failed since it *could* take a shape, but "sphere" wasn't found, thus no blocks were given.

That's kinda inconsistent, I wonder if we should log the shapeless blocks as invalid for a shaped command, then not give those either.

I'm happy to avoid the error in either case, progress! :-)

skaldarnar:
> That's kinda inconsistent, I wonder if we should log the shapeless blocks as invalid for a shaped command, then not give those either.

Requesting a shapeless block with a specific shape should indeed be invalid. I wonder what the difference between "shapeless" and "restricted to one specific shape" is. Or is it actually the same?

Cervator:
So on thinking a bit more I'm realizing something:

* I was thinking billboard plants were invalid to have a shape because they were billboards - but come to think of it, that's just a shape :D Nothing special about billboards shape-wise IIRC..
* Blocks defined with *no* shape automatically becomes eligible for *any* shape
* Blocks defined with an explicit shape cannot be requested in any other shape - if you request a ladder block in a `engine:testSphere` you just get more standard ladder blocks (heck you even just get more standard ladder blocks if you ask for a non-existing shape)

PR #2543, opened 2016-10-14
Made a resettable UIText widget with a button for clearing text.

### LINKED ISSUE #2507: Add option to clear the filter textbox on the Select Modules... screen
At present during game creation you can customize modules, and while picking modules you can filter them by name using the filter textbox at the top. However, you have to manually erase the text if you want to clear the filter.

While that isn't particular laborious it is pretty standard for filter boxes to have a simple clear option of some sort. Two ways to solve this one:

* Easy: Simply add a small independent "Clear" or "X" button by the filter label or textbox. It should clear the filter text, with the module list updating to reflect the cleared filter
* Challenging but more useful: Make a new UI widget, a variant of  `UIText` maybe named `ResettableUIText` that has a tiny little built in `x` at the right side within the text box. Clicking the `x` clears the text and triggers any regular related events as if the user had erased text (`updateOffset()` maybe among other things?)
 * Extra: If this approach is taken also add the new UI element to the UI test screen (see in-game via console using `showScreen migTestScreen`) below `UIText`

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38207267-add-option-to-clear-the-filter-textbox-on-the-select-modules-screen?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

arpan98:
### Contains

Fixes #2507 using the second approach mentioned in the issue, making a new widget `ResettableUIText` based on the `UIText` widget. Also added the widget to the `migTestScreen`, so it can be seen in different states using the console command `showSreen migTestScreen`.

On clicking the button, the text is cleared but focus is lost from the `UIText` box. Any suggestions for a workaround?


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

Thanks for this @arpan98 along with the other two PRs! :-)

Aiming to merge those now since they're easy, maybe @rzats can take a look at this one to provide some feedback

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/SelectModulesScreen.java
```diff
@@ -45,11 +45,7 @@
 import org.terasology.rendering.nui.databinding.Binding;
 import org.terasology.rendering.nui.databinding.ReadOnlyBinding;
 import org.terasology.rendering.nui.itemRendering.AbstractItemRenderer;
-import org.terasology.rendering.nui.widgets.TextChangeEventListener;
-import org.terasology.rendering.nui.widgets.UIButton;
-import org.terasology.rendering.nui.widgets.UILabel;
-import org.terasology.rendering.nui.widgets.UIList;
-import org.terasology.rendering.nui.widgets.UIText;
+import org.terasology.rendering.nui.widgets.*;
```
rzats: No star imports, please. (Go to _Settings → Editor → Code Style → Java → Imports_ and set _"Class count to use import with '*'"_ along with _"Names count to use static import with '*'"_ to 999)

engine/src/main/java/org/terasology/rendering/nui/widgets/ResettableUIText.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+package org.terasology.rendering.nui.widgets;
+
+import org.terasology.math.geom.Rect2i;
+import org.terasology.rendering.nui.Canvas;
+import org.terasology.rendering.nui.SubRegion;
+import org.terasology.rendering.nui.UIWidget;
+
+public class ResettableUIText extends UIText {
+    private UIButton clearButton = new UIButton("clearButton", "X");
```
rzats: This is a working approach, but IMO Terasology's default button doesn't look great when used as a small control element.

`UIDropdown` handles control elements slightly differently - it uses a texture ([dropdown.png](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/textures/ui/dropdown.png)) the right part of which masquerades as a button. I would use this idea for `ResettableUIText` as well:

- Create a modified version of [box.png](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/textures/ui/box.png) (the texture that `UIText` uses as a background) similar to dropdown.png. The left 80% of it should be the same as box.png, but the right 20% will contain a reset button. (Different percentage values are also fine - use what works best!)
- In `ResettableUIText.onDraw()` separate the canvas region into two parts:
  - The left (80%) will display the current text and place the cursor at a specified position when clicked on (same as the default `UIText`).
  - The right (20%) will reset the text field when clicked on. (This should also fix the issue mentioned in the PR as the "button" won't make the text field lose focus!)

engine/src/main/java/org/terasology/rendering/nui/widgets/ResettableUIText.java
```diff
@@ -0,0 +1,62 @@
...
+    public void onDraw(Canvas canvas) {
+        //super.onDraw(canvas);
+        if (super.text.get() == null) {
+            super.text.set("");
+        }
+        super.lastFont = canvas.getCurrentStyle().getFont();
+        super.lastWidth = (int) (canvas.size().x * 0.8f);
+        if (isEnabled()) {
+            canvas.addInteractionRegion(super.interactionListener, Rect2i.createFromMinAndMax(0, 0, (int) (canvas.size().x * 0.8f), canvas.size().y));
+            clearButton.subscribe(new ActivateEventListener() {
```
rzats: This should be lambdified:

```java
clearButton.subscribe(button -> setText(""));
```

IDEA should prompt you to convert anonymous inner classes to lambdas - do this whenever it does 🙂

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
```diff
@@ -68,27 +68,27 @@
...
     @LayoutConfig
-    private Binding<String> text = new DefaultBinding<>("");
+    protected Binding<String> text = new DefaultBinding<>("");
 
     @LayoutConfig
-    private boolean multiline;
+    protected boolean multiline;
 
     @LayoutConfig
     private boolean readOnly;
```
rzats: Consider setting `UIText`'s other private variables and methods (except `logger`, `BLINK_RATE`, `blinkCounter` and `cursorTexture` as they are unlikely to be changed by a child class) to `protected` - don't just make half of class extensible! 😛

engine/src/main/java/org/terasology/rendering/nui/widgets/ResettableUIText.java
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+package org.terasology.rendering.nui.widgets;
+
+import org.terasology.math.geom.Rect2i;
+import org.terasology.rendering.nui.Canvas;
+import org.terasology.rendering.nui.SubRegion;
+import org.terasology.rendering.nui.UIWidget;
+
+public class ResettableUIText extends UIText {
```
rzats: A short multiline comment explaining what the class does would be nice here.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
arpan98: Should I redefine the methods as well? They do have the exact same functionality as in the parent class and will lead to a lot of code repetition.

rzats: No, not at all - just change the access modifiers to `protected`! (The reason for this is that a programmer in the future might want to create their own `UIText` override, and to them it might not be obvious why some fields are marked as `protected` but others aren't.)

engine/src/main/java/org/terasology/rendering/nui/widgets/ResettableUIText.java
arpan98: Nice idea! I'm having some trouble finding where these textures are set to the corresponding UI elements. Any pointers?

rzats: [default.skin](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/skins/default.skin) 🙂 

You'll need to add a new entry looking something like this:

```json
"ResettableUIText": {
    "background": "engine:yourTexture"
}
```

arpan98: A fixed percentage display will look really weird when resized to long lengths. How can I keep the width of the button region fixed like it is in `UIDropDown` and `UIDropDownScrollable`?

GooeyHub:

arpan98:
@rzats I added a default texture and one for disabled as well as the other changes. Have a look.

rzats:
@arpan98: the new widget works great - thanks! 🙂 
Merged.

Issue #1075, opened 2014-04-20
Enhance crouching functionality [$5]


Cervator:
I've moved the oft-forgotten "Crouch" key to CTRL from 'c' although it doesn't actually do any kind of crouching. It does "descend" when you're flying or swimming though.

It would be nice if actual crouching was added. I imagine this is a state that'll render a model in a different pose (if available) and for players change their apparent view height downwards just a bit.

If the player is crouching a nice extra would be prevention from falling off ledges, akin to in MC
##

MaloJaffre:
#2115 can solve the problem of the view height easily.

oniatus:
Follow up for crouching.
With #2961 merged the crouching will not change the camera or the player height but work like a sneak - but it works :)

A clean implementation for crouching can be tracked on this issue.
Actual list of requirements for crouching:

- Multiplayer support (host & connected client)
- 1x1 holes enterable
- Visual Feedback in eye height
- Falling of ledges support
- Toggle/Untoggle + temporary button (as is with this PR)
- Leaves 'LocalPlayerSystem' cleaner instead of making it worse.

PR #2544, opened 2016-10-15
Crouch functionality added

nihal111:
### Contains
Potential fix for #1075 

### How to test

Just hit LCtrl when in game. Keep it pressed to crouch.

There is a "blink" during crouching due to detach and attach of camera with different gazeMountPoint. Possibly can be fixed when a crouching animation is added.
The crouching factor (by which height changes) is kept as 0.5 now.
While crouching I can go in doors of length of 1 block, but on leaving LCtrl, I rise, and while doing so go inside the block above me. (FIXED)

- [ ] Prevent falling off ledges like mc.
- [x] Display message "Cannot stand here!" when crouched under something. (Happening in chat panel now)
- [x] Implement crouch toggles.


GooeyHub:

emanuele3d:
I think Crouching is an important, expected functionality. I am very grateful that you are looking into it nihal. 

But I wouldn't want to miss the opportunity to do it right. Ending up inside the cube above is certainly something requiring fixing: it "breaks" the game. Sounds like the collision capsule around the character needs resizing while crouching.

Also, preventing falls would be quite a bonus if you could manage. I think @coty91 was recently looking at ways of detecting the blocks near the player. It might be easy to check if the block in the direction of motion is below the player and prevent further movement in that direction when reaching the border of the current block.

Finally, crouching should probably slow down the player: 1/2, 1/3 speed perhaps?

nihal111:
Max speed can be adjusted very easily. And yes, I will look into checking blocks around to implement the mc functionality.
As for the collision capsule resizing that's exactly what I am doing, but when I change my stance to walking, the resizing happens again and now I go inside the block above me. 
Basically I need to display a "cannot stand here" when I release the LCtrl and there is a block above me. After that I'll have to check whenever the player comes out of the enclosure and then make him stand.

GooeyHub:

nihal111:
@emanuele3d Rising is prevented while crouching now, if there is something above. :smile: 
Need to know how I can display a "Cannot stand here" message a corner in the screen. Could you help me with that?

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

Zireael07:
One thought: 'keep pressed to crouch' is quite an antiquated way to do it. Most games do it nowadays as a toggle. Press Ctrl to enter crouch, press Ctrl to exit crouch. Same for run mode.

GooeyHub:

Cervator:
Reviewed and tested some, with more thoughts at http://forum.terasology.org/threads/crouch-and-prone-controls.1622/#post-14015 :-)

I think the keys will be the trickiest. Looking for more feedback from others, like @Zireael07 there (thanks!)

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -160,6 +170,43 @@ private void processInput(EntityRef entity, CharacterMovementComponent character
...
+        float height = clientComp.character.getComponent(CharacterMovementComponent.class).height;
+        float eyeHeight = gazeMountPointComponent.translate.getY();
+        Vector3f pos = entity.getComponent(LocationComponent.class).getWorldPosition();
+        // Check for collision when rising
+        CharacterCollider collider = physics.getCharacterCollider(clientComp.character);
+        // height used below = standing_height - crouch_fraction * standing_height
+        Vector3f to = new Vector3f(pos.x, pos.y + height, pos.z);
+        SweepCallback callback = collider.sweep(pos, to, VERTICAL_PENETRATION_LEEWAY, -1f);
+        if (callback.hasHit()) {
+            entity.send(new NotificationMessageEvent("Cannot stand here!", entity));
```
Cervator: Tiny clarification: "Cannot stand *up* here" ? Seems to me like it would slightly better Indicate an inability to take an action to stand up there.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Finally merged! Thanks for keeping at it :-)

Also submitted #2557 and put a couple slight addition in #2555

PR #2550, opened 2016-10-17
Fix for NPE at loading save games

nihal111:
### Contains

Fix for NPE at loading save games

### How to test

Manage to kill the game when saving is in progress.
Open the game. There should be a NPE [here](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/savedGames/GameProvider.java#L69)

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/savedGames/GameProvider.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/savedGames/GameProvider.java
@@ -46,7 +46,7 @@ public final class GameProvider {
         Path savedGames = PathManager.getInstance().getSavesPath();
         SortedMap<FileTime, Path> savedGamePaths = Maps.newTreeMap(Collections.reverseOrder());
         try (DirectoryStream<Path> stream =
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
@rzats you managed to validate this right? Can you confirm or even merge this if so please? :-)

rzats:
@Cervator: merging! Though as discussed on IRC I'd look into making the error visible to the user as a followup.

nihal111:
@rzats Will you? I was thinking of looking into it..

rzats:
@nihal111, what I meant is I'd _[...suggest that you...]_ look into making the error visible to the user 😛

Cervator:
Thanks :-)

nihal111:
I found this again and realized I hadn't looked into making the error visible to the user earlier.
Should we probably create save games such that a temp save game file is created which later replaces the original save game file? Since termination of the game midway during creating the save file causes this error, using this approach at least the second most recent save would be available to the user in such an event. @Cervator  @rzats

Cervator:
I added a comment over to #2614 which is still open and very similar so a better place for discussion :-)

PR #2536, opened 2016-10-12
Added more teleport commands

portokaliu:
This change added the (maybe) much needed commands to influence more than the local player and more than one player's position at the same time.

Commands added with this commit would be:
- `teleportMeToPlayer <String>`
- `teleportPlayerToMe <String>`
- `teleportPlayerToPlayer <String> <String>`
- `teleportAllPlayersToLocation <float> <float> <float>`
- `teleportAllPlayersToPlayer <String>`

This in turn fixes #2534 which required some of these changes. The last two commands were added for event convenience ( like when you want to bring all people on the server to a certain location to see something or try something ) 

The commands that influence other player's locations need 'userManagement' permissions, while the ones that only influence your location require 'cheat' permissions.

To test, you can use any of the command while on a server. 

I've only tested this on a local server and it worked well with 2 players (with different permission levels, and both of them using commands)

PR #2560, opened 2016-10-22
Utility: Include a template override.cfg for use by default in headless launched via IntelliJ

Cervator:
This is a tiny developer convenience PR.
- Contains an `override.cfg` file under `/templates` that'll get copied to the project root on `gradlew idea` - an easy way to get a tiny config file prepared for a server run
- The existing `gradle.properties` under `/templates` will now also get copied to the project root automatically, allowing convenient access to a bunch of Gradle config (turns on the Gradle Daemon and some other stuff by default - which is becoming default in Gradle 3 anyway)
- The headless server run config in IntelliJ will now use the `override.cfg` mentioned above by default (if it isn't present the game falls back on default config - no crash)
- There is a 3rd game client run config in IntelliJ so you can easily run _three_ separate game clients (and a headless server!) for deeper multiplayer testing (like having one player teleport two others - handy for testing #2536 by @portokaliu / Anthodeus!)

I've personally had a `gradle.properties` in my project root dir for ages (since I have all kinds of project secrets in there!) and left a placeholder under `/templates` but it isn't obvious that it is there nor what you can do with it (make Gradle do its thing faster or retrieve/publish things to different GitHub orgs / Artifactories). So this helps expose that better.

@qwc might also be interested in this as I got the idea of a placeholder `override.cfg` from his tinkering with the Docker game server image. While this PR doesn't package the `override.cfg` in the game zip yet that could be done and even customized further if needed. Right now the template just matches current defaults (run CoreSampleGameplay with faceted Perlin) and is _only_ activated on headless server launch via IntelliJ

Like the `gradle.properties` I hope having a placeholder `override.cfg` will encourage using it more for developing, as it is _so much easier_ just adding a few modules to it then recreate a server over and over without having to modify `terasology-server/config.cfg` and then lose it each reset unless you carefully avoid deleting it.

For testing:
- Run `gradlew idea` without `override.cfg` or `gradle.properties` in the root dir, make sure they get copied in from `/templates`
- Make sure altered versions of the files in the project root do not get overwritten
- Make sure a server launched via the headless run config in IntelliJ respects custom modules added to `override.cfg`
- Make sure a server launches with _defaults_ without `override.cfg` present in project root (don't want it to crash if it doesn't find a file despite being told to look for one)

@qwc let me know what you think and if it would help to bundle a placeholder `override.cfg` (with something specific configured?) in the game zip. It likely would at least help me when preparing a server to upload somewhere and configure manually.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,9 +25,10 @@ extensions
 /libs/*
 !/libs/subprojects.gradle
 
-# Ignore Gradle, including local override properties (template file available under /templates)
+# Ignore Gradle and some transient project config files (template versions available under /templates)
 build/
-gradle.properties
+/gradle.properties
+/override.cfg
 
 # Ignore IntelliJ - could use "/**/*.iml" to catch all .iml files, but that requires a somewhat recent Git version
 /out/
--- a/build.gradle
+++ b/build.gradle
@@ -187,8 +187,21 @@ task protobufCompileLinux(type:Exec) {
 // General IDE customization                                                                                         //
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+task copyInMissingTemplates {
+    description = "Copies in placeholders from the /templates dir to project root if not present yet"
+    File gradlePropsFile = new File(rootDir, 'gradle.properties')
+    File OverrideCfgFile = new File(rootDir, 'override.cfg')
+    if (!gradlePropsFile.exists()) {
+        new File(rootDir, 'gradle.properties') << new File(templatesDir, 'gradle.properties').text
+    }
+    if (!OverrideCfgFile.exists()) {
+        new File(rootDir, 'override.cfg') << new File(templatesDir, 'override.cfg').text
+    }
+}
+
 // Make sure the IDE prep includes extraction of natives
 ideaModule.dependsOn extractNatives
+ideaModule.dependsOn copyInMissingTemplates
 
 // For IntelliJ add a bunch of excluded directories
 idea {
--- a/config/gradle/ide.gradle
+++ b/config/gradle/ide.gradle
@@ -242,6 +242,38 @@ ext {
             </configuration>
         '''))
 
+        // Run config for a *third* client in its own data dir (for more extensive multiplayer testing)
+        runManager.append(new XmlParser().parseText('''
+            <configuration default="false" name="TerasologyPC (3rd client)" type="Application" factoryName="Application">
+              <extension name="coverage" enabled="false" merge="false" runner="idea">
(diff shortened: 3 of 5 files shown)


GooeyHub:

qwc:
Afaik is the status quo that on the first start of a terasology server the directory terasology-game-dir/server is populated with the config.cfg and at last on the first join the first world pieces are stored in this directory.

As long the same happens with the override.cfg template file there'll be no problem.

But if you want to ship that file already with the zip, there's one problem:
Normally a user wants to mount a docker-volume somewhere in his own file system to be able to change the configuration files in a convenient way. You can also just leave docker decide where to store volume data, but there's no convenient way to access a volume, afaik. It needs at least quite some more commands than just going to the mounted volume and firing up an cmd-line editor.

So if you already ship a file in this server directory, **a mount will shadow this file** and the user won't see it in the directory he chose for the server data on his server. (See linux/unix mount command, and docker documentation: https://docs.docker.com/engine/tutorials/dockervolumes/#/data-volumes )
So the user would have to create it by himself, but when there are helping comments in this file, he wont have that. 

So, if you keep the way it is with the ./server directory being filled on first run, there's no problem to sail around.

So to the override.cfg itself: **IMMER HER DAMIT!** 
(German, meaning: just give/do it!)
Yes, I think a override.cfg with quite some comments for a newbie server admin to understand what he can configure in there would be a hell of a help for setting up new servers. Also the possibility to save the basic config to the config.cfg and fire up multiple servers with slightly different overrides could be an interesting idea, but still would have to search for a usage strategy, at least for testing the game after an unstable build it would be quite a nice thing. :)

Maybe we could encourage some AI devs to build headless clients which build whatever or just try to survive with their AI, and we could use such AI clients to test different important game parts with such multiple configs ... ohooh, my ideas spread again, search for cover!!! ;)

Cervator:
Anybody able to validate the IntelliJ tweaking - @skaldarnar or @rzats sometime maybe?

@qwc on the Docker piece I'm not entirely catching the shadow mounting thing, but in short yes include an `override.cfg` in the game zip for convenience? Although I'm realizing we aren't actually supposed to comment JSON files, I guess our lenient parsing just lets that slide at the moment. So ideally just need the approach documented - which is already is nicely in the PLAYING doc :-)

Could just leave it as a dev-only convenience.

qwc:
@Cervator If created on first start in the terasology/server directory everything is fine. :)

Cervator:
Went ahead and merged this to get it out of the way. We could do some more fun with Docker later on if it adds convenience :-)

PR #2562, opened 2016-10-24
Fixes issue #2452, tooltips flicker when forced under the mouse

### LINKED ISSUE #2452: Tooltip message that ends up hitting the right side of the window starts "vibrating"
I am guessing it is trying to adjust itself in some fashion to fit on screen while also being centered on the mouse cursor, but should probably instead just chill out and sit still even if it can't be perfectly centered on the cursor anymore :-)

Noticed it with the Potions module with a potion (so a more substantial tooltip message) in the right-most toolbar slot.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37212069-tooltip-message-that-ends-up-hitting-the-right-side-of-the-window-starts-vibrating?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).
</bountysource-plugin>

segfault802:
### Contains

Fixes #2452 
### How to test

Load up the game with the potions module
Give yourself a double jump potion
Place the item in the lower-right inventory slot
Mouse over the item, the tooltip should not flicker or "vibrate" even when the mouse is on top of it

This scenario can occur when a tooltip is unable to render in the preferred
location due to the screen boundaries, for instance if an item with a really
long tooltip is placed in the lower-right most inventory slot.  Because UILists
define InteractionRegions around their contents by default and tooltips are
rendered last, they are seen as the new top-most interactive element on the
following game tick.  This causes an immediate reset of the tooltip timer due
to the top-most region changing, and the process repeats.  This has been fixed
by preventing tooltips from defining interaction regions through a new optional
flag in UIList that sets whether the list is interactive or not.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UIList.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UIList.java
@@ -44,6 +44,7 @@ public class UIList<T> extends CoreWidget {
     private final List<ItemInteractionListener> itemListeners = Lists.newArrayList();
     private final List<ItemActivateEventListener<T>> activateListeners = Lists.newArrayList();
     private final List<ItemSelectEventListener<T>> selectionListeners = Lists.newArrayList();
+    private Binding<Boolean> interactive = new DefaultBinding<>(true);
     private Binding<Boolean> selectable = new DefaultBinding<>(true);
     private Binding<T> selection = new DefaultBinding<>();
     private Binding<List<T>> list = new DefaultBinding<>(new ArrayList<>());
@@ -80,7 +81,9 @@ public class UIList<T> extends CoreWidget {
                 } else {
                     canvas.setMode(DEFAULT_MODE);
                 }
-                canvas.addInteractionRegion(listener, itemRenderer.getTooltip(item), itemRegion);
+                if (isInteractive()) {
+                    canvas.addInteractionRegion(listener, itemRenderer.getTooltip(item), itemRegion);
+                }
             } else {
                 canvas.setMode(DISABLED_MODE);
             }
@@ -134,9 +137,13 @@ public class UIList<T> extends CoreWidget {
         selectable = binding;
     }
 
+    public boolean isInteractive() { return interactive.get(); }
+
     public boolean isSelectable() {
         return selectable.get();
     }
+
+    public void setInteractive(boolean value) { interactive.set(value); }
 
     public void setSelectable(boolean value) {
         selectable.set(value);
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UIList.java
```diff
@@ -44,6 +44,7 @@
     private final List<ItemInteractionListener> itemListeners = Lists.newArrayList();
     private final List<ItemActivateEventListener<T>> activateListeners = Lists.newArrayList();
     private final List<ItemSelectEventListener<T>> selectionListeners = Lists.newArrayList();
+    private Binding<Boolean> interactive = new DefaultBinding<>(true);
```
segfault802: It is not clear why we use DefaultBindings here to wrap around simple primitives.  I did it here only to be consistent.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIList.java
```diff
@@ -80,7 +81,9 @@ public void onDraw(Canvas canvas) {
                 } else {
                     canvas.setMode(DEFAULT_MODE);
                 }
-                canvas.addInteractionRegion(listener, itemRenderer.getTooltip(item), itemRegion);
+                if (isInteractive()) {
```
segfault802: Could also use `isSelectable()` and do away with the separate field entirely.

Cervator:
@GooeyHub: add to whitelist

@segfault802: thanks for the fix and extra comments!

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UIList.java
```diff
@@ -134,10 +137,14 @@ public void bindSelectable(Binding<Boolean> binding) {
         selectable = binding;
     }
 
+    public boolean isInteractive() { return interactive.get(); }
```
Cervator: FYI - this should be on regular lines as per our code style :-)

Cervator:
Finally tested now that I'm back and merged :-)

Fixed a minor style fix I commented on here, too many open PRs and it is too trivial to leave open for just that. Also tweaked another tiny thing or two in the affected files (https://github.com/MovingBlocks/Terasology/commit/79091c6d48e6610aa4d3193ca7fc8325c71b080d) since they're being changed anyway, feel free to do that (and especially improve javadoc!) as a general practice

PR #2563, opened 2016-10-24
falling-off-ledges prevention (mc) feature

nihal111:
### Contains

Potential falling-off-ledges prevention feature.

An extension to PR #2544 and issue #1075.

As of now I have a basic check for preventing the character to fall down any step larger than 0.35f. But if I keep pushing the character off the ledge, at one point the state becomes non-grounded i.e `state.isGrounded() == false` and the game hangs for a frame and the character "appears to" fall off a height (I know because, the landing sound of character fall, is played) landing at the same position with a loss of health.

Could use some help in fixing that.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
+++ b/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
@@ -362,10 +362,8 @@ public class KinematicCharacterMover implements CharacterMover {
         if (!hitBottom && stepHeight > 0) {
             Vector3f tempPos = new Vector3f(position);
             hitBottom = moveDown(-stepHeight, slopeFactor, collider, tempPos);
-            logger.info("hitBottom" + hitBottom);
             // Don't apply step down if nothing to step onto
             if (hitBottom) {
-                logger.info(tempPos.toString());
                 position.set(tempPos);
             }
         }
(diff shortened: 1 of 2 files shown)


GooeyHub:

GooeyHub:

skaldarnar:
Thanks for your work @nihal111. As discussed on Slack, the whole movement-related code could use some good refactoring. I think we might be able to attack this with some joint work by you, @portokaliu (maybe @xrtariq2594 and others). 

I'd be happy to merge a minimal working version of this fix before. It seems like you moved some code and added comments here and there. Strip all that down to a bare minimum for this PR and add the additional work and documentation to the refactoring work.

engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
```diff
@@ -94,6 +94,40 @@ public KinematicCharacterMover(WorldProvider wp, PhysicsEngine physicsEngine) {
         physics = physicsEngine;
     }
 
+    /**
```
skaldarnar: moving this method does not have substantial benefit at this point (could apply it later, though)

engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
```diff
@@ -355,8 +355,10 @@ private MoveResult move(final Vector3f startPosition, final Vector3f moveDelta,
         if (moveDelta.y < 0 || steppedUpDist > 0) {
             float dist = (moveDelta.y < 0) ? moveDelta.y : 0;
             dist -= steppedUpDist;
+            // Returns true if bottom is hit at current position.
```
skaldarnar: this comment should better be a javadoc for `moveDown()` imho

engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
```diff
@@ -623,105 +625,126 @@ private void walk(final CharacterMovementComponent movementComp, final Character
...
-                    logger.debug("Landed at " + landVelocity);
-                    entity.send(new VerticalCollisionEvent(state.getPosition(), landVelocity));
+        if (movementComp.mode == MovementMode.CROUCHING) {
+            if (moveResult.isBottomHit()) {
+                if (state.isGrounded()) {
+                    Vector3f pos = moveResult.getFinalPosition();
+                    Vector3f to = new Vector3f(pos.x, pos.y - 0.35f, pos.z);
+                    SweepCallback callback = collider.sweep(pos, to, VERTICAL_PENETRATION_LEEWAY, -1f);
+                    if (callback.hasHit()) {
+                        state.getPosition().set(moveResult.getFinalPosition());
```
skaldarnar: instead of setting the state here, the `moveResult` (the movement to perform) should be adjusted to prevent falling off the ledge. this probably needs some better structuring of the whole walk method to get a better overview of things.

oniatus:
Is this still in progress/an issue? I did a short test with a 3-block tower on a flat surface with ladders on one side and got no falling sound when hitting the ground or walking on the top block.

there were some other oddities though:
- jumping on a ladder moved the player upwards, i expected him to fall down/backwards or not jump at all
- When climbing on top of the block, the player moves a bit higher than needed, then falls a bit (no sound though)
- compared to minecraft, the controls on the ladder are dependent on the view direction. In Minecraft, you can view downwards and press w until you reach the top. In Terasology, you have to view above the horizon which makes it hard to spot the point where your character moves above the edge.
If it's out of scope of the PR i can move this to a separate feature request issue :)

GooeyHub:

portokaliu:
@oniatus  I think this one is stuck alongside movement refactoring (which would change all of what you mentioned :) )

oniatus:
I think after this period of inactivity we can close this one for now. Feel free to reuse some of the ideas or reopen it as needed.
We also can continue any discussion about the feature on #1075 .

Cervator:
Also linking to #2573

PR #2564, opened 2016-10-24
More work on render nodes

emanuele3d:
### Contains

Refactoring of a number of nodes, primarily to migrate them to the new rendering architecture.
### How to test

Startup a game and check that landscape, water surfaces, shadows, objects and plants render as usual in all circumstances. 

See individual comments for further info on specific bits of code.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
+++ b/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
@@ -26,8 +26,6 @@ import org.terasology.monitoring.PerformanceMonitor;
 import org.terasology.rendering.assets.material.Material;
 import org.terasology.rendering.cameras.Camera;
 import org.terasology.rendering.logic.LightComponent;
-import org.terasology.rendering.primitives.ChunkMesh;
-import org.terasology.rendering.world.WorldRendererImpl;
 import org.terasology.rendering.world.viewDistance.ViewDistance;
 import org.terasology.rendering.world.WorldRenderer;
 import org.terasology.world.WorldProvider;
@@ -39,7 +37,6 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
-import java.util.PriorityQueue;
 
 public class HeadlessWorldRenderer implements WorldRenderer {
 
@@ -86,11 +83,6 @@ public class HeadlessWorldRenderer implements WorldRenderer {
     }
 
     @Override
-    public void renderChunks(PriorityQueue<RenderableChunk> chunks, ChunkMesh.RenderPhase phase, Camera camera, WorldRendererImpl.ChunkRenderMode mode) {
-
-    }
-
-    @Override
     public void onChunkLoaded(Vector3i pos) {
 
     }
@@ -127,9 +119,18 @@ public class HeadlessWorldRenderer implements WorldRenderer {
     }
 
     @Override
+    public void increaseTrianglesCount(int increase) {
+        // we are not going to count triangles in headless
+    }
+
+    @Override
+    public void increaseNotReadyChunkCount(int increase) {
+        // we are not going to count not ready chunks in headless
+    }
+
+    @Override
     public void render(RenderingStage mono) {
         // TODO Auto-generated method stub
-
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
@@ -36,10 +36,17 @@ import org.terasology.world.chunks.RenderableChunk;
 import static org.terasology.rendering.opengl.DefaultDynamicFBOs.READ_ONLY_GBUFFER;
 
 /**
- * TODO: Diagram of this node
- * Alpha reject is used for semi-transparent billboards, which in turn are used for ground plants.
+ * This node uses alpha-rejection to render semi-transparent blocks (i.e. tree foliage) and
+ * semi-transparent billboards (i.e. plants on the ground).
+ *
+ * Alpha-rejection is the idea that if a fragment has an alpha value lower than some threshold
+ * it gets discarded, leaving the color already stored in the frame buffer untouched.
+ *
+ * This is a less expensive way to render semi-transparent objects compared to alpha-blending.
+ * In alpha-blending the color of a semi-transparent fragment is combined with
+ * the color stored in the frame buffer and the resulting color overwrites the previously stored one.
  */
-public class ChunksAlphaRejectNode extends WireframeCapableNode {
+public class AlphaRejectBlocksNode extends WireframeCapableNode {
 
     private static final ResourceUrn CHUNK_SHADER = new ResourceUrn("engine:prog.chunk");
 
@@ -55,6 +62,9 @@ public class ChunksAlphaRejectNode extends WireframeCapableNode {
     private Camera playerCamera;
     private Material chunkShader;
 
+    /**
+     * Initialises the node. -Must- be called once after instantiation.
+     */
     @Override
     public void initialise() {
         super.initialise();
@@ -64,6 +74,17 @@ public class ChunksAlphaRejectNode extends WireframeCapableNode {
         chunkShader = getMaterial(CHUNK_SHADER);
     }
 
+    /**
+     * Renders the world's semi-transparent blocks, i.e. tree foliage and terrain plants.
+     * Does not render fully opaque blocks, i.e. the typical landscape blocks.
+     *
+     * Takes advantage of the two methods
+     *
+     * - WorldRenderer.increaseTrianglesCount(int)
+     * - WorldRenderer.increaseNotReadyChunkCount(int)
+     *
+     * to publish some statistics over its own activity.
+     */
     @Override
     public void process() {
         PerformanceMonitor.startActivity("rendering/chunksAlphaReject");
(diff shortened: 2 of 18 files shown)


engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
```diff
@@ -127,9 +119,18 @@ public void update(float delta) {
...
 
     @Override
+    public void increaseTrianglesCount(int increase) {
+        // we are not going to count triangles in headless
+    }
+
+    @Override
+    public void increaseNotReadyChunkCount(int increase) {
+        // we are not going to count not ready chunks in headless
+    }
```
emanuele3d: These two new methods above are a bit of a hack to maintain existing functionality. They are not horrible and they could probably stay for a while. But I'm thinking we might want to have some kind of StatisticsManager registered in the main context holding all sorts of statistics like these. That would be a more appropriate place I feel. Or we could keep those statistics here and just make something that wraps them a bit better than what we have now.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
```diff
@@ -0,0 +1,142 @@
...
+ * semi-transparent billboards (i.e. plants on the ground).
+ *
+ * Alpha-rejection is the idea that if a fragment has an alpha value lower than some threshold
+ * it gets discarded, leaving the color already stored in the frame buffer untouched.
+ *
+ * This is a less expensive way to render semi-transparent objects compared to alpha-blending.
+ * In alpha-blending the color of a semi-transparent fragment is combined with
+ * the color stored in the frame buffer and the resulting color overwrites the previously stored one.
+ */
+public class AlphaRejectBlocksNode extends WireframeCapableNode {
```
emanuele3d: This class has been renamed from ChunksAlphaRejectNode. Technically the class does iterate through chunks and retrieves from them the mesh to be rendered. However, the mesh only contains the geometry for some of the blocks, the blocks that are semi-transparent. I also feel is more intuitive to think that this node renders blocks rather than chunks, as chunks per se do not have as much meaning from a visual perspective.

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
```diff
@@ -0,0 +1,142 @@
...
+     *
+     * Takes advantage of the two methods
+     *
+     * - WorldRenderer.increaseTrianglesCount(int)
+     * - WorldRenderer.increaseNotReadyChunkCount(int)
+     *
+     * to publish some statistics over its own activity.
+     */
+    @Override
+    public void process() {
```
emanuele3d: This method in this and other classes now include what was previously WorldRendererImpl.renderChunks().

By moving the code in the separate nodes it can be greatly simplified I feel, making it easier to follow. It also makes it easier to eventually move some lines of it toward the new architecture, taking advantage of State Changes.

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
```diff
@@ -0,0 +1,142 @@
...
+                // Effectively this just positions the chunk appropriately, relative to the camera.
+                // chunkPositionRelativeToCamera = chunkCoordinates * chunkDimensions - cameraCoordinate
+                GL11.glPushMatrix();
+                GL11.glTranslatef(chunkPositionRelativeToCamera.x, chunkPositionRelativeToCamera.y, chunkPositionRelativeToCamera.z);
+
+                chunk.getMesh().render(ChunkMesh.RenderPhase.ALPHA_REJECT); // TODO: remove face culling from this method
+                numberOfRenderedTriangles += chunk.getMesh().triangleCount();
+
+                GL11.glPopMatrix(); // Resets the matrix stack after the rendering of a chunk.
+
```
emanuele3d: The content of the IF block above should theoretically be wrapped by a couple of methods, one sertting the chunkShader parameters and the other translating the chunk and rendering it. This would allow for the same code to be reused in other nodes that have identical code. 

The problem is, where to put those two methods?

Some of the affected nodes inherit from WireframeCapableNode, but another inherit from ConditionDependentNode and one more from AbstractNode.  It looks like a case for multiple inheritance but by my understanding it means there is an architectural issue to be resolved.

Suggestions?

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
```diff
@@ -0,0 +1,142 @@
...
+            } else {
+                numberOfChunksThatAreNotReadyYet++; // TODO: verify - should we count them only in ChunksOpaqueNode?
+            }
+        }
+
+        chunkShader.deactivateFeature(ShaderProgramFeature.FEATURE_ALPHA_REJECT);
+
+        worldRenderer.increaseTrianglesCount(numberOfRenderedTriangles);
+        worldRenderer.increaseNotReadyChunkCount(numberOfChunksThatAreNotReadyYet);
```
emanuele3d: Notice this is where the statistics are sent to the WorldRenderer. Other nodes do it the same way.

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueObjectsNode.java
```diff
@@ -16,44 +16,58 @@
...
- * TODO: Diagram of this node
- * Alpha reject is used for semi-transparent billboards, which in turn are used for ground plants.
+ * This node renders the opaque (as opposed to semi-transparent)
+ * objects present in the world. This node -does not- render the landscape.
+ *
+ * Objects to be rendered must be registered as implementing the interface RenderSystem and
+ * take advantage of the RenderSystem.renderOpaque() method, which is called in process().
  */
-public class ChunksAlphaRejectNode extends WireframeCapableNode {
+public class OpaqueObjectsNode extends WireframeCapableNode {
```
emanuele3d: Github is highlighting this as a rename from ChunksAlphaRejectNode to OpaqueObjectsNode. That's false.

engine/src/main/java/org/terasology/rendering/dag/nodes/OverlaysNode.java
```diff
@@ -19,25 +19,50 @@
...
     }
 
+    /**
+     * Iterates over any registered RenderSystem instance and calls its renderOverlay() method.
+     */
     @Override
     public void process() {
         PerformanceMonitor.startActivity("rendering/overlays");
 
+        playerCamera.lookThrough(); // TODO: remove. Placed here to make the dependency explicit.
```
emanuele3d: All nodes are dependent on a camera from which something is rendered. Previously these nodes relied on an identical line placed upstream in a previous node. If a second camera was set between that node and this node, this node would render through the second camera instead. 

By placing this line in this and other nodes I make the dependency on this particular camera explicit and these nodes can move in the render graph slightly more freely. 

Eventually this line (and the identical ones in other nodes) will be replaced by a StateChange and/or a setCamera() method. In turn these ways to set the camera for the node will eventually take in input an entity with an attached camera component rather than an instance of a camera class. But that's still some way out.

engine/src/main/java/org/terasology/rendering/dag/nodes/ShadowMapNode.java
```diff
@@ -75,6 +78,9 @@
     private BackdropProvider backdropProvider;
 
     @In
+    private ShaderManager shaderManager;
```
emanuele3d: This isn't in use anymore! I'll remove it in the next commit.

indianajohn:
I built/ran this and everything seemed to work fine on my machine.

I like the change in file naming scheme to rendering "Blocks" rather than 'Chunks." Somebody who is beginning to read the code has one fewer class to look up before he/she understands what those nodes do. Also, last time I read `WorldRendererImpl` I actually thought that the renderChunks() code belonged in a separate node, I guess I wasn't the only one.

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
emanuele3d: Thank you for the recommendation @indianajohn. Intriguing approach moving those methods to the Mesh interface. I guess one concern I have with this approach is that it makes Mesh objects less of data containers and more of active players in the rendering. On a gut level I'd feel more comfortable with Meshes being pure data and the renderer processing that data. But I haven't looked at the Mesh interface and perhaps it is already -not- like that.

Regarding the individual points:
1. I agree. Eventually I really hope for the camera class to become very lightweight, only storing data and implementing accessors. 
2. Over the two years I have been on the project there have been a number of discussions regarding taking advantage of other libraries (i.e. LWJGL 3.x) or higher OpenGL releases. My hope is that by making the renderer more understandable and more structured it will be easier for people to develop not only additional effects but also more modern renderers. So, I'm with you: we are only limited by manpower.
3. Sounds good to me.
4. I confirm that's the case (dag focusing on state changes), at least so far. With @tdgunes we discussed the possibility for the DAG to be the basis also for runtime generation of shaders but it still seems far out.

Again, thank you for your recommendations. Keep tuned for more commits.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
```diff
@@ -98,35 +99,16 @@ public void process() {
...
-                // Effectively this just positions the chunk appropriately, relative to the camera.
-                // chunkPositionRelativeToCamera = chunkCoordinates * chunkDimensions - cameraCoordinate
-                GL11.glPushMatrix();
-                GL11.glTranslatef(chunkPositionRelativeToCamera.x, chunkPositionRelativeToCamera.y, chunkPositionRelativeToCamera.z);
+
+                chunkMesh.updateMaterial(chunkShader, chunkPosition, chunk.isAnimated());
+                numberOfRenderedTriangles += chunkMesh.render(OPAQUE, chunkPosition, cameraPosition);
 
                 if (renderingDebugConfig.isRenderChunkBoundingBoxes()) {
-                    AABBRenderer aabbRenderer = new AABBRenderer(chunk.getAABB());
```
emanuele3d: The previous commit introduced a change in the bounding box rendering functionality, confirmed again in the new commit: bounding boxes are rendered only for opaque chunks. They are not rendered in reflections, shadows and for blocks that are not opaque. This is debatable and if you really think I should I could re-introduce them in some of nodes modified by this PR.

PR #2567, opened 2016-10-24
added ignoreCase in givepermission command, more work maybe required

kartikey0303:
Added the IgnoreCase condition in player equal tag. I tested it on my computer, locally. More work maybe required.


GooeyHub:

PR #2574, opened 2016-10-26
Added popup if a key has already been assigned to another action.

zeokav:
### Contains

Added a fix for issue #2552 where if a key has already been mapped to another action, a popup is raised which asks the user to confirm the change. It opens a popup instead of flashing a red error message because I think that will look slightly out of style. 

### How to test

To test - 

- Open the Input section in the Settings screen.

- Click on button to map.

- Choose a key that has already been assigned for another task. (Say S when you're rebinding W)

- Select OK to continue with the rebinding or cancel to go back and reconsider.

PR #2579, opened 2016-10-27
i18n fixes for confirmChangePopup

rzats:
### Contains

Minor fixes to code added in #2574 to make UI strings properly internationalizable and a note in the wiki about translating your stuff. Also fixes an unrelated typo I happened to encounter in player settings.

### How to test

Check that the strings added in #2574 (confirmChangePopup) are properly translated.

Nothing! Pinging @skaldarnar and @zeokav for review.


GooeyHub:

engine/src/main/resources/assets/i18n/menu.lang
```diff
@@ -50,6 +50,8 @@
     "category-mouse": "category-mouse",
     "category-movement": "category-movement",
     "category-nui": "category-nui",
+    "change-popup-message": "change-popup-message",
```
skaldarnar: I'd go for a more descriptive key `change-keybind-popup-message` as "change" is quite general and could conflict with other change actions in the menu later. 
Another option would be move it to another .lang file specific for keybindings....

rzats: Noted!

I'd rather not overcomplicate things re: Weblate setup unless we're going for a fully reworked system with strings split into 5-6 files.

GooeyHub:

PR #2580, opened 2016-10-27
Improved detection for penetrable blocks. Also added range check for …

portokaliu:
Contains:
    Second attempt at repairing the tunneling issue #2139 that items might have when traveling at high speed. It's an addition to #2533

This particular fix contains the addition of checking for penetrable blocks and ignoring them whilst checking further blocks for possible collisions ( in case the item has a speed of more than 1 Block / dT )

The collision event now returns a lot more relevant information that can be used by various systems in case the collision is needed (like the StickySystem) and adds a default reaction to the direction of impact to impose a relevant movement and position.

The range at which items are dropped is now considered. Dropping items while looking at a block will now be possible without the item being released inside said block.

How to test:
    Try throwing items of different shapes at different speeds towards surfaces and see a weird reaction they can have, but almost no tunneling.

To consider (and some needed) before merging:
- [x] Verify why the reaction vector has a low value
- [ ] See why items sometimes still do pass through blocks
- [ ] Change dropping items so that they also consider penetrable blocks
- [ ] See what can be done about releasing items close to blocks ( creating and imposing movement on an item during event processing will make it first move to the next position (because physics follow event processing) and only then checking if future blocks will have an effect on it (in the next event processing loop)

### DIFF
--- a/engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
+++ b/engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
@@ -209,6 +209,9 @@ public class BulletPhysics implements PhysicsEngine {
     @Override
     public HitResult rayTrace(org.terasology.math.geom.Vector3f from1, org.terasology.math.geom.Vector3f direction, float distance, Set<EntityRef> excludedEntities,
             CollisionGroup... collisionGroups) {
+        if (excludedEntities == null) {
+            return rayTrace(from1, direction,distance,collisionGroups);
+        }
         Vector3f to = new Vector3f(VecMath.to(direction));
         Vector3f from = VecMath.to(from1);
         to.scale(distance);
--- a/engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
+++ b/engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
@@ -36,6 +36,7 @@ import org.terasology.math.geom.Vector3f;
 import org.terasology.monitoring.PerformanceMonitor;
 import org.terasology.network.NetworkComponent;
 import org.terasology.network.NetworkSystem;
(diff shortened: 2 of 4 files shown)


GooeyHub:

Cervator:
Trying to at least merge one serious PR a day to work down the backlog, even if not quite perfect yet - yours is today's lucky pick! Sorry if it isn't totally ready yet, but I gotta move some product, man!

Tests out fine in any case, the bouncing around is a little more realistic now (no constantly bouncing torches, although a few might still do it just a bit of stand up perfectly occasionally when dropped). Love that items no longer drop inside a block if you're up against it!

Could still provoke the tunneling as expected, but yeah it is harder (have to be very hard throw at a "funneled" corner so to say). And neat new details!

Worked fine in headless + client as well. Physics items were a bit warpy as the server corrected positions, but pretty minor.

portokaliu:
Well, I do have a few more stuff in a shelf I should've commit already, but alas, it will be for next PR. (was working last night with the scope of getting all collisions from the physics to emit sounds on any sort of rigidbody overlapping over a certain error value. Still have a bit of it to work on)

As for the PR itself, it's better that we have it on the main branch since this opens up the possibility for the other modules ( like stickable ) to be completed with the current implementation. The problem would be that stickable sort of changes the event to being consumable (because it sort of should be), so I'll have to work on the branch after I submit the full stickable implementation (I think I haven't done the PR yet for that. trying to get that and ladders done today)

Looking into this version, I remember one issue I had. The entities added to the exclusion group were never considered for some reason. When checking their key values, they were never validated, thus never ignored.

On a side-note, do you think that entities could travel more than a block's length in a single frame? The algorithm was made for checking more blocks out in case there were penetrables and/or the item had too much speed.

Issue #2581, opened 2016-10-28
Switching hotbar item while holding attack button causes an extra attack


Josharias:
### What you were trying to do

I was trying to switch to a better tool to break dirt while joined to a headless server.
### What actually happened

The game instantly attacked extra each time I switched to a different hotbar item,  not regarding the cooldown time.
### How to reproduce
1. Hold the attack button while breaking a block
2. Mash the 1-9 keys to switch between hotbar items
3. Much excitement for increased digging speed!
### Log details and game version

Current develop branch (e08708b48b24ef11bbdd4a52d007ceaf997f6249) in a headless-remote client scenario. 
##

Cervator:
Closing as fixed by #2582

PR #2582, opened 2016-10-28
Fixes for item use cooldowns

Josharias:
### Contains
- A fix to let items use just a single cooldowntime property on the ItemComponent.
- A improvement to the cooldown system that enforces validation on both the client and server side to see if an action is attempted while still cooling down.

As an implication of getting setting this data also on the server side,  this data will also be sent to other remote clients so that they can show other characters doing an action.
### How to test
- Use an item
- Attack with a weapon
- Attack with bare hand
- Press and hold attack to ensure no machine gun-ing happens

- [ ] There is a problem with single player which causes the authority to skip sending the activation event.


GooeyHub:

Josharias:
There is a problem with single player which causes the authority to skip sending the activation event.  Need to address this before proceeding.

GooeyHub:

Cervator:
Interesting!

I can confirm this appears to fix #2581 :-)

However, in an interesting coincidence it seems to allow for an error I saw get reported just yesterday in an unrelated fashion for a different reason. Or it just made it worse. Or it just *felt* worse because humans don't get along with probabilities. See #2597 for details. I was able to replicate with an older Omega build so merging!

Nice catch and fix :-)

Issue #2565, opened 2016-10-24
Crouch mode change bugs


nihal111:
### How to reproduce

*crouch
*Go for a swim while crouching
*Come out and find your height is reduced to half
Hint: Look into `updateMode` method inside KinematicCharacterMover. Probably changing mode to walking before changing to swimming would work.

### How to reproduce

*crouch
*Let the game save
*Exit
*Load the save game. Find that your height is altered.

Josharias:
Also,  I am not sure if it is just me...  when I am joined to a headless server,  I hit [ctrl] my character starts to vibrate up and down.

DPirate:
If you keep repeating the process, after a certain repeats you will start falling down through the blocks(upon crouching) and the you can see the upside down(not all of it though, just certain blocks).

![new bitmap image](https://cloud.githubusercontent.com/assets/24975118/24044058/85f5eec6-0b3f-11e7-8305-4cf1feb082d3.png)

PR #2586, opened 2016-11-02
Attempting fix for #2565

nihal111:
### Contains

Potential fix for #2565 

### How to test

Look at issue for details on how to reproduce.

This doesn't really work out well, and is probably not a good way to solve this problem. Could use some help on how to fix.

Another thing that can be done is that CROUCHING can be remembered, so that after swimming, the character returns in crouched position (However this will change player's height in water which is undesirable). But this will also solve the second part of #2565

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
+++ b/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
@@ -18,6 +18,7 @@ package org.terasology.logic.characters;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.input.binds.movement.CrouchModeButton;
 import org.terasology.logic.characters.events.FootstepEvent;
 import org.terasology.logic.characters.events.HorizontalCollisionEvent;
 import org.terasology.logic.characters.events.JumpEvent;
@@ -195,7 +196,7 @@ public class KinematicCharacterMover implements CharacterMover {
             }
         }
 
-        updateMode(state, newSwimming, newDiving, newClimbing);
+        updateMode(entity, state, newSwimming, newDiving, newClimbing);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Can confirm the original bug (at least the crouch-swimming, loading a save seemed OK to me) and that this fixes it, but yeah it doesn't seem like this would be the way to do it :-)

As written this will conflict with #2515 (pinging @coty91!) which overhauls swimming and may be a better architecture to base this fix on top of. It makes me think about movement modes in general and how switching to swimming should probably just reset crouching, yet doing by triggering a button event seems wrong somehow.

Also, interestingly, probably swimming *should affect the player's shape anyway* - at least for a human a regular swimming stance (while actively moving) is more like a prone stance than standing or crouching, while treading water would be more like standing. Yet a swimming prone would leave the player's eye height the same as treading water, while a landed prone would indeed lower the eye height dramatically. Ahh challenges! :-)

Gets even more fun if you consider other creatures. A crocodile would be prone by nature and that stays unchanged on entering water, whether just floating or actively swimming. And Chessandgo is working on a centaur ...

So maybe crouching and prone should be a subtly different system anyway, yet still reset/change when entering water. Lots of possibilities, but none that are super quick to approach.

In any case I think we need to bake in more flexibility in changing the player's shape during movement, buttons being clicked or not.

nihal111:
@Cervator Yup, this even breaks a test as you can see.
And I can confirm that if you crouch and save and load the save game your standing height turns to half of what it was earlier.
Sending a ButtonEvent was the quickest way around this, since otherwise I would need to @Share LocalPlayerSystem, but is surely not a proper way. It doesn't work smoothly either, I guess there is some lag between the event being sent and height being changed (i.e if you aim for the bottom of the water block in a 1-block deep pool, you might get stuck and would have to jump out). A probably better way would be to change the height before the movement is done, once it is realized that the next step takes you in water.

nihal111:
Let's close this PR for now. I'd get back to this soon :)

Issue #2553, opened 2016-10-20
Feature request: rope ladders


Cervator:
Another fairly easy one here inspired by Hacktoberfest and a recent IRC mention - could go in a module but since ladders are in Core I think we should add this there as well, until a later full piecing out of Core.

The difference between a regular ladder and a rope ladder should be the ability to place a ladder segment _below_ the actively targeted existing rope ladder block, if the spot below is open. Then the player should be able to climb down and continue the ladder, while being able to edge out to either side enough to place additional solid blocks adjacent to the supporting column the ladder is placed against.

This should help the player travel downwards in cases like floating islands, deep vertical caves, and similar situations.

I don't think rope ladders should be able to go up, unsupported, that's what building a column of solid blocks with regular ladder blocks up against should be for.

Probably it should be very limited what can be attached to rope ladders - they should really only be for traveling downward to be able to then start a more proper support platform via other means.

It may be that the logic needs to be in a special ladder system that either catches block placement attempts of the specific block then redirects the placement target to spot below or make the rope ladder block itself special to where `e` clicking one attempts to fetch another rope ladder from the player's inventory and placing it below the target block.

Current ladders have no special placement logic and just movement against them is handled in `KinematicCharacterMover` (probably should remain unchanged and just also count for rope ladders)

marti1125:
Hi @Cervator I would like to work on this. could you give some instructions? please

oniatus:
Hi @marti1125 . 
Climbing is handled in the `KinematicCharacterMover.findClimbable`. Likely, that a block with the climable property (see `Core/assets/blocks/Ladder.block` for an example) would work.
`BlockItemSystem` handles the placement of blocks and notifies the inventory item if the event is not consumed (`OnBlockItemPlaced`-event and `BlockInventorySystem`).
`BlockPlacingSystem` does the actual placement, but I guess you want to build around this system and do a custom placing based on the blocks in the world (instead of attaching a ladder on a ladder, the ladder should be placed below the ladder). This could work with a higher event priority and consume the event for ladders (this also needs a check for the consumed event in `BlockPlacingSystem`). Later on, we could improve this custom ladder placing system and forbid any other blocks that are about to be attached on a ladder.
Consuming the event while placing a block would break the `OnBlockItemPlaced`-sending in the `BlockItemSystem`. One way this could be handled is a separate boolean/int/whatever in the event, to indicate that/how many blocks have been placed and check for this flag instad.
[...]

PR #2587, opened 2016-11-03
Added ladder "roping" functionality.

portokaliu:
Contains:
Functionality of "roping" ladders. Now when trying to place a ladder while looking at a ladder (of the same block family), you will be placing the new block at the bottom of however many ladders down the current one goes. No more getting stuck on flying islands for you :D

This would be a good start towards fixing #2553, if not overdoing some parts of it.

There's currently no special condition for what type of ladder is used, but any property or component can be checked so I'm open to ideas on that side. 

The only "downside", if it can be called that, is that now you cannot place ladders while looking at other same type ladders if there's no space below the one you're looking at (as opposed to putting a ladder block relative to whatever block you're pointing at.

How to test:

Get yourself some ladders, get stuck in a castle with said ladders and proceed to being ladder Rapunzel.

A tl;dr for what changes could be considered:
- [ ] Ladders only try to place below the current one, not at the end of the stack.
- [ ] Only ladders that have a certain component or property should be handled this way
- [ ] Consider moving this to a possible module.
- [x] Ask for what I'm missing

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
+++ b/engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
@@ -88,9 +88,25 @@ public class BlockItemSystem extends BaseComponentSystem {
         }
         Vector3i targetBlock = blockComponent.getPosition();
         Vector3i placementPos = new Vector3i(targetBlock);
+        Vector3i climbablePlacementPos = new Vector3i(targetBlock);
+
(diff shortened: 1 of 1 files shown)


GooeyHub:

GooeyHub:

Cervator:
Closing as per @portokaliu - being put into a module instead :-)

PR #2595, opened 2016-11-06
New handling of Wireframe debug mode and FirstPersonViewNode refactor

emanuele3d:
### Contains
- refactoring of the way Wireframe is handled _internally_: should work the same. This is the bulk of the PR but only because it affects a number of nodes: there is a fair amount of repeated code.
- some refactoring of the FirstPersonViewNode - currently debating if to remove the node completely as it doesn't seem to be in use right now.

### How to test
After loading a game, press F3 and F9 to show wireframe. It should work as usual.


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
```diff
@@ -45,34 +50,57 @@
...
      * Initialises the node. -Must- be called once after instantiation.
      */
     @Override
     public void initialise() {
-        super.initialise();
         playerCamera = worldRenderer.getActiveCamera();
+
+        wireframeStateChange = new SetWireframe(true);
+        renderingDebugConfig = config.getRendering().getDebug();
+        new WireframeTrigger(renderingDebugConfig, this);
```
emanuele3d: I find this line a bit ugly. The containing node doesn't need a reference to the WireframeTrigger instance but the opposite is true, hence the line as it is. Is there a better way to do this?

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
```diff
@@ -45,34 +50,57 @@
...
+            refreshTaskList();
+        }
+    }
+
+    public void disableWireframe() {
+        if (getDesiredStateChanges().contains(wireframeStateChange)) {
+            removeDesiredStateChange(wireframeStateChange);
+            refreshTaskList();
+        }
+    }
```
emanuele3d: These two methods are now repeated identically in a number of nodes implementing the WireframeCapable interface. Without multiple inheritance I can't think of a way to do it differently as these nodes would need to inherit from the capability of multiple classes. The alternative, composition, would expose too much of a node's innards, I feel. 

Suggestions? Or can it stays this way?

engine/src/main/java/org/terasology/rendering/dag/nodes/FirstPersonViewNode.java
```diff
@@ -69,13 +103,9 @@ public void process() {
             playerCamera.updateMatrices();
             playerCamera.loadProjectionMatrix();
 
-            /**
-             * Resets the state after the render of the First Person View.
-             */
-            GL11.glDepthFunc(GL_LEQUAL);
+            //GL11.glDepthFunc(GL_LEQUAL);
             GL11.glPopMatrix();
```
emanuele3d: The GL11.* and the playerCamera.* statements are what I was talking about in the Forum @Josharias: perhaps we could remove them. But right now I'd be more tempted to remove the whole node frankly.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
emanuele3d: Besides, will there be any problem with Garbage Collection?

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
Cervator: I will take one StrawberryEatingCapableNode please. For ... personal reasons. 

:strawberry: :strawberry: :strawberry: 

At first I was thinking about suggesting default methods in the interfaces but needing to include `wireframeStateChange` as a variable would get in the way. Unless ... you used 

`removeDesiredStateChange(getWireframeStateChange());`

But then you'd still need to implement the getter anyway ... nevermind :(

emanuele3d:
Hmmm... @Cervator: I'm thinking it's probably ok to merge it as it is given no better alternatives seem to be available at this time.

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
indianajohn: This made me curious about how GC works, and some research led to the rule of thumb for "safe" objects: an object is safe if that object is reachable by some thread. Since there is a code path for renderingDebugConfig (on which the object is actually being stored) to call `thatObject.propertyChange()`, it should be safe.

Anyway, to make it a little more clear what's happening you could change it to:


```java
renderingDebugConfig.subscribe(RenderingDebugConfig.WIREFRAME,new WireframeTrigger(renderingDebugConfig,this));
```
and take out the `subscribe()` call from the constructor, but that is arguably uglier.

Cervator: Bump for visibility - @emanuele3d I'll merge this now but this might be something to keep in mind :-)

IMHO it could work as a one-liner, even with a couple commas added (space after comma Checkstyle fix) the line would only be 122 deep starting 8 spaces in. I was slightly more confused by the 3-liner (but may have fooled myself thinking `this` appeared twice thanks to the `true`)

Cervator:
Tested it out, seems fine. Merged! :-)

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
emanuele3d: Ok guys, I'll keep all this in mind for the next PRs.

And thank you @indianajohn for the research on the GC and the one-liner suggestion.

PR #2599, opened 2016-11-08
Physics ImpactEvent improvements

portokaliu:
Contains:
The ImpactEvent was made consumable for the purpose of being ignored by other systems (like physics) so that no calculations are done for an entity that wouldn't exist anymore (which is the case for the 'Stickable' module).

Hackish reaction speed multiplier has been removed in favour of a working reaction to hitting blocks as well as a modifier that can be used to change the reaction speed (so the item doesn't constantly jump at the same height it fell from)

Improved calculations and eliminated overrides when trying to send and process an Impact event.

There will be another physics update following this one for the purpose of fixing (literal) edge cases (items thrown into edges have a likelihood of getting into the blocks.

This is once again an addition to #2139 and extends #2580.

How to test:
Try throwing items at the broadside of blocks. The reaction speed will now be proportionate to the speed it impacted the block with. Ever so slight performance boost as well.

Outstanding before merging:
The scope of this particular PR is complete. Changes to fix the final problems will be submitted in another PR.

- [ ] Consider adding a custom reaction speed per item depending on the mass? (maybe material as well)


GooeyHub:

portokaliu:
Agh, wild-thingy... Forgot to deactivate it on this pc. MiniFix inc

GooeyHub:

GooeyHub:

Cervator:
Merging! The bounce is soooo satisfying now, throwing torches at high speed against the side of a cliff, tree, or something. Even better with cheese wheels from FunnyBlocks! :-)

Issues noted for later consideration:

* Torches when dropped very lightly without moving may still just sort of end up sitting in the ground, partially submerged into the block it is sitting on, pointing up. No such problem when it ls laying on its side, although as per the journeys of @emanuele3d and @Josharias through our lighting the torches may still count as being somewhat sunk into blocks when on their sides, dimming the lighting
* When thrown just right the torches may bounce off in unexpected directions even against flat surfaces - probably that literal edge case of them hitting a block seam just right, thinking they're bouncing off a wall instead of the flat ground
* You can throw torches through dropped brick blocks very easily, sort of feels like they only collide with a solid center of the brick blocks but can pass by fairly deep inside the edges. Brick blocks collide pretty well with themselves (can stack them without having them sink into the block below)
* Fixed assorted minor Checkstyle warnings, more than just wildcard imports it can be hugely helpful to do a quick per-file scan in IntelliJ :D I'm doing that myself now when looking at PRs since I'm in there anyway and it gets these PRs going faster than waiting for style-only follow-ups. 

Also: HAH! Bouncing torches off monkeyheads is fun! Great use for that small pickable pause on freshly thrown objects :D New video material.

Cervator:
Style fixes committed separately via https://github.com/MovingBlocks/Terasology/commit/5e5819e5900910c278c760ea38a0c750ccd0907a - take a look at the examples :-)

portokaliu:
@Cervator The receive event should indeed not have anything in the `( )` since the filter itself is the fact that items send the event. This already means that only stuff thrown from inventory or dropped from blocks do this (pretty much any rigidbody entity in the world actually, but that's all I know that has a rigidbody)

Indeed torches do seem to only collide with their center. They can roll forwards (not on their side) very easily. I have yet to explore their shape definitions, but I did work around with fences and those look ok when dropped, so... I'll be looking into it to say the least.

The edge cases again, are something I specified still happened. This is still a mystery to me as to why, but I want to take a break from exploring the reasoning behind that so that I could do some fun stuff in the meantime. This doesn't mean I'm done with it though :)

PR #2602, opened 2016-11-11
Ping Command

ron-popov:
Added ping command , only works on windows

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -55,7 +55,6 @@ import org.terasology.network.ClientComponent;
 import org.terasology.network.JoinStatus;
 import org.terasology.network.NetworkMode;
 import org.terasology.network.NetworkSystem;
-import org.terasology.network.Server;
 import org.terasology.persistence.WorldDumper;
 import org.terasology.persistence.serializers.PrefabSerializer;
 import org.terasology.registry.In;
@@ -77,8 +76,6 @@ import org.terasology.world.block.family.BlockFamily;
 import org.terasology.world.block.items.BlockItemFactory;
 import org.terasology.world.block.loader.BlockFamilyDefinition;
 
-import java.io.BufferedReader;
-import java.io.InputStreamReader;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


GooeyHub:

portokaliu:
For this to make sense on any OS, couldn't we just note the time we sent the packet to the server and then have the server send a response. Measure time difference and presto, ping.

We have commands that need to be run on server (when runOnServer is true). Maybe you can use that to your advantage.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -401,6 +404,75 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
         return "Spawned block.";
     }
 
+
+
```
rzats: Some extra newlines here. The indentation for the entire function is slightly off at places - if you're running Eclipse/IDEA, set your code formatter to Terasology's (`docs/formatter.xml`) and autoformat.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -401,6 +404,75 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+	//Works on windows only , i don't know if you can even play the game on other platforms
```
rzats: Yep, Terasology's crossplatform. `ping` is a UNIX utility but its' output may be not identical to the Windows implementation - I'd use an alternative method of retrieving ping time (like sockets - thanks for the suggestion @portokaliu!)

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -401,6 +404,75 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+    @Command(shortDescription = "Your ping to the server", helpText = "The time it takes the packet " +
+    "to reach the server and back" , runOnServer = false, requiredPermission = PermissionManager.NO_PERMISSION)
+    public String ping(@Sender EntityRef sender ) {
+        //The class is just to tidy up the code , and not having random methods in this file
+        class InnerClass{
```
rzats: Since none of the methods are called more than once, this inner class only makes the function slightly longer. I'd move this code into the `ping` function or `private` helper methods within the `CoreCommands` class.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -401,6 +404,75 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+            arr = temp.split(" ");
+            String ret = "";
+            for(String s : arr){
+              if(s.contains("time=")){
+                String[] secondArr = s.split("=");
+                if(secondArr.length == 2){
+                  ret = secondArr[1];
+                } else {
+                  //incase the string "time=<value>" split by '=' isn't split into 2 strings
+                  return "An error occured";
```
rzats: Needs to be slightly more descriptive. Then again, it shouldn't even occur if you use a different way of retrieving ping 🙂

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -401,6 +404,75 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+                }
+              }
+            }
+
+            return ret;
+        }
+    }
+    //P.S Sorry for my english :D
```
rzats: Looks good to me! 😄 This comment needs to be removed when you finish the PR, though. Generally meta discussion like this should be located in the pull request's description.

Cervator:
@GooeyHub: ok to test

@ron-popov: hi! And thanks for the PR :-)

GooeyHub:

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
ron-popov: So remove the system command method and use a different one ?

rzats: Exactly! Use a socket to ping the server then record the delay on the client.

ron-popov:
I changed the ping command and now it is socket based , can someone take a look ?

GooeyHub:

portokaliu:
Tried on the same system and it seems to be working. Would require more than 1 PC to properly test imo

rzats:
Thanks for the PR! Merged and added you to the credits 😃

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -401,6 +407,50 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+                port = Integer.valueOf(arr[2]);
+            } catch(Exception ex){
+                return "An error occured" + ex.toString();
+            }
+            Instant starts = Instant.now();
+            Socket sock = new Socket(address, port);
+            Instant ends = Instant.now();
+            sock.close();
+            String response = String.valueOf(Duration.between(starts, ends));
+            String millis = "";
```
skaldarnar: Any reason not to use the following instead of the manual loop? @rzats 
```
String millis = response.replaceAll("[^\\d.]", "");
```

PR #2606, opened 2016-11-15
Cleanup of ping command.

skaldarnar:
Please test before merging, I'm on my laptop right now :turtle:

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -63,8 +63,8 @@ import org.terasology.rendering.FontColor;
 import org.terasology.rendering.nui.NUIManager;
 import org.terasology.rendering.nui.asset.UIElement;
 import org.terasology.rendering.nui.editor.layers.NUIEditorScreen;
-import org.terasology.rendering.nui.editor.systems.NUIEditorSystem;
 import org.terasology.rendering.nui.editor.layers.NUISkinEditorScreen;
+import org.terasology.rendering.nui.editor.systems.NUIEditorSystem;
 import org.terasology.rendering.nui.editor.systems.NUISkinEditorSystem;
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -410,43 +409,25 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
     @Command(shortDescription = "Your ping to the server", helpText = "The time it takes the packet " +
         "to reach the server and back", requiredPermission = PermissionManager.NO_PERMISSION)
     public String ping(@Sender EntityRef sender) {
+        Server server = networkSystem.getServer();
+        if (server == null) {
+            //TODO: i18n
```
rzats: Every other console command also needs this TODO note 😃 Though it might be a little annoying with format strings (like the one below).

skaldarnar: that's true 😆 I thought `String.format` looked better with the longer messages, but that might just be a personal thing...

skaldarnar:
@ron-popov thanks again for your initial work on this. I just reworked this method to remove some "smells" from the code. I'd like to explain the changes briefly and hope that will help you understand why they were made.

Your code for gettting the milliseconds from the duration was correct and working, so it was a good start. I first wanted to just replace the *String-to-Float* part with a single line:
```
String millis = response.replaceAll("[^\\d.]", "");
```
I double-checked the `Duration` class and found `toMillis()` which will shorten the whole thing a lot (less code -> less bugs): 
```
long delay = Duration.between(starts, ends).toMillis();
```

In addition, it is always a good idea to be as specific as possible when it comes to exceptions (and only catch those execeptions that are necessary). 

We really appreciate your contribution and hope to see you again soon - hopefully with more good additions :)

Issue #2293, opened 2016-04-13
Potentially unify the giveBlock and giveItem commands


Cervator:
After the work in #2290 and what came before it we might be able to just combine `giveBlock` and `giveItem` into a single `give` command with variants to accept only shapes for blocks, give a sensible warning if you asked for an item that can't be shaped, and so on.

There is a fair bit of overlap between the two commands, hopefully at least we can cut down on the duplication and in the best case completely unify them both.

Main trick is probably still supporting the different method variants based on parameters. Still probably not too hard to do.
##

MaxBorsch:
@Cervator If they were to be completely unified, in which Commands.java file should the new `give` command be? ItemCommands, CoreCommands?

skaldarnar:
@MaxBorsch I think `ItemCommands` fits the purpose better. Maybe we'll find a better name for it, the `giveItem` command seems to be the only thing implemented there. One could argue that the hand-held blocks are items of sort, so that should be fine.
There's always the option to change things later 😉

Cervator:
Closing as completed by #2613! Thanks @MaxBorsch :-)

PR #2613, opened 2016-11-19
Unified giveItem + giveBlock to give

MaxBorsch:
Implemented #2293 

giveBlock and giveItem commands are now just give. (+ better return messages)
Ex.)
```
give <Block or item name> (amount) (block shape)
```

Playing.md has not been adjusted for this.

Possible next steps:
- inventoryManager.giveItem doesn't support item quantity to give
- Uses prefab.getName() for items, should use a display name maybe?

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
@@ -17,7 +17,6 @@
 package org.terasology.world.block.entity;
 
 import com.google.common.base.Joiner;
-import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import org.terasology.assets.ResourceUrn;
 import org.terasology.assets.management.AssetManager;
@@ -267,10 +266,11 @@ public class BlockCommands extends BaseComponentSystem {
         return def.getUrn().toString().toLowerCase().contains(searchLowercase);
     }
 
-
-    @Command(shortDescription = "Adds a block to your inventory",
-            helpText = "Puts a desired number of the given block with the give shape into your inventory",
-            runOnServer = true, requiredPermission = PermissionManager.CHEAT_PERMISSION)
+    /**
+     * Called by 'give' command in ItemCommands.java to attempt to put a block in the player's inventory when no item is found.
+     * Called by 'giveBulkBlock' command in BlockCommands.java to put a block in the player's inventory.
+     * @return Null if not found, otherwise success or warning message
+     */
     public String giveBlock(
             @Sender EntityRef sender,
             @CommandParam("blockName") String uri,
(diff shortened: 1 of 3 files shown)


GooeyHub:

engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
```diff
@@ -45,22 +44,14 @@
 import org.terasology.utilities.Assets;
 import org.terasology.world.BlockEntityRegistry;
 import org.terasology.world.WorldProvider;
-import org.terasology.world.block.Block;
-import org.terasology.world.block.BlockComponent;
-import org.terasology.world.block.BlockExplorer;
-import org.terasology.world.block.BlockManager;
-import org.terasology.world.block.BlockUri;
+import org.terasology.world.block.*;
```
skaldarnar: Please don't use start imporst. If you are using IntelliJ, set the threshold value for start imports to something very high to avoid this.

engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
```diff
@@ -45,22 +44,14 @@
...
 import org.terasology.world.block.loader.BlockFamilyDefinition;
 import org.terasology.world.block.shapes.BlockShape;
 import org.terasology.world.internal.WorldProviderCoreImpl;
 
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
skaldarnar: star import (2)

engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
```diff
@@ -307,36 +299,19 @@ public String giveBlock(
                     return giveBlock(blockManager.getBlockFamily(new BlockUri(def.get().getUrn())), quantity, sender);
                 }
             } else {
-                return "No block found for '" + uri + "'";
+                return "Could not find an item or block matching \"" + uri + "\"";
```
skaldarnar: Showing the message "could not find item ..." within a method called `giveBlock` seems unintuitive. However, I'm not 100% sure how this can be resolved better. 
Since `giveBlock` is not a command anymore, you might use it to return the block URN if found, and do the error handling in the `give` or `giveBulkBlock` commands. Which, in turn, would mean duplication of the code... *sigh*

modules/Core/src/main/java/org/terasology/logic/inventory/ItemCommands.java
```diff
@@ -49,27 +49,37 @@
...
-            @CommandParam(value = "amount", required = false) Integer amount) {
+            @CommandParam(value = "amount", required = false) Integer amount,
+            @CommandParam(value = "blockShapeName", required = false) String shapeUriParam) {
+
+        int itemAmount = amount != null ? amount : 1;
         Set<ResourceUrn> matches = assetManager.resolve(itemPrefabName, Prefab.class);
         switch(matches.size()) {
             case 0:
-                return "Could not find any item matching \"" + itemPrefabName + "\"";
+                return blockCommands.giveBlock(client, itemPrefabName, amount, shapeUriParam);
```
skaldarnar: I'm still thinking about a way to handle this delegation "better". Another might be to indicate that no item was found right here, prepending the info to the output of `giveBlock`:
```Java
return "Could not find a matching item for \"" + itemPrefabName + "\". \n" + blockCommands.giveBlock(client, itemPrefabName, amount, shapeUriParam);
```

skaldarnar:
@GooeyHub: ok to test

GooeyHub:

MaxBorsch:
@skaldarnar I'll change `giveBlock` to a boolean showing whether it succeeded or not and have `give` do the message. It doesn't make for `giveBlock` to return a String any more since it's no longer a command.

skaldarnar:
@MaxBorsch that sound's good, just go ahead and update this PR.

GooeyHub:

MaxBorsch:
@skaldarnar I ended up finding a a way to finesse it into being more intuitive with half the giveBlock() calls and with the "No items or blocks" message being in the ItemCommands.java

MaxBorsch:
@skaldarnar It's updated.

Cervator:
I've updated the related documentation now in https://github.com/MovingBlocks/Terasology/commit/6d997c0a07e8bec663b4a3b464090b3658ad636e - belated thanks for getting this working @MaxBorsch :-)

PR #2618, opened 2016-11-21
Add Socket

SoniEx2:
### Contains

Adds Sockets. Use

```java
@In
SocketManager socket;
```

In a System to get access to sockets. It has TCP clients, but can be expanded later.

### How to test

This is a badly written IRC client that just barely works: https://github.com/SoniEx2/TeraIRC

- [x] Separate whitelists.
- [x] Whitelists need to be user-configurable somehow.
- [x] Rename to "Socket".
- [x] ~~Move DynamicInstanceProvider to a separate PR.~~ Easier to just keep it here.
- [x] ~~Disallow internet access on the client.~~ HARD. Depends on the whole CoreRegistry deprecation/removal thing and splitting the internal server from the client. Not high-priority.
- [x] ~~HTTP, with separate whitelists. Anything with TCP access also has HTTP access, but HTTP access does not give you TCP access.~~ Should be an external "HTTP" module, not part of this PR.
- [x] ~~Global whitelist that any module can use?~~ Would avoid some duplication, but can easily be misused.
- [x] Make the SocketManager thread-safe.

It is *deliberately allowed* to hand your SocketManager (and thus your whitelist) to another module. After all, it's trivial to wrap SocketManager for cross-module use, with or without being able to share SocketManager itself.

Also, if I knew how, I'd use Netty for this. The idea was to make an implementation-agnostic API, maybe with events, but I'm not very good at that.

### DIFF
--- a/engine/src/main/java/org/terasology/config/InternetConfig.java
+++ b/engine/src/main/java/org/terasology/config/InternetConfig.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2016 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.config;
+
+import com.google.common.collect.Maps;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonNull;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+import gnu.trove.set.TIntSet;
+import gnu.trove.set.hash.TIntHashSet;
+import org.terasology.naming.Name;
+
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.Map;
+
+/**
+ * Configuration file for all the internet BS you can think of.
+ */
+// TODO make thread safe and stuff
+public class InternetConfig {
+    /**
+     * All the TCP whitelists. Keyed by module ID.
+     */
+    private Map<Name, Hosts> tcpWhitelists = Maps.newTreeMap();
+
+    /**
+     * Retrieves the TCP whitelists.
+     *
+     * @return The TCP whitelists.
+     */
+    public Map<Name, Hosts> getTcpWhitelists() {
+        return tcpWhitelists;
+    }
+
+    /**
+     * POJO representing the hosts a module is allowed to access.
+     */
+    public static class Hosts {
+        /**
+         * The hosts themselves.
+         */
+        private Map<String, Ports> hosts = Maps.newTreeMap();
+
+        /**
+         * Retrieves the hosts.
+         *
+         * @return The hosts.
+         */
+        public Map<String, Ports> getHosts() {
+            return hosts;
+        }
+
+        public static class Handler implements JsonSerializer<Hosts>, JsonDeserializer<Hosts> {
+
+            @Override
+            public Hosts deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+                // We need this for the treemap.
+                Map<String, Ports> map;
+                try {
+                    // Ugh java
+                    map = context.deserialize(json, Hosts.class.getDeclaredField("hosts").getGenericType());
+                } catch (Exception e) {
+                    // Who cares
+                    throw new JsonParseException(e);
+                }
+                if (map == null) {
+                    return null;
+                }
+                Hosts h = new Hosts();
+                h.hosts = map;
+                return h;
+            }
+
+            @Override
+            public JsonElement serialize(Hosts src, Type typeOfSrc, JsonSerializationContext context) {
+                if (src == null) {
+                    return JsonNull.INSTANCE;
+                }
+                return context.serialize(src.hosts);
+            }
+        }
+    }
+
+    /**
+     * POJO representing the ports a module is allowed to access.
+     */
+    public static class Ports {
+        /**
+         * The ports themselves.
+         */
+        private TIntSet ports = new TIntHashSet();
+
+        /**
+         * Retrieves the ports.
+         *
+         * @return The ports.
+         */
+        public TIntSet getPorts() {
+            return ports;
+        }
+
+        public static class Handler implements JsonSerializer<Ports>, JsonDeserializer<Ports> {
+            @Override
+            public Ports deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+                // We need this custom deserialization (and this class) for 2 reasons:
+                // 1. We need to control TIntHashSet's no_entry_value.
+                // 2. We need to sanitize user input.
+                int[] array = context.deserialize(json, int[].class);
+                if (array == null) {
+                    return null;
+                }
+                Arrays.stream(array).filter(x -> (x < 1 || x > 65535) && x != -1).findAny().ifPresent(x -> {
+                    throw new IllegalStateException("Port must be in range 1 <= port <= 65535, or -1 for wildcard value: " + x);
+                });
+                TIntHashSet hashSet = new TIntHashSet(10, .5f, Integer.MIN_VALUE);
+                hashSet.addAll(array);
+                Ports p = new Ports();
+                p.ports = hashSet;
+                return p;
+            }
+
+            @Override
+            public JsonElement serialize(Ports src, Type typeOfSrc, JsonSerializationContext context) {
+                if (src == null) {
+                    return JsonNull.INSTANCE;
+                }
+                // Just defer to standard TIntHashSet serialization. This is fine.
+                return context.serialize(src.ports);
+            }
+        }
+    }
+}
--- a/engine/src/main/java/org/terasology/context/internal/ContextImpl.java
+++ b/engine/src/main/java/org/terasology/context/internal/ContextImpl.java
@@ -19,8 +19,6 @@ import com.google.common.collect.Maps;
 import org.terasology.context.Context;
 import org.terasology.registry.DynamicInstanceProvider;
 
-import java.lang.reflect.ParameterizedType;
-import java.util.Arrays;
 import java.util.Map;
 
 /**
@@ -78,6 +76,7 @@ public class ContextImpl implements Context {
             return providerClass.cast(provider);
         }
     }
+
     private final Map<Class<?>, DynamicInstanceProviderHolder<?>> providers = Maps.newConcurrentMap();
 
     @Override
--- a/engine/src/main/java/org/terasology/context/internal/ContextImpl.java
+++ b/engine/src/main/java/org/terasology/context/internal/ContextImpl.java
@@ -63,20 +63,6 @@ public class ContextImpl implements Context {
 
     // DynamicInstanceProvider stuff, kinda messy but it works
 
-    private static final class DynamicInstanceProviderHolder<T> {
-        private final Class<? extends DynamicInstanceProvider> providerClass;
-        private final DynamicInstanceProvider<T> provider;
-
-        private DynamicInstanceProviderHolder(Class<? extends DynamicInstanceProvider> providerClass, DynamicInstanceProvider<T> provider) {
-            this.providerClass = providerClass;
-            this.provider = provider;
-        }
-
-        private <U> DynamicInstanceProvider<U> get() {
-            return providerClass.cast(provider);
-        }
-    }
-
     private final Map<Class<?>, DynamicInstanceProviderHolder<?>> providers = Maps.newConcurrentMap();
 
     @Override
@@ -99,5 +85,19 @@ public class ContextImpl implements Context {
             return parent.getInstanceProvider(type);
         }
         return null;
+    }
+
+    private static final class DynamicInstanceProviderHolder<T> {
+        private final Class<? extends DynamicInstanceProvider> providerClass;
+        private final DynamicInstanceProvider<T> provider;
+
+        private DynamicInstanceProviderHolder(Class<? extends DynamicInstanceProvider> providerClass, DynamicInstanceProvider<T> provider) {
+            this.providerClass = providerClass;
+            this.provider = provider;
+        }
+
+        private <U> DynamicInstanceProvider<U> get() {
+            return providerClass.cast(provider);
+        }
     }
 }
--- a/engine/src/main/java/org/terasology/context/internal/ContextImpl.java
+++ b/engine/src/main/java/org/terasology/context/internal/ContextImpl.java
@@ -17,7 +17,10 @@ package org.terasology.context.internal;
 
 import com.google.common.collect.Maps;
 import org.terasology.context.Context;
+import org.terasology.registry.DynamicInstanceProvider;
 
+import java.lang.reflect.ParameterizedType;
+import java.util.Arrays;
 import java.util.Map;
 
 /**
@@ -26,11 +29,9 @@ import java.util.Map;
 public class ContextImpl implements Context {
     private final Context parent;
 
-    private final Map<Class<? extends Object>, Object> map = Maps.newConcurrentMap();
-
+    private final Map<Class<?>, Object> map = Maps.newConcurrentMap();
 
     /**
-     *
      * @param parent can be null. If not null it will be used as a fallback if this context does not contain a
      *               requested object.
      */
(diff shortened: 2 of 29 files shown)


GooeyHub:

Cervator:
@GooeyHub: ok to test

@SoniEx2: good to see some actual code! 

Hoping to see some good review and discussion here, especially for the change to `InjectionHelper` - I am not familiar with that enough to be comfortable reviewing that piece myself. Maybe @immortius @flo @Josharias @msteiger @skaldarnar ... somebody please! :-)

GooeyHub:

engine/src/main/java/org/terasology/internet/InternetManager.java
```diff
@@ -0,0 +1,34 @@
...
+public interface InternetManager {
+
+    /**
+     * Opens a TCP connection.
+     *
+     * @param hostname The target hostname.
+     * @param port     The target port.
+     * @return The socket.
+     * @throws IOException If an I/O error occurs.
+     */
```
oniatus: I would prefer to return the actual socket and leave the in/out-handling to the caller.

engine/src/main/java/org/terasology/internet/internal/InternetManagerImpl.java
```diff
@@ -0,0 +1,100 @@
...
+import java.security.AccessController;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
+import java.util.Map;
+import java.util.TreeSet;
+
+/**
+ * @author soniex2
+ */
+public class InternetManagerImpl implements InternetManager, DynamicInstanceProvider<InternetManager> {
```
oniatus: Please separate the provider from the implementation (i will add an additional comment about the per-module instance on the review)

SoniEx2: There are 2 reasons why I did it this way.

1. Good luck doing it any other way.
2. Uh I forgot the other reason but this feels more natural. We still want a global instance that everyone can use, even if that global instance does nothing, because other parts of the engine can see and access it, and we need something to return from CoreRegistry.get().

SoniEx2:
@oniatus This way is user-friendly and backwards compatible as we can just add permission sets (#1703) without modules needing to know or care about it. It's less things for the programmer to worry about ("does this thing I wanna use have permissions so I need to use that 2nd variant of the dependency injection? or can I use it directly?").

SoniEx2:
@skaldarnar You're meant to have a provider for the client and a provider for the (internal/integrated) server (in the same JVM), but that's not fully implemented yet (part of another issue). And in practice there really is only one instance of the provider: the `private final Map<Name, InternetManager> permCache;`.

I designed this to be future-proof in a way compatible with the current implementation.

GooeyHub:

SoniEx2:
@oniatus Still not gonna do that `@In(scope=Scope.MODULE)` thing. Being able to sandbox old modules is a nice feature to have.

oniatus:
@SoniEx2 I'm fine with that, but we should definitley document the feature somewhere after the PR is merged 😃  Maybe you want to have a look at the [checkstyle report](http://jenkins.terasology.org/job/TerasologyPRs/1108/checkstyleResult/new/) in the meantime, there is probably something more to clean up.

Two more open questions from my side:

1.  What would I have to implement, if i want sandboxed database access (e.g. giving access to a wrapped Connection) for some modules?
As far as I see, this would be:

    1. Implement a DatabaseManager
    2. Implement a DatabaseManagerProvider
    3. Add a new Subsystem to the engine which puts the manager and the provider in the root context
  
    I see some potential code duplication, especially in the Manager and Provider classes if we add more sandboxed content (probably also an ocp violation in the engine with the raising number of subsystems). If we only need one or two (Internet and Database are the only one coming to my mind), this is okay for me. If we need more, we should think about a way to make this open for extension of more modules without touching more classes (especially the engine subsystems).

2. (Maybe out of scope for this change): Where would the actual sandboxing happen?
Let's say we implement a DynamicInstanceProvider which checks if the gameplay module allowed the access of internet for a used module/the user accepted the required extra permissions for the modules somehow, then we would have to replace all existing dynamic instance providers with the new permission strategy.

Just some short ideas about the extensibility of the new feature. Maybe @skaldarnar or @Cervator are the better candidates for such questions, as I do not have the architectual farsight in this scope 😉 

In any case I think that both (Internet Access and general Sandboxing of some API) are very valuable features, keep going 👍

SoniEx2:
@oniatus 

1. For DB:
    1. Create a new **module**, "Database".
    2. Use the Internet subsystem to open database connections. This requires you to write it yourself but eh w/e.
    3. Implement a DatabaseManagerProvider to control access. DynamicInstanceProvider registration is available to modules (however, they can't retrieve them from the registry/Context).
2. For updating instances: Look at the InternetManagerProvider. Note how it contains a Name->InternetManager map. Just access that map, then update the instances.

GooeyHub:

GooeyHub:

GooeyHub:

SoniEx2:
@oniatus How do the configs look?

engine/src/main/java/org/terasology/internet/internal/InternetManagerImpl.java
```diff
@@ -15,9 +15,18 @@
...
+import gnu.trove.impl.unmodifiable.TUnmodifiableIntSet;
+import gnu.trove.set.TIntSet;
+import gnu.trove.set.hash.TIntHashSet;
+import org.terasology.config.Config;
+import org.terasology.config.InternetConfig;
+import org.terasology.context.Context;
 import org.terasology.internet.InternetManager;
 import org.terasology.internet.TCPSocket;
 import org.terasology.naming.Name;
+import org.terasology.registry.CoreRegistry;
```
rzats: Looks like this is unused.

engine/src/main/java/org/terasology/internet/internal/InternetManagerImpl.java
```diff
@@ -26,32 +35,63 @@
...
+        Map<String, TIntSet> allowedHosts = this.allowedHosts;
+        if (!(allowedHosts.containsKey(hostname) || allowedHosts.containsKey("*"))) {
+            // neither hostname nor wildcard in allowedHosts
+            // TODO maybe use a different exception type? (SandboxException or something?)
+            throw new IllegalArgumentException("Blocked hostname");
+        }
+        // TODO We have many options when both a wildcard entry and a specific entry exist. Which one should we pick?
+        // 1. Merge allowed ports.
+        // 2. Wildcard overrides specific hostname.
+        // 3. Specific hostname overrides wildcard. This option pleases me the most. (It also enables blacklists.)
```
rzats: So if I'm reading this correctly, in a whitelist containing `*.example.com` `foo.example.com` will no longer be whitelisted after `bar.example.com` is added to the config? That doesn't sound too intuitive - the second option makes more sense.

engine/src/main/java/org/terasology/internet/internal/InternetManagerImpl.java
```diff
@@ -26,32 +35,63 @@
...
     }
 
     @Override
     public TCPSocket openTCPConnection(String hostname, int port) throws IOException {
-        if (!allowedHosts.contains(hostname)) {
-            throw new IllegalArgumentException();
+        Map<String, TIntSet> allowedHosts = this.allowedHosts;
+        if (!(allowedHosts.containsKey(hostname) || allowedHosts.containsKey("*"))) {
+            // neither hostname nor wildcard in allowedHosts
+            // TODO maybe use a different exception type? (SandboxException or something?)
```
rzats: Sounds good - `IllegalArgumentException` is IMO best used in `Preconditions`-esque checks

```java
public int foo (int a) {
    if (a % 2 == 2) {
        throw new IllegalArgumentException ("wow you messed up");
    }
    /* ... */
}
```

and I can't find a good default exception for this usecase.

engine/src/main/java/org/terasology/config/InternetConfig.java
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * Configuration file for all the internet BS you can think of.
+ */
+// TODO make thread safe and stuff
+public class InternetConfig {
+    /**
+     * All the TCP whitelists. Keyed by module ID.
+     */
+    private Map<Name, Hosts> tcpWhitelists = Maps.newTreeMap();
```
rzats: AFAIK there's some sort of `TObjectObjectMap` class. Might be worth using alongside `TIntSet`.

engine/src/main/java/org/terasology/internet/internal/InternetManagerImpl.java
rzats: (Disregard that, I accidentally commented on the previous commit)

engine/src/main/java/org/terasology/internet/internal/InternetManagerImpl.java
SoniEx2: No, wildcards don't actually work (yet). `*.example.com` would only match a literal `*.example.com`. I only added `*` as an option so a dedicated "HTTP" module (not an engine subsystem) is a possibility.

But if you had `*` set to all ports allowed, and you added `example.com` with only port 80 allowed, then `example.com` would match first and override the `*` match - only port 80 would be valid for `example.com`.

If I do get around to implementing wildcards, then `*.example.com` will match both `foo.example.com` and `bar.example.com`. If `bar.example.com` exists on the whitelist, then `bar.example.com` would only match against it, and `foo.example.com` would still use the `*.example.com` entry.

This allows the "whitelist" to act as both a whitelist *and a blacklist*.

engine/src/main/java/org/terasology/config/InternetConfig.java
SoniEx2: There's no benefit in this case.

engine/src/main/java/org/terasology/internet/internal/InternetManagerImpl.java
rzats: That sounds like a far better idea, thanks for the detailed explanation 🙂

GooeyHub:

GooeyHub:

rzats:
@SoniEx2: looks like the TODO list in the description is fully finished - is this ready for a final review?

SoniEx2:
@rzats Not yet, looks like I overlooked something.

SoniEx2:
@rzats Fixed.

GooeyHub:

engine/src/main/java/org/terasology/context/internal/ContextImpl.java
```diff
@@ -54,12 +53,51 @@ public ContextImpl() {
...
+    @Override
+    public <T, U extends T> void putInstanceProvider(Class<T> type, DynamicInstanceProvider<U> provider) {
+        providers.put(type, new DynamicInstanceProviderHolder<>(provider.getClass(), provider));
+    }
+
+    @Override
+    public <T> DynamicInstanceProvider<T> getInstanceProvider(Class<? extends T> type) {
+        if (System.getSecurityManager() != null) {
+            System.getSecurityManager().checkPermission(new RuntimePermission("permGetInstanceProvider"));
+        }
```
flo: I would like to keep the Context interface and the injection concept as simple as possible.

Some idea I could think of to simplify it a little bit would be to have `Context` have a `inject` method so that we don't need to have a method that we must protect against all but one caller. Although I am not that happy with that idea.

engine/src/main/java/org/terasology/registry/DynamicInstanceProvider.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.terasology.registry;
+
+import org.terasology.context.Context;
+import org.terasology.naming.Name;
+
+/**
+ * A DynamicInstanceProvider provides an instance dynamically.
+ */
+public interface DynamicInstanceProvider<T> {
```
flo: `DynamicInstanceProvider` is kind of a rendundant name as a `Provider` is already someting that provides you dyamically with instances. Also the javadoc currently is kind of redundant to the class name.

When you just look at this class as another developer you would do very hard to guess why we need it and what to do with it.

I think the idea behind it should be described at one of the classes like this one and other API parts should refer to this class for an explanation. Also the class name should give a better indication what the class is good for. For example the class could be called `ModuleSecurityBoundInstanceProvider`.

Example for javadoc:

> Some managers need to be able to check, if a using module has actually the permission to use a certain feature. The idea is to do this check at injection time. At this point of time the module of the class that gets injected can be determined. With this information it is possible to inject a module specific instance to the class. The creation or determination of this module specific instance gets done via this interface. Please note that implementors of this class must make sure that there is no way for other modules to obtain a instance of this interface and by pass thus the injection.
> 
> Of course modules that let a instance be injected should ensure that there is no way to access this instance from outside. This should be document in each type that can be obtained via this interface.

engine/src/main/java/org/terasology/registry/DynamicInstanceProvider.java
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Returns the instance for the given module.
+     *
+     * @param moduleId The module ID.
+     * @param context The current context.
+     * @return An instance.
+     */
+    T getInstanceForModule(Name moduleId, Context context);
```
flo: I think Context should not be an argument, but something that gets passed to the constructor when the object gets created. Reasoning: The provider will live for the same duration like the objects it provides.

engine/src/main/java/org/terasology/socket/internal/SocketManagerImpl.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Map;
+
+/**
+ * Implements SocketManager.
+ */
+public class SocketManagerImpl implements SocketManager {
+    /**
+     * The allowed hosts.
+     */
+    //@Nonnull
```
flo: Comment and redundant javadoc (to name) can be removed.

SoniEx2:
Lots of the javadocs were added so my IDE wouldn't complain about missing javadocs.

engine/src/main/java/org/terasology/registry/CoreRegistry.java
```diff
@@ -69,4 +71,46 @@ public static void setContext(Context context) {
         return context.get(type);
     }
 
+    /**
+     * Returns this class as a Context object.
+     *
+     * @return A Context object representing this class.
+     */
+    public static Context asContext() {
+        return CoreRegistryContext.INSTANCE;
```
SoniEx2: This returns a Context that behaves like CoreRegistry - if you change the CoreRegistry backing context, it reflects on the CoreRegistryContext.

engine/src/main/java/org/terasology/socket/internal/SocketSubsystem.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public class SocketSubsystem implements EngineSubsystem {
+    @Override
+    public String getName() {
+        return "Socket";
+    }
+
+    @Override
+    public void initialise(GameEngine engine, Context rootContext) {
+        rootContext.putInstanceProvider(SocketManager.class, new SocketManagerProvider());
```
SoniEx2: Where should it go? We don't have separate contexts yet...

SoniEx2: Also, if sockets are implicitly closed, how do we give modules enough time to gracefully shutdown?

engine/src/main/java/org/terasology/registry/CoreRegistry.java
flo: As suggested above, I recommend that the thing you register at the context should have the same live span as the context. Thus using something that lives over multiple context lifespans does not make sense.

engine/src/main/java/org/terasology/socket/internal/SocketSubsystem.java
flo: Actually there are 2 `Context` objects usually. One engine context that exists until the program terminates. Besides that there is the context that lives for the duration of the game that inherits from the engine context.

flo: if you explicitly  want to give modules a chance to do something before the disconnect you can send an event (e.g. to world entity) which they can then listen to.

SoniEx2:
Blow up this PR, fork it, idc.

I'm just not gonna implement an unsafe system for something that syncs code.

GooeyHub:

GooeyHub:

skaldarnar:
@SoniEx2 This has been open for some time now, and some of the comments have not been addressed yet (neither have been changed nor commented on further discussion). What is the current state of this PR?

@flo @rzats @oniatus Any comments on whether we can get this merged _soonish_™️ or if it needs more work and/or discussion.

engine/src/main/java/org/terasology/exception/SandboxException.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.terasology.exception;
+
+import org.terasology.module.sandbox.API;
+
+/**
+ * Used when a resource access has been blocked by the sandbox.
+ */
+@API
+public class SandboxException extends Exception {
```
SoniEx2: Actually, SandboxException is to be used when the user chooses to restrict what a module can do. So, e.g. a clipboard API would throw SandboxException instead of returning null, if the user disallows clipboard access.

SoniEx2:
@skaldarnar The PR is uh... dead, mostly.

It needs to be rewritten from scratch and we need a better system than `@In`s for cross-module sandboxing. The internet manager provider wouldn't have to be registered as part of the engine/into the core context, if we had such better system. (Basically, it'd allow for better sandboxing of built-in systems.)

skaldarnar:
Closing in favor of (hopefully soonish) new PR on the matter. We still made good way with this one and all the discussion around it. 

@SoniEx2 Thanks for checking back and letting me know. Feel free to open a new PR whenever ready.

oniatus:
For the interested: The discussion point on the forums would be http://forum.terasology.org/threads/config-context-and-security.1761/

PR #2583, opened 2016-10-29
Added the Turkish translation for i8n #2309

aybuke:
### Contains

Brief description of what the PR does like "Fixes #12345"
### How to test

Brief description of how to test / confirm this PR before merging
### Outstanding before merging

If anything. You can use neat checkboxes! Feel free to delete if not needed
- [ ] Need to consider use case x
- [ ] Still have to adjust the wiki doc
- [ ] Will need translation work

PR #2620, opened 2016-11-22
Merge latest translations

Cervator:
Grabbed everything from our live Weblate server, pushed to the `weblate` branch.

Merged on top of latest `develop`, pushed to a temporary `weblateTest` I'm also going to configure a secondary test Weblate to use (at least for engine). @qwc is looking at maybe doing an upgrade and enabling machine translation.

http://weblate-testing.terasology.net

In this batch Arabic was enabled, as requested a while ago - but I don't think we actually have any translations for it yet.

Turkish was added, we got a somewhat problematic PR submitted for it with #2583 that wasn't written quite right. @tdgunes did a bit of translation from within Weblate resulting in https://github.com/MovingBlocks/Terasology/commit/8a4c86a6daf582b2d59949ba7e4b75de5c873d24 - do you think any other strings from #2583 could be included? It looks like some there have translations that are still untranslated.


GooeyHub:

tdgunes:
@Cervator Sorry but yes, there are. Would it be possible to eliminate other _placeholder_ translations to see what's remaining?

skaldarnar:
re untranslated languages: #2621

Cervator:
@tdgunes I would just copy paste + adjust the translations from the older PR directly into `menu_tr.lang` and do a PR this time, then we'll figure out via #2621 how to make that clearer within Weblate in the future :-)

You could grab this branch then add the additional translations on top, make a new PR, and that one will cover both when merged. Or I'll grab this one later tonight my time, then can just use develop like normal.

Issue #2590, opened 2016-11-05
Crash in multiplayer - chunk loading / rendering null issue, maybe related to dupe chunk loading [$50]


qwc:
### What you were trying to do
Running around and harvesting some sand.

### What actually happened
Crash.
### How to reproduce

Goood question, I will update WHEN I am able to reproduce this.

### Log details and game version

Logfile: http://pastebin.com/9z7FQ7UK

Latest unstable locally and on server.

### Computer details
Arch Linux Client, Ubuntu docker image on server.

Cervator:
Renamed the title a bit, this is probably our biggest known crash issue at the moment, although narrowing down the details or reproducing reliably is problematic. It seems to have something to do with duplicate chunk loading feeding into a problem when then trying to render a newly loaded chunk. Documented some more over in #2477

Cervator:
Interestingly it may have just happened for the first time *without* a reference to dupe chunk loading, but timed superbly to another player joining the server (another way to cause more chunk loads?)

Oddly the log file got cut at one point a few minutes before the crash, started with a `Terasology-.log` then continued in `Terasology-mmo.to-188.68.52.110-25778.log`

```
...
12:37:24.519 [main] INFO  o.t.r.dag.RenderTaskListGenerator - Task list generated in 2.021 ms
12:37:24.520 [main] INFO  o.t.r.dag.RenderTaskListGenerator - 30 nodes, 27 enabled - 88 tasks (excluding marker tasks) out of 115 potential tasks.
12:37:24.520 [main] INFO  o.t.r.dag.RenderTaskListGenerator - ----------------------------------------------------------
12:37:24.668 [main] WARN  o.t.rendering.nui.asset.UIFormat - Field 'emptyIcon' not recognized for interface org.terasology.rendering.nui.UIWidget in {"type":"UIIconBar","id":"breathBar","icon":"engine:icons#bubble","family":"breathBar","emptyIcon":"engine:icons#burstBubble","halfIconMode":"shrink","spacing":2,"maxIcons":10,"layoutInfo":{"use-content-width":true,"use-content-height":true,"position-left":{"target":"CENTER"},"position-right":{"target":"RIGHT","widget":"healthBar"},"position-bottom":{"target":"TOP","widget":"healthBar","offset":1}}}
[...]

PR #2624, opened 2016-11-25
Fix checkstyle violations

Tropid:
### Contains

Fixes all checkstyle error violations.

### How to test

Run ./gradlew checkstyleMain or open the project in eclipse with the checkstyle plugin active (should also work with any other IDE).

### Critical parts

- I added an implementation for equals() in ChunkImpl because the hashCode() method is overwritten
- I renamed some constants in /engine/src/main/java/org/terasology/rendering/openvrprovider/ControllerListener.java
- I changed a lambda expression to an anonymous class instantiation as the checkstyle tool seems to get confused by variable declarations inside lambdas (in my defense, this was a big lambda function so in my opinion it's not really worse to use an anonymous class) -> /engine/src/main/java/org/terasology/logic/console/commandSystem/ConsoleCommand.java


GooeyHub:

engine/src/main/java/org/terasology/world/chunks/internal/ChunkImpl.java
```diff
@@ -446,6 +446,21 @@ public int hashCode() {
...
+            return true;
+        }
+
+        if (!(obj instanceof ChunkImpl)) {
+            return false;
+        }
+
+        ChunkImpl that = (ChunkImpl) obj;
+
+        return Objects.equal(this.chunkPos, that.chunkPos);
```
Cervator: This one makes me wonder. Under regular circumstances you would indeed expect two chunks at the same location to equal each other. But are we setting ourselves up for potential bugs here?

Two cases come to mind:

* We have a known issue about duplicate chunks being loaded (#2590) - could this somehow affect that? If somehow a dupe chunk loads but a system interaction changes only one of them. I don't know if dupe chunks at all persist beyond a timing issue or something causing two load requests to be issued.
* In the future could somebody come up with a way where we could have two *valid* chunks in the same location? Imagine something related to dimensions somehow, either completely separate worlds (why anybody would compare two such chunks I don't know .. but MC does relate the position of Nether chunks vs Overworld chunks so who knows) or in-place chunks where a weird bizarro world shows two different players different chunks. Or phasing maybe (like bits of world with progression locked to individual players)

Perhaps we should add in hashing some additional stuff from each chunk and compare those as well? I don't know how frequently this method is likely used (or really a parent implementation)

engine/src/main/java/org/terasology/rendering/openvrprovider/ControllerListener.java
```diff
@@ -5,16 +5,16 @@
...
-    public static int k_EAxis_Trigger = 1;
-    public static int k_EAxis_TouchPad = 0;
-    public static long k_buttonTouchpad = (1L << JOpenVRLibrary.EVRButtonId.EVRButtonId_k_EButton_SteamVR_Touchpad);
-    public static long k_buttonTrigger = (1L << JOpenVRLibrary.EVRButtonId.EVRButtonId_k_EButton_SteamVR_Trigger);
-    public static long k_buttonAppMenu = (1L << JOpenVRLibrary.EVRButtonId.EVRButtonId_k_EButton_ApplicationMenu);
-    public static long k_buttonGrip = (1L << JOpenVRLibrary.EVRButtonId.EVRButtonId_k_EButton_Grip);
-    public static float triggerThreshold = .25f;
+    int LEFT_CONTROLLER = 0;
+    int RIGHT_CONTROLLER = 1;
+    int EAXIS_TRIGGER = 1;
```
Cervator: Any chance these names are made to match naming in another place, like the VR provider itself?

@indianajohn - ping! :-)

Cervator:
@Tropid Hi there and thanks! Looking good, noted two things I'm curious about :-)

@GooeyHub: ok to test

GooeyHub:

engine/src/main/java/org/terasology/world/chunks/internal/ChunkImpl.java
Tropid: In case of another dimension, we surely should hash another value and also compare another value. But you are correct this could have unforeseen changes.

The important part of the `Object` contract here should be:

`hashCode()` must always return the same code for two objects which are equal according to the `equals()` implementation.

This is true here. I'll adjust the implementation so that `equals` calls only super and provide a comment why the super `equals()` implementation is correct in this case.

engine/src/main/java/org/terasology/rendering/openvrprovider/ControllerListener.java
Tropid: In this case, I'd suggest adding the SuppressionWarnings module to the checkstyle.xml and add a `@SuppressWarnings` annotation to the class.

GooeyHub:

Cervator:
Thanks! Merged and added to credits :-)

Tropid:
Have you noticed I also fixed the violations in the core project? I missed that at first and don't want to sneak in any unnoticed changed.

Cervator:
I looked over all the affected files just in case, but am not too worried exactly where they are. It's all in the same repository after all :-)

PR #2627, opened 2016-11-26
Addressed a number of issues flagged by Jenkins

emanuele3d:
### Contains
Fixes for a number of warnings given by Jenkins during the build process.

Jenkins' output can be found at the end of each PR, under the Details link. The issues this PR has been addressing are just a small portion of the total, but have been found in [**this**](http://jenkins.terasology.org/job/TerasologyPRs/1093/analysisResult/) specific build output.

### How to test
- start Terasology
- load a game
- everything should work normally

### Note
- Given the breadth of the changes, even though many are small and probably inconsequential, we need to check if this PR gives any problem to known modules. I.e. a number of changes regard unused methods and fields. Perhaps modules were using them? This would need verifying in a workspace including all the modules. But who, who could have such a megaworkspace?
- I've used @SuppressWarning annotations in some cases. But I don't know if Jenkins will recognize them as I've used only default strings, i.e. "unchecked" or "access-method".

### DIFF
--- a/engine-tests/src/test/java/org/terasology/rendering/dag/RenderTaskListGeneratorTest.java
+++ b/engine-tests/src/test/java/org/terasology/rendering/dag/RenderTaskListGeneratorTest.java
@@ -17,13 +17,10 @@ package org.terasology.rendering.dag;
 
 import com.google.common.collect.Lists;
 import org.junit.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.List;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 
 public class RenderTaskListGeneratorTest {
 
@@ -106,6 +103,7 @@ public class RenderTaskListGeneratorTest {
         return node;
     }
 
+    @SuppressWarnings("static-access") // actual node classes are not meant to be static
     private class AlphaNode extends AbstractNode {
         @Override
         public void initialise() {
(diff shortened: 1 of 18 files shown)


GooeyHub:

emanuele3d:
The second commit has the same message because I thought I was amending the first but didn't quite work. :confused:

GooeyHub:

emanuele3d:
By the way, @indianajohn a good number of warnings are generated in 

[org.terasology.rendering.openvrprovider](http://jenkins.terasology.org/job/TerasologyPRs/1096/analysisResult/package.-278010644/) (link to the build output)

Can you address them now that the code is still "fresh"? :)

emanuele3d:
Strange. The build for the second commit found a bunch of new warnings. But it doesn't make sense because the second commit only modifies a couple of files. :confused:

Cervator:
@emanuele3d likely that's just from the PR job in Jenkins jumping around between different PRs. So the change in warnings between builds might be reporting a comparison to a completely different branch :-)

More of a safe thing to judge after we merge. Which I'll do momentarily!

Cervator:
Tested out briefly in my mega-workspace - no issues noted. The quirky "Access could be package-private" warnings (by IntelliJ, I think, more than Checkstyle) are all over the place. Not really doing much about them as they're a best practice possibility, but a bit of an edge case I don't think we've thought much about. Really low priority in any case and might even just look weird to some "Huh, shouldn't this explicitly be public or private?"

I ran over 4 classes in the OpenVR package real quick since one change was made to one of them anyway (other style fix PR). Most the warnings are gone but a few trickier ones may remain.

Issue #2208, opened 2016-03-05
Improved particle systems [$25]


flo:
Terasology has a basic particle system implementation. It could be extended to support advanced features like particles that leave a trails.
### Ribbons

Examples of ribbon like effects in other games/engines:
- http://docs.runicgames.com/wiki/Missile_Particles_and_Ribbons
- https://docs.unrealengine.com/latest/INT/Engine/Rendering/ParticleSystems/Reference/TypeData/Ribbon/index.html

Those could be used to make spell like effects.
### Force fields

Other advanced features could be force fields that change the direction of particles.
Like for example force fields in blender:
- https://www.blender.org/manual/de/physics/force_fields.html
### Animated particles

Particles could have an animation that is based on a tile map and that plays in a loop.
## Hints for new developers:

Familiarize yourself wit the Entity-System architecture of terasology: e.g. by reading  ( https://github.com/MovingBlocks/Terasology/wiki/Entity-System-Architecture https://github.com/MovingBlocks/Terasology/wiki/Events-and-systems ) and by trying it out in practice.

The current particle system implementation can be found in the class BlockParticleEmitterSystem. It makes entities with a "BlockParticleEffect" compoent to a particle system.

A "prefab" that uses the component "BlockParticleEffect" is dustEffect.prefab. It can be spawned with the command "spawnPrefab dustEffect". Prefabs (at least those in modules) reload automatically, so you don't have to restart terasology. It is enough to spawn antother instance of the prefab to see the effect of your change in the json file.
##

Cervator:
Quick reference to a force shield that was done in the past: http://forum.terasology.org/threads/light-shadow.766/page-3#post-11413

Likewise there are a few similar usages here and there of particles, but centralizing the system + making better doc/tutorial on how to use it would be great :-)

flo:
@Cervator that is a different kind of force field. With "force field"  I didn't mean something visual, but just something that makes particles take differnt paths.

For example you could have a force field that pulls particle towards a center. Or makes particles start spinning around a center. Or the force field could push particles into a certain direction after they have flown into it.

Cervator:
@flo oooohh! Okay. I get it now :-)

Josharias:
I wonder if any of this is salvageable from Linus: https://github.com/LinusVanElswijk/Terasology/tree/feature/GeneralParticleSystems

Remi05:
[...]

PR #2632, opened 2016-11-28
New Particle System

MaxBorsch:
Discussed heavily in #2208 , initial large commit for reviews, still quite a bit of things to do:

- [ ] **Create ParticleBuilder**
- [ ] **Convert legacy particle stuff (block effects mostly)**
- [ ] **Long-range particlesystem registry, "Chimney-problem"**
- [ ] **Particle system does not stop rendering when far away / unloaded? Unconfirmed.**
- [ ] Multiple renderers?
- [ ] Less hacky "OverheadParticleSystem" implementation for rain-like effects.
- [ ] Optimize updater to partition ParticlePool based on emissionRate of emmiters, not evenly.
- [ ] **Force-fields**
- [ ] Create documentation
- [ ] Annotate code

I'm sure there's more that I forgot to list, any reviews are welcome!

**In-game demo usage commands:**
`spawnExplosion` - spawns everlasting explosion
`addExplosion` - spawns explosions too but always adds to the same system

On another note, I'm actually a high school student and I'm participating in the GCI this year! So I'll continue working on this, but occasionally do some GCI tasks (like that long-overdue introduction post) 
:-) @Cervator @emanuele3d


GooeyHub:

emanuele3d:
Ok, can't go into the details right now (00:35) but I recommend you move the "particles" package one level up, as I imagine most files do not actually belong under "rendering". So, it would become: org.terasology.particles.

Also, I don't think you need to have an "internal" subpackage. The most important classes can go right in "particles", so that they are the first encountered and acquire prominence. I'd also eliminate the "data" package and promote those two files one level up. DisplayListParticleRenderer.java can be left alone in particle/rendering for the time being.

It would be nice to reorganize and reduce the commit list, but I can see that would be quite a bit of work on something that is largely inherited.

I'll go over the rest of the PR in the next few days. Grab a comfy chair as this PR will likely take some time. I mean, you do want to beat @tdgunes record of 172 conversations in a single PR, right? :stuck_out_tongue_winking_eye:

engine-tests/src/test/java/org/terasology/rendering/particles/internal/ParticlePoolTest.java
```diff
@@ -0,0 +1,202 @@
...
+import org.terasology.rendering.particles.internal.data.ParticlePool;
+import org.terasology.utilities.random.FastRandom;
+import org.terasology.utilities.random.Random;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Created by Linus on 28-2-2015.
```
rzats: Terasology doesn't use author headers - please remove these (or replace with detailed and useful Javadoc, wink wink nudge nudge 😉)

engine/src/main/java/org/terasology/rendering/particles/ParticleCommands.java
```diff
@@ -0,0 +1,538 @@
...
+        psEntityBuilder.addComponent(particleSystemComponent);
+
+        EntityRef particleSystem = psEntityBuilder.build();
+        particleSystemComponent.addEmitter(createEmitter(particleSystem));
+
+        return particleSystem;
+    }
+
+    private EntityRef createEmitter(EntityRef particleSystem) {
+        EntityBuilder emmiterEntityBuilder = entityManager.newBuilder();
```
rzats: Did you mean "emitter"?

engine/src/main/java/org/terasology/rendering/particles/ParticleData.java
```diff
@@ -0,0 +1,41 @@
...
+    public final Vector2f textureOffset = new Vector2f();
+
+    // 3d vectors
+    public final Vector3f position = new Vector3f();
+    public final Vector3f previousPosition = new Vector3f();
+    public final Vector3f velocity = new Vector3f();
+    public final Vector3f scale = new Vector3f();
+
+    // 4d vectors
+    public final Vector4f color = new Vector4f();
```
rzats: I wonder if the [`Color`](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/Color.java) class from NUI will be more appropriate here? (If so, it might need to be extracted from NUI into a general utility package - or [TeraMath](https://github.com/MovingBlocks/TeraMath).)

engine/src/main/java/org/terasology/rendering/particles/components/ParticleSystemComponent.java
```diff
@@ -0,0 +1,115 @@
...
+
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * Represents a particle system. A particle system must have emitters in order to produce particles.
+ */
+public class ParticleSystemComponent implements Component {
+
+    // Reference to owner entity
```
rzats: Some weird comment styles in this class. I'd stick to Javadoc style for everything:

```java
/**
 * Stuff
 * More stuff
 */
```

engine/src/main/java/org/terasology/rendering/particles/components/generators/TextureOffsetGeneratorComponent.java
```diff
@@ -0,0 +1,78 @@
...
+import org.terasology.rendering.assets.texture.Texture;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * Created by Linus on 13-4-2015.
+ *
+ * T
```
rzats: T? Should be removed (along with the Created by header)

engine/src/main/java/org/terasology/rendering/particles/functions/affectors/EnergyColorAffectorFunction.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public void beforeUpdates(final EnergyColorAffectorComponent component,
+                              final Random random,
+                              final float delta
+    ) {
+        navigableGradientMap.clear();
+        for(EnergyColorAffectorComponent.EnergyAndColor energyAndColor: component.gradientMap)
+        navigableGradientMap.put(energyAndColor.energy, energyAndColor.color);
```
rzats: Not indenting for-blocks properly gets you [goto fail](https://www.cigital.com/blog/understanding-apple-goto-fail-vulnerability-2/)!

engine/src/main/java/org/terasology/rendering/particles/functions/affectors/EnergyScaleAffectorFunction.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Created by Linus on 11-3-2015.
+ */
+public final class EnergyScaleAffectorFunction extends AffectorFunction<EnergyScaleAffectorComponent> {
+
+    public EnergyScaleAffectorFunction() {
+        super(EnergyScaleAffectorComponent.class, DataMask.ENERGY, DataMask.SCALE);
+    }
+
+    private NavigableMap<Float,Vector3f> scaleMap = Maps.newTreeMap();
```
rzats: Should be placed just before the constructor

engine/src/main/java/org/terasology/rendering/particles/internal/rendering/DisplayListParticleRenderer.java
```diff
@@ -0,0 +1,142 @@
...
+
+import org.lwjgl.opengl.GL11;
+import org.lwjgl.opengl.GL13;
+import org.terasology.math.geom.Vector3f;
+import org.terasology.rendering.assets.material.Material;
+import org.terasology.rendering.particles.components.ParticleSystemComponent;
+import org.terasology.rendering.particles.internal.data.ParticlePool;
+import org.terasology.rendering.particles.internal.data.ParticleSystemStateData;
+
+import static org.lwjgl.opengl.GL11.*;
```
rzats: No wildcard imports, please! Go to your IDE options and set the minimum amount of imports until they're compressed into a * import to something large like 999.

engine/src/main/java/org/terasology/rendering/particles/ParticleData.java
emanuele3d: Hmmm... the internal representation of the Color class is an int. That's 4 bytes vs the 4x4=16 bytes of color precision of a Vector3f. From a performance, storage and bandwidth perspective of course it has its appeal, but I'd maybe offer it as a possible alternative rather than the default, perhaps an "HighPerformanceParticleData" class. 

I'd recommend leaving this as it is for the time being but writing a proposal issue about it, so we don't loose the idea.

rzats: Other advantages I was thinking of are validation (values passed to the constructor have to be between 0-255), preset static colors and helper methods. Though the lower space complexity is nice as well 🙂

emanuele3d: Why are channel values between 0-255 an advantage? Once you are on the GPU everything is float anyway.

rzats: Fair enough - that particular advantage was just a random assumption of mine :)

engine-tests/src/test/java/org/terasology/rendering/particles/internal/ParticlePoolTest.java
MaxBorsch: Yeah that's the Annotate code TODO on the PR, I still have to go over some classes and replace Linus's auto-generated headers with descriptions.

engine/src/main/java/org/terasology/rendering/particles/ParticleCommands.java
MaxBorsch: Nope that's the system to attach emitter to, these functions are not meant to be used or even looked at, I'm removing / replacing them later, these are hacks for debugging.

rzats: I was actually referring to the spelling ("emmiter"), but if this is removed anyway it doesn't really matter 🙂

engine/src/main/java/org/terasology/rendering/particles/internal/rendering/DisplayListParticleRenderer.java
MaxBorsch: Don't blame me that was Linus D: I have my wildcards off, I haven't gotten the chance to fix all his styling issues, I'll fix it next commit though. Thanks for pointing this out.

engine/src/main/java/org/terasology/rendering/particles/components/ParticleSystemComponent.java
MaxBorsch: Yeah my bad, I only remembered how javadoc is supposed to be formatted 3/4 of the way through working on this, I'll go over all my other stuff and make sure it's documented.

Cervator:
@GooeyHub: add to whitelist

@MaxBorsch: thanks! :-)

GooeyHub:

engine/src/main/java/org/terasology/rendering/particles/internal/rendering/DisplayListParticleRenderer.java
skaldarnar: Would this be caught by IntelliJ's format code/optimize imports on PRs? I've checked the boxes for my PRs all the time, I think this will apply to all modified files in the PR...

engine/src/main/java/org/terasology/rendering/particles/DataMask.java
```diff
@@ -0,0 +1,55 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.particles;
+
+/**
+ * Created by Linus on 7-3-2015.
+ */
+public enum DataMask {
```
skaldarnar: Can you think of a better name, describing the kind of data mask better? Maybe `ParticleDataMask` would be enough already...

engine/src/main/java/org/terasology/rendering/particles/ParticleManagerInterface.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.terasology.rendering.particles;
+
+import org.terasology.rendering.particles.functions.affectors.AffectorFunction;
+import org.terasology.rendering.particles.functions.generators.GeneratorFunction;
+
+/**
+ * Created by Linus on 10-3-2015.
+ */
+public interface ParticleManagerInterface {
```
skaldarnar: The `...Interface` suffix looks weird. I'd go for just `ParticleManager`, an implementation would then be named `ParticleManagerImpl` (to follow the style of the project).

engine/src/main/java/org/terasology/rendering/particles/components/OverheadParticleSystem.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package org.terasology.rendering.particles.components;
+
+import org.terasology.entitySystem.Component;
+
+/**
+* The emitters of systems with this component will constantly float above the camera (for rain effects)
+* */
+public class OverheadParticleSystem implements Component {
```
skaldarnar: Hmm, potential confusion incoming. There might be `ParticleSystemComponent`s and `ParticleSystemSystem`s. I'd stick with the naming convention to end all components with `Component` and all systems with `System`, otherwise one would think that this is a system instead of a component.

Edit: the pattern is applied to the `ParticleSystemComponent` class already, so this is a call for consistency.

engine/src/main/java/org/terasology/rendering/particles/ParticleData.java
emanuele3d: @skaldarnar can you elaborate? What you do you mean with "there's also color specs in the range [0..1]"?

engine/src/main/java/org/terasology/rendering/particles/components/ParticleEmitterComponent.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+
+/**
+ * Created by Linus on 1-3-2015.
+ */
+public class ParticleEmitterComponent implements Component {
+    public static final int INFINITE_PARTICLE_SPAWNS = -1;
+
+    // Reference to owner entity
+    public EntityRef entityRef;
```
skaldarnar: probably should be renamed to `ownerEntity` in that case...

engine/src/main/java/org/terasology/rendering/particles/components/ParticleEmitterComponent.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public boolean isEmmiting = true;
+
+    public float maxLifeTime = Float.POSITIVE_INFINITY;
+    public int particleSpawnsLeft = INFINITE_PARTICLE_SPAWNS;
+
+    @Owns
+    private List<EntityRef> generators = new ArrayList<>();
+
+    /*
```
skaldarnar: javadoc formatting

engine/src/main/java/org/terasology/rendering/particles/components/ParticleSystemComponent.java
skaldarnar: Commented on this in the previous class, too. Probably the IDE reformatting applied to all classes should catch this.

engine/src/main/java/org/terasology/rendering/particles/functions/affectors/EnergyColorAffectorFunction.java
skaldarnar: ➕  use braces 😉

engine/src/main/java/org/terasology/rendering/particles/components/OverheadParticleSystem.java
emanuele3d: I haven't looked into the code yet but I'm thinking if a top-level class is needed, that could be a ParticlesManager or a ParticleSystemsManager.

skaldarnar:
General question for @MaxBorsch @emanuele3d: can the particle system be hooked up with general graphics settings, so that the amount of particles is controlled by such option? Or should there be a separate control in settings to control (max) amount of particles drawn on screen.

emanuele3d:
@skaldarnar: tricky tricky tricky. 

Ideally each effect provides its own settings. For some small effects it wouldn't make sense to reduce the already small number of particles. For larger effects, say, rain or snow, it would make sense to have a percentage-based setting, where 100% is the number intended by the developer and 0 is no particles at all. 

For the time being we might want a global 0-100% setting, reducing all particles proportionally. Limiting the maximum number of particles I feel is a blunter option. A 10K particle effect could be as computationally expensive as a 100 particle one, depending on everything associated with those effects.

skaldarnar:
Sounds reasonable. The definite settings control needs to be checked against performance on different systems, nonetheless. The scale could be expanded to 200% for really good systems and particle lovers out there 🚀

emanuele3d:
> The scale could be expanded to 200% for really good systems and particle lovers out there :rocket:

Once in place of course I guess it wouldn't be technically difficult to have slider going all the way to 200% and beyond. 

A simple particle density increase (typical if you just increase emission rate) is not necessarily better though. As an artist, if I wanted to increase the number of raindrops I wouldn't increase their density, I'd keep the density constant and I'd increase the distance limit, for a wider effect. But artistically the relationship is tricky, because until some threshold I might want to increase the rain density instead, and only reaching that density threshold I might switch (more or less abruptly) to increasing distance limits. So, the developer/artist is really best suited to make these kind of decisions. That's why I'd limit it to a 0-100% slider: 100% is the best effect as the developer/artist intended it. Anything less is a compromise due to the available hardware, but anything more wouldn't be beneficial. 

I guess if we really want some kind of generic parameter we could simply call it "quality", then let the developer/artist decide what a 50% quality means: less particles, lower resolution sprites, less sophisticated rendering, you name it. But a global parameter would be really unfair toward computationally inexpensive effects.

engine/src/main/java/org/terasology/rendering/particles/internal/updating/ParticleUpdater.java
```diff
@@ -0,0 +1,48 @@
...
+    void updateStateData(final EntityRef entityRef,
+                         final BiMap<Class<Component>, AffectorFunction> registeredAffectorFunctions,
+                         final BiMap<Class<Component>, GeneratorFunction> registeredGeneratorFunctions);
+
+    void update(float delta);
+
+    Collection<ParticleSystemStateData> getStateData();
+
+    static ParticleUpdater create(Physics physics, Logger logger) {
+        return new ParticleUpdaterImplementation(physics);
```
emanuele3d: Not all particles need physics. And why do you need a logger parameter if you don't use it?

engine/src/main/java/org/terasology/rendering/particles/internal/updating/ParticleUpdater.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Collection;
+
+/**
+ * Created by Linus on 16-4-2015.
+ */
+public interface ParticleUpdater {
+
+    void register(EntityRef entity);
+    void dispose(EntityRef entity);
```
emanuele3d: These are very generic. What kind of entity is expected?

emanuele3d:
It's late and I better go to bed. I'd like to conclude this first round of feedback saying that in my opinion you should split the foundation of the system into another branch and submit it as a separate PR. This will make it vastly easier to review before we get into figuring out exactly how individual effects are achieved. Aim to submit around 10 interfaces/classes which are the basis for everything else. Leave out anything that is not strictly essential, such as affectors/generator implementations, example commands and tests. Just leave the absolutely bare necessities. And please squash the changes into a single commit. We'll go from there.

engine/src/main/java/org/terasology/rendering/particles/ParticleData.java
skaldarnar: > color range in [0..1]
that was just a hint that there probably is too much implicit information on the color format using just a Vec3f.

engine/src/main/java/org/terasology/rendering/particles/internal/ParticleSystemManager.java
```diff
@@ -0,0 +1,213 @@
...
+        registerGeneratorFunction(new VelocityRangeGeneratorFunction());
+        registerGeneratorFunction(new TextureOffsetGeneratorFunction());
+
+        registerAffectorFunction(new DampingAffectorFunction());
+        registerAffectorFunction(new EnergyColorAffectorFunction());
+        registerAffectorFunction(new EnergyScaleAffectorFunction());
+        registerAffectorFunction(new ForceAffectorFunction());
+        registerAffectorFunction(new PointForceAffectorFunction());
+        registerAffectorFunction(new ForwardEulerAffectorFunction());
+        registerAffectorFunction(new TurbulenceAffectorFunction());
```
skaldarnar: Maybe go the full distance and don't have example affectors and generators in engine but only in a new ParticleTutorialModule. That would match our setup with other things like world gen. 
On the other hand, there might be some very basic affectors/generators that shoul be available per default(?)

engine/src/main/java/org/terasology/rendering/particles/internal/rendering/DisplayListParticleRenderer.java
```diff
@@ -0,0 +1,142 @@
...
+
+        glPushMatrix();
+        glTranslatef(-camera.x(), -camera.y(), -camera.z());
+
+        for (int i = 0; i < particlePool.livingParticles(); i++) {
+            final int i2 = i * 2;
+            final int i3 = i * 3;
+            final int i4 = i * 4;
+
+            material.setFloat3("position",
```
skaldarnar: > biiiig TODO

```
 _____         _       
|_   ____   __| | ___  
  | |/ _ \ / _` |/ _ \ 
  | | (_) | (_| | (_) |
  |_|\___/ \__,_|\___/ 
                       
```

PR #2635, opened 2016-12-03
add clear command for console

Avalancs:
Add "clear" command to console. Unit test included.


GooeyHub:

Issue #2629, opened 2016-11-27
Follow-up to #2625 - broken unit tests and entity behavior confirmation


Cervator:
#2625 "fixed" the cause for a lot of `NullPointerExceptions` around interaction screen bearing blocks (like the assembly table in JoshariasSurvival) and who knows what else in some multiplayer scenarios. However, it also broke a few unit tests and could lead to unexpected behavior in some other cases.

For now I merged it prematurely as I need to get Alpha 6 released already, particularly with Google Code-In starting tomorrow. But I want to get this some more attention in the near future.

We may not need to do anything other than update the unit tests to match the new behavior. But I did notice if you try to add logging to the affected block of code in `EntityAwareWorldProvider` *it triggers a whole lot*

```java
    @Override
    public EntityRef getBlockEntityAt(Vector3i blockPosition) {
        if (GameThread.isCurrentThread()) {
            EntityRef blockEntity = getExistingBlockEntityAt(blockPosition);
            if ((!blockEntity.exists() || !blockEntity.hasComponent(NetworkComponent.class)) && isBlockRelevant(blockPosition.x, blockPosition.y, blockPosition.z)) {
                logger.info("Found that an entity either didn't exist or it had no NetworkComponent. Entity {}", blockEntity);
                Block block = getBlock(blockPosition.x, blockPosition.y, blockPosition.z);
                blockEntity = createBlockEntity(blockPosition, block);
            }
            return blockEntity;
        }
        logger.error("Attempted to get block entity off-thread");
        return EntityRef.NULL;
    }
```

In fact it spammed the log enough to lock up my IntelliJ. Yet oddly the game seems able to handle it well enough to where it in theory fixed #2594. Taking the `NetworkComponent` check out of the condition still produced massive log spam in some conditions, but not enough to freeze my IDE.

I started the event server with a build containing this, running JoshariasSurvival, and no issues noted so far.

Getting the unit tests fixed ASAP is the first priority as all PRs until then will return results as unstable due to said unit tests.

Poking for potential interest: @portokaliu, @Josharias

Cervator:
Also wanted to throw in a link here to https://github.com/Terasology/JoshariasSurvival/issues/15 on the double `e` click issue, since it might not be in the right repo but was kinda-sorta related to the PR that spawned this follow-up.

Unsure if they relate functionally at this point.

Cervator:
[...]

PR #2625, opened 2016-11-26
Fixing JS

portokaliu:
This should fix  #2594 .
The issue was that the client told the server to create an entity he never created locally. For some reason (TBA), the created block doesn't exist for the host player but exists for the client player after it's seen in the world.

Seeing as any network component was created only if the entity the player was looking at (this as well may prove a problem and shouldn't be relied on, since currently the network component isn't created on block placement) doesn't exist, the client never considered it a networked entity, thus it had none of the proper components. As the systems relied on the block having these components, it caused a crash when trying to access null values.

As of this change, any block existence check will create the block if it doesn't have a network component. This might have the side-effect of creating network components for some blocks that don't need them, but I'm not entirely sure about that side of things. (still in testing :) )

How to test:
Place any machine block, or pretty much anything that should have a widget, as a client while the host isn't looking at said block. No crash == no problem.

Outstanding before merging:
- I've only tried it on my system with 2 players and I had 2 servers that worked pretty wonky (probably due to the 1.5gb limit. After increasing the limit I had no more problems)
- It might deserve further testing before fully merging. My personal ones were conclusive, but it's just one system

PR #2638, opened 2016-12-03
Engine fixes: code formatter, unit tests, create game screen, config

rzats:
### Contains

Small fixes for various issues mentioned in IRC and elsewhere:

- Set the Eclipse (IDEA-compatible) code formatter to indent with 4 spaces, which is what our indentation is, instead of 8.
- Ignored the known failing unit tests (see #2625 and #2629) for now so that they don't potentially confuse GCI students.
- Removed the save game path from the Create Game screen, as it would otherwise break the interface on resolutions lower than 1366x768:

a98ra0j

- Set the Film Grain setting to false in the default config - it's been widely reported as an annoyance.

### How to test

Fairly self-explanatory:

- IDEA should use 4-space indentation when set to the Terasology formatter.
- No unit tests should fail.
- The Create Game screen should work fine on 1366x768.
- A new game install should have film grain off by default.

No outstanding issues!


GooeyHub:

nihal111:
Looks good to me.
Tests out as stated :)

Cervator:
Thanks @rzats ! Especially for the unit test bit :-)

Issue #2601, opened 2016-11-11
Integration of entity-based LightComponents with Light Propagation / Voxel Lighting


emanuele3d:
Currently Terasology handle point-like lights in two separate ways, the output of which is combined via shaders.

The first method is somewhat more traditional (fully shader-based) and handles the lighting produced by entity with a LightComponent attached to them. An example of this is the held-in-hand torch of a player. 

The second method handles lights placed in the landscape, i.e. torches placed on the ground or attached to a tree. These lights are not entity-based but at are light-emitting blocks instead. These are handled via Voxel Lighting / Light Propagation, a method involving a cellular automata-like algorithm.

Ideally, we'd like to eliminate the first method and handle all or most lighting through the second method, probably making exceptions for purely directional lights i.e. the sun/moon. 

The rationale for this is that the Voxel Lighting / Light Propagation system produces nice visual features such as "around the corner" lighting and shadows with reasonable performance (1). Conversely, it would be quite difficult to have those features with the first method and they would be very computationally expensive. 

(1) See https://www.youtube.com/watch?v=OkPL3HvYqVM

OvermindDL1:
While still having a method to get the light level of a given block?

emanuele3d:
Dear @OvermindDL1, your question is obviously very clever but my smallmind just cannot make sense of it in the context of the description of this issue. Would you care to elaborate and give a chance to individuals with lesser intellects  to understand your concern?

portokaliu:
I think he might be referring to the method in "WorldProvider"
/**
     * Returns the light value at the given position.
     *
     * @param pos The position
     * @return The block value at the given position
     */
    byte getLight(Vector3f pos);

OvermindDL1:
> I think he might be referring to the method in "WorldProvider"

Precisely, yes, sorry.  Busy day.  :-)

> smallmind

What is that about?

emanuele3d:
smallmind: self-deprecating flunked humor - nevermind :smile: 

Thank you both for clarifying the concern.

If all lighting goes through the voxel lighting method per-block light levels should still be very easy to obtain: the per-block data will still be there but it will also take in account the input of dynamic lights, which right now I don't makes it into the per-block light levels.

No need to worry about imminent changes though: unless somebody picks up this issue I do not intend to work on it any time soon.

eviltak:
Do all the entities that use the first method use `LightComponent`? If not, what's the best way to identify these objects?

[...]

PR #2641, opened 2016-12-04
Refresh Render Task List only once every frame

### LINKED ISSUE #2610: Refactor Refreshing of Rendering TaskList
Currently the renderer goes through the list of nodes in the render graph and generates a task list. The task list is iterated every frame and the content of the .execute() method in each task object is what renders the 3D world on screen.

When something changes, i.e. a video setting, one or more  nodes can trigger the refresh of the task list. In some cases the same video setting change can prompt multiple nodes to trigger a task list refresh. Each time the the task list is refreshed, probably unnecessarily. 

Ideally nodes only _request_ a task list refresh so that a refresh is "pending". Then (and only then!) at the end of the frame or at the beginning of the next, if the pending flag is on the task list is actually refreshed.

For relevant classes, see org.terasology.rendering.world.WorldRendererImpl.initRenderGraph() and org.terasology.rendering.dag.AbstractNode.refreshTaskList().

eviltak:
Closes #2610. Shaves off a minimum of 1 ms from time taken for task list refreshes according to the logs before and after the change.

### How to test

In debug mode (`F3`), enable and disable wireframe mode (`F9`) to trigger task list refreshes (I don't know of any other method). The speedup isn't noticeable, but on comparison between the logs before and after the change, in the worst case a minimum of 1 ms is shaved off the task list refresh time.

### Details

 * Nodes now *request* refreshes (`RenderTaskListGenerator.requestRefresh()`). If requested, the task list will be refreshed in the next `preRenderUpdate`.
 * `WorldRendererImpl` stores a reference to `RenderTaskListGenerator` as a field. This has no performance implications since all nodes store the  same reference anyway.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
@@ -79,10 +79,9 @@ public abstract class AbstractNode implements Node {
         desiredStateResets.remove(stateChange.getDefaultInstance());
     }
 
-    // TODO: to be refactored - nodes should request a refresh but this should be pending until the end of the frame
     protected void refreshTaskList() {
         if (taskListGenerator != null) {
-            taskListGenerator.refresh();
+            taskListGenerator.requestRefresh();
         }
     }
 
--- a/engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
@@ -23,10 +23,12 @@ import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.assets.ResourceUrn;
+import org.terasology.registry.In;
 import org.terasology.rendering.assets.material.Material;
 import org.terasology.rendering.opengl.BaseFBOsManager;
 import org.terasology.rendering.opengl.FBO;
 import org.terasology.rendering.opengl.FBOConfig;
+import org.terasology.rendering.world.WorldRenderer;
 import org.terasology.utilities.Assets;
 
 /**
@@ -39,8 +41,10 @@ public abstract class AbstractNode implements Node {
     private Set<StateChange> desiredStateResets = Sets.newLinkedHashSet();
     private Map<ResourceUrn, BaseFBOsManager> fboUsages = Maps.newHashMap();
     private NodeTask task;
-    private RenderTaskListGenerator taskListGenerator; // TODO: investigate ways to remove nodes influence on taskList
     private boolean enabled = true;
+
+    @In
+    protected WorldRenderer worldRenderer;
 
     protected FBO requiresFBO(FBOConfig fboConfig, BaseFBOsManager frameBuffersManager) {
         ResourceUrn fboName = fboConfig.getName();
@@ -79,10 +83,8 @@ public abstract class AbstractNode implements Node {
         desiredStateResets.remove(stateChange.getDefaultInstance());
     }
 
-    protected void refreshTaskList() {
-        if (taskListGenerator != null) {
-            taskListGenerator.requestRefresh();
-        }
+    protected void requestTaskListRefresh() {
+        worldRenderer.requestRenderTaskListRefresh();
     }
 
     public Set<StateChange> getDesiredStateChanges() {
@@ -105,7 +107,6 @@ public abstract class AbstractNode implements Node {
     }
 
     public void setTaskListGenerator(RenderTaskListGenerator taskListGenerator) {
-        this.taskListGenerator = taskListGenerator;
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
@@ -23,12 +23,10 @@ import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.assets.ResourceUrn;
-import org.terasology.registry.In;
 import org.terasology.rendering.assets.material.Material;
 import org.terasology.rendering.opengl.BaseFBOsManager;
 import org.terasology.rendering.opengl.FBO;
 import org.terasology.rendering.opengl.FBOConfig;
-import org.terasology.rendering.world.WorldRenderer;
 import org.terasology.utilities.Assets;
 
 /**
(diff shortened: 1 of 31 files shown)


GooeyHub:

rzats:
@GooeyHub: ok to test

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -227,4 +228,25 @@ public void refresh() {
...
+     * Refreshes the task list if a refresh has been requested.
+     */
+    public void refreshIfNeeded() {
+        if (refreshRequested) {
+            refresh();
+
+            // Clear refresh requests to prevent refreshing unless needed
+            refreshRequested = false;
+        }
+    }
```
eviltak: That was actually my initial guess on how to do this, but won't the task list generator require a reference to the `WorldRendererImpl` now? I tried looking for existing ways that nodes could obtain a reference to the `WorldRenderer`, but couldn't find any.

eviltak: These changes break the `ConditionDependentNode` class, which now requires a `WorldRenderer` reference.

eviltak: Can be worked around by using `CoreRegistry.get()` or adding a `WorldRenderer` to `AbstractNode` but the first one is a hack and the second one is a bit too liberal, but is the best option. Going with the latter.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
emanuele3d: Thank you @eviltak for the changes. We are nearly there with this. 

I'd prefer that nodes that need a reference to the WorldRenderer get it through injection rather than have it by default, extending the AbstractNode class: there are nodes that do not need a reference to the WorldRenderer  (or won't need it after I refactored them). So I'd revert the changes you did to some nodes and have them look like this:

```
@In
private WorldRenderer worldRenderer
...
worldRenderer.requestTaskListRefresh()
```

Also, I'd like the method ```WorldRenderer.requestRenderTaskListRefresh()``` to be renamed to ```requestTaskListRefresh()```: the renderer has only one task list and the documentation will have explicit references to it as _the_ task list.

engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
```diff
@@ -79,10 +83,8 @@ protected void removeDesiredStateChange(StateChange stateChange) {
         desiredStateResets.remove(stateChange.getDefaultInstance());
     }
 
-    protected void refreshTaskList() {
-        if (taskListGenerator != null) {
-            taskListGenerator.requestRefresh();
-        }
+    protected void requestTaskListRefresh() {
+        worldRenderer.requestRenderTaskListRefresh();
```
emanuele3d: In line with other comment, I recommend to remove this method and have the `worldRenderer.requestRenderTaskListRefresh();` line right in the nodes that need it. Indeed there is some code repetition but it is minimal.

engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
```diff
@@ -105,7 +107,6 @@ public String toString() {
     }
 
     public void setTaskListGenerator(RenderTaskListGenerator taskListGenerator) {
```
emanuele3d: This method can be removed now! Yay for simplifying code!

engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
```diff
@@ -53,7 +53,7 @@ private boolean checkConditions() {
     public void propertyChange(PropertyChangeEvent evt) {
         boolean conditionsChanged = checkConditions();
         if (conditionsChanged) {
-            refreshTaskList(); // TODO: Think about `pending` mode for the nodes, for not refreshing task list more than once
+            requestTaskListRefresh(); // TODO: Think about `pending` mode for the nodes, for not refreshing task list more than once
```
emanuele3d: You can remove the comment now. 

On the other hand, while AbstractNode no longer needs a WorldRenderer reference ConditionDependentNode now certainly does. But its subclasses won't, so you can keep some as they are in this PR.

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -58,6 +58,11 @@
     void onChunkUnloaded(Vector3i chunkPos);
 
     /**
+     * Request a refresh of the render task list. The refresh takes place before the next frame.
+     */
+    void requestRenderTaskListRefresh();
```
emanuele3d: Please rename to requestTaskListRefresh(): there are no plans for additional task lists within the renderer and as such the term task list does not require disambiguation.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -359,6 +360,11 @@ public void onChunkUnloaded(Vector3i pos) {
     }
 
     @Override
+    public void requestRenderTaskListRefresh() {
+        requestedRenderTaskListRefresh = true;
```
emanuele3d: In line with comment above this variable can just be `requestedTaskListRefresh`.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/AbstractNode.java
emanuele3d: No matter how hard I try another pair of eyes will always spot things I miss. And that's what code reviews are for! :smiley:

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -47,6 +47,7 @@
     private static final Logger logger = LoggerFactory.getLogger(RenderTaskListGenerator.class);
     private List<RenderPipelineTask> taskList;
     private List<Node> nodeList;
+    private boolean refreshRequested;
```
emanuele3d: Whoops. I think we both missed this.

GooeyHub:

emanuele3d:
Just approved 879ee7ef and subsequently made one tiny change about a line that slipped through the cracks. @rzats, if Jenkins doesn't complain from my perspective this can be merged.

Good work @eviltak, much appreciated. I hope you'll stick around and take the chance to get your hands dirty with the rendering engine again. :wink:

rzats:
Merged and added @eviltak to [the credits](https://github.com/MovingBlocks/Terasology/blob/develop/docs/Credits.md) - thanks for your contribution!

eviltak:
@emanuele3d Fortunately, I do. #2601 is in my sights!

Issue #2557, opened 2016-10-21
Improve auto move support (autorun)


Cervator:
Edit: Initially implemented by #2647 - list below updated to cover improvements

Came up while thinking about crouching functionality (see http://forum.terasology.org/threads/crouch-and-prone-controls.1622/#post-14015) and seems to generally be an appreciated feature.

At present we have switch default walk/run ("change speed permanently") on caps lock, which isn't great. We don't have auto-move. Suggestion as noted in the forum:
- [x] Add auto-move functionality (bound to `r` by default)
  - [x] Needs an entry in the input settings + translation key
- [x] Move the walk/run toggle from caps lock to ALT-R (depends on #2551 - can leave on caps lock until that one is ready)
- [x] Consider alternatives for the movement to not break on jumping or strafing
- [x] Consider options to achieve full regular speed (auto-move is slower than a regular run, and toggling walk/run has no impact)

One thing to keep an eye out for: with auto move available you could start moving, open a UI, and keep moving. Can this cause trouble? Interaction screens tied to blocks should likely auto-close after a certain distance, while the player's own inventory screen could stay open.

zeokav:
Hi! I'd like to work on this. Where do I start implementing this idea?
Also, about the opening UI and continuing to move issue, can't we just drop the auto move state of the character when an interaction screen connected to a block is opened? I have used auto move in a game that I used to play and as far as I remember, opening a menu used to stop the movement.

rzats:
@zeokav: you can look into https://github.com/MovingBlocks/Terasology/pull/2544/files which is a PR in a fairly similar area 🙂 

You'll need to have a new `BindButtonEvent` bound to `r`, then add a new event to `LocalPlayerSystem` that will start periodically sending move forward events when the button is pressed (and stop when it's pressed again).

Cervator:
This feature was made functional in #2647 but I guess I'll leave this open for a bit for additional improvements so we don't lose them (comments in the closed PR aren't very visible.

Will update the checklist in the top comment.

zeokav:
Aye @Cervator ! I got back to trying to get this to work without having to give an ImpulseEvent every time. But I kinda need an explanation on why my approach doesn't work. Here's what I put in - 

```
@ReceiveEvent
    public void onAutoMove(PeriodicActionTriggeredEvent event, EntityRef entity) {
        if (event.getActionId().equals(AUTO_MOVE_ID)) {
            if (isAutoMove) {
[...]

PR #2647, opened 2016-12-07
Added auto move feature

zeokav:
### Contains

A PR to address #2557 . I have added a binding to enable auto run. 

### How to test

Open up a game and press R (default binding - can be changed from the inputs menu). Any other action will stop the auto run state. 

### Notes

More like questions before merging - 
- I have made impulse events fire up every 200 milliseconds. Will this have a negative impact on the performance? If I increase the deltas, the movement looks more jerky than how it does if I keep it at 200ms. 
- I would like some feedback though because using this method makes the movement looks slightly sluggish.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
@@ -56,6 +56,8 @@ import org.terasology.logic.characters.events.OnItemUseEvent;
 import org.terasology.logic.characters.events.SetMovementModeEvent;
 import org.terasology.logic.characters.interactions.InteractionUtil;
 import org.terasology.logic.debug.MovementDebugCommands;
+import org.terasology.logic.delay.DelayManager;
+import org.terasology.logic.delay.PeriodicActionTriggeredEvent;
 import org.terasology.logic.location.LocationComponent;
 import org.terasology.logic.notifications.NotificationMessageEvent;
 import org.terasology.logic.players.event.OnPlayerSpawnedEvent;
@@ -101,6 +103,8 @@ public class LocalPlayerSystem extends BaseComponentSystem implements UpdateSubs
     @In
     private PhysicsEngine physics;
     @In
+    private DelayManager delayManager;
+    @In
     NetworkSystem networkSystem;
 
     @In
@@ -123,13 +127,13 @@ public class LocalPlayerSystem extends BaseComponentSystem implements UpdateSubs
     @In
     private Time time;
 
-    private long lastItemUse;
-
     private BlockOverlayRenderer aabbRenderer = new AABBRenderer(AABB.createEmpty());
 
     private int inputSequenceNumber = 1;
 
     private AABB aabb;
+
+    private static final String AUTO_MOVE_ID = "AUTO_MOVE";
 
     public void setPlayerCamera(Camera camera) {
         playerCamera = camera;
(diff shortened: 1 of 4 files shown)


GooeyHub:

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -212,6 +215,33 @@ private void standPlayer(EntityRef entity) {
...
+    }
+
+    /**
+     * Start a thread that moves the entity in the camera direction at every period of 0.2 seconds.
+     * @param entity
+     */
+    private void startAutoMove(EntityRef entity) {
+        isAutoMove = true;
+        ClientComponent clientComponent = entity.getComponent(ClientComponent.class);
+        new Thread(() -> {
```
skaldarnar: Imho this should not be done with a new Thread, but with the help of `DelayManager`. You can schedule events there, and cancel the periodic events if auto-move should stop.

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -212,6 +215,33 @@ private void standPlayer(EntityRef entity) {
...
+            while(isAutoMove) {
+                try {
+                    Vector3f viewDir = entity.getComponent(LocationComponent.class).getWorldDirection();
+                    clientComponent.character.send(new CharacterImpulseEvent(viewDir.mul(8f)));
```
skaldarnar: The `CharacterImpulseEvent` likely makes the movement feel _sluggish_. The impulse will actually push the player forward, instigated by some external force, not actually the player moving by himself.

zeokav: @skaldarnar : What can I use to make it feel smooth? I was looking for something that could simulate a key press event with the Forwards key pressed at all times but that _drops_ inputs because of too many inputs from the entity.

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
zeokav: Not an issue. I'll fix this bit! 😄

zeokav: @skaldarnar I looked up DelayManager but I could not find any implementation of it. What should my action Id be? And where exactly am I supposed to define the action that has to happen periodically?

oniatus: @zeokav I have some usage of DelayManager in [one of my repos](https://github.com/oniatus/UniqueServerNames/blob/master/src/main/java/org/terasology/uniqueservernames/system/ServerUniqueNameSystem.java).
Check out Line 40 (get an instance), Line 78 (register delayed action) and Line 82F (handle the event). I think there are two different events for delayed and periodic action, so if you need the other one, you may want to inspect the source of delaymanager directly :)

zeokav: Thanks @oniatus! Some small changes and I got it working. 👍

zeokav:
@skaldarnar I have made the change you asked for (Using delayed manager instead of making a new thread). :)

skaldarnar:
@zeokav merged! 

What do you think about staying in automove mode when jumping? That way you could travel long distances in non-flat worlds as well. I still have to look into the key press event vs impulse event.

zeokav:
@skaldarnar I was thinking of that too since the world is more multi-levelled than flat. Just have to change a line or two for that anyway. I'll club it along with my next PR. 😄

PR #2652, opened 2016-12-09
Add ping to server browser

iojw:
### Contains

- Server browser now shows ping in addition to other server information. 

It didn't make sense to bind this to the UIText like the other fields so this information is retrieved from inside the JoinGameScreen class whenever the user selects a server or clicks refresh.

~~Player overlay [tab] now shows the ping of other players in multiplayer.~~ 

~~To implement this, I went with adding a PingComponent and a PingAuthoritySystem which constantly updates the PingComponents of all connected players in a multiplayer server. The overlay screen then displays this information. I was unable to test this since the system needs to be on the server to receive the onConnected event so I'm not entirely sure if this work~~
 
There seem to have been some auto-formatting changes by IntelliJ as well.
### How to Test

Open the Join Game screen. Verify that the ping is shown for each server.

capture

Do let me know what areas of the code can be improved on 🙂.


GooeyHub:

portokaliu:
I feel the need to mention https://github.com/MovingBlocks/Terasology/commit/d964e4c961afb53643b1e03bd763bfd369d56e57 that was related to issue https://github.com/MovingBlocks/Terasology/pull/2602

iojw:
@portokaliu Yup, the code used to retrieve the ping is from #2602. Should probably have mentioned that in the PR 😐.

iojw:
There currently seems to be a problem with the ping showing in the player overlay due to one of the events not firing. I will try my best to fix it, so this is probably not ready for a review yet... 😐

Cervator:
No problem @iojw take your time - not like we're finding a lot of chances to review and merge engine PRs these days, so busy in module land! Thank you so much for getting this going in any case :-)

iojw:
I was initially working on storing each player's client-to-server ping locally using the ping code available, then allowing other clients connected in the same server to retrieve this ping. 

However, after some discussion on IRC with oniatus and some research, it seems the better solution to this would be to do it server-side instead, where the server would time the round-trip time for packets to the client and then making this information available to all clients through some sort of subscriber system or event. I tried taking a look at the network code, and it seems this might require some sizeable architectural changes. 

I'm not very experienced with networking and I'm afraid that I might do more harm that good to the architecture. Therefore, I'm reducing the scope of this PR to only include the changes to the Join Game Screen, leaving the main issue open for someone more experienced to tackle. The relevant classes are probably `NetClient` and `NetworkSystemImpl`.

GooeyHub:

oniatus:
I think it's okay to split both features (ping in server browser and ping overview for all players) in two PRs.

msteiger:
I believe that the ping-server functionality should be re-factored in a new class and references from the two mentioned points (UI and game command).

It would be great, if the code could be adjusted to just read the time instead of parsing the `toString()` text.

iojw:
That does seem like a better and cleaner approach. Will leave it to someone who is more experienced with all the networking-related stuff. Should I just go ahead and close this then? 🙂

msteiger:
You can leave it here so we don't forget. I'll look into this as soon as I can.

iojw:
Closing in favour of #2768.

Cervator:
No need to close! The other PR contains your commits from here so this should auto-close as merged when the other one is merged. That way both get the shiny indicator of something being merged :D

GooeyHub:

PR #2660, opened 2016-12-11
Add Snowball block

mjuvekar7:
Add snowball block (a block with the pure snow texture on all sides).


GooeyHub:

PR #2661, opened 2016-12-12
Fix moon/sun cycles and nights being totally dark

Avalancs:
### Contains
Hopefully the position of the sun/moon and the light levels respective to the time of day now works correctly. Also now there is some dimmed light during the night, so the world is no longer totally black, see attached image.
fullmoon
Unfortunately the direction of shadows are sometimes in the wrong direction, and there are random groups of blocks which do not get illuminated in the night.

### How to test

Use the console with the following commands:
setWorldTime 0.0 -- Moon should be above, fairly lit night
setWorldTime 0.25 -- Moon and Sun not visible, almost total darkness
setWorldTime 0.5 -- Sun should be above, maximum visibility

OR if you use setTimeDilation 25.0 then you can watch as the moon/sun rises and sets.

- [x] Sun and Moon cycle based on time of day
- [x] No longer totally dark during the night
- [ ] Simplify calculations
- [ ] Colored light for the moon?
- [ ] Weird shadow directions/patches

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/backdrop/Skysphere.java
+++ b/engine/src/main/java/org/terasology/rendering/backdrop/Skysphere.java
@@ -101,16 +101,30 @@ public class Skysphere implements BackdropProvider, BackdropRenderer {
         return celSystem.getSunPosAngle();
     }
 
+    /**
+     * Get the strength of the light for the current day part, based on sun/moon visibility.
+     * @return the strength of the current light in [0;1] range.
+     */
     @Override
     public float getDaylight() {
-        float angle = (float) Math.toDegrees(TeraMath.clamp(Math.cos(getSunPositionAngle())));
-        float daylight = 1.0f;
+        float lightBasedOnTimeOfDay = 1.0f;
+        final float ambientLight = 0.2f; // not totally black when the sun goes down, but the moon is not yet up
+        double radSunAngle = getSunPositionAngle(); // expected [-PI;PI] for the day cycle to work correctly!
 
-        if (angle < 24.0f) {
-            daylight = 1.0f - (24.0f - angle) / 24.0f;
+        // if the degree of the sun/moon is less than "lessLightUnderAngle", then we gradually increase the light, instead of full intensity
+        final double lessLightUnderAngle = Math.sin(Math.toRadians(15.0));
+        double absSinSunAngle = Math.abs(Math.sin(radSunAngle));
+        if (absSinSunAngle < lessLightUnderAngle) {
+            lightBasedOnTimeOfDay = (float)(absSinSunAngle / lessLightUnderAngle);
+        }
+        // otherwise the light remains full intensity
+
+        // if the moon is out, we want to reduce the brightness
+        if(radSunAngle < 0.0) {
+            lightBasedOnTimeOfDay *= 0.2f;
         }
 
-        return daylight;
+        return ambientLight + (lightBasedOnTimeOfDay*(1.0f-ambientLight));
     }
 
     @Override
@@ -126,7 +140,7 @@ public class Skysphere implements BackdropProvider, BackdropRenderer {
     @Override
     public Vector3f getSunDirection(boolean moonlightFlip) {
         float sunAngle = getSunPositionAngle() + 0.0001f;
-        Vector3f sunDirection = new Vector3f(0.0f, (float) Math.cos(sunAngle), (float) Math.sin(sunAngle));
+        Vector3f sunDirection = new Vector3f(0.0f, (float) Math.sin(sunAngle), (float) Math.cos(sunAngle));
 
         // Moonlight flip
         if (moonlightFlip && sunDirection.y < 0.0f) {
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -537,18 +537,18 @@ public final class WorldRendererImpl implements WorldRenderer {
 
     @Override
     public String getMetrics() {
-        String stringToReturn = "";
-        stringToReturn += renderableWorld.getMetrics();
-        stringToReturn += "Empty Mesh Chunks: ";
-        stringToReturn += statChunkMeshEmpty;
-        stringToReturn += "\n";
-        stringToReturn += "Unready Chunks: ";
-        stringToReturn += statChunkNotReady;
-        stringToReturn += "\n";
-        stringToReturn += "Rendered Triangles: ";
-        stringToReturn += statRenderedTriangles;
-        stringToReturn += "\n";
-        return stringToReturn;
+        StringBuilder stringToReturn = new StringBuilder("");
+        stringToReturn.append(renderableWorld.getMetrics());
+        stringToReturn.append("Empty Mesh Chunks: ");
+        stringToReturn.append(statChunkMeshEmpty);
+        stringToReturn.append("\n");
+        stringToReturn.append("Unready Chunks: ");
+        stringToReturn.append(statChunkNotReady);
+        stringToReturn.append("\n");
+        stringToReturn.append("Rendered Triangles: ");
+        stringToReturn.append(statRenderedTriangles);
+        stringToReturn.append("\n");
+        return stringToReturn.toString();
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/backdrop/Skysphere.java
```diff
@@ -101,16 +101,30 @@ public float getSunPositionAngle() {
...
 
-        if (angle < 24.0f) {
-            daylight = 1.0f - (24.0f - angle) / 24.0f;
+        // if the moon is out, we want to reduce the brightness
+        if(radSunAngle < 0.0) {
+            lightBasedOnTimeOfDay *= 0.2f;
         }
 
-        return daylight;
+        return ambientLight + (lightBasedOnTimeOfDay*(1.0f-ambientLight));
```
emanuele3d: In light of what I wrote above I'd rewrite the whole method as an if/elseif/elseif/else statement dependent on the main light angle. Also, I'd use linear interpolation rather than sin and cos to do the light intensity math. It would make sense to use trigonometry with a more realistic lighting model (i.e. taking in account atmospheric thickness) but in this case we just need to interpolate light intensities between 15/5 degrees and viceversa: I am 99% certain linear interpolation won't make a difference there.

engine/src/main/java/org/terasology/world/sun/BasicCelestialModel.java
```diff
@@ -29,9 +29,21 @@
...
 
+    /**
+     * Return the direction of the sun based on the time of the day.<br />
+     * At days=0 midnight the direction of the sun will be -90 degrees, while at
+     * days=0.5 noon it will be +90 degrees.
+     * @param days the time of day
+     * @return the direction of the sun in radians in the range [-PI;PI]
+     */
     @Override
     public float getSunPosAngle(float days) {
```
emanuele3d: Sorry, we need to rename this method to getMainLightElevation as elevation is what it returns. Also, we need to rewrite the javadoc in terms of both sun and moon as the MainLight implements both. Talking only in terms of sun would confuse a reader unfamiliar with the simplified celestial model Terasology uses.

engine/src/main/java/org/terasology/world/sun/BasicCelestialModel.java
```diff
@@ -29,9 +29,21 @@
...
-        return (float) (days * 2.0 * Math.PI - Math.PI);  // offset by 180 deg.;
+        double sunRad = (days%1.0) * 2.0 * Math.PI; // [0;2PI]
+        double shiftedBy90 = sunRad - Math.PI/2.0; // shift by -90 so midnight means moon is on top
+        if(shiftedBy90 > Math.PI) { // expected output is [-PI;PI]
+            return (float)(shiftedBy90 - 2.0*Math.PI);
+        }
+        return (float)shiftedBy90;
```
emanuele3d: Right now the math is a little hard to read. I'd refactor this method as an if/else block dependent on `days % 1.0`: if time is between 0.25 and 0.75 (daytime) you calculate and return a value, if it isn't you return another. This way it's easier to understand the differences in the calculation depending on daytime/nighttime.

emanuele3d:
@Avalancs : thank you for your efforts. I have given you some feedback that should help you to simplify the code and the calculations. 

Shadows are important. We cannot merge this if the shadows are not working properly. Shadow generation is done in the ShadowMapNode. Please check the code there to see what it expects and modify it as needed, to be compatible with the new handling. 

Regarding moonlight color: that's a tricky visual/perceptual issue. 

Moon soil is grey/silvery: effectively color-neutral. As such it gets the color tinge of the sun, which is a yellowish star, but only astronomically speaking: for our eyes it is pretty much white at high elevations and yellowish/orangish when it has to go through a lot of atmosphere.

If you look at the landscape at night its colors feels... "different", and hard to describe. That's because at night you are using light-sensing cells (the rods) that do not make much distinction between different colors: they work with whatever light they can get (unlike the cones which are red-green-blue specialized). That's why the nocturnal landscape is a dark, grainy, considerably desaturated and perhaps with a slight green tinge as our increased sensitivity in that part of the spectrum probably gets more noticeable in low light conditions relative to our red/blue sensitivity. 

So, natural night vision is a tricky topic. Ideally this would be handled via a look-up table used by a tone-mapping filter. And there is already a tone-mapping node in use but I would suggest this aspect could be left to a separate PR. You might want to experiment with almost imperceptible greenish/bluish tinges if you like, but I'm skeptical it will get you far: our eyes are amazingly good and it's just very tricky to trick them!!! :smile:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -537,18 +537,18 @@ public float getBlockLightIntensityAt(Vector3f position) {
...
-        stringToReturn += statChunkMeshEmpty;
-        stringToReturn += "\n";
-        stringToReturn += "Unready Chunks: ";
-        stringToReturn += statChunkNotReady;
-        stringToReturn += "\n";
-        stringToReturn += "Rendered Triangles: ";
-        stringToReturn += statRenderedTriangles;
-        stringToReturn += "\n";
-        return stringToReturn;
+        StringBuilder stringToReturn = new StringBuilder("");
```
Avalancs: No, I'm using Intellij... I was using a rule of thumb that if you have to concatenate many variables into a single string, a builder is more efficient.

engine/src/main/java/org/terasology/rendering/backdrop/Skysphere.java
```diff
@@ -101,16 +101,30 @@ public float getSunPositionAngle() {
...
     @Override
     public float getDaylight() {
-        float angle = (float) Math.toDegrees(TeraMath.clamp(Math.cos(getSunPositionAngle())));
-        float daylight = 1.0f;
+        float lightBasedOnTimeOfDay = 1.0f;
+        final float ambientLight = 0.2f; // not totally black when the sun goes down, but the moon is not yet up
+        double radSunAngle = getSunPositionAngle(); // expected [-PI;PI] for the day cycle to work correctly!
+
+        // if the degree of the sun/moon is less than "lessLightUnderAngle", then we gradually increase the light, instead of full intensity
+        final double lessLightUnderAngle = Math.sin(Math.toRadians(15.0));
```
Avalancs: In a few lines later, I check the absolute value of the sine compared to this, so that means fullIntensityThresholdInRadians == fullDarknessThresholdInRadians for now. I will experiment if using different thresholds for the sun and the moon would make it more pleasing overall (sun sets slower, but it takes more time for the moon to get to full intensity?

engine/src/main/java/org/terasology/world/sun/BasicCelestialModel.java
Avalancs: I totally agree!

Avalancs:
I will look into the shadows, and do the changes you recommended.

Natural night vision sounds very hard to implement, but we don't have to go that far for it to be fun. I really liked how in Splinter Cell Conviction, when you were in cover in the dark, the dark areas appeared in a black and white fashion, while normally lit areas had normal color. It would be fun to have it as an option at least, when the shader pipeline is ready for it.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: I'm not going to fight over this but this is a good reference:

http://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

engine/src/main/java/org/terasology/rendering/backdrop/Skysphere.java
emanuele3d: @Avalancs: I don't know if an asymmetrical ramping up/down of intensity is going to make a significant difference. If I remember correctly MC has a 7 minutes night and a 13 minutes day, but I don't remember a defined transition between them and the light intensity never really goes down too much. In Terasology at this stage we can do whatever we want and it would be interesting from a gameplay perspective to have at least a few minutes of complete darkness. It's an easy way to be different from MC.

@skaldarnar: My apologies good Sir. I shall go to my cell and repent by reading for 100 times the CheckStyle rules files out loud.

emanuele3d:
Interesting approach the one used in Splinter Cell. We'd have to use the stencil buffer to mark areas to be desaturated to black and white, leaving other areas untouched by the desaturation filter. I plan to do something along that line for the above/underwater vision. Sometimes the camera is right at the level of the water surface but the underwater effect kicks in and affects the whole frame, including above-water portions of it. Perhaps we could make that effect more generic and useful in both circumstances.

Avalancs:
I think I found the problem: ShadowMapNode:getQuantizedMainLightDirection() also calculates the direction of the sun/moon, and I did not change the sine/cosine there like in SkySphere. I will be tinkering a bit more before I submit my changes.

Avalancs:
This needs more work, maybe re-thought, and might already be obsolete with the node refactors, so I'm going to close it, and start a new PR when I have something.

Issue #2376, opened 2016-06-27
Add toggle / option to improve visibility of chat


Cervator:
As suggested last multiplayer test session: the chat window (either in its full version or even the mini single line version) could have its visibility improved. At present it'll tend to be suppressed (even the one-line mini chat) when any other UI is up, and there's no way to leave it persistently visible (probably should be a toggle of some sort)

Probably not terribly difficult when used to our UI a little bit, although I'm not sure exactly how the ideal design would look / work. Open for ideas! :-)
##

iojw:
@Cervator: Interesting issue 🙂. Not sure if I understood this issue correctly, but are you referring to how the chat is invisible unless you are typing something? From my tests alone, I did notice that you only see the chat history when you are typing in chat, which could annoying for some.

If that's the case, I think an possible idea to consider would be the leave the chat visible after anyone types something in chat for a period of time before fading away (maybe around 10 seconds). I managed to find an example of this in TF2 here - [https://youtu.be/VSKGX6yBWgQ?t=52 ](https://youtu.be/VSKGX6yBWgQ?t=52 )- you just need to watch it for a few seconds. We could also add a toggle button to toggle the visibility as you said if some prefer to hide the chat.

I'm not sure what you mean by the mini single-line chat though. How do you make it show up?

Cervator:
That is exactly it yes. The window is only visible when you are actively interacting with it *or* briefly (as the mini single line chat) when somebody else sends a message. But that fades pretty quick and is only one line of text :-)

iojw:
@Cervator : Hmm, would it then be better to keep the current behaviour and add a button to toggle the chat history or make the chat history fade out after a certain amount of inactivity (eg: https://youtu.be/VSKGX6yBWgQ?t=15)? Also, it seems the fade out times for the mini chat are hardcoded integers so it should be simple to change those values.

iojw:
@Cervator: I attempted to address this issue in #2667 which got merged a while back, adding a few tiny features to bring the chat more in line with other games. 

I'm wondering if there are any additional features you guys would like to see on top of the changes already made or any other thoughts you might have.

As @oniatus mentioned, there's currently an issue whereby too much text might overrun the screen on the mini overlay, since it shows up to the last 6 messages with no char limit on individual messages. I'm leaning towards a line limit for the mini overlay to tackle this since I feel that the mini overlay is something that the player should not have to interact much with. I could fix this together with any new features you guys would like to see in the chat.

[...]

PR #2667, opened 2016-12-14
Chat UI tweaks

iojw:
Simple fix for #2376 

### Contains

- Chat overlay now shows up to last 6 messages
- Sending a message now closes the chat window and opens the chat overlay instead of remaining focused on the text field
- Tweaked base time values of chat overlay so it lasts longer before fading out

### How to test

Open up a game and type something in chat. Here's an image of the updated chat overlay:

Some ideas I had but not really sure if they should be put in the game:
- Settings for max no. of lines to be shown on overlay
- Similar to other games, where the chat history is separate from the text box and pressing the chat button just makes the text field visible and turns the chat history scrollable (my concern with this is that it might take up too much space since the new messages appear at the bottom but we could change this such that new messages appear on top)

I adjusted the chat system based on my own judgement and experience with other games so definitely let me know if you would like any changes and I can add them in 🙂.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/ui/MiniChatOverlay.java
+++ b/engine/src/main/java/org/terasology/logic/console/ui/MiniChatOverlay.java
@@ -24,9 +24,10 @@ import org.terasology.rendering.nui.CoreScreenLayer;
 import org.terasology.rendering.nui.databinding.ReadOnlyBinding;
 import org.terasology.rendering.nui.widgets.UILabel;
 
+import java.util.ArrayList;
+
 /**
  * The miniaturized chat console widget
- *
  */
 public class MiniChatOverlay extends CoreScreenLayer {
 
@@ -35,9 +36,11 @@ public class MiniChatOverlay extends CoreScreenLayer {
      */
     private static final float TIME_VISIBLE_PER_CHAR = 0.08f;
 
-    private static final float TIME_VISIBLE_BASE = 1.0f;
+    private static final float TIME_VISIBLE_BASE = 5.0f;
 
     private static final float TIME_FADE = 0.3f;
+
+    private static final int MAX_LINES = 6;
 
     private enum State {
         FADE_IN,
@@ -62,13 +65,21 @@ public class MiniChatOverlay extends CoreScreenLayer {
             @Override
             public String get() {
                 Iterable<Message> msgs = console.getMessages(CoreMessageType.CHAT, CoreMessageType.NOTIFICATION);
-                String last = "";
+                ArrayList<Message> msgsList = new ArrayList<>();
+                StringBuilder messageHistory = new StringBuilder();
 
                 for (Message msg : msgs) {
-                    last = msg.getMessage();
+                    msgsList.add(msg);
                 }
 
-                return last;
+                //checks if number of messages > MAX_LINES, else starts from 0
+                for (int i = msgsList.size() >= MAX_LINES ? msgsList.size() - MAX_LINES : 0; i < msgsList.size(); i++) {
+                    messageHistory.append(msgsList.get(i).getMessage());
+                    if (i < msgsList.size() - 1) {
+                        messageHistory.append(Console.NEW_LINE);
+                    }
+                }
+                return messageHistory.toString();
             }
         });
     }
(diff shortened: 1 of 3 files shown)


GooeyHub:

engine/src/main/java/org/terasology/logic/console/ui/MiniChatOverlay.java
```diff
@@ -62,13 +65,21 @@ public void initialise() {
...
 
-                return last;
+                //checks if number of messages > MAX_LINES, else starts from 0
+                for (int i = msgsList.size() >= MAX_LINES ? msgsList.size() - MAX_LINES : 0; i < msgsList.size(); i++) {
+                    messageHistory.append(msgsList.get(i).getMessage());
+                    if (i < msgsList.size() - 1) {
+                        messageHistory.append(Console.NEW_LINE);
+                    }
+                }
+                return messageHistory.toString();
```
oniatus: Not sure what is on the classpath, but this looks like a good use for [EvictingQueue](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/EvictingQueue.html) or [CircularFifoQueue](http://commons.apache.org/proper/commons-collections/javadocs/api-release/org/apache/commons/collections4/queue/CircularFifoQueue.html).
Or even better: Build the string while iterating over the messageList, if you use the counter anyway.

iojw: What do you mean by "Build the string while iterating over the messageList"? Do you mean using Iterables.size() and a counter to append the last 6 messages?

oniatus: That sounds preferable to copying the entire buffer first and pruning the copy to the last n messages afterwards.

GooeyHub:

engine/src/main/java/org/terasology/logic/console/ui/MiniChatOverlay.java
oniatus: Checked out and worked well so far.
Small experiment on readability (not sure what to prefer, I would leave this up to you):
```
        message.bindText(new ReadOnlyBinding<String>() {
            @Override
            public String get() {
                Iterable<Message> messages = console.getMessages(CoreMessageType.CHAT, CoreMessageType.NOTIFICATION);
                int messagesCount = Iterables.size(messages);
                int messagesToSkip = messagesCount - MAX_LINES;
                if (messagesToSkip > 0) {
                    messages = Iterables.skip(messages, messagesToSkip);
                }

                Stream<Message> mostRecentMessages = StreamSupport.stream(messages.spliterator(), false);
                String messagesHistory = mostRecentMessages.map(Message::getMessage).collect(Collectors.joining(Console.NEW_LINE));
                return messagesHistory;
            }
        });
```

In any case, `MAX_LINES` should probably be `MAX_MESSAGES` or `MAX_VISIBLE_MESSAGES`. Fun fact: The overlay can get really huge, when large messages wrap over multiple lines (tested with large `BANANA...AS` 👓 . However adding a scrollpane or a real line limitiation/char is likely out of scope for this PR, so this is fine from my side :)

oniatus:
Whoops, put my [last comment](https://github.com/MovingBlocks/Terasology/pull/2667/files/b447ec38829b0710b9fceb3b1464a4c0f7a6ead2#r92462459) in the first commit :)

Just give me a short ping if you want to do some more stuff on the code, otherwhise I think we can merge this one tomorrow :)

iojw:
@oniatus: Ah, didn't think about that. Would like to deal with it before it gets merged. I don't think we should add a scrollbar since that's already present in the full chat history or a char limit since that might lead to one message filling the whole overlay.. It does seem strange that there's no limit for chat messages - what do you about limiting all messages sent to 100 chars? 🙂

oniatus:
@iojw I think the overlay has to deal with the message size, not vice versa :) If one wants to copy-paste a novel or a large link, why not.
But as the issue was about improving the visibility and adding a fade out, this is okay for the moment.
We can add a follow-up issue about a maximum size or more polishing like different or configurable fade out times if you want to continue on this 👍

portokaliu:
The only thing I can think about regarding this is having a fade to a larger text area (with previously buffered messages) that can only be seen when trying to type stuff. Probably, as you said, out of the current scope though

iojw:
@oniatus Cool, I guess you can just merge this in then 🙂

@portokaliu That should be what what happens currently - this overlay is only shown when the player is not actively typing and someone else sends a message. When the player is actively typing, there's a larger scrollable view visible with the full chat history.

GooeyHub:

PR #2670, opened 2016-12-16
Fix auto move cancel action when not triggered

oniatus:
### Contains

Short follow up to #2557.
The current implementation will cancel the periodic action each time **any** periodic action is triggered.
The PeriodicActionComponent is not null safe on removal (note the null -> long autoboxing), which causes periodic NPEs.

### How to test

Enable any module that triggers periodic action, like the portals module (place one portal block in the world, then watch the log).


GooeyHub:

PR #2672, opened 2016-12-16
Fix #2605 Create new client side entity for wielded item

### LINKED ISSUE #2605: Host Items at (0,0,0)
What you were trying to do:
I was investigating physics on a client.

What actually happened:
While working on said physics (and items going through the ground), I saw an entity really low in the ground (with the client). After getting there, I saw it was an axe that I tried to collect and couldn't. I changed the item on the host to a torch to also go down and investigate, and now the client saw a torch (host doesn't see anything)

How to reproduce:
1. Join a local server (not headless) with a client. 
2. Go to (0,0,0). 
3. Observe the host wielded item.

Details:
I only had Core+CheatsForAll active

Tropid:
### Contains

Fixes #2605

This fixes 2605 by creating a new entity (client-side) for the wielded items.

This pull request creates a client side entity for first person view display instead of reusing the entity from the inventory. This makes reasoning easier and fixes the bug where the host held item spawns at (0, 0, 0) for the clients.

### How to test

Create a listen server, join with a new client. Pick up an item as host and see how the client will not see the host wielded item at (0, 0, 0). Also dropping items and placing blocks should still work.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterHeldItemAuthoritySystem.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterHeldItemAuthoritySystem.java
@@ -28,7 +28,6 @@ public class CharacterHeldItemAuthoritySystem extends BaseComponentSystem {
     @ReceiveEvent
     public void onChangeHeldItemRequest(ChangeHeldItemRequest event, EntityRef character,
                                         CharacterHeldItemComponent characterHeldItemComponent) {
-        EntityRef oldItem = characterHeldItemComponent.selectedItem;
         characterHeldItemComponent.selectedItem = event.getItem();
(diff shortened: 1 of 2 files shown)


GooeyHub:

rzats:
@GooeyHub: ok to test

Sorry for the late response - things here are a little hectic because of Google Code-In. I'll review this within a few hours 🙂

GooeyHub:

flo:
Sorry that this one got forgotten, tested and merged it.

Cervator:
For reference: reverted in #2806 due to other issues. Fix might still be viable but needs some additional work and testing before being put back in :-)

Removed from the Alpha 7 milestone and re-opened #2605

PR #2673, opened 2016-12-17
Remove dependency on CoreRegistry from Skysphere

Avalancs:
Remove dependency on CoreRegistry from Skysphere by passing in the CelestialSystem directly. 
I also added it to the HeadlessEnvironment, with a mock to WorldProvider, and a static WorldTime reference, so unit testing the day/night cycle in the future becomes possible

### How to test
* Run all the unit tests and verify nothing got broken by this change.
* Open up the game, and create a world
* Open up the game, and load a world

### DIFF
--- a/engine-tests/src/main/java/org/terasology/Environment.java
+++ b/engine-tests/src/main/java/org/terasology/Environment.java
@@ -80,6 +80,10 @@ public class Environment {
 
         setupComponentManager();
 
+        setupWorldProvider();
+
+        setupCelestialSystem();
+
         loadPrefabs();
     }
 
@@ -138,6 +142,14 @@ public class Environment {
     }
 
     protected void setupStorageManager() throws IOException {
+        // empty
+    }
+
+    protected void setupWorldProvider() {
+        // empty
+    }
+
+    protected void setupCelestialSystem() {
         // empty
     }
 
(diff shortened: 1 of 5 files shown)


GooeyHub:

emanuele3d:
In the description of this PR you correctly mention what you did. Another aspect that would be interesting to know however, is the why. What is the reasoning that led you to propose this set of changes?

flo:
Any CoreRegistry usage removal is good.

Using the Ciregegistry is like using a static variable and comes with the same restrictions. E.g currently we can't run server and client in same JVM

emanuele3d:
Fair enough, but then why not to use the Context instead? 

I guess my question is about the advantage of passing the CelestialSystem (which couples the SkySphere to it) rather than Skysphere implementations obtaining it from a Context only if they need it, as there could be game modes that do not need a skysphere (i.e. undeground-only) or need a Skysphere that does not use a CelestialSystem, i.e. creative-mode daytime-only modes.

Avalancs:
@emanuele3d It makes testing more easier for my work in progress sun/moon cycles. I won't need a HeadlessEnvironment, or to instantiate a Context just to initialize a Skysphere.

emanuele3d:
Your reasoning did not quite convince me: you are suggesting that making it easy for your tests is more important than the game modes I suggested. 

But never mind. Looking at the context of this code it seems to be pretty inflexible anyway. It's not like a module can currently provide a different SkySphere implementation if it had one.

Avalancs:
Sorry, I had to re-read your comment, and now I get it! For those cases it makes more sense to have it with the Context, I'll change this.

GooeyHub:

Cervator:
Since this seems like it has been worked out in a good way while making progress on CoreRegistry -> Context I've gone ahead and merged it after testing :-)

emanuele3d:
Thank you @Avalancs for the last minute changes and thank you @Cervator for the merge!

PR #2678, opened 2016-12-18
Use multiple stacks to give blocks. This closes #2607

### LINKED ISSUE #2607: give more than 127 blocks in one command (use multiple stacks?)
What you were trying to do:
Give myself more than 99 blocks with the giveBlock command with a single command.

What actually happened
The sound of collecting items played, no error was returned and the 999 blocks I asked for were nowhere in sight.

How to reproduce:
* Open console and type "give stone 999"
* Observe no stone has been given.

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/39261328-giveblock-multiple-stacks?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5279&utm_medium=issues&utm_source=github).

J0hnCena:
### Contains
Makes the giveBlock command send multiple times, until quantity runs out. The command is limited to 99 blocks per stack, since that is the max stack size. Fixes #2607. 

### How to test

Press (F1) to bring up the console. Use the giveBlock command by typing give <item> <quantity>. Pull up the inventory with (i) to see if none of the stack sizes are above 99 and if the correct number of blocks were given.

Sorry I had to reopen this again. I accidentally closed it when I force pushed. To answer @portokaliu 's question, unless I change stackCount in ItemComponent to an integer or create another component to contain the number of blocks to give, then I can't use InventoryAuthoritySystem to send blocks more than 127. If there's another way to do it, then please let me know. I apologize for the late response; I've been a bit busy with finals.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
@@ -327,17 +327,21 @@ public class BlockCommands extends BaseComponentSystem {
         if (quantity < 1) {
             return "Here, have these zero (0) blocks just like you wanted";
         }
+        //continue giving blocks until there are no more blocks to give
+        //TODO reference maxStackSize instead of explicitely subtracting 99
+        for(int quantityLeft = quantity; quantityLeft > 0; quantityLeft -= 99) {
+            EntityRef item = blockItemFactory.newInstance(blockFamily, quantityLeft > 99 ? 99 : quantityLeft);
+            if (!item.exists()) {
+                throw new IllegalArgumentException("Unknown block or item");
+            }
 
-        EntityRef item = blockItemFactory.newInstance(blockFamily, quantity);
-        if (!item.exists()) {
-            throw new IllegalArgumentException("Unknown block or item");
-        }
-        EntityRef playerEntity = client.getComponent(ClientComponent.class).character;
+            EntityRef playerEntity = client.getComponent(ClientComponent.class).character;
 
-        GiveItemEvent giveItemEvent = new GiveItemEvent(playerEntity);
-        item.send(giveItemEvent);
-        if (!giveItemEvent.isHandled()) {
-            item.destroy();
+            GiveItemEvent giveItemEvent = new GiveItemEvent(playerEntity);
+            item.send(giveItemEvent);
+            if (!giveItemEvent.isHandled()) {
+                item.destroy();
+            }
         }
 
         return "You received " + quantity + " blocks of " + blockFamily.getDisplayName();


GooeyHub:

portokaliu:
@J0hnCena I'm pretty sure changing the stackCount from bye to int wouldn't have that much of an impact. Don't forget that you also have to change the blockItemFactory settings when you want to improve this system (since it casts the int quantity as a byte specifically).

J0hnCena:
I moved the logic of dividing the blocks into multiple stacks into InventoryAuthoritySystem.java and changed stackCount from an int to a byte. I had to disable some modules such as Workstation, since they depend on stackCount being a byte. This implementation works, but should I just add a separate field to ItemComponent just to contain the blocks when they are given? It would make these changes compatible with other modules, and it would be less lines changed overall, since I could revert the byte casting.

Cervator:
Heya @J0hnCena and sorry for the delay in catching back up with regular engine PRs :-)

In this case I would recommend sticking with byte for minimal changes and accept the limitation of 99 items, at least for now. Maybe then submit an issue for later consideration if we should increase the stack size limit. I pretty much thought 99 was the max stack size anyway, was surprised to see we could get to 127 (makes sense with the byte thing)

And that's actually more for API compliance than anything. It would technically be a breaking change as per SemVer requiring us to bump the major version number for the next release.

modules/Core/src/main/java/org/terasology/logic/inventory/InventoryAuthoritySystem.java
```diff
@@ -170,25 +172,24 @@ private boolean giveItemToSlots(EntityRef instigator, EntityRef entity, EntityRe
                 if (toAdd > 0) {
                     consumableCount.put(slot, toAdd);
                     toConsume -= toAdd;
-                    if (toConsume == 0) {
+                    if (toConsume <= 0) {
```
flo: This change isn't necessary as toAdd got set to a value originally <= toConsume

modules/Core/src/main/java/org/terasology/logic/inventory/InventoryAuthoritySystem.java
```diff
@@ -205,12 +206,15 @@ private boolean giveItemToSlots(EntityRef instigator, EntityRef entity, EntityRe
             InventoryUtils.adjustStackSize(entity, slot, itemAt.stackCount + count);
         }
 
-        if (emptySlotNo > -1) {
-            ItemComponent sourceItem = item.getComponent(ItemComponent.class);
-            sourceItem.stackCount = (byte) emptySlotCount;
-            item.saveComponent(sourceItem);
-
-            InventoryUtils.putItemIntoSlot(entity, item, emptySlotNo);
+        if (!emptySlots.isEmpty()) {
```
flo: This if condition looks wrong to me. If we put copies into the empty slots then the original item will remain without location and does not get destroyed.

engine/src/main/java/org/terasology/logic/inventory/ItemComponent.java
```diff
@@ -49,7 +49,7 @@
      * How many of said item are there in this stack
      */
     @Replicate(FieldReplicateType.SERVER_TO_CLIENT)
-    public byte stackCount = 1;
+    public int stackCount = 1;
```
flo: You make stackCount and int but maxStackCount remains a byte. It's always good when a PR changes just one thing and not multiple. So I would suggest you revert this change.

Cervator:
@J0hnCena are you still out there by chance?

J0hnCena:
Sorry about the long delay. I changed stackCount back to an int. I also removed the else statement surrounding item.destroy(), so that item is cleaned up.

J0hnCena:
Actually hold up, there is a bug with this where if I pick up items thrown with the give command, they get put back onto the same stack even if the stack is at max size. Should I just revert back to the multiple commands logic or create new entities for the stacks in the if(!emptySlots.isEmpty()) statement? If I do the former, the problem of giving more blocks than 127 is also solved, since it is taken care of in multiple commands.

Cervator:
No worries nor rush :-)

Yeah probably just keep it simple with the multiple commands. We can always enhance it more later.

J0hnCena:
I reverted my changes from InventoryAuthoritySystem and BlockCommands.

Cervator:
Thanks @J0hnCena :-)

Looks nice and simple now, with one code style issue remaining: looks like you've got tab characters in there instead of spaces. We're a 4-spaces kinda group. Not a big deal to fix during merge if needed but might want to tweak project setup if you've got more stuff coming (hopefully!)

@flo: any remaining concerns or looking good for merge?

Cervator:
Tested again both single and multi and merged!

Did note one sort of edge case: you can request a *huge* amount of blocks (like 9999999) and it'll stall the game while going through the command a whole lot of times. Probably an ideal fix there is to catch when you run out of inventory space then break the loop and stop giving more blocks.

Unsure how to ideally detect that and it may change with @jellysnake's work in #2789 - so I just noted it in the TODO for us to catch up on later :-)

Thanks @J0hnCena! Added you and a bunch of other recent additions to the Credits file finally.

PR #2682, opened 2016-12-19
Fixed a bunch of Stylechecker issues

jellysnake:
### Contains
Fixes for the following Stylechecker issues

Add @override to methods with @inheritDoc javadoc tags (8)
Remove redundant private modifiers (18)
Remove redundant public modifiers (51)
Set value is the same as default (3)
Character not followed by a space (1)
Variables defined in wrong order (7)
Line too long (2)
Remove redundant static modifer (1)
'-' on wrong line (1)
Class should be declared as final (1)
@Depracated javadoc tag should be added (7)

### How to test

Ensure that the terasology engine is unchanged from last pull

### DIFF
--- a/engine/src/main/java/org/terasology/audio/openAL/staticSound/OpenALSound.java
+++ b/engine/src/main/java/org/terasology/audio/openAL/staticSound/OpenALSound.java
@@ -124,7 +124,7 @@ public final class OpenALSound extends StaticSound {
         private int bufferId;
         private final WeakReference<OpenALSound> asset;
 
-        public DisposalAction(ResourceUrn urn, OpenALSound openALSound) {
+        DisposalAction(ResourceUrn urn, OpenALSound openALSound) {
             this.urn = urn;
              asset = new WeakReference<>(openALSound);
         }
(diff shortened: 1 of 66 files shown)


GooeyHub:

emanuele3d:
Hi @jellysnake, and thank you for your effort.

I think the access-related changes, i.e. switching things from/to public/package/private might be a bit heavy-handed. It's true the checkstyle will flag them as issues because, say, a public method is not used by anything in the engine or is not used outside of its package. But it might be used by a module and CheckStyle might not know about it. 

So, I would avoid demoting anything from public to package-private or fully private unless you really mean to make something less accessible.

oniatus:
@emanuele3d 
I did a short review and the public markers from checkstyle seem to be okay to me. Most of them are enums where the constructor should be private or package private or package private classes where a public constructor won't increase the accessibility.

Let me know if i missed some false positives. In this case I would try to update the checkstyle config accordingly.

emanuele3d:
I see what you mean. Ok, please disregard my last comment then: I hadn't realized they were enums.

jellysnake:
Ah, okay. I get what you mean.  
I was under the impression however that the public/private I removed were ones that were not needed.  
Ie the default is public and hence explicitly writing public is not needed.

Cervator:
Like noted over in #2686 IMHO the access level warnings are a little dubious, yeah. Although it does seem like Checkstyle and IntelliJ differ in the amount of complaining they do - Checkstyle seems far more picky and it appears most or all warnings addressed in these two PRs relate to enums, inner classes, etc, unlike the spurious places IntelliJ complains from just thinking something isn't in public use.

So with that in mind I'll look to merge both. The Checkstyle config might actually be solid, but maybe there's a way we can make IntelliJ listen a bit more and throw out its overly chatty warnings that we've tried to avoid messing with so far.

PR #2691, opened 2016-12-21
InputEvent, PathManager, RenderSystem javadocs

MaxBorsch:
Added javadocs for RenderSystem, PathManager, InputEvent, and TerasologyConstants. 
No particular target issue, just roamed around the engine and found some undocumented important classes.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/TerasologyConstants.java
+++ b/engine/src/main/java/org/terasology/engine/TerasologyConstants.java
@@ -24,18 +24,52 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 
 /**
+ * This class contains various important file-related constants used by the Terasology engine.
  */
 public final class TerasologyConstants {
+    /**
+     * Name and extension of an entities data file.
+     */
     public static final String ENTITY_DATA_FILE = "entity.dat";
+    /**
+     * Terasology's default server port.
+     */
     public static final int DEFAULT_PORT = 25777;
+    /**
+     * Name and extension of a world data file.
+     */
     public static final String WORLD_DATA_FILE = "world.dat";
+    /**
+     * The name of the default world.
+     */
     public static final String MAIN_WORLD = "main";
+    /**
+     * Default charset used by Terasology.
+     */
     public static final Charset CHARSET = Charsets.UTF_8;
+    /**
+     * Name of the engine module.
+     */
     public static final Name ENGINE_MODULE = new Name("engine");
+    /**
+     * Name of the core gameplay module.
+     */
     public static final Name CORE_GAMEPLAY_MODULE = new Name("coresamplegameplay");
+    /**
+     * Name of a module's assets directory.
+     */
     public static final String ASSETS_SUBDIRECTORY = "assets";
+    /**
+     * Name of a module's overrides directory.
+     */
     public static final String OVERRIDES_SUBDIRECTORY = "overrides";
+    /**
+     * Name of a module's deltas directory.
+     */
     public static final String DELTAS_SUBDIRECTORY = "deltas";
+    /**
+     * Path to a module's info file.
+     */
     public static final Path MODULE_INFO_FILENAME = Paths.get("module.txt");
 
     private TerasologyConstants() {
--- a/engine/src/main/java/org/terasology/engine/TerasologyConstants.java
+++ b/engine/src/main/java/org/terasology/engine/TerasologyConstants.java
@@ -24,11 +24,11 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 
 /**
- * This class contains various important file-related constants used by the Terasology engine.
+ * This class contains various important constants used by the Terasology engine.
  */
 public final class TerasologyConstants {
     /**
-     * Name and extension of an entities data file.
+     * Name and extension of an entities data file. (Not currently in use) Used to contain state data on world entities for loading/saving. See AbstractStorageManager for details.
      */
     public static final String ENTITY_DATA_FILE = "entity.dat";
     /**
@@ -48,11 +48,12 @@ public final class TerasologyConstants {
      */
     public static final Charset CHARSET = Charsets.UTF_8;
     /**
-     * Name of the engine module.
+     * Name of the engine module. The engine module contains engine features, without any gameplay elements.
+     * Examples of specific features include the entity system, noise functions, core blocks, and utilities.
      */
     public static final Name ENGINE_MODULE = new Name("engine");
     /**
-     * Name of the core gameplay module.
+     * Name of the core gameplay module. The core gameplay module adds a minimal environment for some basic gameplay.
      */
     public static final Name CORE_GAMEPLAY_MODULE = new Name("coresamplegameplay");
     /**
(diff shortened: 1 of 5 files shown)


GooeyHub:

engine/src/main/java/org/terasology/input/events/InputEvent.java
```diff
@@ -50,6 +58,10 @@ public Vector3i getTargetBlockPosition() {
         return targetBlockPosition;
     }
 
+    /**
+     *
+     * @return The time since the event was fired (also the game update loop's delta time).
```
skaldarnar: What is the connection between the update delta and this one? Is the event's delta always updated, or is it just the delta of the update pass processing it?

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
```diff
@@ -24,18 +24,52 @@
 import java.nio.file.Paths;
 
 /**
+ * This class contains various important file-related constants used by the Terasology engine.
  */
 public final class TerasologyConstants {
+    /**
+     * Name and extension of an entities data file.
```
emanuele3d: I think I could use some further information: what is an entity data file?

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
```diff
@@ -24,18 +24,52 @@
...
+    /**
+     * Default charset used by Terasology.
+     */
     public static final Charset CHARSET = Charsets.UTF_8;
+    /**
+     * Name of the engine module.
+     */
     public static final Name ENGINE_MODULE = new Name("engine");
+    /**
+     * Name of the core gameplay module.
```
emanuele3d: It would be useful here and in the next javadoc, to give a short sentence about what the engine module provides vs what the core gameplay module provides.

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
```diff
@@ -24,18 +24,52 @@
...
+     */
     public static final Name CORE_GAMEPLAY_MODULE = new Name("coresamplegameplay");
+    /**
+     * Name of a module's assets directory.
+     */
     public static final String ASSETS_SUBDIRECTORY = "assets";
+    /**
+     * Name of a module's overrides directory.
+     */
     public static final String OVERRIDES_SUBDIRECTORY = "overrides";
```
emanuele3d: What's an override and what's a delta (below)? How does they fit in the context of Terasology? Potentially you could give this kind of information pointing to a wiki page that explains all these things. Sometimes links break though: a short description would at least allow to narrow the search.

engine/src/main/java/org/terasology/entitySystem/systems/RenderSystem.java
```diff
@@ -16,16 +16,32 @@
 package org.terasology.entitySystem.systems;
 
 /**
+ * Interface for component systems that needs to be updated every time certain game elements are rendered (generally to render something).
```
emanuele3d: Tricky tricky tricky. I'm struggling to find a proper description for this.

I guess we could write:

_Instances implementing this interface see their methods called at specific stages during the rendering process, with an OpenGL state compatible with what needs to be rendered - i.e. opaque objects (a fairly standard opengl state) vs semi-transparent objects (opengl has blending enabled). This allows a RenderSystem to update its own entities or inject its own objects for rendering._  

That been said, long term this interface might disappear.

engine/src/main/java/org/terasology/entitySystem/systems/RenderSystem.java
```diff
@@ -16,16 +16,32 @@
...
  */
 public interface RenderSystem extends ComponentSystem {
 
+    /**
+     * Called when opaque objects should be rendered.
```
emanuele3d: I'd write:

_Called with an opengl state useful to the rendering of opaque objects. See OpaqueObjectsNode for more information._

I'd describe the other methods in a similar fashion.

engine/src/main/java/org/terasology/entitySystem/systems/RenderSystem.java
```diff
@@ -16,16 +16,32 @@
...
 
+    /**
+     * Called when overlay objects should be rendered.
+     */
     void renderOverlay();
 
+    /**
+     * Called when first person objects should be rendered.
+     */
     void renderFirstPerson();
```
emanuele3d: Leave this as it is: we'll get rid of this method at some point. Everything should be placed in-world. If it's not in world it's either background, overlay, or on the very top UI. Which potentially could be an UI anyway. But I'm thinking out loud here.

engine/src/main/java/org/terasology/entitySystem/systems/RenderSystem.java
```diff
@@ -16,16 +16,32 @@
...
 
+    /**
+     * TODO: Implement RenderSystem.renderShadows()
```
MaxBorsch: renderShadows() is never called by a rendering node, so implementing it is useless.

engine/src/main/java/org/terasology/input/events/InputEvent.java
MaxBorsch: It's the delta of the update loop processing it, which is equal to the delta since the event was fired since all events are processed every loop.

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
MaxBorsch: I'm having trouble finding a wiki page that explains overrides and deltas. The way I understand overrides is changed versions of existing classes. Not quite sure what deltas are.

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
MaxBorsch: It's never used in the project :) We may never know! A lot of classes seem to redefine these constants because they didn't know this class existed. For example StoragePathProvider defines GLOBAL_ENTITY_STORE as "global.dat" as well as new world directory names and player data file names.

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
oniatus: You can link to this one: [Deltas and Overrides](https://github.com/Terasology/TutorialAssetSystem/wiki/Deltas-and-Overrides). I have a section for deltas and overrides in another (well hidden 😃 ) wiki draft and will create a page from it in the next hour.

engine/src/main/java/org/terasology/entitySystem/systems/RenderSystem.java
skaldarnar: @emanuele3d do you have to say anything to the rendering system's defense❓ Should this part of the interface be properly documented, is it obsolete and can be removed, or what are the plans for this?

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
skaldarnar: Then the constant could either be removed, or it should be used in the spots you mention. IMO, it would be a good idea to include those fixes in this PR.

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
skaldarnar: @oniatus to the rescue! We could grep for wiki links in the code base and check if they are existing in our build scripts to have at least some safety there...

engine/src/main/java/org/terasology/entitySystem/systems/RenderSystem.java
skaldarnar: You could even use the `@Deprecated` annotation and leave what Manu said as reference.

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
MaxBorsch: @skaldarnar Does it make sense to describe the core gameplay module's features here? After all this is a constant storage class, so if you need a constant reference to the core gameplay module then you must already know what the core gameplay module does. Shouldn't that documentation go in the core module?

skaldarnar:
@emanuele3d some doc, even the most simple one, helps us to get some discussion started. If we cannot come up with something better than proposed in a PR, I guess the proposed comment is not worse than none comment. 
That said, I see you point, where we could easily look out for not documented parts. If you feel like some comment is leading in the wrong direction, but you cannot think of something else, I'm fine with requesting to remove the comment (for now) or add a TODO for later reference.

GooeyHub:

emanuele3d:
Looks good to me too!

PR #2699, opened 2016-12-22
Fixed issue with untranslated text

### LINKED ISSUE #2698: Label in "Join Game" screen is broken
The label under the server list in the **Join Game** screen seems to be broken as it displays the i18n key instead of a proper label. This should be fixed so that the correct message is displayed.

iojw:
### Contains

Closes #2698 with a very simple one line fix 🙂 

Just an issue with the text being untranslated. It does seem a bit unintuitive that the translation is done in JoinGameScreen.java instead of ServerListDownloader.java but I went with this since the TranslationSystem object cannot be instantiated in ServerListDownloader.

### How to test

Open the Join Game screen. Verify that the translated text is shown.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
@@ -379,7 +379,7 @@ public class JoinGameScreen extends CoreScreenLayer {
             downloadLabel.bindText(new ReadOnlyBinding<String>() {
                 @Override
                 public String get() {
-                    return downloader.getStatus();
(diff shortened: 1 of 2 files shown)


GooeyHub:

skaldarnar:
@iojw Please add Javadoc to the `ServerListDownloader` stating that the `status` String/`getStatus()` method return a i18n key. This is not very intuitive and should be noted somewhere to prevent this bug in the future.

iojw:
@skaldarnar I've added the  javadoc to ServerListDownloader. Let me know if there's anything you would like me to change.

GooeyHub:

skaldarnar:
Thanks, looks all good now 😄

PR #2701, opened 2016-12-23
Change how the inventory is opened and closed

jellysnake:
Currently, the inventory is opened via [`InventoryUIClientSystem`](https://github.com/MovingBlocks/Terasology/blob/develop/modules/Core/src/main/java/org/terasology/logic/inventory/InventoryUIClientSystem.java#L46) in `onToggleInventory`. For seemingly no reason it is closed in [`InventoryScreen`](https://github.com/MovingBlocks/Terasology/blob/develop/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryScreen.java#L46) in `onBindEvent`.

The closing section in `InventoryScreen` appears to be redundant as when `onBindEvent` is removed the inventory opens and closes as normal. I am aware that I might simply be missing something which explains the logic behind this.

Also, removing `onBindEvent` also allows for other screens to be opened and closed alongside the inventory screen, as I've done in [my fork](http://github.com/jellysnake/Potions) with `potionStatusScreen` (See files [`PotionStatusScreen`](https://github.com/jellysnake/Potions/blob/master/src/main/java/org/terasology/potions/PotionStatusScreen.java) and [`PotionStatusUISystem`](https://github.com/jellysnake/Potions/blob/master/src/main/java/org/terasology/potions/PotionStatusUISystem.java))

(Also, I'm sorry if this is the wrong place for this, but I didn't think it fit on the forum)

### DIFF
--- a/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryScreen.java
+++ b/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryScreen.java
@@ -16,15 +16,12 @@
 package org.terasology.rendering.nui.layers.ingame.inventory;
 
 import org.terasology.entitySystem.entity.EntityRef;
-import org.terasology.input.BindButtonEvent;
-import org.terasology.input.binds.inventory.InventoryButton;
 import org.terasology.logic.players.LocalPlayer;
 import org.terasology.registry.In;
 import org.terasology.rendering.nui.CoreScreenLayer;
 import org.terasology.rendering.nui.databinding.ReadOnlyBinding;
 
-/**
- */
+
 public class InventoryScreen extends CoreScreenLayer {
 
     @In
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
This is a fine place for this and I'm very curious how it works out as we have a known issue about opening/closing some screens - see https://github.com/Terasology/JoshariasSurvival/issues/15 for the primary example and some more thoughts.

Might take a little extra testing to figure this one out (unless maybe @portokaliu can spot it immediately)

@GooeyHub: add to whitelist

GooeyHub:

iojw:
I've tested this out with a few interaction screens and it seems to work the same as before. 

`onBindEvent` seems to be called at a higher level in `NUIManagerInternal` whenever a raw input event is received. However, both pieces of code seem to accomplish the same thing.

 This `onBindEvent` code was added in 3c02cfb0f4f688cad207b33c6adefbd063315e93 along with `ContainerScreen` even though the part about `nuiManager.toggleScreen()` already existed then. Additionally `InventoryScreen` seems to be the only class that actually implements `onBindEvent`

I'm not exactly sure what all this means, but just throwing what I found out here. 🙂

GooeyHub:

GooeyHub:

Cervator:
Tested this out along with #2703 and inventory stuff seems working as well and better (no double `e`) than before :-)

@jellysnake do you have your potions UI piece ready? @iojw mentions in #2703 that we might want to make any one interaction screen opening first close all other ones which may well make sense. Would your stuff count as a background element like the toolbar? Which would sound sensible.

PR #2702, opened 2016-12-24
Use dimensions instead to compare SetViewportSizeOf instances

### LINKED ISSUE #2688: SetViewportToSizeOf class equality should be dependent on width/height, not fbo name
Ensure glViewport is called only if necessary.

### What actually happened
It gets called whenever the FBO changes from one node to the next. Should change only when the width/height dimensions of the FBOs in consecutive nodes change.

### How to reproduce
- Run Terasology, start a game, exit terasology.
- Check log for RENDER TASK LIST: consecutive nodes should have SetViewportToSizeOfTask lines with different width/height values, or the following nodes should have no SetViewportToSizeOfTask line at all.

### Cause
This is due to the SetViewportToSizeOf class comparing instances based on the name of the FBO they get the dimensions from. Instead, they should compare the width and height values. Methods hashCode and equals must be refactored accordingly. Obtaining the fbo dimensions is the architectural challenge.

It shouldn't be too difficult but must be handled carefully.

eviltak:
Fixes #2688.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
@@ -97,7 +97,7 @@ public final class RenderTaskListGenerator {
         taskList.clear();
         Node nextEnabledNode;
         StateChange persistentStateChange;
-        Map persistentStateChanges = Maps.newHashMap();  // assuming we can't make it a private field for the time being
+        Map<Class<?>, StateChange> persistentStateChanges = Maps.newHashMap();  // assuming we can't make it a private field for the time being
 
         int enabledNodes = 0;
         int potentialTasks = 0;
@@ -118,7 +118,7 @@ public final class RenderTaskListGenerator {
 
                     // State changes persist beyond the node that request them if following nodes
                     // require the exact same change.
-                    persistentStateChange = (StateChange) persistentStateChanges.get(currentStateChange.getClass());
+                    persistentStateChange = persistentStateChanges.get(currentStateChange.getClass());
 
                     // for a state change to be necessary there can't be an identical one already persisting
                     if (persistentStateChange == null || !currentStateChange.equals(persistentStateChange)) {
@@ -142,8 +142,16 @@ public final class RenderTaskListGenerator {
                     // will bind another FBO.
                     for (StateChange currentStateReset : node.getDesiredStateResets()) {
                         if (sameClassStateChangeNotFoundInThe(nextEnabledNode, currentStateReset)) {
-                            taskList.add(currentStateReset.generateTask());
-                            persistentStateChanges.remove(currentStateReset.getClass());
+                            persistentStateChange = persistentStateChanges.get(currentStateReset.getClass());
+
+                            // A reset state change is necessary only if an identical state change isn't already persisting.
+                            if (persistentStateChange == null || !currentStateReset.equals(persistentStateChange)) {
+                                taskList.add(currentStateReset.generateTask());
+
+                                // This makes sure that the reset state change persists so that future state changes
+                                // which are identical to the reset state change are not added.
+                                persistentStateChanges.put(currentStateReset.getClass(), currentStateReset);
+                            }
                         }
                     }
 
@@ -177,10 +185,11 @@ public final class RenderTaskListGenerator {
         return taskList;
     }
 
+    // TODO: Rename method and remove comments
     private boolean sameClassStateChangeNotFoundInThe(Node nextEnabledNode, StateChange stateChangeReset) {
 
         for (StateChange stateChange : nextEnabledNode.getDesiredStateChanges()) {
-            if (stateChange.getClass() == stateChangeReset.getClass()) {
+            if (stateChange.getClass() == stateChangeReset.getClass() ) {
                 return false; // we did find it! And yes, returning false is correct, see method name.
 
                 // note: we don't worry about the details of the two state changes (i.e. if they are value-identical)
--- a/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
@@ -189,7 +189,7 @@ public final class RenderTaskListGenerator {
     private boolean sameClassStateChangeNotFoundInThe(Node nextEnabledNode, StateChange stateChangeReset) {
 
         for (StateChange stateChange : nextEnabledNode.getDesiredStateChanges()) {
-            if (stateChange.getClass() == stateChangeReset.getClass() ) {
+            if (stateChange.getClass() == stateChangeReset.getClass()) {
                 return false; // we did find it! And yes, returning false is correct, see method name.
 
                 // note: we don't worry about the details of the two state changes (i.e. if they are value-identical)
--- a/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
@@ -185,7 +185,6 @@ public final class RenderTaskListGenerator {
         return taskList;
     }
 
-    // TODO: Rename method and remove comments
     private boolean sameClassStateChangeNotFoundInThe(Node nextEnabledNode, StateChange stateChangeReset) {
 
         for (StateChange stateChange : nextEnabledNode.getDesiredStateChanges()) {
--- a/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
@@ -189,7 +189,7 @@ public final class RenderTaskListGenerator {
     private boolean sameClassStateChangeNotFoundInThe(Node nextEnabledNode, StateChange stateChangeReset) {
 
         for (StateChange stateChange : nextEnabledNode.getDesiredStateChanges()) {
-            if (stateChange.getClass() == stateChangeReset.getClass()) {
+            if (stateChange.getClass() == stateChangeReset.getClass() ) {
                 return false; // we did find it! And yes, returning false is correct, see method name.
 
                 // note: we don't worry about the details of the two state changes (i.e. if they are value-identical)
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
@@ -67,7 +67,6 @@ public final class SetViewportToSizeOf implements FBOManagerSubscriber, StateCha
         return task;
     }
 
-    // TODO: change equals and hashCode to use dimensions instead.
     @Override
     public int hashCode() {
         // Generates a unique 32 bit signed integer from two integers.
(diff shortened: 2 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -177,10 +185,11 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
         return taskList;
     }
 
+    // TODO: Rename method and remove comments
     private boolean sameClassStateChangeNotFoundInThe(Node nextEnabledNode, StateChange stateChangeReset) {
 
         for (StateChange stateChange : nextEnabledNode.getDesiredStateChanges()) {
-            if (stateChange.getClass() == stateChangeReset.getClass()) {
+            if (stateChange.getClass() == stateChangeReset.getClass() ) {
```
rzats: This should be reverted.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -65,12 +69,20 @@ public RenderPipelineTask generateTask() {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(fboName);
+        // Generates a unique 32 bit signed integer from two integers.
+        // This will return unique values for the following reasonable ranges:
+        // width < (1 << 16) (65536) and height < (1 << 15) (32768)
+        return getWidth() << 15 | getHeight();
```
rzats: The ranges _are_ reasonable, but why impose arbitrary restrictions when http://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#hash(java.lang.Object...) is a thing? 😛

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -65,12 +69,20 @@ public RenderPipelineTask generateTask() {
...
 
     @Override
     public boolean equals(Object obj) {
-        return (obj instanceof SetViewportToSizeOf) && this.fboName.equals(((SetViewportToSizeOf) obj).getFboName());
+        if (!(obj instanceof SetViewportToSizeOf))
+            return false;
+
+        SetViewportToSizeOf other = (SetViewportToSizeOf) obj;
+
+        return getWidth() == other.getWidth() && getHeight() == other.getHeight();
```
rzats: This method could be converted into a single (three-line) boolean statement.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -80,20 +92,28 @@ public boolean isTheDefaultInstance() {
...
     @Override
     public String toString() { // TODO: used for logging purposes at the moment, investigate different methods
         return String.format("%30s: %s", this.getClass().getSimpleName(), fboName);
     }
 
-    public ResourceUrn getFboName() {
-        return fboName;
+    public int getWidth() {
+        FBO fbo = getFbo();
+        return fbo.width();
```
rzats: A one-liner is fine here.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -80,20 +92,28 @@ public boolean isTheDefaultInstance() {
...
+    }
+
+    public int getHeight() {
+        FBO fbo = getFbo();
+        return fbo.height();
```
rzats: A one-liner is fine here. [2]

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -80,20 +92,28 @@ public boolean isTheDefaultInstance() {
...
+    }
+
+    private FBO getFbo() {
+        if (defaultDynamicFBO == null)
+            return frameBuffersManager.get(fboName);
+
+        return defaultDynamicFBO.getFbo();
```
rzats: A ternary statement is fine here.

Cervator:
@GooeyHub: ok to test

@eviltak: thanks! :-) And yeah we like keeping to the code style, hehe :D

GooeyHub:

emanuele3d:
As suggested in issue #2688 I feel there might be a misunderstanding at play here. I will hold on reviewing this PR until @eviltak and I are reasonably certain we understood each other in that issue.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
eviltak: I was looking for something like that, thanks.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
eviltak: I'm not sure whether a ternary statement would make the code more beautiful or any more easier to read, but if this project's code formatting guidelines requires it, I'll do it.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
eviltak: I think this is cleaner and _much_ more easily understandable than casting `obj` to `SetViewportToSizeOf` twice in a huge boolean statement.

GooeyHub:

GooeyHub:

eviltak:
As per discussion in #2688, this PR is limited to just using FBO dimensions as a comparison heuristic for `SetViewportToSizeOf` state changes. The original problem stated in #2688 (i.e. generation of duplicated or redundant tasks) will be fixed in a future PR.

emanuele3d:
Just to be clear, let's hold on this PR for a little longer. I suspect we'll merge at least some code from it.

GooeyHub:

Cervator:
Bump! Have merged #2710 and can confirm it plus this one together work fine and result in a lowered task count.

What's next for this one? And other tweaks needed or can we merge this as well?

@emanuele3d @eviltak ping :-) And Happy New Year!

eviltak:
@Cervator I'm not sure, I've implemented all the requested changes except the one which I think is unnecessary and the code works. Probably should wait for @emanuele3d or @rzats to review the PR again.

emanuele3d:
I think I have identified a problem with this PR but I don't have the time to verify it. Please stand-by. I hope I will get around working on it tonight, when the family has gone to bed and the fireworks lit up the sky. =)

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
emanuele3d: In fairness it does make the code in this particular case _a little_ easier to read. 

I'm tempted to say that we do not actually expect compared objects being anything else but StateChanges. So, potentially we could cast obj straight into a StateChange and skip the instanceof check. We'd only have to include in the javadoc of this method that it will generate an exception if obj isn't actually a StateChange. What do you guys think?

In any case state changes `SetBlendFunction`, `SetInputTexture `and `SetInputTextureFromFBO` could benefit from the same treatment.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -80,20 +89,24 @@ public boolean isTheDefaultInstance() {
...
 
-    public ResourceUrn getFboName() {
-        return fboName;
+    public int getWidth() {
+        return getFbo().width();
+    }
+
+    public int getHeight() {
+        return getFbo().height();
+    }
```
emanuele3d: It was a conscious decision to _not_ provide much in the way of accessors in all StateChange classes. 

It was tempting but I came to the conclusion that all that matters in the Intermediate List (listing nodes and state changes rather than tasks) is where that information comes from. In this class the width and height come from an FBO and that's why the toString method publishes the FBO's name to the log. The associated task (if it makes it into the task list) then publishes the deeper information, in this case the width and height.

So I'd prefer to remove these methods and use getFbo().width() straight into the update() method. 

Good idea about adding the getFbo() method though.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -142,8 +142,16 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
...
+                            persistentStateChange = persistentStateChanges.get(currentStateReset.getClass());
+
+                            // A reset state change is necessary only if an identical state change isn't already persisting.
+                            if (persistentStateChange == null || !currentStateReset.equals(persistentStateChange)) {
+                                taskList.add(currentStateReset.generateTask());
+
+                                // This makes sure that the reset state change persists so that future state changes
+                                // which are identical to the reset state change are not added.
+                                persistentStateChanges.put(currentStateReset.getClass(), currentStateReset);
+                            }
```
emanuele3d: All good, thank you.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -19,12 +19,16 @@
...
+
 import java.util.Objects;
+
 import org.terasology.rendering.dag.RenderPipelineTask;
 import org.terasology.rendering.dag.StateChange;
 import org.terasology.rendering.dag.tasks.SetViewportToSizeOfTask;
 import org.terasology.rendering.opengl.FBO;
+
 import static org.terasology.rendering.opengl.DefaultDynamicFBOs.READ_ONLY_GBUFFER;
+
```
emanuele3d: Good idea removing these.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -65,12 +69,17 @@ public RenderPipelineTask generateTask() {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(fboName);
+        return Objects.hash(getWidth(), getHeight());
```
emanuele3d: .hash is very convenient, isn't it.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -177,6 +185,7 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
         return taskList;
     }
 
+    // TODO: Rename method and remove comments
```
eviltak: That seems to have not been removed from back when I thought this method was causing the problems 😄. Removed.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
eviltak: I don't really think that an `equals` method should throw an `Exception`, and it is in good practice to return `false` if the compared object is not an instance of the same class (unless that is not the intended behavior).

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
eviltak: That's reformat code for you. I'll remove the new lines.

GooeyHub:

emanuele3d:
@rzats & @Cervator: from my perspective this PR is good to go.

Cervator:
Merged! Thanks to all involved :-)

PR #2704, opened 2016-12-24
Add javadocs and sometimes comment

gkaretka:
### Contains

Add javadocs and sometimes comment

### DIFF
--- a/engine/src/main/java/org/terasology/logic/actions/PlaySoundAction.java
+++ b/engine/src/main/java/org/terasology/logic/actions/PlaySoundAction.java
@@ -31,6 +31,7 @@ import org.terasology.utilities.random.FastRandom;
 import org.terasology.utilities.random.Random;
 
 /**
+ * Play sound nearby player
  */
 @RegisterSystem(RegisterMode.ALWAYS)
 public class PlaySoundAction extends BaseComponentSystem {
@@ -43,6 +44,10 @@ public class PlaySoundAction extends BaseComponentSystem {
     @In
     private LocalPlayer localPlayer;
 
+    /**
+     * @param event
+     * @param entity
+     */
     @ReceiveEvent(components = {PlaySoundActionComponent.class})
     public void onActivationPredicted(ActivationPredicted event, EntityRef entity) {
         PlaySoundActionComponent playSound = entity.getComponent(PlaySoundActionComponent.class);
@@ -64,6 +69,10 @@ public class PlaySoundAction extends BaseComponentSystem {
         }
     }
 
+    /**
+     * @param event
+     * @param entity
+     */
     @ReceiveEvent(components = {PlaySoundActionComponent.class})
     public void onActivate(ActivateEvent event, EntityRef entity) {
         if (event.getInstigator().equals(localPlayer.getCharacterEntity())) {
--- a/engine/src/main/java/org/terasology/logic/actions/PlaySoundAction.java
+++ b/engine/src/main/java/org/terasology/logic/actions/PlaySoundAction.java
@@ -31,7 +31,8 @@ import org.terasology.utilities.random.FastRandom;
 import org.terasology.utilities.random.Random;
 
 /**
- * Play sound nearby player
+ * This class is responsible for managing audio in-game
+ * when triggered by some event on client side and also on server side.
  */
 @RegisterSystem(RegisterMode.ALWAYS)
 public class PlaySoundAction extends BaseComponentSystem {
@@ -45,8 +46,8 @@ public class PlaySoundAction extends BaseComponentSystem {
     private LocalPlayer localPlayer;
 
     /**
-     * @param event
-     * @param entity
+     * @param event contains the details for the predicted event, used here for location purposes
+     * @param entity is source of the playsound
      */
     @ReceiveEvent(components = {PlaySoundActionComponent.class})
     public void onActivationPredicted(ActivationPredicted event, EntityRef entity) {
@@ -70,8 +71,8 @@ public class PlaySoundAction extends BaseComponentSystem {
     }
 
     /**
-     * @param event
-     * @param entity
+     * @param event contains the details for the active event, used here for location purposes
+     * @param entity is source of the playsound
      */
     @ReceiveEvent(components = {PlaySoundActionComponent.class})
     public void onActivate(ActivateEvent event, EntityRef entity) {
--- a/engine/src/main/java/org/terasology/logic/actions/PlaySoundAction.java
+++ b/engine/src/main/java/org/terasology/logic/actions/PlaySoundAction.java
@@ -46,6 +46,7 @@ public class PlaySoundAction extends BaseComponentSystem {
     private LocalPlayer localPlayer;
 
     /**
+     * This method plays sound in prediction for preventing not playing song because of server lags
      * @param event contains the details for the predicted event, used here for location purposes
      * @param entity is source of the playsound
      */
@@ -71,6 +72,7 @@ public class PlaySoundAction extends BaseComponentSystem {
     }
 
     /**
+     * This method plays sound if it wasn't played by prediction system
      * @param event contains the details for the active event, used here for location purposes
      * @param entity is source of the playsound
      */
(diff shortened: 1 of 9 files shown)


GooeyHub:

engine/src/main/java/org/terasology/logic/actions/PlaySoundActionComponent.java
```diff
@@ -34,6 +34,10 @@
     public PlaySoundActionComponent() {
     }
 
+    /**
+     * Add sounds to list of sounds
```
Cervator: This again could be written with a bit more meaning, and the `sounds` param needs a line as well.

In this case `sounds` is actually a "vararg" parameter, meaning you can pass any number of StaticSound variables to this method. Which is in fact a constructor, not just a regular method. So you could say something like "Creates a new instance of `PlaySoundActionComponent` adding all provided sounds via vararg to the instance variable `sounds`" and the `@param sounds` could say "the provided sounds via vararg"

engine/src/main/java/org/terasology/logic/ai/AICommands.java
```diff
@@ -25,26 +25,38 @@
 import com.google.common.collect.Iterables;
 
 /**
+ * Artificial Intelligence command
```
Cervator: Really vague. Yes, "AI Commands" naturally contains commands related to AI, spelling that out doesn't really help :-)

In this case I believe the commands all relate to entities that have particular AI components attached, interacting with them somehow.

engine/src/main/java/org/terasology/logic/ai/AICommands.java
```diff
@@ -25,26 +25,38 @@
...
  */
 @RegisterSystem
 public class AICommands extends BaseComponentSystem {
     @In
     private EntityManager entityManager;
 
+    /**
+     * Lists all AIs in the world
```
Cervator: This is pretty close but note how the method explicitly returns *counts* of AIs, not a *list* of AIs. There is a crucial difference between those. Your `@return` statement is spot on :-)

engine/src/main/java/org/terasology/logic/ai/AICommands.java
```diff
@@ -25,26 +25,38 @@
...
+     */
     @Command(runOnServer = true, shortDescription = "Count all AIs in the world")
     public String countAI() {
         int simpleAIs = Iterables.size(entityManager.getEntitiesWith(SimpleAIComponent.class));
         int hierarchical = Iterables.size(entityManager.getEntitiesWith(HierarchicalAIComponent.class));
         return "Simple AIs: " + simpleAIs + ", Hierarchical AIs: " + hierarchical;
     }
 
+    /**
+     * Destroys all AIs in the world
```
Cervator: This needs to include what is meant by destroying AIs. In this case the command goes through all *entities* with AI components attached, then destroys those entities. The return doc is unchanged from the former method, which is bad since it does differ, in this case in returns counts for how many were *destroyed*

engine/src/main/java/org/terasology/logic/ai/AICommands.java
```diff
@@ -25,26 +25,38 @@
...
+     * @return String string containing number of simple AIs and hierarchical AIs
+     */
     @Command(runOnServer = true, shortDescription = "Destroys all AIs in the world")
     public String destroyAI() {
         int simpleAI = 0;
+        // go through all simple AIs and destroy reference
```
Cervator: For once this comment is probably not needed - the code itself is pretty clear (and technically it is destroying the entities that have the component attached - not references. EntityRef is a reference to an entity)

engine/src/main/java/org/terasology/logic/ai/HierarchicalAISystem.java
```diff
@@ -59,6 +59,9 @@
 
     // TODO add way to recognize if attacked
 
+    /**
+     * @param delta The time (in seconds) since the last engine update.
```
Cervator: Here you do not need to add JavaDoc as the method is overriding the `update` from `UpdateSubscriberSystem` which has a JavaDoc statement on it. That statement will automatically apply for child methods like this one.

engine/src/main/java/org/terasology/logic/console/commands/ClientCommands.java
```diff
@@ -26,6 +26,7 @@
 import org.terasology.world.WorldProvider;
 
 /**
+ * Client command usable in F1 in-game console
```
Cervator: This is too vague again

engine/src/main/java/org/terasology/logic/console/commands/ClientCommands.java
```diff
@@ -36,12 +37,21 @@
     private WorldProvider worldProvider;
 
 
+    /**
+     * Displays debug information on the target entity
```
Cervator: This is good but it just mimics what's in the description. Technically what is meant here is that debug information is displayed *for the target the camera is pointing at* which could help explain the command a little better. In other words it displays information for what the player is looking at.

engine/src/main/java/org/terasology/logic/console/commands/ClientCommands.java
```diff
@@ -36,12 +37,21 @@
...
+     * @return String
+     */
     @Command(shortDescription = "Displays debug information on the target entity")
     public String debugTarget() {
         EntityRef cameraTarget = cameraTargetSystem.getTarget();
         return cameraTarget.toFullDescription();
     }
 
+    /**
+     * Sets the current world time of the in days fr the local player
```
Cervator: This is again simply copied from the description, including the weird typo. Probably fix both the description and the javadoc.

"Sets the current world time for the local player in days"

Be sure to also have a line for the parameter and return

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -142,27 +142,39 @@
...
+     * @param searched String
+     * @return String
+     */
     @Command(shortDescription = "Search commands/prefabs/assets",
              helpText = "Displays commands, prefabs, and assets with matching name, description, "
                  + "help text, usage or required permission")
     public String search(@CommandParam("searched") String searched) {
         String searchLowercase = searched.toLowerCase();
+
+        // finds commands, prefabs, blocks that match searched
```
Cervator: Comments should add something - this and comments like it do not :(

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -171,11 +183,21 @@ public String search(@CommandParam("searched") String searched) {
         return result;
     }
 
+    /**
+     * @param searchLowercase String
```
Cervator: Need a description of the method itself

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -309,11 +393,19 @@ public void destroyEntitiesUsingPrefab(@CommandParam("prefabName") String prefab
         }
     }
 
+    /**
+     * Exits the game
```
Cervator: This is another good example of javadoc vs description.

* Javadoc: Triggers a graceful shutdown of the game after the current frame, attempting to dispose all game resources
* Description: Closes the game client

engine/src/main/java/org/terasology/logic/actions/PlaySoundAction.java
```diff
@@ -31,6 +31,8 @@
 import org.terasology.utilities.random.Random;
 
 /**
+ * This class is responsible for managing audio in-game
```
gkaretka: Yep, absolutely right

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
gkaretka: well, I kinda like this description :D i'll use it

Cervator:
Nice job - merged!

PR #2710, opened 2016-12-27
Value comparison instead of reference comparison in StateChange.isDefaultInstance

### LINKED ISSUE #2706: Use value comparisons instead of reference comparisons in StateChange.isDefaultInstance implementations
Solve the original problem (i.e. removal of duplicate `SetViewportToSizeOf` state changes) #2688 highlighted.

### What actually happened
State changes which are value identical to the default state change are not being reported as such to `AbstractNode.addDesiredStateChange`, its only caller.

### How to reproduce

* Run the game, load a world.
* Check the logs for the generated render task list with `SetViewportToSizeOfTask` tasks which are identical to the default (reset) state change task (i.e. the `SetViewportToSizeOfTask` which is run _after_ the node's `process` task).

### Cause
This happens because `StateChange.isDefaultInstance` incorrectly returns `false` for all instances of a state change as it performs a _reference_  comparison instead of a value comparison.

cc @emanuele3d

eviltak:
### Contains
Fixes #2706. Instead of the existing reference based comparison, all `StateChange.isDefaultInstance` implementations now make use of value based comparison by calling the `equals` method instead.

### How to test
- Run the game, load a world.
- Check the logs for the render task list. Note the difference in number of generated tasks (114 from 123).

### How it works
All implementations of the `isDefaultInstance` method used reference comparisons using the equality operator (`this == defaultInstance`). This is modified to call the overridden `equals` method instead (`this.equals(defaultInstance)`).

### DIFF
--- a/engine-tests/src/test/java/org/terasology/rendering/dag/SetName.java
+++ b/engine-tests/src/test/java/org/terasology/rendering/dag/SetName.java
@@ -53,7 +53,7 @@ public class SetName implements StateChange {
 
     @Override
     public boolean isTheDefaultInstance() {
-        return this == defaultInstance;
+        return this.equals(defaultInstance);
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
@@ -63,7 +63,7 @@ public final class BindFBO implements FBOManagerSubscriber, StateChange {
 
     @Override
     public boolean isTheDefaultInstance() {
-        return this == defaultInstance;
+        return this.equals(defaultInstance);
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthMask.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthMask.java
@@ -56,7 +56,7 @@ public final class DisableDepthMask implements StateChange {
 
     @Override
     public boolean isTheDefaultInstance() {
-        return this == defaultInstance;
+        return this.equals(defaultInstance);
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthTest.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthTest.java
@@ -73,7 +73,7 @@ public final class DisableDepthTest extends SetStateParameter {
 
     @Override
     public boolean isTheDefaultInstance() {
-        return this == defaultInstance;
+        return this.equals(defaultInstance);
     }
 
 }
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableBlending.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableBlending.java
@@ -76,7 +76,7 @@ public final class EnableBlending extends SetStateParameter {
 
     @Override
     public boolean isTheDefaultInstance() {
-        return this == defaultInstance;
+        return this.equals(defaultInstance);
     }
 
 }
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableFaceCulling.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableFaceCulling.java
@@ -76,7 +76,7 @@ public final class EnableFaceCulling extends SetStateParameter {
 
     @Override
     public boolean isTheDefaultInstance() {
-        return this == defaultInstance;
+        return this.equals(defaultInstance);
     }
 
 }
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableMaterial.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableMaterial.java
@@ -71,7 +71,7 @@ public final class EnableMaterial implements StateChange {
 
     @Override
     public boolean isTheDefaultInstance() {
-        return this == defaultInstance;
+        return this.equals(defaultInstance);
     }
 
     private static Material getMaterial(String assetId) {
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableStencilTest.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableStencilTest.java
@@ -70,7 +70,7 @@ public final class EnableStencilTest extends SetStateParameter {
 
     @Override
     public boolean isTheDefaultInstance() {
-        return this == defaultInstance;
+        return this.equals(defaultInstance);
     }
 
 }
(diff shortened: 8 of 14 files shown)


GooeyHub:

Cervator:
Thanks :-) Will merge after testing completes.

@GooeyHub: ok to test

emanuele3d:
@Cervator: wait for merge please.

GooeyHub:

emanuele3d:
@eviltak, @indianajohn and @Cervator. 

The code changes in this PR are fairly straightforward. But I'd like to solicit your thoughts on a more... "philsophical" issue.

First, let me tell you about submarines. I know: what do submarines have to do with this? Well, they have compartments. And the sailors are trained to close them in case of flooding. If the whole submarine floods it sinks. If only some compartments flood, the submarine has at least a fighting chance.

In this PR @eviltak is kindly implementing a change I myself suggested. If a render node needs an opengl state change, say, disabling face culling, it shouldn't request one that is already the default. 

Currently there is no check to enforce this. A node can request and obtain a state change disabling face culling even though it is disabled by default. This PR addresses that.

However, as I thought about this, I thought of the following scenario.

Let's say an external module injects Node X into the renderer. Let's also say that an unrelated module, developed by somebody else, injects Node Y, which takes the output of Node X and does something with it. Let's now say that the developer of Node X has been naughty. Instead of setting face culling "properly", using an EnableFaceCulling instance, he did it the old way, with a direct opengl command: `glEnable(GL_CULL_FACE)`. Even naughtier, he didn't clean after his act and forgot to disable it again.

Node Y developer quickly realizes this, but has no influence on Node X code other than sending a politely cursing email to Node X developer. Node X developer is not fond on polite curses: he would have changed the code right away had he received some sound, creative, verbal smacking. Instead, he put the task in the pile of things to do and forgets it. Result Node X remains unchanged and every node after it has to cope with a parameter that might or might not be the default, depending if Node X is in use or not.

With the existing code Node Y and any other node following Node X can use a StateChange to reset an opengl parameter to the default, compartmentalizing the problem caused by Node X: not ideal but it would work within the philosophy of the new architecture. With the change suggested in this PR a node wouldn't be allowed to use a StateChange to cause a reset to default. Instead it would be forced to use a direct opengl call within its process() method, something the new architecture is trying to avoid as much as possible. 

So, while no node _should_ directly request a state change resetting a parameter to its default, it could be useful to leave that door open, to compartmentalize problems arising in other nodes, in a way that is consistent with the current architecture. 

What do you guys think?

Cervator:
Interesting metaphor and philosophy! I especially love that it is a submarine example when you're living in a landlocked mountain country :D Although I suppose there is at least one large lake.

It is a very thorough and good argument. From a high level project leadership type perspective I would consider it with a somewhat (surprising perhaps?) iron-fisted approach: if anybody leaves a poorly configured node anywhere in the engine or the Iota series of modules we simply fix it. Engine + Iota (which covers anything you might expect in a vanilla game like MC - so all the basics and maybe a few more advanced foundations for additional systems) are expected to be under community ownership and shared maintenance. So as long as we have anybody capable we'd fix mistakes like that (which we would if we were able to consider a hacky alternative to fix it in a different node instead).

For further out modules, say a whacky post-apocalyptic version of Alice in Wonderland where everything went even *more* wrong, an naughty author might sneak in errors in his or her fundamental nodes, which other authors then might in turn depend on while extending that world. At that point though I would say it would be on that author's quality of work, and I wouldn't necessarily throw options into the engine to give us options to deal with it. I'd rather we send the author a PR offering a fix, and if they ignore it, well, maybe that series of content just doesn't play as nice with the central community. And that's okay, really.

So I would keep it simple, and sharp. I am but one voice, however, and would invite additional feedback :-)

@eviltak while I won't merge this yet then I'll still approve it in the GCI system. You've clearly put plenty of thought into the system and @emanuele3d agreed on that. Thank you very much and please keep at it! 👍

eviltak:
I agree with what @Cervator said. If we keep adding safety checks and features while compromising efficiency or good practices for something that a future contributor _might_ do, we never will be able to stop. Besides, if someone wants to add a new `Node`, she must send a PR to to this repository if I'm not mistaken. So any and all such code smells will be caught by a reviewer.

emanuele3d:
Thank you @Cervator and @eviltak for your comments. 

A couple of clarifications: indeed I was only talking about nodes developed in the context of external modules. Nodes within the engine won't have that kind of problems as long as I'm the one nurturing the renderer's code. 

Also, it is true @eviltak that _right now_ adding a Node to the renderer implies submitting a PR to this repository. However the medium-term goal is to make an API available for modules to inject their own Nodes into the rendering process.

In any case, thank you for your comments. Let's merge this as it is then.

Thanks @eviltak for the work!

Cervator:
Hmm, interesting, on now merging this for final sign-off testing I oddly get 124 nodes both before and after the merge.

```
02:52:08.778 [main] INFO  o.t.r.dag.RenderTaskListGenerator - Task list generated in 10.617 ms
02:52:08.778 [main] INFO  o.t.r.dag.RenderTaskListGenerator - 33 nodes, 29 enabled - 124 tasks (excluding marker tasks) out of 169 potential tasks.
```

@eviltak could you double check that? Did another node sneak by or did something change the graph where the extra nodes this skipped are no longer skippable? Or something weird in my workspace?

@emanuele3d I think we'll be casting a greater net than just what's in the engine. Even outside the Iota module series I want to offer contributors/modders the option to keep their work within the Terasology module org, where we'd help maintain whatever if need be (and justified), along with offering builds, release management, etc. Naturally some people would want to do their own thing entirely, but I think our infrastructure and process could be a tempting safety net to trade for a little authority on fixing obvious issues :-)

(not that I want us obligated to fix anything outside Iota but this is pretty much the perfect example of what I could see us step in to fix if an author doesn't on their own)

eviltak:
@Cervator The only duplicates in the render task list (for now) are `SetViewportToSizeOf` instances. Since #2702 hasn't been merged yet, these duplicates aren't reported as value equal (see #2688) and hence aren't culled. To see a difference, _both_ #2710 and #2702 must be merged.

Cervator:
Ahh! Okay, so need both to be able to properly validate this one.

Or with this one and the related discussion should #2702 be merged as well?

(I'm done for tonight but will check back tomorrow)

Cervator:
Tested both PRs merged together. Worked as expected with the task number changing :-)

Re-merged this one solo and pushing it. Will bump the other PR to see what's next.

emanuele3d:
@Cervator - you wrote:
>  I think we'll be casting a greater net than just what's in the engine. Even outside the Iota module series I want to offer contributors/modders the option to keep their work within the Terasology module org, where we'd help maintain whatever if need be (and justified), along with offering builds, release management, etc. Naturally some people would want to do their own thing entirely, but I think our infrastructure and process could be a tempting safety net to trade for a little authority on fixing obvious issues :-)

I understand what you wrote here but I don't quite understand why you wrote it. Did I write something that is not quite fully aligned with this?

Cervator:
@emanuele3d nah :-) I think we're on the same page, just communicating via text in a language native to neither of us. That, and I write too much :D

PR #2714, opened 2016-12-27
Ui javadoc

jellysnake:
### Contains

Javadoc for all NUI Widgets

### How to test

Confirm no functional changes have been made and that the javadoc works fine


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UIBox.java
```diff
@@ -57,19 +57,39 @@ public void update(float delta) {
         }
     }
 
+    /**
+     * Gets the widget/layout inside the box.
```
rzats: The getter/setter descriptions are way too repetitive and the boolean ones ("gets if the contents..." / "sets if the contents...") are just weird.

A better solution is to simply remove the description everywhere:

```java
// getter
/**
 * @return Whether the content is updated at the same time as the box.
 */

// setter
/*
 * @param value Whether the content should be updated at the same time as the box.
 */
```

(If Checkstyle complains about this, make it shut up by setting `allowMissingPropertyJavadoc` to `false` in [our config](https://github.com/MovingBlocks/TeraConfig))

engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
```diff
@@ -20,11 +20,7 @@
 import org.terasology.math.geom.Rect2i;
 import org.terasology.math.geom.Vector2i;
 import org.terasology.rendering.assets.font.Font;
-import org.terasology.rendering.nui.BaseInteractionListener;
-import org.terasology.rendering.nui.Canvas;
-import org.terasology.rendering.nui.CoreWidget;
-import org.terasology.rendering.nui.InteractionListener;
-import org.terasology.rendering.nui.SubRegion;
+import org.terasology.rendering.nui.*;
```
rzats: No wildcard imports, please! (Set min. # of imports before wildcards are used to 999 in IDEA settings)

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImage.java
```diff
@@ -25,8 +25,6 @@
 import org.terasology.rendering.nui.databinding.Binding;
 import org.terasology.rendering.nui.databinding.DefaultBinding;
 
-/**
```
rzats: Should be filled in (along with the other class-level Javadocs)

GooeyHub:

PR #2723, opened 2017-01-01
Documentation of Code (GCI)

TheJYKoder:


engine/src/main/java/org/terasology/config/PlayerConfig.java
```diff
@@ -25,66 +25,126 @@
...
+     */ 
     public void setHasEnteredUsername(boolean entered) {
         this.hasEnteredUsername = entered;
     }
 
+    /**
+     * Returns a String representing the player's default name
+     *
+     * @return A String representing the player's default name
+     */
```
oniatus: In this case, i would like to see a different javadoc. 
I know that the method returns the players default name by reading the method name. However i don't know, that the default name will be "Player", followed by a 5-digit random number.

engine/src/main/java/org/terasology/config/AudioConfig.java
```diff
@@ -20,49 +20,89 @@
...
+    private float musicVolume;          // The music volume
+    private boolean disableSound;       // The boolean property of whether sound is disabled or not
 
     private transient PropertyChangeSupport propertyChangeSupport = new PropertyChangeSupport(this);
-
+    /**
+     * Returns the current sound volume
+     *
+     * @return a float representing the current sound volume
+     */
```
oniatus: Add new information if you document a method.
the method signature already explains, that this returns the sound volume, also that this will be a float.
What would be interesting to know, is the value range of the float (yes, this requires code-digging ;) ).

engine/src/main/java/org/terasology/config/AudioConfig.java
```diff
@@ -20,49 +20,89 @@
...
+ * For configuring the audio of the game
  */
 public class AudioConfig {
     public static final String SOUND_VOLUME = "soundVolume";
     public static final String MUSIC_VOLUME = "musicVolume";
 
-    private float soundVolume;
-    private float musicVolume;
-    private boolean disableSound;
+    private float soundVolume;          // The sound volume
```
oniatus: No need to comment all private fields.
In some cases it makes sense to explain a field but `soundVolume` is pretty self-explanatory right? :)

engine/src/main/java/org/terasology/logic/health/DestroyEvent.java
```diff
@@ -21,27 +21,48 @@
...
+     * @return an EntityRef object which is the instigator of the destruction
+    */
     public EntityRef getInstigator() {
         return instigator;
     }
-
+    
+    /**
+     * Returns the instigator data of the DestroyEvent object
+     *
```
oniatus: also take care of copy-pasting javadoc :)

TheJYKoder:
My apologies, I will redo this task. It did seem weird though when I was attempting to write Javadocs, since I thought to myself "Isn't this self-explanatory?"

TheJYKoder:
I will resubmit via different PR

PR #2733, opened 2017-01-04
Added hindi language

VaibhavBajaj:
## Purpose
Added Hindi language to the list of compatible languages.

**Note:** The flag has a different extension (flag_in.png) from the lang file (menu_hi.lang) because that is what is specified in the link:
[List of ISO 639-1 Codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
[FamFamFam flag pack ISO 3166-1 alpha-2 country codes](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#IN)

## Testing
Unfortunately, the code can't be tested until this Devanagari script is included since Hindi has a couple hundred unique computer characters.
Link: https://www.google.com/get/noto/#sans-deva
After that is done, testing can be done by opening Terasology and going to:
Settings -> Player -> Language -> हिंदी (Hindi)


GooeyHub:

Cervator:
Nice work! As discussed I'll approve this for GCI early and see if we can figure out what to do about the font situation soonish :-)

@GooeyHub ok to test

GooeyHub:

msteiger:
I've added 128 Devanagari glyphs plus all 28 from the Devanagari Extended region. It looks like some of them are used for composition, so it doesn't look quite right yet:

image

@VaibhavBajaj can you spot the problem here?

VaibhavBajaj:
As far as I can see, it has not been calibrated properly, so all the upper characters (we call them mathra) are one step ahead of all lower characters. Could I please see the code that has been added?

VaibhavBajaj:
According to this: There are over 200 charachters. I can confirm that seeing these: 
[http://indiatyping.com/index.php/typing-tips/hindi-code](http://indiatyping.com/index.php/typing-tips/hindi-code)
Might want to check it out.

emanuele3d:
Regardless of the problems, it is wonderful to see such a beautiful script in Terasology.

VaibhavBajaj:
Thanks :P

Cervator:
While reviewing #2737 and #2739 the saga deepened 😕 

Dutch is easy - nl & nl. Indonesia came in as id & id. No flag icon working in-game. Huh. Flag has to be "in". Yet India in the flag pack is "in" (and Indonesia "id"), but really it needs to be "hi", and certainly not "id" - that's confusing :D

When we arrive at merging this we might have to rename the flag icon here from "id" to "hi" (looks like the flag is missing in @msteiger's test, even) - but honestly I'd have to wait and see what our internationalization system comes up with when running the game to be sure.

VaibhavBajaj:
@Cervator You do mean "in" to "hi", don't you? Got a bit confused in all that, :)? Well, I just hope that someday I will see Hindi in Terasology, the same Hindi that I spent 8 sweating hours translating, :P.

Cervator:
Er, yeah. Maybe. I think. I don't know anymore!?!? 😕 😕 😕 😕 😕 😕

msteiger:
Sorry for the delay - here's the commit: https://github.com/MovingBlocks/Terasology/commit/93f85ebf973fdfe50e74f568639a283cad4e0f45

I did not change code, just added the glyphs. The chars are expected to be 16bit only. Is that the case here? The glyphs are rendered one by one as specified in the file. Maybe that commit helps you find out - I don't know what it's supposed to look like, but I'm happy to assist.

What Unicode ranges are you referring to? Is there more than "Devangari" and "Devangari Extended" or does the Noto font not support all glyphs?

VaibhavBajaj:
Since I am completely new to this, I don't think I have much hope of spotting the error. Sorry, sir.

However, could you humor me so that I may understand this better? I am confused. When I downloaded Noto font, I only saw 40 or so characters. Here I see you have added hundreds. Also, is there a way I can download the .png file or is it self-generated?

VaibhavBajaj:
I have viewed the hindi characters:
Characters
I now only have a few questions:
1. Is it possible some characters were cut off screen like the | charachter. I cannot spot it anywhere and that may be why some of the characters are misplaced. 
2. Could it be that the characters have different codes in Noto and thus, some characters are messed up. I would love to see more .png files like the one you shared in the topmost post but from what I can see, I would say the EE(ी) character and the AA ( ा) character have been messed up.

msteiger:
About the flags: we might need a (bi-directional) mapping from language to country such as Hindi <-> India.

I'm using BMFont to generate those images. You can download the installer and the config files we use here: https://github.com/MovingBlocks/TeraMisc/tree/master/fonts

For Noto Sans Devanagari, I see the following glyphs:
image

This seems to be in line with the Unicode standard: https://en.wikipedia.org/wiki/Devanagari_(Unicode_block)

Afaics, the issue is with the dotted circle (a "Spacing Combining Mark"). Maybe we can construct a simple example consisting of 2-3 glyphs? Maybe reversing the order of glyphs already solves the issue.

msteiger:
@VaibhavBajaj as I see it, the translation for "Hindi" has 5 chars. The f-shaped one is the second, but apparently it should be put before the first one. Is that correct? I wonder where the offsets are stored in the font ..

VaibhavBajaj:
@msteiger Yep That's about correct. Here they are on your BMFont chart (x, y):
1. ह
2. ि
3. Half of न as in न्द (This can also be replaced by a dot on top of the second alphabet)
4. द
5. ी
हिन्दी

VaibhavBajaj:
@msteiger Have you come across any new discoveries. I just wanted to let you know that I was sticking around after GCI as well. So, if I can help in any way, please ping me. 👍

msteiger:
Yes, I tried Hiero, but it even fails to render the word "Hindi" in Devanagari typescript. We need another tool to generate the textures properly.

VaibhavBajaj:
@msteiger I will try my best to start understanding this code's true working. Then I would be able to explain why the print is failing.

For starters, could you tell me what Hiero is? Also, just so you could experiment, I have added a Devanagari translator that you can fool around with to see exactly how the Devanagari is written.
Link: http://www.lexilogos.com/keyboard/devanagari.htm

Also, here are some guidelines on using it:
1. The main thing I want you to see is how upper characters are joined to the lower characters. 
2. For starters, type: h i n d I (Cap-sensitive Last char is Cap i). 
3. Next you can try using the letters: a A e E o O etc. to see how words are NORMALLY written when writing Devanagari. 

I have come to think that instead of putting the upper character above the one right below it, the current program might be putting it AFTER the current character which is causing issues. Check it out whenever free.
Meanwhile, I will try understanding the exact working of the language system.

msteiger:
That's a great tool. I still haven't figured out completely how the characters are composed. We're using [BMFont](http://www.angelcode.com/products/bmfont/) to create textures from typescript glyphs. This step already fails as it renders the dotted circles as well. The website lists a few alternative tools and one of them is Hiero. Unfortunately, it doesn't work either. Maybe you can find one that does work?

VaibhavBajaj:
@msteiger I believe rendering is the only thing failing us here. Is there a way that instead of loading the text as, well, text, we could load it instead as an image of some sort?

If not, I guess we will have to keep searching for tools. What is the name of the website you mentioned?

nihal111:
@msteiger I tried using the BMFont tool. Need some help on how to use it. From what it looks to me, I suppose it converts a ttf to bmp, however when I import a ttf file from the "Font Settings" nothing happens.  How did you generate those Devanagari glyphs?
Also, the problem with the dotted circles being rendered can be solved if somehow we can get all character combinations with all the different 'mathras' (i.e ड़ी की are stored separately, even though both have a common ी character). Is this possible?

VaibhavBajaj:
@nihal111 You mean combining each of the normal characters with each of the mathras and storing it? Well. That is a possibility but it will increase the number of characters to around 600 or so. Probably more. What software would you use? Thanks for the suggestion.

@msteiger Could you please answer my earlier post? Thanks.

nihal111:
No of characters wouldn't be a restriction, would it? :)
I'm wondering if it's possible using BMFont.

Cervator:
Number of characters probably wouldn't be terrible with 3 digits of combinations. 4 or 5 digits ... that gets a little crazier :-)

Also, dunno why this got closed and the branch deleted. We need to figure it out sooner or later, good research here.

msteiger:
Sorry, I didn't see the question in the upper post. The website is this one: http://www.angelcode.com/products/bmfont/

I suppose that we could render all pairs, but this would require special treatment for this typescript (and maybe others). We probably need to incorporate the offsets, but it would be good if it was consistent across all languages somehow.

What I usually do to add a new typescript font is this: create it using BMFont and merge it into the NotoSans-Regular.fnt definition file that is currently used as a default for the menu. You might need to adjust page number and glyph count.

Cervator:
For reference this commit has been merged now :-)

Fixed up the language file and verified its entry shows up in-game, although without symbols due to the known issue. Then I merged #2784 on top to promptly hide it in-game until we resolve said symbol issue.

@VaibhavBajaj thanks again! I look forward to when we find a solution that'll let Hindi show in-game.

PR #2736, opened 2017-01-06
Added death screen instigator and death screen damageType

VaibhavBajaj:
## Purpose

Added Death Screen DamageType for any event. Also added DeathScreen instigator for any death where player is killed by other AI or player.

## Testing
You can test DamageType by simply falling from a great height. Instigator can be tested by being killed from Skeleton in GooeysQuests.

PR #2738, opened 2017-01-06
Dying in detail is hard - fixes on top of #2736

Cervator:
NOT FOR MERGING!

This enhances #2736 by @VaibhavBajaj by using the `ClientInfoComponent" to store death info instead of the `CharacterComponent` as the character is dead and gone by the time the `DeathScreen` tries to retrieve the info for display.

However, I'm totally not sure this is the right way to do it - might be. But it is also a mess since I was updating in pieces and just kinda threw fixes on top. So at the least it should be cleaned up and resubmitted.

Curious if anybody else has any ideas of the correctness of this approach. Primary commit with the final fixes is https://github.com/MovingBlocks/Terasology/commit/4d4f5a691dbefde088459bb324c0ea13f2a8b63f


iojw:
@Cervator: If you do go with this approach, here is an example of how you could bind the data in `DeathScreen.java` as discussed on IRC earlier: https://hastebin.com/unemicanid.java. This also replaces the `CoreRegistry` usage with `@In` 🙂

VaibhavBajaj:
@iojw Works great. @Cervator Is this considered any better. I tested iojw's code and it works fine. 

## Testing
Enable potions module and Gooeys Quests.
1. Teleport to 0 80 0. Death due to Physical Damage.
2. Get harmPotions. Death due to Direct Damage.
3. Get hemlockPotions. Death due to Poison Damage.
4. Die by skeleton. Death due to Physical Damage.

jellysnake:
@VaibhavBajaj @Cervator
Would a cleaner approach not to simply listen for the DoDamage event in one class and then when the DeathScreen is shown simply grab the last damage type from this class and display it. I've made a bare bones example [here](https://github.com/jellysnake/Terasology/tree/ExampleDeathDying).

VaibhavBajaj:
@jellysnake That is much like the code I created to start with. The problem is not in the fact that the damageType is not stored in the code. If I were to log my damage type in the first code I created, I would have seen all the different kinds. The problem is that the code does not sync with the UI file. Even if you were to complete this code and run it, the UI file would only sync with the code once.

Thanks for the attempt. However, this does not solve the problem this thread was created to solve.

Cervator:
Closing this since it has been consumed back into #2736 - will aim to merge that in the near future.

PR #2745, opened 2017-01-08
Korean Language

TheJYKoder:


GooeyHub:

Cervator:
Drat. This is in the same boat as #2733 and needs new font files included as well.

Also, the `"this-language-English"` entry needs to include the name of the *target* language, in English. So "Korean" :-) Since the language gets listed in both native and English in the drop-down.

Will let this PR sit around until we get a change to add more font files. Maybe @msteiger can find those too :D

msteiger:
The glyphs should already be present in the CJK (Chinese, Japanese, Korean) png file. So it might work already?

msteiger:
I just checked myself: the current collection does not cover Hangul. Do we need to cover the entire range of 11'184 glyphs or is there a reasonably small subset?

https://en.wikipedia.org/wiki/Hangul_Syllables

emanuele3d:
:scream:

msteiger:
@TheJYKoder do you think all glyphs that are shown on the Wikipedia page are needed?

TheJYKoder:
@msteiger Nope, the Korean I used is relatively simple, so characters like 걞 and 걟

msteiger:
Can you specify a range of glyphs that we should cover? Alternatively, would it make sense to use Hangul Jamo instead?

https://en.wikipedia.org/wiki/Hangul_Jamo_(Unicode_block)

TheJYKoder:
There is no "range" of glyphs in all honesty. The language uses all the characters, but it's just that I used simpler ones. However, if Koreans were to type in the chat, I would say that the first Wikipedia page is necessary. The reason being that Korean is based on sound, and depending on sound, any of the glyphs can be used.

msteiger:
Too bad. It takes 44 textures (512x512) to cover the entire range of glyphs. Maybe we need to load or even create them on-demand.

Cervator:
How about we find a way to allow modules to define fonts, much like how we moved out the large music files?

The language settings page could scan for modules identifying themselves somehow as including a language

Admittedly this would require module support within the main menu, which has been raised as a desired feature from time to time. 

Might be a good thing to pair with a rework of world preview (also goes looking for modules) in a GSOC level item

Cervator:
Included an item for some of the stuff in my previous comment at http://forum.terasology.org/threads/gsoc-2017-ideas-phase-dev-meeting-sat-jan-21st.1770 - although some of that is fairly long term.

msteiger:
I agree that we might want to add support for non-default content such as music files or Asian typescripts. 
The Hangul textures are pretty large unfortunately. We should investigate optional extra content before merging this.

I'll look into this some more..

Sidenote: Cleaning the git repo can be done with filter-branch or this tool: https://rtyley.github.io/bfg-repo-cleaner/

Cervator:
> Sidenote: Cleaning the git repo can be done with filter-branch or this tool: https://rtyley.github.io/bfg-repo-cleaner/

Unfortunately that includes rewriting history, which is a big no-no in public since it'll put all existing clones into a bad state :( May also cloud up places where we have commits saved/references in issues and PRs. I'm not sure. Kinda scary.

That's much more suitable for a private/enterprise type Git setting, where you have a small set of well known workers and can force-push with some certainty that you can reach all users that matter and support them if needed :-)

I did put an item in my [summary of possible new GSOC ideas](http://forum.terasology.org/threads/gsoc-2017-ideas-phase-dev-meeting-sat-jan-21st.1770/) that would cover some of the module magic that would allow this to work better - but not any time soon, alas :-)

msteiger:
Valid point, @Cervator. No easy solution then, I guess.

Maybe we can setup a module for Korean language support. For now, we can at least use it for in-game chat. A bit limited, but using it later at "menu time" is then less effort.

Cervator:
Merged this together with #2784 so the language file is present, but excluded from the language drop-down until we find a solution :-)

Can still test by just commenting out the exclusion in `PlayerSettingsCreen.java`

Thanks again @TheJYKoder !

PR #2748, opened 2017-01-09
Tracked controller

indianajohn:
### Contains

Support for a single tracked controller in OpenVR. In the location of your tracked controller, your equipped tool model will show up. For the moment, this is fun but useless. Shortly, I will make it so that the tool's action affects the block that the tool is colliding with, allowing for mining with the tracked motion controller.

### How to test

Ideally, start up your game with vrSupport enabled in `config.cfg` and turn on your controller. However, since VR is not officially supported yet, I don't think there's a need for anybody to test that.

However, it would be good to test that I haven't broken anything by starting the game in normal mode and using your tools, and confirming they act normally.

In my understanding, I should be able to put the lines:

```
@In
OpenVRProvider vrProvider;
```

in `FirstPersonHeldItemMountPointComponent.java`, and it should give me the instance created in `WorldRendererImpl.java` (and subsequently `put()` into the context). However, if I do that, the reference following `@In` is null. Due to the lack of other options here, I put a public static instance in the `WorldRenderer` interface. I am aware how ugly that is, and I am open to other options on how to get that done.


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -33,6 +34,13 @@
  * the update and render methods as they are central to a rendering implementation.
  */
 public interface WorldRenderer {
+    // This is exposed so that controller-related functions can create callbacks.
+    // Ideally, these classes would be able to do so with dependency injection, but that
+    // is broken for at least FirstPersonHeldMountPointComponent.java. Could be reltaed to:
+    // https://github.com/MovingBlocks/Terasology/issues/2336A
+    // TODO: take this out and do dependency injection properly for any classes that depend on this being exposed.
+    public OpenVRProvider VR_PROVIDER = new OpenVRProvider();
```
emanuele3d: Yes, I confirm this is pretty ugly. :smile: 

So, in FirstPersonHeldItemMountPointComponent you can get the WorldRender via @In but not the OpenVRProvider?

engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
```diff
@@ -15,18 +15,66 @@
...
+import org.terasology.registry.In;
+import org.terasology.rendering.openvrprovider.ControllerListener;
+import org.terasology.rendering.world.WorldRenderer;
+import org.terasology.math.geom.Quat4f;
 
 /**
  * Only used by the client side so that held items can be positioned in line with the camera
  */
-public class FirstPersonHeldItemMountPointComponent implements Component {
+public class FirstPersonHeldItemMountPointComponent implements Component, ControllerListener {
```
emanuele3d: Tagging @Josharias and @skaldarnar to review changes to this class as they are much more knowledgeable than I am on ES stuff, i.e. components.

emanuele3d:
Apart from the known problem we'll be trying to address the rest of the code per se seems ok to me.

Architecturally, what concerns me more is the way VR code is getting intertwined again with non-VR code, similar to what we had with the Oculus VR code that you @indianajohn had to painstakingly eradicate before you could proceed with the OpenVR work. 

Granted, so far the extent of the VR "intrusions" is much smaller as much of the VR code is neatly contained in its own package. But let's keep in mind the potential long-term goal of everything VR sitting in its own module so that people without VR hardware don't have their software bloated by it. 

With that goal in mind I'm wondering if the changes to FirstPersonClientSystem and FirstPersonHeldItemMountPointComponent will ever be extractable. Should perhaps VR offer replacement classes instead of adding code to the existing ones? I.e. a VRcapableFirstPersonClientSystem, perhaps simply extending FirstPersonClientSystem, kicking-in if VR is available?

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
indianajohn: It's worse than that; I can get neither. The reference to `WorldRenderer` is null in this case. Here a bit of my Java ignorance is shining through, because,

* I'm sure that my public static instance to `VR_PROVIDER` is the same instance (otherwise the callback that I registered would not be hit), but
* I'm not 100% what that object's lifecycle is.

The best I can tell, 'WorldRendererImpl' is instantiated and VRProvider is instantiated for the first time when `WorldRendererImpl` accesses the instance, or when the program is first run; I'm not real sure on that part. Then that instance is accessible anywhere via `WorldRenderer.VR_PROVIDER`. Then, as long as `WorldRenderer.VR_PROVIDER` is reachable by some thread, that instance escapes garbage collection. I guess the exact same effect can be achieved with something like:

```
import org.terasology.rendering.world.WorldRenderer.
...
VR_PROVIDER; VR_PROVIDER.addControllerListener();
```

The code ended up how it did just because I was trying to get an `@In` reference to work and that's when I gave up. Anyway, that's a long way of saying that the code as it is written is very confusing and possibly a future minefield, and it would be good if we could find an alternative. Any ideas, anyone, on:

* how I can get dependency injection to work, or
* if there is an alternative?

indianajohn:
So I considered your suggestion about making another `FirstPersonMountPointComponent` that works for VR, and switched on/off based on config options, and began implementing it only to find that I would have to introduce references to `OpenVRProvider` into several different classes. While it might be a bit more organized, who it's written now seems to touch fewer parts of the engine.

Anyway, point taken about trying to keep things more-or-less isolated inside VR-related packages. That's a principal I'll be extra sure to follow as I make more changes.

GooeyHub:

GooeyHub:

indianajohn:
Okay, as you can see, I've replaced my workaround with a slightly more palatable one. We should be able to delete one file and replace 4 lines when we're able to properly inject the dependency.

Also, thinking back on it, I guess it would be possible to have a separate class that extends `FirstPersonHeldMountPointComponent`.

GooeyHub:

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProviderSingleton.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+package org.terasology.rendering.openvrprovider;
+
+/**
+ * Created by john on 1/9/17.
+ */
+public class OpenVRProviderSingleton {
+    public static OpenVRProvider vrProvider = new OpenVRProvider();
+}
```
emanuele3d: Hmmm... I know what you are trying to do but this isn't a singleton. Anybody with access to OpenBRProviderSingleton can replace the content of vrProvider with a new provider!

I'm guessing you could mitigate this by adding the final keyword but a minimally modified version of the classical singleton java pattern is this:

```
public final class ClassicSingleton {
   private static ClassicSingleton instance = null;
   private ClassicSingleton() {
      // Exists only to defeat instantiation.
   }
   public static ClassicSingleton getInstance() {
      if(instance == null) {
         instance = new ClassicSingleton();
      }
      return instance;
   }
}
```

That been said, can't you make OpenVRProvider a singleton of it own?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -114,6 +115,9 @@
     private final ShaderManager shaderManager;
     private final Camera playerCamera;
 
+    // TODO: @In
+    private final OpenVRProvider vrProvider = OpenVRProviderSingleton.vrProvider;
```
emanuele3d: A pattern I have noticed in Terasology's codebase is to initialize variables not on declaration (which might happen much sooner than instantiation) but on construction or initialization. I'd recommend that also in this case.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProviderSingleton.java
emanuele3d: P.S. some interesting consideration on this deceptively simple pattern can be found here:

http://www.javaworld.com/article/2073352/core-java/simply-singleton.html

indianajohn:
Alright, I took your suggestion and made VRProvider have its own `getInstance()` method. After that, I think it actually looks fine to use the `getInstance()` rather than `@In` keyword; in the spirit of keeping OpenVRProvider fairly independent, it makes sense to put that there as the prescribed usage pattern in the event that OpenVRProvider were used in another program.

Is there anything outstanding before we merge this?

GooeyHub:

GooeyHub:

indianajohn:
I should also mention that I'm in the middle of writing a new subsystem for OpenVR. When I do that, I will change `FirstPersonHeldItemMountPointComponent` back to its nearly virgin state, aside from the quaternion component. Then I'll have my controller listener handle the updating of the mount point. Having this subsystem will hopefully allow me to keep all of the stuff connecting Terasology to OpenVR away from the rest of the engine.

Cervator:
@indianajohn: Sounds good :-)

With GCI ending I'm finally starting to catch up with engine PRs again! And I now have a nice new PC to test my Oculus and other stuff with, and can mess with my graphics drivers on the old one without worrying I'll put myself out of commission for a few days :D

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -70,6 +70,19 @@
     private final VRTextureBounds_t texBounds = new VRTextureBounds_t();
     private float nearClip = 0.5f;
     private float farClip = 500.0f;
+    private static OpenVRProvider instance = null;
```
emanuele3d: CheckStyle should have caught this: on declaration this variable is null anyway. This line can just be:

`private static OpenVRProvider instance;`

engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
```diff
@@ -40,7 +39,7 @@
     public float scale = 1f;
 
     // TODO: @In
-    private OpenVRProvider vrProvider = OpenVRProviderSingleton.vrProvider;
+    private final OpenVRProvider vrProvider = OpenVRProvider.getInstance();
```
emanuele3d: I must confess I am not well versed in the arcane arts of java timelines, i.e. what gets instantiated when. 

But I think by filling the vrProvider variable on declaration you might be forcing the OpenVRProvider instance to exist when the FirstPersonHeldItemMountPointComponent _class_ is loaded instead of waiting until an instance of the component is created. Tagging @flo for confirmation.

I'm guessing this might make a huge difference in this particular case but I'd create a default constructor and I'd initialize vrProvider in there.

emanuele3d:
Just a couple of minor comments - one on an outdated commit but still relevant. 

Nearly there!

indianajohn:
So you were referring to this I guess:

```
public final class ClassicSingleton {
   private static ClassicSingleton instance = null;
   private ClassicSingleton() {
      // Exists only to defeat instantiation.
   }
   public static ClassicSingleton getInstance() {
      if(instance == null) {
         instance = new ClassicSingleton();
      }
      return instance;
   }
}
```
With the last commit, I think the class fits that pattern exactly.

GooeyHub:

indianajohn:
I'm working out some bugs, but I've got full trackpad movement and interaction with the controllers (selecting the block to hit using motion tracking) pretty much up and running, based on this commit, in a different branch:

https://github.com/indianajohn/Terasology/tree/controller_buttons

Movement is pretty nausea-inducing, though, so I'll be implementing teleportation too. But it's nice to have trackpad movement for small movements, such as when mining.

Cervator:
Heya @indianajohn - curiously, and I probably should've asked this earlier, when you say "tracked controller" what could that actually mean? I take it in your case it is one of the hand-held Vive controllers?

We had somebody express interest in the forum in reviving support for the Leap Motion controller, and I was curious if this would aid that at all, or if it is conceptually too different (a controller that can provider its location to the input system vs a controller that can parse what your hands are doing)

http://forum.terasology.org/threads/reviving-leap-motion.1763/

I also just wrote up an initial thread on GSOC 2017 where I wonder if there's enough VR / Motion Controller potential to make up an item, probably focused on in-game UIs and/or related content.

http://forum.terasology.org/threads/gsoc-2017-ideas-phase-dev-meeting-sat-jan-21st.1770/

Am curious what you would think, and whether you'd consider mentoring such an item if we can write one up well. Planning to run a GSOC prep day / dev meeting tomorrow (starting in about 12 hours from now - all day Saturday my time)

Also re-poking @Josharias over the held item thing and @tdgunes for good measure on VR stuff, also wondering if we should consider any of the OSVR controller stuff for anything (come join us tomorrow if you can, both!)

indianajohn:
By "tracked controller(s)", I refer to OpenVR controller support. This means that the SteamVR / HTC Vive position / orientation tracking system works with the game. I can stand in the middle of my living room with my motion tracked controllers, look down in my headset, and see the tool in the same position the controller would be in the real world as long as I stay inside of the tracking volume covered by the lighthouses (infrared emitters).

I'm not hugely familiar with OSVR, but from what I've read the API will support room scale. In that case, the work I'm doing in these pull requests will be conceptually identical.

Leap Motion is a different problem. I'm unaware of any API support for Leap Motion at the moment in the two main VR libraries, and room scale VR depends on absolute tracking of objects. So while their API could probably be used to trigger certain actions based on hand movements, all of that would not be visible inside of the VR world because registering hands to their location with respect to the Vive tracking system is a nontrivial problem.

Sorry for not showing up for the meeting, but if you guys are thinking of writing some proposals for a VR user interface that would definitely tie into my work. At the moment, with my other PR and this one, it's possible to move your tool around the world, target objects, and trigger actions in-game; this is the core of any VR experience, so I think that's actually pretty good progress. However, there's still plenty of work to be done.

* Teleportation-based locomotion.
* Other sorts of locomotion - climbing? Skiing? There's lots of thoughts on this. This video is pro-watch: https://www.youtube.com/watch?v=p0YxzgQG2-E
* The creation of a canvas renderer that allows menu usage through VR.
* Archery - using the motion controllers to pick up an arrow, draw it into the bow, pull back, and release. It works very well in many VR games and is very cool.
* Melee combat. See https://www.youtube.com/watch?v=Grzo18EFfJM

If all of the above worked it would definitely attract a bit of attention among people with headsets.

I might be able to help out with GSOC if it didn't involve a huge time commitment. As you have probably noticed I disappear for days at a time due to my job. :)

engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
```diff
@@ -234,13 +234,22 @@ public void update(float delta) {
...
+                    TeraMath.DEG_TO_RAD * (mountPointComponent.rotateDegrees.x + addPitch),
+                    TeraMath.DEG_TO_RAD * mountPointComponent.rotateDegrees.z));
+            Vector3f offset = new Vector3f(0.25f * animateAmount, -0.12f * animateAmount, 0f);
+            offset.add(mountPointComponent.translate);
+            locationComponent.setLocalPosition(offset);
+        } else {
+            // In the case that we have obtained the quaternion directly, just set it.
+            // Receiving a quaternion directly is also signalling to us that we received
+            // it from OpenVR, so we omit the animation (since we want the item to track
+            // the player's hand.
```
emanuele3d: Good comment!

engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
emanuele3d: Looks like @Josharias and @skaldarnar are unavailable. @Cervator: anybody else who can take a look at this to validate the ES-side of things?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -63,6 +63,7 @@
     //keyboard
     private static boolean keyboardShowing;
     private static boolean headIsTracking;
+    private static OpenVRProvider instance;
 
     public OpenVRState vrState = new OpenVRState();
```
emanuele3d: Whoa. I almost missed this one. How come this is public? Shouldn't it be readonly via an accessor and perhaps even final?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -71,7 +72,24 @@
...
+    * As a general rule, we should use this class as a singleton, because multiple instantiation
+    * will likely cause problems in the upstream native library. This provides a convenient method
+    * of using OpenVRProvider as a singleton.
+     */
+    public static OpenVRProvider getInstance() {
+        if (instance == null) {
+            instance = new OpenVRProvider();
+        }
+        return instance;
+    }
```
emanuele3d: :+1:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -161,11 +163,11 @@ public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
...
         if (renderingConfig.isVrSupport()) {
-            this.vrProvider = new OpenVRProvider();
             context.put(OpenVRProvider.class, vrProvider);
-            if (this.vrProvider.init()) {
-                playerCamera = new OpenVRStereoCamera(this.vrProvider);
+            if (vrProvider.init()) {
+                playerCamera = new OpenVRStereoCamera(vrProvider);
                 currentRenderingStage = RenderingStage.LEFT_EYE;
             } else {
                 playerCamera = new PerspectiveCamera(renderingConfig.getCameraSettings());
```
emanuele3d: Let's add a comment here to make more explicit what's happening here, i.e.:
"If the vrProvider fails to initialize we assume it is unsupported or there is a problem with the hardware and we fall back on a standard, mono camera."

Feel free to rephrase/reword as needed.

emanuele3d:
Regarding your last post, thank you @indianajohn for the links to the videos. I particularly appreciated the review of the various movement methods currently being in use or being tested for VR.

I'd recommend you open a thread on the forum regarding this matter, to get an impression of what people think would be 1-2 best methods to implement and their specific flavors. I.e. I think simple teleportation could be a first step, followed by the capability to set the orientation after the jump. But others might have different ideas and there are gameplay details to keep into consideration, i.e. a straightforward teleportation would be considerably faster than walking, affording a big advantage to teleporters.

I am puzzled by your mention to a Canvas Renderer. What do you mean by that? Again, you might want to either open a thread in the forum specifically for this or even an issue here on github.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
indianajohn: There's a comment in there now, but it seems to not have gotten picked up by Github since it's one line above this one.

GooeyHub:

indianajohn:
Re: CanvasRenderer. The easiest way to get something 100% playable as soon as possible is going to be to make a very simple system in OpenVRProvider that renders to a Framebuffer and displays that Framebuffer as a plane in the room scale space. The controller will be captured and a ray will pass between the front of the controller and the plane, and that will be the "mouse".

By creating a new CanvasRenderer implementation that renders to this surface, we could allow the ability to interact with menus from inside the VR headset. That would fit very neatly inside of the subsystem that I am writing for controller support on my other branch, and it seems to fit with the overall structure of the code the best.

At some later point it might make sense to get a little more sophisticated with the menu system - some people have made objects that you interact with in the game world serve as a menu system, which is very cool / intuitive - but I think the above is a fairly easy change and makes good sense as a first step.

I'll register for the forums and make a post summarizing some design aspects for the VR system pretty soon.

emanuele3d:
@indianajohn: can you open an issue specifically on the matter of in-game UI for VR? Looks like it will require some more discussion and I can see this PR is going to land soon.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -63,15 +63,37 @@
...
+    // Call getInstance() to instantiate
+    private OpenVRProvider() {
+    }
+
+    // Get a singleton instance.
+    /*
+    * As a general rule, we should use this class as a singleton, because multiple instantiation
+    * will likely cause problems in the upstream native library. This provides a convenient method
+    * of using OpenVRProvider as a singleton.
+     */
```
emanuele3d: Can you cleanup this javadoc? The comment above it should be integrated in the javadoc and the javadoc should be in the form:

```java
/**
 * profoundly meaningful and masterfully written documentation
 */
```
Notice the double asterisks.

engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
```diff
@@ -15,18 +15,62 @@
...
+    // The hand/tool models seem to have an origin other than the pivot point. This is a best-effort correction,
+    // in the form of a 4x4 homogeneous transformation matrix
+    private Matrix4f toolAdjustmentMatrix = new Matrix4f(
+            1.0f, 0.0f, 0.0f, 0.0f,
+            0.0f, (float) Math.cos(230.* TeraMath.DEG_TO_RAD), (float) Math.sin(230. * TeraMath.DEG_TO_RAD), 0.0f,
+            0.0f, (float) -Math.sin(230. * TeraMath.DEG_TO_RAD), (float) Math.cos(230. * TeraMath.DEG_TO_RAD), 0.0f,
+            0.0f, -0.05f, -0.2f, 1.0f
+    );
+
+    public void trySubscribeToControllerPoses() {
```
emanuele3d: I think the code can be considered stable enough now. Can you please write some javadocs for the public methods of this class?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: Yep, I can see it. It's good, thank you.

indianajohn:
I started writing Javadocs and just kept writing them so you get some bonus Javadocs for OpenVRProvider et al. I'll write the issue up too.

GooeyHub:

Cervator:
@indianajohn on the GSOC bit you wouldn't need to be constantly available every day to help mentor. I'd expect 2-3 mentors per student so there is a fair overlap. You've got a good overview of the VR related topics and potential content so being available for occasional guidance/vision (like in writing up #2761 or eventually working with a student to write a GSOC proposal for a VR content feature) would be valuable and mentor-worthy :-)

@emanuele3d for some reason I don't have a reply box anymore on the line comment for `FirstPersonHeldItemMountPointComponent` so replying here: adding a second interface on a Component class that expects a method implemented is *technically* against the architectural spirit of the entity system. However, it is not without precedent in very "fundamental" cases.

For instance `InventoryComponent`, `LocationComponent`, and `LightComponent` all also implement `ReplicationCheck` which forces some additional networking magic in a `shouldReplicate` method. That results in very minimal checks though, while in this case System-like logic is being added that for sure belong better in a System class (where `@In` might also have an easier time working)

I wonder if we could shrink down the VR handling in `FirstPersonHeldItemMountPointComponent` to minimally update some sort of state then handle it regularly in `InputSystem` or a variant thereof? It has an `update` method that has a tidy lineup of inputs to process, could a new VR related entry just be added there, which would probably just use a dummy empty implementation or a quick conditional check most of the time?

Or would it go better back in `FirstPersonClientSystem` ? It does seem like most of it should be in a System rather than in a Component

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -89,10 +91,18 @@ public static OpenVRProvider getInstance() {
...
+     * @return the VR state.
+     */
     public OpenVRState getState() {
         return vrState;
     }
 
+    /**
+     * Initialize the VR system. Note that calling this method will cause OpenVR to launch.
+     * @return true if successful.
+     */
```
emanuele3d: I'd add a few words about the circumstances in which this method will return false.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -142,18 +152,34 @@ public boolean init() {
...
+     *
+     * @return true if initialized.
+     */
     public boolean isInitialized() {
         return initialized;
     }
 
+    /**
+     *
+     * @return true if the pose of the headset is currently considered reliable.
```
emanuele3d: Define "reliable" in this context?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -142,18 +152,34 @@ public boolean init() {
...
+     */
     public boolean isHeadTracking() {
         return headIsTracking;
     }
 
+    /**
+     *
+     * @param controllerIndex - 0 for left, 1 for right
+     * @return true if the pose of the controller is currently considered reliable.
+     */
```
emanuele3d: Same comments as above. And I'd also append ", as an integer." to the controllerIndex line.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -163,12 +189,87 @@ public void shutdown() {
         initialized = false;
     }
 
+    /**
+     * Query the VR library and update the VR state, gettable with getState(). This method should be called once per
+     * render cycle.
```
emanuele3d: I'd write:

```
Queries the VR library and updates the VR state. The VR state can then be retrieved via getState(). 

This method should be called once per frame.
```

Assuming I understood it right.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -163,12 +189,87 @@ public void shutdown() {
...
     public void updateState() {
         updatePose();
         pollControllers();
         pollInputEvents();
     }
 
+    /**
+     * Make the specified controller vibrate
+     * @param controller - the hand index, 0 for left and 1 for right.
+     * @param strength - the strength of the pulse.
```
emanuele3d: Please provide information regarding the type of the `controller` and `strength` arguments.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -163,12 +189,87 @@ public void shutdown() {
...
+     */
+    public static void triggerHapticPulse(int controller, int strength) {
+        if (controllerDeviceIndex[controller] == -1) {
+            return;
+        }
+        vrSystem.TriggerHapticPulse.apply(controllerDeviceIndex[controller], 0, (short) strength);
+    }
+
+    public void submitFrame() {
```
emanuele3d: No javadoc? All public methods except simple accessors usually should have one.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -163,12 +189,87 @@ public void shutdown() {
...
+        }
+        if (vrCompositor.PostPresentHandoff != null) {
+            vrCompositor.PostPresentHandoff.apply();
+        }
+    }
+
+    /**
+     * Set the near clip. This will influence the projection matrix that comes from the
+     * vrProvider.getState().getProjectionMatrix(...) method.
+     * @param nearClipIn - the near clip to set.
```
emanuele3d: I'd write at least:

`The distance from the camera, in opengl units, of the near clipping plane, as a float/double.` (?)

I had a pause here. I presume negative values, which are sometimes used with orthogonal cameras, would not be a good idea here?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -163,12 +189,87 @@ public void shutdown() {
...
+     */
+    public void setNearClip(float nearClipIn) {
+        nearClip = nearClipIn;
+    }
+
+    /**
+     * Set the far clip. This will influence the projection matrix that comes from the
+     * vrProvider.getState().getProjectionMatrix(...) method.
+     * @param farClipIn - the near clip to set.
+     */
```
emanuele3d: Previous comment also applies here.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -163,12 +189,87 @@ public void shutdown() {
...
+        }
+    }
+
+    /**
+     * Set the near clip. This will influence the projection matrix that comes from the
+     * vrProvider.getState().getProjectionMatrix(...) method.
+     * @param nearClipIn - the near clip to set.
+     */
+    public void setNearClip(float nearClipIn) {
+        nearClip = nearClipIn;
```
emanuele3d: I'd write:

`this.nearClip = nearClip`

And I'd use the same also in the following method. Or did you get a checkstyle warning about it?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -163,12 +189,87 @@ public void shutdown() {
...
+     * @param farClipIn - the near clip to set.
+     */
+    public void setFarClip(float farClipIn) {
+        farClip = farClipIn;
+    }
+
+    /**
+     * Turn on the keyboard overlay. This is a keyboard that hovers in front of the user.
+     * @param showingState - true or false
+     * @return - true if successful.
```
emanuele3d: I'd add some information regarding the circumstances in which this would fail.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -163,12 +189,87 @@ public void shutdown() {
...
+     * Set the far clip. This will influence the projection matrix that comes from the
+     * vrProvider.getState().getProjectionMatrix(...) method.
+     * @param farClipIn - the near clip to set.
+     */
+    public void setFarClip(float farClipIn) {
+        farClip = farClipIn;
+    }
+
+    /**
+     * Turn on the keyboard overlay. This is a keyboard that hovers in front of the user.
```
emanuele3d: Is the keyboard functionality provided by the OpenVR library? And how does it looks like?

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
```diff
@@ -55,10 +55,34 @@
         }
     }
 
+    /**
+     * Add a controller listener. This listener will receive pose and button state updates for the controller.
+     * @param toAdd - the controller listener. Implements the ControllerListener interface.
```
emanuele3d: I'd write: 

`@param listener - an object implementing the ControllerListener interface.`

And yes, I recommend a change of variable name.

This way the code becomes `controllerListeners.add(listener);` which is marginally closer to plain English.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
```diff
@@ -55,10 +55,34 @@
...
+     */
     public void addControllerListener(ControllerListener toAdd) {
         controllerListeners.add(toAdd);
     }
 
+    /**
+     * Get the pose of eye eyeIndex.
+     * @param eyeIndex - the eye index. 0 is left, 1 is right.
+     * @return
```
emanuele3d: I'd write (if I'm understanding this correctly):

```
* Get an eye's pose
* @param eyeIndex - an integer specifying which eye to obtain the pose for: 0 for the left eye, 1 for the right one.
* @return the specified eye's pose, as a Matrix4f.
```

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
```diff
@@ -55,10 +55,34 @@
...
+        Matrix4f matrixReturn = new Matrix4f(headPose);
+        matrixReturn.mul(eyePoses[eyeIndex]);
+        return matrixReturn;
+    }
+
+    /**
+     * Get the projection matrix for eye eyeIndex.
+     * @param eyeIndex - The target eye. 0 is left, 1 is right.
+     * @return
+     */
```
emanuele3d: I'd rewrite in line with the previous comment.

emanuele3d:
Regarding Cervator's comment, I don't have enough experience with the ES to add much to it. The only thing I know is that the general rule of thumb is that components should be pretty dump, i.e. data containers with simple accessors at most. All the functionality to update the component's data should be done through a system. 

While we might not want to be overly dogmatic about this, it's a good rule to keep in mind.

engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
```diff
@@ -50,14 +50,28 @@
...
 
+    /**
+     * If possible, make this object listen for controller pose upates.
+     */
     public void trySubscribeToControllerPoses() {
         vrProvider.getState().addControllerListener(this);
     }
 
+    /**
+     * A callback target for the controller listener. Not intended to be called manually.
```
indianajohn: the method is public by necessity - so that the vrProvider can call it when the listener triggers - but it's not really designed as a "public" method. It's not intended to be part of the public interface.

engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
```diff
@@ -50,14 +50,28 @@
...
+     * @param stateBefore - the state before the state change.
+     * @param stateAfter - the state after the state change.
+     * @param nController - the hand index, 0 for left and 1 for right.
```
indianajohn: Actually, the upstream API provides up to 10 tracked devices at this time (and I think that's an arbitrary number), and some additional tracked devices are expected soon:
Main article:
http://www.roadtovr.com/valve-third-party-lighthouse-api-steamvr-tracking-htc-vive-dev-kit-license-royalty/
A tracked "puck" that you could put on the end of a baseball bat or something (lock up your cats).
http://www.roadtovr.com/htc-vive-tracker-enables-a-bevy-of-bats-guns-gloves-and-other-tracked-accessories/

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
indianajohn: For VR head sets, it's very important to use the projection matrix / pose matrices given by the SDK when rendering. I've never tried to put a negative value in, but an orthographic projection in general is a no-go for VR. The reason is that, if you draw a ray from the user's left eye to the apparent location of the object in 3D space, and then draw a line from that object back to the user's right eye, the angle that ray enters the user's eye had better agree with where the left eye is perceiving the object to be. If not, the user will get double vision and headaches.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
indianajohn: I think the CheckStyle warning was not visible because "method not used" was blocking it. (I'm leaving this here because it could possibly be useful in the future, even if it's not currently used).

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
indianajohn: Yes, it is. I've added few more details. This image shows the keyboard in action:
http://www.valvetime.net/attachments/steamvr-10-jpg.25552/

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
indianajohn: To be honest, I think this will always succeed. The native function returns true or false so it's handled in the function. But I don't know why that would return false.

indianajohn:
> I wonder if we could shrink down the VR handling in FirstPersonHeldItemMountPointComponent to minimally update some sort of state then handle it regularly in InputSystem or a variant thereof? It has an update method that has a tidy lineup of inputs to process, could a new VR related entry just be added there, which would probably just use a dummy empty implementation or a quick conditional check most of the time?
> 
> Or would it go better back in FirstPersonClientSystem ? It does seem like most of it should be in a System rather than in a Component

In my next pull request, I will be creating a subsystem where this stuff goes. That subsystem will contain a listener that tracks down the `FirstPersonHeldItemMountPointComponent`, stores a reference, and edits it when the callback fires. The only remaining code in both `FirstPersonHeldItemMountPointComponent` and `FirstPersonPlayerSystem` will be the quaternion and associated handling.

Anyway, I think I got all of the comments on the Javadocs, and answered and made changes where appropriate.

GooeyHub:

Cervator:
> In my next pull request, I will be creating a subsystem where this stuff goes.

Works for me :-) Also would give another chance to poke @Josharias some more for ideas.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -211,6 +214,11 @@ public static void triggerHapticPulse(int controller, int strength) {
         vrSystem.TriggerHapticPulse.apply(controllerDeviceIndex[controller], 0, (short) strength);
     }
 
+    /**
+     * Submit the frame stored in the frame buffers for the left and right eyes to the compositor. When this method is
+     * called, the contnts of those frame buffers will show up in the head set. This method should be called exactly
```
emanuele3d: *cont-E-nts

Cervator:
Went ahead and merged this then, purging the evil typo of ultimate evil while I was at it! :-)

Thanks @indianajohn! Sent you a channel invite on Slack to our #gsoc2017, so I can remember inviting you as part-time mentor later, if we get in.

PR #2749, opened 2017-01-09
Make it possible to use more icon types for UI elements like tree views.

flo:
### Contains

A change that makes it possible to use atlas icon URLs like `engine:items#whiteRecipe` with the `StringTextIconRenderer`. The latter can for example be used by a TreeView to render its icons.

Note that this change should not break anything since the TextureRegion is a interface implemented by the class Texture: Thus a StringTextIconRenderer sub class is still allowed to return a Texture and does not need to be adjusted.

### How to test

No test needed, it should be enough that it compiles.


GooeyHub:

flo:
@rzats Ut is slightly related to the treeview

PR #2750, opened 2017-01-10
Surface Depth Facet

MaxBorsch:
1. Added an optional surface depth facet that when added limits the default surface rasterizer from generating below the depth.

2. Made spawning point selection return a value even if no "good" spawn could be found, because it's better to at least spawn in the ground than crash the game.

Part of GCI task creating an Underworld.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
+++ b/engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
@@ -1,11 +1,11 @@
 /*
- * Copyright 2013 MovingBlocks
+ * Copyright 2017 MovingBlocks
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *  http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -42,6 +42,8 @@ import org.terasology.physics.shapes.SphereShapeComponent;
 import org.terasology.rendering.logic.MeshComponent;
 import org.terasology.world.WorldProvider;
 
+import java.util.Optional;
+
 /**
  * Creates new player instances.
  */
@@ -71,7 +73,7 @@ public class PlayerFactory {
         float entityHeight = getHeightOf(builder) + extraSpace;
 
         LocationComponent location = controller.getComponent(LocationComponent.class);
-        Vector3f spawnPosition = findSpawnPos(location.getWorldPosition(), entityHeight);
+        Vector3f spawnPosition = findSpawnPos(location.getWorldPosition(), entityHeight).get(); // TODO: Handle Optional being empty
         location.setWorldPosition(spawnPosition);
         controller.saveComponent(location);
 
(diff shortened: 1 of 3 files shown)


GooeyHub:

Cervator:
I can confirm this lets Perlin + Underworld create a secondary world layer for the underworld :-)

I'm not sure about the longer search for a spawn position. I think I'd prefer timing out and rather than having that stall/crash the game return gracefully to the main menu with a sensible error message.

Forcing a spawn (possibly after an excessive amount of time searching a spiral of max size 250?) in an invalid position could be very confusing for a new player. Better to have a clean error message :-)

And not that it is a big deal but you've got a 2014 license header in there :D

Haven't tested a lot of other worlds yet (sleepy). Might be good with some bits of review from elsewhere so throwing a few reviewers on here.

engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
```diff
@@ -134,15 +134,15 @@ private Vector3f findSpawnPos(Vector3f targetPos, float entityHeight) {
         int targetBlockY = TeraMath.floorToInt(targetPos.y);
         int targetBlockZ = TeraMath.floorToInt(targetPos.z);
         Vector2i center = new Vector2i(targetBlockX, targetBlockZ);
-        for (BaseVector2i pos : SpiralIterable.clockwise(center).maxRadius(3).scale(2).build()) {
+        for (BaseVector2i pos : SpiralIterable.clockwise(center).maxRadius(250).scale(2).build()) {
```
skaldarnar: 3 might be a bit low, but 250 seems too high. Maybe 16 or 32 is a better compromise? What is the default chunk size nowadays?

engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
```diff
@@ -134,15 +134,15 @@ private Vector3f findSpawnPos(Vector3f targetPos, float entityHeight) {
...
 
             Vector3i testPos = new Vector3i(pos.getX(), targetBlockY, pos.getY());
             Vector3i spawnPos = findOpenVerticalPosition(testPos, entityHeight);
             if (spawnPos != null) {
                 return new Vector3f(spawnPos.getX(), spawnPos.getY() + entityHeight, spawnPos.getZ());
             }
         }
-        return null;
+        return targetPos;
```
skaldarnar: Should we go for `Optional[Vector3i]̀  as return type?

See: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

engine/src/main/java/org/terasology/world/generation/facets/SurfaceDepthFacet.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2014 MovingBlocks
```
skaldarnar: 2014 license header.

engine/src/main/java/org/terasology/world/generation/facets/SurfaceDepthFacet.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package org.terasology.world.generation.facets;
+
+import org.terasology.math.Region3i;
+import org.terasology.world.generation.Border3D;
+import org.terasology.world.generation.facets.base.BaseFieldFacet2D;
+
+/**
+ * Stores the surface depth limits
```
skaldarnar: Please briefly explain what the "surface depth" actually is…

engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
emanuele3d: 32x64x32 if I remember correctly.

MaxBorsch:
@skaldarnar I fixed all the stuff you mentioned. Made the getSpawn() return an optional. Added a TODO for handling an empty optional, since that is a whole other task and we haven't really agreed on what should happen.

GooeyHub:

Cervator:
Retested, this time used railgun to drill a tunnel straight down .... woo! We might need to increase the cooldown a bit on that thing.

Added a comment to clarify what the new line does in SolidRasterizer. Also checked out Polyworld just in case (different player spawn location picking)

Will submit a follow-up for figuring out a player spawn timeout. Anybody else with comments please submit here or on the new issue :-)

VaibhavBajaj:
@MaxBorsch Maybe the ceiling of this world should be mantleRock so that no matter how crazy a guy is using a rail gun, he can't reach the bottom. Just a suggestion.

Note: MantleRock is indestructible.

MaxBorsch:
@VaibhavBajaj You're supposed to be able to dig to the bottom. Also the underworld then has no bottom.

VaibhavBajaj:
@MaxBorsch Then why the portals? Also, why would the mantle rock disappear from the bottom? I was suggesting adding it in your WorldRasterizer, not actually changing the position of the whole Underground to below the mantle rock.

MaxBorsch:
@VaibhavBajaj The portals are useless, just a demo for portals, or to help people discover the underworld. Ideally they'd spawn in some necromancer castle etc.

PR #2757, opened 2017-01-15
FlexibleConfig

### LINKED ISSUE #2668: FlexibleConfig
***Introduction***
The current rendering config is a relatively straightforward (if long) Java class with hardcoded video settings and their handling. This includes accessors (get*(), set*(), is*()), but also functionality to subscribe to the config to be notified of changes - to specific values or to the whole config. 

This works, but is inflexible. Ideally engine and modules should be able to publish their configuration settings, as needed. Here is a couple of simple scenarios to ponder about.

****Scenario 1 - ShadowMapNode****
The ShadowMapNode is responsible for generating a ShadowMap used to cast shadows from the main light (sun/moon). This node wishes to publish two settings to the rendering config: a) a boolean defining its enabled/disabled state b) an integer defining the size of the (square) shadow map, in pixels. 

****Sceneario 2 - RealCameras module****
The RealCameras module (so far a fictional one) introduces a number of realistic effects for the camera. Among them are lens flares. These are generated by a specialized node injected into the render graph on module startup. The node renders up to 20 lens flare elements, each having a set of parameters to play with, such as the texture asset used, the distance along the flare axis and the maximum opacity of the element. All these settings need to be published at runtime on the rendering config, for the user to play with. 

***FlexibleConfig and Setting classes***
Enter the FlexibleConfig class, barely more than a <ResourceURN, Setting> map:
* `void add(Setting)` - doesn't allow overwriting a stored setting object with a new one using the same ResourceUrn
* `boolean remove(ResourceUrn)` - return true if the setting has been removed from the config. This happens only if the setting no longer has subscribers.
* `Setting get(ResourceUrn)`
* `boolean has(ResourceUrn)`

Individual Setting instances would contain:
* an id - of type ResourceURN: the prefix is engine-set and is either "engine" or the fully qualified name of the module publishing the setting.
* a stored value
* a type - i.e. boolean, integer, float, double, one-selection/multiple-selection lists
* a default value
* a range - where applicable, i.e. integers, float, double 
* a list of choices - where applicable, i.e. list-type settings
* a comma-separated list of ui-path
* a human-readable name - translation friendly, to be displayed in the UI
* a description - translation-friendly, potentially verbose, describing in plain language what a setting does.

`Setting` is largely immutable and contains the information from the FlexibleConfigs Features section. But it would have a setter methods for the stored value. It also provides subscribe()/unsubscribe() methods and a `boolean hasSubscribers()` method. 

A few clarifications about the ui-path feature. A ui path would provide a hint to the ui on how to group settings together. I.e. all settings having a path /engine/shadows/shadowmap *could* be displayed together when the "shadows" list item is selected in the "engine" tab of the rendering config. It could be meaningful however for the same setting to be also be displayed in a separate tab/menu, hence the possibility for multiple ui paths. Ultimately however, it is up to the UI code to decide how to take advantage of them. I.e. a simpler UI might not provide "tabs" and might group settings between horizontal separators instead, potentially ignoring the deeper sections of a path.

How does this all sound? And notice that this issue specifically avoids dealing with the UI needed to display the content of a FlexibleConfig instance. That will be for a separate issue.

eviltak:
Closes #2668. Implements a `FlexibleConfig` and `Setting` class which allows for an extensible configuration system which does not have hard-coded settings.

### Testing

- Run tests (all relevant tests are in the `org.terasology.config.flexible` package)

- [x] Add tests for each method in `Setting`
- [x] Add tests for each method in `FlexibleConfig`
- [x] Add JavaDoc
   - [x] Document `RangedValueValidator`
   - [x] Document `Setting`
   - [x] Document `SettingImpl`
   - [x] Document `FlexibleConfig`
   - [x] Document `FlexibleConfigImpl`
 - [x] Update tests to be more "test" like (check @emanuele3d review)


GooeyHub:

eviltak:
A few questions I'd like reviewers to answer:

- [ ] Should `FlexibleConfig.add` raise an `Exception` if the key is already present in the map?
- [ ] Should `FlexibleConfig.get` raise an `Exception` if the type of the `Key` and the type of the `Setting` don't match (i.e. `key.vClass.isInstance(setting.getValue())` returns `false`)? I'm leaning towards yes on this one, but it depends on the project architecture.
- [ ] Should `SettingValueValidator` be renamed to `SettingValueConstraint` (or something better)?

Cervator:
@GooeyHub: ok to test

@eviltak: thanks for all the work and consideration you've put into this topic :-)

Poking @emanuele3d and @flo for feedback

GooeyHub:

emanuele3d:
@eviltak: this PR might need some more patience than usual unfortunately.

With the other folks interested in architecture we are currently looking at some high-level concepts that have been mentioned a number of time and never fully discussed. Configs is one of those concept and it's not quite clear yet if we are even going in the right direction with issue #2668 and therefore this PR.

So, you might have to wait a bit longer than usual to get a proper review.

eviltak:
@emanuele3d that's alright. Any other things (that you mentioned) that you think I can help out in?

Cervator:
Heya @eviltak - you can check out some of the related architecture talk over here if you're interested: 

http://forum.terasology.org/threads/config-context-and-security.1761/

A ton of it is beyond the initial Config work which is sort of a phase 1 and the first thing we hope to get done, with this PR the focal point.

So far the main solid point I've taken to have some impact on this PR is the idea of moving to use `SimpleUri` instead of `ResourceUrn` - the latter is way more advanced than what we likely need in a key, and actually comes from `gestalt-asset-core` rather than the engine itself. Could you evaluate that possibility and see if it would cause any trouble to switch?

It might also give us a nice key naming convention simply consisting of a module name + object, with the module name possibly relating to an actual module or something else when appropriate (global rendering config or other stuff - more of a category there)

The idea behind that is covered in the thread: in short it would allow easy extraction of Config+Context+Injection into its own little external library at some point, without having to depend on `gestalt-asset-core` - maybe even including `SimpleUri` itself

On the type stuff and UI auto-building then my brain is still trying to digest that. I *think* I get what you mean but stuff like `<V> Setting<V> get(Key<V> key)` makes my brain melty. I would like to think we can get to where you can feed a settings object containing enough stuff including value ranges to a UI widget for it to auto-create the UI nicely. The config stuff shouldn't depend on any UI piece.

Thanks for sticking with this so far, and apologies for any missteps I may make in trying to understand all this, I really don't get enough exposure to fancy Java these days :-)

eviltak:
@Cervator Switching to `SimpleUri` should just be a simple refactor, not taking more than a couple of minutes.

The reason for creating a separate generic `Key` class for the `FlexibleConfig` is primarily to be able to obtain the type `T` of the `Setting<T>` value corresponding to the key. This is done for the same reason that Florian on the forums (@temsa on GitHub, I guess) posted on that thread. By using classes as part of the keys you can let the `FlexibleConfig` know what the correct type of the corresponding `Setting<T>` should be, and let it itself verify that you always get the correct type. Of course, you could perform an unchecked cast to the generic variant that you _think_ is the type and simply storing `ResourceUrn`/`SimpleUri` keys, but this opens up the possibilities to many `ClassCastException`s. 

The reason for making `Setting<T>` generic is so that the compiler can tell you what type is stored inside the `Setting` instead of you having to guess it. Generics are used for the same reason `CoreRegistry.get` is generic over `T` and accepts a `Class<T>`, so that the code and the compiler can prevent human error. IMO `WorldRenderer renderer = CoreRegistry.get(Context.class)` making the compiler complain is much better than `WorldRenderer renderer = (WorldRenderer) CoreRegistry.get("Context")` causing a runtime exception. For the same reason, I prefer `Setting<DisplayMode> setting = flexibleConfig.get(SettingKeys.RESOLUTION)` over `Setting<DisplayMode> setting = (Setting<DisplayMode>) flexibleConfig.get(SettingUris.RESOLUTION)`.

emanuele3d:
I understand you reasons @eviltak.

Your approach is a virtuous one but not one that I support. I will write
more about it later.

Cervator:
Our Florian is @flo, like listed over in the Reviewers section in the sidebar. Dunno who that temsa is :-)

Thanks for the further examples! Is that getting into how you'd prepare `SettingKeys.RESOLUTION` and others to be able to easily know what type you're wanting to get back? From the earlier issue discussion I remembered:

> The only pitfall of this method and using a UI-side map to build Setting UIs is that the Keys will either have to be stored as static final fields of a helper class or be generated at runtime, preventing the usage of loops.

Is that what `SettingKeys.RESOLUTION` is an example of there? The helper class? It is of type `DisplayMode` ?

I do like it when in existing code you supply a `Something.class` when you know that's the class you want to get back.

eviltak:
@Cervator You got that right. In my example, I wanted to show how my approach makes the compiler point out an error in the choice of type instead of a runtime exception (I'm sorry I didn't make that clear). `SettingKeys.RESOLUTION` is something similar to `new Key<>("engine:ResolutionSetting", Resolution.class)`. Since `setting` in my example is a `Setting<DisplayMode>` but the `get` call returns a `Setting<Resolution>`, the compiler raises an error. But in the case where the returned setting is manually casted, it would result in a runtime exception.

Cervator:
So if I understand it right that's then also what gets in the way of easy looping. You could loop over the map, but not easily get the type of what you were looking at. 

A user wanting to get a single typed object would already know what they want, much like when we retrieve something from the CoreRegistry.

Cervator:
A few more thoughts: I think I am seeing better what @emanuele3d is getting at. The very beginning of the original issue was very focused on the details for every individual config item, like the UI hint for instance. The types would be very limited as there just aren't that many - generally you have numbers, strings, and booleans (which sometimes may be in lists for stuff like radio buttons or checkboxes).

The discussion lead to all kinds of more advanced topics which introduced plenty of confusion and got us off track somewhat. Whitelists, security, all kinds of fancy stuff. That led to the forum thread to try splitting stuff apart again so we could just focus on plain config again here.

While the code here is super cool and I think I understand better what it does now it sort of ends up almost reimplementing `Context` by becoming a container for arbitrary classes that in turn could contain config (or anything, really). It is way more generic than the original vision for just handling relatively "dumb" config details.

Rather than having an exceedingly focused config object we store in a `Context`, we sort of have a config-flavored `Context` we store in a `Context`.

Hopefully I'm on track here and making some sort of sense. Maybe other config frameworks are leaking in, from situations where you really only have one sort of container of arbitrary objects that could contain config, but are classes in their own right. While here we have a split between the high level `Context` and a simple map of config items.

emanuele3d:
Just wanted to inform you all I am reviewing the code. Hopefully I will finish the review tonight (CET).

eviltak:
@Cervator The reason I'm going for a generic approach is type safety and flexibility. The following types were found by a script that I wrote in the `RenderingConfig` class, which was my primary inspiration for `FlexibleConfig`:

```
PixelFormat
PerspectiveCameraSettings
int
DisplayModeSetting
boolean
ViewDistance
float
ScreenshotSize
String
```

Instead of having 9 (and possibly more for other `Config` implementations) different `Setting` implementations that do basically the same thing with variation only in specific areas (where the `SettingValueValidator` implementations come in), I thought that it'd be better to just require `Setting` adders to define a `SettingValueValidator` that defines the constraints of the data (I really think it should be renamed to `SettingValueConstraint` now!) Also, this approach makes it much easier to define `Setting` instances of different types with the same constraints like no constraints, for example. Instead of creating multiple `Setting` impls for each type, you just pass in the same `SettingValueValidator` to a `Setting` instance generic over a different type.

Going with using separate implementations of a `Setting` interface for each object won't prevent storage of arbitrary classes that could contain absolutely everything either, since an arbitrary implementation of `Setting` can be created by the `Setting` publisher.

Regarding the similarity between the implementations of `Context` and `FlexibleConfig`, I'd say that `FlexibleConfig`/`Setting` is more than just a reimplementation of `Context`, the biggest addition being handling subscribers to value change events. If we do manage to create implementation/s of `Context` to use to store settings, it stretches the definition of "Context". As an external viewer of the code, I wouldn't expect `Context` implementations to have anything to do with storing configuration settings. If I have misunderstood what you meant to say, please correct me as what you wrote wasn't very clear on how you think `Context` can replace `FlexibleConfig`/`Setting`.

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private ResourceUrn id;
+
+    private String idString;
+    private T value;
+
+    private String name;
+
+    private String description;
+    private SettingValueValidator<T> valueValidator;
```
emanuele3d: Let's rename this variable to just `validator`. In the context of this class we know it validates the value.

emanuele3d:
By the way @eviltak, feel free to hold on changing the test files. You can wait until the other classes have stabilized.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+package org.terasology.config.flexible;
+
+import com.google.common.collect.Maps;
+import org.terasology.assets.ResourceUrn;
+
+import java.util.Map;
+
+public class FlexibleConfig {
+    private Map<Key<?>, Setting> settingMap;
```
eviltak: I can change the `Key` class to use `SimpleUri` instead, but if we completely remove the `Key` class, we won't get typed `Setting<T>` from `FlexibleConfig.get`.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -0,0 +1,93 @@
...
+    public FlexibleConfig() {
+        this.settingMap = Maps.newHashMap();
+    }
+
+    public <V> Key<V> add(Setting<V> setting) {
+        Key<V> key = new Key<>(setting.getId(), setting.getValue());
+
+        // Maybe throw an exception?
+        if (has(key))
+            return null;
```
eviltak: I return the `Key` because the `Key` is created by the `FlexibleConfig` itself, and needs to be used to access the setting from `FlexibleConfig.get`. Useful in cases like:

```
Key<Float> key = flexibleConfig.add(someFloatSetting);
// ...
Setting<Float> setting = flexibleConfig.get(key);
```

Of course, if keys are going to be manually created (like in a static helper class), I can create an overload of `add`: `boolean add(Key<V>, Setting<V>)` that returns true or false showing whether the `Key` was added. What do you think I should do with the existing `add(Setting)` method?

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @SuppressWarnings("unchecked")
+    public <V> Setting<V> get(Key<V> key) {
+        if (key == null || key.vClass == null) return null;
+
+        Setting setting = settingMap.get(key);
+
+        if (setting == null) return null;
+
+        // Possibly raise an exception?
```
eviltak: Same as above: if we stop using `Key` as the map key type, we won't get typed `Setting<T>` instances from `FlexibleConfig.get`. We'll just get the raw type `Setting`.

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,124 @@
...
+        return defaultValue;
+    }
+
+    public T getValue() {
+        return value;
+    }
+
+    public boolean setValue(T value) {
+        if (!valueValidator.isValid(value))
+            return false;
```
eviltak: Okay. The way I'm thinking of going about this is to add a method to `SettingValueValidator` that logs a message _or_ returns a logger friendly message that can be used to log a warning.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
emanuele3d: Yes, that's ok with me.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
emanuele3d: The key is the SimpleUri stored in the setting, so we don't need the config to create it and return it.

Of course we need to make sure that the map is aware of the fact that

`new SimpleUri("module:setting") == new SimpleUri("module:setting")`

that is, SimpleUris are considered identical if their content is identical.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
emanuele3d: Yep, I confirm that's ok with me, user will have to know what type it is and cast it.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
eviltak: I presume the `SimpleUri.equals` method handles that.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
eviltak: Alright. One thing though: Making the user cast it will cause loads of unchecked cast warnings. Instead, I propose to make the `get` method perform the cast for them and suppress the unchecked cast warning for it. Instead of `Setting<Integer> setting = (Setting<Integer>) config.get(uri)` polluting the codebase with unchecked cast warnings, we can have either `Setting<Integer> setting = config.get(uri, Integer.class)` or `Setting<Integer> setting = config.get<Integer>(uri)`. Since we don't have a need for the Class (unless you want to return null instead of raising an exception if the types are invalid), I propose the second method.

eviltak:
@emanuele3d Do you think a fully flexible and customizable `ConditionalValidator` (with a builder class) that chains custom conditions with AND or OR is required? It'd remove the requirement to define a new `SettingValueValidator` implementation (almost) every time you want to use a different condition and make this config **truly** flexible 😛 Maybe it can be the scope of a future issue and PR.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
emanuele3d: I understand the idea of confining the warning to the get method. It's a good idea. Can you show me how you'd write the get method for this form:

`Setting<Integer> setting = config.get(uri, Integer.class)`

emanuele3d:
Regarding the ConditionalValidator, I feel you are getting fancy on me again. :wink: But show me some examples of code using it.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
eviltak: It would be defined as 
```java
<V> Setting<V> get(SimpleUri uri, Class<V> valueClass) {
    return (Setting<V>) settingMap.get(uri);
}
```
Since the `Class<V>` will not be used (unless you want to return `null` instead of causing a class cast exception when the types do not match), the parameter can be removed. That's why I prefer `config.get<Integer>(uri)` instead of this signature. Plus, `get<Integer>(uri)` makes it more clear than `get(uri, Integer.class)` that the fetched type will be a `Setting<Integer>`, since the class parameter could be used for absolutely anything.

emanuele3d: Ok @eviltak: let's proceed with option 2 and let's see how that looks like in the code.

eviltak: My inexperience with Java got to me again 😛 Turns out that the `get` type argument is inferred from the type of the variable to be assigned. It simply is `Setting<Integer> setting = config.get(uri);` Simple, isn't it?

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,124 @@
...
+    private String idString;
+    private T value;
+
+    private String name;
+
+    private String description;
+    private SettingValueValidator<T> valueValidator;
+    private List<PropertyChangeListener> subscribers;
+
+    public Setting(ResourceUrn id, T defaultValue, SettingValueValidator<T> valueValidator) {
```
eviltak: I think a default value should compulsorily be added to the `Setting` as all types contained in the `Setting` will have `null` as the default since they are classes.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/validators/RangedNumberValueValidator.java
```diff
@@ -0,0 +1,89 @@
...
+package org.terasology.config.flexible.validators;
+
+/**
+ * Validates a {@link Number} within the specified range.
+ * @param <T> The type of the {@link Number} to validate.
+ */
+public class RangedNumberValueValidator<T extends Number & Comparable<? super T>> implements SettingValueValidator<T> {
+    private boolean inclusive;
+    private T low;
+    private T high;
```
eviltak: I think `min` and `max` are more concise while being understandable.

GooeyHub:

eviltak:
@emanuele3d `ConditionalValidator` would ideally be a tree of `Condition<T>`. It would be built in a way similar to the following:

```java
Condition<Integer> baseCondition = Condition.new(x -> MathUtil.isPrime(x))
                                                                          .or(Condition.new(x -> min < x && x <= max)
                                                                                               .and(x -> x % 2 == 0));

ConditionalValidator<Integer> validator = new ConditionalValidator<>(baseCondition);
```

Which is equivalent to a validator implementation with the following method:

```java
boolean validate(Integer value) {
    return MathUtil.isPrime(x) || (min < value && value <= max && value % 2 == 0);
}
```

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,124 @@
...
+        this.value = value;
+
+        dispatchChangedEvent(event);
+
+        return true;
+    }
+
+    public String getName() {
+        return name;
+    }
```
Cervator: We don't have a clear strategy on how to expand internationalization coverage beyond the main (static) menu. I would just make sure we have clear non-ambiguous keys, maybe prefixed with something to tie them to their origin config, so we can associate that with a translation entry at some point later.

emanuele3d:
I agree with your assessment regarding the ConditionalValidator. 

What I'm aiming for is something we know it's finished when we don't have anything to remove from it. So, as much as possible, let's try not to add things to it. :wink:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

eviltak:
@emanuele3d Can you please put forward your thoughts on the following:
- Should `FlexibleConfig.has` be renamed to `FlexibleConfig.contains`?
- Should `SettingValueValidator` be renamed to `SettingValueConstraint` (or something better)?
- Should `FlexibleConfig` and `Setting` be refactored into an interface and its implementation?

And yes, I know the build is failing. I forgot to update the tests 😛

emanuele3d:
- I'm ok with FlexibleConfig.contains if you prefer it. I'd prefer the shorter "has" but I'm not too concerned either ways.
- Validator vs Constraint: I feel that "constraint" is the description of a limit while "validator" is what checks that something is within the limits. So, I'd stick to Validator unless you want to split the description from the validation part - which would seem overkill to me. 
- FlexibleConfig and Setting becoming interfaces+implementations: sure.

GooeyHub:

GooeyHub:

eviltak:
Thanks @emanuele3d. Also, when logging a warning when `newValue` in `Setting.setValue` is invalid, do you want the `SettingValueValidator` to _return_ a message that will be logged or _directly_ log a warning?

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public FlexibleConfigImpl() {
+        this.settingMap = Maps.newHashMap();
+    }
+
+    public <V> boolean add(Setting<V> setting) {
+        SimpleUri key = setting.getId();
+
+        if (key == null || contains(key))
+            return false;
```
emanuele3d: Can we have logger warnings similar to what you did in the `remove` method please?

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -0,0 +1,68 @@
...
+
+        settingMap.put(key, setting);
+        return true;
+    }
+
+    public boolean remove(SimpleUri id) {
+        Setting setting = get(id);
+
+        if (setting == null) {
+            LOGGER.warn("The setting to remove with the id \"{}\" does not exist in the config.", id);
```
emanuele3d: I'd write:

`LOGGER.warn("Setting \"{}\" does not exist.", id);`

And below:

`LOGGER.warn("Setting \"{}\" cannot be removed while it has subscribers.", id);`

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -0,0 +1,68 @@
...
+            LOGGER.warn("The setting with the id \"{}\" still has subscribers and cannot be removed.", id);
+            return false;
+        }
+
+        settingMap.remove(id);
+        return true;
+    }
+
+    @SuppressWarnings("unchecked")
+    public <V> Setting<V> get(SimpleUri id) {
```
emanuele3d: I am puzzled by the first in this line. Wouldn't that be necessary only to restrict the type of V somehow? I.e. `<V extends Number>`?

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
```diff
@@ -0,0 +1,123 @@
...
+        return value;
+    }
+
+    public boolean setValue(T newValue) {
+        if (!validator.validate(newValue))
+            return false;
+
+        PropertyChangeEvent event = new PropertyChangeEvent(this, id.toString(), this.value, newValue);
+
+        this.value = newValue;
```
emanuele3d: I'd remove the empty lines before and after this one.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
eviltak: We need it to tell the compiler that the method is generic, but without imposing any type constraints.

eviltak: @emanuele3d Ping: when logging a warning when `newValue` in `Setting.setValue` is invalid, do you want the `SettingValueValidator` to _return_ a message that will be logged or _directly_ log a warning?

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
emanuele3d: Regarding <V>, I know the basics of generics but here there's something I'm still fuzzy about. What's the difference between a generic method and one that just returns a type, i.e. Setting<V> using generics?

Regarding Setting.setValue: it's enough logging a warning directly for the time being. If, upon use, we find out we could use the validators return specific error messages, we'll change it.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -32,24 +32,28 @@ public FlexibleConfigImpl() {
     }
 
     public <V> boolean add(Setting<V> setting) {
-        SimpleUri key = setting.getId();
+        SimpleUri id = setting.getId();
 
-        if (key == null || contains(key))
+        if (id == null) {
+            LOGGER.warn("The id of the setting can not be null.");
```
emanuele3d: The id of **a** setting `cannot` be null.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
eviltak: A generic method is simply a method that uses a type parameter with optional constraints, defined by the angle brackets at the start of the method definition. A generic method can do whatever it wants with a type. It may choose to accept a parameter of type of the type parameter, return an instance of that type or simply use a local variable of that type. Here we are using generics to define the type parameter of the `Setting` that we will return.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -32,24 +32,28 @@ public FlexibleConfigImpl() {
...
             return false;
+        } else if (contains(id)) {
+            LOGGER.warn("A Setting with the id \"{}\" already exists in the config.", id);
```
emanuele3d: "A Setting with the id \"{}\" already exists."

We can omit "in the config" because the logger already displays the name of the class from which the warning originates from.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
emanuele3d: So, it would be an error if we didn't include that first <V>?

eviltak: Yes, it would be an error.

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
```diff
@@ -101,13 +104,13 @@ public T getValue() {
     }
 
     public boolean setValue(T newValue) {
-        if (!validator.validate(newValue))
+        if (!validator.validate(newValue)) {
+            LOGGER.warn("The passed value {} is invalid.", newValue);
```
emanuele3d: Ah, that's what you meant. =)

No, this isn't good: it's too generic, it wouldn't help you debug a failure. The validator itself has to log a warning and then return false, so that the warning can be very specific on what went wrong, i.e. the new value being beyond the maximum value allowed.

eviltak: That's why I was asking. Do you want the `validate` method to log a warning itself or do you want another method that encapsulates `validate` and logs a warning if `validate` returns false? Or maybe we can pass in a flag to `validate` that states whether it should log a warning or not. Personally I wouldn't log a warning by _default_ since you wouldn't expect a validating method to log a warning if the value is invalid since its job is just telling you whether the value is valid or not.

emanuele3d: In my earlier comment I thought of suggesting option 1, but then I thought you'd figure it out yourself... which you did! :smiley: 

If a developer is setting a value and the operation return false the developer will wonder: why was I not able to set the value? As such, I'd explain in the javadoc of the method to check the log for the exact nature of the problem when the method returns false.

eviltak: @emanuele3d but what if someone doesn't read the method doc before implementing it in their implementation of `SettingValueValidator` and does not log a warning? Instead, we can add another method to `SettingValueValidator` called `validateWithWarning` (or something better) which first returns whether the value is valid while logging an error if it is invalid, like follows:
```java
public boolean validateWithWarning(T value) {
    boolean isValid = this.validate(value);
    if (!isValid) {
        logger.warn("[Why the value is invalid]");
    }
    return isValid;
}
```

emanuele3d: My apologies. In my earlier comment I did mean option 2, which is what you suggest. Specifically, I'd like:

```java
public final boolean validate(T value) {
    boolean isValid = fastValidate(value);
    if (!isValid) {
        this.issueWarnings(value);
    }
    return isValid;
}
```

where fastValidate and issueWarnings are methods that _must_ be implemented by subclasses.

Notice I could have suggested issueWarnings() - with no arguments- which would assume some preliminary works gets done in plainValidate, i.e. to select which message to send to the logger. 

In practice I would prefer for the fastValidate method to remain as fast possible. issueWarnings can then do whatever it needs, potentially replicating some of the checks done in fastValidate, to determine the exact message to log. In the most straightforward cases however, i.e. intervals, it might not involve any duplicate checks and be a one-liner looking like this:

```java
private void issueWarnings(T value) {
    // neglecting actual syntax for simplicity
    logger.warn( Cannot set module:myParam to 2.0: allowed interval [0.0, 2.0[ );
}
```

eviltak: That's what I prefer too. But from your examples I'm just going to rename `validate` to `validateWithWarnings` and `fastValidate` to `validate`.

eviltak:
@emanuele3d should I use annotations like `@Nullable` to show that a value can be null while also stating it in the docs?

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Represents a setting stored in a {@link FlexibleConfig} identified by an id. Contains a value that can
+ * optionally be validated by a {@link SettingValueValidator<T>} and notifies its subscribers when the value
+ * stored is changed.
+ * @param <T> The type of the value this {@link Setting} contains.
+ */
+public interface Setting<T> extends GeneralSubscribable {
+    /**
+     * Returns the id of this {@link Setting}.
```
emanuele3d: Is this a link to this exact interface?

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    SimpleUri getId();
+
+    /**
+     * Returns the {@link SettingValueValidator<T>} used by this {@link Setting<T>}.
```
emanuele3d: ` Returns the {@link SettingValueValidator<T>} used by this {@link Setting<T>}, if present. Returns null otherwise.`

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,71 @@
...
+    T getDefaultValue();
+
+    /**
+     * Returns the value stored in this {@link Setting}.
+     */
+    T getValue();
+
+    /**
+     * Sets the value stored in this {@link Setting<T>} if the passed value is valid. If it is invalid, the stored
+     * value is not updated. On successfully updating the value, the subscribers are notified of the change.
```
emanuele3d: I'd write:

`When no Validator is present the new value immediately replaces the stored one and any subscriber is notified of the change. If a Validator is present, the value is first validated. Only if the value is valid it replaces the stored one and subscribers are notified.`

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param newValue The new value to store.
+     * @return True if the value was valid and was stored successfully, false otherwise.
```
emanuele3d: I'd write:

`@return True if the value was was stored successfully, false otherwise.`

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    boolean setValue(T newValue);
```
emanuele3d: I just realized we never got around adding a resetValue() method. Without it it doesn't make sense to have a default value. Can you add it up really quick? With javadoc and all?

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Returns the human readable name of this {@link Setting}.
+     */
+    String getHumanReadableName();
+
+    /**
+     * Returns the description of this {@link Setting}.
```
emanuele3d: Returns _a potentially verbose_ description of this {@link Setting}.

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    String getDescription();
```
emanuele3d: It's not quite clear just yet how we'll handle internationalization. 

Would you like to ignore this aspect and wrap up this PR sooner or would you like to deal with this issue in this PR? I'd recommend the first option: let's proceed toward the finish line and we'll take care of the internationalization issue in a separate PR.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
```diff
@@ -0,0 +1,174 @@
...
+    private T value;
+
+    private String humanReadableName;
+
+    private String description;
+    private SettingValueValidator<T> validator;
+    private List<PropertyChangeListener> subscribers;
+
+    /**
+     * Creates a new {@link SettingImpl} that does not use a validator with the given id and default value.
```
emanuele3d: `Creates a new {@link SettingImpl} with the given id and default value but no validator.`

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
```diff
@@ -0,0 +1,174 @@
...
+    private boolean validate(T value) {
+        return validator == null || validator.validate(value);
+    }
+
+    /**
+     * Subscribe a listener that will be notified when the value stored in the setting changes.
+     */
+    public void subscribe(PropertyChangeListener listener) {
+        if (subscribers == null) {
+            subscribers = Lists.newArrayList();
```
emanuele3d: Just thought of something here. Could this be a set? So that the same listener cannot be registered twice?

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        subscribers.add(listener);
```
emanuele3d: I'm also wondering if we should be nice to the users of this code and do the usual return false+log message combo if adding/removing a subscriber didn't succeed for whatever reason. What do you think?

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
```diff
@@ -0,0 +1,174 @@
...
+     * {@inheritDoc}
+     */
+    public T getValue() {
+        return value;
+    }
+
+    /**
+     * Sets the value stored in this {@link Setting<T>} if the passed value is valid. If it is invalid, the stored
+     * value is not updated and a warning is logged. On successfully updating the value, the subscribers are
+     * notified of the change.
```
emanuele3d: See an earlier comment on how to improve this.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param newValue The new value to store.
+     * @return True if the value was valid and was stored successfully, false otherwise.
```
emanuele3d: `@return True if the value was stored successfully, false otherwise.`

engine/src/main/java/org/terasology/config/flexible/validators/SettingValueValidator.java
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.config.flexible.validators;
+
+/**
+ * Validates the given value and optionally issues warnings if the value is invalid.
```
emanuele3d: I'd remove the `optionally`.

engine/src/main/java/org/terasology/config/flexible/validators/SettingValueValidator.java
```diff
@@ -0,0 +1,50 @@
...
+ * @param <T> The type of values this {@link SettingValueValidator} validates.
+ */
+public interface SettingValueValidator<T> {
+    /**
+     * Checks whether the given value is valid or not.
+     * @param value The value to validate.
+     * @return True if the value is valid, false otherwise.
+     */
+    boolean validate(T value);
```
emanuele3d: Rename to fastValidate please.

engine/src/main/java/org/terasology/config/flexible/validators/SettingValueValidator.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Issues (logs) appropriate warnings if the given value is invalid.
+     * @param value The value to issue warnings for.
+     */
+    void issueWarnings(T value);
```
emanuele3d: Hmmm. This way these methods (fastValidate/issueWarnings) are public. So, somebody could use fastValidate instead of validate. I seem to remember I recommended for these methods to be protected and abstract, so that they must be implemented but they cannot be used directly.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
eviltak: Yes, that is a good idea.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
eviltak: Yeah, I'll do that in tandem with making `subscribers` a `Set`. If the listener already exists/is null, we can return false and log a warning.

eviltak:
Sorry for the inactivity, been busy with school. I'll hopefully fix the aforementioned issues in a day or two.

engine/src/main/java/org/terasology/config/flexible/Setting.java
eviltak: I'd do it in a separate PR too.

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Bump. Some new changes were made a few days ago, may have snuck under the radar :-)

@emanuele3d / @eviltak: think this is about ready for merging?

eviltak:
@Cervator this PR is pretty close to completion. Just the documentation and test fixes remain. I won't be able to work on this as much as I'd like to, unfortunately, because school examinations start tomorrow.

Cervator:
No worries @eviltak! I just wanted to make sure the latest changes didn't slip through a crack with nobody aiming to check on this again soon :-)

RL always comes first!

emanuele3d:
Definitely getting close, but I'm not convinced all issues have been addressed. At this stage I'd recommend to @eviltak to address all the issues and then ping me so that I can review the code again. Of course do let me know if I haven't provided an answer yet to any question you might have asked.

engine/src/main/java/org/terasology/config/flexible/Setting.java
eviltak: @emanuele3d Yes, it is.

engine/src/main/java/org/terasology/config/flexible/Setting.java
eviltak: @emanuele3d Sure can do. I was feeling like we were missing something the whole time, I guess this is what it was 😄

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -0,0 +1,55 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.config.flexible;
+
+import org.terasology.engine.SimpleUri;
+
+/**
+ * Stores multiple potentially related {@link Setting} instances that can be retrieved using their id.
```
eviltak: Like all settings in a `FlexibleConfig` could be related to rendering, etc.

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/validators/SettingValueValidator.java
eviltak: @emanuele3d I'd like to keep fastValidate public as it can be used when warnings needn't or shouldn't be logged. `issueWarnings` can be made protected, however.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Bump - I see some commits snuck in since the last review :-)

@eviltak are you happy with the current state and wanting hopefully final review from @emanuele3d?

eviltak:
@emanuele3d I'd like you to review the documentation while I revise the unit tests.

GooeyHub:

GooeyHub:

GooeyHub:

emanuele3d:
@eviltak: review request noted. Please stand-by as I'm in holiday for a week. Should be able to review this one of these days or as soon as I come back.

eviltak:
FYI, I'm waiting on @emanuele3d's review. I've fixed the `FlexibleConfigTests` and the documentation, but am waiting further approval and suggestions.

emanuele3d:
Apologies. RL is very dense right now. Let's see if I can do a review this weekend.

engine/src/main/java/org/terasology/config/flexible/validators/RangedNumberValidator.java
```diff
@@ -78,4 +82,10 @@ public boolean validate(T value) {
 
         return withinMinBoundary && withinMaxBoundary;
     }
+
+    @Override
+    public void issueWarnings(T value) {
+        LOGGER.warn("Value {} is not in the range {}{}, {}{}", value, minInclusive ? "[" : "(",
+                min != null ? min : "UNBOUNDED", max != null ? max : "UNBOUNDED", maxInclusive ? "]" : ")");
```
emanuele3d: This hasn't been addressed.

engine/src/main/java/org/terasology/config/flexible/validators/SettingValueValidator.java
emanuele3d: What scenario(s) do you have in mind in which validation make sense but issuing a warning doesn't?

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public interface FlexibleConfig {
+    /**
+     * Adds a {@link Setting} to the config if the id of the {@link Setting} is valid.
+     * @param setting The {@link Setting} to add.
+     * @return True if the {@link Setting} was added, false otherwise.
```
emanuele3d: Why did you not like this phrasing?

engine/src/main/java/org/terasology/config/flexible/Setting.java
emanuele3d: Then what's the point having it there? If you click on it you get to... the same file!

engine/src/main/java/org/terasology/config/flexible/Setting.java
emanuele3d: You left the "Sets the value..." part, which is incorrect because it assumes values are checked for validity. I'd start directly with "When...", which clarifies what happens in both the presence of absence of a validator.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
emanuele3d: This didn't happen yet. :slightly_smiling_face:

emanuele3d:
Apologies I couldn't review this as quickly as I'd hoped in the past couple of comments. 

It is reviewed now. Please have a look at the last block of "outdated" comments. A few haven't been addressed or there is some lingering discussions. 

Looking good though. I wouldn't be surprised if we manage to merge it by the weekend.

engine/src/main/java/org/terasology/config/flexible/Setting.java
eviltak: I just like to annotate all types and code references in documentation with the link tag, because it is syntax highlighted in the documentation comments and looks better when performing a documentation lookup through an IDE. If you don't think it's appropriate I can remove them.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
eviltak: Must have sneaked under the radar. This will require `Setting` to not implement `GeneralSubscribable`, though. If that's okay, I can put the `subscribe` and `unsubscribe` methods in the `Setting` interface itself.

engine/src/main/java/org/terasology/config/flexible/validators/SettingValueValidator.java
eviltak: Say for example the UI sets the value in a `Setting`. The UI controller simply wants to check whether the value set by the user is valid or not, and prompt the _user_ if the value is invalid. There is no need of polluting the log file by logging a warning when the change is not being done through code.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
eviltak: Since consumers will be working with the interface itself, I prefer "warnings _will_ be issued" rather than "warnings _should_ be issued by an implementation" as it is more clear to the consumer and does not leave him wondering that the specific implementation of `FlexibleConfig` he refers to issues warnings or not. This way, all (good) implementations of `FlexibleConfig` will _have_ to issue warnings to comply with the base interface documentation.

engine/src/main/java/org/terasology/config/flexible/Setting.java
eviltak: Don't you think I should keep "Sets the value stored in this {@link Setting<T>}" (only up to that point)? I prefer the first line of any method's docs to outline what it does, even if it is clear from the method name. The resulting docs would be "Sets the value stored in this {@link Setting<T>}. When no Validator..."

engine/src/main/java/org/terasology/config/flexible/validators/RangedNumberValidator.java
eviltak: Your first suggestion does seem more intuitive. I'll make the code more readable.

As for your second suggestion, it will require adding a parameter to both `issueWarnings` and `validate`, either of the type `Setting` or of `SimpleUri` (for the setting's id). I prefer the former with a default value of null, so that if someone wishes to execute `validate` outside of a `Setting`, he can do so while still being able to log warnings.

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
eviltak: Also, do you want the logged messages to include the id of the setting?

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/validators/RangedNumberValidator.java
emanuele3d: Hmmm. I see what you mean. 

Could we provide the validator instance with the settings's SimpleURI? I.e. in a validator.setSettingURI(SimpleUri) to be called just once in the setting's constructor? I guess this prevents validator sharing between settings, but is that a probable scenario? I'm guessing not.

engine/src/main/java/org/terasology/config/flexible/validators/SettingValueValidator.java
emanuele3d: Alright. You convinced me. :smile:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
emanuele3d: Fair points. How about "warnings _-must-_ be issued by an implementation" so that it's clear we are not attempting to predict the future?

engine/src/main/java/org/terasology/config/flexible/Setting.java
emanuele3d: Oh, I see, just to highlight it. Ok fair enough, perhaps there is a more orthodox way to do the highlighting but I'm not too fussed about it, let's leave it as it is.

engine/src/main/java/org/terasology/config/flexible/Setting.java
emanuele3d: Fair point: "Sets the value stored in this {@link Setting}. When..." it is then.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
emanuele3d: Yes to both, thank you.

engine-tests/src/test/java/org/terasology/config/flexible/FlexibleConfigTest.java
```diff
@@ -0,0 +1,154 @@
...
+            config = new FlexibleConfigImpl();
+        }
+
+        @Test
+        public void testRemove() throws Exception {
+            SimpleUri id1 = new SimpleUri("engine-tests:TestSetting1");
+            SimpleUri id2 = new SimpleUri("engine-tests:TestSetting2");
+
+            config.add(new MockSetting(id1));
+
```
emanuele3d: I'd remove this empty line.

engine-tests/src/test/java/org/terasology/config/flexible/FlexibleConfigTest.java
```diff
@@ -0,0 +1,154 @@
...
+            assertTrue(config.add(new MockSetting(id)));
+        }
+
+        @Test
+        public void testAddExisting() throws Exception {
+            SimpleUri id = new SimpleUri("engine-tests:TestSetting");
+
+            assertTrue(config.add(new MockSetting<Integer>(id)));
+
```
emanuele3d: I'd remove this empty line.

engine-tests/src/test/java/org/terasology/config/flexible/FlexibleConfigTest.java
```diff
@@ -0,0 +1,154 @@
...
+        @Test
+        public void testSubscribedRemove() throws Exception {
+            SimpleUri id = new SimpleUri("engine-tests:TestSetting");
+
+            Setting setting = new MockSetting(id);
+
+            config.add(setting);
+
+            setting.subscribe(propertyChangeEvent -> {
+            });
```
emanuele3d: Can we put the previous line in line with the one above it?

engine-tests/src/test/java/org/terasology/config/flexible/SettingTest.java
```diff
@@ -0,0 +1,144 @@
...
+@RunWith(Enclosed.class)
+public class SettingTest {
+    public static class SetValue {
+        private Setting<Integer> setting;
+
+        private int eventResult;
+
+        @Before
+        public void setUp() {
+            setting = new SettingImpl<>(new SimpleUri("engine-tests:TestSetting"),
```
emanuele3d: Let's put the SimpleUri in a variable, to keep the constructor call cleaner visually.

engine-tests/src/test/java/org/terasology/config/flexible/SettingTest.java
```diff
@@ -0,0 +1,144 @@
...
+                    50, new RangedNumberValidator<>(0, 100, false, false));
+
+            eventResult = -1;
+
+            setting.subscribe(propertyChangeEvent -> eventResult = (int) propertyChangeEvent.getNewValue());
```
emanuele3d: Can you describe this to me in plain english please? I don't see any flaw in it but I'm not agile with this newish syntax yet.

engine-tests/src/test/java/org/terasology/config/flexible/SettingTest.java
```diff
@@ -0,0 +1,144 @@
...
+            assertFalse(setting.hasSubscribers());
+        }
+
+        @Test
+        public void testSetEventCall() {
+            setting.subscribe(listener);
+
+            Random random = new FastRandom();
+
+            final int n = 50;
```
emanuele3d: n, i, r: let's not abbreviate please, let's use meaningful names for all of them.

engine-tests/src/test/java/org/terasology/config/flexible/validators/RangedNumberValidatorTest.java
```diff
@@ -0,0 +1,210 @@
...
+            validator = new RangedNumberValidator<>(min, max, minInclusive, maxInclusive);
+        }
+
+        @Test
+        public void testAllInclusive() {
+            initValidator(0, 100, true, true);
+
+            for (int i = validator.getMin(); i <= validator.getMax(); i++) {
+                assertTrue(String.format("%d returned invalid", i), validator.fastValidate(i));
+            }
```
emanuele3d: Looping through all 100 numbers seems a bit overcomplicated here and below. Just check a much smaller range, 1 to 3 for example, using simple, straightforward code.

engine-tests/src/test/java/org/terasology/config/flexible/validators/RangedNumberValidatorTest.java
```diff
@@ -0,0 +1,210 @@
...
+        @Test
+        public void testMinExclusive() {
+            initValidator(0, 100, false, true);
+
+            for (int i = validator.getMin() + 1; i < validator.getMax(); i++) {
+                assertTrue(String.format("%d returned invalid", i), validator.fastValidate(i));
+            }
+
+            assertFalse(validator.fastValidate(validator.getMin()));
+            assertTrue(validator.fastValidate(validator.getMax()));
```
emanuele3d: In line with my previous comment I'd simplify here and elsewhere using explicit numbers rather than getMin()/getMax().

engine-tests/src/test/java/org/terasology/config/flexible/validators/RangedNumberValidatorTest.java
```diff
@@ -0,0 +1,210 @@
...
+            assertTrue(validator.fastValidate(50000));
+            assertTrue(validator.fastValidate(50));
+
+            assertTrue(validator.fastValidate(-1000));
+            assertTrue(validator.fastValidate(-50000));
+        }
+    }
+
+    public static class DoubleValidator {
+        private static final double EPSILON = 0.000001d;
```
emanuele3d: Epsilon is indeed often a synonym with "error", but I'd name it more eplicitly, i.e. ALLOWED_ERROR, for the benefit of younger students, i.e. GCI, which may or may not have encountered yet that notion.

engine-tests/src/test/java/org/terasology/config/flexible/validators/RangedNumberValidatorTest.java
```diff
@@ -0,0 +1,210 @@
...
+            validator = new RangedNumberValidator<>(min, max, minInclusive, maxInclusive);
+        }
+
+        @Test
+        public void testAllInclusive() {
+            initValidator(0d, 100d, true, true);
+
+            for (double i = validator.getMin(); i <= validator.getMax(); i++) {
+                assertTrue(String.format("%f returned invalid", i), validator.fastValidate(i));
+            }
```
emanuele3d: Same comment as above: use a much smaller range and a few validation lines rather than a loop please.

engine-tests/src/test/java/org/terasology/config/flexible/validators/RangedNumberValidatorTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+            assertFalse(validator.fastValidate(validator.getMin() - EPSILON));
+            assertFalse(validator.fastValidate(validator.getMax() + EPSILON));
```
emanuele3d: Same as in a previous comment, I'd turn the getMax()/getMin() calls into actual numbers. We don't want to test anything else but the `validator.fastValidate` method here.

engine-tests/src/test/java/org/terasology/config/flexible/validators/RangedNumberValidatorTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+            assertTrue(validator.fastValidate(1000d));
+            assertTrue(validator.fastValidate(50000d));
+            assertTrue(validator.fastValidate(50d));
+
+            assertTrue(validator.fastValidate(-1000d));
+            assertTrue(validator.fastValidate(-50000d));
+        }
+    }
+}
```
emanuele3d: Missing newline!

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2016 MovingBlocks
```
emanuele3d: Just noticed you are still outputting 2016 copyright notices. I think a running `gradlew idea` might do the trick.

engine/src/main/java/org/terasology/config/flexible/Setting.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    T getDefaultValue();
+
+    /**
+     * Returns the value stored in this {@link Setting}.
+     */
+    T getValue();
+
+    /**
+     * Sets the value stored in this {@link Setting<T>} if the passed value is valid. When no Validator is
```
emanuele3d: Remember to remove the "if the passed value..." please.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
```diff
@@ -0,0 +1,201 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public boolean subscribe(PropertyChangeListener listener) {
+        if (subscribers == null) {
+            subscribers = Sets.newHashSet();
+        }
+
+        if (listener == null) {
+            LOGGER.warn(MessageFormat.format(this.warningFormatString,
```
emanuele3d: What's the advantage of using the format string compared to the seemingly simpler version of this line we had earlier?

eviltak: I was thinking of passing this format string as a base format string to the SettingValueValidator so that the id of the corresponding string can be printed out in the logs too.

engine-tests/src/test/java/org/terasology/config/flexible/FlexibleConfigTest.java
eviltak: I'm not sure I understand what you mean.

engine-tests/src/test/java/org/terasology/config/flexible/SettingTest.java
eviltak: Lambdas are equivalent to new anonymous object (with a _single_ method) creation statements, but are much cleaner. 

`propertyChangeEvent -> eventResult = (int) propertyChangeEvent.getNewValue()` evaluates to
 ```
new PropertyChangeListener() {
    @Override
    void propertyChange(PropertyChangeEvent propertyChangeEvent) {
        eventResult = (int) propertyChangeEvent.getNewValue();
    }
}

engine/src/main/java/org/terasology/config/flexible/Setting.java
eviltak: Ah, must've missed it. That's why I prefer using inheritDoc.

engine-tests/src/test/java/org/terasology/config/flexible/SettingTest.java
emanuele3d: Understood, thank you.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
emanuele3d: Alright, let's leave it as it is.

engine-tests/src/test/java/org/terasology/config/flexible/FlexibleConfigTest.java
emanuele3d: `setting.subscribe(propertyChangeEvent -> {});`

GooeyHub:

eviltak:
I just realized that I used a default method in the `Setting` interface but converted `SettingValueValidator` to an abstract class to provide a default method implementation 😄

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
eviltak: That didn't make any difference whatsoever to the copyright template. In fact, I think it has broken my project. I may have to re-clone it.

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
Cervator: I believe we solved this via IRC/Discord :-)

emanuele3d:
@eviltak: please let me know when it's time for me to review the changes.

GooeyHub:

engine-tests/src/test/java/org/terasology/config/flexible/FlexibleConfigTest.java
eviltak: @emanuele3d the code reformat option automatically reformats it to the existing form, so I'd prefer to leave it that way.

GooeyHub:

eviltak:
@emanuele3d Please review when possible.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

emanuele3d:
Review in progress....

emanuele3d:
@eviltak: additional question for you. 

A couple of FlexibleConfigs-related issues (#2893, #2894) are in the works to determine where to go next, after FCs get merged in the codebase. @vampcat seem to have a bit of free time available and would be happy to take care of them or at least help.

You have given a lot of time to this PR and it's only fair we ask you first: would you like to continue being the coding lead on this effort? Do you have enough time and energy to proceed with other FC-related issues? I'd be happy for you to do so. 

Alternatively you could help with the review process, which tends to be somewhat less time-consuming compared to coding but just as important. 

Let us know!

eviltak:
@emanuele3d I'm more than fine with others pitching in. If @vampcat (or anyone else, for that matter) is ready to tackle those issues, they are free to do so. I may not have time to actually implement the changes, but I can and will post recommendations and answer/pose any questions if needed.

I'd love to help with the review process, if needed.

engine-tests/src/test/java/org/terasology/config/flexible/FlexibleConfigTest.java
eviltak: @emanuele3d I let the IDE/editor do the formatting for me because it has _never_ failed me, and because open source projects (like this one) usually define a set of formatting rules for their IDE of choice following the project's conventions which I don't want to deviate from.

GooeyHub:

emanuele3d:
@Cervator: this PR is finally good to go. At 236 conversations (with this one), I believe it's the new record? :blush: 

@eviltak: thank you for your work and your persistence with this. I believe when Flexible Configs serialization is in place FCs will slowly become widespread in many parts of Terasology's codebase and your work will greatly pay off. 

Regarding future FCs-related efforts, especially on Serialization, I then recommend for @vampcat to lead the charge. I will then be close behind leading/coordinating the review process and eviltak will take part in it in the measure that his schedule and level of interest will allow.

Again, @eviltak, thank you.

Cervator:
Thanks @eviltak and reviewers! Merged at last :-) Especially appreciated are the new unit tests, ran them and all appeared well.

This is definitely another of our record PRs, there are a few though and they each tend to excel in one or more unique ways :D

Looking forward to more work in this area, as well as maybe a small chapter in a tutorial somewhere.

SoniEx2:
How do I use this?

eviltak:
@SoniEx2 I'd recommend you wait until all the supporting structures (like serialization #2893 and UI support #2894) are in place, because this won't be much use otherwise. Expect a tutorial on usage after FlexibleConfig is completely ready for consumption both internally and externally.

SoniEx2:
I need *something* for #2901, even if it's just experimental.

eviltak:
@SoniEx2 FlexibleConfig hasn't been integrated into the engine yet (i.e. there are no ways to create, store or retrieve a FlexibleConfig), so using it _in its current state_ is out of the question. This PR only deals with the implementation, and the integration is going to be covered in future issues and PRs. I suggest you go with the typical hard coded class in the `org.terasology.config` package, or wait. All such hard coded configuration classes should be migrated to FCs in the near future.

PR #2728, opened 2017-01-03
Particle System

MaxBorsch:
New core particle system. Currently fixing & optimizing & planning.

- [x] **Multiple renderers**
- [ ] **Particle system does not stop rendering when far away (particlesystems don't dispose)**
- [x] **Convert legacy particle stuff (block effects mostly)**
- [ ] **Long-range particlesystem registry, "Chimney-problem"**
- [ ] Optimize updater to partition ParticlePool based on emissionRate of emmiters, not evenly.
- [ ] **Force-fields**
- [ ] Create documentation
- [ ] Finish annotating code
- [ ] Create tests

fireworks

PR #2758, opened 2017-01-19
Added postWorldGeneration event / WorldGenerationSubscriberSystem

MaxBorsch:
Currently systems have no events that fire after world generation. I went and added one, postWorldGeneration(). 
More events related to worldgen can be added to the WorldGenerationSubscriberSystem.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/modes/loadProcesses/PostWorldGenerationSystems.java
+++ b/engine/src/main/java/org/terasology/engine/modes/loadProcesses/PostWorldGenerationSystems.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2016 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.engine.modes.loadProcesses;
+
+import org.terasology.context.Context;
+import org.terasology.engine.ComponentSystemManager;
+import org.terasology.entitySystem.systems.ComponentSystem;
+import org.terasology.entitySystem.systems.WorldGenerationSubscriberSystem;
+
+import java.util.Iterator;
+import java.util.LinkedList;
+
+/**
+ */
+public class PostWorldGenerationSystems extends StepBasedLoadProcess {
+
+    private final Context context;
+
+    private Iterator<WorldGenerationSubscriberSystem> componentSystems;
+
+    public PostWorldGenerationSystems(Context context) {
+        this.context = context;
+    }
+
+    @Override
+    public String getMessage() {
+        return "Post-World Generation Systems";
+    }
+
+    @Override
+    public boolean step() {
+        if (componentSystems.hasNext()) {
+            componentSystems.next().postWorldGeneration();
+        }
+        return !componentSystems.hasNext();
+    }
+
+    @Override
+    public void begin() {
+        ComponentSystemManager csm = context.get(ComponentSystemManager.class);
+        LinkedList<WorldGenerationSubscriberSystem> systems = new LinkedList<>();
+        for (ComponentSystem system : csm.iterateAll()) {
+            if (system instanceof WorldGenerationSubscriberSystem) {
+                systems.add((WorldGenerationSubscriberSystem) system);
+            }
+        }
+        componentSystems = systems.iterator();
+    }
+
+    @Override
+    public int getExpectedCost() {
+        return 1;
+    }
+}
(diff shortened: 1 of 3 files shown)


Cervator:
Came out of some discussion on Slack. Idea was to have a place that triggered once after, I guess, the world gen setup phase, right as chunks start generating, but before `update()` in `SubscriberSystems` begin.

Started from a joke about `postPostBegin()` but our world gen types can probably comment on that more closely, pinging a few via requesting review ... :-)

GooeyHub:

flo:
I think it would be good to have a concrete example that needs this event hook.

I could imagine that if you want to introduce a way to do something after a system A has done it's post begin code then you could let that system A fire an event during it's post begin.

flo:
Just for discussion:Maybecwe could even think about making post begin a event so that some systems can simply subscribe with a lower prio if they want to be called afterwards

flo:
(first comment was also only for discussion. Some feedback frim the others would be good)

MaxBorsch:
@flo As an example, this could be for systems that need to grab a block entity / look at the world somehow. Since systems start before there are blocks this isn't currently possible. Ask @vaibhavbajaj on Slack for info on why he needed this. I needed it earlier to have a system place a block in the world at startup, etc.

flo:
Blocks need to have systems to be placed/loaded so systems must come first. Also on a headless server there might not be any blocks loaded until a player joins. Also the question would be how many chunks of the world you need loaded to decide if your block can be placed.

The block chould be placed by a world generator (plugin). Or it could be placed like we place dungeons in Gooey's Quests.

VaibhavBajaj:
@flo I am just quickly dropping in to explain what I had in mind which required the system.
I have a few Updraft blocks that push the player up and I wanted to animate them to move up using the delayManager periodicActionEvent. That is when I requested that there be a process that runs only once in the whole game and only runs AFTER everything in the world has been rasterized.

I have since proved unsuccessful in my attempts to create such an animation (since I realized the block is activated only after the player aims his cursor at it. Also, @ForceBlock ( or @ForceActive... whichever) does not work as intended) Nonetheless, there is need for such a process which runs only once the whole game and can be used for similar purposes, say block teleportation when play clicks E or something of the like.

iojw:
If you would like to deal with blocks generated during the world gen phase, wouldn't it make more sense to do so inside the `onChunkGenerated` event since the Terasology world is *technically* infinite? 

This is actually what Terasology/SimpleLiquids#2 does to apply liquid propagation to water blocks generated by rasterizers.

@VaibhavBajaj : Sorry I'm not very sure what you mean, could you elaborate a bit more on how this post world-gen event would allow you to animate blocks or allow for block teleportation when the player clicks E? Could this not be done through an `ActivateEvent` instead? 🙂

Also, @ForceBlockActive seemed to work alright the last time I tested it in persisting the entity containing the component. How did it not work for you?

VaibhavBajaj:
@iojw. As you might know, onChunkGenerated() and onChunkLoaded() events are for each chunk. I wanted to move each updraft block up by 1 position. There would be chunks where this would give an OutOfBoundsException. Also, exactly how many delayManager instances do you plan to make using that? 
In short, I tried that and failed miserably. It just doesn't suit my idea.

flo:
Actually I think your upwind thing could actually be done better via entity. There is a way to schedule entity placement at world generation time. Like a name tag gets generated in this class:

https://github.com/Terasology/StaticCities/blob/master/src/main/java/org/terasology/staticCities/SettlementEntityProvider.java

flo:
The entity could then either have child entity that you move around, or it could have a particle system, or it could have an animated model.

flo:
I added some further links

VaibhavBajaj:
@flo Thanks a lot. I am looking into this right now and will change my approach as appropriate.

flo:
Oh and please be aware that the particle systems are currently relativly short lifed.

This will change once #2728 also from @MaxBorsch got merged, which will allow for particle systems that continuously spawn particles.

skaldarnar:
I know I'm late to the discussion, but let me add my 2 cents anyways.

I remember having problems when trying to get the magic dome thingy working in Light and Shadow. The problem was to add the dome entity during the world generation - even the targeted chunk was not generated (think of a dome at (0,0,0) but the player spawns far away). Maybe the `postWorldGeneration` event would be a more logical approach than what is currently implemented - see [MagicDomeSystem](https://github.com/Terasology/LightAndShadow/blob/master/src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java#L54-L72).

engine/src/main/java/org/terasology/engine/modes/loadProcesses/PostWorldGenerationSystems.java
```diff
@@ -0,0 +1,67 @@
...
+import org.terasology.context.Context;
+import org.terasology.engine.ComponentSystemManager;
+import org.terasology.entitySystem.systems.ComponentSystem;
+import org.terasology.entitySystem.systems.WorldGenerationSubscriberSystem;
+
+import java.util.Iterator;
+import java.util.LinkedList;
+
+/**
+ */
```
skaldarnar: Why empty Javadoc? Just so that some code analysis tool does not complain? ;-)

MaxBorsch:
@skaldarnar That's practically the same reason I wanted this event added. In my [WorldPortals](https://github.com/Terasology/WorldPortals/) module, each portal is an entity, and since I had some portals generated during world generation I wasn't able to create the entities for those portals since the world was not created yet, so I had to do a workaround where I would add the positions for the to-be-generated portal entities to a static queue and then have the system create all the entities once the game loaded.

flo:
How about spawning them at the `OnChunkGenerated` event like in the WildAnimals module:

https://github.com/Terasology/WildAnimals/blob/master/src/main/java/org/terasology/wildAnimals/WildAnimalsSpawnSystem.java

MaxBorsch:
@flo That would work I think, still not ideal for once-in-a-world events but in most cases I think it makes sense.

skaldarnar:
> How about spawning them at the OnChunkGenerated event [..]

@flo I tried to point out that this is not feasible if the entity affects more than the chunk it is in. The dome in LaS is always active which keeps it alive when the chunk is unloaded, but you cannot enforce it to be generated in the first place. 
I briefly looked for a way to force (or at least schedule) a specific chunk (or position) or generation, but couldn't find anything in that direction. 

Please re-open if you think this needs further discussion for this use case...

PR #2767, opened 2017-01-29
Optional module dependencies solution for issue #2754

### LINKED ISSUE #2754: Optional module dependencies
Modules can have optional dependencies by adding the flag `"optional" : true`.
However, this is not handled when component systems are registered and likely in some other places too.

I will take the [EquipmentSmithing](https://github.com/Terasology/EquipmentSmithing) module as example:

The module will register two systems: [RegisterEquipmentSmithingRecipes](https://github.com/Terasology/EquipmentSmithing/blob/master/src/main/java/org/terasology/equipmentSmithing/system/RegisterEquipmentSmithingRecipes.java) and [ForgingStationWindowSystem](https://github.com/Terasology/EquipmentSmithing/blob/master/src/main/java/org/terasology/equipmentSmithing/ui/ForgingStationWindowSystem.java) to wire the equipment module to the smithing module.

This could be integrated into smithing or equipment if we were able to register systems based on a condition, e.g. if a module is present on the classpath. I am __not__ saying that's the way we should structure our modules, the wiring modules have the benefit of keeping the original modules smaller.
But there may be other cases where the module integration is a 1-liner in one system, which would be too much for one module and could go in a single package like `org.terasology.smithing.optional.equipment.*`.

I'd like to write something like `@RegisterSystem(requiresOptional = {"ModuleA", "ModuleB"}))` which will only register the component system if the optional modules are present in the current environment.
As the component systems are the main entry points to wire modules together, this would allow us to loosely couple modules together and keep some adapter logic directly in the parent modules.

I think this can be achieved with very little code changes at `org.terasology.engine.ComponentSystemManager.loadSystems()`

Small test scenario:
- create two modules `TestA` and `TestB`. Create an interface in TestA, then create a componentsystem in TestA that implements the interface and use `@Share` to share the service.
- Add TestA as optional dependency to TestB
- Implement a System in TestB that injects the service from TestA with `@In` and log something from the service to the console after startup.
- Start the game with only TestB enabled -> Game crash with a NoClassDefFoundError.

GabrielXia:
### Contains

"Fixes #2754"

### How to test
-  Clone two testing modules [OptionalDependencyTestA](https://github.com/GabrielXia/OptionalDependencyTestA) and [OptionalDependencyTestB](https://github.com/GabrielXia/OptionalDependencyTestB)
- I've set `OptionalDependencyTestB` has optional dependency  on `OptionalDependencyTestA`. And the system `ComponentSystemTestB` in module `OptionalDependencyTestB`is set as `@RegisterSystem(requiresOptional = {"OptionalDependencyTestA"})`
- Test in the case where both modules `OptionalDependencyTestA` and `OptionalDependencyTestB` are selected : we can find that the system `ComponentSystemTestB` is registered (it logs information in console when initializing)
- Test in the case  where only module TestB is selected : we find `ComponentSystemTestB` is not registered(the `ComponentSystemTestC` logs information about whether `ComponentSystemTestB` is registered)

- Modify in RegisterSystem annotation to get the optional dependency from this annotation, like `@RegisterSystem(requiresOptional = {"ModuleA", "ModuleB"})`
- Modify in `loadSystems()` in `ComponentSystemManager.java`, it only loads the component system whose optional modules are present in the classpath(in the ModuleEnvironment instance).


Cervator:
Thanks for the PR @GabrielXia !

Three requests on this one:

* You've got a bunch of new `.gitignore` files that shouldn't be needed as those `.iml` files should be covered by the `.gitignore` at the root directory (it has entries like `engine/engine.iml` in it) - could you double check that you really need those extras, and if needed maybe vary the ones at the root instead? May be a moot point as we're trying to get to the new `.idea` directory approach with IntelliJ soon anyway meaning all the `.iml` files will go away
* You've got a bunch of what seems like automated case changes switching `system` to `System` all over the place - stray refactoring request that applied to more than just a single file? Those should be reverted as they don't seem to relate to this PR (you can undo in a new commit, then optionally Git rebase and squash the undesired changes later)
* Can you please create two super simple modules (under your own GitHub account to start with) that contain the test cases for trying out the optional dependency setup, with clear logging indicating what happens under which circumstances? It would also be good to get the remaining two steps completed before review :-)

GabrielXia:
Hi @Cervator, It's very nice of you replying me ! 
- I have a question about `.gitignore` : I have deleted the new `.gitignore` files, but it seems that I have committed the `.iml` files while normally it should be ignored.
- I'm doing the test sample now. I have one question : should I create two repositories in my account, for example `TestA` and `TestB`, and then how can I commit these two modules ?

Thanks a lot !!!

Cervator:
No problem :-)

Yep now the `.iml` files need to be deleted - you *may* want to do that outside of IntelliJ so you can do the deletions + commit without IntelliJ recreating the files. After that you could `gradlew cleanIdea idea` one more time to have only the ones you need recreated. At that point you could try changing the `.gitignore` some to better exclude those files, or just personally ignore them by not re-adding to Git

On the modules yep two repositories then just commit and push to your repos. Anybody testing would then clone your repos locally to set them up as regular modules. Later on if we come up with a tutorial on the topic we might move those repositories under the Terasology org so they can be used as general examples.

GabrielXia:
Hi , I have added two testing modules in my account : [TestA](https://github.com/GabrielXia/OptionalDependencyTestA) and [TestB](https://github.com/GabrielXia/OptionalDependencyTestB), where TestB has optional dependency on TestA.

Thanks to tell me where is not good!!!    :-)

GabrielXia:
Heya, I have updated `How to test` in my first comment :-)

However, I don't know what to do to facilite the setting up of two test modules in my Github account :-(, any comments are welcomed 👍

engine/src/main/java/org/terasology/engine/ComponentSystemManager.java
```diff
@@ -84,7 +84,14 @@ public void loadSystems(ModuleEnvironment environment, NetworkMode netMode) {
             Name moduleId = environment.getModuleProviding(type);
             RegisterSystem registerInfo = type.getAnnotation(RegisterSystem.class);
             if (registerInfo.value().isValidFor(netMode, isHeadless)) {
-                systemsByModule.put(moduleId, type);
+                if(registerInfo.requiresOptional().length != 0){
+                    for(String s : registerInfo.requiresOptional()){
+                        if(environment.get(new Name(s))!=null)systemsByModule.put(moduleId, type);
+                    }
```
flo: woudln't it make sense to have the system register only if all dependencies are present?

(Also I am not sure if this is a good concept in general, but since some people want it  i won't object)

GabrielXia: Hi @flo , I'm sorry that I have made a stupid mistake. Now I've modified a little, it runs in multiple optional dependencies tests. 


( It may be not a good concept is that because it may make dependency relations unclear ? but I think this modification may help in the case where the module integration is a 1-liner in one system, as mentioned in #2767.)

engine/src/main/java/org/terasology/engine/ComponentSystemManager.java
```diff
@@ -85,6 +85,14 @@ public void loadSystems(ModuleEnvironment environment, NetworkMode netMode) {
             RegisterSystem registerInfo = type.getAnnotation(RegisterSystem.class);
             if (registerInfo.value().isValidFor(netMode, isHeadless)) {
                 systemsByModule.put(moduleId, type);
+                if(registerInfo.requiresOptional().length != 0) {
```
msteiger: I think it would be good to put this in a new method - also to clarify what it does exactly. I find it hard to follow why it inserts it first and removes it immediately after if a condition is met.

Could this become a boolean method that decides whether to insert it or not?

GabrielXia: Hi @msteiger , thank you for reviewing !!! 
I've added a method to decide wether to insert the system, could you please help me review it again ? What it does : 

- The method `ifRegister(RegisterSystem registerSystem,ModuleEnvironment enviromnent)` get a list of `ModuleName` that this system has option dependency on, using `RegisterSystem.requiresOptional()`
- If this list is length 0, so this system hasn't optional dependency, it should be registered(true)
- If there is any `ModuleName` aren't present, then it shouldn't be registered, return false
- If all these `ModuleName` are present in environment, then return true

I still have some questions :

- Should I change the name of this method ?
- Should I change the position of this method in class ?

Cervator:
Okay so this topic has gotten a fair bit of related discussion, but elsewhere - I decided to resurrect a forum thread that had some earlier thoughts on the matter, then added a recent Slack log on top :-)

http://forum.terasology.org/threads/optional-or-supported-modules-for-a-module.939/#post-14544

@GabrielXia: Let me know what you think about the forum thread, feel free to reply there. I am especially curious what sort of overall scope we might have for the feature vs. what this PR aims to achieve.

Also pinging @oniatus here for review

engine/src/main/java/org/terasology/engine/ComponentSystemManager.java
```diff
@@ -71,6 +71,14 @@ public ComponentSystemManager(Context context) {
...
     }
 
+    private boolean ifRegister(RegisterSystem registerSystem, ModuleEnvironment environment) {
+        if (registerSystem.requiresOptional().length == 0) return true;
+        for (String moduleName :registerSystem.requiresOptional()) {
+            if(environment.get(new Name(moduleName)) == null) return false;
+        }
+        return true;
+    }
+
```
oniatus: Could need a more expressive name and some formatting updates :)

```
    private boolean areOptionalRequirementsContained(RegisterSystem registerSystem, ModuleEnvironment environment) {
        if (registerSystem.requiresOptional().length == 0) {
            return true;
        }
        for (String moduleName : registerSystem.requiresOptional()) {
            if (environment.get(new Name(moduleName)) == null) {
                return false;
            }
        }
        return true;
    }
```

It is also a common pratice to define helper methods below their call so that a reader of the class can scroll downwards instead of jumping back and forth.
Therefore this should be below `loadSystems`

engine/src/main/java/org/terasology/entitySystem/systems/RegisterSystem.java
```diff
@@ -26,5 +26,6 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
 public @interface RegisterSystem {
+    String[] requiresOptional() default {};
```
oniatus: With the additional feature, the class could need some JavaDoc (It should have javadoc anyway 😄  )
Something like this?
```
/**
 * This annotation is used to mark a {@link ComponentSystem} to be registered in the environment.
 * <br><br>
 * A system can be conditionally registered, depending on optional dependencies by passing a list of optional module IDs:
 * <br>
 * <code>@RegisterSystem(value = RegisterMode.ALWAYS, requiresOptional = {"ModuleA","ModuleB"})</code>
 * <br>
 * In this case, the system would only be registered if both, <code>"ModuleA"</code> 
 * and <code>"ModuleB"</code> are contained in the environment.
 * <br><br>
 * By default, a system is registered with {@link RegisterMode#ALWAYS} and no optional requirements.
 */
```

GooeyHub:

PR #2768, opened 2017-01-29
Ping

msteiger:
Supersedes #2652 by refactoring the code into a new service class that both the in-game command and the menu GUI now use.

Quite a few Runnables and Callables here, but they all make sense (at least for me).

Feedback very welcome - first PR in ages :-)


GooeyHub:

emanuele3d:
Not quite my field of expertise, sorry... but it's good to have you back. :smiley:

iojw:
Thanks for this. 😄

Actually, the displaying of ping in the server list was just something added on the side. The main issue that the PR was initially targetting was the ability for clients to retrieve the client-server pings of all other clients connected to the multiplayer server, and then displaying this in the player list overlay. I wasn't familiar or confident enough to implement that one so I left it out in the end. The idea was that the server would ping all connected clients and then make this information available to the clients through some sort of subscriber system or event. Any comments or plans on implementing this one?

msteiger:
I believe that the server should perform the ping to all connected clients in the frame of a regular message. This could be done on a regular basis to discover stalled clients and also to serve ping stats to interested parties such as a client viewing the player overview screen.

@emanuele3d thanks :-)

flo:
Btw. creating and throwing connections away en mass looks like denial of service attacks. So a router inbetween might decide to act by blocking.

So what you suggest (using regular messages of the existing connections) can be much better.

Cervator:
Tested this out locally - seems to work fine :-)

Can we merge this now or do we want to change the connection / message setup? Is that more of a separate follow-up issue for later?

One minor side note that may not relate: On an server-client the `ping` console command complains it can't ping the server as single player doesn't count :D

msteiger:
I think that we can merge this with two follow-up issues:
1) Fix the error message for the ping console command (for single-player/server client)
2) Implement the ping for all clients. from the server

GooeyHub:

Cervator:
Submitted follow-up issue #2787 - just one as I suspect the ping command should probably just be clearer until that changes more later (split apart server/client more even in single player?)

Re-tested real quick, validates fine, merging! Thanks for getting back on PRs, hope you find time for some more! :-)

PR #2769, opened 2017-01-29
Remove Arabic - no translation present

msteiger:
Maybe the content was not transferred properly from Weblate?


GooeyHub:

Cervator:
Other way around I believe - we got a request in to *add* Arabic so a translator could work from within Weblate?

Looks like that was #2620

Weblate looks to have gone down again so I can't check if anything is there. It looks like its branch is getting pretty outdated at this point though.

msteiger:
I'm happy to add it as soon as we have the translation and the typescript. That said, I suggest removing the broken translation until then.

Cervator:
Here's a thought - and I think I noted this in another related PR/issue but I can't find it right now - would it be wise for us to include a config option to exclude a language from the drop-down specifically so we can add new ("empty") languages and the other language PRs missing fonts, to continue working on them (and letting them show up in Weblate for translator work) but not show in-game until they're suitably complete?

Pretty much give the language drop-down an extra logic hook to check a variable for excluded languages (could even simply be a constant in the relevant class) and if such a language file is found simply ignore it.

I think I'll submit a feature request for it, then we can work it separately and if somebody gets it fast enough we can get the pending language stuff out of the way :-)

Cervator:
#2784 got worked real fast and now excludes Arabic. Closing this as completed via workaround I guess :-)

One quirk: Arabic is a language without a single country to source a suitable flag icon from. I found a couple options via Google and on Wikipedia for ways to connect a flag to Arabic, but they seemed dubious. Either go with Arabic League or a typical design in more or less direct use by a bunch of countries but not a bunch of others. I went ahead and made a pure green flag instead - hopefully that'll work for now.

The language code `ar` for Arabic sort of conflicts with the country code `ar` for Argentina, but luckily they speak Spanish so we'll never need a language entry with that flag icon. Conflict avoided!

Issue #2764, opened 2017-01-28
NUI Editor: layoutInfo can be added via context menu to a layoutInfo


flo:
### What you were trying to do
I was checking out what options the context menu offers me when I right click a layoutInfo node in the NUI Editor.
### What actually happened
The context menu offered me the option to add a layoutInfo node to a layoutInfo node.
### How to reproduce
1. Open the NUI editor. To do so press ESC ingame and then click at the bottom right on "Developer Tools". Then select "NUI Editor Screen".
2. Select a screen (e.g. addServerPopup from the combo box at top left
3. Navigate to a layoutInfo node
4. Right click it and checkout what context menu option it offers in the "Add ..." sub menu. It offers to add a layoutInfo which is wrong.

rzats:
(Should have been closed by #2270, but wasn't - closing manually)

Issue #2763, opened 2017-01-28
NUI Editor: family node has wrong type


flo:
### What you were trying to do
I tried to give a UILabel in the NUI editor a family attribute with the value "title"
### What actually happened
A family node got added that has no value, but allows for the addition of a layoutInfo node.
### How to reproduce
1. Open the NUI editor. To do so press ESC ingame and then click at the bottom right on "Developer Tools". Then select "NUI Editor Screen".
2. Select a screen (e.g. addServerPopup from the combo box at top left
3. Navigate to the first UILabel
4. Remove the family node if it has one by selecting it and then pressing delete
5. Right click the UILabel node and select in the context menu "Add.." and in the sub menu "familiy".
6. Notice how the added node has no value and no way to set it. Check out what the context menu offers when you right click it.

flo:
@rzats: since you made it you probably know a quick fix for this.

flo:
Probably related bug: https://github.com/MovingBlocks/Terasology/issues/2764

rzats:
(Should have been closed by #2270, but wasn't - closing manually)

PR #2770, opened 2017-01-30
Fix NUI editor issues

rzats:
### Contains

Fixes for a few minor NUI editor issues, specifically #2763 and #2764 - read these for a nice detailed description. Also adds an autosave file created by the editor to `.gitignore`, since we never want to include it in git.

### How to test

Attempt to reproduce the bugs from #2763 and #2764, then verify that you can't reproduce them because they've been fixed!

No outstanding issues! /cc @flo

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/editor/utils/NUIEditorMenuTreeBuilder.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/editor/utils/NUIEditorMenuTreeBuilder.java
@@ -28,6 +28,7 @@ import org.terasology.rendering.nui.AbstractWidget;
 import org.terasology.rendering.nui.Color;
 import org.terasology.rendering.nui.LayoutConfig;
 import org.terasology.rendering.nui.NUIManager;
+import org.terasology.rendering.nui.UILayout;
 import org.terasology.rendering.nui.UIWidget;
 import org.terasology.rendering.nui.contextMenu.MenuTree;
 import org.terasology.rendering.nui.databinding.Binding;
@@ -91,7 +92,7 @@ public class NUIEditorMenuTreeBuilder {
 
         // Create the ADD_EXTENDED level.
         if ((type == JsonTreeValue.Type.ARRAY && !"contents".equals(node.getValue().getKey()))
-            || type == JsonTreeValue.Type.OBJECT) {
+                || type == JsonTreeValue.Type.OBJECT) {
             MenuTree addTree = createAddContextMenu(node);
             primaryTree.addSubmenu(addTree);
         }
@@ -107,8 +108,8 @@ public class NUIEditorMenuTreeBuilder {
 
         // Unless the node is an OBJECT child of an ARRAY (should always have an empty key), add the edit option.
         if (type != JsonTreeValue.Type.NULL && !(type == JsonTreeValue.Type.OBJECT
-            && !node.isRoot()
-            && node.getParent().getValue().getType() == JsonTreeValue.Type.ARRAY)) {
+                && !node.isRoot()
+                && node.getParent().getValue().getType() == JsonTreeValue.Type.ARRAY)) {
             primaryTree.addOption(OPTION_EDIT, externalConsumers.get(OPTION_EDIT), node);
         }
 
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/editor/utils/NUIEditorMenuTreeBuilder.java
```diff
@@ -250,7 +252,7 @@ private void populateContextMenu(JsonTree node, MenuTree addTree, boolean isSkin
                 // If the node is part of a layout, add an option to add the layoutInfo node.
                 if (nodeInfo.getLayoutClass() != null) {
                     String layoutInfo = "layoutInfo";
-                    if (!node.hasChildWithKey(layoutInfo)) {
+                    if (!Objects.equals(node.getValue().getKey(), layoutInfo) && !node.hasChildWithKey(layoutInfo)) {
```
flo: There are other cases too where we don't want to offer a layoutInfo as option that can be added.

e.g. the context menu for a position-horizontal-center still offers with that change to add a "layoutInfo" which is similarly wrong.
Some classes to look at: 
* UIWidgetTypeAdapter.deserialize for the actual way of how layoutInfo gets used
* NUIEditorNodeUtils.getNodeInfo for logic you wrote that does something in the direction of detecting what is what.

rzats: Right, so in this case having a "whitelist" of nodes that can have `layoutInfo` added to them would make more sense. Can you think of any criteria?

flo: Since you already have a way to calculate the possible child nodes for layoutInfo. Maybe the criteria for adding a layoutInfo node could be that it is possible to add child nodes to it. So that it is not possibe to add layoutInfo nodes that can't have any child.

Otherwise based on UIWidgetTypeAdapter the logic I think the condition for adding a layoutInfo node should be:
* The node represents a UIWidget
* The parent is `UIFormat.CONTENTS_FIELD`
* The parent of the parent implements UILayout
* The hint type of the layout can be instanced (e.g. no layoutInfo for CardLayout)

I added comments to the following code part of `UIWidgetTypeAdapter` and added comments what it means for our condition:
```
if (UILayout.class.isAssignableFrom(elementMetadata.getType())) { // => parent of parent must implement UILayout
                UILayout<LayoutHint> layout = (UILayout<LayoutHint>) element;

                Class<? extends LayoutHint> layoutHintType = (Class<? extends LayoutHint>)
                    ReflectionUtil.getTypeParameter(elementMetadata.getType().getGenericSuperclass(), 0);
                if (jsonObject.has(CONTENTS_FIELD)) { // parent => is CONTENTS_FIELD
                    for (JsonElement child : jsonObject.getAsJsonArray(CONTENTS_FIELD)) {
                        UIWidget childElement = context.deserialize(child, UIWidget.class); // => element is UIWidget
                        if (childElement != null) {
                            LayoutHint hint = null;
                            if (child.isJsonObject()) {
                                JsonObject childObject = child.getAsJsonObject();
                                if (layoutHintType != null && !layoutHintType.isInterface() && !Modifier.isAbstract(layoutHintType.getModifiers()) // => layoutHintType can be instanced (no interface no abstract class)
                                    && childObject.has(LAYOUT_INFO_FIELD)) { 
                                    hint = context.deserialize(childObject.get(LAYOUT_INFO_FIELD), layoutHintType);
                                }
                            }
                            layout.addWidget(childElement, hint);
                        }
                    }
                }
            }
```

rzats: Couldn't manage to retrieve `layoutHintType` - for some odd JVM-y reason it resolves to null every time. All the other criteria should now be met, though.

GooeyHub:

flo:
Thanks for improving the condition further.

There is one thing that is still missing but it is not dramatic: You don't check if the layout hint is instance able so it offers you still wrongly the option to add a layoutInfo to a ColumnLayout.

However I think it is good enough now to be merged. That special case can be handled in a separate PR if you want to address it.

Issue #2677, opened 2016-12-18
Rendering Chinese characters properly in-game may require additional font support


Cervator:
With #2666 merged we have Chinese translations but they do not render in-game (for instance in the language drop-down. I wonder if that's because there is a whole other Noto font family for that character set: https://www.google.com/get/noto/help/cjk/

If we need that to be able to properly display Chinese we may need to build in the possibility to do a full font swap when changing language. Although I wonder what that'll do for untranslated fields? 

We also need to add a Chinese flag icon and anything else for the language to be complete, and to push it to the weblate branch.

smsunarto:
Added flag and change localisation file name to follow ISO 639-1 convention #2697

msteiger:
I think that having this Chinese dropdown entry should be fixed/removed asap, because it's broken.

I'd like to add typescript support, if possible. What is the smallest subset of character glyphs? It looks like [the Table of General Standard Chinese Characters](https://en.wikipedia.org/wiki/Table_of_General_Standard_Chinese_Characters) still covers over 8000 glyphs.  Do we need them all? Ping @smsunarto

smsunarto:
@msteiger We won't need them all, chinese (Mandarin) doesn't use a system similar to alphabet hence their crazy amount of glpyhs since every word have their own unique glyph. Ideally we should only use the glpyh that is used by the translation. But I'm pretty sure this one should suffice https://en.wikipedia.org/wiki/List_of_Commonly_Used_Characters_in_Modern_Chinese really depends on whether or not the translator uses complex words though.

Cervator:
Here's perhaps an odd thought: should we merge the outstanding languages even without proper font support and just leave them disabled in the drop-down somehow? So we can get the PRs out of the way, keep working on them via Weblate if needed, and focus down to a single effort on better supporting fonts?

msteiger:
Sounds reasonable. I wonder whether it should be in the main project or should live in a module.

smsunarto:
@msteiger I'm pretty sure glyph support is an integral part of the system. Im not sure how we can activate a new module outside of the world creation menu too.

MarkDerosier:
To overcome this, we can first move some of the text into icons (a wrench for settings, a globe for language, etc), then use a simpliefied character set. (simplified Chinese vs traditional), and then to really save the space, all text is actually vectors. 

Translate our words, load the font up into inkscape, make the labels, then convert font to path. 

[...]

PR #2777, opened 2017-02-07
CrashReporter improvement : IssueReporter ?

### LINKED ISSUE #2765: CrashReporter improvements: Full reporting system?
This is a follow-up to #2577 which was completed by #2584

At this point you can launch the CrashReporter from inside the game without actually having crashed so you can easily access the game logs from the running session. However, this is still a bit awkward and looks funny since the CR was, well, designed specifically to handle crashes. There are some short term possibilities to tidy up the setup as well as some long term possibilities that could be GSOC scope for an overall problem reporting system..

Short term easy fixes:

- [x] Rename the button to "Issue Reporter" (for the in-game context it isn't for reporting crashes)
- [x] Add the button anywhere else needed / move it under Developer Tools? Rename that to "Extras" or so?
- [x] At present the CrashReporter opens in a modal window mode locking up the game itself while open, until the user clicks the `X` to close it again. Could it perhaps be left non-modal, especially in a multiplayer situation? If a player is hosting a server it goes unresponsive if they open the CR. If non-modal stream log updates right into the log tabs?
- [ ] The ~Crash~ IssueReporter now opens in the background, which can be confusing (should go to the front, just not be modal)
- [x] Improve the CrashReporter itself to adopt a larger scope beyond just crashes: https://github.com/MovingBlocks/CrashReporter/issues/35
- [ ] Clarify the `-noCrashReporter` command line argument. It now only matters for actual crashes - opening the CR via menu item is something totally different

Longer term possibilities (GSOC stuff):

- Relation to #1402 - steal @iojw's improvement from https://github.com/MovingBlocks/TerasologyLauncher/pull/392 to also apply in the engine (warning icon showing problems may be active). But could that actually be a new way of opening the CrashReporter in a non-modal mode, which could also be retrofitted to the launcher and elsewhere?
- Implement an outright back-end supported issue reporting system beyond just dumping logs on pastebin or a Google Drive. This would probably involve some existing framework - need to research what's available that could work for us (think a user support style system like ZenDesk, but naturally it has to be at least free and ideally open source. Bonus for integrating with GitHub)
- Implement some metrics gathering users could opt-in for to better understand how the game (or other projects) is being used. Could include how frequently certain warnings or error conditions happen. Could include a more full picture of the user's hardware and such (GPU driver versions?)

As a clarification vs #1402 this item should focus on the support for better reporting issues, not avoiding them in the first place via better visibility (the stated goal of #1402). However the two are quite close, especially if the CR becomes a more generic review tool you can access in-game.

GabrielXia:
### Contains

Fixes #2765 

### How to test

1. Rename the button to "Issue Reporter".

- However I still don't know why my other text on button disappear. Appreciate badly someone can tell me the reason 👍 
- I just changed the text value in `.lang` file but haven't changed others, like `"crash-reporter": "Issue Reporter"`, I'm wondering if it's better change all "crash reporter" to "issue reporter", like `"Issue-reporter": "Issue Reporter"`
2. I've moved this button under `Developer Tools` menu

- I'm wondering if it's necessary rename `Developer Tools` button to `Extras`

- [ ] Release CrashRreporter v4.1.0 and bump the Terasology dependency to match it 
- [ ] I've tried to set crash reporter window to non modal windows simply by setting set `jDialog` to modeless type, sadly log doesn't update @Cervator  What's more, in the non modal windows, you can reopen this windows, and there can be more than one CR windows :( Other than making this window to update log, I think we should think about how to make sure there is only one CR window, like if CR window is open, press the CR button to make this windows in front of all other windows.
- [ ] I think that there are some texts to be changed from "Crash Reporter" to "Issue Reporter" : 

-  Ah it's mentioned in [MovingBlocks/CrashReporter#35](https://github.com/MovingBlocks/CrashReporter/issues/35)

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
@@ -33,7 +33,7 @@ public class PauseMenu extends CoreScreenLayer {
         WidgetUtil.trySubscribe(this, "settings", widget -> getManager().pushScreen("settingsMenuScreen"));
         WidgetUtil.trySubscribe(this, "mainMenu", widget -> CoreRegistry.get(GameEngine.class).changeState(new StateMainMenu()));
         WidgetUtil.trySubscribe(this, "exit", widget -> CoreRegistry.get(GameEngine.class).shutdown());
-        WidgetUtil.trySubscribe(this, "devTools", widget -> getManager().pushScreen("devToolsMenuScreen"));
         WidgetUtil.trySubscribe(this, "crashReporter", widget -> CrashReporter.report(new Throwable("There is no error."), LoggingContext.getLoggingPath(), CrashReporter.MODE.ISSUE_REPORTER));
+        WidgetUtil.trySubscribe(this, "devTools", widget -> getManager().pushScreen("devToolsMenuScreen"));
     }
 }
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
@@ -15,9 +15,7 @@
  */
 package org.terasology.rendering.nui.layers.ingame;
 
-import org.terasology.crashreporter.CrashReporter;
 import org.terasology.engine.GameEngine;
-import org.terasology.engine.LoggingContext;
 import org.terasology.engine.modes.StateMainMenu;
 import org.terasology.registry.CoreRegistry;
 import org.terasology.rendering.nui.CoreScreenLayer;
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
@@ -33,7 +33,6 @@ public class PauseMenu extends CoreScreenLayer {
         WidgetUtil.trySubscribe(this, "settings", widget -> getManager().pushScreen("settingsMenuScreen"));
         WidgetUtil.trySubscribe(this, "mainMenu", widget -> CoreRegistry.get(GameEngine.class).changeState(new StateMainMenu()));
         WidgetUtil.trySubscribe(this, "exit", widget -> CoreRegistry.get(GameEngine.class).shutdown());
-        WidgetUtil.trySubscribe(this, "crashReporter", widget -> CrashReporter.report(new Throwable("Report an error."), LoggingContext.getLoggingPath()));
         WidgetUtil.trySubscribe(this, "devTools", widget -> getManager().pushScreen("devToolsMenuScreen"));
     }
 }
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
@@ -15,7 +15,9 @@
  */
 package org.terasology.rendering.nui.layers.ingame;
 
+import org.terasology.crashreporter.CrashReporter;
 import org.terasology.engine.GameEngine;
+import org.terasology.engine.LoggingContext;
 import org.terasology.engine.modes.StateMainMenu;
 import org.terasology.registry.CoreRegistry;
 import org.terasology.rendering.nui.CoreScreenLayer;
@@ -32,5 +34,6 @@ public class PauseMenu extends CoreScreenLayer {
         WidgetUtil.trySubscribe(this, "mainMenu", widget -> CoreRegistry.get(GameEngine.class).changeState(new StateMainMenu()));
         WidgetUtil.trySubscribe(this, "exit", widget -> CoreRegistry.get(GameEngine.class).shutdown());
         WidgetUtil.trySubscribe(this, "devTools", widget -> getManager().pushScreen("devToolsMenuScreen"));
+        WidgetUtil.trySubscribe(this, "crashReporter", widget -> CrashReporter.report(new Throwable("There is no error."), LoggingContext.getLoggingPath(), CrashReporter.MODE.ISSUE_REPORTER));
     }
 }
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/MainMenuScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/MainMenuScreen.java
@@ -56,7 +56,7 @@ public class MainMenuScreen extends CoreScreenLayer {
         WidgetUtil.trySubscribe(this, "settings", button -> triggerForwardAnimation(SettingsMenuScreen.ASSET_URI));
         WidgetUtil.trySubscribe(this, "credits", button -> triggerForwardAnimation(CreditsScreen.ASSET_URI));
         WidgetUtil.trySubscribe(this, "exit", button -> engine.shutdown());
-        WidgetUtil.trySubscribe(this, "crashReporter", widget -> CrashReporter.report(new Throwable("Report an error."), LoggingContext.getLoggingPath()));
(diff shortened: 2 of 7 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/MainMenuScreen.java
```diff
@@ -56,7 +56,7 @@ public void initialise() {
         WidgetUtil.trySubscribe(this, "settings", button -> triggerForwardAnimation(SettingsMenuScreen.ASSET_URI));
         WidgetUtil.trySubscribe(this, "credits", button -> triggerForwardAnimation(CreditsScreen.ASSET_URI));
         WidgetUtil.trySubscribe(this, "exit", button -> engine.shutdown());
-        WidgetUtil.trySubscribe(this, "crashReporter", widget -> CrashReporter.report(new Throwable("Report an error."), LoggingContext.getLoggingPath()));
+        WidgetUtil.trySubscribe(this, "crashReporter", widget -> CrashReporter.report(new Throwable("Report a error."), LoggingContext.getLoggingPath(), false));
```
Cervator: I believe "Report an error" is more correct in this case than "Report a error" - English is weird that way. You can *generally* go with `an` if the next word starts with a vovel and otherwise `a` but there are exceptions ...

GabrielXia: Hey, I'm just thinking about deleting all the text "Report a error" or changing it, because there is actually no error in issue reporter.

GabrielXia:
Problem found ! It's from system language setting, seems that Chinese support doesn't work well ! :(

Cervator:
@GabrielXia just responded with a big comment over on #2677 - just to clarify, you picked Chinese yourself under the Player Settings, right? It didn't get picked by the game automatically from simply running on your system? I'd think it would start out English in any case :-)

msteiger:
Why are you removing the button from the in-game menu?

GooeyHub:

Cervator:
> Why are you removing the button from the in-game menu?

@msteiger that one is on me, I raised it as an idea over in #2765 but it was honestly more of a stray thought / suggestion I had hoped to get some more opinions on first :-) 

To be clear the button just moved one level deeper, it didn't go away entirely. But I am curious what the most ideal setup there would be. Should reporting an issue be right there as a top level button users can easily spot, or grouped with other stuff under "Extras" ?

I would love to instead see the thing I mentioned over in #2765 about what @iojw did in the CR, so we can have a little warning icon in a bottom corner instead of a full button. Make it highlight if anything has occurred we think might be worth reporting. Otherwise it is just a small icon hiding in a corner, not adding much to clutter.

GabrielXia:
Hello :-) I've moved this button to the Developer Tools. I think the issue reporter might rather be a tool for developer, but you are right, put it in in-game menu is more convenient for user to check. Maybe we can ask @Cervator here to discuss about where we put this button :-)

p.s. Why there is a red cross ? Scared ...

Cervator:
The red x is because the build in Jenkins failed - in this case that's just the unfortunate reality of making a change that requires a new version of an external project like the CrashReporter. Since the engine build test uses the latest CR without your change there is a simple syntax error where the new thing is missing :-)

Poking @rzats for an additional opinion on the button since he does UI stuff!

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/MainMenuScreen.java
```diff
@@ -27,6 +27,8 @@
 import org.terasology.rendering.nui.widgets.UILabel;
 import org.terasology.version.TerasologyVersion;
 
+import javax.print.attribute.standard.Severity;
```
rzats: The enum you're using is intended for a printer API and currently _happens_ to contain the three values we need, but can be arbitrarily changed in the future. Can you replace it with a custom enum located within the `CrashReporter` so that we "own" the code?

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
```diff
@@ -33,7 +31,6 @@ public void initialise() {
         WidgetUtil.trySubscribe(this, "settings", widget -> getManager().pushScreen("settingsMenuScreen"));
         WidgetUtil.trySubscribe(this, "mainMenu", widget -> CoreRegistry.get(GameEngine.class).changeState(new StateMainMenu()));
         WidgetUtil.trySubscribe(this, "exit", widget -> CoreRegistry.get(GameEngine.class).shutdown());
-        WidgetUtil.trySubscribe(this, "crashReporter", widget -> CrashReporter.report(new Throwable("Report an error."), LoggingContext.getLoggingPath()));
```
rzats: I'd keep the button in the in-game menu - it's fairly well-hidden as is, no reason to obscure it.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -77,7 +77,7 @@
  * <tr><td>-headless</td><td>Start headless.</td></tr>
  * <tr><td>-loadlastgame</td><td>Load the latest game on startup.</td></tr>
  * <tr><td>-noSaveGames</td><td>Disable writing of save games.</td></tr>
- * <tr><td>-noCrashReport</td><td>Disable crash reporting.</td></tr>
+ * <tr><td>-noCrashReport</td><td>Disable crash reporting when having a crash.</td></tr>
```
rzats: No need to change the documentation - it still makes sense (the option disables crash reporting, but not issue/feedback reporting)

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
Cervator: Yeah I'm in no rush to move stuff around. More curious what we should be aiming for :-)

facades/PC/src/main/java/org/terasology/engine/Terasology.java
Cervator: My thought was emphasizing the flag doesn't disable the usage of the CR for issue reporting (just on triggering actual crashes) ... but yeah this line as written kinda ends up working anyway, since the new option is indeed more about issue/feedback reporting, not crashes :-)

GooeyHub:

GooeyHub:

GooeyHub:

oniatus:
https://github.com/MovingBlocks/CrashReporter/pull/36 has been merged some time ago. Is there something outstanding here except the new release and version bump?
The conflict in the language file seems to be pretty small.

GabrielXia:
Thanks for keeping on this, I think it's ok for the moment. :-)
The enhancement of CR could be:
- Show warning icon somewhere in the game when there is an error
- Based on the telemetry system, maybe a new page showing all the information sent to the server

skaldarnar:
I've just released a new version of CR (4.1.0). We can probably merge this after the next Alpha release.

@GabrielXia could you please update the dependency in this PR?

GooeyHub:

GabrielXia:
I'm so glad to see this to be merged :-)

@skaldarnar which dependency do you refer to? Is that the exclusion of `jackson` in CR: https://github.com/MovingBlocks/CrashReporter/pull/39 or the one to import `snowplow`: https://github.com/MovingBlocks/Terasology/pull/2968 ?

Cervator:
I went ahead and did the version bump (just the CR itself in the engine's `build.gradle` from 4.0.1 to 4.1.0) while merging. Conflicts were benign as expected and it tested out fine :-)

0shine0:
@Cervator There seems to be a problem with my build. Although you bumped the version of CR, do i need to do anything else manually like download the new version jar for CR?

skaldarnar:
@0shine0 running `gradlew idea` or `gradlew jar` should put you on the safe side and download the updated dependency. Be sure that you don't have the Crash Reporter checked out as source in case you run into problems.

PR #2735, opened 2017-01-05
Added Farsi

mdj117:
Well that took a while, added almost 100% of the translated version of Terasology in Farsi
The flag is named flag_fa.png and the file is menu_fa.lang

PR #2784, opened 2017-02-11
Add an exclusion list for unready languages to the player settings

### LINKED ISSUE #2781: Add an exclusion list for unready languages to the player settings
This one should be easy!

Not all our supported languages are ready enough to be shown to players - this can happen when we get new languages that need a custom font or somebody requests a new language be set up in Weblate yet it can't be worked on there before we actually add an "empty" language file (which would promptly show up in-game, empty).

I think it might be sensible for us to add the ability to exclude one or more languages until they are ready. This could be done as simply as adding a list variable in `PlayerSettingsScreen.java` and varying the language lookup block inside `initialise` to check if a scanned language is to be excluded.

Three things to be careful about:

* The exact code/name of the language in the exclusion variable should match the format we get during scanning. Maybe both could be using `Locale` ?
* Maybe there would be a better way than hard coding the exclusion list inside `PlayerSettingsScreen.java` - it might be worthwhile to have it be more configurable, however if we add it to config.cfg we can't re-enable a language for any user who already has a `config.cfg`
* If we do this it needs to be *very carefully* spelled out in documentation for contributors testing languages ("Go remove language x from the exclusion list when it is ready for testing")

If this results in a PR soonish we could resolve/merge #2733, #2769, #2745, #2735 *temporarily* until we get needed font support to actually display them in-game. A few other languages like Chinese have also already been merged yet do not work (see #2677) - so they should be excluded until ready.

Alternatively for the new "empty" language scenario I wonder if Weblate supports a possibility to "park" a translation inside its app without us first having to manually add the right language files, letting an interested translator get started immediately.

GabrielXia:
### Contains

Fixes #2781 

### How to test

- I've added Hindi #2733, Arabic #2769, Korean #2745, Farsi #2735, Chinese #2677 to the exclusion list.
- I've tested in Hindi and Chinese.
Before the change, Chinese appears in the list : 
screen shot 2017-02-11 at 1 15 35 pm
After the change, Chinese disappeared : 
screen shot 2017-02-11 at 1 14 23 pm

- [ ] I'm not sure about the best place to put documentation for contributors testing languages, for now I put it before the declaration of `languagesExcluded`
- [ ] We might have to consider the scenario : when a user's system language is an unready language  e.g. Chinese, when  this user start the game, there will be problems in menu text(as mentioned in #2777). We might need to add an exclusion list in an other class. Now I still don't find that class, please tell me if someone knows :-)

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/settings/PlayerSettingsScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/settings/PlayerSettingsScreen.java
@@ -58,6 +58,12 @@ public class PlayerSettingsScreen extends CoreScreenLayer {
 
     private final List<Color> colors = CieCamColors.L65C65;
 
+    /**
+     * Remove language x from this languagesExcluded table when it is ready for testing
+     */
+    private final Locale[] languagesExcluded = {Locale.forLanguageTag("zh"), Locale.forLanguageTag("hi"),
+            Locale.forLanguageTag("ar"), Locale.forLanguageTag("ko"), Locale.forLanguageTag("fa")};
+
     private UIText nametext;
     private UISlider slider;
     private UISlider heightSlider;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Thanks for this! Glad it was that easy :-)

Interesting note with #2777. I didn't think the game would automatically launch in the OS language, huh.

@GooeyHub: add to whitelist

GooeyHub:

PR #2785, opened 2017-02-11
Work on render nodes: exposure and bloom

emanuele3d:
### Contains
- Continues refactoring work of the renderer, to take advantage of the new node-based architecture.
- See commit descriptions below for an overview of the changes

### How to test
- Run Terasology
- Create a new game or load an existing one
- Verify nothing has changed in the rendering, especially when activating/deactivating Eye Adaptation and Bloom.

### Notes:
- While the bloom effect is in now taking advantage of the new architecture... in practice it doesn't seem to work. This seems to be a pre-existing issue however and I won't worry about fixing it now.
- In general a number of problems have crept into the renderer while the overhaul is taking place no matter our best intention to keep everything working. 
- These problems will be solved during planned additional refactoring passes through all the nodes in the renderer. These refactoring passes are anyway needed as work on the later nodes has shown ways to improve the code in earlier nodes.
- Furthermore, an overhaul of the visual debug system (showing the content of a number of FBOs) is also planned and will help pinpoint problems that have visual consequences.


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/BloomBlurNode.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2016 MovingBlocks
```
Cervator: We've been in 2017 for like weeks now, man! :D

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerForExposureNode.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.terasology.assets.ResourceUrn;
+import org.terasology.config.Config;
+import org.terasology.config.RenderingConfig;
+import org.terasology.registry.In;
+import org.terasology.rendering.opengl.FBO;
+import org.terasology.rendering.opengl.FBOConfig;
+
+/**
+ * Extends the DownSamplerNode class purely to:
```
Cervator: As a tiny best practice note it is ideal to have a full sentence ended with a `.` here so it can parse out cleanly for the one-sentence summary used on some javadoc pages. In this case I dunno if we'd end up with something like:

`DownSamplerForExposureNode | Extends the DownSamplerNode class purely to:`

As the one liner explanation of the class in some places :-) Or maybe it would multi-line until it finds a . somewhere (in this case the whole thing)

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
```diff
@@ -0,0 +1,121 @@
...
+    private FBOConfig outputFBOconfig;
+    private FBO outputFBO;
+    private Material downSampler;
+
+    /**
+     * Throws a RuntimeException if invoked. Use initialise(...) instead.
+     */
+    @Override
+    public void initialise() {
+        throw new RuntimeException("Please do not use initialise(). For this class use initialise(...) instead.");
```
Cervator: This seems like something to cover differently ... but I don't really have any ideas or much desire to worry about that just yet, knowing there's still plenty of nodes to play with. Patterns will become clearer later :-)

engine/src/main/java/org/terasology/rendering/dag/nodes/BloomBlurNode.java
emanuele3d: My apologies. There's something in IntelliJ about updating the copyright notice on commit. Should I have that enabled? Or should I just do it manually?

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
emanuele3d: Yes, it's not nice. There was some thought behind it, in the sense that this was the best either I could come up with or both tdgunes and I could come up with - I don't remember at what stage it was introduced.

I think it has to do with the fact that the superclass has an initialise() method that needs to be overridden, but this particular class (and others like it) need initialization parameters that vary from class to class. In this context, to create a NodeFactory specific for each of these type of nodes seemed overkill and we settled for a slightly improved NodeFactory capable of delaying initialization if necessary. This allow for a class-specific initialise(...) method to be called after the NodeFactory has instantiated (but not initialised) a node. The WorldRendererImpl class has a number of example usages of this in the initRenderGraph method.

Ultimately I'd welcome suggestion: it's not exactly elegant.

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerForExposureNode.java
emanuele3d: Understood. I'll try and come up with something that takes this in account.

engine/src/main/java/org/terasology/rendering/dag/nodes/BloomBlurNode.java
Cervator: It is a nitpick, in large part just so I'd have something to say in a review ... :D

I checked out the IDE customization in Gradle (`/config/gradle/ide.gradle`) and remembered we do have a copyright template there with a hard set year of 2016. So after we fix that newly generated copyright headers should appear with the right year. In the meantime it is just in case you notice an outdated header go ahead and fix. But trivial importance :-)

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
indianajohn: Isn't this the classic case of checked vs. unchecked exceptions? Since the programmer is never meant to run `subClassMethod.initialize()` in this case, by making `Node.initialize()` and `DownSamplerNode.initialize()` throw a child of `Exception` (Maybe `NoSuchMethodException`?), the Java compiler will throw an error if somewhere in the code somebody tries to call the `initialize()` method on a child of `DownSamplerNode`. This will avoid having to watch the program crash and read the stack trace at least once for a future programmer trying to construct a child of `DownSamplerNode` with the wrong constructor.

indianajohn: Never mind, I see what you mean about the factory now. Disregard the java novice over here.

emanuele3d: Don't worry @indianajohn, in this particular context (and in many others) I'm also a java novice. :blush:

emanuele3d:
@indianajohn: thank you for the feedback. Indeed having smaller, simpler nodes is one of my aims. It appears switching shaders on modern hardware is not as horribly expensive as I thought it would be and in later PRs I will try to get deeper into the shaders and break those too, as some shaders do a few too many things for my taste.

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/BloomBlurNode.java
```diff
@@ -0,0 +1,72 @@
...
+    private static final float BLUR_RADIUS = 12.0f;
+
+    @In
+    private Config config;
+
+    @In
+    private DisplayResolutionDependentFBOs displayResolutionDependentFBOs;
+
+    /**
+     * Initializes a BloomBlurNode instance.This method must be called once shortly after instantiation
```
tdgunes: A space after the period?

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerForExposureNode.java
```diff
@@ -0,0 +1,57 @@
...
+ * encode the average brightness of the rendering, which in turn is used to tweak the exposure parameter later nodes use.
+ */
+public class DownSamplerForExposureNode extends DownSamplerNode {
+
+    public static final FBOConfig FBO_16X16_CONFIG = new FBOConfig(new ResourceUrn("engine:fbo.16x16px"), 16, 16, FBO.Type.DEFAULT);
+    public static final FBOConfig FBO_8X8_CONFIG = new FBOConfig(new ResourceUrn("engine:fbo.8x8px"), 8, 8, FBO.Type.DEFAULT);
+    public static final FBOConfig FBO_4X4_CONFIG = new FBOConfig(new ResourceUrn("engine:fbo.4x4px"), 4, 4, FBO.Type.DEFAULT);
+    public static final FBOConfig FBO_2X2_CONFIG = new FBOConfig(new ResourceUrn("engine:fbo.2x2px"), 2, 2, FBO.Type.DEFAULT);
+    public static final FBOConfig FBO_1X1_CONFIG = new FBOConfig(new ResourceUrn("engine:fbo.1x1px"), 1, 1, FBO.Type.DEFAULT);
+
```
tdgunes: I am really glad to see that these are properly and explicitly named, as opposed to `downsampledScene` array.

emanuele3d:
@Cervator, I think we are good to go with this one...

GooeyHub:

Cervator:
Roger that! Merged latest after a quick test :-)

Also took the chance to switch the IntelliJ copyright header year to 2017 real quick (should apply after the next `gradlew idea`)

PR #2789, opened 2017-02-15
Extract and Overhaul inventory

jellysnake:
See also [The repo for the extracted files](https://github.com/jellysnake/JellyInventory).

This pull request serves to remove the inventory from core and into it's own module. It also refactors both inventory and items.

## Outstanding before merging

- [ ] Make held items visible in players hand
- [ ] Make inventory persistent upon saves
- [ ] Make sure inventory is network compatible
- [ ] Add ability to drag items around inventory
- [ ] Make stacking more reliable
- [ ] Add scrollwheel as a method to select slots


GooeyHub:

flo:
1.) I would prefer the module being part of the main source repository analogue to Core since it is a very basic functionality. At others: What do you think about that?

2.) I think it would be a good idea to state what the purpose/advantages of this PR are. e.g. what advantages do we have from this change. A reason against it I can think of would be that when you try the game out without modules you get something that misses even more. A reason for this change could be an alternative inventory implementation. Do we have/need that?

3.) Also it would be good if you could say if you have made any changes that could have broken other modules. If your PR results in a lot of effort in fixing of other modules then we should evaluate if that change is really worth the effort that it causes. So that's why point 2 is so important. If it does not break any module then you should also highlight that in your PR as it would be a huge plus.

emanuele3d:
Responding to @flo:

1. sounds reasonable to me, as long as the module can be disabled if needed.

2. I can certainly envision alternative inventories. I.e. instead of a single inventory a player could have multiple ones: one for items worn on the body (clothes, armor, jewels) and one for each container he's carrying (pockets, backpack). Each inventory could have separate characteristics and implementations, i.e. with some keeping track of the volume and weight of the contained items. Another scenario is a much more spartan inventory, say only five items, for adventure-modes in which the player is very constrained in what he/she can carry. Would the current inventory system allow for a module to customize it to provide these features? If not, the ability to replace the inventory could be useful.

3. I completely agree.

Cervator:
So from some past discussion and what I can pull out of my head on a whim here are some answers and additional comments :-)

1. A soft dependency on this is hooking up something like the previously discussed Iota module series as present by default in a source workspace - potentially as simple dependencies on the PC facade so they'll retrieve like any other library jar and you can override with `gradlew fetchModuleInventory`. This is something I was working on around the end of GCI, but as is often the case with me I've been derailed from it again.

2. Alternative inventory implementation yep and further slimming down of the engine *repo*. A couple alternatives have been suggested, like Ultima Online style "loose bags" where you just drop in items by their icon or even taking it a step further where you have you be able to fit said item icons without overlapping any pixels (unlike UO)

3. It breaks all kinds of modules :-) Yet IMHO it'll be worth the pain while we still haven't locked down *module* level API (Inventory is now in Core, not engine). I believe @jellysnake has already put a substantial amount of thought/work into that.

As a general comment on inventory it has gotten kinda rough in places over time. Its extraction out of engine and assorted patching over time to keep it running well in multiplayer has gone part of the distance to refactoring it, almost like what we're doing to the rendering system (but fortunately far smaller)

Lots of little outstanding issues and annoyances are still present with inventory and this is probably one good (if a bit effort intensive) way to really overhaul it. I think we lack some clean design and good practices on the feature that this will help rectify.

I suspect it'll still take quite a while before we consider merging it, paired with some related support like more easily getting a "complete" source workspace. I haven't had a chance to look at the code here yet, but am optimistic based on some occasional discussion on IRC.

jellysnake:
1. Cervator essentially covered this. Having it fetch as a binary like other module dependencies would be the best solution to the issue of having it there. Perhaps by having Iota as a dependency for the PC facade, because this lets us easily solve similar situations (like Health) by adding those modules to Iota

2. There are a couple of benefits to extraction. It lets alternative inventory implementations, as well as extensions, actually exist. Along the same vein it makes working on or with the inventory much easier.

3. Any module that uses an inventory related function is going to be broken by sheer fact that they'll need to depend on Inventory as well as core now. However the changes needed are rather small and as such can actually be fixed quickly.

I also just want to add that the main reason why I started was because the state inventory was in, it was too hard to try and work with it. For instance, I wanted to have a potion effects screen display next to the inventory. Because the inventory apparently uses three different methods in a bunch of different places to open/close the inventory this was much harder than it should have been. Now you can simply check from nuiManager or trivially add an event.  

Similarly I was planning on adding a pipes module which was going to be horridly difficult to get it to work with block Inventories. So I decided to refactor it and Cervator suggested that I also extract it at the same time.

flo:
1. Downloading dependencies is not the issue. You can already download them ingame. What I see as issue is it becomes more and more effort to work on the game. E.g. when you want to have a look at the inventory you have to download it. So you add it as source, then you work on something else and at some point you get a strange compile error because you forgot to update the inventory module. So you basically have to regularly update all the basic modules you have downloaded in source form. Also as a new developer you can't just look right at the source it, but have to get modules. So I think it would be great to have core like modules being part of the main repository.

2. Can you explain "Along the same vein it makes working on or with the inventory much easier." further?

Previously adding a different inventory screen was as easy as just defining a UI screen that overrides (via overrides folder) the existing inventory UI screen. You won't have to write any opening/closing code so I can't follow that argument.

What difficulties did you have with your pipe module? What did it try to archive and where have there been problems? 

I would rather have an improved inventory that is compatible with your ideas then having 2 separate implementations that are incompatible to each other it basically will split up all modules in 2 sections. Those that work with your inventory and those that work with the other.

Cervator:
@flo you make absolutely valid points. It is a balancing act between "noisy repo" or "noisy logistics" more or less.

Waaaaay back before the Grand Convergence (the legacy days without even the entity system or multiplayer) we had a dozen or more modules in the main repo so PRs were all over the place with a ton of noise and occasional problems like merge history going wonky. That was part of the motivation for splitting into independent module repos in the first place.

Core was left behind with temporary intent and as a transition ground for when we extracted stuff out of the engine. Really we could leave Inventory in the engine repo for a while too. But eventually we're likely to get back to a dozen or more modules in the engine repo again - we have a couple new members already in the two sample gameplay modules. Health could also get split out.

For some time now we've informally been drifting toward extracting *everything* content-related into module repos with the expectation that around when that happens we'll have the extra utility (better source fetching and maintenance) and API docs to make it a pretty seamless process. But partly that direction has come from a lack of counter arguments.

I agree that without super convenient contributor utility splitting out basic functionality would cause added pain and not be worth it. I'm open minded to alternatives like keeping a few more modules in this repo, but since I've usually been the one driving the slow long term projects and am logistics-centered the current approach has admittedly been colored a fair bit by logistics over code

As for the functionality @jellysnake might be able to better speak to that. I think the "makes it easier" angle is more about having Inventory split off on its own just for whomever cares about it while the engine gets smaller and smaller so there'll be less code to worry about. Maybe better API separation somehow.

I imagine there'll be some basic inventory "offering" logic shared between our current classic inventory and a new school, say, Ultima Online "open bag" approach.

Cervator:
Oh, a small extra. I'm playing with a sketch of a potential new introduction to a variety of tutorials:

https://sketchboard.me/tAlG29BYVSTN#/

Only one of the "tracks" actually starts with an engine source work space at all. Naturally it is a common path to arrive at, but to start out a full workspace can be intimidating.

I'm wondering if we can develop a new approach for modding starting out of a game zip - no source code needed at all. That might lessen the concerns about how approachable a full engine workspace is.

Really, when you think about it how many contributors will ever even need to see for instance Inventory code if we have good enough tutorials, guides, and reference docs? :-)

Cervator:
One more addition (I've found a few of these over time, just hard to keep them all in mind!): release management is more difficult for embedded modules as the current setup has them all increment when we release the engine - even if, say, Core has no changes at all. Can't keep it at the same number in the same build as then the publish to Artifactory would fail due to non-unique release builds.

Chances are that could be done differently and not with a ton of difficulty, it is just one more small logistical item I've kept in mind due to my focus :-)

jellysnake:
Closing this as I have no time to work on it and hitting a couple of big walls. May re-open in future if I find a chuck of time for it

PR #2793, opened 2017-02-19
OpenVR motion controller gameplay through controller button API and e…

indianajohn:
### Contains

Support for gameplay using OpenVR-compatible motion-tracked controllers. The controller can be used to target in-game objects and trigger actions (i.e. mine blocks that you are touching with the controller).

A video is worth a thousand words: https://youtu.be/zkZ04NFy73U

### How to test

If you have a VR headset, start the program, load/create a game, and put the headset on. Put on a controller. Try to do the same thing in the video.

If you do not, try to turn play the game as you normally would. If you notice any problems with the targeting system, it could be a side effect of this commit. (However, I have tested non-VR gameplay, and everything seems to work as normal). Also pay attention to the animations / position of the tool.

A couple of things that I should mention.
* Before this commit, the target of actions and the target as seen by the renderer were being determined by two different instances of ray casting in the gaze direction. I changed it so that the target assigned to both is given by the targeting system instead of being done separately. This made it easier to hijack the targeting system for my own purposes.
  * When VR is being used, the target system is used to assign a target based on the OpenVR controller location.
  * When VR is not being used, the target system follows the gaze location, and the behavior is the same as it used to be (although what is happening behind the scenes is a little more efficient).
* I found some time ago that, when VR mode was active, I was only able to see the world rendered in the headset. The rendering to the computer monitor was not working, because the `CopyImageToHMD` node was binding a different FBO without telling the DAG about it. I made a temporary fix - to bind FBO 0 after `CopyImageToHMD` terminates. It's at least not much uglier than before.
* Controller support is done through the existing controller API. Each controller has a trigger, a grip button, a menu button, and a touchpad with joystick axes. On the left controller, the first three buttons are bound to buttons 0 - 2, and the touchpad is bound to movement. On the second controller, the first three buttons are bound to 3-5 and the four directions of the touchpad are bound to buttons 6-9. I haven't given much thought to what the best layout for gameplay is, but the back-end is in place.
* There are a couple of random VR fixes in there too. Before, the eye height was not being accounted for and the floor plane didn't match the floor plane in the room. That is fixed. There was also a minor problem with the camera that regarding the inverse projection matrix that could have been causing some problems that is now fixed.


GooeyHub:

GooeyHub:

GooeyHub:

emanuele3d:
I will strive to review this sometimes mid-week.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Very cool looking video! Nicely done :-)

Wondering if @tdgunes got a motion controller via the OSVR setup and might have time to try it out. Also poking @Josharias if available on feedback on gaze-related and item holding stuff.

Am curious: is the sort of free-hand wielding of the pickaxe purely a construct on the owning player's screen, or could the pickaxe be an independent entity that could be replicated so a second player could see it get swung around wildly? I think Josharias was working toward that with a more basic "Show what each player is holding" approach.

tdgunes:
After watching the video, that made me think about picking how we can make picking blocks much cooler. 😄 How about an actual swing motion with the pickaxe? If you hit it fast enough you can pick the block in an instant, if not you need to swing it again.

Well, I forgot to write here that I finally setup OSVR and tried it with OpenVR. As far as I remember, with OpenVR (SteamVR) it was really blurry and also since I don't have any controllers, it's hard to make it move. 

What I failed to do is to configure the headset for only OSVR native libraries. It's really time consuming and guides are not there yet. I managed to use it with OpenVR by an unofficial guide from an reddit thread. So I think moving forward with OpenVR makes sense, since I guess you can support almost any headset.

Cervator:
Heya @tdgunes good to hear from you! Glad you had a chance to test the headset, even if there is still work to do. Maybe we should poke our OSVR rep with some feedback and ask what might be a suitable motion controller from them?

I also had a thought hit me about the Leap Motion vs these kinds of motion controllers that you hold. Probably the way you could make the Leap work in this case is register the *user's hands* as motion controllers, with the Leap sort of serving as a proxy to translate hand movement and gestures into simulated input for a standard held motion controller. That could be an interesting project.

indianajohn:
Re: passing tool information over the network. Yeah, I think it's possible. I dug into how entity state is sent over the network and got a bit confused, and temporarily gave up. As a network game should do, Terasology is doing motion prediction using kinematics to smooth over the effects of network latency, and this makes the problem a bit more complex than just sending positional data whenever it's available. And besides, that data is coming in at a very high rate and would likely increase the network overhead for a player. I didn't have an immediate answer for making that possible, so I glossed over it for now. Any suggestions from those more familiar would be welcome.

Re: swinging. Yes, it's absolutely possible to keep track of the velocity of the controller and trigger mining based on the velocity of the controller. At that point it becomes a balance issue. If a VR user is co-existing with a boring keyboard-and-mouse user, then he/she might have an advantage at being able to mine faster.

Re: blurriness. Try maximizing your window before making a new game / loading your game. I know this isn't ideal, but at the moment, the image per-eye is being rendered at the same resolution as the on-screen window size. Also try to disable pretty much all image-processing effects in your settings - blur, film grain, etc...  I know this is less than ideal at the moment, but I am going to eventually make a VR preset that should handle this in the settings dialogue.

Re: leap motion, it could work, but I must stress that, absent a system that has a shared frame for the controllers and headset, it will feel frustrating to sense that the controller isn't moving right relative to your perspective, so it is absolutely vital to get that right.

Re: using your OSVR for control: At the moment, since no tracking events come through, gaze-based targeting should continue to work.  You can then use either a controller or just your keyboard to do all of the actions you would normally do.  If you press the forward button, I am guessing that you will move in the direction you are facing. If that's not the case, let me know. (this is by design - motion sickness is much reduced if there is no strafing motion, but especially no camera rotation).  I guess the way things are currently set up, it's impossible to use an XBox controller or whatever. It's pretty easy to fix this, and a decent cherry-pickable issue for someone else to tackle, I think.

emanuele3d:
On the issue of swinging the axe to break the block: I also felt it is needed, as just placing the axe to see the block break is highly unrealistic - would work for a pneumatic hammer instead. 

That been said I would refrain from speed-based breaking. First because there will be people with all sorts of motor problems that won't be able to swing the controller fast enough. Second because even healthy people will get into RSI problems as the controller has no significant weight and will allow for unnaturally abrupt movements straining the articulations. Third because swung controllers will fly. At objects, at people, no matter the wristbands - think wiimotes stuck in plasma screens.

That doesn't mean we can't take advantage of the novelty of these controllers. For example rather than simply measuring speed we could require a reasonable minimum swing amplitude AND a specific hit frequency for a given material. Fragile things would be broken with small, relatively frequent movements while hard rock might require large, infrequent but regular hits - think chopping a tree with an axe for example.

I do agree on the need to balance things between keyboard-based and controller-based players. _However_ if we found a way to reliably require controller-based players to spend a fair amount of calories in their chopping action, I'd be happy to reward them with a higher effectiveness/efficiency in their actions. This could provide an incentive for players to move more.

emanuele3d:
By the way, @indianajohn, you mention the problem of shared frame between the Leap Motion and the VR setup. Why do you think it will be hard to put the LM data into the same frame/world as the VR setup?

indianajohn:
In order to put something in the frame of OpenVR, it's necessary to know the transformation matrix between the leap motion device frame and the SteamVR coordinate system origin.

Without having looked at the SDK, I can only judge by pictures and videos I've seen. But the Leap Motion hand sensor device is designed to either be put on a desk or on the front of a headset. This makes every setup unique, and the transformation between the two reference frames different each time the sensors (lighthouses and leap motion) are moved with respect to one another. This can be gleaned either through:

* Some SDK magic (maybe the leap motion SDK has some existing procedures for mounting/calibrating the thing).
* A calibration process where 4+ positional correspondences between the two frames are gathered, and the 12-variable system of 12+ linear equations simultaneously, retrieving `r_ij`, `t_i` of the transformation matrix.

The second is annoying to write, and both necessitate that the user calibrates whenever the leap motion or the light houses are moved. It would be cool if the SDK offered a calibration process, but even then, the user would be forced to calibrate a lot. This is the reason why SteamVR made the lighthouse system at great expense and complication - your average user doesn't want to mess with it.

Anyway, it looks like it starts out as more of a research project. You're not the first to go down this road.

https://github.com/cbuchner1/driver_leap

forked from:

https://github.com/ValveSoftware/driver_hydra

If either of these ends up working well, then instead of working support into Terasology, the user could just install a driver for the Razer Hydra or the Leap Motion, and OpenVR would just detect controllers for a user's hands. Presumably these drivers have some method of calibration built in? Definitely worth looking at for whoever decides to take responsibility for the task.

indianajohn:
Good video on the subject: https://www.youtube.com/watch?v=Qk2QlvKLWOA

Cervator:
Lots of neat information, thanks! Getting back to testing this I've done some non-VR validation and it seems good so far :-) Haven't looked at the code in detail though, leaving that for @emanuele3d at some point convenient to him.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
```diff
@@ -204,6 +206,22 @@ public Vector3f getViewingDirection() {
         return viewingDirection;
     }
 
+    /**
+     * Get the gaze direction.
+     * @return the gaze direction, a quaternion.
+     */
+    public Quat4f getGazeDirection() {
```
emanuele3d: Hmmm.... I find that the term "gazing" implies an agent, i.e. a human, an animal or an autonomous robot.

Cameras however, even though right now are mostly related to the player, could be attached to inanimate
objects, such as in-game security cameras. 

In this context I wouldn't use the term "gaze" in here. If you really need a gaze direction I'd retrieve a vector from whatever dictates the camera orientation and in your code I'd then name it gazeDirection.

But even before that, why isn't getViewingDirection() enough?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToHMDNode.java
```diff
@@ -117,5 +119,14 @@ private void renderFinalStereoImage(RenderingStage renderingStage) {
...
         }
+
+        // Bind the default FBO. The DAG does not recognize that this node has
+        // bound a different FBO, so as far as it is concerned, FBO 0 is still
+        // bound. As a result, without the below line, the image is only copied
+        // to the HMD - not to the screen as we would like. To get around this,
+        // we bind the default FBO here at the end.  This is a bit brittle
+        // because it assumes that FBO 0 is bound before this node is run.
+        // TODO: break this node into two different nodes that use addDesiredStateChange(BindFBO...))
+        new BindFBOTask(0,DEFAULT_FRAME_BUFFER_URN).execute();
```
emanuele3d: Eeeeeew. :stuck_out_tongue: 

Yes, this is pretty ugly. :smile: 

There is no advantage in using the BindFBOTask here - which, by the way, might become a private class as is not intended for direct instantiation. I therefore recommend for the time being simply using

`glBindFramebufferEXT(GL_FRAMEBUFFER_EXT, 0)`

and then, as you suggest, breaking this node into two nodes, each binding the appropriate FBO.

By the way, are you sure you need to clear the left/rightEye buffers? And if yes, do they even have a depth attachment?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -176,6 +176,13 @@ public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
             // the log.
             if (vrProvider.init()) {
                 playerCamera = new OpenVRStereoCamera(vrProvider);
+                /*
+                * The origin of OpenVR's coordinate system lies on the ground of the user. We have to move this origin
+                * such that the ground plane of the rendering system and the ground plane of the room the VR user is
+                * in match.
+                 */
+                //
```
emanuele3d: Former single-line comment?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -176,6 +176,13 @@ public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
...
+                vrProvider.getState().setGroundPlaneYOffset(-0.80f - context.get(Config.class).getPlayer().getEyeHeight());
```
emanuele3d: - no magic numbers: what's -0.80f?
- you might want to store the result of getPlayer() somewhere, so that you don't have to get it through the context every time this method is called.

engine/src/main/java/org/terasology/input/cameraTarget/TargetSystem.java
```diff
@@ -44,6 +44,15 @@ public TargetSystem(BlockEntityRegistry blockRegistry, Physics physics) {
         this.physics = physics;
     }
 
+    /**
+     * Gets the position of the block that is currently targeted. If there is currently no target, this will return
+     * a null reference (isTargetAvailable() would have returned false in this case).
+     * @return the target block position in world coordinates, a vector of 3 integers.
+     */
+    public Vector3i getTargetBlockPosition() {
```
indianajohn: I'll refer you to a comment I wrote on line 130 in `CharacterSystem.java`:

```
// It would seem to make more sense to just call getTargetBlock() and get the entity here, but that causes
 // a NPE up in the EntityRef.
 // Vector3i positionVector = targetSystem.getTargetBlockPosition();
 Vector3i positionVector = targetSystem.getTargetBlockPosition();
```

I'll expand that explanation here to note that replacing the  contents of the `PlayerTargetSystem.getBlockPosition()` method with:

```
Vector3i positionVector = targetSystem.getTargetBlock().
LocationComponent location = target.getComponent(LocationComponent.class);
return location.getWorldPosition();
```

causes the same NPE that is referenced above. Specifically, the NPE happens when `PlayerTargetSystem.getBlockPosition()` is called from `CharacterSystem as it was above; curiously, any calls to this method from `updateTarget()` succeed.

I traced the issue up the stack and was unable to resolve it at the time. I may report it in an issue, since it seems like this should be a valid use case of the `getTarget()` method. (I'm too lazy to go turn on my headset to get a stack trace right now, or I'd include it here :/)

engine/src/main/java/org/terasology/input/cameraTarget/PlayerTargetSystem.java
```diff
@@ -56,6 +58,14 @@ public EntityRef getTarget() {
         return targetSystem.getTarget();
     }
 
+    /**
+     * Get the position of the block that is currently targeted.
+     * @return the position of the block, a vector.
+     */
+    public Vector3i getTargetBlockPosition() {
+        return targetSystem.getTargetBlockPosition();
+    }
```
indianajohn: This is related to the comment I made directly below this - I'm having to bend over backwards to avoid calling `getTarget()` from `CharacterSystem` to get around a presumed bug in the entity system.

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
```diff
@@ -0,0 +1,168 @@
...
+        ids.add("OpenVR");
+        return ids;
+    }
+
+    /**
+     * Get all queued actions registered since this method was last called.
+     * @return a queue of actions.
+     */
+    @Override
+    public Queue<ControllerAction> getInputQueue() {
```
indianajohn: In this case I'm following the method name from the interface I'm implementing, `ControllerDevice`. Renaming the superclass and the other implementation of that interface (`JInputControllerDevice`) would involve editing code which I didn't originally write, which I avoid whenever I can.

I have added a note to the effect in the class Javadocs, to be pushed shortly.

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
```diff
@@ -0,0 +1,168 @@
...
+        Queue<ControllerAction> result = new ArrayDeque<>();
+        result.addAll(queuedActions);
```
indianajohn: Yup, done in the next commit.

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
```diff
@@ -0,0 +1,168 @@
...
+        queuedActions.clear();
+        return result;
+    }
+
+    private void handleController0(VRControllerState_t stateBefore, VRControllerState_t stateAfter) {
```
indianajohn: I refactored this more or less how you said. Having written it, I was a bit blinded to the readability issues there. Good idea.

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRInput.java
```diff
@@ -0,0 +1,98 @@
...
+import org.terasology.input.lwjgl.LwjglMouseDevice;
+import org.terasology.rendering.openvrprovider.OpenVRProvider;
+
+public class OpenVRInput implements EngineSubsystem {
+
+    private Context context;
+
+    private OpenVRProvider vrProvider;
+
+    private OpenVRControllers controllerDevice = new OpenVRControllers();
```
indianajohn: Yep, no harm in it.

engine/src/main/java/org/terasology/input/cameraTarget/PlayerTargetSystem.java
```diff
@@ -56,6 +58,14 @@ public EntityRef getTarget() {
         return targetSystem.getTarget();
     }
 
+    /**
+     * Get the position of the block that is currently targeted.
+     * @return the position of the block, a vector.
```
indianajohn: Done.

engine/src/main/java/org/terasology/input/cameraTarget/TargetSystem.java
```diff
@@ -44,6 +44,15 @@ public TargetSystem(BlockEntityRegistry blockRegistry, Physics physics) {
...
+        return targetBlockPos;
+    }
+
```
indianajohn: Are you referring to this?

```
LocationComponent location = target.getComponent(LocationComponent.class);
        if (location != null && targetBlockPos != null) {
            location.setLocalPosition(targetBlockPos.toVector3f());
        }
```
As I understand this code, the `LocationComponent` is a reference to the one stored within the entity. So by calling `setLocalPosition()` on that block, it mutates the location component on the actual entity. I'm not sure what happens downstream of / surrounding this code. Once again, the Law of Least Intervention is a decent one to follow on large code bases being worked on by many different people.

engine/src/main/java/org/terasology/logic/characters/CharacterSystem.java
```diff
@@ -118,16 +127,14 @@ public void onAttackRequest(AttackRequest event, EntityRef character, CharacterC
...
-            Vector3f direction = gazeLocation.getWorldDirection();
-            Vector3f originPos = gazeLocation.getWorldPosition();
-
-            HitResult result = physics.rayTrace(originPos, direction, characterComponent.interactionRange, Sets.newHashSet(character), DEFAULTPHYSICSFILTER);
-
-            if (result.isHit()) {
-                result.getEntity().send(new AttackEvent(character, event.getItem()));
+            // It would seem to make more sense to just call getTargetBlock() and get the entity here, but that causes
+            // a NPE up in the EntityRef.
+            Vector3i positionVector = targetSystem.getTargetBlockPosition();
```
indianajohn: Fair enough, changed.

engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
```diff
@@ -51,6 +52,15 @@
     );
 
     /**
+     * Sometimes, a tracking system (i.e. for room-scale VR) provides a pose for the player hand position. In this
+     * case, special handling is necessary, and this accessor gives a way to check.
+     * @return true if this class is receiving tracking updates, false if not.
+     */
+    public boolean isTracked() {
+        return trackingDataReceived;
+    }
```
indianajohn: I've tried to skirt the line between specificity and generality here. If I leave my comments too vague (is tracked? How?), then reading the javadocs doesn't make any sense.

The way I see it, a "tool mount point" is almost always a hand. so I used the two interchangeably in this case. However, I've adjusted the Javadoc here to only reference "tool mount point", because I can see how referring to a hand here could be confusing.

The suggestion I made is that I would make a class inside of the new VR subsystem that implements `ControllerListener`, holds reference to the mount point, and edits the quaternion on the mount point. We could then remove the implementation of `ControllerListener` here. That seems to jive more with the "systems modify components" architecture.This is still on my road map.

I still see this method as being necessary after I perform the above change. Regardless of what is doing the updating of the location info on the component, the special handling needed by other components (e.g. `FirstPersonClientSystem`, `PlayerTargetSystem`) hinges around the quaternion / position being updated dynamically rather than needing special handling as dumb objects. More specifically, without tracking information, animations are needed when actions are triggered, and there is no information that contributes to targeting. My language here seems general enough to reflect that, and also includes examples for clarification.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
indianajohn: Honestly, I just needed to have this process:
LocalPlayerSystem: Hey camera, please look at X.
VR camera: Can't do that buddy, my camera only moves where my player moves his head.
LocalPlayerSystem: Alright camera, where are we looking?
VR camera: Here's the player's head pose.

vs.

LocalPlayerSystem: Look here, camera.
Normal Camera: Okay.
LocalPlayerSystem: Where are we looking?
Normal Camera: (gives the direction that was given earlier).

I agree that the viewing direction is probably redundant here, and in fact, there are no references elsewhere in the code, so it's probably fine to merge the two. When I initially decided not to use `getViewingDirection()` to store that info, I was just worried that I would break something else that depended on it working as it did earlier.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToHMDNode.java
indianajohn: 1. Yeah, the direct OpenGL call here is reasonable. I just thought that directly calling OpenGL functions inside nodes was discouraged. Fixed.

2. Yep, we definitely need two frame buffers. the call order for VR rendering in OpenVR is:
1. pass FBO ids into OpenVR to be used for left/right eyes.
2. In the past I have noticed weird behavior without a clear() statement here. I'm sorry, but I can't really elaborate - not enough pattern to the weirdness to really articulate it, just crashes and strange frame rate issues.
3. I'll have to get back to you on the depth attachment question. I think the depth attachment is optional but useful. OpenVR is beginning to implement Asynchronous warping, in which if a frame is not provided by the application within a certain time frame, it can use the contents of the color / depth buffer along with the head pose to interpolate another frame. This can increase the effective frame rate, which is a very good thing for VR games, not only because they require a lot of rendering power, but also because low frame rate contributes to motion sickness.

But that knowledge is from reading articles - I have yet to actually poke OpenVR and see how to make that work. I know that some games are currently using it.

indianajohn: I updated the comment above due to a editing mistake. What you got through email might be different, sorry. :)

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
indianajohn: Fixed.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
indianajohn: For the moment I've just removed that factor. I need to fire the rig up and see the relationship between the setting for eye height and where the ground plane ends up. All I remember is that there was an issue where the expected `setGroundPlaneYOffset(-eyeHeight);` did not work properly and I had to add a random offset to fix it.

(I'll update this in a bit).

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
emanuele3d: Ok, fair enough. Indeed if there are constraints on the naming of something because it has a relationship with something else let's make it clear in the javadocs or in very visible comments.

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
emanuele3d: We are all blinded by our own code in one measure or another. That's what additional pair of eyes are for. :smiley:

engine/src/main/java/org/terasology/input/cameraTarget/TargetSystem.java
emanuele3d: Tagging @Josharias: he might be more knowledgeable than I am about this part of the codebase.

engine/src/main/java/org/terasology/input/cameraTarget/TargetSystem.java
emanuele3d: Oooohhh. I get it. Ok, never mind. Sorry about that. I guess it would be nice to rename the location variable to targetLocation so that it's a bit clearer.

engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
emanuele3d: I must admit that went over my head a bit. This is a part of the codebase I'm highly unfamiliar with. I'm just concerned with getting too much VR-related code peppering the codebase, to the point that it can't be easily extracted and then 1-2 years down the line somebody will have to do work similar to what you did to eradicate Oculus support, just to add support for some other VR platform.

Let's continue to keep this in mind and hopefully one day we'll get to a polish phase where we can look again at the big picture and find ways to smooth these issues out.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
emanuele3d: Hmmm... here I'm concerned the camera is getting a bit "too smart". 

I appreciate your VR camera vs Normal Camera clarification: creative and to the point. But I'm thinking: the camera shouldn't argue if it can or cannot do something. The camera should  be pretty dumb and do what it is told. Systems perhaps could fight each other to determine who has control. Or perhaps the LocalPlayerSystem should be more VR-aware than it is now.

I mean: remember that one of the goal we have about cameras is that they should become even dumber components. So, if you add this functionality, we'll then have to strip it away and put it back into a system. 

I'd therefore like to ask you if you can go the extra-mile and handle these issues at system level rather than camera level. Who controls the camera should not be relevant for the camera.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToHMDNode.java
emanuele3d: 1. you are right it's deprecated, but this is meant to be temporary and to misuse tasks is more deprecated. :smile: 

2. Hmm. Strange. Might be worth to pop a question in an OpenVR forum perhaps? My impression is that part of the code completely overwrites the content of those buffers, so there shouldn't be a need clear them. Meanwhile you might want to leave a `// TODO - review need for clearing` in there? 

3. I feel by the time we get to this part of the code we are squarely in 2D processing mode: the only input data you are using is from the FINAL fbo which only has a color attachment: depth data at this stage is long gone. So, I'm struggling to see how the depth buffer might be involved even for the purpose you mention.

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
```diff
@@ -51,90 +59,83 @@
...
-            queuedActions.add(new ControllerAction(InputType.CONTROLLER_AXIS.getInput(ControllerId.X_AXIS),
-                    "OpenVR", ButtonState.DOWN, -stateAfter.rAxis[0].x));
-            queuedActions.add(new ControllerAction(InputType.CONTROLLER_AXIS.getInput(ControllerId.Y_AXIS),
-                    "OpenVR", ButtonState.DOWN, stateAfter.rAxis[0].y));
-        } else if (OpenVRUtil.switchedUp(ControllerListener.BUTTON_TOUCHPAD, stateBefore.ulButtonPressed, stateAfter.ulButtonPressed)) {
-            queuedActions.add(new ControllerAction(InputType.CONTROLLER_AXIS.getInput(ControllerId.X_AXIS),
-                    "OpenVR", ButtonState.UP, 0));
-            queuedActions.add(new ControllerAction(InputType.CONTROLLER_AXIS.getInput(ControllerId.Y_AXIS),
-                    "OpenVR", ButtonState.UP, 0));
+    private boolean switchedUp(long buttonIndex) {
```
emanuele3d: Could this be `buttonUp`? When I read this I just keep wondering _what_ is switching up. And could buttonIndex be `buttonId`?

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
```diff
@@ -51,90 +59,83 @@
...
+        queuedActions.add(new ControllerAction(InputType.CONTROLLER_AXIS.getInput(controllerButton),
+                    "OpenVR", buttonState, axisValue));
+    }
+
+    private void addButtonAction(int controllerButton, ButtonState buttonState) {
+        queuedActions.add(new ControllerAction(InputType.CONTROLLER_BUTTON.getInput(controllerButton),
+                "OpenVR", buttonState, 1.0f));
+    }
+
+    private void handleController0() {
```
emanuele3d: Thank you for improving this method, now I can see things better.

Question: these two methods seems to be polling all the buttons to check which has changed state. So, there is no way to know which button was pressed without asking each of them?

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
```diff
@@ -51,90 +59,83 @@
...
+        if (switchedUp(ControllerListener.BUTTON_TRIGGER)) {
+            addButtonAction(ControllerId.ZERO, ButtonState.UP);
+        } else if (switchedDown(ControllerListener.BUTTON_TRIGGER)) {
```
emanuele3d: I'm looking at constants such as ControllerId.ZERO and ControllerListener.BUTTON_TRIGGER: are these again matching those in somebody's else code?

indianajohn:
Alright, I feel like I'm done making commits in response to your review. If you see further issues, please let me know. My last few commits were mainly bug fixes to address sloppy coding during my revisions. Sorry for the mail bomb.

One nontrivial change is my changing of the way that the camera orientation is handled. I should clarify a bit.

A "pose" as a SLAM / robotics student would understand it is how position/orientation is represented for an object in 3D space. Itg is naturally 6 degrees of freedom - x, y, z, roll, pitch, and yaw. People shy away from using RPY because of a phenomenon known as gimbal lock: singularities exist in certain configurations.

Another way to represent rotation/orientation is X,Y, Z, VECTOR_1, VECTOR_2, where VECTOR_1 and VECTOR_2 are orthonormal. Using VECTOR_1 and VECTOR_2, it's possible to get VECTOR_3, the third vector of an orthonormal basis. This is how the camera encoded its "pose" before I touched it.

I found that the ``up()` method was not used anywhere in the code, so I just took it out and switched this representation to an axis and an angle, which is convertible to a quaternion. The accessor `getOrientation()` makes sense because a quaternion represents a unique rotation. The `viewingDirection()` accessor still works, as it should, since it is used elsewhere in the code. `getOrientation()` forms a quaternion by using the `viewingDirection` and `viewingAngle` - the combination of which fully parameterizes the orientation of the camera.

I've tested this in VR mode and normal mode and everything seems to work as expected. Let me know if you notice anything different.

Alright, so I'll now request that this gets another look-over, and I'll address any further comments I get.

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
```diff
@@ -147,10 +148,14 @@ private void handleController1(VRControllerState_t stateBefore, VRControllerStat
      */
     @Override
     public void buttonStateChanged(VRControllerState_t stateBefore, VRControllerState_t stateAfter, int handIndex) {
+        cachedStateBefore = stateBefore;
+        cachedStateAfter = stateAfter;
+        queuedActions.add(new ControllerAction(InputType.CONTROLLER_BUTTON.getInput(ControllerId.X_AXIS),
```
emanuele3d: This seems odd here. What is this for?

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
```diff
@@ -207,23 +207,20 @@ public Vector3f getViewingDirection() {
...
      */
-    public void requestSetGazeDirection(Quat4f gazeDirectionIn) {
-        gazeDirection = gazeDirectionIn;
-    }
-
-    public Vector3f getUp() {
-        return up;
+    public void requestSetOrientation(Quat4f direction) {
+        viewingDirection = direction.getAxis();
+        viewingAngle = direction.getAngle();
```
emanuele3d: Naming is better now, but I'm still not convinced this method should be here, see my reply to your comment on the matter, in an outdated commit. 

If we do go down this road I'd prefer setDesiredOrientation or perhaps suggestOrientation: requestSetOrientation can be interpreted as "request the orientation that has been set".

GooeyHub:

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
indianajohn: I should clarify that the only data getting passed is a single `long` value. `isDown` and `isUp` both use `isPressed`, which is just a simple bit mask check. This allows for a long value to have the entire state of the controller. So `TRIGGER & buttonState == true` and `GRIP & buttonState == true` can simultaneously be satisfied.

The just is that even though it looks like there are a lot of function calls / work being done there the code will reduce to several bit mask checks, which is very cheap. And they all need to be checked every time.

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
indianajohn: Well, `ControllerId.ZERO` is part of Terasology's controller interface. `BUTTON_TRIGGER` comes from the following line:

```
long BUTTON_TRIGGER = (1L << JOpenVRLibrary.EVRButtonId.EVRButtonId_k_EButton_SteamVR_Trigger);
```

which I wrote. the programmer-esque word order comes from the original SDK, and I converted it over to our check style guidelines.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToHMDNode.java
indianajohn: 2. I added a TODO.

3. As I understand ASW, it uses the depth buffer to reproject results from the previous frame into the missed frame using the tracking data. By knowing the position of a particle at time `t_1` and the head pose at time `t_2`, it's possible to project that particle into the reference frame needed for time `t_2`. If there were no depth buffer provided, I'd guess that the SDK just wouldn't be able to do that.
Once again, just speculation here. In order to get ASW working I would have to sit down and read about it, as well as verify it were working through experimentation.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
indianajohn: So in this case, where the `LocalPlayerSystem` manages player movement, it is necessary to have some way of checking whether the orientation needs to be updated by events or whether it is being set by a tracking system. I have named this method `isTracked()`. To respond to your desire to have less work/logic happening in the camera (as it is moving toward being a `Component`), I have moved the logic relating to how the camera movement is handled into `LocalPlayerSystem`.

Moving forward I think there will be a `RoomscaleVRPlayerSystem` and `SeatedVRPlayerSystem` to take the burden of making this decision away from this VR-nonspecific system. In that case it won't be necessary to have the check on the camera, because we will already know that a VR camera is active.

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
indianajohn: I've now put in what is essentially a fudge factor here. The issue is that if I move `getEyeHeight()` units down from the camera height, I don't encounter the ground plane. To me this doesn't seem right.

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
indianajohn: In response I have only "oops"

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
indianajohn: So this is gone, and I have commented above on why and how I've refactored things.

indianajohn:
I've addressed your second round of comments fully now, I think.

GooeyHub:

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
indianajohn: How did I miss this? Done.

GooeyHub:

jzacharias-associa:
Pinging in!  I am still catching up on the conversation.  This is cool stuff, @indianajohn!  ETA for my comments is later tonight PST.

Josharias:
This is cool stuff, @indianajohn!  ETA for my comments is later this evening PST.

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -162,16 +162,22 @@ private void processInput(EntityRef entity, CharacterMovementComponent character
         Quat4f viewRot;
         switch (characterMovementComponent.mode) {
             case WALKING:
-                viewRot = new Quat4f(TeraMath.DEG_TO_RAD * lookYaw, 0, 0);
-                viewRot.rotate(relMove, relMove);
+                if (!playerCamera.isTracked()) {
```
emanuele3d: Hmmmm. Very unconvinced by this. When cameras will be components it won't be the camera component to see its data change as the player moves his head and walk around the room, but the orientation and position components in the same entity. What should we have then? A tracked flag in both position and orientation components? I would recommend not: if a tracking system is enabled it should take exclusive control of those components and simply write into them. In this context it would be perhaps more appropriate that the LocalPlayerSystem or the CharacterSystem store a tracked flag. That been said: when is the player's head really _not_ tracked? Isn't changing the orientation of a character with the mouse a form of (hand) tracking?

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -162,16 +162,22 @@ private void processInput(EntityRef entity, CharacterMovementComponent character
...
+                    viewRot = new Quat4f(TeraMath.DEG_TO_RAD * lookYaw, 0, 0);
```
emanuele3d: I'd write: `viewRotation =  new Quat4f(DEG_TO_RAD * lookYaw, 0, 0);`

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
```diff
@@ -38,6 +38,7 @@
     // In the head frame
     private Matrix4f[] eyePoses = new Matrix4f[2];
     private Matrix4f[] projectionMatrices = new Matrix4f[2];
+    private float groundPlaneYOffset = 0.0f;
```
Josharias: Is this the same as the height of the character's eyes?  If it is, probably the more resilient way to get this value would be from the character that the player is controlling.  However,  I suppose that making this different from the character is good so that VR nausea is avoided?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -122,6 +122,12 @@
     private final ShaderManager shaderManager;
     private final Camera playerCamera;
 
+    /*
+    * presumably, the eye height should be context.get(Config.class).getPlayer().getEyeHeight() above the ground plane.
+    * It's not, so for now, we use this factor to adjust for the disparity.
+     */
+    private final float GROUND_PLANE_HEIGHT_DISPARITY = -0.7f;
```
Josharias: Its odd,  this value is the negative value of PlayerConfig.DEFAULT_PLAYER_EYE_HEIGHT.   Are we doing too much math for these eye height values?

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
Josharias: Do I understand correctly that this is here so that it does not give extra information to the camera so that the VR stuff can hook into a lower system and provide this information?  If so (which I give myself only a 50% chance of being on the right track), could the playerCamera just ignore this data instead of making 2 different logic paths through these systems?

engine/src/main/java/org/terasology/engine/subsystem/openvr/OpenVRControllers.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    private void addButtonAction(int controllerButton, ButtonState buttonState) {
+        queuedActions.add(new ControllerAction(InputType.CONTROLLER_BUTTON.getInput(controllerButton),
+                "OpenVR", buttonState, 1.0f));
+    }
+
+    private void handleController0() {
+        if (buttonUp(ControllerListener.BUTTON_TRIGGER)) {
+            addButtonAction(ControllerId.ZERO, ButtonState.UP);
```
indianajohn: 1. Iterating through xor of bit mask:
By doing what you suggest I think we could knock down a few lines of code, but I personally think it would end up being a little less readable. That sort of if statement block, while a bit ugly, is an idiom I've seen many times. Since the efficiency difference is minimal, I personally think that sacrificing a little bit of efficiency for readability would be better.

2. ControllerId.ZERO->ButtonId.ZERO: this goes once again into other folks' code. If we're fine with me editing those defines throughout the whole codebase, I'm on board for refactoring it.

3. BUTTON_TRIGGER->TRIGGER_BUTTON: that I can do. It'll be in my next commit.

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
indianajohn: @Josharias That was the case before my recent change. The idea was that the `trySetOrientation` (or something) method was specific per-camera and the VR camera ignored input to that method, where a normal camera did not.

Re: keyboard/mouse controls as a form of "head tracking" - to some extent, yes, it makes sense. That is why it's more appropriate to have a separate `LocalPlayerSystem` for VR. I would honestly prefer to not do that in this pull request.

Anyway, I leave it up to the two of you to decide among these three options:
1. Camera class contains a `setOrientation`method to handle the logic of optionally using or disregarding the pose received from `LocalPlayerSystem`.
2. The current state.
3. Scrap this PR until I do more work on a new player system for VR.

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
indianajohn: I changed this.

engine/src/main/java/org/terasology/rendering/openvrprovider/OpenVRState.java
indianajohn: I'm assuming this is outdated by the below comment.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
indianajohn: What I'm finding is this. Presumably using `-context.get(Config.class).getPlayer().getEyeHeight()` should result in the ground plane matching between the actual room and the world, right? But I'm finding that's not the case. If I just use that value, then a controller that is set on the ground hovers 0.7 meters off of the ground.

I'm stumped as to why this could be, but there's something fishy going on if you think about it (or my understanding is flawed). 1 world unit is 1 block high, right? Presumably 1 meter? So 0.7 world units would be less than the height of a single block. But the default eye height in a new game seems to be about 1.5 blocks high, which is consistent with my observations. Anyway, 1.5 meters makes more sense in terms of real-world units too - 1.5 meters is a more realistic height.

engine/src/main/java/org/terasology/input/cameraTarget/PlayerTargetSystem.java
```diff
@@ -66,6 +76,11 @@ public void update(float delta) {
             if (charComp != null) {
                 Vector3f dir = player.getViewDirection();
                 float maxDist = charComp.interactionRange;
+                FirstPersonHeldItemMountPointComponent heldItemMountPoint = player.getCameraEntity().getComponent(FirstPersonHeldItemMountPointComponent.class);
```
indianajohn: In my reading of the code, a 'PlayerTargetSystem` class does not seem to be associated with the eye ball. `PlayerTargetSystem` disambiguating between using gaze-based targeting and hand-based targeting is something I would expect.

I feel like the next logical step here is to create `GazeBasedPlayerTargetSystem` and `TrackedControllerPlayerTargetSystem` or something. As above, we can push this through for now and I can add that to my bigger "vr subsystem" PR, or we can nuke this and just wait for that one.

Anyway, everyone's reading of a piece of code can be different, so I'm up for revising it in a way that makes more sense. I just can't figure out a better place to put this. Suggestions welcome. :)

GooeyHub:

indianajohn:
I addressed what I felt was within the scope of this PR, and there are two issues remaining which I'll need further guidance on.

1. `isTracked()` vs. camera subclasses ignoring `setRotation()` as a method for having different behavior between VR and non-VR cameras.

2. Is there a need to extricate `FirstPersonMountPointComponent` references from `PlayerTargertSystem`? If so, how?

Downstream, I have plans to:

1. Make it so `FirstPersonMountPointComponent` is directly edited by a subsystem to further isolate VR-specific code.

2. Create `FirstPersonVRPlayerSystem` for similar reasons.

Finally,

* If you guys feel like we're unable to resolve some of the conflicts above, we can just drop this PR and pick up the subject in a later PR.

Cervator:
> If you guys feel like we're unable to resolve some of the conflicts above, we can just drop this PR and pick up the subject in a later PR.

It sounds like it is getting pretty close, and there'll be more work later in future PRs. I'd say hang in there :-)

indianajohn:
I think it looks fine to just check for VR support, and also avoids dirtying up the camera class with a new accessor. So for now I've done that, according to your suggestion @emanuele3d. As this proceeds into a separate system we can continue to move away from references to VR in non-VR related code.

I've also gotten this PR merge-ready again after upstream changes.

At this point I'm going to ask for a quick check-in on the reviewers on whether there are any changes that are needed before merging this PR.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -162,7 +162,7 @@ private void processInput(EntityRef entity, CharacterMovementComponent character
         Quat4f viewRotation;
         switch (characterMovementComponent.mode) {
             case WALKING:
-                if (!playerCamera.isTracked()) {
+                if (!config.getRendering().isVrSupport()) {
```
emanuele3d: Any way to get the renderingConfig only once and assign it to a field? I noticed this class does not have a constructor. This way we don't have to fetch the rendering config every time and we can write:

`if (!renderingConfig.isVrSupport()) {`

emanuele3d: Other than this I think we are good to go. Please take care of this so that I can greenlight everything. At this stage I don't think we should wait anymore for further reviews from Josharias, obviously he's very busy right now.

Cervator:
Looks like the final single change was resolved - merge time? :-)

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
indianajohn: It's possible, as you can see in this commit (which I've left out of this branch because I don't know if it's the best thing to do):

https://github.com/indianajohn/Terasology/commit/6bca73285992fa5a91849a9ce89245a4282c2a9b

Anyway, if we think we should add that I can push it to this branch.

emanuele3d: @indianajohn: so you are saying that the only way to get the renderingConfig injected in the field is to add it to the context in RegisterInputSystem? Otherwise it stays null? 

Bummer. I really don't like the coupling between rendering and input system. Conceptually they should really be separate. Let's leave things as they are then, so that there is less stuff to remove if one day we get to revisit this.

emanuele3d:

Cervator:
Cool, thanks @emanuele3d !

Tested it out and noticed two things

1: Before the PR I was able to run the game with a junky old generic controller and got the following log entry on startup, which revealed a controller section under the Input Settings named after it (and I could use the controller)

`01:52:42.624 [main] INFO  o.t.i.lwjgl.JInputControllerDevice - Registered controller: Afterglow controller for PS3`

After the change I do get the same log entry, but I get an OpenVR section instead of the explicitly named controller. It looks about the same (mostly just gives you options about axis inversion and dead zones) so that's probably fine. However, none of the controller buttons seems to work - not in-game despite the new controller mappings enabled by default, nor can I successfully bind any single key from the controller (seems like none of them are registering at all)

I'm not sure if that's partly just because my controller is junky? Or is the OpenVR support too geared toward OpenVR compatible controllers of some sort?

2: On startup until I look in any direction I get log spam probably from something be initialized in a way that causes trouble until looking somewhere changes a value:

```
01:57:02.815 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
java.lang.ArithmeticException: input is infinite or NaN
	at com.google.common.math.DoubleMath.roundIntermediate(DoubleMath.java:59)
	at com.google.common.math.DoubleMath.roundToInt(DoubleMath.java:132)
	at org.terasology.math.geom.Vector3i.<init>(Vector3i.java:87)
	at org.terasology.logic.characters.KinematicCharacterMover.step(KinematicCharacterMover.java:108)
	at org.terasology.logic.characters.ServerCharacterPredictionSystem.stepState(ServerCharacterPredictionSystem.java:168)
	at org.terasology.logic.characters.ServerCharacterPredictionSystem.onPlayerInput(ServerCharacterPredictionSystem.java:126)
	at org.terasology.logic.characters.ServerCharacterPredictionSystemMethodAccess.invoke(Unknown Source)
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:506)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.sendStandardEvent(EventSystemImpl.java:269)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:260)
	at org.terasology.entitySystem.entity.internal.BaseEntityRef.send(BaseEntityRef.java:145)
	at org.terasology.logic.players.LocalPlayerSystem.processInput(LocalPlayerSystem.java:187)
	at org.terasology.logic.players.LocalPlayerSystem.update(LocalPlayerSystem.java:152)
	at org.terasology.engine.modes.StateIngame.update(StateIngame.java:160)
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:413)
	at org.terasology.engine.Terasology.main(Terasology.java:153)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
```

Let me know if you can spot anything @indianajohn - would be so very thrilled to get this merged, and honestly I'm not worried about generic controller support working right this moment, I can't imagine it being used much, so if it does work for your motion controllers and we can look at the generic ones later I'm fine with that. We just did a stable release so it'll be a while till the next one. Even though these PRs do tend to take a while - but I really appreciate you keeping at it! :-)

Cervator:
Coincidental related update: backed a Kickstarter for some motion control stuff that then ended up with an update that they're adding OSVR support :-)

https://www.kickstarter.com/projects/243372678/nolo-affordable-motion-tracking-for-mobile-and-ste/posts/1844403

indianajohn:
It's a goal of mine to un-break regular controller support by the end of the weekend. I have a few ideas what could be causing it, and it should be fairly easy to fix. I have a controller to test with at home.

indianajohn:
It's a goal of mine to un-break regular controller support by the end of the weekend. I have a few ideas what could be causing it, and it should be fairly easy to fix. I have a controller to test with at home.

GooeyHub:

GooeyHub:

GooeyHub:

indianajohn:
Okay, 1 from above is fixed; it's now possible to use normal controller support again. There is a caveat.

If you start the game with `vrSupport` enabled, the game will assume you want to use an OpenVR motion controller, and therefore won't use/recognize your regular controller. It is possible to make it use both/either with `vrSupport` enabled, and probably fairly easily, but I haven't tested for this scenario, so we'll leave it inactive for the moment and consider vrSupport targeting motion controllers. Once we've specifically tested for the use case of standard controller + VR, we'll need to add it in a way that makes sense in terms of flow - i.e. as a config option, always on whenever the controller is, or both.

2 from above is also fixed as well, although I'm not super happy with how I had to do it. `Quat4f.rotate()` is returning `NaN` if it is near the value of a unit Quaternion. This should pretty much *never* happen for a quaternion class; the advantage of using quaternions for rotation is that there are no singularities.  Anyway, fixing that particular class is likely outside of the scope of this PR, so I've just filtered the 'NaN's out.

I've also merged develop in, so this branch is mergeable (hint hint :P ).

Let me know if there's anything else I can do to make this happen.

Cervator:
That all sounds fair :-)

GSOC app prep has been eating us alive but the deadline is in about 10-11 hours so it should calm down. Will merge as soon as I get a chance to retest, if nobody else beats me to it.

Cervator:
Did some final validation both with and without a boring old controller, can report both appeared fine!

Merging at long last, thank you for your patience @indianajohn and thorough review @emanuele3d (and @Josharias!) 🎉

emanuele3d:
Congrats @indianajohn!!

PR #2794, opened 2017-02-19
Renderer work on camera-related StateChanges

emanuele3d:
Adds and takes advantage of a number of StateChanges and associated tasks:
- LookThrough
- LookTroughNormalized
- SetCameraToReflected

### How to test
- run Terasology
- create a new game or load an existing one
- verify nothing has changed in the graphic output

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCameraToReflected.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCameraToReflected.java
@@ -19,10 +19,10 @@ import java.util.Objects;
 import org.terasology.rendering.cameras.Camera;
 import org.terasology.rendering.dag.RenderPipelineTask;
 import org.terasology.rendering.dag.StateChange;
-import org.terasology.rendering.dag.tasks.SetCameraReflectedMode;
+import org.terasology.rendering.dag.tasks.SetCameraReflectedModeTask;
 
 /**
- * TODO
+ * Instances of this state change generate the tasks that set or reset the reflected flag of a given camera.
  */
 public class SetCameraToReflected implements StateChange {
 
@@ -33,6 +33,11 @@ public class SetCameraToReflected implements StateChange {
 
     private RenderPipelineTask task;
 
+    /**
+     * Constructs an instance of this class initialized with a given Camera instance.
+     *
+     * @param camera An instance implementing the Camera interface.
+     */
     public SetCameraToReflected(Camera camera) {
         this(camera, true);
         defaultInstance = new SetCameraToReflected(camera, false);
@@ -43,10 +48,13 @@ public class SetCameraToReflected implements StateChange {
         this.reflected = reflected;
     }
 
+    /**
+     * @return a RenderPipelineTask configured to set or reset the reflection flag of the camera given on construction.
+     */
     @Override
     public RenderPipelineTask generateTask() {
         if (task == null) {
-            task = new SetCameraReflectedMode(camera, reflected);
+            task = new SetCameraReflectedModeTask(camera, reflected);
         }
         return task;
     }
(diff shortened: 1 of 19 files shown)


engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCameraToReflected.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Objects;
+import org.terasology.rendering.cameras.Camera;
+import org.terasology.rendering.dag.RenderPipelineTask;
+import org.terasology.rendering.dag.StateChange;
+import org.terasology.rendering.dag.tasks.SetCameraReflectedModeTask;
+
+/**
+ * Instances of this state change generate the tasks that set or reset the reflected flag of a given camera.
+ */
+public class SetCameraToReflected implements StateChange {
```
emanuele3d: I wasn't particularly inspired in the choosing of this class name. Alternatives?

GooeyHub:

emanuele3d:
Thank you @indianajohn for the review and also the high-level description of this PR. I should have been more verbose in my initial description: what's wrong with me? :open_mouth: In any case I confirm your understanding of it.

Regarding documenting the available StateChanges, would the html javadocs generated out of the classes in the StateChanges package be enough? But yes, of course we'd have to write some tutorial on how to use the whole thing.

indianajohn:
Nope, your description was fine. Part of the point of these things is to make sure things are self-documenting, and my summary was just so you could verify that my reading of everything was correct. Also I feel bad just saying "yep everything looks good!" over and over again. :)

Re: docs, I was just noting that, if I were to look at this DAG stuff without having read it before, I wouldn't really know where to start reading. As long as there is some plan for a tutorial in the future, there's no need to worry about it at this stage.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCameraToReflected.java
Cervator: `ReflectCamera` ? `SetReflectedCamera` ? Nothing less obvious jumping out at me ...

engine/src/main/java/org/terasology/rendering/dag/tasks/LookThroughDefaultCameraTask.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.lwjgl.opengl.GL11.GL_MODELVIEW;
+import static org.lwjgl.opengl.GL11.GL_PROJECTION;
+import static org.lwjgl.opengl.GL11.glLoadIdentity;
+import static org.lwjgl.opengl.GL11.glMatrixMode;
+
+/**
+ * Instances of this class reset the ModelView and Projection matrices to identity matrices,
+ * as per opengl default.
+ *
+ * WARNING: RenderPipelineTasks are not meant for direct instantiation and manipulation.
```
Cervator: If these classes shouldn't be used directly should there be some clever way to prohibit a user from acquiring an instance? Like a package private factory method, more access limited `execute`, or something otherwise sneaky.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCameraToReflected.java
emanuele3d: I did consider ReflectCamera. It does sound nice and concise. I'm only concerned about the "Reflect" term as it could be interpreted as related to java Reflection, while here it's more about mirroring/flipping an image rather than deep-inspecting a java object.

Did you interpret it that way by any chance?

engine/src/main/java/org/terasology/rendering/dag/tasks/LookThroughDefaultCameraTask.java
emanuele3d: Good point. I've been wondering on that. I have to verify if there is a 1-to-1 relationship between state changes and tasks. If there is I think I will keep the RenderPipelineTask interface public but the specific implementations could become internal private classes in the respective state changes. I believe most tasks can be handled that way.

For those tasks that for whatever reason might be related to more than one state change I'll follow your suggestion and limit access via package-private methods/classes.

Meanwhile, I've added issue #2796 to the Renderer Overhaul epic, so that I don't loose sight of this.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetCameraToReflected.java
Cervator: Just for a brief second - but the inclusion of "Camera" makes it pretty clear cut I think :-)

Any additional confirmation of details would be in the Javadoc which is close by hand anyway.

GooeyHub:

emanuele3d:
Hey guys, submitted a new commit to address a few minor issues. Have a look. I don't expect it to be controversial. 

If that's ok I think this PR is good to go.

Cervator:
Retested and merged :-)

Cervator:
I failed at subtle testing :D See #2800

PR #2798, opened 2017-02-22
Fixed auto movement

zeokav:
### Contains

A couple of additions for #2557.
- Fixed the jerky movement because of CharacterImpulseEvent. This approach simply simulates the forwards key pressed down.
- The character no more stops after jumping or strafing.
- Added code to simulate key presses. This can be extended to other tasks that require pressing of a key through code as if it was pressed on the keyboard.

### How to test

1. Start a game
2. Press R to start moving.
3. Confirm that character starts moving and stops when one of forward move button, backward move button or auto move button is pressed.

Nothing as of now. Let me know if there are any issues with this implementation while testing. I couldn't find any.

### DIFF
--- a/engine/src/main/java/org/terasology/input/InputSystem.java
+++ b/engine/src/main/java/org/terasology/input/InputSystem.java
@@ -17,6 +17,9 @@ package org.terasology.input;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Queues;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.terasology.config.BindsConfig;
 import org.terasology.config.Config;
 import org.terasology.config.ControllerConfig.ControllerInfo;
@@ -63,6 +66,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Queue;
 
 /**
  * This system processes input, sending it out as events against the LocalPlayer entity.
@@ -107,6 +111,10 @@ public class InputSystem extends BaseComponentSystem {
     private Map<Input, BindableRealAxis> controllerAxisBinds = Maps.newHashMap();
     private BindableButtonImpl mouseWheelUpBind;
     private BindableButtonImpl mouseWheelDownBind;
+
+    private Logger logger = LoggerFactory.getLogger(InputSystem.class);
+
+    private Queue<KeyboardAction> simulatedKeys = Queues.newArrayDeque();
 
     private boolean capturingMouse = true;
 
@@ -420,8 +428,39 @@ public class InputSystem extends BaseComponentSystem {
         }
     }
 
+
+    /**
+     * Simulated key strokes: To simulate input from a keyboard, we simply have to extract the Input associated to the action
+     * and this function adds it to the keyboard's input queue.
+     * @param key The key to be simulated.
+     */
+    public void simulateSingleKeyStroke(Input key) {
+        /* TODO: Perhaps there is a better way to extract the character.
+            All the simulate functions extract keyChar by getting the first character from it's display string.
+            While it works for normal character buttons, might not work for special buttons if required later.
+        */
+        char keyChar = key.getDisplayName().charAt(0);
+        KeyboardAction action = new KeyboardAction(key, ButtonState.DOWN, keyChar);
+        simulatedKeys.add(action);
+    }
+
+    public void simulateRepeatedKeyStroke(Input key) {
+        char keyChar = key.getDisplayName().charAt(0);
+        KeyboardAction action = new KeyboardAction(key, ButtonState.REPEAT, keyChar);
+        simulatedKeys.add(action);
+    }
+
+    public void cancelSimulatedKeyStroke(Input key) {
+        char keyChar = key.getDisplayName().charAt(0);
+        KeyboardAction action = new KeyboardAction(key, ButtonState.UP, keyChar);
+        simulatedKeys.add(action);
+    }
+
     private void processKeyboardInput(float delta) {
-        for (KeyboardAction action : keyboard.getInputQueue()) {
+        Queue<KeyboardAction> keyQueue = keyboard.getInputQueue();
+        keyQueue.addAll(simulatedKeys);
+        simulatedKeys.clear();
+        for (KeyboardAction action : keyQueue) {
             boolean consumed = sendKeyEvent(action.getInput(), action.getInputChar(), action.getState(), delta);
 
             // Update bind
--- a/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
@@ -15,8 +15,6 @@
  */
 package org.terasology.logic.players;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.terasology.assets.ResourceUrn;
 import org.terasology.config.BindsConfig;
 import org.terasology.config.Config;
--- a/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
@@ -18,7 +18,9 @@ package org.terasology.logic.players;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.assets.ResourceUrn;
+import org.terasology.config.BindsConfig;
 import org.terasology.config.Config;
+import org.terasology.engine.SimpleUri;
 import org.terasology.engine.Time;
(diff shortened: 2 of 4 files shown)


GooeyHub:

Cervator:
Oh cool! You figured it out then, well done :-)

No time to check this out tonight, but will trigger the bot and try it out when I get a chance!

@GooeyHub: ok to test

GooeyHub:

engine/src/main/java/org/terasology/input/InputSystem.java
```diff
@@ -420,8 +428,39 @@ private void processControllerInput(float delta) {
...
+     * Simulated key strokes: To simulate input from a keyboard, we simply have to extract the Input associated to the action
+     * and this function adds it to the keyboard's input queue.
+     * @param key The key to be simulated.
+     */
+    public void simulateSingleKeyStroke(Input key) {
+        /* TODO: Perhaps there is a better way to extract the character.
+            All the simulate functions extract keyChar by getting the first character from it's display string.
+            While it works for normal character buttons, might not work for special buttons if required later.
+        */
+        char keyChar = key.getDisplayName().charAt(0);
```
skaldarnar: Should the `Input` interface be extended to provide the required information, i.e., outright provide the character/key code?

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -222,28 +227,21 @@ private void standPlayer(EntityRef entity) {
      */
     private void stopAutoMove() {
         isAutoMove = false;
+        List<Input> inputs = bindsConfig.getBinds(new SimpleUri("engine:forwards"));
+        Input forwardKey = inputs.get(0);
```
skaldarnar: not very likely, but can the `inputs` list be empty (no forward key defined)?

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -222,28 +227,21 @@ private void standPlayer(EntityRef entity) {
...
-            } else {
-                // If isAutoMove turns false, cancel the action.
-                delayManager.cancelPeriodicAction(entity, AUTO_MOVE_ID);
-            }
-        }
+        bindsConfig = config.getInput().getBinds();
+        List<Input> inputs = bindsConfig.getBinds(new SimpleUri("engine:forwards"));
+        Input forwardKey = inputs.get(0);
+        inputSystem.simulateSingleKeyStroke(forwardKey);
+        inputSystem.simulateRepeatedKeyStroke(forwardKey);
```
skaldarnar: is it always necessary to trigger a single key stroke before a repeated key stroke? if so, this should be documented in the Javadoc, otherwise triggering a repeated key stroke should be sufficient(?)

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
```diff
@@ -33,7 +33,7 @@ public void initialise() {
         WidgetUtil.trySubscribe(this, "settings", widget -> getManager().pushScreen("settingsMenuScreen"));
         WidgetUtil.trySubscribe(this, "mainMenu", widget -> CoreRegistry.get(GameEngine.class).changeState(new StateMainMenu()));
         WidgetUtil.trySubscribe(this, "exit", widget -> CoreRegistry.get(GameEngine.class).shutdown());
-        WidgetUtil.trySubscribe(this, "crashReporter", widget -> CrashReporter.report(new Throwable("Report an error."), LoggingContext.getLoggingPath()));
+//        WidgetUtil.trySubscribe(this, "crashReporter", widget -> CrashReporter.report(new Throwable("Report an error."), LoggingContext.getLoggingPath()));
```
skaldarnar: why did you comment this? is it a mistake or is there a bug? if it should be removed the code can be deleted.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/MainMenuScreen.java
```diff
@@ -16,7 +16,7 @@
 
 package org.terasology.rendering.nui.layers.mainMenu;
 
-import org.terasology.crashreporter.CrashReporter;
+//import org.terasology.crashreporter.CrashReporter;
```
skaldarnar: same here.

engine/src/main/java/org/terasology/input/InputSystem.java
zeokav: That's what I thought, but what about the case where the user has move forward set to the up arrow on his keyboard? It will return a string "UP" and not a single character. Also, no matter what the value of keyChar is in the KeyboardAction instance is, it works. I added this since there was no constructor without it. So, I kinda left it the way it is right now.

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
zeokav: Could be, but why would anyone want that? 😛 
I tested and it throws a NPE but it doesn't really affect the game. It works only if forward is defined. What exactly are the other means of moving forward, if the user has no keyboard binding for forward movement, hypothetically?

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
zeokav: Yep, it's required. I logged how keyboard events occur and the natural order of occurrence is - Down, Repeat, Repeat, Repeat... , Up. I'll mention it in the docs.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
zeokav: The code I pulled from develop didn't have the CrashReporter class, so I commented it out until that's implemented (?)
Doesn't build if it's there.

GooeyHub:

engine/src/main/java/org/terasology/input/InputSystem.java
skaldarnar: What is the `keyChar` used for then? An alternative might be to provide a different constructor for `KeyboardAction` (or a new sub-class for `AutomaticKeyboardAction`.

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
skaldarnar: yeah, that's exactly what I was thinking about (the question was a bit unclear, though). Since you cannot assume that the player has defined that key, you should check for the empty list and do nothing (instead of provoking the NPE). 
It _should_ not happen, but a hard crash is not what we want here.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
skaldarnar: hmm, that sounds weird. @Cervator do you know what's up with this?

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
zeokav: It's not hard crashing, but I'll put a check for an empty input (plus the new constructor) and push it soon!

engine/src/main/java/org/terasology/input/InputSystem.java
zeokav: So I can't extend the KeyboardAction class because it's a final class and I don't want to risk breaking anything by making it non static. Instead I decided to make a constructor that doesn't take inputChar. 

> public KeyboardAction(Input input, ButtonState state) {
>         this.input = input;
>         this.state = state;
>     }

But then, inputChar is final too! Do I just give any random character? I just don't know if this inputChar value is required anywhere later. 
As of now, I don't see it getting used anywhere. It's added in LwjglKeyboardDevice, but like not really used anywhere else...

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/PauseMenu.java
Cervator: Might have pulled latest code without doing a fresh `gradlew idea` ? The CR dependency was recently moved around a bit (from PC facade back to engine) so it could be hit from the main menu screen. 

That would fit the symptom pretty perfectly so hopefully that's it :-)

zeokav: Oops. It was that. I reverted back and pushed the code. :)

GooeyHub:

skaldarnar:
@zeokav I'll merge this once the NPE is checked - everything else looks good and works well! 👍

zeokav:
@skaldarnar Alright, done. :)

GooeyHub:

GooeyHub:

PR #2799, opened 2017-02-22
Ping to all clients

### LINKED ISSUE #2787: Improvements for ping system
Follow-up to #2768 - an improvement is desired to where the server pings all connected clients both for health-check type functionality and so that information can be provided to clients to show a ping overlay for each player (holding down `tab` in a multiplayer session)

Should be done without excessive connection usage and disposal, as indicated in #2768

There is also a small side-issue where a server-client player cannot use the ping command in the console as it believes said player is not engaged in a multiplayer session. That might just be a cosmetic issue to clear up for now since that setup may change in the future anyway (improvements to `Context` and possibly splitting apart single player to having a local server)

This shouldn't be *too* hard so marking bite-size, although it probably isn't a super beginner task. Check the linked PR to see some related code and consider reading https://github.com/MovingBlocks/Terasology/wiki/Entities,-Components-and-Events-on-the-Network

GabrielXia:
### Contains

"Fixes #2787", implement ping to all clients

### How to test

- I think we can use `networkSystem.getMode()` to split apart single player to a local server, it works in ping command

-  As my understanding of #2768(tell me if I'm wrong), we can use Event to perform ping. In this code, `ServerPingSystem` will send `PingFromServerEvent` to every client in every 10 second. Every client will send back `PingFromClientEvent`.
screen shot 2017-02-22 at 10 40 44 pm

The code is not finished yet, there are several questions, very glad if someone can help me :-)

- [x] For now the ping time information is only performed in server, it doesn't update in client. I've used 

```
@Replicate 
public String ping = "";
```
it doesn't work :-( 
- [ ] Using regular event to perform ping seems to be much slower than real ping. For example, in the same game, client ping(ping command) to server returns 5ms, but server ping(event ping) to this client returns 300ms. Is that because I don't use a new thread ? But I think the comparison between different clients may still be reliable.  
- [ ] Need to consider case where one client doesn't response


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/logic/chat/ChatSystem.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 MovingBlocks
+ * Copyright 2016 MovingBlocks
```
Cervator: Feel free to bump all the way to 2017 any time you modify a file with out of date years :D

engine/src/main/java/org/terasology/logic/chat/ChatSystem.java
```diff
@@ -172,4 +172,4 @@ public String whisper(
 
         return senderMessage;
     }
-}
+}
```
Cervator: Best practice: Leave a single empty new line after the final `}` (otherwise GitHub will give you that angry red symbol)

engine/src/main/java/org/terasology/network/ServerPingSystem.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.network;
+
+
```
Cervator: Stray empty line. Literally unmergable! (not really - just kidding. Just a general practice to only leave one empty line between blocks)

engine/src/main/java/org/terasology/network/ServerPingSystem.java
```diff
@@ -0,0 +1,105 @@
...
+import org.terasology.network.events.PingFromClientEvent;
+import org.terasology.network.events.PingFromServerEvent;
+import org.terasology.registry.In;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.HashMap;
+
+/**
+ * This system implement the server ping to all clients(include server local player)
```
Cervator: Ideally leave a `.` at the end of the first sentence in Javadoc to indicate where it'll end the initial summary of the class (the parser can usually figure it out anyway, just another good practice).

Also space between "clients" and "(include"

Maybe a second sentence to help describe a little more how it works?

engine/src/main/java/org/terasology/network/ServerPingSystem.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+@RegisterSystem
```
Cervator: More a question for other reviewers - should this system have a specific registration mode (client, server, or both)? It'll default this way to `ALWAYS` but I figure maybe being explicit helps clarity.

engine/src/main/java/org/terasology/network/ServerPingSystem.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @ReceiveEvent(components = ClientComponent.class)
+    public void onConnect(ConnectedEvent event, EntityRef entity) {
+        if(networkSystem.getMode().isServer()) {
+                delayManager.addPeriodicAction(entity, PING_ACTION_ID, 10000, 10000);
+        }
+    }
+
+    @ReceiveEvent(components = ClientComponent.class)
+    public void onDisconncet(DisconnectedEvent event, EntityRef entity) {
```
Cervator: Typo: `onDisconncet` should be `onDisconnect`

engine/src/main/java/org/terasology/network/events/PingFromClientEvent.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.network.events;
+
+import org.terasology.entitySystem.event.Event;
+import org.terasology.network.ServerEvent;
+
+/**
+ * This event, from client to server, implement one part of ping to all clients
```
Cervator: Which part? :-) This is probably actually good enough but it is written a little awkwardly. Maybe:

"This event handles the ping from a client to the server"

engine/src/main/java/org/terasology/network/events/PingFromServerEvent.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.network.events;
+
+import org.terasology.entitySystem.event.Event;
+import org.terasology.network.OwnerEvent;
+
+/**
+ * This event, from server to client, implement one part of ping to all clients
```
Cervator: ... while this one handles from server to all clients

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/OnlinePlayersOverlay.java
```diff
@@ -50,7 +52,8 @@ private String determinePlayerListText() {
                 sb.append("\n");
             }
             ClientComponent clientComp = clientEntity.getComponent(ClientComponent.class);
-            sb.append(PlayerUtil.getColoredPlayerName(clientComp.clientInfo));
+            sb.append(PlayerUtil.getColoredPlayerName(clientComp.clientInfo) +
```
Cervator: No real need to line break here. It is only 138 characters, well within our code style max line length.

Better yet might be fixing the pre-existing inspection warning over using String concatenation with a StringBuffer (could use two append calls). Also missing a space between `"` and `+` (super tiny issue)

GooeyHub:

GabrielXia:
Thanks a lot for comments on code with patience, it really helps a lot !!!

- I tested code in branch `develop` (without functionality ping to all clients), and encountered the same error. I'm wondering if it's because we tested two instances locally.
~~~
17:19:40.272 [New I/O worker #1] INFO  o.t.n.i.ServerHandshakeHandler - Sending Server Hello
17:19:40.376 [New I/O worker #2] INFO  o.t.n.i.ServerHandshakeHandler - Sending Server Hello
17:19:40.424 [New I/O worker #1] WARN  o.t.network.internal.ServerHandler - Unexpected exception from client
java.io.IOException: Connection reset by peer
	at sun.nio.ch.FileDispatcherImpl.read0(Native Method)
...
~~~
- With the help of @iojw, now the ping information has updated in every clients :D
 
Actually, "0ms" means that the ping is really fast, in the case server ping to itself; No text means there is no ping response. @Cervator Could you please test again to see whether the same problem exists ?  

I think we also need to fix the case where clients doesn't response. In that case it will show the last ping time for now, but in fact it has to show sometime like "player stalled". Any comments are welcomed !

engine/src/main/java/org/terasology/network/ServerPingSystem.java
```diff
@@ -0,0 +1,106 @@
...
+    public void onPingFromServer(PingFromServerEvent event, EntityRef entity){
+        ClientComponent client = entity.getComponent(ClientComponent.class);
+        if (client.local) {
+            entity.send(new PingFromClientEvent());
+        }
+    }
+
+    @ReceiveEvent(components = ClientComponent.class)
+    public void onPingFromClient(PingFromClientEvent event, EntityRef entity){
+        if (networkSystem.getMode().isServer()) {
```
flo: In the `@ReceiveEvent` annotation and in the `@RegisterSystem` annotation you can specify if a event handler / all event handlers of that class are for a server or client. So you don't need to check the mode of networkSystem.

engine/src/main/java/org/terasology/network/ServerPingSystem.java
```diff
@@ -0,0 +1,106 @@
...
+    private DelayManager delayManager;
+
+    private HashMap<EntityRef, Instant> startMap = new HashMap<>();
+
+    private HashMap<EntityRef, Instant> endMap = new HashMap<>();
+
+    @ReceiveEvent(components = ClientComponent.class)
+    public void onConnect(ConnectedEvent event, EntityRef entity) {
+        if(networkSystem.getMode().isServer()) {
+                delayManager.addPeriodicAction(entity, PING_ACTION_ID, 10000, 10000);
```
flo: I would let the system implement UpdateSubscriberSystem and then simply do the ping when a field called  now > timeOfNextPing. However your solution works as well.

engine/src/main/java/org/terasology/network/ServerPingSystem.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @In
+    private EntityManager entityManager;
+
+    @In
+    private DelayManager delayManager;
+
+    private HashMap<EntityRef, Instant> startMap = new HashMap<>();
+
+    private HashMap<EntityRef, Instant> endMap = new HashMap<>();
```
flo: The maps do not get cleaned up yet so we have a minor memory leak here.

GooeyHub:

GabrielXia:
@flo Thanks :-)
I did some modification :
- Changed this function on need base, use command `subscribePing` to subscribe or to cancel
- Split `PingSystem` to two system : `ServerPingSystem` on server and `ClientPingSystem` on client
- Now it only displays the ping information of player itself, and I put it in the `DebugOverlay`
screen shot 2017-03-01 at 2 54 25 pm


I'm pretty curious about the further use of this function !!
@flo Yeah I tested in a headless server, the ping time is much smaller.
Any comments :-)

GooeyHub:

GabrielXia:
@flo Do you mean that server ping to all clients and give ping information of all clients to the client who subscribes ?

flo:
yes and of course it would only do the pings if there is someone interested.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
```diff
@@ -170,6 +162,15 @@ public Boolean get() {
         metricsLabel = find("metrics", UILabel.class);
     }
 
+    private String getPing() {
+        if (localPlayer.getClientEntity().hasComponent(PingStockComponent.class)) {
+            PingStockComponent pingStockComp = localPlayer.getClientEntity().getComponent(PingStockComponent.class);
```
flo: This could be further simplified by first getting pingStockComp and then checking if the result is != null. Saves one lookup and simplifies the code.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/DebugOverlay.java
```diff
@@ -170,6 +162,15 @@ public Boolean get() {
...
+            return new StringBuffer(pingStockComp.pingValue.toString()).append(" ms").toString();
```
flo: why not just `return pingStockComp + " ms";`

GooeyHub:

GooeyHub:

GabrielXia:
@flo Thanks for advice, now we can see all clients' ping in online player list :-)

For the easy usage I just simplement tell the user that ping is in the online player overlay, we may modify it after.

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/network/ServerPingSystem.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.HashMap;
+
+/**
+ * This system implement the server ping to clients on need base.
+ * It runs on the server, pings to all clients who subscribe this function.
+ */
+@RegisterSystem(RegisterMode.AUTHORITY)
+public class ServerPingSystem extends BaseComponentSystem implements UpdateSubscriberSystem {
+
+    private static final long pingPeriod = 10000;
```
flo: Just noticed this: Wouldn't it be cooler if it would be a lower value like 200?

To make it real time you could change the initialise method of OnlinePlayersOverlay to:
```
        text.bindText(new ReadOnlyBinding<String>() {
            @Override
            public String get() {
                PingStockComponent pingStockComp = localPlayer.getClientEntity().getComponent(PingStockComponent.class);
                if (pingStockComp == null) {
                    String playerListText = determinePlayerListText();
                    return playerListText;
                } else {
                    String playerAndPing = determinePlayerAndPing(pingStockComp);
                    return playerAndPing;
                }
            }
        });
```
instead of setting the text during the opening. During the opening it could subscribe.to pings and during closing it could unsubscribe.

oniatus:
Also an update since the last review batch after some IRC talk: The Networking depends now on a replication of the PingStockComponent. The Server updates the component batchwise for all clients, then saves it once. This should redue the amount of network events, as the component update triggers only one event instead of single events per client to all other clients.
One drawback is a bit of serialization overhead, as the data format is a `Map<EntityRef,Long>` which is not supported by network serialization. Therefore the Map is transferred as two Lists, one for the keys and one for the values (see the TODO in PingStockComponent). We could open an enhancement issue, if one wants to dig into serialization and improve the data type support but for the moment I am fine with the solution :)

+1 to merge from my side 👍

GabrielXia:
Thanks @flo @oniatus  :)
- Yeah 10,000 is too big, but I think it might not be too small : if a player's ping is large, like 300ms, then the server receive the older pingFromClient after sending the newer pingToClient, so that it will consider the ping value is 100 ms. It depends really on the client's ping value ... I test in my pc, 400 looks cool :)
- Play in real time looks cool ! @flo What do you mean by "During the opening it could subscribe.to pings and during closing it could unsubscribe." ?

GooeyHub:

Cervator:
@flo be my guest, please merge any time you're happy with something! :-)

@GabrielXia thanks for keeping at this!

flo:
@GabrielXia 

I mean the open and close events of the `CoreScreenLayer` which you can subscribe to by overriding the following methods:
```
 @Override
    public void onOpened() {
      // 
    }

    @Override
    public void onClosed() {

    }

```
With subscribing/unsubscribing to the pings I meant that you do the logic that is currently in your `showPingInPlayerlist` command. Since the command currently runs on the server it could be that you need to create a new server event (Class that implements Event and is annotated with `@ServerEvent`) so that you can let the client inform the server. It might be possible that you don't need an event as there is a way to synchronize changes of a owner of a entity to the server (    `@Replicate(FieldReplicateType.OWNER_TO_SERVER)`). 

About the issue with pings beeing faster than X ms: I think ideally you would make your solution roboust against that by either:
1. Sending no further ping to a client until it has replied
2. Sending the current server time in the ping to the client and have the client reply with the original server time in it's response. Then you could compare server sent time with server receive time.

GooeyHub:

flo:
Tested it and looks good. Works even properly with headless server + client.

It's now merged, thanks for your PR and your great work.

PR #2801, opened 2017-02-24
Fix for Issue 2800

### LINKED ISSUE #2800: Subtle rendering issue introduced by #2794 - billboard floating position off
Took me testing two additional PRs after #2794 to notice this sneaky issue: the position a billboard is rendered at seems off based on how the player is moving. I found two ways in which it is exhibited (may be more):

1. When breaking a billboard plant while the player is moving toward and nearly has crossed into the blockspace holding the plant the breaking effect can weirdly float off to the side at the bottom of the window. Similar oddities can happen but that's the easiest to reproduce. Simply walk into billboard plants while breaking them right as you're entering their block.
1. In a multiplayer session walk one player up to another one. Observe their floating name tag acting weird (floating off to the side again). Also easy to reproduce but does take running at least two game clients connected in a world together.

Example of the name tag issue: 

terasology-170223230853-1152x720

Confirmed it did not exist before the introduction of #2794 so it is definitely caused by that. Also makes sense considering the topic of the PR.

Sending this your way @emanuele3d - you're keeping me on my toes! :D I suspect this is quirky but probably not very complex to solve. Nor is it a big deal so feel free to ship it with something else if you already have pending changes for other stuff.

emanuele3d:
### Contains
Fixes #2800

### How to test
Break some bushes: a small cloud of rapidly disappearing particles/leaves should appear where the bush used to be - this is the normal behavior. 

The Bug detailed in 2800 had the particle appear in the lower portion of the field of view of the player, in a frame of coordinates seemingly unrelated to the visible world.

### Notes
This problem was due to the SimpleBlendMaterialsNode implicitly relying on other nodes to set the correct modelview and projection matrices. After the merging of PR #2794 nodes _must_ either add a StateChange to look through a specific camera or run their code assuming the projection and modelview matrices being identity matrices - opengl's default. 

As testing of #2794 didn't include semi-transparent objects, this problem was not noticed.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/SimpleBlendMaterialsNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/SimpleBlendMaterialsNode.java
@@ -19,6 +19,7 @@ import org.terasology.engine.ComponentSystemManager;
 import org.terasology.entitySystem.systems.RenderSystem;
 import org.terasology.monitoring.PerformanceMonitor;
 import org.terasology.registry.In;
+import org.terasology.rendering.cameras.Camera;
 import org.terasology.rendering.dag.AbstractNode;
 import org.terasology.rendering.dag.stateChanges.BindFBO;
 import static org.lwjgl.opengl.GL11.GL_ONE_MINUS_SRC_ALPHA;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Tested out, both scenarios are fine now, and the code pretty simple and cool to see. Thanks! :-)

Issue #2797, opened 2017-02-22
Refactor Camera to be Component based


Cervator:
One of our GSOC items and a long standing desired feature. Rather than having a hard coded camera attached to the player let *any* arbitrary entity hold a CameraComponent so a camera viewpoint can be sourced from it (imagine a security camera and a monitor block that can show the view from said security camera)

May have API impact but @emanuele3d would be the expert there :-)

Maybe related: #1272, #1126, http://forum.terasology.org/threads/associating-renderableworlds-to-cameras.1221, http://forum.terasology.org/threads/cinematic-module.1112, http://forum.terasology.org/threads/skysphere.1004

zeokav:
Hi @Cervator ! The idea of having a separate camera component seems interesting. (Didn't know there was just one fixed camera :o)
I'm going to try my hand at GSoC this time, and would like to work on something that visibly affects the game. Is there somewhere I can start looking to have a better understanding of the code-base and help with this already? Starting slightly early will help me write a better application, I hope.

skaldarnar:
While trying out my new GPU w/ recording the last day I noticed that the player's hand/held item is out of sync when using a smooth camera mode and high speed multiplier. Maybe you can look into that to get your hands on the rendering/camera topics.

Cervator:
@zeokav very cool to hear about GSOC potential! Yes getting an early start is great and a huge advantage as selection is likely to be competitive.

Other than the few links I added in I'm unsure about deeper technical details. I know @emanuele3d would know some stuff, as likely would @immortius and certainly @begla if he were to coincidentally fly by one of these days :-)

emanuele3d:
@zeokav: I confirm I'd be available for some support/mentoring for this project. 

On top of what Cervator already said I'd like to give you a glimpse of the goal. 

Generally speaking in an architecture based on an Entity System (ES) components are pretty dumb and it is systems where all the magic happens: systems get the data from components and if necessary update the data on those components. 

Right now cameras are pretty "clever" in the sense that they contain a lot of code in them. The goal is to have a camera component that only contains the data describing the characteristics proper of a camera. For example, if you think about a physical camera, its position and orientation are _not_ characteristics proper of a camera. They are attributes of pretty much any existing object which is why Terasology already have position and orientation components that can be associated with any object. On the other hand, focal length and, given we are in an opengl world, near/far distances _are_ characteristics of a camera. As is interpupillary distance if we are dealing with a VR camera.

[...]

PR #2803, opened 2017-03-01
Head underwater check function moved (following to-do comment in-code)

fabriond:
### Contains

Moves isHeadUnderWater function from WorldRenderer.java to Camera.java (function name changed to isCameraUnderWater) and other minor changes made accordingly.
(This change was in the to-do comment right over the isHeadUnderWater function in the WorldRenderer class)

### How to test

This change isn't supposed to make a change in-game, but to make the code more readable so, to test it, just compare the functionality of the underwater camera with and without the change.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
+++ b/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
@@ -25,6 +25,7 @@ import org.terasology.math.geom.Vector3i;
 import org.terasology.monitoring.PerformanceMonitor;
 import org.terasology.rendering.assets.material.Material;
 import org.terasology.rendering.cameras.Camera;
+import org.terasology.rendering.cameras.SubmersibleCamera;
 import org.terasology.rendering.world.viewDistance.ViewDistance;
 import org.terasology.rendering.world.WorldRenderer;
 import org.terasology.world.WorldProvider;
@@ -87,8 +88,8 @@ public class HeadlessWorldRenderer implements WorldRenderer {
     }
 
     @Override
-    public Camera getActiveCamera() {
-        return noCamera;
+    public SubmersibleCamera getActiveCamera() {
+        return (SubmersibleCamera) noCamera;
     }
 
     @Override
--- a/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
+++ b/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
@@ -17,7 +17,6 @@ package org.terasology.engine.subsystem.headless.renderer;
 
 import com.google.common.collect.Lists;
 import org.terasology.config.Config;
-import org.terasology.config.RenderingConfig;
 import org.terasology.context.Context;
 import org.terasology.logic.players.LocalPlayerSystem;
 import org.terasology.math.Region3i;
@@ -294,16 +293,5 @@ public class HeadlessWorldRenderer implements WorldRenderer {
             return distance2 > distance ? -1 : 1;
         }
     }
-
-	@Override
-	public RenderingConfig getRenderingConfig() {
-		// TODO Auto-generated method stub
-		return null;
-	}
-
-	@Override
-	public WorldProvider getWorldProvider() {
-		return worldProvider;
-	}
 
 }
--- a/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
+++ b/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
@@ -17,6 +17,7 @@ package org.terasology.engine.subsystem.headless.renderer;
 
 import com.google.common.collect.Lists;
 import org.terasology.config.Config;
+import org.terasology.config.RenderingConfig;
 import org.terasology.context.Context;
 import org.terasology.logic.players.LocalPlayerSystem;
 import org.terasology.math.Region3i;
@@ -175,12 +176,6 @@ public class HeadlessWorldRenderer implements WorldRenderer {
     }
 
     @Override
-    public boolean isHeadUnderWater() {
-        // TODO Auto-generated method stub
-        return false;
-    }
-
-    @Override
     public float getMillisecondsSinceRenderingStart() {
         // TODO Auto-generated method stub
         return 0;
@@ -299,5 +294,16 @@ public class HeadlessWorldRenderer implements WorldRenderer {
             return distance2 > distance ? -1 : 1;
         }
     }
+
+	@Override
+	public RenderingConfig getRenderingConfig() {
+		// TODO Auto-generated method stub
+		return null;
+	}
+
+	@Override
+	public WorldProvider getWorldProvider() {
+		return worldProvider;
+	}
 
 }
--- a/engine/src/main/java/org/terasology/rendering/cameras/Camera.java
+++ b/engine/src/main/java/org/terasology/rendering/cameras/Camera.java
@@ -239,5 +239,4 @@ public abstract class Camera {
     public boolean hasInSight(AABB aabb) {
         return viewFrustum.intersects(aabb);
     }
-    
 }
--- a/engine/src/main/java/org/terasology/rendering/cameras/Camera.java
+++ b/engine/src/main/java/org/terasology/rendering/cameras/Camera.java
@@ -76,6 +76,11 @@ public abstract class Camera {
     /* (Water) Reflection */
     private boolean reflected;
     private float reflectionHeight = 32;
+    
+    /* Used for Underwater Checks */
+    // TODO: Get these variables using injection
+    private WorldProvider worldProvider;
+    private RenderingConfig renderingConfig;
 
     /**
      * Applies the projection and modelview matrix.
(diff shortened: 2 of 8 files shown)


GooeyHub:

emanuele3d:
Please stand-by. I should be able to provide feedback in the next 12 hours or so.

Cervator:
Triggering our bot for a test build :-)

@GooeyHub: ok to test

Hi @fabriond! Prospective GSOC student by chance?

GooeyHub:

engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
```diff
@@ -300,4 +295,15 @@ public int compare(RenderableChunk o1, RenderableChunk o2) {
         }
     }
 
+	@Override
+	public RenderingConfig getRenderingConfig() {
```
emanuele3d: Have you looked into obtaining RenderingConfig and WorldProvider via injection instead? Have a look at the RenderNodes listed in WorldRendererImpl.initGraph() to see how it's done. It involves taking advantage of the @In annotation.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
```diff
@@ -239,4 +242,30 @@ public boolean isReflected() {
...
+    /**
+     * Returns True if the head of the player is underwater. False otherwise.
+     *
+     * Implementations must take in account waves if present.
+     *
+     * @param World Rendering Configuration and the World Provider
+     *
+     * @return True if the head of the player is underwater. False otherwise.
+     */    
+    public boolean isCameraUnderWater(RenderingConfig renderingConfig, WorldProvider worldProvider) {
```
emanuele3d: Let's rename this method to isUnderWater(), so that we do not get the repetition of camera.isCameraUnderWater(). Also, if we get the config and the provider via injection into the camera we'll be able to avoid passing arguments to this method.

engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
fabriond: I tried doing it via injection, as you suggested, but the game keeps crashing on me when I do it, I'll keep trying to implement it with injections, but for now I've managed to use setters to make it so that the method doesn't need parameters

GooeyHub:

engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
emanuele3d: Ok, I'm asking around on how to do this: there are people who know injection better. @msteiger? How do you make @In work in a class where it doesn't? We'd like to write something like:

```java
@In 
```

but worldProvider is null when you use it, even though it is alive.

Meanwhile, @fabriond, let's not use setters. WorldRendererImpl has the Context passed-in on construction. The Context has references to both WorldProvider and RenderingConfig stored in it. Pass the Context to the camera constructor and you have everything you need.

emanuele3d: Ok, meanwhile I found a bit more about injection.

Let's assume you want a PerspectiveCamera instance to have a worldProvider injected in it.

Within PerspectiveCamera you just write:

```java
@In 
private WorldProvider worldProvider.
```

Within the class that instantiates the camera (WorldRendererImpl in our case) you have to write something like:

```java
Camera myCamera = new PerspectiveCamera();
InjectionHelper.inject(myCamera, context);
```

The second line will fill the worldProvider field with the instance it finds in the context - the context is basically a <Class, Instance> map. Notice that in this scenario the @In annotated fields will be null on construction. So, if something needs to be done before the camera can be used, it will need a method called after injection to do it, i.e. an initialise() method. 

So, as far as I can see, you'd save a few lines of code if you just gave the context directly to the constructor and initialized the fields within it instead of using the @In annotations - I'm asking confirmation of this.

emanuele3d: Scrap all of the above. Having discussed these matters over the past 90 minutes or so, I now believe you should pass the WorldProvider and the RenderingConfig instances to the camera's constructor. 

No point in using injection in this circumstance given that the WorldRenderer instance already has references to them.

Sorry about the flip-flopping.

fabriond: for the OrthographicCamera (which calls super without parameters in its constructor) should I change it to call the new super(with worldProvider and renderingConfig as parameters) or just leave it like that? that camera is instanced in ShadowMapNode.java

fabriond: I assume it won't need the worldProvider or the renderingConfig, since the method that needs them is only called for the playerCamera in worldRenderer(which is never an OrtographicCamera) so I'll leave it as is (but if it does need the worldProvider and renderingConfig, it'd be better to inject them, since ShadowMapNode.java doesn't instance the worldProvider)

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
```diff
@@ -73,6 +76,18 @@
...
     private float reflectionHeight = 32;
+    
+    /* Used for Underwater Checks */
+    private WorldProvider worldProvider;
+    private RenderingConfig renderingConfig;
+    
+    public Camera(){
+    }
+    
+    public Camera(WorldProvider worldProvider, RenderingConfig renderingConfig){
```
emanuele3d: My concern here is that, as you rightly noticed, there are cameras that are not going underwater, i.e. the ShadowMapCamera.

One way to handle this would be to make a SubmersibleCamera class extending the Camera class and adding only the underwater check functionality. PerspectiveCamera and the OpenVRCamera would then extend that rather than the plain Camera class. If we look at the way the code works now, this would be enough.

We could use more sophisticated methods, so that cameras can acquire the underwater check functionality at runtime rather than have it statically compiled in. But I think for the time being let's keep it simple. Cameras are slated to go through some major changes anyway (see #2797) and it is best if we don't invest toooo much time on them.

engine/src/main/java/org/terasology/rendering/cameras/PerspectiveCamera.java
```diff
@@ -49,7 +51,8 @@
 
     private Vector3f tempRightVector = new Vector3f();
 
-    public PerspectiveCamera(PerspectiveCameraSettings cameraSettings) {
+    public PerspectiveCamera(PerspectiveCameraSettings cameraSettings, WorldProvider worldProvider, RenderingConfig renderingConfig) {
```
emanuele3d: The cameraSettings can be obtained via renderingConfig.getCameraSettings() - no need to pass them in explicitly anymore.

Cervator:
I missed the earlier question:

> I'm not so sure the camera itself should check if it's underwater.

Unsure here if I have good direct feedback for that. I do have an opinion on cameras *around* water - but it is more at the level of whether something like a third person camera could be swung around and above/under water, something that can be abused in other games to "peak" above or under water when you're on the other side, often due to the special effects that trigger.

If I picture a 3rd party camera you can move around the player's avatar in typical fashion I would say the logical placement of the avatar's eyes should matter somewhat. Sure a 3rd person camera isn't actually realistic unless you have camera drones, but still, we need to allow some limited freedom.

So the avatar's eyes is in one spot and probably using the mouse you could spin around the camera pointing at the avatar to get 3rd person mode. If you're near water I wouldn't want the camera to be able to be placed on the other side of the surface - so I'd want *either* location to be checking. If the avatar's eyes are above water then an independently moving camera shouldn't be allowed to move to a position *under* water, and vice versa.

How that translates into this exact PR, `Camera` classes and possible variants, or whether this little rant actually makes any sort of difference I am unsure of ... :-)

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/cameras/SubmersibleCamera.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public SubmersibleCamera(WorldProvider worldProvider, RenderingConfig renderingConfig) {
+	this.worldProvider = worldProvider;
+    	this.renderingConfig = renderingConfig;
+    }
+	
+   /**
+    * Returns True if the head of the player is underwater. False otherwise.
+    *
+    * Implementations must take in account waves if present.
```
emanuele3d: Well, this class -is- taking in account waves if present. So, I'd change to:

` * Takes in account waves if present.`

emanuele3d:
@Cervator: thank you for your feedback. Your design/interaction considerations are very reasonable and I share the gameplay concerns. I'd suggest to wrap them in a design issue to be tackled after cameras become components-based. In this context the movement constraints you are talking about would become constraints on the Location and Orientation components of the camera entity rather than the camera components themselves. And I'm guessing it would need some kind of ThirdPersonCameraController system to handle it all, which is somehow aware of where obstacles are and considers the water surface a hard or even a soft obstacles - can go through but perhaps never too much.

In any case @fabriond you don't have to worry about these concerns for now.

engine/src/main/java/org/terasology/rendering/cameras/PerspectiveCamera.java
```diff
@@ -49,8 +51,9 @@
 
     private Vector3f tempRightVector = new Vector3f();
 
-    public PerspectiveCamera(PerspectiveCameraSettings cameraSettings) {
-        this.cameraSettings = cameraSettings;
+    public PerspectiveCamera(WorldProvider worldProvider, RenderingConfig renderingConfig) {
+    	super(worldProvider, renderingConfig);
+        this.cameraSettings = renderingConfig.getCameraSettings();
```
fabriond: I'm not sure why, but that extra space is only showing in the commit changes, if you go into the actual file it doesn't have that extra space

emanuele3d: Ah. Peculiar. Never mind then! Thanks for taking a look.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
```diff
@@ -239,4 +239,5 @@ public boolean isReflected() {
     public boolean hasInSight(AABB aabb) {
         return viewFrustum.intersects(aabb);
     }
+
```
emanuele3d: You might want to remove this too, so that the Camera files remains unmodified.

GooeyHub:

GooeyHub:

GooeyHub:

emanuele3d:

engine/src/main/java/org/terasology/rendering/cameras/PerspectiveCamera.java
Cervator: It might be due to tab characters. It could look right on your system but not here via diff or in other workspaces. Tweak at your IntelliJ code style settings for the project and make sure to use 4 spaces instead of tabs :-)

Cervator:
Tested it out, fixed a few tabs to spaces, added you to the credits @fabriond - thanks! :-)

Cervator:
Oh, do you have a forum account? I'll get you a contributor badge and such if so, let me know. Helps to post an intro - unless you already have but just under a different name, happens sometime :-)

emanuele3d:
Good job @fabriond! Thank you for sticking with it. I look forward to other contributions from you.

fabriond:
yeah, I do have a forum account, it's fabriond as well and thanks @emanuele3d for the help :)

Cervator:
Excellent! You now have a contributor badge in the forum, just no posts to show it off with :D

PR #2809, opened 2017-03-06
Add damage type and instigator to death screen

iojw:
### Contains

This cleans up the changes made in #2736, which adds the damage type and instigator that caused the death to the death screen. The data is retrieved from `DoDestroyEvent`, so to ensure that accurate death information is displayed in the death screen in the future, different classes/modules have to specify the correct instigator and damage type there.

### How to test

You can kill the character directly through the `kill` command or through fall damage with the `hjump` command.


GooeyHub:

Cervator:
Thanks @iojw and @VaibhavBajaj :-)

After some workspace cleanup I could validate this in single player or as host+client in multiplayer, but I crash on a pure client in multiplayer with the following error:

```
23:51:59.737 [main] INFO  o.t.network.internal.ServerImpl - Destroying entity: EntityRef{id = 12724, netId = 1280, prefab = 'Potions:PoisonPotion'}
23:51:59.741 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
	at org.terasology.logic.players.MenuControlSystem.loadDeathDetails(MenuControlSystem.java:89)
	at org.terasology.logic.players.MenuControlSystemMethodAccess.invoke(Unknown Source)
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:506)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.sendStandardEvent(EventSystemImpl.java:269)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:260)
	at org.terasology.entitySystem.entity.internal.BaseEntityRef.send(BaseEntityRef.java:145)
	at org.terasology.network.internal.ServerImpl.processEvent(ServerImpl.java:244)
	at org.terasology.network.internal.ServerImpl.processMessages(ServerImpl.java:276)
	at org.terasology.network.internal.NetworkSystemImpl.update(NetworkSystemImpl.java:317)
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:406)
	at org.terasology.engine.Terasology.main(Terasology.java:152)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
```

It seems that `nuiManager.getScreen("engine:deathScreen")` on the pure client might be resolving as `null` - or my quick debugging session didn't work right.

Either of you able to confirm and take a look if it seems a valid problem?

engine/src/main/java/org/terasology/logic/health/DoDestroyEvent.java
```diff
@@ -36,11 +34,34 @@ public EntityRef getInstigator() {
...
     }
 
+    public String getInstigatorString() {
+        if (instigator == null) {
+            return "Unknown";
+        }
+        if (instigator.getParentPrefab() != null) {
+            String instigatorString = instigator.getParentPrefab().getName();
+            instigatorString = instigatorString.replaceAll("[A-Za-z]*:([A-Za-z]*)", "$1");
+            instigatorString = instigatorString.replaceAll("([A-Z])", " $1");
```
skaldarnar: Could you shortly explain what the format of the `parentPrefab.getName()` is, and what the `replaceAll` calls do?

engine/src/main/java/org/terasology/logic/players/MenuControlSystem.java
```diff
@@ -28,19 +28,26 @@
...
 import org.terasology.registry.CoreRegistry;
 import org.terasology.registry.In;
 import org.terasology.rendering.nui.NUIManager;
+import org.terasology.rendering.nui.layers.ingame.DeathScreen;
 import org.terasology.rendering.nui.layers.ingame.OnlinePlayersOverlay;
 import org.terasology.rendering.opengl.ScreenGrabber;
+import org.terasology.utilities.Assets;
 
 /**
  */
```
skaldarnar: Since you worked in this class, could you please provide a short javadoc?

engine/src/main/java/org/terasology/logic/players/MenuControlSystem.java
```diff
@@ -28,19 +28,26 @@
...
 @RegisterSystem(RegisterMode.CLIENT)
 public class MenuControlSystem extends BaseComponentSystem {
 
+    private String lastDamage = "Unknown";
+    private String lastInstigator = "Unknown";
```
skaldarnar: I'm not sure how far i18n is for the game, but "Unknown" should either be a constant reused or a key to a translatable string.

Cervator:
Note to self / when somebody merges this: I've been meaning to enter a bite-sized issue for getting a console command that does damage of type x to the player (for debugging stuff like this exact sort of functionality, or testing resist potions). Figured it would be best to do so after this is merged.

Oddly I could've sworn I already submitted such an issue, but I can't readily find it. So .. odd.

Also: @VaibhavBajaj & @iojw ping! :-) Want to merge this already!

GooeyHub:

GooeyHub:

iojw:
Sorry this took a while, I had to go through a few iterations to fix the issue with pure clients.

The issue as a pure client is that all the damage and death logic is handled on the server, with the character entity on the server being linked to the *server* client entity. However, the death screen is displayed by the *local* client entity, and this local client entity has no reference to the character entity on the server at all. As I had initially stored the details in a component on the character for replication, the local client entity could not retrieve the relevant details to display on the death screen locally. 

Thankfully I realised I could just store the details in the event itself - finally works and it's pretty short too. This should fix all the issues.

GooeyHub:

engine/src/main/java/org/terasology/logic/health/DoDestroyEvent.java
```diff
@@ -36,11 +38,36 @@ public EntityRef getInstigator() {
...
+
     public EntityRef getDirectCause() {
         return directCause;
     }
 
     public Prefab getDamageType() {
         return damageType;
     }
+
+    public String getDamageTypeString() {
```
oniatus: Same as above: We should support a display name component for damage types before falling back to the prefab name.
Assuming a module registers a damage type like `InternalBossEventPoisonDamage`, this is nothing we want to show the player directly.
Instead, the prefab for such a damage could look like this:
```
{
	"DisplayName" : {
		"name" : "Deadly Poison Aura"
	}
}

```
We can still keep the fallback code for prefabs without a name and maybe log a info that the damage type is missing a readable name. As player deaths should not occur multiple times per second, this would not create too much noise in the logs.
This would also leave an option to i18n the damage type names.

engine/src/main/java/org/terasology/logic/health/DoDestroyEvent.java
```diff
@@ -36,11 +38,36 @@ public EntityRef getInstigator() {
         return instigator;
     }
 
+    public String getInstigatorString() {
+        if (instigator.getParentPrefab() != null) {
```
oniatus: If the instigator was a player, this will always show "Player". I would like to see the actual player name here.
For the kill command, this does not work because `HealthCommands` adds only `clientComp.character` instead of `clientComp.clientInfo` (clientInfo is an Entity with the `DisplayNameComponent`, character does not contain a display name). Can you check if it makes more sense to pass the client info instead? Another option would be to duplicate the DisplayNameComponent to the character entity but this place may need some more investigation :)

It would also be nice to support a general DisplayNameComponent, the prefab name may not always be human readable (thinking of something like `SmeltingCapableCampFire`). Keeping the logic with the name parsing as fallback is still okay.

engine/src/main/java/org/terasology/logic/characters/CharacterSystem.java
```diff
@@ -73,10 +74,21 @@
...
     @ReceiveEvent(components = {CharacterComponent.class})
     public void onDeath(DoDestroyEvent event, EntityRef entity) {
         CharacterComponent character = entity.getComponent(CharacterComponent.class);
-        character.controller.send(new DeathEvent());
+        DeathEvent deathEvent = new DeathEvent();
+        //Store the details of the death in the event for display on the death screen
+        deathEvent.lastDamageType = event.getDamageTypeString();
+        if (event.getInstigatorString() != null) {
+            deathEvent.lastInstigator = event.getInstigatorString();
+        } else {
```
oniatus: I think this class is a better place to convert the prefabs/entities to readable names, i'd expect little to none logic in events but in systems instead. Move the two conversion methods from the event to this system.

engine/src/main/java/org/terasology/logic/characters/CharacterSystem.java
```diff
@@ -73,10 +74,21 @@
...
+    @In
+    private TranslationSystem translationSystem;
+
     @ReceiveEvent(components = {CharacterComponent.class})
     public void onDeath(DoDestroyEvent event, EntityRef entity) {
         CharacterComponent character = entity.getComponent(CharacterComponent.class);
-        character.controller.send(new DeathEvent());
+        DeathEvent deathEvent = new DeathEvent();
+        //Store the details of the death in the event for display on the death screen
+        deathEvent.lastDamageType = event.getDamageTypeString();
```
oniatus: rename both, method and field to `damageTypeName`, `getDamageTypeName`. Same for the instigator.

engine/src/main/java/org/terasology/logic/characters/CharacterSystem.java
```diff
@@ -73,10 +74,21 @@
...
+        if (event.getInstigatorString() != null) {
+            deathEvent.lastInstigator = event.getInstigatorString();
+        } else {
+            deathEvent.lastInstigator = translationSystem.translate("${engine:menu#unknown}");
+        }
```
oniatus: Is `unknown` appropriate for no instigator?
I tested it with the fall damage and `Unknown killed you with physical damage` is not what i would expect, more like `You died to physical damage`. Are there other cases, where `unknown` fits better?

oniatus:
ping @iojw, still around on this one? I think we are 1 or 2 iterations away from merging 👍

iojw:
@oniatus: yup, will get to it in a few days!

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

iojw:
Thanks @oniatus! Ping to @Cervator to maybe add the issue regarding the console command to inflict damage of a specific type since this has finally been merged.

PR #2810, opened 2017-03-06
Use client-side entity for held item, fixes #2605

### LINKED ISSUE #2605: Host Items at (0,0,0)

Tropid:
### Contains

Fixes #2605

This commit clones the picked up entity and creates a client-only entity for displaying the selected item. It does so by cloning the entity and removing the NetworkComponent and RigidBodyComponent. By creating the new entity no other systems like inventory and pickup systems should be affected.

Notes:
1) Item is only visible for the client wearing it (as far as I know it was not visible beforehand also).

2) In the previous commit nearly all components but some rendering-related ones where removed (whitelist). This also incorrectly removed the component responsible for emitting the light from a torch. So in this commit only the NetworkComponent and RigidBodyComponent are removed from the cloned entity (blacklist). This may lead to other problems.

3) Other clients still won't see the light from the torch, but I think that is no change to the previous behavior.

### How to test

I do not have a lot of spare time at the moment and am also missing knowledge about all the features of Terasology. Basically this commit creates a new client-only entity when picking up items, so everything regarding picking up items and placing them in a multiplayer game is in danger of being affected.

What did not work before this commit:
* Host wielded items are visible at (0, 0, 0), see #2605
* The previous fix removed the light effect from torches and some other side-effects, this one should not have the side-effects of the previous fix (see #2605)

Stuff I noticed was affected while developing:
* Client and Host should be able to destroy blocks, pick up the items, select them and place them
* All changes should be synchronized between host and client (item disappears when picked up, etc.)

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
@@ -19,7 +19,6 @@ import org.terasology.engine.Time;
 import org.terasology.entitySystem.entity.EntityBuilder;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
-import org.terasology.entitySystem.entity.lifecycleEvents.BeforeDeactivateComponent;
 import org.terasology.entitySystem.entity.lifecycleEvents.OnActivatedComponent;
 import org.terasology.entitySystem.entity.lifecycleEvents.OnChangedComponent;
 import org.terasology.entitySystem.event.ReceiveEvent;
@@ -37,11 +36,14 @@ import org.terasology.math.TeraMath;
 import org.terasology.math.geom.Quat4f;
 import org.terasology.math.geom.Vector3f;
 import org.terasology.network.ClientComponent;
+import org.terasology.network.NetworkComponent;
+import org.terasology.physics.components.RigidBodyComponent;
 import org.terasology.registry.In;
 import org.terasology.rendering.world.WorldRenderer;
 
 @RegisterSystem(RegisterMode.CLIENT)
 public class FirstPersonClientSystem extends BaseComponentSystem implements UpdateSubscriberSystem {
+
     private static final int USEANIMATIONLENGTH = 200;
 
     @In
@@ -54,8 +56,12 @@ public class FirstPersonClientSystem extends BaseComponentSystem implements Upda
     private Time time;
 
     private EntityRef handEntity;
-    private EntityRef currentHeldItem;
-    private boolean relinkHeldItem;
+
+    // the item from the inventory synchronized with the server
+    private EntityRef currentHeldItem = EntityRef.NULL;
+
+    // the client side entity representing the hold item
+    private EntityRef clientHeldItem = EntityRef.NULL;
 
     private EntityRef getHandEntity() {
         if (handEntity == null) {
@@ -67,6 +73,7 @@ public class FirstPersonClientSystem extends BaseComponentSystem implements Upda
         return handEntity;
     }
 
+    // ensures held item mount point entity exists, attaches it to the camera and sets its transform
     @ReceiveEvent
     public void ensureClientSideEntityOnHeldItemMountPoint(OnActivatedComponent event, EntityRef camera,
                                                            FirstPersonHeldItemMountPointComponent firstPersonHeldItemMountPointComponent) {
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Thanks for this! Pinging some potential reviewers to give this a deeper look than I can readily provide.

@GooeyHub: add to whitelist

GooeyHub:

engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
```diff
@@ -131,50 +138,61 @@ public void onHeldItemChanged(OnChangedComponent event, EntityRef character, Cha
...
                 }
 
                 // use the hand if there is no new item
-                EntityRef heldItem = newItem;
-                if (!heldItem.exists()) {
+                EntityRef heldItem;
+                if (!newItem.exists()) {
                     heldItem = getHandEntity();
+                } else {
+                    heldItem = newItem;
```
portokaliu: Was this really neccessary?

Tropid: I think so, it references the currently held item which is copied in line 179 to be used as a client-side model. Also in line 188 the `ItemIsHeldComponent` is added to mark the inventory item as being equipped.

The basic idea is to not change any of the previous inventory logic, but create a new client-side entity for the equipped items. So there are two entities, the "inventory entity" (which was replicated to the server and thus shown at 0, 0, 0)  and the new "client-side entity" which is only shown to the current client.
So this code determines the currently held item (hands or inventory item) which is afterwards cloned to the client. This may not be necessary if the hands are equipped, but this way it is consistent and simpler.

portokaliu: agh, it didn't select the amount of lines I tohught it would. Changing the few lines after the comment `// use the hand if there is no new item` seems to just do the same thing as before, just structured in a different way.

Tropid: Oh, I see. Of course, you're right. I guess I thought it was easier to read that way because you immediately see both possible options for the reference instead of setting it to one and then possibly overwriting it.
I can change it back if you'd like me to.

engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
```diff
@@ -37,11 +36,14 @@
...
 import org.terasology.network.ClientComponent;
+import org.terasology.network.NetworkComponent;
+import org.terasology.physics.components.RigidBodyComponent;
 import org.terasology.registry.In;
 import org.terasology.rendering.world.WorldRenderer;
 
 @RegisterSystem(RegisterMode.CLIENT)
 public class FirstPersonClientSystem extends BaseComponentSystem implements UpdateSubscriberSystem {
+
     private static final int USEANIMATIONLENGTH = 200;
```
skaldarnar: Slightly unrelated: Don't we use a naming scheme with underscores for static members, i.e., `USE_ANIMATION_LENGTH` or `USE_ANIMATION_DURATION`? Easy change alongside other changes in this file.

skaldarnar:
Overall, the changes look good to me. If anyone is sure enough that this won't break or complicate things pleas just approve this PR ;-)

GooeyHub:

portokaliu:
The only reason I'm hesitating to accept the changes is, as we spoke a while ago, that we may want to have the current held item visible to other players (in combat for example I'd like to know if the enemy was blocking on the general stance of the player (defending/attacking) or if he has no weapon/shield out, case in which I'd attack). For now we could integrate the changes if only to test the proof of concept for the fix. but I'd rather get the other idea going. Your move @Cervator :)

Cervator:
Alright, merged at long last :-)

Thanks for sticking with it @Tropid and thanks for testing & review @portokaliu & @skaldarnar !

PR #2820, opened 2017-03-11
Fix headless server build

Calinou:
This pull request fixes the server build (the regression was likely caused by https://github.com/MovingBlocks/Terasology/pull/2803). It does not seem to affect the client noticeably (tested with a local client and server).

## How to test

`./gradlew jar server` used to cause a build error, now it works.

Pinging @fabriond about this, since this fix is a bit hacky.


GooeyHub:

emanuele3d:
I guess that's the way to go short of having a good rethink of headless. 

Green light from me @Cervator. Thanks @Calinou!

Cervator:
Thanks! Yeah there might ultimately be a cleaner solution but in that case we can go bigger about just what the headless needs to do in renderland anyway.

PR #2822, opened 2017-03-12
Removed CTS dependency. Issue - #1126. [WIP]

### LINKED ISSUE #1126: Reduce coupling between Input System and CameraTargetSystem
As per conversation in PR https://github.com/MovingBlocks/Terasology/pull/1094:

The Input System has a dependency on the _CameraTargetSystem_ (CTS), which crops up in places like _StateMainMenu.init(GameEngine)_, where the CTS is initialized and registered for the Input System to work correctly. While Immortius has features in mind requiring camera(s) in the Main Menu (i.e. to deal with 3d content - customizable avatar perhaps?), the Main Menu currently has no need for this. In any case the dependency should be reduced, to avoid having to deal with the CTS when there are no cameras involved.
##

zeokav:
### Contains
An attempt to fix #1126 . It's still slightly half baked - needs some more refinements before being merged though. 

### How to test
i) Run Terasology.
ii) Confirm that CTS is not required anymore on the main menu and is only initialized before the game starts (when it's really required).

**_Don't merge yet_**
There's still quite a few things pending. I'm pushing this branch anyway, for a review and some more pointers and tips. 
- [x] CTS has been removed as a dependency from the menu
- [ ] Keyboard bindings must be added to regain control of in-menu keyboard shortcuts (NUI editor, Console, etc.)
- [ ] Refactor some more code which uses InputSystem arguments.
- [ ] Figure out how to deal with the NPE thrown by isCaptureMouse() (called by NUIManagerInternal). [Line 91 of MenuInputSystem.java]
- [ ] Figure out why the CrashReporter crashes the game. (Perhaps related to the NUIManagerInternal issue above?)

Tagging @emanuele3d and @Cervator for a review. Is there anything else apart from all that in the checklist that I need to consider? 😃

### DIFF
--- a/engine/src/main/java/org/terasology/input/InputSystem.java
+++ b/engine/src/main/java/org/terasology/input/InputSystem.java
@@ -284,7 +284,7 @@ public class InputSystem extends BaseComponentSystem {
     }
 
     public boolean isCapturingMouse() {
-        return capturingMouse && display.hasFocus();
+        return true;
     }
 
     public void setCapturingMouse(boolean capturingMouse) {
--- a/engine/src/main/java/org/terasology/input/MenuInputSystem.java
+++ b/engine/src/main/java/org/terasology/input/MenuInputSystem.java
@@ -1,0 +1,203 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.input;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.terasology.config.Config;
+import org.terasology.engine.Time;
+import org.terasology.engine.module.ModuleManager;
+import org.terasology.engine.subsystem.DisplayDevice;
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.systems.BaseComponentSystem;
+import org.terasology.input.device.MouseAction;
+import org.terasology.input.device.MouseDevice;
+import org.terasology.input.device.nulldevices.NullMouseDevice;
+import org.terasology.input.events.InputEvent;
+import org.terasology.input.events.LeftMouseDownButtonEvent;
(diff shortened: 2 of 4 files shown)


GooeyHub:

Cervator:
@GooeyHub: ok to test

@zeokav: thanks for the PR and I think you got a good checklist. Just in case you might want to run with multiplayer (server+client and a 2nd client and/or headless + client) just to make sure there's no difference in behavior.

Will try to test it some when able to find some time :-)

GooeyHub:

engine/src/main/java/org/terasology/input/MenuInputSystem.java
```diff
@@ -0,0 +1,203 @@
...
+import org.terasology.input.events.MouseAxisEvent;
+import org.terasology.input.events.MouseButtonEvent;
+import org.terasology.input.events.MouseDownButtonEvent;
+import org.terasology.input.events.MouseUpButtonEvent;
+import org.terasology.input.events.MouseWheelEvent;
+import org.terasology.input.events.MouseXAxisEvent;
+import org.terasology.input.events.MouseYAxisEvent;
+import org.terasology.input.events.RightMouseDownButtonEvent;
+import org.terasology.input.events.RightMouseUpButtonEvent;
+import org.terasology.input.internal.BindableButtonImpl;
```
Cervator: Looks like this import is unused - that's the only thing that sticks out though :-)

engine/src/main/java/org/terasology/input/InputSystem.java
```diff
@@ -284,7 +284,7 @@ public void update(float delta) {
     }
 
     public boolean isCapturingMouse() {
-        return capturingMouse && display.hasFocus();
+        return true;
```
Cervator: Although maybe I could also point this out. It appears to have the side effect that if you de-focus the window you can move the mouse cursor over it (still without giving the window focus) and the game will react. Previously the game entirely ignored input while not having focus. Dunno if we'd want that change?

zeokav: Oh this I had to change for the 4th checklist item. It's temporary. The display.hasFocus() method is dependent on the old input system. I'll change it back as soon as I figure out how to change that bit and not have any side effects on the game. :)

engine/src/main/java/org/terasology/input/MenuInputSystem.java
zeokav: My bad! I forgot to remove those.
The difficult part now is to get the keyboard input on the menu. The key bindings class is slightly (quite) dependent on the original input system class. I was playing around with that, hence all the imports. 
Will remove.

Cervator:
Poking @zeokav to see if more work is expected on this sometime :-)

zeokav:
@Cervator I've been pretty caught up with my internship and placements of late! But yes, I will get back to it once I get the time. 😃

zeokav:
@Cervator Free again! 😄 I'm sure that the code has gone through some changes over the summer. I think I'll close this branch and start fresh. Will save me from conflicts.

PR #2823, opened 2017-03-12
Cancel velocity on teleport

skaldarnar:
This is a small fix up for discussion. Currently, the teleport command will maintain the player's velocity. This fix will cancel the current velocity to zero.

On the one hand, maintaining velocity might be desired for a portal, where the player walks in and out. However, in the current implementation teleporting is a "normal" movement action with a (possible) huge delta in position. Thus, it is hard for system acting on player movement to identify a teleport.  

See https://github.com/Terasology/LightAndShadow/pull/52

I don't know if we could/should link the movement to a TeleportEvent of sorts so that systems can correctly identify the source of movement.

Discussion welcome: @krishnarb3 @msteiger

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/DeferredMomentumComponent.java
+++ b/engine/src/main/java/org/terasology/logic/characters/DeferredMomentumComponent.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2018 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.logic.characters;
+
+import org.terasology.entitySystem.Component;
+import org.terasology.math.geom.Vector3f;
+
+public final class DeferredMomentumComponent implements Component {
+    private Vector3f velocity = new Vector3f();
+
+    public Vector3f getVelocity() {
+        return velocity;
+    }
+
+    public void setVelocity(Vector3f newVelocity) {
+        velocity.set(newVelocity);
+    }
+}
--- a/engine/src/main/java/org/terasology/logic/characters/ServerCharacterPredictionSystem.java
+++ b/engine/src/main/java/org/terasology/logic/characters/ServerCharacterPredictionSystem.java
@@ -147,7 +147,7 @@ public class ServerCharacterPredictionSystem extends BaseComponentSystem impleme
         // stores the old character momentum so that it might be re-applied after the teleport event
         DeferredMomentumComponent deferredMomentum = new DeferredMomentumComponent();
         deferredMomentum.setVelocity(lastState.getVelocity());
-        entity.saveComponent(new DeferredMomentumComponent());
+        entity.saveComponent(deferredMomentum);
 
         stateBuffer.add(newState);
         characterMovementSystemUtility.setToState(entity, newState);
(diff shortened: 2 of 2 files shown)


GooeyHub:

krishnarb3:
This definitely helps with handling teleportation (Especially with magic dome, I'll test it out once), but I have a concern about teleport implementations in the future (Portals for example as you mentioned). 
I strongly feel we should have a TeleportEvent.
I'll try to create one if you guys are fine with it.
Discussions welcome ☺️

skaldarnar:
@krishnarb3 there is `CharacterTeleportEvent` - the question is how to handle the position change.

krishnarb3:
Sorry, my bad.
Is there a way to check if the player is currently undergoing the event?

skaldarnar:
I think that's the problem. The event is not tied to the entity, thus we cannot know for sure what caused movement. 

In general, the event can also be listened for in the dome system which would allow us to manually keep track of a specific entity. However, I don't think this should be the solution. There could be other cases where movement is not the right thing to express what is going on.

0shine0:
Can't we set the position of the teleported to object to the given position. No need for any movement action and if the portal is an entity than we can introduce a new component say RigidPortal that if present on the given portal entity will set the velocity to zero and change position and if not present than will just change position to specified value.

skaldarnar:
@0shine0 Just changing position data will interfere with the character prediction system, that's the reason why teleport is handled as movement.

krishnarb3:
@skaldarnar I am a bit confused, if we link the movement to a TeleportEvent how do we find if a character is teleporting?

skaldarnar:
That's the point: we currently can't 😞 I'm not sure what a good design for this case would be - that's what this discussion is for.

Cervator:
Not that my comment here will help add clarity, but I just wanted to mention that it would be sweet to have momentum be *configurable* - so you could have velocity cancelling, maintaining, or even modifying teleportation. Good old Portal had some fun puzzles :D

Pretty minimal PR so I guess we could just merge it in a day or two unless discussion suggests otherwise?

skaldarnar:
I'd like to wait with merging until we tried to use this for LaS dome. Might be that it won't help and we'll need to find something else.

I think the general problem is that teleportation _should not_ be general movement, but somehow different. Or we have to provide means to track movement (by player input) in order to only react on those events.

Cervator:
Bump! Digging around and noticed this plus PR #2849 and was wondering if we should merge this yet.

Also had an idea about teleporting and momentum: what if we negate all momentum as this PR suggests but then attach a `DeferredMomentumComponent` to the teleported target that triggers a few moments later and applies a suitable impulse to get the target moving again? That would get around character prediction, right?

That should give us a lot of flexibility as well. *Should* momentum for a particular teleport be resumed? Should it be enhanced or diminished? How long should the "teleportation freeze" last? And so on. It seems reasonable to accept as gameplay logic that a teleportation effect can have a momentary freeze. Could even make it longer on purpose if the player is going to a new place that needs to load a bunch of chunks. Cue a paired cut scene and momentary invincibility/invisibility/ghost mode so they player doesn't get eaten while waiting for the game to catch up :D

GooeyHub:

GooeyHub:

Cervator:
Bump - I see new commits on this @skaldarnar - are you happy with it / is there related module work somewhere meant to put the deferred momentum to use? :-)

skaldarnar:
No related work in any module (yet). Also, I'm not really happy as it is right now - Instead of the `DeferredMomentumComponent` why are we not sending a `DeferredMomentumEvent`? If it is a consumable event, the only code required in L&S should be to cancel the event to stop the movement after teleport, right?

Cervator:
On `DeferredMomentumComponent` vs `DeferredMomentumEvent` ... good question. It might be one of those *shrug* moments about which option you go for :-)

It is a subtle difference and I don't know if it is worth it or a good practice. Might be. Thinking out loud:

_Component scenario_: "Hey, this player is teleporting. Standard procedure is to freeze during teleport then use the deferred momentum component to *resume* the player's movement. If you want to change that please intercept the teleport and use the component's data as you see fit"

**Response in L&S**: Intercept teleport (which?), remove the deferred momentum from the player before it is used

_Event scenario_: "Hey, this player is teleporting. Standard procedure is to freeze during teleport then use the deferred momentum event to *match* the player's intended movement. You can cancel that event to avoid the movement or otherwise examine it for its data"

**Response in L&S**: Intercept ... deferred momentum event? Or first one then the other? At some point cancel the deferred momentum event, but be sure it is the right one somehow

Pretty similar scenarios, but maybe the responses show it. One advantage to the component scenario IMHO: you *know* you're reacting to the teleport event. With the deferred momentum event you could be reacting to *any* deferred momentum event since the two events are separate. You might have to connect them in some way (field in the deferred momentum event to indicate what triggered it?) to validate you're only cancelling the teleport's momentum.

However, even if you just worry about the teleport event you might only want to cancel momentum on *some* teleports ... Damned if you do, damned if you don't :-)

At first I thought maybe a convention could be that a component would be used if the desire is to modify what came first while an event would be to extend or cause a consequence but after the termination of what came first. But now I don't know.

jellysnake:
I am going to close this due it it's age & the lack of discussion. If people are interested in continuing this then we can re-open or move to an issue

Issue #2614, opened 2016-11-20
Game Save System Request


Feralmouse:
Technically not a bug, but it'd probably a) help with replicating bugs and b) prevent people from rage quitting every time some bug destroys their world. 

I propose that, instead of automatically overwriting the autosave, that after x time, such a save is copy archived into the users local save directory (wherever that is). Allowing multiple saves to exist. Such that, in the event of a bug happening due to using shapes...or perhaps a chunk exploding causing a permanent dead chunk, a player would be able to "restore" their world to an earlier point by reloading an old save file. 

Sure they'd lose fifteen minutes work, but  it's better than having to start all over again. Further, the user could attempt to recreate the bug in that world and if they can, then the save file could be useful evidence for a bug report as well and potential testing for developers.

Cervator:
Linking this to a comment @nihal111 made over at https://github.com/MovingBlocks/Terasology/pull/2550#issuecomment-283933830 - we still need to make this more visible, harden it further, and support multiple saves if desired yep :-)

PR #2829, opened 2017-03-19
Quick save and load hotkeys - addresses #2614

danosaur98:
### Contains
Addresses #2614. Creates two hotkeys: quick save (defaulted as the "[" key) and quick load (defaulted as "]" key). They can be changed in the Input Settings menu under "General". When quick save is pressed, it creates a quick save folder within the original save folder. When quick load is pressed, it loads the game found in this quick save folder. Autosave still updates the original save folder. 

### How to test
Enter any game. Press quick save button (creates save game with your items still in your inventory). Drop items.  Exit to main menu. Reload game, and items are still dropped.  Press quick load button, and you'll load the game with your items still in you inventory!

- [ ] Will need translation work for new Quick Save and Quick Load options in settings menu

### Thoughts for the future
- [ ] I had to create new constructors for some classes. Maybe restructure so that they continue to have only one constructor.
- [ ] Have a gamescreen that asks the user if they're sure they want to load the new game (in case they accidentally press the quick load button)
- [ ] Multiple game states instead of only one. Maybe have a gamescreen to navigate through these save paths.


GooeyHub:

engine/src/main/java/org/terasology/engine/modes/GameState.java
```diff
@@ -33,6 +33,8 @@
 
     void dispose();
 
+    void dispose(boolean saveRequested);
```
flo: SAme as with changeState: Something like a save system should not have a impact on those generic interfaces.

engine/src/main/java/org/terasology/input/binds/general/LoadButton.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.terasology.input.binds.general;
+
+import org.terasology.input.BindButtonEvent;
+import org.terasology.input.DefaultBinding;
+import org.terasology.input.InputType;
+import org.terasology.input.Keyboard;
+import org.terasology.input.RegisterBindButton;
+
+@RegisterBindButton(id = "load", description = "${engine:menu#binding-load}", category = "general")
```
flo: I think a quick load button is something dangerous to have. One might hit it accidentally why just trying out what keys do what.

engine/src/main/java/org/terasology/logic/players/MenuControlSystem.java
```diff
@@ -64,6 +85,31 @@ public void onScreenshotCapture(ScreenshotButton event, EntityRef entity) {
...
+    }
+
+    @ReceiveEvent(components = {ClientComponent.class})
+    public void onLoad(LoadButton event, EntityRef entity) {
+        if (event.getState() == ButtonState.DOWN) {
+            GameManifest quickSaveGameManifest = getQuickSaveGameManifest();
+            if (quickSaveGameManifest != null) {
+                boolean saveRequested = false; //quick load shouldn't cause the game to save
+                boolean isQuickLoad = true;
+                CoreRegistry.get(GameEngine.class).changeState(new StateLoading(quickSaveGameManifest, NetworkMode.NONE, isQuickLoad), saveRequested);
```
flo: CoreRegistry is deprecated. Use an `@In` annoated field of the wanted type instead. Alternativly you can if available use a Context object instance to get a value.

engine/src/main/java/org/terasology/persistence/internal/SaveTransaction.java
```diff
@@ -138,6 +144,8 @@ public void run() {
             mergeChanges();
             result = SaveTransactionResult.createSuccessResult();
             logger.info("Save game finished");
+            GameManifest latestManifest = getLatestGameManifest();
+            storageManager.setSavePath(PathManager.getInstance().getSavePath(latestManifest.getTitle()));
```
flo: Not sure if:
* It is a good idea to have 2 classes use each (SaveTransaction and storageManger) other. Maybe it would be cleaner if for example the storageManger would update it's save path itself.
* If changing the save path is design wise the correct approach. Maybe StorageManager could have a `saveAs` method that takes a path and does a regular save and then copy the save to that location.
* That changing the save path will create a proper copy as far as I can remember the save system saves only the chunks that have changed.

dannyzhou98:
Hi! Thank you for taking the time to review my pull request. 

I will change the code as requested later, but I have some questions about the quick save and load itself. 
I don't entirely understand what you mean by the world not being saved properly. When I was testing, it seemed like the entire world was properly saved. Do you mind testing the PR yourself and telling me how the world is incorrectly saved?

>  #2614 is about the problem that the automatic save might actually save a state that is corrupted and that you then have no way of getting back. 

My pull request does provide a solution. If you quick save before your game corrupts, you can "restore" your world by quick loading this save state. Auto save has no effect on the quick save. In fact, after you load your quick save, auto save will continue overwriting your original save state. This implies that if your game corrupts, but you quick saved beforehand, you can quick load, and auto save will overwrite your corrupted game. I hope this makes more sense to you after you test the PR and view the save transactions occurring in the save folders.

engine/src/main/java/org/terasology/input/binds/general/LoadButton.java
dannyzhou98: I thought it would be more convenient for the user. To safeguard against this, perhaps it should lead to a screen asking for confirmation from the user if they truly want to load their quick save state?

flo:
It can't by theory not work but I tested it anyway. 

Try this:
1. Start a fresh world and place some blocks at spawn
2. Teleport to 10000 50 0 via the command "teleport 10000 50 0"
3. Wait about 3 min so that the auto save runs and all spawn point chunks got properly unloaded
4. Quick save
5. Quick load
I tested it and first it seems to be saving it, but I am not sure if it loads the correct save. Try this:
6. Teleport to spawn (teleport 0 50 0)
7. Check if the things you made there are still there
8. Compare what chunk files got saved

Expected results:
* Chunk files equal
* At the start location there are things you build there even when those chunks were not loaded when you did the quick save

Actual result during my test:
* The quick save does not contain the chunks at origin
* The quick load somehow still managed to restore the blocks at spawn

About "saving before the world is corrupt": Normally you don't know when a world gets corrupted, so you would never know when to save. You can't even periodically do manual quick saves as you don't know if a world got corrupted already without trying to load it first.  

In  #2614 two aspects where mentioned: 1. helping with debugging those issues. 2. preventing players from loosing their work. Point 1 is not so important in my opinion and you also don't know if quick loads actually will help you to debug that issue as the bug might actually be in the saving system and your quick save might do it differntly or the bug might be related to multiplayer and your quick save is probably not multiplayer ready yet. Point 2 is a something I can understand that some people want something for it, but 99% of the bugs cause just crashes and not corruption and thus the auto save is enough. There are special action that might trigger world corruption but i guess once we get more testers there will be a stable and not stable version and hopefully those bugs get found and fixed in the unstable version.

So personally I think that actually this bug is not even that important to fix. I just gave you a review since I made the autosave system (and at that point avoiding the loss of work was exactly the reason why I made it).

skaldarnar:
Closing as stale PR. 

If you made progress or want to take this up again with the provided feedback please feel free to re-open.

PR #2830, opened 2017-03-19
Fix dvorak command

Vizaxo:
Dvorak command had left and right inverted. This simple fix swaps them back to be the correct way around.

### Contains

Fix dvorak command (left and right were inverted)

### How to test

Set keyboard to be a dvorak layout. Open up a game world, and run the dvorak command. The movement should have the same keys as a normal qwerty keyboard.


GooeyHub:

xrtariq2594:
From testing, this works as intended. Good work.

Merging in.

Issue #2833, opened 2017-03-19
Various isses with movement inputs persisting through the chatbox


Vizaxo:
There are various issues concerning movement inputs persisting through the chatbox:

1) If you hold down a movement key, then press 't' to open the chatbox, the movement will persist until you let go of it (expected: movement will stop moving the player upon opening the chatbox).
2) If you hold down two movement keys in different planes (i.e. left or right, and forward or backward), they will both persist when the chatbox is opened. Letting go of one of them will stop movement in that direction, and then letting go of the other will stop movement completely.
3) Holding down up and down at the same time will not allow you to open the chatbox.
4) Holding down left and right at the same time will work the same as (2), but with you standing still at the start, then starting to move when you let go of one direction.
5) Pressing 'r' to automove will keep you moving after opening the chatbox, but will be turned off when the key bound to forward is pressed (expected: either automove will stop upon opening the chatbox, or it will continue until you exit the chatbox and stop it manually).
6) Holding shift and opening the chatbox (whether moving at the time or not), then releasing shift and closing the chatbox will make your character move slowly (as if shift was held) until you press and release shift again.

### Log details and game version
Game version: tested both on the latest source tree (commit 7e3ba2c), and the Omega 691 binaries from http://forum.terasology.org/threads/play-test-session-sunday-march-19th-1-pm-us-eastern-time.1814/.

### Computer details
GNU/Linux Ubuntu 16.10

Cervator:
One additional (if barely different) case: immediately after opening the chat box with `t` (or the console) even after having let go of said button the next button press like one of `wasd` will trigger movement. After letting go once everything is fine.

Pinging @ThompsonTyler due to extensive research on potentially adding modifier key support requested in #2551, also a includes a great summary of involved classes - https://github.com/MovingBlocks/Terasology/issues/2551#issuecomment-285990629 - curious if it rings a bell or additional feedback/ideas may be available :-)

ThompsonTyler:
Will take a look, see if I can see anything. (Sorry bout the absence lately, messed up my hand and gotta do everything 1 handed atm)

Was able to replicate all problems other than Cervators(Not sure if I am understand it wrong, but if I open chat box and press was or d I don't see any movement)

Also problem 3 I wasn't able to replicate (holding forward/back preventing chat from opening)

Vizaxo:
[...]

PR #2839, opened 2017-03-19
Fix first letter in chatbox moving player

Vizaxo:
Fix part of issue #2833 (Various isses with movement inputs persisting
through the chatbox) where the first letter typed into an empty chatbox was
causing the player to move.

This was caused by the event falling through to the normal event system,
because a path was not marked with eventHandled = true.

Note: there are still parts of #2833 that this doesn't fix, but might be related.

### Contains

Fixes part of #2833

### How to test

Open up a chatbox, making sure it is empty. Type a movement command (e.g. 'w'), and observe what happens.

Before the patch, the character typed would go into the chatbox, but the player would move, even though the chat box is open. After the patch, the character typed just goes into the chatbox, and doesn't make the player move.


GooeyHub:

Issue #2835, opened 2017-03-19
Improve bulk and quantity-based giving of items (as opposed to blocks)


Cervator:
At present we've simplified down to a single `give` command where we used to have separate `giveBlock` and `giveItem` commands. They still exist as methods and the `give` command simply tries to deduce which one to use.

However, quantity based giving isn't working for items, even for those that are stackable. Try for instance `give axe 42` - the console insists that it gave you 42 items of `Core:axe` but you only get one item. Same thing if the item is actually stackable (`Core:axe` is not but `Core:door` is). Repeated commands with a stackable item will happily increase the stack by one each time while telling you that you were given x items.

We need:

* Improve the ability to *actually* grant you multiples of a stackable item
* Improve the logging for when you requested an unstackable item without a given quantity (default quantity is 16 for blocks, but for an unstackable item (or block like `core:chest`) it should be 1 and the console should indicate that)
* Allow quantity-based requests for unstackable items to put one item in each available inventory slot (up to the max inventory size)

Additionally we could also use a variant of `bulkGiveBlock` that works for items. For instance with NeoTTA enabled you get 29 item hits for `search seed` right now and it would be great to be able to `bulkGiveItem seed` to get one (or a stack) of each seed so you can get on with life already and plant all the things :-)

... although at that point do we fold the two into `bulkGive` and react intelligently to blocks vs items?

Finally it would be good to actually respond to the max inventory space when giving a large quantity of anything. See #2607 and its implementation in #2678 - that setup simply keeps going so if you ask for ten million of something it will literally loop through the entire range even after the inventory is full and couldn't possibly accept more items.

See `ItemCommands` and `BlockCommands` to get started and keep an eye out for `TODO` tags that relate.

Vizaxo:
Oops, sorry for spamming this thread with commits; I was rebasing my branch and didn't realise they showed up here.

Anyway, I think #2841 and #2842 cover everything in this. There are a couple of notes on #2841 (such as checking performance on servers), and I left bulkGiveBlock and bulkGiveItem seperate for now (instead of implementing bulkGive to do both).

xrtariq2594:
Good work @Vizaxo!

PR #2841, opened 2017-03-20
Fix quantity-based giving of items #2835

Vizaxo:
### Contains

Fix most of #2835 (everything apart from the bulkGiveItem command), including the lag on large numbers of items introduced in #2678.

### How to test

Run the command "give axe 1000", and observe that your inventory is filled with axes, and the command window reported the number of axes actually given.

Run the command "give dirt 10000000" and observe that the game no longer lags, but the command executes successfully, and reports the number of dirt blocks given.

This way of giving multiple items/blocks is a much more robust solution than the previous one, but it does it by giving them one at a time, rather than a stack at a time. I'm not sure if that's an issue, but it might be worthwhile to test giving a full inventory of items on a remote server, to make sure it's not too slow.

Also, the default number of items given is always 1; I'm not sure if you want this to be 16 as well. I tried to change it, and found about 4 different ways to get the stack size of an item, but they all returned the same for stackable and unstackable items, so I'm not sure how to determine whether or not an item is stackable.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
@@ -296,18 +296,20 @@ public class BlockCommands extends BaseComponentSystem {
             @CommandParam("blockName") String uri,
             @CommandParam(value = "quantity", required = false) Integer quantityParam,
             @CommandParam(value = "shapeName", required = false) String shapeUriParam) {
-        int quantity = quantityParam != null ? quantityParam : 16;
         Set<ResourceUrn> matchingUris = Assets.resolveAssetUri(uri, BlockFamilyDefinition.class);
+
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
@GooeyHub: ok to test

I dunno what to do about quantity. I guess the default being 1 could be perfectly fine and less confusing than the somewhat random 16.

If you're curious to get a better read on whether the singular item adder is worse off than before you could probably try wrapping each command with some timer logging then run them a bunch of times to get some stats. If it doesn't make a big difference it is probably fine, since console usage isn't really meant to be heavy anyway (there should be better ways to get blocks, like via gameplay or the BlockPicker)

You could try running two local clients to check out the same for both the host and non-host client.

GooeyHub:

PR #2842, opened 2017-03-20
Add bulkGiveItem command (from #2835)

Vizaxo:
### Contains

Adds the bulkGiveItem command (the same as bulkGiveBlock, but for items) mentioned in #2835.

It also moves the listItem command into ItemCommands, so it can be used in other commands relating to items.

### How to test

Type the command "bulkGiveItem axe", and you will be given a copy of every item with "axe" in the name.

### Notes

This complements #2841, but they are completely separate, and work individually or together.

As this command uses give in its implementation, using bulkGiveItem with a quantity only works if both patches are applied together.

Also, I changed uriStartsWithAnyString to be a static method because it is a useful utility function, but it might be better suited to be in a separate file, as it is used in both BlockCommands and ItemCommands.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
@@ -22,7 +22,6 @@ import org.terasology.assets.ResourceUrn;
 import org.terasology.assets.management.AssetManager;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
-import org.terasology.entitySystem.prefab.Prefab;
 import org.terasology.entitySystem.prefab.PrefabManager;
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
@GooeyHub: ok to test

@Vizaxo: thanks for both PRs! Especially from them being separate since they could be, yay topic branches!

I have a vague recollection that @xrtariq2594 talked about maybe checking these two out, so pinging for great justice :D Even if he didn't well.. tag you're it!

GooeyHub:

Vizaxo:
@xrtariq2594 Great, thanks for reviewing and merging all of these :)

I've applied the change (and made bulkGiveBlock only print one message, rather than one message per block, if you give it a quantity of < 1). I can't update this PR now that it's merged, so it's in #2866.

Issue #2837, opened 2017-03-19
Add a simple command that makes the player immune to damage


Cervator:
More or less the traditional "god mode" - unless that term has become too loaded by now. What's it called in some other recent games? Maybe it should just be `damageImmunity` instead (keep reading, more further down)

We already have "flight" and "ghost" (flight + noclip) so this command should just focus on cancelling all damage affecting the player.

See `HealthCommands` where `softLanding` heavily restricts fall damage from a player (and is probably where the new command should go) or maybe the physical resist system in the AlterationEffects/Potions modules.

Alternatively for a nicer approach support including a damage type to make the player specifically immune to, rather than only support global damage immunity. That (along with a command to inflict typed damage - see #2809) would go a long way to make it easier to distinguish between and test out different kinds of and sources for damage.

So we could have a simple command for damage immunity against everything, with a variant taking a parameter for the damage type. Same for doing damage.

Interestingly enough the `hjump` command over in `MovementDebugCommands` once upon a time *did* include damage immunity from falling. It wasn't taken off with `restoreSpeed` so #1247 fixed that - and made me realize that isn't there anymore (fix was from late 2014). Huh!

So if interested in digging deeper that might be something additional to look at. We used to have damage immunity applied in some cases, and more damage *dealing* in others, like smashing into walls with `hspeed` on could actually kill you from "horizontal fall damage" - why'd we ever take that off .... :D

krishnasism:
Hi!
I have been able to implement the feature, but I am unable to do a pull request due to some network issues. Could you suggest an alternate way via which I can show you the feature? Thanks.

krishnasism:
#2845

portokaliu:
@krishnasism I don't thing that putting in something this specific would be wise. Why not just make it a general player attribute? This way any mod that wants to fiddle with the setting would just have that to hook on to. (like say, a feather fall potion - which btw, you might wanna check if it exists in the potions mod)

krishnasism:
@portokaliu I'm trying to do that. I'm currently understanding the code. I have worked on it for 2 hours or so. So I will just request some time from you :P yes thank you for your response. I will get back to work on it!

Weilin1992:
@portokaliu what do you think of my implementation?

portokaliu:
@Weilin1992 I did read your version at about the same time. I have my doubts on the whole "new component" idea, since adding more components wouldn't be the best idea imo. 

[...]

PR #2843, opened 2017-03-20
Add noDamage Command

Weilin1992:
### Contains
add god command to immune to all damage mentioned in #2837

### How to test

Type the command "noDamage", and you will be immutable to all damage. Type "GodMode" again to remove God mode, need cheat permission

### Notes
This is a simple command make player immutable to all types of damage. Later I will submit another PR to add another command to make player immutable to some particular type of damage

### DIFF
--- a/modules/Core/src/main/java/org/terasology/logic/health/HealthComponent.java
+++ b/modules/Core/src/main/java/org/terasology/logic/health/HealthComponent.java
@@ -41,7 +41,6 @@ public final class HealthComponent implements Component {
     @Replicate
     public boolean immutable = false;
 
-    @Replicate
     @TextField
     public int currentHealth = 20;
 
--- a/modules/Core/src/main/java/org/terasology/logic/health/HealthComponent.java
+++ b/modules/Core/src/main/java/org/terasology/logic/health/HealthComponent.java
@@ -39,6 +39,9 @@ public final class HealthComponent implements Component {
     public float excessSpeedDamageMultiplier = 10f;
 
     @Replicate
+    public boolean immutable = false;
+
+    @Replicate
     @TextField
     public int currentHealth = 20;
 
--- a/modules/Core/src/main/java/org/terasology/logic/health/HealthComponent.java
+++ b/modules/Core/src/main/java/org/terasology/logic/health/HealthComponent.java
@@ -21,7 +21,7 @@ import org.terasology.rendering.nui.properties.TextField;
 
 /**
  */
-public final class HealthComponent implements Component {
+public class HealthComponent implements Component {
     // Configuration options
     @Replicate
     public int maxHealth = 20;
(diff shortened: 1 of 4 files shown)


GooeyHub:

OvermindDL1:
`nodamage` or `noDamage` might be a better command name?

Also instead of adding a new field to the health component, making it larger, what about some kind of event hook to cancel damage applied to the health component, or better yet, remove the health component entirely?

Weilin1992:
@OvermindDL1 I don't think remove health component is better, since we will lost some information, such as current health, current max health. If the component can be disable like unity3d, that will be a good choice.But If we want to add a disable property it will make all component larger.

I have another idea, how about add a noDamage component. check if the entity have this noDamage component, we can use a command to remove or add this component. This will not make health component  larger.Further we can develop a buff system based on such component.

OvermindDL1:
> @OvermindDL1 I don't think remove health component is better, since we will lost some information, such as current health, current max health. If the component can be disable like unity3d, that will be a good choice.But If we want to add a disable property it will make all component larger.

You can always make a new 'DisabledHealthComponent' just to store the information for recovery later.

> I have another idea, how about add a noDamage component. check if the entity have this noDamage component, we can use a command to remove or add this component. This will not make health component larger.Further we can develop a buff system based on such component.

This could work, as long as all the systems that affect the HealthComponent checked it too, but two issues I see is lots of 'if's around for that and not everything will check it, hence why I think it is better to just remove the health component at all, can store the information in a dummy component instead if it is really wanted to be stored (not all situations would need that, though would be useful for this specific command).

Weilin1992:
@OvermindDL1 I change the code as you suggested. remove HealthComponent and store the information to DisabledHealthComponent.

OvermindDL1:
I'm not the Authors, just someone that programs on Terasology at times so I know how I would prefer to use these changes.  ^.^

It looks good though, should be quite efficient, reducing the number of checks, and can be used for more than just players with ease.

And idea though, I wonder if the disabled component could just subclass the normal health component or if the internal ECS would identify them the same...  If this were a static language I'd say just make the healthcomponent an internal field (not a pointer), and that might still be useful here, hmm...

I wonder if it is worth making a `DisabledComponentsComponent`, that is just a list of components that are 'disabled' but you want to hold on to later.  This would allow it to be far more generic and useful for any component where you want to keep the state when you put it back on but still want it 'gone' for now...  Ideas anyone?

modules/Core/src/main/java/org/terasology/logic/health/HealthComponent.java
```diff
@@ -21,7 +21,7 @@
 
 /**
  */
-public final class HealthComponent implements Component {
```
OvermindDL1: Hmm, making this non-final could have a performance issue...  Java needs a way to use the same memory as a different type as is easy in C++, or to not use individual components, hmm...

oniatus:
What's the state of this PR?

Most recent discussion about disabling components would be #2896 which seems to favor a separate component to mark an entity as immune.

Weilin1992:
@oniatus I will work on this issue #2896 recently. Maybe this pr can be closed because i have to rewrite my code

Cervator:
Good to hear from you @Weilin1992 !

Yep probably easier to rewrite. Will close this. Hoping to see more from you in the future :-)

Issue #2744, opened 2017-01-07
All graphics become white when switching from fullscreen to windowed


Metatricity:
### What you were trying to do

Play on the Terasology event server

### What actually happened

I played for a little bit and looked around and switched from windowed to fullscreen. Then I realized that I liked it better windowed, on the computer I was on, and so I switched it back. When I clicked BACK, everything was white.

![image](https://cloud.githubusercontent.com/assets/16510596/21745691/f85b3a42-d4fe-11e6-823c-b3b339ff510f.png)

### How to reproduce

* Step 1
Join Terasology Event Server
* Step 2
Set display to Fullscreen
* Step 3
Set display to Windowed

### Log details and game version

**Game**:
https://gist.githubusercontent.com/Metatricity/3b7f8c6d1b008ca270d8a37b36c39de4/raw/706787d3c63ca6b42f74e675e9e580a4c5fd9d33/TerasologyLog.txt
**Init:**
https://gist.github.com/Metatricity/44e1f5403f57e4fe5f1ef43468cdfd91

### Computer details

Windows 7
Dell Latitude 3550, x64-bit
Intel Core i3-5005U @ 2.00GHZ, 2000 Mhz, 2 Cores, 4 Logical Processors
Intel HD Graphics 5500
7.64 GB of Physical Memory
15.3 GB of Virtual Memory

Metatricity:
This might just be for my model of graphics card, but it is stil worthy of an issue :-1:

iojw:
I believe that this issue related to display resizing and rendering is known and has already been mentioned a few times. 🙂 

See: #2528, #2659 and #2591

emanuele3d:
Just wanting to confirm we are aware of the problem. 

Technically speaking this is most likely due to changes @tdgunes and I have done to the renderer, in particular in the area of FrameBufferObjects management. When resizing occurs the FBOs are regenerated, their id changes but most of the nodes in the renderer seems to be unaware of the new id and keep using the old FBOs which are no longer valid.

I can't tackle this just yet as I have some intermediate goals before I can take care of it, one of which is to simplify FBO management. But it will be taken care of. Soonish I hope.

Metatricity:
Thank you for looking into this :)

Metatricity:
@iojw However, there is no known record of there being a white screen rendered, there is only a black screen, or a water render change. #2528 and #2659

iojw:
[...]

Issue #2659, opened 2016-12-11
Black screen on changing from windowed mode to fullscreen when playing


zeokav:
### What you were trying to do
Change a few video settings when the game was running. Switching from windowed to full screen causes the game to display a black screen with nothing else on it.
### What actually happened
As soon as I switch to full screen mode, everything disappears. However, the game doesn't really crash - it's just a black screen. I can still see the game being saved. I am attaching a copy of the logs. Search for '!!!' in Terasology-Game_2 to jump directly to the part where the black screen issue happened.

### How to reproduce

* Step 1 Launch the game in windowed mode, and start playing (load or create a new game)
* Step 2 Open video settings and change from windowed to full screen
* Step 3 The darkness.

### Log details and game version
Version - built from the latest sources.
Logs - 
[2016-12-11_10-28-22.zip](https://github.com/MovingBlocks/Terasology/files/644241/2016-12-11_10-28-22.zip)

### Computer details
8 GB RAM, 
i7-6700HQ 2.6 GHz
Nvidia GTX 960m 4GB VRAM
Windows 10

Edit: Could not get a screenshot of the black screen because it doesn't work for some reason. (But trust me, it's just a black screen with nothing on it.)

emanuele3d:
Thank you for the report. We are aware of issues related to any kind of display-resizing, i.e. from full screen to window, viceversa and even just interactively resizing the window. I.e. see #2528. At this stage I believe this problem is connected to those issues. As soon as I have the time this will get fixed.

Meanwhile, the simple workaround is to reload the game or, in the worst case scenario, restart Terasology. 

Apologies for the inconvenience.

zeokav:
@emanuele3d Let me know if you need some minor assistance, I'll be happy to help. 😃

emanuele3d:
@zeokav: thank you for the offer. Unfortunately this problem has to do with FrameBufferObjects (FBOs) and their regeneration anytime the display is resized. This makes this issue a moderately difficult one to understand even though it might not be overly difficult to fix: it's probably a hard to find/(potentially) easy to fix kind of thing. So, I wouldn't consider it "minor". But of course you'd be welcome to try. =)

You'd have to look into the classes BindFBO, SetViewportToSizeOf and probably also in DisplayResolutionDependentFBOs and DefaultDynamicFBOs. I suspect the problem is somewhere there.

zeokav:
[...]

Issue #2591, opened 2016-11-05
Resizing not working properly


Halamix2:
### What you were trying to do
Resize window

### What actually happened
window resized but not the "rendering area" (something like #2490 but for blocks)
Skybox and water are rendered as expected
outline is visible on line between rendered zone and empty one (when outline is enabled in options)

### How to reproduce

* Step 1 turn on outilne in graphic settings
* Step 2 resize window before joining server or opening local game (after loading world it won't work)
* Step 3 see weird things going on

###More sophisticated explanation

After starting Terasology, entering game and resizing this bug occurs. After quitting game without quitting Terasology, entering game again and resizing bug #2490 occurs but not this one, and bug 2490 stays on the screen even after quitting game and resizing window again (skybox rendering width stays the same all the time when bug #2490 occurs)

### Log details and game version  

version 1 836 alpha 2016-11-05T03:55:30Z
[Terasology-.log.txt](https://github.com/MovingBlocks/Terasology/files/573349/Terasology-.log.txt)
[Terasology-init.log.txt](https://github.com/MovingBlocks/Terasology/files/573350/Terasology-init.log.txt)
[Terasology-menu.log.txt](https://github.com/MovingBlocks/Terasology/files/573348/Terasology-menu.log.txt)
[Terasology-mmo.to-188.68.52.110-25778.log.txt](https://github.com/MovingBlocks/Terasology/files/573351/Terasology-mmo.to-188.68.52.110-25778.log.txt)

### screens  
http://i.imgur.com/2KSFJex.png
http://i.imgur.com/NhzE3DE.png

### Computer details

OS: Ubuntu 16.10 yakkety
Kernel: x86_64 Linux 4.8.0-26-generic
Resolution: 1366x768
CPU: Intel Core i5-5200U CPU @ 2.7GHz
GPU: GeForce 920M (Nvidia drivers 370.28)
RAM: 7886MiB
Java version: 1.8.0_111

emanuele3d:
Thank you @Halamix2 for reporting this. I managed to reproduce it. I'll see what I can do about it.

emanuele3d:
While I did manage to reproduce it, it took me a while to reproduce it reliably. For posterity, the steps I found most reliable are:

- Start Terasology in Windowed Mode with Outline OFF
- Go to the Video Settings and turn it ON
- Go to the Main Menu
- Resize the window to nearly full screen 
- Load  up a game (everything will be normal)
- Bring up the In-game menu
[...]

Issue #1747, opened 2015-05-27
Rendering Freezes/Breaks when Resizing Window when In-Game


prestidigitator:
Found in b2bcc42 on AMD FX-8350 with GeForce GTX 650 Ti Boost, running Arch Linux 64-bit (4.0.2-1-ARCH kernel with nvidia-340xx 340.86-9 driver) and OpenJDK 1.8.0_45.  Steps to reproduce:
1. Launch Terasology normally (gradlew runPC).
2. Create or load a single-player world and start game.
3. Maximize or unmaximize window.

Expected: window resizes and rendering continues as usual.

Observed: window resizes, but rendering becomes very odd and updates strangely or not at all.  Only crosshairs, HUD/menus, and possibly sun rays render normally.  Note smear of blue/white (sky?) in screenshot below.  It's the only thing that changed as I tried to look around or move.

Workaround: Return to menu and reload world.  Rendering resets to normal.
## Normal Rendering Prior to Resize

![terasologyresizerenderingbug_preresize](https://cloud.githubusercontent.com/assets/3741838/7831149/a447653e-0408-11e5-88c1-35e0d87fafb0.png)
## Broken Rendering after Resize

![terasologyresizerenderingbug_postresize](https://cloud.githubusercontent.com/assets/3741838/7831154/b4acf902-0408-11e5-967f-aff10458b519.png)
##

emanuele3d:
Uh-oh, that looks squarely up my alley.

I can reproduce the problem in develop. It appears some of the FrameBuffers are not being replaced with new ones matching the new size of the window. It doesn't surprise me as the resizing code, and especially its triggers, are a bit of a mess. And don't even think about taking screenshots with sizes not matching the current display! That might format your hard drive. Or worse!

Fear not however! I seem to have solved the problem in this [work branch](https://github.com/emanuele3d/Terasology/tree/IntroducingPostProcessorAndBuffersManager) of mine. Can you by any chance grab it and confirm? There still seems to be an issue as the image is slightly stretched on the Y axis after the resize/maximize/unmaximize, but the stretching disappears as soon as the player moves or rotates, which is perhaps something we can live with for the time being.

[...]

PR #2844, opened 2017-03-20
Resizing issues

vampcat:
- Fixes the resizing issues.
- Replaced all explicit bind calls with StateChanges
- Some code refactoring to make the code more consistent across files

### How to test
Run the game, try resizing the game in windowed mode or going from fullscreen to windowed or vice versa (I've noticed an issue with the game sometimes not being able to go from windowed to fullscreen, but that's an issue outside the scope of this PR). Everything else related to rendering should be same (no weird water bugs, no screenshot crashes, no new visual artifacts).

This PR fixes the following:
#2744 
#2659
#2591
#1747
And quite possibly a few others related to resizing.


vampcat:
Pinging @emanuele3d and @Cervator for review!

GooeyHub:

GooeyHub:

Cervator:
Awesome! This is way improved over the buggy previous state. I *did* still find ways to provoke the quirks, but it seems way harder, and may only happen after reloading an existing world?

I don't have exact reproducing steps but I would start the game then resize a bunch, toggle between screen modes (fullscreen etc), exit to main menu, load the game back up, then resize some more. One time after a reload the game looked fine but resizing would reproduce the old blackness in the resized areas. Two times *immediately* after reloading a world it came out immediately off ratio or something, like so:

2017-03-23 00_09_07-terasology - c__dev_terasology_git_terasology - intellij idea 2016 3 3

vampcat:
Okay, that's definitely not good. I'll be travelling through that part of the code for quite some time, so I'll keep an out for anything else that might cause these issues :D

engine/src/main/java/org/terasology/rendering/dag/nodes/FirstPersonViewNode.java
```diff
@@ -69,10 +71,13 @@ public void initialise() {
...
 
         addDesiredStateChange(new SetViewportToSizeOf(READONLY_GBUFFER, displayResolutionDependentFBOs));
-        sceneOpaqueFbo = displayResolutionDependentFBOs.get(READONLY_GBUFFER);
+        addDesiredStateChange(new BindFBO(READONLY_GBUFFER, displayResolutionDependentFBOs));
+        update();
 
         // this guarantee the objects drawn by this node are always drawn in front of everything else
         addDesiredStateChange(new SetDepthFunction(GL_ALWAYS));
+
+        displayResolutionDependentFBOs.subscribe(this);
```
emanuele3d: Let's move this up please.

engine/src/main/java/org/terasology/rendering/dag/nodes/FirstPersonViewNode.java
```diff
@@ -114,4 +117,9 @@ public void process() {
 
             PerformanceMonitor.endActivity();
     }
+
+    @Override
+    public void update() {
+        readOnlyGBufferFbo = displayResolutionDependentFBOs.get(READONLY_GBUFFER);
```
emanuele3d: I might be wrong but this looks like another case where we are not actually using this and readOnlyGBufferFbo is not necessary?

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
```diff
@@ -83,10 +81,13 @@ public void initialise() {
...
 
         addDesiredStateChange(new SetViewportToSizeOf(READONLY_GBUFFER, displayResolutionDependentFBOs));
-        sceneOpaqueFbo = displayResolutionDependentFBOs.get(READONLY_GBUFFER);
+        addDesiredStateChange(new BindFBO(READONLY_GBUFFER, displayResolutionDependentFBOs));
+        update();
 
         addDesiredStateChange(new EnableMaterial(CHUNK_SHADER.toString()));
         chunkShader = getMaterial(CHUNK_SHADER);
+
+        displayResolutionDependentFBOs.subscribe(this);
```
emanuele3d: Let's move this up please.

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
```diff
@@ -68,7 +66,7 @@
     private Material chunkShader;
     private SetWireframe wireframeStateChange;
     private RenderingDebugConfig renderingDebugConfig;
-    private FBO sceneOpaqueFbo;
+    private FBO readOnlyGBufferFbo;
```
emanuele3d: It appears this variable might be unnecessary in this node.

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueObjectsNode.java
```diff
@@ -57,7 +59,7 @@
     private Camera playerCamera;
     private SetWireframe wireframeStateChange;
     private RenderingDebugConfig renderingDebugConfig;
-    private FBO sceneOpaqueFbo;
+    private FBO readOnlyGBufferFbo;
```
emanuele3d: Same as above: it doesn't look like we are using this.

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueObjectsNode.java
```diff
@@ -72,7 +74,10 @@ public void initialise() {
         addDesiredStateChange(new LookThrough(playerCamera));
 
         addDesiredStateChange(new SetViewportToSizeOf(READONLY_GBUFFER, displayResolutionDependentFBOs));
-        sceneOpaqueFbo = displayResolutionDependentFBOs.get(READONLY_GBUFFER);
+        addDesiredStateChange(new BindFBO(READONLY_GBUFFER, displayResolutionDependentFBOs));
+        update();
+
+        displayResolutionDependentFBOs.subscribe(this);
```
emanuele3d: Let's move this up please.

engine/src/main/java/org/terasology/rendering/dag/nodes/OverlaysNode.java
```diff
@@ -59,7 +57,7 @@
 
     private Camera playerCamera;
     private SetWireframe wireframeStateChange;
-    private FBO sceneOpaqueFbo;
+    private FBO readOnlyGBuffer;
```
emanuele3d: Same as above: variable seemingly not in use.

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
```diff
@@ -81,10 +81,9 @@ public void initialise() {
         playerCamera = worldRenderer.getActiveCamera();
         addDesiredStateChange(new LookThrough(playerCamera));
 
-        requiresFBO(new FBOConfig(REFRACTIVE_REFLECTIVE, FULL_SCALE, FBO.Type.HDR).useNormalBuffer(), displayResolutionDependentFBOs);
+        refractiveReflectiveFbo = requiresFBO(new FBOConfig(REFRACTIVE_REFLECTIVE, FULL_SCALE, FBO.Type.HDR).useNormalBuffer(), displayResolutionDependentFBOs);
         addDesiredStateChange(new BindFBO(REFRACTIVE_REFLECTIVE, displayResolutionDependentFBOs));
-        sceneOpaqueFbo = displayResolutionDependentFBOs.get(READONLY_GBUFFER);
-        refractiveReflectiveFbo = displayResolutionDependentFBOs.get(REFRACTIVE_REFLECTIVE);
+        readOnlyGBufferFbo = displayResolutionDependentFBOs.get(READONLY_GBUFFER);
```
emanuele3d: Do we need this line here? readOnlyGBufferFbo seems to be used only in update(), where it is refetched.

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
```diff
@@ -81,10 +81,9 @@ public void initialise() {
...
 
         addDesiredStateChange(new EnableMaterial(CHUNK_SHADER.toString()));
         chunkShader = getMaterial(CHUNK_SHADER);
```
emanuele3d: Let's move the subscribe line below up, where the FBOs are.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
```diff
@@ -114,20 +112,13 @@ public StateChange getDefaultInstance() {
...
             switch (textureType) {
-
                 case ColorTexture:
                     return inputFbo.colorBufferTextureId;
 
-                case DepthRenderBuffer:
-                    return inputFbo.depthStencilRboId;
-
-                case DepthTexture:
+                case DepthStencilTexture:
```
emanuele3d: Why the removal of a case here?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
```diff
@@ -136,36 +127,29 @@ private int fetchTextureId() {
...
-
-        inputFbo = fbosManager.get(fboURN);
-
-        // If a node taking advantage of this state change is disabled when a game is started, task is null.
-        // This is due to the task list generator not having yet called the generateTask() method.
-        if (task != null) {
-            task.setTextureId(fetchTextureId());
-        }
+        inputFbo = fboManager.get(fboUrn);
+        task.setTextureId(fetchTextureId());
```
emanuele3d: Why the removal of the comment and the if block?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
```diff
@@ -136,36 +127,29 @@ private int fetchTextureId() {
...
     }
 
     @Override
     public String toString() {
         if (this != defaultInstance) {
             return String.format("%30s: slot %s, fbo %s, textureType %s, fboManager %s, material %s, parameter '%s'", this.getClass().getSimpleName(),
-                    textureSlot, fboURN.toString(), textureType.name(), fbosManager.toString(), materialURN.toString(), parameterName);
+                    textureSlot, fboUrn.toString(), textureType.name(), fboManager.toString(), materialURN.toString(), parameterName);
```
emanuele3d: fbosManager -> fboManager: picky, picky, picky. 

:smile:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
```diff
@@ -193,27 +177,24 @@ public boolean equals(Object other) {
...
-        private String parameterName;
 
         /**
          * Constructs an instance of this inner class, according to the given parameters.
          *
          * @param textureSlot an integer indirectly identifying a texture unit on the GPU (textureUnit = GL_TEXTURE0 + textureSlot).
          * @param textureId the opengl id of a texture, usually obtained via glGenTextures().
          * @param materialURN an URN identifying a material.
-         * @param parameterName the name of a variable in the shader program used to sample the texture.
+         * @param materialParameter a string
```
emanuele3d: Why the change of name and the shortening of the parameter description?

engine/src/main/java/org/terasology/rendering/opengl/AbstractFBOsManager.java
```diff
@@ -259,7 +262,13 @@ public FBOConfig getFboConfig(ResourceUrn fboName) {
      */
     @Override
     public boolean subscribe(FBOManagerSubscriber subscriber) {
-        return fboManagerSubscribers.add(subscriber);
+        // TODO: Add sanity checks to ensure no object can resubscribe
```
emanuele3d: I'd like this to happen in this PR please.

engine/src/main/java/org/terasology/rendering/opengl/AbstractFBOsManager.java
```diff
@@ -259,7 +262,13 @@ public FBOConfig getFboConfig(ResourceUrn fboName) {
...
+        // return fboManagerSubscribers.add(subscriber);
+
+        // VAMPCAT : TODO: RESTORE!!!
+        if (!fboManagerSubscribers.add(subscriber))
+            return false;
+        return true;
```
emanuele3d: Let's use a proper if/else statement here please.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.terasology.rendering.world;
 
+import org.terasology.rendering.dag.*;
```
emanuele3d: No star imports please.

engine/src/main/java/org/terasology/rendering/dag/nodes/ApplyDeferredLightingNode.java
```diff
@@ -57,22 +54,37 @@
...
         addDesiredStateChange(new SetInputTexture(
-                textureSlot++, sceneOpaqueFbo.depthStencilTextureId,  DEFERRED_LIGHTING_MATERIAL, "texSceneOpaqueDepth"));
+                textureSlot++, readOnlyGBufferFbo.depthStencilTextureId,  DEFERRED_LIGHTING_MATERIAL, "texSceneOpaqueDepth"));
         addDesiredStateChange(new SetInputTexture(
-                textureSlot++, sceneOpaqueFbo.normalsBufferTextureId, DEFERRED_LIGHTING_MATERIAL, "texSceneOpaqueNormals"));
+                textureSlot++, readOnlyGBufferFbo.normalsBufferTextureId, DEFERRED_LIGHTING_MATERIAL, "texSceneOpaqueNormals"));
         addDesiredStateChange(new SetInputTexture(
-                textureSlot,   sceneOpaqueFbo.lightBufferTextureId,   DEFERRED_LIGHTING_MATERIAL, "texSceneOpaqueLightBuffer"));
+                textureSlot,   readOnlyGBufferFbo.lightBufferTextureId,   DEFERRED_LIGHTING_MATERIAL, "texSceneOpaqueLightBuffer"));
+*/
```
vampcat: Ah, it was removed in the next PR. I'll take care of it here too.

engine/src/main/java/org/terasology/rendering/dag/nodes/ApplyDeferredLightingNode.java
```diff
@@ -85,8 +97,7 @@ public void initialise() {
     public void process() {
         PerformanceMonitor.startActivity("rendering/applyDeferredLighting");
 
-        sceneOpaquePingPongFbo.bind(); // TODO: remove and replace with a state change
-        sceneOpaquePingPongFbo.setRenderBufferMask(true, true, true);
+        writeOnlyGBufferFbo.setRenderBufferMask(true, true, true);
```
vampcat: That won't be needed, I converted this to a StateChange in the next PR.

engine/src/main/java/org/terasology/rendering/dag/nodes/BackdropNode.java
```diff
@@ -86,10 +81,9 @@ public void initialise() {
         new WireframeTrigger(renderingDebugConfig, this);
 
         addDesiredStateChange(new SetViewportToSizeOf(WRITEONLY_GBUFFER, displayResolutionDependentFBOs));
```
vampcat: I planned on doing this either way after asking for your opinion, glad you mentioned it on your own! :D

engine/src/main/java/org/terasology/rendering/dag/nodes/BackdropNode.java
```diff
@@ -123,12 +119,11 @@ public void disableWireframe() {
...
         PerformanceMonitor.startActivity("rendering/backdrop");
 
-        sceneOpaqueFbo.bind(); // TODO: remove when we can bind this via a StateChange
-        sceneOpaqueFbo.setRenderBufferMask(true, false, false);
+        readOnlyGBufferFBO.setRenderBufferMask(true, false, false);
 
         glCallList(skySphere); // Draws the skysphere
 
-        sceneOpaqueFbo.setRenderBufferMask(true, true, true); // TODO: handle these via new StateChange to be created
+        readOnlyGBufferFBO.setRenderBufferMask(true, true, true); // TODO: handle these via new StateChange to be created
```
vampcat: Adding to what I wrote in an earlier comment: The next PR takes care of this :D

engine/src/main/java/org/terasology/rendering/dag/nodes/BackdropReflectionNode.java
```diff
@@ -50,7 +50,6 @@
  *
  */
 public class BackdropReflectionNode extends AbstractNode {
-
```
vampcat: I usually do it manually.. Should I be avoiding that? I do it mostly to maintain consistency between files.

engine/src/main/java/org/terasology/rendering/dag/nodes/DeferredPointLightsNode.java
```diff
@@ -92,7 +85,12 @@ public void initialise() {
 
         addDesiredStateChange(new DisableDepthTest());
 
+        addDesiredStateChange(new BindFBO(READONLY_GBUFFER, displayResolutionDependentFBOs));
```
vampcat: Not really. I'll add it for now for the sake of consistency, but since the default viewPort is the size of screen, I would recommend that we remove all setViewportToSizeOf(READONLY_GBUFFER) StateChanges, seeing how redundant they are.

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
```diff
@@ -58,28 +57,24 @@ public void initialise() {
...
-        this.outputFBOconfig = outputConfig;
-        this.outputFBOmanager = outputManager;
-
-        requiresFBO(inputConfig, inputManager);
-        requiresFBO(outputConfig, outputManager);
-
-        outputFBO = outputManager.get(this.outputFBOconfig.getName());
-        outputManager.subscribe(this);
+        requiresFBO(inputFboConfig, inputFboManager);
+        outputFbo = requiresFBO(outputFboConfig, outputFboManager);
```
vampcat: I was the one who added the update() function to other nodes, most of them didn't even have an update() earlier.
However, the reason I have not used update() in CopyImageToHMDNode and in DownSamplerNode is because it isn't applicable.

Over here, requiresFbo() creates the Fbo, and returns it's FboId. It would be inefficient to ignore this return value, and then ask the fboManager for it's fboId - hence i broke the pattern at times.
Similarly in CopyImageToHMDNode, leftEye and rightEye are generated using requireFbo(), no need to query them again using update().

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
```diff
@@ -113,9 +110,6 @@ public void process() {
 
     @Override
     public void update() {
-        // Note: we don't need to update the inputFBO because only SetInputTextureFromFBO
-        // needs to keep it up to date and takes care of that internally.
-        outputFBO = outputFBOmanager.get(outputFBOconfig.getName());
+        outputFbo = outputFboManager.get(outputFboUrn);
```
vampcat: That comment would have made sense, had there been an inputFbo variable.

(I removed it, because it was useless)

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
vampcat: No, this line is essential here. The reason is, readOnlyGBufferFbo has not been initialized till now. It can either be manually initialized here (as I've done) or it can be initialized by calling update() here (not recommended, since that re-fetches refractiveReflectiveFbo again).

vampcat: Although since the cost of re-fetching the fboId isn't too high in the grand scheme of things, I guess we can call update() in such nodes to maintain a pattern.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
vampcat: The enum (and the corresponding retrieving function) are used for a BindTexture StateChange, so it should only support Textures. depthRenderBuffer is NOT a texture, and so logically shouldn't even be in the list. Even if I return the depthRenderBuffer Fbo, you can't just use it like the other returned values (in a glBindTexture(TEXTURE_2D, ...) call). So, for the sake of clear and logical code, I removed it.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
vampcat: Because, that was just an edge case that had to be handled because of badly designed code. As things stand, the update() function will only be called after subscription to fboManager, and the subscription fo fboManager takes place only after creation of task (in generateDefaultTask). As things stand, task can never be null inside update(), no matter what the conditions.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
vampcat: I hope you don't mind too much? I like clear code, and clear and informative names.

In this case, "Singular noun + Manager" should be the pattern (It's never sheepsManager, it's always sheepManager),and so I did it. I plan on renaming the FBOsManagers to FBOManagers in a separate PR too.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
vampcat: Okay, I can't figure out myself what I was thinking. I'm sorry, I'll revert it back.

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/BackdropReflectionNode.java
emanuele3d: If you did it manually this in particular is a matter of personal preferences and I'm not too bothered either ways. The important thing is that it's not some kind of automated functionality you are unaware of. So, no worries on this one.

engine/src/main/java/org/terasology/rendering/dag/nodes/DeferredPointLightsNode.java
emanuele3d: That's a good point: agreed. Let's remove them completely please.

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
emanuele3d: Sorry, I didn't explain myself clearly enough. I'm thinking the initialization pattern:

`fbo = requiresFBO(config, manager);`

instead of using update(), could be more widespread. For example, have a look in DeferredPointLightsNode. I think it could be applicable there, to initialize the readOnlyGBuffer variable. And the same could be done in other nodes, skipping the update() call within initialize.

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
emanuele3d: Sorry, I got confused here. You are right, it is no longer relevant.

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
emanuele3d: I'm not sure I'm understanding your first comment. That variable is only needed/used in update() from what I can see. The FBO is instantiated both CPU/GPU side in requiresFBO if I remember correctly. =?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
emanuele3d: Understood, so it was actually a bug to keep it there as it wouldn't have worked with glBindTexture.

Can you put a comment about it? Something like:

```java
// depthStencilRboId is a possible FBO attachment but is not covered by a case here 
// as it wouldn't work with the glBindTexture(TEXTURE_2D) call.
```

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
emanuele3d: No, I don't mind, if I did I would have been explicit about changing it. :wink:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
emanuele3d: I see what you mean. Good catch!

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
emanuele3d: I think it's inaccurate. The parameter is really on the shader, not on the material. And it's not really the parameter itself, but its name. I guess the most accurate name would be shaderParameterName. But I thought parameterName + the description was enough.

emanuele3d:
I responded to a number of "outdated" comments. Some require action, some not.

I did have a look at the new commits and I saw that they addressed most issues, thank you.

I think we are about 1-2 rounds of communication away from merging. Good job so far!

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
vampcat: Okay, in the last round of comments I realized that using update() to ensure consistency was better, but if you want, I'll replace it with fbo = requiresFbo() wherever required.

That being said, it cannot be applied in most nodes. The reason being, they simply use the FBOs without creating them. Point in case, the DeferredPointLightsNode. It does not use a requiresFbo(), so we have to use update() or manually do it. I decided to adopt the cheeky way ;)

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
vampcat: The compiler did not throw a warning for redundant variable, so I just assumed that the variable was used in process().. Turns out it wasn't.

As it turns out, I converted this to a update() call, so that the readOnlyGBufferFbo.attachDepthBufferTo(refractiveReflectiveFbo); part can also run, even for the first frame. That, and I made the two FBOs local to the update() method, since we don't use them anywhere else.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
vampcat: I'll add the comment as you suggest for now.

However, and I *might* be wrong here, I don't think it's particularly required, mainly because the enum is FboTextureTypes, and not FboAttachmentTypes, and depthStencilRboId is definitely not the former.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
emanuele3d: Sure. But the question is: why does it not use the requiresFbo() method? Because if a node binds an FBO, it surely requires it.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
emanuele3d: Giving a little bit more information, even if it's not strictly required, doesn't hurt.

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
```diff
@@ -137,8 +135,8 @@ public void process() {
 
     @Override
     public void update() {
-        readOnlyGBufferFbo = displayResolutionDependentFBOs.get(READONLY_GBUFFER);
-        refractiveReflectiveFbo = displayResolutionDependentFBOs.get(REFRACTIVE_REFLECTIVE);
+        FBO readOnlyGBufferFbo = displayResolutionDependentFBOs.get(READONLY_GBUFFER);
+        FBO refractiveReflectiveFbo = displayResolutionDependentFBOs.get(REFRACTIVE_REFLECTIVE);
```
emanuele3d: Why did you do this?

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/DownSamplerNode.java
vampcat: That's an interesting question, and I think the answer is pretty simple - since it's a static renderGraph right now (if you can call it a graph), and so the programmer just used the requiresFbo() in the first Node that uses the particular Fbo. That, and all nodes assume the default Fbos don't have to explicitly asked-for as requirement.

I inherited the problem :P

Cervator:
Tested again and merged! Huzzah!

emanuele3d:
Well done @vampcat!

PR #2845, opened 2017-03-21
Toggle Fall Damage. Command - damageStop

krishnasism:
### Contains

Added damageStop command to toggle damage infliction by fall.
### How to test

Start game with core gameplay.
Fall and test that there is damage. Add command to console - "damageStop" and damage infliction by fall will be toggled. To stop the feature simple enter "damageStop" again to the console.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/editor/layers/WidgetSelectionScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/editor/layers/WidgetSelectionScreen.java
@@ -64,6 +64,7 @@ public class WidgetSelectionScreen extends CoreScreenLayer {
 
     @Override
     public void initialise() {
+
         availableWidgets = find("availableWidgets", UIDropdownScrollable.class);
 
         // Populate the widget list.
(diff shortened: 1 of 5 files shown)


GooeyHub:

nihal111:
Also went through the issue and found that your solution is only specific to Vertical Fall damages instead of a complete God-mode. Refer to @portokaliu's comment [here](https://github.com/MovingBlocks/Terasology/issues/2837#issuecomment-288017321)

portokaliu:
If we were to nitpick problems in the code, the above mentioned issue about vertical collisions is true, the text for the command is not exactly its description (I kinda see you copy-pasted), thus also related to the permission level to stop damage, which should be cheat.

I'm also not exactly sure that the new HealthCommands class created when the event is processed would ever have the correct values of the player's current settings (the new class would listen to the event after it's created I think)

SkySom:
I feel like toggleFallDamage or something would be be a better command name?

Cervator:
Closing from inactivity and from being a relatively minor change that would be easy to remake in a new branch without the undesired style changes and with a better logic approach. Conflicts in the way now too.

I appreciate the effort @krishnasism and hope you might still be interested in trying out some new code sometime. It seems like we still are forming an opinion on how to best approach damage negation (see #2843 or #2837) so it would probably be easier to restart this from scratch at some point with more discussion done.

Issue #2364, opened 2016-06-20
Music abruptly stops during game startup, is fine when returning from a world


Cervator:
I suspect there is a context change of some sort with the music starting to play in one then gets killed off before the menu loads fully. As such might relate a little to #2074 

Disappears between these two statements:

```
14:14:02.011 [main] WARN  o.t.rendering.nui.skin.UISkinFormat - Failed to resolve UIWidget class UIInputBindButton, skipping style information
14:14:03.038 [main] WARN  o.t.engine.internal.TimeBase - Delta too great (16124), capping to 1000
```

When returning to the main menu from a world the menu theme starts and keeps playing as expected, it only stops abruptly on initial game load.
##

Cervator:
Interestingly this seems specific to running the game from source, at least in my mega workspace. If I run the Alpha 4 binary the music continues playing in the main menu.

Cervator:
Additional bit of detail: this *may* be instead about letting the game start in the background (which is the case at least for me running from IntelliJ in Windows 8 on my tablet). If I grab the game window before it fully initializes and it loads while the active window the menu music keeps playing after initial startup just fine.

Cervator:
Closing as possibly obsolete or fixed by #2945 - we can always reopen later. Likely some needed context/environment tweaks will effectively reset this anyway sooner or later.

PR #2846, opened 2017-03-21
Fixed issues #2527 and #2364, now main menu music looping infinitely

### LINKED ISSUE #2527: Main menu background track not looping
### How to reproduce
- Open game
- Turn up the audio and find some work to do :stuck_out_tongue: 
- Wait

The background track should play once and not loop back. That is the game goes silent after a while, once the song finishes.

Probably an easy fix.
##

omomthings:
### Contains

Fixes #2527, now music looping infinitely, simple and efficient.

### How to test

Go to main menu, enjoy the music till it ends and  restarts again.. the music is about 3 minutes..

### DIFF
--- a/engine/src/main/java/org/terasology/engine/modes/StateMainMenu.java
+++ b/engine/src/main/java/org/terasology/engine/modes/StateMainMenu.java
@@ -136,7 +136,7 @@ public class StateMainMenu implements GameState {
     }
 
     private void playBackgroundMusic() {
-        context.get(AudioManager.class).playMusic(Assets.getMusic("engine:MenuTheme").get());
+        context.get(AudioManager.class).playMusic(Assets.getMusic("engine:MenuTheme").get(),() -> playBackgroundMusic());
     }
 
     private void stopBackgroundMusic() {
--- a/engine/src/main/java/org/terasology/engine/modes/StateMainMenu.java
+++ b/engine/src/main/java/org/terasology/engine/modes/StateMainMenu.java
@@ -16,6 +16,7 @@
 package org.terasology.engine.modes;
 
 import org.terasology.audio.AudioManager;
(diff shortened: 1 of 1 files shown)


GooeyHub:

ThompsonTyler:
The music does loop infinitely, but not only in the main menu, when you start playing it just re-loops as well....

nihal111:
You would want to kill the thread in instances when the game is started etc.

omomthings:
Ah ok I thought that it would be nice to have a background music all along hahaha, I'll make the modifs!

omomthings:
@nihal111  the problem is solved for all the possible navigations from the menu and stops when we start the game, however, I found another bug that stops the music if the main menu is kept open for too long in the background, looks like if the game goes on a hibernation mode or something. I'll push when I find out if it's normal. I'll keep you informed.

nihal111:
Are you sure that you can reproduce it reliably?
There is also the case when the music stops abruptly (mostly at start), and I believe that is IntelliJ doing some optimizations, followed by a "Delta too great.." message. Have a look at the discussion in the issue and a related issue- https://github.com/MovingBlocks/Terasology/issues/2364.
The above issue doesn't seem to happen when running the game from the binary though.

Not sure whether what you discovered is the same, just fyi.

omomthings:
Yes I think the two are related, got the same log trace, will probably solve both at the same time.

omomthings:
Hello Nihal, tried to reproduce the problem in various ways, I think that it should be fine now, what happens is that for some reason the music was stopped, now the Listener ensures that the music is looping till we explicitly quit the menu.

Cervator:
Closing as finished by #2945 - added you to the credits for your efforts @omomthings thanks again :-)

Issue #2834, opened 2017-03-19
User can submit whitespace only messages to chat + enter with empty message should re-close the chat ui


Cervator:
We need to add a `trim()` somewhere and a check to see if a message is non-empty before sending it :-)

Additionally pressing `enter` with a *completely* empty chat entry should close the ui but at present inserts a single space instead (then a second `enter` then would manage to send that as an empty message)

So: Trim the message the user is attempting to send and if it is at that point empty close the chat ui instead of sending a blank message

Probably should review `ChatScreen` and/or `ChatSystem` to get started on this.

ThompsonTyler:
Hey Cervator, was able to get this fixed, but wanted some input(Will probably ask you on irc before you see this), should console close on an empty string as well? (The enter adding a space was a problem for all UIText, not just chat)

ThompsonTyler:
Will send in PR once I am back at computer and available to fix the branching on my end :P

Cervator:
Closing as completed by #2847 :-)

PR #2847, opened 2017-03-22
Empty space input for chat/console handling

ThompsonTyler:
### Contains

Fixes the problems addressed in #2834 

Inputting a string of empty spaces into either console or chat treats it as an empty message. (Does not send that message)
Inputting an empty message into console does nothing.
Inputting an empty message into chat closes the chat box.
Any messages are trimmed(white space before and after messages is removed)

### How to test

Open up chat, try sending an empty message.
Open up chat, try sending a message with only spaces in it.
Open up console, try sending an empty message.
Open up console, try sending a message with only empty spaces in it.

Check that console and chat both trim the message before it is sent.

Nothing that I know of!

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/ui/ChatScreen.java
+++ b/engine/src/main/java/org/terasology/logic/console/ui/ChatScreen.java
@@ -15,6 +15,7 @@
  */
 package org.terasology.logic.console.ui;
 
+import org.codehaus.plexus.util.StringUtils;
 import org.terasology.input.MouseInput;
 import org.terasology.logic.console.Console;
 import org.terasology.logic.console.CoreMessageType;
@@ -73,9 +74,9 @@ public class ChatScreen extends CoreScreenLayer {
         getManager().setFocus(commandLine);
 
         commandLine.subscribe(widget -> {
-            String text = commandLine.getText().trim();
+            String text = commandLine.getText();
 
-            if (!text.isEmpty()) {
+            if (StringUtils.isNotBlank(text)) {
                 String command = "say";
                 List<String> params = Collections.singletonList(text);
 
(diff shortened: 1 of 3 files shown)


GooeyHub:

ThompsonTyler:
Hey Nihal, Fixed up the console a bit, UI does run first, but it calls the super onto UIText, so implementing in UIText causes an empty input to trigger the above and let it do with it as it wants to. I changed the console a bit because it was still sending commands of empty strings, just wasn't doing anything with them, so I added a check for it.

I also caused it to not write the commandline to "" because UICommandEntry already does it, and that would instead just be back to back writing the exact same thing.

As for the hinttext, When the UIText is completely empty it is in a state that will show the hint text if available(Nothing has this actually used from what I can tell, so can't really show it) But if the text is empty then isShowingHintText is true. With out the check and actions in UIText it results in the problem of hitting enter on an empty string triggering the if state that is now an if-else and adding a space(Enter's "key character" is a space).


As for console not exiting, this was something I asked @Cervator  about for what it should do, he had said both should just act as if you typed nothing/did nothing, but shouldn't close the console window. Could add this logic if you feel that it is the more correct approach, I wasn't really sure which would feel more correct, both seem fine to me.

skaldarnar:
@GooeyHub ok to test

GooeyHub:

jellysnake:
@ThompsonTyler How's the work going on this? Are you able to add the features oniatus requested/have comments on those features?

ThompsonTyler:
Hey, sorry. Been busy lately and then kind of forgot what issues were addressed in which way to go back and fix stuff up.

Should be able to make that fix real quick, makes sense to me.

GooeyHub:

Issue #2821, opened 2017-03-12
Oddities surrounding maintaining momentum after hitting a ceiling


Cervator:
Noticed during some unrelated testing as by chance I spawned under a tree.

1. Be under a tree or some other low ceiling
1. Enable `hjump` in the game console
1. Jump!
1. Die horribly but *not* from smashing into the ceiling at high speed. Instead you seem to be redirected to the ground with full momentum kept meaning you splat as if you had fallen a hundred blocks down or so.

Indicates two possible issues:

1. The player shouldn't retain so much momentum (if any) on hitting a ceiling, unless the ceiling is in fact made out of upside-down trampolines or something crazy. Probably instead momentum should be reset and gravity allowed to take over to gradually assume full regular fall speed
1. The player *should* take damage on any high velocity impact to a surface other than the ground. This used to actually be the case where with `hspeed` enabled you could run horizontally into a wall so fast you'd turn yourself into a squished fly on a windshield

@portokaliu might be interested in this since it would seem similar to his work on reacting to velocity and such on items hitting blocks. Interested in any thoughts on the topic at least :-)

skaldarnar:
I'd like to chime in with a general question on retaining momentum of movement for players, since the magic dome over in LaS addresses a similar topic. We should find a uniform way to deal with these things so that it feels naturally (in all game modes). 
What would be physically the most accurate action to happen here? Is velocity redirection even a thing if you hit a (solid) wall or ceiling?

0shine0:
retaining momentum depends upon Coefficient Of Restitution (e). [https://en.wikipedia.org/wiki/Coefficient_of_restitution#Range_of_values_for_e_.E2.80.93_treated_as_a_constant](url)

As COR is a pair property. I will depend on both, the player and the surface the player hits upon. COR will depend upon yield strength, elastic modulus and other things of both colliding materials. But since we cannot reproduce those things in our game without complications, we can try to introduce our own property for material that decides overall COR for collision.

what I suggest is that we make a property say hardness for a single material ranging from -1 to 1. the average of this property between the 2 colliding materials will determine COR. if COR comes to be -ve than we can take it to be 0. as COR ranges from 0-1. 0 is completely inelastic collision and no momentum or energy is convserved after collision and 1 is completely elastic and no energy is lost during collision. Values in between determine the loss of energy.

Eg. object 1: hardness = -1
object 2: hardness = 1
object 3: hardness = 0
object 4: hardness = 0.5

Collision between:
[...]

PR #2848, opened 2017-03-22
Solved Oddities surrounding maintaining momentum after hitting a ceiling

0shine0:
### Contains

Fixes PR #2821 
Makes velocity of character = 0 when the character hits a ceiling. The gravity than does the rest.
Character takes damage if he/she hits the ceiling above allowable speed limit.
Made crashing speed generation more consistent and better than previously(Does not guarantee if player will take damage upon horizontal collision but is certainly better and consistent than the previous one)

### How to test

Run the game. Go under a low ceiling environment eg. under a tree.
Enable hjump.
Jump and die due to hitting on ground at a speed as if falling from a mountain.


GooeyHub:

PR #2849, opened 2017-03-22
Fixes Oddities surrounding maintaining momentum after hitting a ceiling

0shine0:
###Contains

Fixes PR #2821 
Makes velocity of character = 0 when the character hits a ceiling. The gravity than does the rest.
Character takes damage if he/she hits the ceiling above allowable speed limit.
Made crashing speed generation more consistent and better than previously(Does not guarantee if player will take damage upon horizontal collision but is certainly better and consistent than the previous one)

###How to test

Run the game. Go under a low ceiling environment eg. under a tree.
Enable hjump.
Jump and die due to hitting on ground at a speed as if falling from a mountain.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
+++ b/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
@@ -673,7 +673,13 @@ public class KinematicCharacterMover implements CharacterMover {
             }
         } else {
             if (moveResult.isTopHit() && endVelocity.y > 0) {
-                endVelocity.y = -0.5f * endVelocity.y;
+                if (input.isFirstRun()) {
+                    Vector3f hitVelocity = new Vector3f(state.getVelocity());
+                    hitVelocity.y += (distanceMoved.y / moveDelta.y) * (endVelocity.y - state.getVelocity().y);
+                    logger.debug("Hit at " + hitVelocity);
+                    entity.send(new VerticalCollisionEvent(state.getPosition(), hitVelocity));
+                }
+                endVelocity.y = -0.0f * endVelocity.y;
             }
 
             // Jump again in mid-air only if a jump was requested and there are jumps remaining.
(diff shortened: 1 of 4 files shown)


0shine0:
Pinging @Cervator  and @portokaliu for review

GooeyHub:

Cervator:
Also pinging @skaldarnar especially since #2823 seems sorta similar :-)

engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
```diff
@@ -673,7 +673,13 @@ private void walk(final CharacterMovementComponent movementComp, final Character
...
         } else {
             if (moveResult.isTopHit() && endVelocity.y > 0) {
-                endVelocity.y = -0.5f * endVelocity.y;
+                if (input.isFirstRun()) {
+                    Vector3f hitVelocity = new Vector3f(state.getVelocity());
+                    hitVelocity.y += (distanceMoved.y / moveDelta.y) * (endVelocity.y - state.getVelocity().y);
+                    logger.debug("Hit at " + hitVelocity);
+                    entity.send(new VerticalCollisionEvent(state.getPosition(), hitVelocity));
+                }
+                endVelocity.y = -0.0f * endVelocity.y;
```
0shine0: @skaldarnar I am keeping the multiplication because if at a later stage a variable Coefficient of Restitution(cor) is introduced than we can just put the variable name instead of zero and the collisions would work like a charm.

modules/Core/src/main/java/org/terasology/logic/health/HealthComponent.java
```diff
@@ -34,7 +34,7 @@
     @Replicate
     public float fallingDamageSpeedThreshold = 20;
     @Replicate
-    public float horizontalDamageSpeedThreshold = 100;
+    public float horizontalDamageSpeedThreshold = 20;
```
0shine0: I think we should have same speed damage in all directions. That's why I changes horizontalSpeed damage to 20. and I think we should just have 1 constant i.e. damageSpeedThreshold that works for all directions.

PR #2850, opened 2017-03-23
Fix #2836

### LINKED ISSUE #2836: ResettableUIText does not allow backspace or arrow key usage
Seems a somewhat amusing oversight: a UI element the purpose of which is to be easy to reset cannot easily be edited normally :D

Example: On the Modules screen (from Create Game) the module filter box can be used to filter the shown modules. You can *enter* text just fine - but seemingly not use backspace to then erase the text normally (partially for instance) nor other controls like arrow keys.

Interestingly you *can* still do text selection and even copy paste stuff in (including erasing whatever is selected)

Simple outcome desired: Arrow keys should work to navigate the text and backspace or delete should be allowed to erase individual characters.

Start by looking at `ResettableUIText ` and maybe similar text type widgets (the "Seed:" field back on the create game screen for instance supports regular text editing fine)

gianluca-nitti:
### Contains

As described in issue #2836, the ResettableUIText widget was not reacting properly to some keyboard events (e.g. backspace) due to `isShowingHintText` in the parent class never being updated. This solves the issue by setting `isShowingHintText` in `ResettableUIText.draw(Canvas)` like it's done in `UIText.draw(Canvas)`.

### How to test

Typing the backspace key in the search/filter field in the Modules screen (from Create Game) now works correctly, while before it was not erasing any text.

### Notes

I'm not sure if this is the right way to fix the bug, considering I'm duplicating ~8 lines of code from a class to a child class and changing the visibility of a couple of fields in the superclass to make them accessible in the subclass. A more sophisticated fix could be splitting the code in the `UIText.onDraw` method in multiple smaller methods and overriding only the strictly necessary one in `ResettableUIText` but I'm not sure how it could be done in practice.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/ResettableUIText.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/ResettableUIText.java
@@ -42,39 +42,12 @@ public class ResettableUIText extends UIText {
     @Override
     public void onDraw(Canvas canvas) {
         Rect2i clearButtonRegion = Rect2i.createFromMinAndSize(0, 0, 30, canvas.size().y);
-
-        if (text.get() == null) {
-            text.set("");
-        }
-        if (isShowingHintText) {
-            setCursorPosition(0);
-            if (!text.get().equals(hintText) && text.get().endsWith(hintText)) {
-                text.set(text.get().substring(0, text.get().length()-hintText.length()));
-                setCursorPosition(text.get().length());
-                isShowingHintText = false;
-            }
-        }
-        lastFont = canvas.getCurrentStyle().getFont();
         lastWidth = canvas.size().x - clearButtonRegion.size().x;
         if (isEnabled()) {
             canvas.addInteractionRegion(interactionListener, Rect2i.createFromMinAndMax(0, 0, canvas.size().x, canvas.size().y));
             canvas.addInteractionRegion(clearInteractionListener, Rect2i.createFromMinAndMax(canvas.size().x, 0, canvas.size().x +
                     clearButtonRegion.size().x, canvas.size().y));
         }
-        correctCursor();
-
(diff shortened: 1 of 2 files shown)


GooeyHub:

jerrickc:
If you're going to be bringing those 8 lines of code in, onDraw would be the same in both classes except for the first line in ResettableUIText so we could just reuse the parent method and an identical fix would be doing `@Override
Rect2i clearButtonRegion = Rect2i.createFromMinAndSize(0, 0, 30, canvas.size().y);
super.onDraw(canvas);
}` for onDraw in ResettableUIText.

gianluca-nitti:
Hi @jerrickc, thanks for the input.

I was thinking of doing as you describe, but the two methods aren't exactly the same: for example, the calculations for `lastWidth` and `widthForDraw` are different because `ResettableUIText` needs to keep in consideration the space used by the reset button. So, in my understanding, simply calling the superclass method will result in those values not being computed correctly.
I'll later see if there is a way to refactor this without breaking anything.

gianluca-nitti:
In the latest commit I removed the duplicated code by moving the common parts in a new method in `UIText`, `drawAll`.

An alternative way I considered was to completely reimplement `ResettableUIText` to extend `ColumnLayout` instead of `UIText`, and inserting in the layout one column for a `UIText` and one for a `UIButton` that would work as the reset button. However I dropped this idea since this would have probably broken the usage of the class, specifically when dynamic loading from JSON; as far as I understand, to make it work this way, one should define the properties fields (e.g. `hintText`) in the `ResettableUIText` class, marking them with the `@LayoutConfig` annotation, then at load time copy their values in the fields of the underlying `UIText`, then same thing with the methods (such as `getText`), definitely not a practical task.

Cervator:
@GooeyHub: ok to test

@gianluca-nitti: thanks for the PR! :-)

@rzats: this look OK to you?

GooeyHub:

flo:
Thanks for the PR, I tested it out and fixed the issue. Good work.

PR #2851, opened 2017-03-23
Add more unit tests

iaronaraujo:
### Contains

Adds unit tests for  org.terasology.config generated automatically with EvoSuite and edited to make it work only using JUnit and removing tests that didn't make sense.

### How to test

Run tests on engine-tests/src/test/evosuite/org/terasology/config (These are the added tests).


GooeyHub:

iaronaraujo:
I made a post about this if anyone want to see it: http://forum.terasology.org/threads/adding-more-unit-tests.1827/

skaldarnar:
@GooeyHub ok to test

First of all: Thanks for the PR and the wrap up post in the forum. I did not know about Evosuite before, that's why I'm really interested in what this tool does and how the generated tests look like. 😄 

Without having a look at _all_ the generated test cases, I can already note down some aspects I dislike about it:

* The generic names do not convey a good idea of what they are supposed to test. I understand that this is a problem that comes from the nature of generating test cases, but I still don't like it.
* The test cases seem a bit random at times. Usually, one tries to cover edge cases in unit tests. While I value generator-driven property checks (e.g., in [ScalaTest](http://www.scalatest.org/user_guide/generator_driven_property_checks)) I am not convinced by single, randomly generated tests. 

To give an example, look at the following test:
```Java
@Test(timeout = 4000)
public void test3()  throws Throwable  {
    AudioConfig audioConfig0 = new AudioConfig();
    audioConfig0.setMusicVolume((-774.0F));
    float float0 = audioConfig0.getMusicVolume();
    assertEquals((-774.0F), float0, 0.01F);
}
```
Imho, a better API for the `AudioConfig` would be only allow volume levels greater or equal to 0, or at least the volume retrieved from the getter cannot be smaller 0. That's why a developer has to put some thought into carefully crafting test cases most the time.

Please don't get me wrong, I still see benefits we get from this kind of tool. I really like how it works for boolean properties, there is not much else to test on these.
Furthermore, I'd like to understand the tool better. Why is it generating the following test case for `NUIEditorConfig`? What is its source of knowing (or guessing?) that the default for disabling autosave is false?

```Java
@Test(timeout = 4000)
public void test4()  throws Throwable  {
    NUIEditorConfig nUIEditorConfig0 = new NUIEditorConfig();
    boolean boolean0 = nUIEditorConfig0.isDisableAutosave();
    assertFalse(boolean0);
}
```


Let's keep this PR open for further discussion on this topic. I'll ping a few people to request their review on this.

GooeyHub:

iaronaraujo:
I completely agree with your dislikes. When I generated the tests I also noticed that Evosuite sometimes create tests that don't make sense, especially if it's trying to test methods with some random features. I also noted some redundancies on a few tests.

While it certainly has some problems, I think we could still use Evosuite to generate tests to the classes that don't have any tests yet, so we could have at least a few tests for them. I strongly think that we **mustn't** only use these tests, though. We could use the automatically generated tests until we make more detailed tests manually. We could also keep both!

If we decide to have automatically generated tests, I can also explore EvoSuite more, especially because I think we can create things like regression and mutation tests with it!

iaronaraujo:
@rzats Thank you very much for your review! I didn't find a way to change the names automatically, but I am looking forward writing a script that does at least some renamings. I was able to change all you asked manually, though. Could you check that up, please?

GooeyHub:

iaronaraujo:
After GSoC's proposal submission, I plan on doing something a little bit different. During the discussion on IRC, we noticed that the real value of these tests is not the test itself, but the discussion that emerges when we analyze it carefully. We see some strange things like setting a very high volume or setting negative heights and then we clearly see that EvoSuite is not capable of knowing when a certain value makes sense or not. But these are the best tests to discuss because it makes us ask ourselves "Wait, what should we actually do if this method receives this weird value?".

So, on this matter, we decided that when I get back on working on this, I should probably close this issue and open a new one with two specific commits. The first one would be the tests generated by EvoSuite just like the ones I committed on this PR (Weird Values, Unmeaningful Names, etc). Then I would make a second commit fixing the test, improving its quality and also making code changes that answer the question asked in the previous paragraph. By doing this, I would be creating an approach for new testers wanting to contribute to terasology because this would give them an interesting place to start by generating weird tests and trying to fix them to further understand terasology's whole code! When I finish this new PR, I will also write a document or a forum post explaining how to follow this approach! 

I hope you all like this idea, and feel free to give feedback and suggestions!

GooeyHub:

iaronaraujo:
I will begin to work on the new PR mentioned on my comment above

Issue #2855, opened 2017-03-24
[suggestion] MOTD for public servers


qwc:
As I am contributing/running two terasology servers (latest stable, latest dev; aka "official stable/unstable" in the server list) I would like to inform new players about some rules/circumstances which apply for these servers.
For example, that when a new unstable/stable version is released that these servers are automatically immediatly updated and restarted, so that everyone playing is warned that this can happen anytime. 
Or that sometimes a complete server-world reset is done due to bugs or incompatibilities between versions.

This would be best achieved with a "Message of the Day"-like approach, just a window popping up on join of the player, either with an okay button or that the window is closed with the good old "any key". ;)

It would be nice if some stylesheeting would be possible, some html/css maybe or markdown. 
Thinking about that a second time I would rather like **markdown** instead of complicated html/css. :+1: 

Could be a bite-sized little task for gsoc-ers or for other bored devs. ;)

skaldarnar:
> [...] or for other bored devs. ;)

Whatever that is 😛

gianluca-nitti:
@qwc It's a nice feature in my opinion. I implemented it in #2856, however that's just plain text for now since I'm not sure how styling should be implemented. In my implementation the MOTD is displayed as a `MessagePopup` which internally uses an `UILabel` to display text. AFAIK this widget doesn't support styling, however it might be possible to set a global style for the entire displayed string using UI skins.

(To experienced devs) Maybe the `BrowserWidget` class may be useful to render markup?

Here is a simple proof-of-concept of what I implemented:
![motd](https://cloud.githubusercontent.com/assets/15270303/24322463/33ddd654-1165-11e7-9589-25629061364f.gif)

qwc:
How to configure:
<gianluca_n> you just add a voice named "serverMOTD" under "network", for example i put it after the "serverPort"

* using this issue as my private notepad... *

qwc:
Huge thanks to @gianluca-nitti for the first draft. I'm looking forward to the next unstable builds when the PR is integrated.

Possible extensions are now: 
* Maybe BrowserWidget to **render Markdown**
* Transfer of the MOTD into an own config file (as Cervator said, this may not be bite-sized)

Naman-sopho:
I would like to work on the modification this MOTD for public servers namely the

- Styling of the MOTD.
- Render the MOTD from a file if its available.

jellysnake:
@Naman-sopho.  
[...]

PR #2856, opened 2017-03-25
Implemented server MOTD (Message Of The Day)

gianluca-nitti:
### Contains

Implemented server MOTD (Message Of The Day) as suggested in #2855 (no html/markdown formatting for now).

### How to test

Spin up a server (for example launch with `-headless -homedir=server-dir`) to create a configuration. Wait for the server to finish the initialization and start listening on network interfaces, then stop it. Now, open the `config.cfg` in the server's home directory (`server-dir` in the example), go in the `network` section and add a `serverMOTD` field, for example:
```
...
    "serverPort": 25777,
    "serverMOTD": "Hi, I'm the server's MOTD! (change it as you like)",
    "masterServer": "meta.terasology.org"
...
```
Now start the server again, start a client and join the server (`localhost` if on the same machine). As soon as you get to the `StateInGame` state (the loading screen finishes) you'll see a message dialog with the MOTD.

Again as in my previous PR, I'm not sure if it's the right way of implementing this feature, any kind of improvement suggestion is obviously welcome. In particular, I modified the protobuf message schema `NetMessage.proto` and recompiled it to java with the appropriate gradle task, and the output is included in the commits (`NetData.java` and `EntityData.java`). I'm not sure why these generated files are checked in into git despite the fact they can built automatically.

### DIFF
--- a/engine/src/main/java/org/terasology/config/NetworkConfig.java
+++ b/engine/src/main/java/org/terasology/config/NetworkConfig.java
@@ -69,7 +69,7 @@ public class NetworkConfig {
         this.serverPort = serverPort;
     }
 
-    public String getServerMOTD(){
+    public String getServerMOTD() {
         return serverMOTD;
     }
 
--- a/engine/src/main/java/org/terasology/config/NetworkConfig.java
+++ b/engine/src/main/java/org/terasology/config/NetworkConfig.java
@@ -73,6 +73,10 @@ public class NetworkConfig {
         return serverMOTD;
     }
 
+    public void setServerMOTD(String serverMOTD) {
+        this.serverMOTD = serverMOTD;
+    }
+
     public void addServerInfo(ServerInfo serverInfo) {
         servers.add(serverInfo);
     }
--- a/engine/src/main/java/org/terasology/config/NetworkConfig.java
+++ b/engine/src/main/java/org/terasology/config/NetworkConfig.java
@@ -40,6 +40,11 @@ public class NetworkConfig {
     private int serverPort;
 
     /**
+     * The server's Message of The Day
+     */
+    private String serverMOTD;
+
+    /**
      * The master server URL
      */
     private String masterServer = "meta.terasology.org";
@@ -62,6 +67,10 @@ public class NetworkConfig {
 
     public void setServerPort(int serverPort) {
         this.serverPort = serverPort;
+    }
+
+    public String getServerMOTD(){
+        return serverMOTD;
     }
 
     public void addServerInfo(ServerInfo serverInfo) {
--- a/engine/src/main/java/org/terasology/engine/modes/StateIngame.java
+++ b/engine/src/main/java/org/terasology/engine/modes/StateIngame.java
@@ -102,8 +102,9 @@ public class StateIngame implements GameState {
             }
         });
 
-        if (networkSystem.getMode() == NetworkMode.CLIENT)
+        if (networkSystem.getMode() == NetworkMode.CLIENT) {
             nuiManager.pushScreen(MessagePopup.ASSET_URI, MessagePopup.class).setMessage("Server MOTD", networkSystem.getServer().getInfo().getMOTD());
+        }
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


GooeyHub:

Cervator:
@GooeyHub: ok to test

@gianluca-nitti: well done! Will check this out when able :-)

GooeyHub:

Cervator:
Works great! Thanks @gianluca-nitti :-)

As a nice convenience option you can use a tiny JSON file to adjust config for a headless server without first having to generate a `config.cfg` - for instance I have an `override.cfg` in my workspace root (which is covered by the `.gitignore` yet automatically provided locally when you run Gradle) and the headless server launch config in IntelliJ will apply it automatically. So I adjusted my copy to look like so for testing:

```
{
  "defaultModSelection": {
    "modules": [
      "JoshariasSurvival",
      "CheatsForAll"
    ],
    "defaultGameplayModuleName": "JoshariasSurvival"
  },
  "worldGeneration": {
    "defaultGenerator": "Core:facetedperlin"
  },
  "network": {
    "serverMOTD": "Hi, I'm the server's MOTD! (change it as you like)"
  }
}
```

Worked first try no problem!

As for why the protobuf files are checked in - it was done so long ago and they change so rarely. This might be the first change in a year or two? But yes we could probably just make Gradle auto-generate them if needed :-)

I wonder if this change would make it so you cannot successfully connect to a new server with this included from an older game client.

gianluca-nitti:
> As a nice convenience option you can use a tiny JSON file to adjust config for a headless server without first having to generate a config.cfg

Handy suggestion, thanks.

> I wonder if this change would make it so you cannot successfully connect to a new server with this included from an older game client.

Should work. From [here](https://martin.kleppmann.com/2012/12/05/schema-evolution-in-avro-protocol-buffers-thrift.html), about protobuf:
> You can add a field to your record, as long as it is given a new tag number. If the Protobuf parser parser sees a tag number that is not defined in its version of the schema, it has no way of knowing what that field is called. But it does roughly know what type it is, because a 3-bit type code is included in the first byte of the field. This means that even though the parser can’t exactly interpret the field, it can figure out how many bytes it needs to skip in order to find the next field in the record.

So the old client should simply ignore the MOTD field.

I tested it now and it worked, however the first time I tried I had a crash but it's probably due to my workspace: to avoid keeping two copies of the codebase on my machine, I tried to test with this quick&dirty solution:
* Checked out `server-motd` branch
* Started server (headless run configuration)
* When it finished initialization, switched to the `develop` branch (the relevant binary code should be in memory now, changing them on disk shouldn't affect)
* Launched a client and tried connecting

The first time I had a `NullReferenceException` when I clicked on the server's name in the list; then I shut the server down and retried the whole process a couple of times and everything worked OK (launched the server built from `server-motd`, then the client built from `develop` and it joined correctly - without showing the MOTD obviously - and finally, without touching the running server, switched back to `server-motd` and rebuilt and launched the client that correctly joined and showed the MOTD). No idea what exactly happened the first time though, could be that I switched the branches too fast? Sounds very unlikely since the server was running so in my understanding changes to the files on disk shouldn't have affected.

Cervator:
Cool to hear about the protobuf thing :-)

I did find a crash as well: if I use a *non-headless* server I crash on clicking the entry in the server list like you mentioned above. So in other words start two regular clients, make one host the game, make the other connect.

I tried it a few times and it seemed consistent. Or can you confirm you can get past that somehow?

Headless seemed fine :-)

gianluca-nitti:
I just successfully reproduced the problem you described. I launched two instances, from the same branch, and tried joining. If the hosting instance has the `serverMOTD` defined in it's configuration file everything goes well, while if the value is not present the client crashes with a `NullPointerException` when trying to join and the server reports an exception of the same type (but keeps running).

I'm sorry I haven't tested this scenario before opening the PR. I'm now going to further investigate and try to fix it (should be quite easy, probably I just have to check for nulls and handle them by assigning default values).

gianluca-nitti:
As I was suspecting the problem was not depending on the server being headless or not, but on the server having a `serverMOTD` configuration entry. This new commit adds checks to avoid attempting to access the value when it's null, and also avoids displaying the popup when joining an "old" server or a "new" server without a MOTD. A `setServerMOTD` method has been added added in `NetworkConfig` too to allow to later implement the possibility to set the message from the GUI (for non-headless servers).

GooeyHub:

Cervator:
Looks fixed now! Thanks for that :-)

I'm seeing @msteiger signed up for review so will wait a bit to give him a chance. Otherwise I'm good with merging any time now (feel free to merge after you review)

Cervator:
Merged :-)

Thanks again! Added to credits too.

gianluca-nitti:
@Cervator Thanks to you.

Issue #2804, opened 2017-03-04
Oddities with lighting from held objects (held torches yield no light)


Cervator:
This may be two separate issues but they are so similar I wanted to document both in one issue for now.

1) Held torches yield no light for the immediate surroundings, while thrown or placed torches (blocks) light up stuff as expected

![2017-03-04 13_03_00-terasology _ alpha](https://cloud.githubusercontent.com/assets/1063833/23581040/f1391ce0-00da-11e7-8c54-adbab8bffb20.png)

2) Some items when held appear fully lit regardless of surroundings, while others are lit appropriately by the environment's light level. This seems to be affecting primarily (or only?) billboard type blocks, so plants in other words

![2017-03-04 12_57_53-terasology _ alpha](https://cloud.githubusercontent.com/assets/1063833/23581019/b03bd430-00da-11e7-9d97-4646588b3dcc.png)

Maybe worth noting is that the held torch is *itself* lit despite being a shaped block (or is it an item when held? I forget) yet other shaped blocks (like stair blocks) or items are not lit.

This broke some weeks ago, during development of Alpha 7, but unsure of a more exact time frame.

Perhaps of interest to @emanuele3d, @tdgunes, or @Josharias

DPirate:
Hi,
Where can i get started on this?
How and where does terasology differentiate between item / shaped-block? Where are held blocks/items rendered?
Cactus block also seems to be lit while held so I think its not just billboard shaped.

dannyzhou98:
Hey all!
I was trying to familiarize myself with the code base and examine this issue as well. It looks like @DPirate is correct in terms of the translucent property causing this oddity. According to this [page](https://github.com/Terasology/TutorialAssetSystem/wiki/Block-Attributes), translucency tells "Whether the texture is masked". Is the solution as simple as turning all translucent attributes to false, or is the translucent property itself currently incorrect?

DPirate:
@dannyzhou98 Given that translucent objects are needed , its about how translucent property of the held object is being handled.

emanuele3d:
I find "Whether the texture is masked" as a somewhat unclear description. What does that really mean? That the alpha channel is taken into account? How? Via alpha rejection which is a blunt tool, or via proper blending, which is more "correct" but also more expensive? 

[...]

PR #2857, opened 2017-03-25
Moving SimpleBlendMaterialsNode from post rendering to world generation

DPirate:
SimpleBlendMaterialsNode renders semi-transparent objects, therefore its appropriate for it to be under world-generation section. It partially fixes issue #2804  

### Contains
Moved SimpleBlendMaterialsNode, to world-generation section (above deferred lighting section) in initRederGraph() function of WorldRendererImpl class.

### Arguments 
- As I mentioned , it renders semi-transparent objects.

- Putting this node after deferred lighting nodes meant semi-transparent objects were being flatly lit.

- Putting this node before deferred lighting nodes gives this node a flatly lit world to blend from but deferred lighting nodes also enable blending, so I think that incorporates  the later lighting effects.

- This node shares the same render code as opaque objects, except the fact that GL_Blend is active.

### Outstanding issues

- [ ] blending against sky, still lit up the objects
- [ ] depth issue with water blocks, objects seem to be rendering behind them(chunksRefractiveReflectiveNode)

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -281,6 +281,9 @@ public final class WorldRendererImpl implements WorldRenderer {
         Node overlaysNode = nodeFactory.createInstance(OverlaysNode.class);
         renderGraph.addNode(overlaysNode, "overlaysNode");
 
+        Node simpleBlendMaterialsNode = nodeFactory.createInstance(SimpleBlendMaterialsNode.class);
+        renderGraph.addNode(simpleBlendMaterialsNode, "simpleBlendMaterialsNode");
+
         // TODO: remove this, including associated method in the RenderSystem interface
         Node firstPersonViewNode = nodeFactory.createInstance(FirstPersonViewNode.class);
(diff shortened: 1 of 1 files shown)


GooeyHub:

emanuele3d:
@DPirate: I am sorry but I really do not think this is the right approach. 

1. By shifting the addition of semi-transparent objects to an earlier stage in the pipeline those objects won't take advantage of a number of effects that are added afterwards, i.e. ambient occlusion and outline. 

2. Suspension of disbelief is maintained reasonably well if errors, i.e. visual problems, are consistent and do not change depending on the circumstances. The introduction of new issues such as lit-up rendering against the sky or semi-transparent objects behind the water surface goes against this concept and makes suspension of disbelief more likely to be disrupted. In this context I'd prefer for the rendering of semi-transparent objects to remain consistently wrong than sometimes right/wrong depending on the circumstances.

I'd therefore recommend you to close this PR and work on understanding which shader exactly gets to render the problematic objects, why it doesn't do its job and how can it and/or the node be modified so that it does.

emanuele3d:
Closing PR in the meantime: further discussions can be had here, in the forum or in chat as necessary.

PR #2870, opened 2017-03-31
Hot-reloading of shaders

vampcat:
As things stand, gestlat will load the changed shaders in memory, but not compile them. This PR changes that, by adding a command "recompileShaders" to the console.
Note: The reason they are not automatically compiled are because in case the developer makes a mistake, the shader log is shown as a warning, and he can rectify it before attempting to recompile.


vampcat:
Tagging @emanuele3d for review.

GooeyHub:

Cervator:
This seemed simple enough so I went ahead and tested the command (didn't crash anything!) and merged it :-)

Thanks!

PR #2876, opened 2017-04-02
Fixed the backwards axe image so that axe head hits blocks when used.

ianbrobin:
### Contains

Axe used to be backward, so it looks like the player is smashing the back of the axe into the tree. I flipped the image so that it looks correct.

### How to test

Create world with core gameplay and use axe on anything.


GooeyHub:

jellysnake:
Just tested this. Axe is no longer backwards but is still not orientated correctly.
Your axe
The handle is not facing the player, even though the blade is indeed no longer backwards.

Cervator:
Thanks for the initiative @ianbrobin and the testing @jellysnake :-)

Yes this is indeed an issue with "older" icons as some time ago the icon rendering was modified in a way that reversed their appearance when "held" by the player. There is a ton of older icons now reversed, and newer icons that are drawn the new way. Also a bunch that just plainly don't align (especially on some of the GCI art assets)

I think one GCI task became writing up guidelines for making icons point correctly but ... I can't find where it went :/

One way to make this PR more useful would be to fix the orientation entirely then apply it to the other related icons (same asset location / texture atlas)

jellysnake:
@ianbrobin how's this pull request going? I'd like to be able to merge this.

jellysnake:
Pinging @Cervator to close this due to inactivity

PR #2879, opened 2017-04-02
Fixes all resizing issues

vampcat:
Contains fixes for the remaining resizing issues.

Existing Problem 1: When you start a game, exit to the main menu and start the game again, resizing will become corrupt and any resizing (maybe even no resizing) that you do will lead to problems.
Cause and Solution: The reason behind that is a static defaultInstance object of class SetViewportToSizeof, and this PR resets that object to null when disposing the RenderGraph.

Existing Problem 2: When playing with scale set to anything other than 100%, the game would not render to entire screen, and would draw only to a small part if scale < 100%, or outside bounds if scale > 100%.
Cause and Solution: This was caused because setViewportToSizeOf used an FBO to obtain the size, but since scale was not 100%, size of FBO was not same as size of Display. This PR fixes that by forcing last Node to set viewport to size of Display (there is still scope for improvement).

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
@@ -15,6 +15,7 @@
  */
 package org.terasology.rendering.dag.nodes;
 
+import org.lwjgl.opengl.Display;
 import org.terasology.assets.ResourceUrn;
 import org.terasology.rendering.dag.ConditionDependentNode;
 import org.terasology.config.Config;
@@ -27,6 +28,8 @@ import org.terasology.rendering.opengl.FBO;
 import org.terasology.rendering.opengl.FBOManagerSubscriber;
 import org.terasology.rendering.opengl.fbms.DisplayResolutionDependentFBOs;
 import org.terasology.rendering.world.WorldRenderer;
+
+import static org.lwjgl.opengl.GL11.glViewport;
 import static org.terasology.rendering.opengl.OpenGLUtils.renderFullscreenQuad;
 import static org.terasology.rendering.opengl.fbms.DisplayResolutionDependentFBOs.FINAL_BUFFER;
 import static org.terasology.rendering.world.WorldRenderer.RenderingStage.LEFT_EYE;
@@ -45,6 +48,8 @@ public class CopyImageToScreenNode extends ConditionDependentNode implements FBO
     private DisplayResolutionDependentFBOs displayResolutionDependentFBOs;
 
     private FBO sceneFinalFbo;
+    private int displayWidth;
+    private int displayHeight;
 
     @Override
     public void initialise() {
@@ -60,6 +65,7 @@ public class CopyImageToScreenNode extends ConditionDependentNode implements FBO
     public void process() {
         PerformanceMonitor.startActivity("rendering/copyImageToScreen");
         sceneFinalFbo.bindTexture(); // TODO: Convert to a StateChange
+        glViewport(0, 0, displayWidth, displayHeight); // TODO: Convert to a StateChange
         renderFullscreenQuad();
         PerformanceMonitor.endActivity();
     }
@@ -67,5 +73,7 @@ public class CopyImageToScreenNode extends ConditionDependentNode implements FBO
     @Override
     public void update() {
         sceneFinalFbo = displayResolutionDependentFBOs.get(FINAL_BUFFER);
+        displayWidth = Display.getWidth();
+        displayHeight = Display.getHeight();
     }
 }
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
@@ -65,7 +65,10 @@ public class CopyImageToScreenNode extends ConditionDependentNode implements FBO
     public void process() {
         PerformanceMonitor.startActivity("rendering/copyImageToScreen");
         sceneFinalFbo.bindTexture(); // TODO: Convert to a StateChange
-        glViewport(0, 0, displayWidth, displayHeight); // TODO: Convert to a StateChange
+        // The way things are set-up right now, we can have FBOs that are not the same size as the display (if scale != 100%).
+        // However, when drawing the final image to the screen, we always want the viewport to match the size of display,
+        // and not that of some FBO. Hence, we are manually setting the viewport via glViewport over here.
+        glViewport(0, 0, displayWidth, displayHeight);
         renderFullscreenQuad();
         PerformanceMonitor.endActivity();
     }
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
@@ -96,9 +96,8 @@ public final class SetViewportToSizeOf implements FBOManagerSubscriber, StateCha
         return fboManager.get(fboName);
     }
 
-    public static void ResetDefaultInstance() {
-        // TODO: Make this a method of StateChange, and override it here.
-        // TODO: Maybe rename this to dispose() ?
+    public static void disposeDefaultInstance() {
+        // TODO: Make a generic dispose() method for StateChange, and override it here.
         defaultInstance = null;
     }
 
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
@@ -96,6 +96,12 @@ public final class SetViewportToSizeOf implements FBOManagerSubscriber, StateCha
         return fboManager.get(fboName);
     }
 
+    public static void ResetDefaultInstance() {
+        // TODO: Make this a method of StateChange, and override it here.
+        // TODO: Maybe rename this to dispose() ?
+        defaultInstance = null;
+    }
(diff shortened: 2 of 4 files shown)


GooeyHub:

emanuele3d:
@vampcat: can you please better justify this fix? In which circumstances problems arise?

vampcat:
I believe the description is much more elaborate now.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
```diff
@@ -60,12 +65,15 @@ public void initialise() {
     public void process() {
         PerformanceMonitor.startActivity("rendering/copyImageToScreen");
         sceneFinalFbo.bindTexture(); // TODO: Convert to a StateChange
+        glViewport(0, 0, displayWidth, displayHeight); // TODO: Convert to a StateChange
```
emanuele3d: The problem with this solution is that it side-steps the problem instead of solving it. 

Can the SetViewportToSideOf class be made scale-aware?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -96,6 +96,12 @@ private FBO getFbo() {
         return fboManager.get(fboName);
     }
 
+    public static void ResetDefaultInstance() {
```
emanuele3d: How about "disposeDefaultInstance".  Yes, please no capitalized first letter.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -118,4 +118,8 @@ public void swapReadWriteBuffers() {
         fboLookup.put(WRITEONLY_GBUFFER, fbo);
         notifySubscribers();
     }
+
+    public FBO.Dimensions getFullScale() {
+        return fullScale;
+    }
```
emanuele3d: Why did you add this? Where is this used?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
```diff
@@ -60,12 +65,15 @@ public void initialise() {
     public void process() {
         PerformanceMonitor.startActivity("rendering/copyImageToScreen");
         sceneFinalFbo.bindTexture(); // TODO: Convert to a StateChange
```
vampcat: Ideally, we should be able to. I tried converting it, but for some reason I could not get it to work (I think it was mid-PR, lots of broken files. I'll try again, and post an update!)

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
vampcat: Not really. The problem is, SetViewportToSizeOf should only consider the size of FBO, and not whether the size is because the display is actually that size or because the display * scale is that size. Which is the case right now. We can make a new SetViewportToSizeOfDisplay StateChange, but that's it. I tried to integrate this into SVTSO, but found no plausible way.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
vampcat: I'm really sorry for the capitalization, idk what I was thinking.
Must be my left-over habit from the pre-Terasology era :P

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
vampcat: This was an oversight, I earlier tried to do fbo.size / scale, and then I relized it's more efficient to do display.size. I'll fix this!

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
emanuele3d: Ok. Then, let's remove the TODO comment and let's add an _extensive_ comment explaining why we do this this way rather than through SVTSO.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
vampcat: Okay, so I remembered why I could not convert this to a SetInputTextureFromFBO call earlier.

SetInputTextureFromFbo expects a slot, an fbo an attachment type, a materialUrn and a materialParameter. fbo.bindTexture() operates on an fbo, and uses the ColorBufferTextureId. The missing component between bindTexture() and SITFF is the slot (which I assume will be 0 for color?), the materialUrn (no idea) and the materialParameter.

Hence, I gave up mid-conversion.

vampcat: Also, the bindTexture() is used by other Nodes as well, I think I'll create a separate PR for it if you don't mind :)

GooeyHub:

vampcat:
I believe that should take care of it all, feel free to correct me.

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
emanuele3d: The slot is likely to be 0 but you should check the relevant ShaderParameters class.
The materialUrn is the ResourceUrn uniquely identifying the material/shader being enabled via EnableMaterial. The materialParameter is the name of the sampler variable in the shader receiving the texture.

I'm ok if you'd like to change these type of lines altogether rather than here.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -551,8 +551,8 @@ public void dispose() {
         renderableWorld.dispose();
         worldProvider.dispose();
         // TODO: Shift this to a better place, after a RenderGraph class has been implemented.
-        // TODO: Call ResetDefaultInstance() for every StateChange after it is a method of StateChange.
-        SetViewportToSizeOf.ResetDefaultInstance();
+        // TODO: Call disposeDefaultInstance() for every StateChange after it is a method of StateChange.
```
emanuele3d: I'm not sure what you mean with this line of comment. Can you rephrase?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
vampcat: So I'll have to check the materialParameter from the relevant shader?
Oh boy, so fun! :D

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
vampcat: By this, I mean that we ideally should have a disposeDefaultInstance() for every single StateChange, rather than just for SVTSO (most of them will probably be empty for now). Will make it possible to call dispose() for every single StateChange/Node.
Does that sound good?

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
emanuele3d: Yep, that's less than ideal. Hopefully one day we'll get to sort that out too, i.e. doing some runtime editing of the shaders source code.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: I'm not sure I'd want it at StateChange level. Some StateChanges, if not many, do have an absolute default that might not be necessary to remove. I.e. the default BindFBO state change binds to fboId 0, which is the default frame buffer. Would there be a need to dispose it?

vampcat: Because, objects should only live for as long as they are useful. The user quits game screen, all objects useful only in game screen should be disposed.

emanuele3d: That's dogmatic. 

If the user continues its way out and closes the app everything gets trashed. If the user starts a new game startup will be a smidgen faster because it doesn't have to regenerate those default instances. 

In this case it is the right thing to do. In other cases I remain unconvinced.

vampcat: Fair enough, I'll drop the comment (and the idea) for the time being, and will try again when I am more convinced of the idea myself xD

GooeyHub:

emanuele3d:
@Cervator: this PR is good to go. 

I also just tried something new. I reasoned that vampcat will always be tied to this PR as he's the author. But  I thought of changing the assignee to you as the PR to progress requires you merging it and no longer requires vampcat's input. 

The idea here being that the assigned people are only those who have something to do for a PR. It's just a process idea I'm testing, does not need to be rolled out anywhere fast.

Cervator:
Tried my best to break it, including resizing, scaling, reloading, and screenshotting - seemed to handle it all!

@emanuele3d yep shuffling assignee around that sounds fine :-)

Sometimes I assign it to reviewers as well as request said review ... but that might be overkill. But yeah switching to a merge type person makes sense (we just need more of them than just me, hehe)

PR #2882, opened 2017-04-03
Split up the graph intiaization into multiple functions.

hybrideagle:
### Contains

Split up the initRenderGraph() function in WorldRenderImpl into multiple smaller functions.

### How to test

Run the game.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -217,7 +217,41 @@ public final class WorldRendererImpl implements WorldRenderer {
         NodeFactory nodeFactory = new NodeFactory(context);
         RenderGraph renderGraph = new RenderGraph();
 
+        FBOConfig sceneOpaqueFboConfig = displayResolutionDependentFBOs.getFboConfig(READONLY_GBUFFER);
+
         // ShadowMap generation
+        addShadowMapNodes(renderGraph,nodeFactory);
+        // (i.e. water) reflection generation
+        addWaterReflectionNodes(renderGraph,nodeFactory);
+        // sky rendering
+        addSkyNodes(renderGraph,nodeFactory,sceneOpaqueFboConfig);
+
+        // world rendering
+        addWorldNodes(renderGraph,nodeFactory);
+
+        // TODO: remove this, including associated method in the RenderSystem interface
+        Node firstPersonViewNode = nodeFactory.createInstance(FirstPersonViewNode.class);
+        renderGraph.addNode(firstPersonViewNode, "firstPersonViewNode");
+
+        // lighting
+        addLightingNodes(renderGraph,nodeFactory);
+
+        // TODO: consider having a none-rendering node for FBO.attachDepthBufferTo() methods
+
+        // 3d-based decorations (versus purely 2d, post-production effects)
+        add3dDecorationNodes(renderGraph,nodeFactory);
+
+        
+        // Pre-post-processing, just one more interaction with 3D data (semi-transparent objects, in SimpleBlendMaterialsNode)
+        // and then it's 2D post-processing all the way to the image shown on the display.
+        addPostProcessingNodes(renderGraph,nodeFactory,sceneOpaqueFboConfig);
+
+        renderTaskListGenerator = new RenderTaskListGenerator();
+        List<Node> orderedNodes = renderGraph.getNodesInTopologicalOrder();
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -217,7 +217,41 @@ private void initRenderGraph() {
...
+
+        // lighting
+        addLightingNodes(renderGraph,nodeFactory);
+
+        // TODO: consider having a none-rendering node for FBO.attachDepthBufferTo() methods
+
+        // 3d-based decorations (versus purely 2d, post-production effects)
+        add3dDecorationNodes(renderGraph,nodeFactory);
+
+
```
emanuele3d: Let's remove this unnecessary empty line.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -217,7 +217,41 @@ private void initRenderGraph() {
...
+        // TODO: consider having a none-rendering node for FBO.attachDepthBufferTo() methods
+
+        // 3d-based decorations (versus purely 2d, post-production effects)
+        add3dDecorationNodes(renderGraph,nodeFactory);
+
+        
+        // Pre-post-processing, just one more interaction with 3D data (semi-transparent objects, in SimpleBlendMaterialsNode)
+        // and then it's 2D post-processing all the way to the image shown on the display.
+        addPostProcessingNodes(renderGraph,nodeFactory,sceneOpaqueFboConfig);
```
emanuele3d: For this and other new methods below, please order them in the same order as they appear in initRenderGraph().

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -217,7 +217,41 @@ private void initRenderGraph() {
...
+
+        renderTaskListGenerator = new RenderTaskListGenerator();
+        List<Node> orderedNodes = renderGraph.getNodesInTopologicalOrder();
+        renderPipelineTaskList = renderTaskListGenerator.generateFrom(orderedNodes);
+    }
+
+    private void addShadowMapNodes(RenderGraph renderGraph, NodeFactory nodeFactory) {
```
emanuele3d: I just had a look at the NodeFactory code to see if we could make its createInstance methods static, to avoid having to pass it as an argument but no, the node factory instance stores a context which is needed for injection. 

For your and @vampcat information however, and having discussed the matter with immortius, in general we want to move away from annotation-based injection and into passing contexts and potentially subcontexts to constructors. Something for you guys to keep on the backburner and adopt whenever possible.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -239,14 +274,53 @@ private void initRenderGraph() {
...
+        renderGraph.addNode(deferredPointLightsNode, "DeferredPointLightsNode");
+
+        Node deferredMainLightNode = nodeFactory.createInstance(DeferredMainLightNode.class);
+        renderGraph.addNode(deferredMainLightNode, "deferredMainLightNode");
+
+        Node applyDeferredLightingNode = nodeFactory.createInstance(ApplyDeferredLightingNode.class);
+        renderGraph.addNode(applyDeferredLightingNode, "applyDeferredLightingNode");
+
+        Node chunksRefractiveReflectiveNode = nodeFactory.createInstance(RefractiveReflectiveBlocksNode.class);
+        renderGraph.addNode(chunksRefractiveReflectiveNode, "chunksRefractiveReflectiveNode");
```
emanuele3d: This is more of a world node but needs to stay after lighting so that lit, refracted underwater objects are visible under the surface. I'm guessing it would be neater in its own method even though it's just two lines.

hybrideagle:
My bad for taking so long, exam week just ended. @vampcat seems to have made quite a few changes, so I'll just wipe my changes and rewrite them. Should have that up in a few hours.

vampcat:
Yeah, sorry for making you re-write the entire commit ^.^

Cervator:
Bump - @hybrideagle any update? Should we close this in favor of a new PR?

I hope your exams went well! :-)

hybrideagle:
@Cervator Yeah, we should probably close this. Fixing the merge conflicts seems completely impossible at this point, so it's kinda pointless anyway.

Cervator:
Ok, thanks for the reply @hybrideagle ! I hope your assignments go well and we see a new PR soon :-)

PR #2884, opened 2017-04-05
Ignore out of bounds exception in NearestSortingList, Closes #2742

### LINKED ISSUE #2742: ArrayIndexOutOfBoundsException in NearestSortingList
Maybe an issue in the sort logic or an issue in the add-remove logic of the list?

### What actually happened
Random Multiplayer Testing with the gci content and NeoTTA with ~6 players.

### How to reproduce

I have no idea :) It occured only once so far.

### Log details
Only this one, all other lines were unrelated (chat events, destroyed entities, multiple chunks loaded)
19:58:01.762 [Timer-1] ERROR o.t.r.logic.NearestSortingList - Uncaught exception in sorting thread: java.lang.ArrayIndexOutOfBoundsException: 200

oniatus:
### Contains

Possible fix for #2742.
Destroyed entities were not visible in the sorting thread, causing the lookup to fail.

### How to test

Start with core gameplay and use the large railgun while walking forward/diagonal downwards and collect the destroyed blocks (should lag a lot, make sure to assign enough memory. ~2,5GB worked for me).
Without the PR, you should get the exception from https://github.com/MovingBlocks/Terasology/issues/2742#issuecomment-271735663 after some time.
With the PR, you should get a log message like `10:15:08.759 [Timer-1] WARN  o.t.r.logic.NearestSortingList - Entities destroyed during sorting process. Sorting is skipped this round.` at a similar time.

We should discuss if the `ComponentTable` has to be thread safe in the first place, before merging.
If the entire environment is expected to work only in the main thread, this may lower the overall performance and we could just swallow the exception in the sorting code.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/logic/NearestSortingList.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/NearestSortingList.java
@@ -40,7 +40,7 @@ import java.util.TimerTask;
  * sorting of the entities. This class only tries to keep the elements sorted,
  * but does not guarantee it.
  * <br><br>
- * It it therefor use full for graphics purposes, to keep track of the nearest
+ * It it therefore use full for graphics purposes, to keep track of the nearest
  * entities to draw.
  *
  */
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Tests out and can confirm easy error before, can't reproduce after :-)

No opinion on the threading, maybe @immortius can advise.

flo:
This would only be a partial support for thread savety so I would tent to say that we shouldn't try to make it thread save and just state that it is not thread save.

Also wrapping TLongObjectHashMap with a synchronized map defeats the purpose of that map as then it would become necessary to create Long objects just to pass a long value through the synchronized map. TLongObjectHashMap can work with the primitive type long.

oniatus:
In this case i would update the docs and just ignore this one exception type. As it occurs very rarely and has little to none visible impact on the game.

GooeyHub:

Cervator:
With the updated commit is this one OK to merge?

oniatus:
It's fine for me.

Cervator:
Gave it another quick test and merged - thanks! :-)

PR #2886, opened 2017-04-07
Remove testing line from Radial UI

jellysnake:
This really should have been removed earlier.

Also change the sections to use bindings following discussion with rzats and immortius


GooeyHub:

Cervator:
Works for me! And by works I mean it compiles, which is good enough without anything actually using the radial menu yet :-)

Thanks!

PR #2895, opened 2017-04-12
Task removal

vampcat:
This PR removes the Task classes from the Rendering Pipeline, instead making the Nodes and StateChanges extend the RenderingPipelineTask interface, simplifying the code.

Part of the Rendering Pipeline Refactor Movement
#RPRM

How to test: Run the game, and ensure no weird NPEs, no rendering artifacts, and everything appearing as you would expect it to.


vampcat:
Tagging @emanuele3d and @tdgunes for review!

Manu, I have tried my best to make the commits sensible and atomic, but I know I somehow messed up. Please, do tell me how I can make them more organized in the future (mainly the second commit, which is the biggest of them all).

GooeyHub:

GooeyHub:

emanuele3d:
Started review but can't complete it right now. I should go through it in about 7 hours.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTexture.java
```diff
@@ -111,40 +104,16 @@ public StateChange getDefaultInstance() {
...
+                textureSlot, textureId, material.getUrn().toString(), materialParameter);
+    }
 
-        @Override
-        public String toString() {
-            return String.format("%30s: slot %s, texture %s, material %s, parameter %s", this.getClass().getSimpleName(),
-                    textureSlot, textureId, material.getUrn().toString(), materialParameter);
-        }
+    @Override
+    public void process() {
```
emanuele3d: If you had listed the process() method before the toString() method github wouldn't be confused and wouldn't have flagged toString() as changed. This is something you should be able to notice also in the IDE. Again, one of the goal is to reduce the number of changed lines for the reviewer to go through as it all adds up. In this case I had to figure out what changed and it took me a minute or so to be sure nothing has.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
```diff
@@ -76,33 +76,26 @@ public SetInputTextureFromFBO(int textureSlot, ResourceUrn fboUrn, FboTexturesTy
...
+        this.inputFbo = fboManager.get(fboUrn);
+        this.material = getMaterial(materialUrn);
+
+        // Cheeky way to initialise textureId
+        update();
+        fboManager.subscribe(this);
     }
 
-    private SetInputTextureFromFBO(int textureSlot, ResourceUrn materialUrn, String parameterName) {
+    private SetInputTextureFromFBO(int textureSlot, ResourceUrn materialURN, String parameterName) {
```
emanuele3d: Whoops? What happened here? Reverting to the old convention?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
```diff
@@ -175,55 +168,14 @@ public boolean equals(Object other) {
...
-        @Override
-        public String toString() {
-            return String.format("%30s: slot %s, texture %s, material %s, parameter %s", this.getClass().getSimpleName(),
-                    textureSlot, textureId, material.getUrn().toString(), shaderParameterName);
-        }
+    @Override
+    public void process() {
+        glActiveTexture(GL_TEXTURE0 + textureSlot);
+        glBindTexture(GL_TEXTURE_2D, textureId);
+        material.setInt(parameterName, textureSlot, true);
```
emanuele3d: Similar to a previous comment (but not yet to be tackled), this is a class that mixes opengl state changes with material state changes. Perhaps we can add a TODO in both cases? I.e.:

// TODO: split this class into two - one for opengl's global state change and one for the specific material state change.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetStateParameter.java
```diff
@@ -48,37 +50,24 @@ public int hashCode() {
...
     @Override
     public boolean equals(Object obj) {
-        return (obj instanceof SetStateParameter) && this.enabled == ((SetStateParameter) obj).isEnabled() && this.glParameter == ((SetStateParameter) obj).glParameter;
+        return (obj instanceof SetStateParameter) && this.enabled == ((SetStateParameter) obj).enabled && this.glParameter == ((SetStateParameter) obj).glParameter;
     }
 
-    public boolean isEnabled() {
-        return enabled;
+    private String getStatus() {
+        return enabled? "Enabled": "Disabled";
```
emanuele3d: Space after "enabled" please.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetStateParameter.java
```diff
@@ -48,37 +50,24 @@ public int hashCode() {
...
-
-    public String getStatus() {
-        String status = "disabled";
+    @Override
+    public void process() {
         if (enabled) {
-            status = "enabled";
+            glEnable(glParameter);
+        } else {
+            glDisable(glParameter);
```
emanuele3d: Similar to previous comments this class now needs to be split into an `EnableStateParameter `and `DisableStateParameter`. As the defaults vary from parameter to parameter you'll probably have to extend a common class which could be the existing `SetStateParameter`.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -35,12 +34,17 @@
...
-    private SetViewportToSizeOfTask task;
     private ResourceUrn fboName;
+    private int fboWidth;
+    private int fboHeight;
 
     public SetViewportToSizeOf(ResourceUrn fboName, BaseFBOsManager frameBuffersManager) {
         this.fboManager = frameBuffersManager;
         this.fboName = fboName;
+
+        // Cheeky way to initialise fboWidth, fboHeight
```
emanuele3d: This line used to be beside update() for good reason. Now it seems to related to both lines below. I think something similar happens in another class, earlier.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
```diff
@@ -29,14 +28,17 @@
...
     private static SetWireframe defaultInstance = new SetWireframe(false);
-    private static SetWireframeTask enablingTask;
-    private static SetWireframeTask disablingTask;
 
     private boolean enabled;
+    private int mode;
 
     public SetWireframe(boolean enabled) {
         this.enabled = enabled;
+        this.mode = enabled? ENABLED_MODE: DISABLED_MODE;
```
emanuele3d: Space after "enabled" please.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
```diff
@@ -29,14 +28,17 @@
  * TODO: Add javadocs
  */
 public final class SetWireframe implements StateChange {
+    private static final int ENABLED_MODE = GL_LINE;
+    private static final int DISABLED_MODE = GL_FILL;
```
emanuele3d: I am not convinced about these two variables - they made more sense in the task perhaps. Here I'd recommend to use the opengl constants directly.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetWireframe.java
```diff
@@ -73,41 +60,17 @@ public boolean isEnabled() {
         return enabled;
     }
 
+    private String getStatus() {
+        return enabled? "Enabled": "Disabled";
```
emanuele3d: Space after "enabled" please.

engine-tests/src/test/java/org/terasology/rendering/dag/SetName.java
```diff
@@ -58,4 +49,7 @@ public String toString() {
     public String getName() {
         return name;
     }
+
+    @Override
```
emanuele3d: If this method took the place of the removed generateTask() method this diff would be slightly easier to browse. No big deal though.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
```diff
@@ -19,34 +19,40 @@
...
 
 import static org.lwjgl.opengl.EXTFramebufferObject.GL_FRAMEBUFFER_EXT;
 import static org.lwjgl.opengl.EXTFramebufferObject.glBindFramebufferEXT;
 
 /**
- * TODO: Add javadocs
+ * Binds the given FBO to the GL_FRAMEBUFFER target, making it the read framebuffer target as well as
+ * the draw framebuffer target.
+ *
+ * This is useful for operations such as off-screen rendering.
```
vampcat: Sounds reasonably, should be as easy as changing a single flag.. I'll take care of it in a separate commit.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthWriting.java
```diff
@@ -61,61 +55,26 @@ public StateChange getDefaultInstance() {
...
-
-    public String getStatus() {
-        String status = "disabled";
-        if (enabled) {
-            status = "enabled";
-        }
-
-        return status;
+    private String getStatus() {
+        return enabled? "Enabled": "Disabled";
```
vampcat: The reason I removed this from here is, the isEnabled() you talk about will be a generic Node/StateChange funciton, not a DisableDepthWriting StateChange function.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFacesToCull.java
```diff
@@ -30,12 +29,12 @@
...
  */
 public final class SetFacesToCull implements StateChange {
-
-    private static SetFacesToCull defaultInstance = new SetFacesToCull(GL_BACK); // also specified in OpenGL documentation
     private static Map<Integer, String> modeNameMap = ImmutableMap.of(GL_BACK, "GL_BACK",
             GL_FRONT, "GL_FRONT",
             GL_FRONT_AND_BACK, "GL_FRONT_AND_BACK");
-    private SetFacesToCullTask task;
+
+    private static SetFacesToCull defaultInstance = new SetFacesToCull(GL_BACK); // also specified in OpenGL documentation
```
vampcat: ..I just moved code around, no extra information added from my side :P
But I'll take care of this comment! :D

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFBO.java
vampcat: I swear I don't know which elf did this.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
```diff
@@ -19,34 +19,40 @@
...
     private static final ResourceUrn DEFAULT_FRAME_BUFFER_URN = new ResourceUrn("engine:display");
 
-    private static BindFBO defaultInstance = new BindFBO(DEFAULT_FRAME_BUFFER_URN);
+    private static BindFBO defaultInstance = new BindFBO();
 
-    private BindFBOTask task;
-    private BaseFBOsManager fboManager;
     private ResourceUrn fboName;
+    private BaseFBOsManager fboManager;
+    private int fboId;
```
vampcat: The change I made was to make the structure more consistent. Agreed, it should have been in a separate commit.. but it will happen nonetheless. Do you mind if I skip the reverting this one time, and make sure to do it in a separate commit next time onwards?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
```diff
@@ -88,36 +76,16 @@ public boolean equals(Object obj) {
...
-            this.fboId = fboId;
-        }
-
-        @Override
-        public String toString() {
-            return String.format("%30s: %s (fboId:%s)", this.getClass().getSimpleName(), fboName, fboId);
-        }
+    @Override
+    public void process() {
+        glBindFramebufferEXT(GL_FRAMEBUFFER_EXT, fboId);
```
vampcat: Strangely enough, there is a `EXTFramebufferObject.GL_FRAMEBUFFER_EXT` constant, but no `EXTFramebufferObject.GL_DRAW_FRAMEBUFFER_EXT` constant.
I'll skip this for now?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
vampcat: Yeah, I don't know what was going through my head when I did this.. :sweat:

emanuele3d:
@vampcat: I have noticed a number of replies to individual comments, but please let me know when you'd like me to reply to them. Another way to do it is to thumb-up every comment you are happy to take on board as it is. This way, between your thumbs up and your actual replies, I'll know you have gone through all the comments and it might be time for me to reply to any outstanding issue. 

Ideally any comment to a PR should have a thumb-up or a discussion ending with a thumb-up, so that we are clear what has been addressed and what hasn't been.

vampcat:
My usual work flow involves taking in every single comment, and just commenting on the ones i don't understand or don't fully agree with. Either way, I'd be happy for you to go through my comments now, since I have the rest of them addressed in my local codebase.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthTest.java
```diff
@@ -32,10 +29,8 @@
  */
 public final class DisableDepthTest extends SetStateParameter {
     private static final int PARAMETER = GL_DEPTH_TEST;
```
vampcat: In line with one of your previous comment, should I delete this variable too?

(Just fyi, it's used only once in the constructor to pass the parameter to SetStateChange's constructor)

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthWriting.java
```diff
@@ -61,61 +55,26 @@ public StateChange getDefaultInstance() {
...
-            return enablingTask;
-        } else {
-            return disablingTask;
-        }
-    }
-
-    @Override
     public boolean equals(Object obj) {
-        return (obj instanceof DisableDepthWriting) && this.enabled == ((DisableDepthWriting) obj).isEnabled();
+        return (obj instanceof DisableDepthWriting) && this.enabled == ((DisableDepthWriting) obj).enabled;
```
vampcat: Noted, I'll be more careful in the future :)

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableFaceCulling.java
```diff
@@ -54,23 +49,11 @@ public EnableFaceCulling() {
     }
 
     private EnableFaceCulling(boolean enabled) {
-        super(GL_CULL_FACE, enabled);
-        disablingTask = new DisableStateParameterTask(PARAMETER, PARAMETER_NAME);
-        enablingTask = new EnableStateParameterTask(PARAMETER, PARAMETER_NAME);
+        super(PARAMETER, enabled);
```
vampcat: Can you explain what you mean?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
emanuele3d: It isn't in the scope of this PR, so a separate PR would make sense. That been said, it is probably a single line change in the code and perhaps some rewriting of the javadocs, which needs to be done anyway. So, it wouldn't be horribly out of place in this PR as its own commit. But I'm not fussed either way.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
emanuele3d: Bummer. GL_DRAW_FRAMEBUFFER is only available from OpenGL 3.0. :disappointed: Never mind then. Let's just add:

// TODO: change the target argument to GL_DRAW_FRAMEBUFFER when we switch to OpenGL 3.0 and beyond.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFBO.java
emanuele3d: Note: previous comment is obsolete: GL_DRAW_FRAMEBUFFER is only available from OpenGL 3.0.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthWriting.java
emanuele3d: Ah, ok, fair enough. Is this already happening in this PR?

vampcat: No, I'll need a PR or two more for that to happen.

vampcat:
@emanuele3d I've pushed 3 commits (yes, 3, not 2 - one is above your last comment). The first is fairly trivial, and should be easy enough to skim through. The second one is the proposed rework of TaskListGenerator (for the List Based Pipeline), and should be satisfactory for the time being (I plan on making even more changes and making it capable of handling even more types of set-reset class pairs, like LookThrough-LookThroughNormalized-LookThroughDefault, but that's for a future PR) and the third one is a proof-of-concept that uses the new TaskListGenerator to use a set-reset StateChange pair.

If you're fine with these changes, I'll push the 5-6 other commits I have waiting in my local branch for other set-reset pair splits.

P.S. The code is much clearer after the split, so kudos for coming up with the idea!
( Even though it made me re-write the TaskListGenerator :P )

GooeyHub:

vampcat:
Also, don't worry about failing builds - it's an issue with the Tests, which expect a VERY rigid order, so my shifting of state sets/resets around has messed it up.
I'll hack the tests for now after you've approve these changes, and once I get a bit of time, write some meaningful tests that aren't quite as.. hardcoded and rigid.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -123,49 +120,44 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
...
-                    for (StateChange currentStateReset : node.getDesiredStateResets()) {
-                        taskList.add(currentStateReset);
-                        persistentStateChanges.remove(currentStateReset.getClass());
-                    }
+                // Finally add all the desirable state changes for this iteration into the task list.
+                // The delay ensures that the state resets take place before new state sets, which can
+                // be necessary when dealing with State Changes like LookThrough and LookThoughNormalized
+                // that are essentially the same, but not according to the TaskListGenerator.
+                for (StateChange stateChange : stateChangesToAdd) {
+                    taskList.add(stateChange);
```
vampcat: Ah, excuse my java-noobness :P

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableStateParameter.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+package org.terasology.rendering.dag.stateChanges;
+
+import static org.lwjgl.opengl.GL11.glEnable;
+
+/**
+ * TODO: Add javadocs
+ */
+abstract class EnableStateParameter extends SetStateParameter {
```
vampcat: Because of the process() method.

The EnableStateParameter calls glEnable(), and DisableStateParameter calls glDisable(). The best I can do is remove SetStateParameter (I wanted to do it either way..) to simplify things a little.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableBlending.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.lwjgl.opengl.GL11.GL_BLEND;
+import org.terasology.rendering.dag.StateChange;
+
+/**
+ * Instances of this class enable OpenGL's blending, i.e. to render transparent objects or to make
+ * a composite of different images.
+ *
+ * See also StateChange implementation SetBlendFunction to set the source and destination factors
+ * used by the blending process.
+ */
```
vampcat: Noted, I'll take care of this.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetStateParameter.java
```diff
@@ -28,46 +28,20 @@
...
-    public boolean equals(Object obj) {
-        return (obj instanceof SetStateParameter) && this.enabled == ((SetStateParameter) obj).enabled && this.glParameter == ((SetStateParameter) obj).glParameter;
-    }
-
-    private String getStatus() {
-        return enabled ? "Enabled" : "Disabled";
-    }
-
-    @Override
-    public String toString() {
```
vampcat: getStatus was a public accessor not used anywhere, SetStateParameter still has a toString() (although much more generic that only prints class name, looks quite pretty in the taskList) and equals.. I don't know.

I'll take care of this one too.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableMaterial.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.terasology.registry.CoreRegistry;
+import org.terasology.rendering.ShaderManager;
+import org.terasology.rendering.dag.StateChange;
+
+/**
+ * TODO: Add javadocs
+ */
+public final class DisableMaterial implements StateChange {
+    private ShaderManager shaderManager = CoreRegistry.get(ShaderManager.class);
```
vampcat: Hmm, this is a tricky one. The EnableMaterial() originally got it from the CoreRegistry too, and since the DisableMaterial is a static object in class EnableMaterial, it's kinda hard..

Right now, I don't think I can do it easily. Just adding a comment for now, but I'll try to make the entire DAG stop relying on CoreRegistry in the near future.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthTest.java
```diff
@@ -19,13 +19,13 @@
...
  *
- * Notice that by default OpenGL has depth testing disabled but in Terasology's defaults it is enabled,
- * as depth testing is used by many nodes. It's important then to use this StateChange so that each
- * node requiring no depth testing disables it and re-enables it after the process() method has been
- * executed.
+ * OpenGL Default: DisableDepthTest
+ * Terasology Default: EnableDepthTest
+ *
+ * Type: Reset
+ * Corresponding Set: DisableDepthTest
```
emanuele3d: This javadoc highlight the difficulties with terse javadocs: it's easy to get confused. 

Specifically in this case the "type" of this state change is actually "set" as it sets the depth testing as disabled. The "reset" state change resets it to its (Terasology) default - depth testing is enabled. 

So, I'd write:

```
/**
* Disables OpenGL depth testing.
*
* Notice that OpenGL has depth testing disabled by default. Terasology however enables it by default as depth 
* testing is used in many nodes. It's important then to use this StateChange if a node requires the OpenGL default.
*/
```

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableBlending.java
```diff
@@ -19,11 +19,16 @@
 import org.terasology.rendering.dag.StateChange;
 
 /**
- * Instances of this class enable OpenGL's blending, i.e. to render transparent objects or to make
- * a composite of different images.
+ * Enables blending in the OpenGL pipeline.
```
emanuele3d: I'd just write `Enables OpenGL Blending.` The term pipeline in our context can be a bit ambiguous I feel.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableBlending.java
```diff
@@ -19,11 +19,16 @@
...
+ * Can be used to render transparent objects or to make a composite of different images.
```
emanuele3d: I'd write:

```
* Blending can be used to render transparent objects on top of those already rendered in a buffer 
* or to make a composite of different images.
```

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableBlending.java
```diff
@@ -19,11 +19,16 @@
...
  *
- * See also StateChange implementation SetBlendFunction to set the source and destination factors
- * used by the blending process.
+ * Note that SetBlendFunction can be used to set the source and destination factors used by the blending process.
```
emanuele3d: I like my wording better, please revert.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableBlending.java
```diff
@@ -19,11 +19,16 @@
...
+ *
+ * OpenGL Default: DisableBlending
+ * Terasology Default: DisableBlending
```
emanuele3d: The term "DisableBlending" is a reference to a specific state change class, not OpenGL's nor Terasology's defaults. Keeping your format I'd write:

```
* OpenGL default: blending disabled
* Terasology default: blending disabled
```

In practice I'd simply write: 

`* By default both OpenGL and Terasology keep blending disabled as it is a relatively expensive operation.`

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableDepthTest.java
```diff
@@ -20,9 +20,14 @@
...
- * Instances of this enable OpenGL's depth test, which allows fragments that fail the test to be discarded.
+ * Enables depth testing in OpenGL pipeline.
+ * This allows the processing of fragments that are hidden behind another fragment, and would normally fail the test and therefore be discarded.
  *
- * This StateChange is used to reset the effects of DisableDepthTest.
+ * OpenGL Default: DisableDepthTest
+ * Terasology Default: EnableDepthTest
+ *
+ * Type: Set
+ * Corresponding Reset: EnableDepthTest
```
emanuele3d: Similarly to a previous comment, this is actually a reset class given that Terasology's default is for the depth test to be enabled.

emanuele3d:
@vampcat: please notice that I make a number of comments that github collapses into "outdated commits", even though they are still relevant. Make sure to read them all!

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableDepthTest.java
vampcat: This was a copy-pasta from the DisableDepthTest class.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableStateParameter.java
emanuele3d: Ah yes, of course. We didn't have this problem earlier because the task classes were segregating the glEnable/glDisable calls. Ok, please go ahead with the removal of the SetStateParameter class in favor of two Enable/DisableStateParameter classes.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -102,41 +115,41 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
...
                     // Marker tasks just add a dividing line to the logger output
                     taskList.add(new MarkerTask(node.getClass().getSimpleName()));
                     enabledNodes++; // we count them only for statistical purposes
                     potentialTasks += 2 * node.getDesiredStateChanges().size() + 1;
                 }
 
-                // generating tasks for the desired state changes
+                // Examine the StateChanges requested by current Node, and insert/update them in persistentStateChanges.
                 for (StateChange currentStateChange : node.getDesiredStateChanges()) {
+                    StateChange persistentStateChange = persistentStateChanges.get(currentStateChange.getClass());
```
vampcat: Regarding the first point: the persistentStateChange is used ONCE in the entire method, and that is here. It can technically be removed by making the next line `if (!currentStateChange.equals(persistentStateChange))` much longer, but I prefered not doing that. With that being said, I really don't think the type of a variable is the kind of redundant information that needs to be removed, at the expense of increasing the scope of a variable unneccesarily. When I originally encountered this code, I was confused why this variable was defined at top and not locally.
Can you please consider letting the type information remain here? Pretty please? :D

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -102,41 +115,41 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
...
                     persistentStateChangesRelevant.put(currentStateChange.getClass(), true);
                 }
 
-                // set the state change as relevant for this iteration
-                for (Iterator<Map.Entry<Class<?>, StateChange>> iterator = persistentStateChanges.entrySet().iterator(); iterator.hasNext(); ) {
-                    Map.Entry<Class<?>, StateChange> stateChange = iterator.next();
+                // Iterate through all StateChanges
+                Iterator<Map.Entry<Class<?>, StateChange>> iterator = persistentStateChanges.entrySet().iterator();
+                while (iterator.hasNext()) {
+                    Map.Entry<Class<?>, StateChange> entry = iterator.next();
```
vampcat: Not sure if Java supports that. I tried simply changing it, but the compiler started throwing errors later on in code. I think it's much easier to leave it like this.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -102,41 +115,41 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
...
-                    // set the state change as relevant for this iteration
+                    // Mark the StateChange as relevant for this iteration.
                     persistentStateChangesRelevant.put(currentStateChange.getClass(), true);
                 }
 
-                // set the state change as relevant for this iteration
-                for (Iterator<Map.Entry<Class<?>, StateChange>> iterator = persistentStateChanges.entrySet().iterator(); iterator.hasNext(); ) {
-                    Map.Entry<Class<?>, StateChange> stateChange = iterator.next();
+                // Iterate through all StateChanges
+                Iterator<Map.Entry<Class<?>, StateChange>> iterator = persistentStateChanges.entrySet().iterator();
```
vampcat: Same as last comment, since this is the only place this variable is used, can I leave this as it is?

I really don't want to go back to the C-days when all the variables had to be declared on top, overscoping them and making it hard to determine where they are used (provided they are).

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -102,41 +115,41 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
...
+                while (iterator.hasNext()) {
+                    Map.Entry<Class<?>, StateChange> entry = iterator.next();
 
-                    if (persistentStateChangesRelevant.get(stateChange.getKey())) {
-                        persistentStateChangesRelevant.put(stateChange.getKey(), false);
+                    if (persistentStateChangesRelevant.get(entry.getKey())) {
```
vampcat: `stateChangeAgainstClass` doesn't quite seem right to me. The Entry is <Class, StateChange>, so to me, `entry.getKey()` should be `className` or 'key` or something, and the entry.getValue() makes more sense as `stateChangeAgainstClass`.
Thoughts?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthTest.java
vampcat: I cannot make sense of my own javadocs now 😓

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
emanuele3d: I am aware of the fact that this is the only place this variable is used. And of course in most cases I would agree with you. But this method is the core of the generator. It is not an easy piece of code to understand and I want to give the reader the best chance to understand it. 

So, stop being dogmatic and keep in mind that readability is a very high priority in an open source project.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
emanuele3d: Alright, then let's give it the same treatment as other variables of this method: declare it early, well out of the loop(s) so that this line becomes just `entry = iterator.next();`

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
emanuele3d: Regarding getKey(): it's a tricky case. From my perspective the most important line is

`if (persistentStateChangesRelevant.get(stateChangeAgainstClass))) {`

that's where I want to be clear(er). By writing it this way reads closer to plain English. If we wrote:

`if (persistentStateChangesRelevant.get(aClass))) {`

the reader first would think we are trying to get(aClass), and then would go through the mental step of reminding himself he's dealing with a map and therefore aClass is the key. My emphasis on plain English is just a way to remove as many mental steps as possible. Given that this is not hot code we have some flexibility. 

Regarding .getValue(), it make sense to me.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFbo.java
```diff
@@ -32,6 +32,9 @@
  *
  * When this state change is reset opengl's default framebuffer (usually the display) is bound again.
  * Similarly, nodes that do not take advantage of this state change will normally write to the default framebuffer.
+ *
+ * Type: Set
+ * Corresponding Reset: UnbindFbo
```
emanuele3d: Can you respond on this one?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthTest.java
emanuele3d: No worries. That's what code reviews are for.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/DisableDepthWriting.java
```diff
@@ -38,6 +38,7 @@
      * two task instances frame the execution of a node's process() method unless they are deemed redundant,
      * i.e. because the upstream or downstream node also disables depth buffer writing.
      */
+    // TODO: Remove this method?
```
emanuele3d: It does look unnecessary. Perhaps it's something that made sense in earlier commits and just tagged along, unnoticed. Let's try and remove it.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFbo.java
emanuele3d: Actually, I just noticed you made a number of private classes in line with this comment. Never mind then!

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableBlending.java
```diff
@@ -19,20 +19,18 @@
...
- *
- * Type: Set
- * Corresponding Reset: DisableBlending
+ * By default both OpenGL and Terasology keep blending disabled as it is a relatively expensive operation.
  */
 public final class EnableBlending extends EnableStateParameter {
     private static StateChange defaultInstance = new DisableBlending();
 
+    // TODO: Do we really need to add such a description for every single StateChange?
+    // TODO: IMHO, we can just add the common parts to StateChange interface, and add class-specific info as needed.
```
emanuele3d: This kind of comment, sort-of directed to the reviewer, can be added as a comment on github rather than in code. I often add comments of my own shortly after I publish a PR, to point the attention of the reviewers to specific parts of the code, to illustrate concerns I might have and to make sure I get additional opinions about specific issues.

That been said, you make a good point here. Make a separate commit to show me what you mean in practice: keep it small, don't invest too much time on it just yet, i.e. modify the StateChange interface javadoc and the javadoc for _just_ this EnableBlending class. Only if we both like the changes we can then roll out the same javadoc format to other classes.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFbo.java
vampcat: Sorry, should have made it clear I had made changes according to your comment.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableFaceCulling.java
```diff
@@ -19,7 +19,10 @@
 import org.terasology.rendering.dag.StateChange;
 
 /**
- * Instances of this class enable OpenGL's Face Culling, i.e. to render only the inside faces of the skysphere.
+ * Enables OpenGL's Face Culling.
+ *
+ * This can be used to discard the fragments that face away from the camera, and can be used to improve
+ * the performance in certain situations like rendering only the inside faces of the skysphere.
```
emanuele3d: Hmmm. It's a good effort. The only problem is that it narrows the meaning of face culling a bit. I'd rewrite as:

```
* Face Culling is used to discard triangles facing away from the camera, reducing the total number of triangles
* to be processed. In normal circumstances the GL_BACK triangles are culled, but in some circumstances, i.e. inside
* a mesh such as the skysphere, the triangles facing the camera are the GL_BACK ones and the GL_FRONT triangles need
* to be culled instead. Use SetFacesToCull to set which triangles gets culled - by default GL_BACK.
```

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetBlendFunction.java
```diff
@@ -38,7 +38,7 @@
 import static org.lwjgl.opengl.GL11.glBlendFunc;
 
 /**
- * This StateChange generates the tasks that change and reset the blend function factors.
+ * Changes the blend function factors used by OpenGL.
```
emanuele3d: "Changes" -> "Sets"

I'd also add in the middle of this javadoc:

```
* See https://www.khronos.org/opengl/wiki/Blending for details
* Also see http://www.andersriggelsen.dk/glblendfunc.php to experiment with different factors.
```

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTexture.java
```diff
@@ -30,7 +30,7 @@
 // TODO: split this class into two - one for opengl's global state change and one for the specific material state change.
 
 /**
- * This StateChange generates the tasks that set and reset input textures.
+ * Sets the input textures.
```
emanuele3d: `* Sets a texture asset as the input for a material.`

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
vampcat: Actually, right now I'm using this:
`                    Class<?> key = entry.getKey();`
`                    if (!requestedStateChanges.contains(key)) {`
Does that seem good enough?
I don't want to have the `stateChangeAgainstClass`, which as you pointed out makes the code clearer, but is misleading at the same time.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableFaceCulling.java
vampcat: Using this as it is, with a small change:
`Face Culling is used to discard triangles facing -a particular direction-, reducing..`
to make things more generic.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableFaceCulling.java
emanuele3d: Hmmm... I see where you are aiming at. 

The problem I see is that "a particular direction" is very, very open. As in: the set of all unit vectors starting from the origin: it's a lot of them! Here we are dealing with directions that are orthonormal to a triangle. And there's only two of them possible in that context.

I can see how there might be reasons to discard triangles facing toward the camera, so perhaps my text was too narrow. See if you can find a wording that suggest what's the typical use (culling triangles facing away from the camera for optimization purposes) and what the alternative uses could be.

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -99,6 +99,12 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
 
         List<StateChange> stateChangesToAdd = Lists.newArrayList();
 
+        // The following variables have been declared here to make the code clearer.
+        // P.S. manu3d made me do it!
```
emanuele3d: :laughing:

engine/src/main/java/org/terasology/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -133,15 +139,16 @@ private void logIntermediateRendererListForDebugging(List<Node> orderedNodes) {
                 }
 
                 // Reset all the persistentStateChanges that are not requestedStateChanges.
-                Iterator<Map.Entry<Class<?>, StateChange>> iterator = persistentStateChanges.entrySet().iterator();
+                iterator = persistentStateChanges.entrySet().iterator();
                 while (iterator.hasNext()) {
-                    Map.Entry<Class<?>, StateChange> entry = iterator.next();
+                    entry = iterator.next();
                     Class<?> key = entry.getKey();
+                    StateChange stateChangeAgainstKey = entry.getValue();
```
emanuele3d: Ok, I think we are getting to a compromise here. Please replace the previous two lines with:

```
key = entry.getKey();
stateChange = entry.getValue();
```

by moving the variable declarations at the beginning of the method, as you did with other variables, and I'll be ok with it.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableFaceCulling.java
```diff
@@ -21,13 +21,13 @@
 /**
  * Enables OpenGL's Face Culling.
  *
- * This can be used to discard the fragments that face away from the camera, and can be used to improve
- * the performance in certain situations like rendering only the inside faces of the skysphere.
+ * Face Culling is used to discard triangles facing a particular direction, reducing the total number of triangles
+ * to be processed. In normal circumstances the GL_BACK triangles are culled, but in some circumstances, i.e. inside
+ * a mesh such as the skysphere, the triangles facing the camera are the GL_BACK ones and the GL_FRONT triangles need
+ * to be culled instead. Use SetFacesToCull to set which triangles gets culled - by default GL_BACK.
```
emanuele3d: Actually, disregard previous comment on this: this javadoc is ok as it is.

engine/src/main/java/org/terasology/rendering/dag/StateChange.java
```diff
@@ -16,7 +16,14 @@
 package org.terasology.rendering.dag;
 
 /**
- * TODO: Add javadocs
+ * A StateChange denotes a change in the OpenGL state, that is required by a particular Node.
```
emanuele3d: I'd rewrite:

`A StateChange instance encapsulate a change in the OpenGL state that is required by one or more consecutive nodes.`

engine/src/main/java/org/terasology/rendering/dag/StateChange.java
```diff
@@ -16,7 +16,14 @@
...
+ *
+ * Any of the concrete classes extending the StateChange interface can be used in a node's initialise() method in the form:
+ * addDesiredStateChange(new StateChangeImplementation());
```
emanuele3d: I like this part, very clear.

engine/src/main/java/org/terasology/rendering/dag/StateChange.java
```diff
@@ -16,7 +16,14 @@
...
+ *
+ * This triggers the inclusion of an StateSet and a StateReset instance in the rendering task list.
```
emanuele3d: I'd rewrite as:

```
* Each state change instance is included for processing in the task list, alongside a second state change 
* called the "default instance" resetting the state back to its OpenGL or Terasology default. 
```

engine/src/main/java/org/terasology/rendering/dag/StateChange.java
```diff
@@ -16,7 +16,14 @@
...
+ * These two task instances frame the execution of a node's process() method unless they are deemed redundant
+ * by the RenderTaskListGenerator because the upstream or downstream node also requires the exact same StateChange.
```
emanuele3d: I'd rewrite as:

```
* These two state change instances effectively become tasks and frame the execution of a node's process() method unless they are deemed redundant by the RenderTaskListGenerator.
* State changes are deemed redundant when the upstream and/or downstream nodes require the exact same StateChange: no point in sending the same opengl directive to the GPU multiple times if the opengl state doesn't change as a result.
```

engine/src/main/java/org/terasology/rendering/dag/stateChanges/EnableBlending.java
```diff
@@ -29,22 +29,6 @@
...
-     * addDesiredStateChange(new EnableBlending());
-     *
-     * This trigger the inclusion of an EnableStateParameterTask instance and a DisableStateParameterTask instance
-     * in the rendering task list, each instance enabling/disabling respectively the GL_BLEND mode. The
-     * two task instance frame the execution of a node's process() method unless they are deemed redundant,
-     * i.e. because the upstream or downstream node also enables blending.
-     *
-     * See also StateChange implementation SetBlendFunction to set the source and destination factors
-     * used by the blending process.
-     */
```
emanuele3d: I see what you mean with this changes, thank you for providing a sample. 

I still think we should have a short constructor javadoc along these lines:

```
/**
Constructor, to be used in the initialise() method of a node:

E.g.   `addDesiredStateChange(new EnableBlending());`
*/
```

In this particular case it's fairly trivial, but there will be state changes with constructor having arguments: in those cases providing an example will be more helpful. 

Anyway, please go ahead with updating/writing javadocs for all the touched or new classes, along these lines.

GooeyHub:

Cervator:
Oh hey! My turn to shine! ... *clicks button* woo I helped! :D

Well, merged and tested locally and all that too. Looks good! Thanks a lot both :-)

(feel free to prod me on Slack if I don't notice pings on GitHub right away - so many notification emails)

PR #2897, opened 2017-04-12
CopyImageToHMDNode Constructor Rework

vampcat:
This PR prevents the creation of 2 unnecessary buffers for most casual players, the leftEye FBO and the rightEye FBO, which are used for HMD (Head Mounted Displays (VR)).

~~However, this PR is probably not complete, as we might want to consider re-working the ConditionDependentNodes. Currently, ConditionDependentNode::checkConditions() returns true if the state of Node was toggled, and false otherwise. Do we really need a return value, since we can just query it with Node.isEnabled() in the calling function (which returns the enabled state, which might be of more interest than toggleState, depending on the particular case)?~~

Pinging @indianajohn and @emanuele3d for review.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToHMDNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToHMDNode.java
@@ -62,18 +62,18 @@ public class CopyImageToHMDNode extends ConditionDependentNode implements FBOMan
     public CopyImageToHMDNode(Context context) {
         super(context);
 
-        worldRenderer = context.get(WorldRenderer.class);
-
         vrProvider = context.get(OpenVRProvider.class);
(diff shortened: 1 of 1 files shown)


GooeyHub:

vampcat:
Okay, I'll change the description for this PR from a WIP to a finished one, and create a new PR for the changes I think we can make.

emanuele3d:
@Cervator: this PR is good to GOOOOOOOOO!

@indianajohn: thank you for your quick response/review!

Cervator:
Thanks for the work as usual all :-)

I'm finally set up on the laptop I brought with me on my business trip, and am ready to start merging stuff again. Based on Slack logs I think some pending PR may be getting held back? But based on the contents of this one it doesn't seem like the one being talked about ...

So: merged!

PR #2898, opened 2017-04-12
Added gson to whitelist

nihal111:
This is linked to https://github.com/Terasology/Books/pull/14
The referenced PR makes a change to the Books module such that recipes (icon images) are now loaded dynamically. The book's page contains a mixture of text and recipes.
Such a page looks like this: 

> Following are some recipes <recipe **{\"blockIngredients\": 2, \"itemIngredients\": [\"WoodAndStone:stone\", \"WoodAndStone:stone\"], \"blockResult\": null, \"itemResult\": \"WoodAndStone:toolStone\", \"resultCount\": 1}**><l><l><l>Some More Text<l><l><recipe **{\"blockIngredients\": 2, \"itemIngredients\": [\"WoodAndStone:stone\", \"WoodAndStone:stone\"], \"blockResult\": null, \"itemResult\": \"WoodAndStone:toolStone\", \"resultCount\": 1}**>More Text

The recipe tags are extracted and the content is parsed as JSON. This is used to dynamically create new recipes (editable books).
Gson is required for this purpose of parsing and needs to be added to the whitelist.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
+++ b/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
@@ -53,6 +53,7 @@ public final class ExternalApiWhitelist {
             .add("com.google.common.math")
             .add("com.google.common.primitives")
             .add("com.google.common.util.concurrent")
+            .add("com.google.gson")
             .add("gnu.trove")
             .add("gnu.trove.decorator")
             .add("gnu.trove.function")


GooeyHub:

SoniEx2:
Is this really needed? Wouldn't it be better to allow GSON to be a module?

Cervator:
IMHO it is a fair concern to double check that nothing dangerous happens if we add GSON to the whitelist. Anybody want to research/comment on that one?

On the other hand the JSON bundled via module thing ... *maybe* that would make sense one day, or maybe instead allow modules to declare external dependencies via `module.txt` (still subject to the whitelist). But I think dealing with that level of complexity is probably beyond the scope of this PR (especially keeping two segregated copies)

flo:
I would recommend to use Prefabs to represent the recipes and then reference those prefabs by name in the translatable text. Makes it more readable and makes it use the standard way in terasology to represent data

nihal111:
@flo Sound like a good idea. Closing this PR in an attempt to do this the prefab way.

Issue #2872, opened 2017-03-31
Allow image scaling in UIImage


nihal111:
As of now, if the layout parameters i.e height and width are set for a UIImage object, the image does resize, but only as much as it can without breaking its aspect ratio.

That is if I have a 100x100 pixel image. I can easily scale it to say, 250x250 or 23x23. But I cannot scale it to 300x200 (i.e. stretch the image). In such a case the image would simply resize to 200x200, the maximum it can without breaking its aspect ratio.

This was found while working on https://github.com/Terasology/Books/blob/master/assets/ui/bookScreen.ui in an attempt to increase the height of the cover beyond what its aspect ratio allowed.

PR #2899, opened 2017-04-13
Added a Scaled Image widget

vampcat:
Adds a new UIImageScaled NUI widget, that allows arbitrary scaling of images regardless of ScaleMode set in the skin.

How to test: Change any UI file to use UIImageScaled instead of UIImage, and change the width/height out of propotion and watch it get stretched.

There are other ways to do this (most notably creating a new skin), so this is more of a test.

Created in response to issue #2872.


GooeyHub:

Issue #2651, opened 2016-12-08
Crouching doesn't properly uncrouch in multiplayer


Cervator:
When crouching in multiplayer everything is fine *if you're the host* - but if you are a connected client and attempt to crouch you sort of get stuck vibrating slightly and rapidly up and down, even long after no longer crouching. Something in the process does not seem to terminate correctly when purely a client. Maybe you get stuck in a crouch/uncrouch loop.

You can "fix" the issue temporarily by flying / ghosting as that gets you ungrounded, at which point you are incapable of crouching. But the issue returns after you land and takes logging out then back in to solve permanently.

gianluca-nitti:
For reference, here is a gif showing the bug: http://imgur.com/a/K83Jn

oniatus:
Closing as proposed in https://github.com/MovingBlocks/Terasology/pull/2961#issuecomment-341993601
(Issue should be fixed with the PR)
Follow up issue is #1075

PR #2904, opened 2017-04-17
Crouching mode change in server/client correctly

ThompsonTyler:
### Contains

Begins to solve #2651

Something I identified when doing debugging was that the KineticCharacterMover(What is used to allow players to simulate across multiple clients/server) did not actually have any knowledge of crouching. I made an addition that would allow for crouching to pass, by adding a parameter to the CharacterMoveInputEvent.

This properly allows for the server and clients to know each other's actual mode and state correctly(crouching, walking, climbing, etc). But the visuals and actual representation of moving is beyond my knowledge as it is not directly tied to the mode of the player and therefore I am currently unable to solve the visual glitches.

The problem for the visual glitches can be seen if you look at the states received  from the server in the onCharacterStateReceived in ClientCharacterPredictionSystem. The current Mover doesn't correctly translate the player's position for the state based on the crouch and therefore the client is constantly translating between it's intended location(the crouched position), and the server's knowledge(in crouch, but location wise still standing). This results in the server as treating the clients as always standing visually(and then physically if they go to do any action while crouching and possibly causing desync issues).

### How to test

Addition of a few new unit tests for the new CharacterMoveInputEvent/KineticCharacterMover.

Inside of localPlayerSystem have a debug for outputting the character's movementMode, then also in ServerCharacterPredicitionSystem output the movementMode at each update for each entity. Verify the matching for both client and servers. (Let me know if more clarification is needed)

In summation this is the current progress

- [X] MovementModes transferred correctly between client/server
- [ ] Correct state locations transferred between client/server
- [ ] Possible refactoring of location/movement based on the movement modes and a centralized location of the character


GooeyHub:

GooeyHub:

flo:
Added API label as it requires an update of the NPCMovementSystem of https://github.com/Terasology/Pathfinding

flo:
I propose you keep the old constructor as deprecated so that it does not break the NPCMovmentSystem and then we need to create an Pathfinding issue afterwards to not use that constructor.

engine/src/main/java/org/terasology/logic/characters/CharacterMoveInputEvent.java
```diff
@@ -80,6 +83,8 @@ public boolean isRunning() {
         return running;
     }
 
+    public boolean isCrouching() { return crouching; }
```
Cervator: Style issue: needs a couple newlines :-)

engine/src/main/java/org/terasology/logic/characters/CharacterMoveInputEvent.java
```diff
@@ -36,11 +37,12 @@
     protected CharacterMoveInputEvent() {
     }
 
-    public CharacterMoveInputEvent(int sequence, float pitch, float yaw, Vector3f movementDirection, boolean running, boolean jumpRequested, long delta) {
+    public CharacterMoveInputEvent(int sequence, float pitch, float yaw, Vector3f movementDirection, boolean running, boolean crouching, boolean jumpRequested, long delta) {
```
Cervator: Agreeing with @flo that it would be nice to just add a new constructor and deprecate this one. Other modules than Pathfinding might also break.

Really you'd keep this change but then add a new deprecated constructor with the old signature + `@deprecated` that just calls the new one and nothing else.

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -118,7 +118,9 @@
     private Vector3f relativeMovement = new Vector3f();
     private boolean isAutoMove = false;
     private boolean runPerDefault = true;
+    private boolean crouchPerDefault = false;
```
Cervator: Any reason to really have this one rather than just initialize `crouch` to `false` ?

ThompsonTyler:
I definitely agree a movement refactoring is the way to go, but I can check this out and do a bit more testing to get it "usable" for now for the sole purpose of basically not breaking the game when a client crouches. 

Will check out fixing up anything that could cause problems in the mean time and also update the constructors sometime today hopefully.

GooeyHub:

ThompsonTyler:
I believe this should be setup to work for functionality. Reverted the changes made for AI and updated with a deprecation on the old constructor. Fixed the few style changes and just did another quick test to make sure desyncing doesn't occur.

The targeting was fixed in another PR prior to the last play-test. This does still bug out when crouch is held, but stops as soon as it's let go and doesn't break the client anymore. It still doesn't really function for crouching, but it stops breaking the game when a client crouches, which in the meantime until movement is refactored, might be worthy enough to warrant this PR still.

GooeyHub:

nihal111:
Why does this PR have so many commits and file changes?
Was it merged improperly?

flo:
I wondered about that too. Maybe it was because it was merged and then another commit made?

I did a manual diff compare with git to actually see the delta

Cervator:
@flo @nihal111 @ThompsonTyler: So a classical reason for this kinda commit mess is when somebody starts out on a branch that's behind the upstream develop, commits to it occasionally but keeps it alive for some time (in this case a bit more than a month), then merges *from upstream* into their branch, which then in turn gets merged again *into* upstream.

Every commit that got merged into the originating branch like that are now in a different order due to the way Git tracks stuff (each commit is a state for the entire tree). So they show up for re-integration but Git *is* smart enough to realize they were already merged so the diff remains tiny. That's on a good day, anyway, mix in even more merges and you might hit conflict hell when Git starts seeing everything as unique new changes.

Some tools are smart enough to ignore such "merge commits" in some cases, like Jenkins only listed this entire PR as two unique commits: http://jenkins.terasology.org/job/Terasology/1991/changes

The PR commit listing is not and the GitHub network graph now also draws a second line way back to a month plus change with lots of new dots, but luckily doesn't jump some 189 commit gap forward when the merge is done. No idea if the commits end up counting double for stats and such.

In conclusion: try to avoid using merge to update a branch - use rebase :-) Or even just leave the branch out of date and we fix any conflicts during the merge into upstream (resulting in no spammy history). 

In a case like this using GitHub's squash-merge is an option to rebase-in-place to get rid of the excess history. I'm somewhat reluctant to trust the UI though and tend to then rebase locally and push to a temporary integration branch, just in case.

PR #2906, opened 2017-04-18
Tutorial: How to use EvoSuite to create tests and understand the project better

iaronaraujo:
### Contains

This PR contains some changes and commits to aid new Testers in the task of using Evosuite to automatically generate unit tests, improving the generated tests and making changes to the project based on the tests. The tutorial consists of the following commits:
https://github.com/iaronaraujo/Terasology/commit/aa3ae1559e57dcedae25707e1e59992b62d8a726
https://github.com/iaronaraujo/Terasology/commit/41ca661d395d92e033a276ded57ab8d4f58140e6
https://github.com/iaronaraujo/Terasology/commit/140cfc656236a6703b7f8376914939184cf96b6e

### How to test

Previous Unit Tests are still passing. The best way to test the tutorial is to follow it and see if anything could possibly add a fault to the project and if anything is unclear.

### Tutorial

**Before we begin** 

For this tutorial, it is necessary to download EvoSuite's plugin for IntelliJ. To do so, follow these steps:

1 - Open IntelliJ
2 - Open File > Settings
3 - On this new window click on "plugins" at the left side of the window, then click on the "Browse repositories..." button at the bottom of the window.
4 - On this new window, search for "EvoSuite Plugin" then install it.
5 - Download "evosuite-1.0.3.jar" on this link http://www.evosuite.org/downloads/

After these steps, it should be possible to automatically generate tests using EvoSuite.

**First part of the Tutorial: Generating Tests with Evosuite**

Commit referent to this part: https://github.com/iaronaraujo/Terasology/commit/aa3ae1559e57dcedae25707e1e59992b62d8a726

With the plugin installed, to generate tests you just need to right click a class or package and select "Run Evosuite". The "EvoSuite Options" window will appear, and there you can customize some aspects of the test generation. On "EvoSuite location", select the "evosuite-1.0.3.jar" you just downloaded. After everything is alright, press "enter" and EvoSuite will begin the test generation.

When the generation ends, two classes are created for each class under test: the real test class and a scaffolding class. We should ignore the scaffolding for this tutorial. It is possible to see that you cannot execute the tests because they are dependent on an EvoSuite's library. Since this library has a dependency with maven, we will manually modify these tests to improve them and make them work without this library.

**Second part of the Tutorial: Fixing and improving Evosuite's tests**

Commit referent to this part: https://github.com/iaronaraujo/Terasology/commit/41ca661d395d92e033a276ded57ab8d4f58140e6

First of all, we need to make these tests executable. To do that, remove the first line above the class declaration. Then, remove the "extends" on the class declaration, since we do not want a dependency with the scaffolding generated by EvoSuite. Then, remove the imports that can't be found, since they come from EvoSuite's library, and remove the lines that need this library. If you prefer, you can also remove the whole test method.

The test should be executable by now, but as we can see, the methods' names aren't very good since they don't show the purpose of the test. We should check each method and see if they make sense. If they do, we should rename it to show its purpose. If they don't or makes us have some doubts, we go to the third part of this tutorial. By doing this, we understand better the class under test. 

**Third part of the Tutorial: Making code changes and adapting the tests**

Commit referent to this part: https://github.com/iaronaraujo/Terasology/commit/140cfc656236a6703b7f8376914939184cf96b6e

Note: It is possible to see on the commit that I added a new method to the test. I did that because on the previous commit I deleted it by mistake. But by following this tutorial, it is not necessary to create new methods on the test class.

When analyzing the test, sometimes we wonder if some methods make sense and then we wonder what should the code really do on the situation being tested. That's when we should look better at the code,, ask questions and then decide what to change to improve the project's quality.

As an example, when I generated tests for the AudioConfig class, some test methods used negative numbers and numbers higher than 100 on the setMusicVolume method. So I went to Terasology's IRC channel and talked about this with some contributors. We got to the conclusion that these tests didn't make sense since the volume minimum value is 0 and its maximum value is 100. But the tests clearly showed us that it was indeed possible to pass those values to the setMusicVolume method even if they didn't make sense. Therefore, one solution we had was to check the input of this method, so if it was higher than 100, we would change it to 100, and if it was lower than 0, we would change it to 0. To reflect these changes, I also had to change the tests too. All these changes can be seen on the commit referent to this part of the tutorial.

**Conclusion Remarks**

By following this tutorial, it is possible to generate a lot of tests and get a lot of discussions about cases that nobody thought about. Even though it takes a considerable amount of time to make the test changes and analyze them, we also learn a lot about the code by doing this, hence why this tutorial is recommended especially for parts of the project that you do not know much about.

### DIFF
--- a/engine-tests/src/test/evosuite/org/terasology/AudioConfig_ESTest.java
+++ b/engine-tests/src/test/evosuite/org/terasology/AudioConfig_ESTest.java
@@ -51,7 +51,15 @@ public class AudioConfig_ESTest{
         AudioConfig audioConfig0 = new AudioConfig();
         audioConfig0.setMusicVolume(2476.09F);
         float float0 = audioConfig0.getMusicVolume();
-        assertEquals(2476.09F, float0, 0.01F);
+        assertEquals(100F, float0, 0.01F);
+    }
+
+    @Test(timeout = 4000)
+    public void testAudioConfigSetMusicVolumeNegative()  throws Throwable  {
+        AudioConfig audioConfig0 = new AudioConfig();
+        audioConfig0.setMusicVolume((-1022.3F));
+        float float0 = audioConfig0.getMusicVolume();
+        assertEquals((0.0F), float0, 0.01F);
     }
 
     @Test(timeout = 4000)
--- a/engine/src/main/java/org/terasology/config/AudioConfig.java
+++ b/engine/src/main/java/org/terasology/config/AudioConfig.java
@@ -39,6 +39,8 @@ public class AudioConfig {
      * @param soundVolume Sets the sound volume offering both the new as well as prior value if needed
      */
     public void setSoundVolume(float soundVolume) {
+        if(soundVolume > 100) soundVolume = 100;
+        if(soundVolume < 0) soundVolume = 0;
         float oldValue = this.soundVolume;
         this.soundVolume = soundVolume;
         propertyChangeSupport.firePropertyChange(SOUND_VOLUME, oldValue, soundVolume);
@@ -55,6 +57,8 @@ public class AudioConfig {
      * @param musicVolume Sets the music volume offering both the new as well as prior value if needed
      */
     public void setMusicVolume(float musicVolume) {
+        if(musicVolume > 100) musicVolume = 100;
+        if(musicVolume < 0) musicVolume = 0;
         float oldValue = this.musicVolume;
         this.musicVolume = musicVolume;
         propertyChangeSupport.firePropertyChange(MUSIC_VOLUME, oldValue, musicVolume);


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

Thanks again for this. Might take a bit to find the time to review deeper - off on a business trip that's keeping me busy :-)

GooeyHub:

oniatus:
I think there is some value in generating this api-exercising tests for review purpose. But the generated tests are tied a lot to the actual implementation and it will require a lot of review effort to make use of this tests which is one thing we are lacking at most for the time being :)

Going to close this PR for now as it has been kind of inactive and I don't see a point where we can merge it or benefit from it. Feel free to re-open or add further improvement to it.

PR #2907, opened 2017-04-18
Placement of slope blocks

elwe74:
### Contains

So far "slope" shapes can be placed only horizontally.
Now slope shapes can be placed in all possible orientations.

The new EdgeBlockFamily will place the slope block according to the hit edge.

### How to test

Do not use an existing saved game: Testing this will destroy already used slope blocks!

In a new game try "give core:dirt 50 engine:slope" or "engine:halfSlope" and place the slope blocks.
Depending on where you hit the existing blocks the slope block will be placed in different orientations.

* What about existing saved games? "core:dirt:engine:slope.FRONT" is not understood with the new EdgeFamily.
* Are the different slope-shapes really core functionality? Or move them to a module?


GooeyHub:

vampcat:
I'm not sure how Terasology handles such things, but can't we add a temporary condition somewhere to convert all FRONT blocks to the new slope blocks, and finally remove that condition (and hence break the compatibility) with the next major release?

skaldarnar:
> What about existing saved games? "core:dirt:engine:slope.FRONT" is not understood with the new EdgeFamily.

I don't think backwards compatibility of game worlds is high on our agenda. For now, it's fine if old worlds break. This could be an idea for future GSoC projects, implementing a mechanism to automatically convert/update worlds...

> Are the different slope-shapes really core functionality? Or move them to a module?

Hmm... this could potentially go into [StructuralResources](https://github.com/Terasology/StructuralResources). I'll have a quick review and let you know what I think about it 😉

skaldarnar:
@GooeyHub ok to test

engine/src/main/java/org/terasology/math/BlockHitDetector.java
```diff
@@ -0,0 +1,76 @@
...
+                return Edge.forSides(sideZ,sideY);
+            case 6: //Y>Z>X
+                return Edge.forSides(sideY,sideZ);
+            case 7: //Y>X>Z
+                return Edge.forSides(sideY,sideX);
+            default: //2: //X>Y>Z>X invalid, 5: //Y>X>Z>Y invalid
+                return Edge.BottomBack;
+        }
+    }
+
```
skaldarnar: remove empty lines

engine/src/main/java/org/terasology/math/Edge.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.EnumSet;
+import java.util.Set;
+
+/**
+ * A cube has 12 edges. For asymmetric cubes each edge can have 2 orientations, giving 24 different combinations.
+ * First 12 edges are for symmetric blocks, all 24 are for asymmetric blocks.
+ */
+public enum Edge {
+
+    BottomBack(Rotation.none()),
```
skaldarnar: I'm not sure about the naming here, but finding better names isn't easy, either...

engine/src/main/java/org/terasology/math/Edge.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private final Name name;
+    private final Rotation rotationFromBottomBack;
+
+    private Edge(Rotation rotationFromBottomBack) {
+        this.name = new Name(name());
+        this.rotationFromBottomBack = rotationFromBottomBack;
+    }
+
+
```
skaldarnar: empty lines again

engine/src/main/java/org/terasology/math/Edge.java
```diff
@@ -0,0 +1,143 @@
...
+            return symmetricSubset;
+        }
+        symmetricSubset = Collections.unmodifiableSet(EnumSet.range(Edge.BottomBack, Edge.TopLeft));
+        return symmetricSubset;
+    }
+
+    public static Edge of(Name name) {
+        for (Edge e : values()) {
+            if (e.name.equals(name)) {
+                return e;
```
skaldarnar: Maybe I worked to much with Scala - can this be simplified using streams/a more functional interface?

engine/src/main/java/org/terasology/world/block/Block.java
```diff
@@ -670,7 +670,7 @@ public void renderWithLightValue(float sunlight, float blockLight) {
 
     @Override
     public String toString() {
-        return uri.toString();
+        return String.valueOf(uri);
```
skaldarnar: ?

engine/src/main/java/org/terasology/world/block/family/EdgeBlockFamily.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * Block group for blocks that can be oriented on 12 edges.
+ *
+ */
+public class EdgeBlockFamily extends AbstractBlockFamily {
+
+    Logger logger = LoggerFactory.getLogger(EdgeBlockFamily.class); //fixme delete...
```
skaldarnar: you can keep it for debugging messages 😉

GooeyHub:

engine/src/main/java/org/terasology/world/block/family/EdgeBlockFamily.java
```diff
@@ -0,0 +1,100 @@
...
+            if (block == null) {
+                throw new IllegalArgumentException("Missing block for edge: " + edge.toString());
+            }
+            block.setBlockFamily(this);
+        }
+    }
+
+    public Block getBlockForPlacement(WorldProvider worldProvider, BlockEntityRegistry blockEntityRegistry, Vector3i location, Vector3f direction, Vector3f hitNormal, Vector3f hitPosition) {
+        Edge edge = BlockHitDetector.detectEdge(hitPosition, location);
+        logger.info( "orientation "+  edge ); //fixme delete...
```
skaldarnar: yes, info level logs should indeed be removed

Cervator:
Very nice :-)

Marking this as API just because it stands to break world compatibility, which indeed isn't a critical factor yet - that's more of a Beta item.

skaldarnar:
I'm looking forward to test this out this weekend ... eventually... 😂

engine/src/main/java/org/terasology/math/Edge.java
flo: The method seems redundant to me as there is already a method provided by the java vm called Edge.valueOf

engine/src/main/java/org/terasology/math/Edge.java
flo: The names should follow the java naming convention that constants are upper case and separted with _. Otherwise the names are fine as they match the names of the Side enum.

engine/src/main/java/org/terasology/world/block/family/EdgeBlockFamily.java
flo: yes but please make it a private static final field.

engine/src/main/java/org/terasology/world/block/family/EdgeBlockFamily.java
```diff
@@ -0,0 +1,100 @@
...
+        Block b = blocks.get( edge );
+        if( b == null ) {
+            b = getArchetypeBlock();
+        }
+        return b;
+    }
+
+    @Override
+    public Block getBlockForPlacement(WorldProvider worldProvider, BlockEntityRegistry blockEntityRegistry, Vector3i location, Side attachmentSide, Side direction) {
+        //FIXME throw Exception?
```
flo: Either remove comment or fix. Maybe all implementations of this method should lbe switched to your new method siganture?

engine/src/main/java/org/terasology/world/block/family/FreeformBlockFamilyFactory.java
```diff
@@ -34,10 +35,13 @@ public BlockFamily createBlockFamily(BlockFamilyDefinition definition, BlockBuil
 
     @Override
     public BlockFamily createBlockFamily(BlockFamilyDefinition definition, BlockShape shape, BlockBuilderHelper blockBuilder) {
-        if (shape.isCollisionYawSymmetric()) {
-            return symmetric.createBlockFamily(definition, shape, blockBuilder);
-        } else {
-            return horizontal.createBlockFamily(definition, shape, blockBuilder);
+        switch( shape.getBlockShapePlacement().getType() ) { //Fixme others...
```
flo: The fixme really doesn't help here. 

Instead it is better to add an explict handling for all known cases and have no default section. That way bug finding tools can warn about it once  there is a new constant that does not get handled

engine/src/main/java/org/terasology/world/block/shapes/BlockShapePlacement.java
```diff
@@ -0,0 +1,96 @@
...
+            type = PlacementType.Cube;
+        }
+        return new BlockShapePlacement(type, Symmetry.Symmetric, new Name(type.getArchetype()));
+    }
+
+    public enum PlacementType {
+        Cube("none"),
+        Side("FRONT"),
+        Edge("BottomBack"),
+        Corner(""), //fixme
```
flo: If a placement type isn't supported yet, please don't add it to enum. It will just make other developers might assume that it does exist when they see it in the autocompletion.

engine/src/main/java/org/terasology/world/block/shapes/JSONUtils.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+package org.terasology.world.block.shapes;
+
+import com.google.gson.JsonObject;
+
+import java.util.Optional;
+
+/**
+ * FIXME find better location
```
flo: Move to org.terasology.utilities.gson ?

engine/src/main/java/org/terasology/world/block/family/EdgeBlockFamilyFactory.java
```diff
@@ -0,0 +1,100 @@
...
+    public BlockFamily createBlockFamily(BlockFamilyDefinition definition, BlockShape shape, BlockBuilderHelper blockBuilder) {
+        if (!definition.isFreeform()) {
+            throw new IllegalStateException("A shape cannot be provided when creating a family for a non-freeform block family definition");
+        }
+        BlockUri uri = new BlockUri(definition.getUrn(), shape.getUrn());
+
+        Edge archetypeEdge = Edge.of(shape.getBlockShapePlacement().getArchetype());
+        if( archetypeEdge != Edge.BottomBack ) {
+            Rotation rotToBottomBlock = archetypeEdge.getRotationFromBottomBack();
+            //Fixme: invert and create base for following rotations...
```
flo: What about this fixme?

engine/src/main/java/org/terasology/math/Edge.java
elwe74: It is redundant, but will ease the usage. Mostly valueOf will be used with a Name-Object. So I will have to use Edge.valueOf( name,toString().toUppercase() ). 
I have replaced the iteration with a faster Map-Lookup.

engine/src/main/java/org/terasology/world/block/Block.java
elwe74: It prevents a NPE if uri is null because Block is not fully initialized.

engine/src/main/java/org/terasology/world/block/family/EdgeBlockFamily.java
elwe74: I will remove it. It was only for developement and for the first testers.

engine/src/main/java/org/terasology/world/block/shapes/BlockShapePlacement.java
elwe74: You are right, I think the same way. It was for just my development, therefore the empty fixme-comment. In the meantime Corner exists...

engine/src/main/java/org/terasology/world/block/family/EdgeBlockFamily.java
elwe74: I don't know what implementations exists. There are so many modules...  So I have added a new method with default implementation. So the existing mehtod in my new EdgeBlockFamily is not called anymore, but I can't  remove it without loosing comaptibility.

Throw IllegalStateException?

engine/src/main/java/org/terasology/world/block/shapes/JSONUtils.java
elwe74: Good location.

engine/src/main/java/org/terasology/world/block/family/EdgeBlockFamilyFactory.java
elwe74: Maybe better a TODO. I will not implement this, becaue it is somewhat complicated. A Rotation cannot give its inverse. This functionality will be used if someone builds a shape with edge placement, which is not oriented that BOTTOM_BACK is a proper base orientation. As far as I see all existing slope shapes doesn't need this to be implemented.

flo: Why is arche type not always BottomBack. What did you plan to do what is so complicated?

To me it seems like archetypeSide of HorizontalBlockFamily does not provide any real benefit either but was just something planed for the future that was never needed. I guess you tried to do something analog.

I would suggest to make the arche type edge of the edge block family a constant.

engine/src/main/java/org/terasology/world/block/family/EdgeBlockFamily.java
flo: How about calling your new getBlockForPlacement variant then?

flo: Well, it's bad when a interface offers a method and you can't rely on it to work with all implementation.

I think it would be better if the new method would have all the parameters of the old method + 1 additional that specifies the edge. Then a deprecated version of the old method could be kept that calls the new method with the bottom edge of the provided side. All existing engine code should be changed to call the new method and all implementations should be changed to new method.

skaldarnar:
Just noticed this PR has been inactive for some time now, but has some very nice things in it. @elwe74 are you still around and occasionally working on it? I guess it is nearly ready to merge once Flo's comments are addressed 😃

GooeyHub:

elwe74:
After a long time with computer problems and too few spare time I have managed to process the requested changes. Additionally there is analogous code for Sides and Corners.

I would propose to revert the changes to all shapes in Terasology/resources, maybe delete the shapes from Terasology/resources and transfer them to StructuralResources. 

I think that for good usage in StructuralTemplates rotations on a vertical axis must be supported.

oniatus:
@elwe74 There are still some pending change requests on this PR. Any plans how to continue here?

pollend:
This will have to reconciled with #3343

jellysnake:
I'm closing this as it's now incompatible with the block family changes.

PR #2908, opened 2017-04-19
alter some interfaces to support GameEngine testability

kaen:
### Contains

Interface changes to support ModuleTestingEnvironment (creation and manipulation of TerasologyEngines)
see https://github.com/kaen/ModuleTestingEnvironment

### How to test

Engine should work exactly as it did before :)

### DIFF
--- a/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
+++ b/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
@@ -53,6 +53,7 @@ import org.terasology.logic.behavior.asset.BehaviorTree;
 import org.terasology.logic.behavior.asset.BehaviorTreeData;
 import org.terasology.monitoring.Activity;
 import org.terasology.monitoring.PerformanceMonitor;
+import org.terasology.network.NetworkSystem;
 import org.terasology.persistence.typeHandling.TypeSerializationLibrary;
 import org.terasology.reflection.copy.CopyStrategyLibrary;
 import org.terasology.reflection.reflect.ReflectFactory;
@@ -414,6 +415,7 @@ public class TerasologyEngine implements GameEngine {
 
         Iterator<Float> updateCycles = timeSubsystem.getEngineTime().tick();
         CoreRegistry.setContext(currentState.getContext());
+        rootContext.get(NetworkSystem.class).setStateContext(currentState.getContext());
 
         for (EngineSubsystem subsystem : allSubsystems) {
             try (Activity ignored = PerformanceMonitor.startActivity(subsystem.getName() + " PreUpdate")) {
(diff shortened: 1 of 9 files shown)


GooeyHub:

Cervator:
Neat! Pinging @immortius @flo @skaldarnar to maybe get some quick review :-)

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
```diff
@@ -389,45 +389,56 @@ public synchronized void run(GameState initialState) {
...
+        PerformanceMonitor.endActivity();
+    }
 
-            processPendingState();
+    /**
+     * Runs a single "tick" of the engine
+     * @return true if the loop requesting a tick should continue running
+     */
+    public boolean tick() {
+        if(shutdownRequested) {
```
immortius: Pedantic comment, but missing a space between if and bracket.

GooeyHub:

GooeyHub:

bkconrad:
Getting a client to connect to a host required some additional changes. Chiefly I had to normalize the behavior of States and Engines regarding when CoreRegistry is set. Now by definition a State is required to expose a context, and that context will be set as CoreRegistry when the engine is `tick`d.

Additionally, NetworkSystem now uses a local context reference rather than CoreRegistry in order to make it thread-safe (since some parts are apparently executed off of the main thread). The engine now sets the NetworkSystem's local contex on each `tick`. This is equivalent to the old behavior of relying on the State to set its own context as CoreRegistry, but without the CoreRegistry reference it becomes thread safe in the test environment. engine-tests have been updated to reflect this requirement.

Additionally I exposed the cleanup method to allow proper shutdown/cleanup of the Engine instances between test cases

kaen:
Getting a client to connect to a host required some additional changes. Chiefly I had to normalize the behavior of States and Engines regarding when CoreRegistry is set. Now by definition a State is required to expose a context, and that context will be set as CoreRegistry when the engine is `tick`d.

Additionally, NetworkSystem now uses a local context reference rather than CoreRegistry in order to make it thread-safe (since some parts are apparently executed off of the main thread). The engine now sets the NetworkSystem's local contex on each `tick`. This is equivalent to the old behavior of relying on the State to set its own context as CoreRegistry, but without the CoreRegistry reference it becomes thread safe in the test environment. engine-tests have been updated to reflect this requirement.

Additionally I exposed the cleanup method to allow proper shutdown/cleanup of the Engine instances between test cases

GooeyHub:

Cervator:
Tested it out, both in single, server-client + client, and headless + client, seemed solid! Well, found a pre-existing bug I hadn't noticed before, but that's something else :-)

Thanks!

PR #2910, opened 2017-04-21
Fix reloading of assets on shutdown

vampcat:
This PR removes a redundant call to the function `cheapAssetManagerUpdate` on shutdown, which causes a reload of assets during shutdown.

How to test: Run the game, do something (or don't) and ensure that you get no weirdness when quitting the game.

Pinging @immortius for review.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
+++ b/engine/src/main/java/org/terasology/engine/TerasologyEngine.java
@@ -435,7 +435,7 @@ public class TerasologyEngine implements GameEngine {
         changeStatus(StandardGameStatus.SHUTTING_DOWN);
 
         if (currentState != null) {
-            currentState.dispose();
+            currentState.dispose(true);
             currentState = null;
         }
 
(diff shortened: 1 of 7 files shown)


GooeyHub:

Cervator:
Tested this out real quick but it does seem to dispose a little *too* well - try hosting a game, exiting to main menu, then hosting a game again. Suddenly lots of stuff is missing, like the toolbar :-)

Rolled back to test without this PR and that scenario was fine again.

vampcat:
As always, you're right :D

The latest commit is (hopefully) a proper fix, but involves changing 4 more files than what I would have liked.

If the new version of Gestalt fixes the redundant loading issues and can be integrated into Terasology anytime in the near future, we can just close this PR.. otherwise, I think you can test this again :D

GooeyHub:

Cervator:
Certainly I'm not always right, I just have a decent testing routine and some basic intuition for where to go looking after this long :-)

Change looks right and in-game is fine now. I discovered #2915 while testing but confirmed it as an preexisting issue. Thanks to that I did seem to notice a difference in shutdown speed with this PR (although sadly in neither case did the game actually fully dispose after a multiplayer session)

Issue #2905, opened 2017-04-17
Crash while loading (solo mode) world + modules


magali-demorangies:
### What you were trying to do
After first launch of the game, for my first game I chose the solo game, and there the Sample something world. To customize this world I took several modules, as nothing was warning this could cause a trouble...
### What actually happened
When the game I selected started to load, it suddenly crashed.
### How to reproduce
The problem is I don't remind exactly which modules I selected to customize the world...
* Step 1 : take the first world from the list in Solo mode of the game, named "something Sample something"
* Step 2 : add modules, not all of them but several, and all the ones which were shown in yellow as I supposed they were needed by the modules I took (total around 20 certainly, maybe a little bit more)
* Step 3 : click on "play", the game showed the loading screen a short moment then crashed.

### Log details and game version
Terasology - Version
Jenkins:
  URL: http://jenkins.terasology.org/job/TerasologyStable/71/
  Build number: 71
GIT branch: origin/master
Created at: 2017-03-20T05:04:33Z
http://terasology.org
https://github.com/MovingBlocks/Terasology
### Computer details
Processor Intel Core i7 - 6700 CPU @ 3.40 GHz (x8)
Graphic Card NVIDIA GeForce GTX 1060 6 GB
RAM 8 G
Network : Eth, WiFi
OS : Windows 10
Log details : https://pastebin.com/cgTFuZvw

gianluca-nitti:
That's an `OutOfMemoryError` so the Java Virtual Machine running the game ran out of memory while perfotming the task.
You could try to allow the JVM to use more of your machine's physical memory.

How do you run the game? If you use the launcher, go in its settings (by clicking on the gear icon) and tweak with the Max and Initial memory settings; since you have 8GB total I would set them to 6GB, provided you run the game without other programs in background.

Also you should ensure that you are using a 64-bit JVM (using a 32-bit JVM on a 64-bit CPU on Windows works due to [WOW64](https://en.wikipedia.org/wiki/WoW64), but with performance limits). Since you are on Windows, you can do it by going in Programs and Features in the Control Panel and searching for "Java". Look if you see something like `Java 8 Update 121 (64-bit)`. If you only have a `Java 8 Update 121` it's the 32-bit JRE; in this case, to get optimal performance you should download and install `jre-8u121-windows-x64.exe` from [here](http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html).

magali-demorangies:
[...]

PR #2911, opened 2017-04-21
Show warning if running in 32-bit JVM on 64-bit system

gianluca-nitti:
Idea for this patch is from issue #2905.

### Contains

If the game is running in a 32-bit JVM on a 64-bit system, a warning is shown in the console and on the main menu screen.
This can help users which get limited performance due to running in a 32-bit JVM when they could install a 64-bit one (thus native for their CPU).

This is not uncommon especially on Windows, I remember seeing people playing other Java games on x86 JVMs despite being on a 64-bit CPU and OS.

On Windows, this is checked by simply verifying if the [PROCESSOR_ARCHITEW6432](https://blogs.msdn.microsoft.com/david.wang/2006/03/27/howto-detect-process-bitness/) is set.
On POSIX OSs, the JVM architecture is read from the "os.arch" property and the real system architecture by reading the output of "uname -m".

### How to test

You need to be on a 64-bit CPU and OS, and have installed both a 64-bit JVM ("native") and a 32-bit one ("non-native"). Start the game on this branch with the 64-bit JVM (your usual run configuration) and everything should be as before. Now start Terasology with the 32-bit one - some examples of how to do it:
* in IntelliJ: create a new run configuration by duplicating the default one ("Terasology"); then, change the JRE path pointing it to the directory where the 32-bit one is installed (for example, on Windows should be something like `C:\Program Files (x86)\Java\jre1.8.0_121`), then run with this new configuration; if there are memory parameters you may need to remove or tweak them, for example during my tests the `-Xms256m -Xmx1536m` copied from the normal configuration didn't worked with the x86 JVM;
* from the command line: run `gradlew game` or `gradlew server` (with the second one you will obviously see the warning in the terminal only) with the JAVA_HOME env variable set to the 32-bit JVM home; for example on my Ubuntu 16.04 I did `JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/ ./gradlew server`.

In this case, you should see the warning. On stdout, it's printed right after the system info (after `Processors`); on the main menu screen, it's shown in red in the bottom-left corner.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/NonNativeJVMDetector.java
+++ b/engine/src/main/java/org/terasology/engine/NonNativeJVMDetector.java
@@ -46,11 +46,10 @@ public final class NonNativeJVMDetector {
 
     private static boolean posixSystemIs64() throws IOException, InterruptedException {
         Process unameProc = new ProcessBuilder("uname", "-m").start();
-        BufferedReader unameStdout = new BufferedReader(new InputStreamReader(unameProc.getInputStream()));
         unameProc.waitFor();
-        boolean result = unameStdout.readLine().endsWith("64");
-        unameStdout.close();
-        return result;
+        try (BufferedReader unameStdout = new BufferedReader(new InputStreamReader(unameProc.getInputStream()))) {
+            return unameStdout.readLine().endsWith("64");
+        }
     }
 
     private static boolean jvmIs64() {
--- a/engine/src/main/java/org/terasology/engine/NonNativeJVMDetector.java
+++ b/engine/src/main/java/org/terasology/engine/NonNativeJVMDetector.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

@gianluca-nitti: thanks! :-)

Poking @skaldarnar @msteiger @MaxBorsch @qwc for potential review

GooeyHub:

engine/src/main/java/org/terasology/engine/NonNativeJVMDetector.java
```diff
@@ -0,0 +1,59 @@
...
+                return !jvmIs64() && posixSystemIs64();
+            } catch (Exception ex) {
+                return false;
+            }
+        }
+    }
+
+    private static boolean posixSystemIs64() throws IOException, InterruptedException {
+        Process unameProc = new ProcessBuilder("uname", "-m").start();
+        BufferedReader unameStdout = new BufferedReader(new InputStreamReader(unameProc.getInputStream()));
```
oniatus: you could use the try-with-ressources statement here.

gianluca-nitti: Oh right, I forgot about it, thanks for the hint!

GooeyHub:

PR #2914, opened 2017-04-22
add startBehavior and related tests

kaen:
### Contains

A new `BehaviorSystem` helper to programmatically start a behavior on an entity. Adds a new `getInterpreter` method for getting a specific entity's interpreter for manipulation or inspection.

Both of these were possible before but required knowledge of the BehaviorSystem's internals and required several lines of code. This PR makes these operations officially supported and tested.

### How to test

Covered by new unit tests. Also used in unit tests for some modules I'm currently developing.


GooeyHub:

Cervator:
Seems fine to me - poking a few behavioral experts just for reference and review if desired, otherwise will merge tomorrow or so :-)

flo:
What I miss in this PR is an explanation why this PR is something we need. E.g. why is this API access needed? Aren't we just exposing implementation details? Exposing implementation details can be bad for multiple reasons. For example it can result in an "API" break when we want to do a refactor later on. Also when there is no user it is just added complexity for nothing. So the big question is: Is there a module that needs this API do something cool?

I think it is great that you made a PR that is even unit tested I think we should only merge it if it is good for something except being a new possible API to use.

Cervator:
@flo yep something super cool is connected with this: programmatically driving in-game entities for automated testing :)

@kaen curiously could you add links to the pending modules?

flo:
@Cervator uhm, why wouldn't be it enough to just change the behavior of the entity via component modification? That would archive the same or not?

Cervator:
@flo that I'm less sure about. I just recall the topic came up earlier on IRC or Slack for what @kaen was trying to achieve. What options there are for achieving it I haven't had time to dig into :/

If your option is better than the one in this PR then that's great! I just don't know. Part reason why I'm curious to see the example modules and what they add to the picture :)

oniatus:
We can probably save the exposed interpreter and use `BehaviorSystem.update` instead but I like the idea to test the entire system.

However, I would like to delegate this to @dkambersky as this issue fits perfectly into the gsoc topic :)

GooeyHub:

kaen:
Out of date and no longer needed, I believe. Even if it is I'll just graft it onto the changes after the v2 merge. Closing to clean up PRs a bit

Issue #2912, opened 2017-04-22
If a player has already picked a custom name don't prompt them to pick a custom name


Cervator:
The first time a player enters either Host Game or Join Game they get a prompt to pick a name - but should really only get one if they *haven't already* picked a name.

At present you can start the game, go select a custom player name, then still have the game helpfully try to offer to set you up with a custom name. That's silly. Probably an easy fix :-)

Malanius:
I'll take a look onto it.

Cervator:
Completed by #2920

PR #2920, opened 2017-04-29
Disable prompt if username was picked in Join/Host screens

ar0ne:
### Contains

Fix of issue #2912

I prefer to change greeting phrase for first visit of Join/Host screens.
**From**: Would you like to use a non-default username?
**To**: Would you like to change default username?

Because as I see it's a wrong question to ask. I'd like to propose ask user to change default nickname. Because before we had "Cancel" button which in this case means - I don't want to use default username, but it doesn't save user choose, it just skips it. And "Yes" button means - I want to use another one, but you can do not change it, etc. In my proposal we won't have this misunderstandings.
Also I've commented wrong translation at present for few languages which I don't know. And hope we will make right translation for them soon.

And add simple validation for usernames - I guess we shouldn't accept usernames from white spaces and longer then 100 chars at least.

### How to test

Remove **config.cfg** file, or change value of "hasEnteredUsername" field to false. And open Join/Host Game menu, and Settings as well.


GooeyHub:

Cervator:
Thanks! Checking this out :-)

@GooeyHub: add to whitelist

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/EnterUsernamePopup.java
```diff
@@ -38,18 +44,59 @@ public void initialise() {
...
                 playerConfig.setHasEnteredUsername(true);
                 getManager().popScreen();
-            }
-        });
+            });
+            okButton.bindEnabled(new ReadOnlyBinding<Boolean>() {
+                @Override
+                public Boolean get()
+                {
+                    return Strings.isNullOrEmpty( validateUserName() );
```
Cervator: Tiny code style tweak: no extra whitespace within parenthesis

engine/src/main/resources/assets/i18n/menu_cs.lang
```diff
@@ -86,7 +86,7 @@
     "drop-item": "Zahodit předmět",
     "edit-server": "Upravit",
     "edit-server-title": "Přidat/Upravit server",
-    "enter-username-message": "Ahoj a vítej v muliplayeru Terasology!\nChceš použít vlastní jméno?",
+//    "enter-username-message": "Ahoj a vítej v muliplayeru Terasology!\nChceš použít vlastní jméno?",
```
Cervator: Not good to have comments inside JSON let alone commented out strings - better to leave an associated TODO in a .java that relates or submit an issue to figure out what we need to do :-)

GooeyHub:

GooeyHub:

ar0ne:
Hi @Cervator, I created new issue to track this problem with translation. And tried to mark word as "fuzzy" as you mentioned above, but unsure, does that Weblate actual? Because I can't find all languages which we support.

Cervator:
Thanks :-)

Oddly after we added a bunch of new languages during Google Code-In, they didn't all seem to register for Weblate :(

It feels like it is stuck trying to rescan for languages or a cache needs to be cleared, or something. I haven't had time to dig into it. 

In the meantime the issue should be good for keeping track, I appreciate it!

On a side note I just realized it would be unusually difficult to encounter the new string in a non-English language - since you pick the language on the same settings screen as the player name, meaning just changing the language "counts" :-)

That may change though so I'm not worried about that right now. And maybe you could pick the language via the game launcher one day.

PR #2930, opened 2017-05-03
BehaviorSystem- changed Map to ConcurrentHashMap

nihal111:
This is done for the new changes merged in https://github.com/Terasology/WildAnimals/ to work.

The additions to the WildAnimals module make changes to the entity's behavior component to switch behavior.

The current system uses events to trigger changes in the Behavior Tree. This means that if the deer is hit, the deer's `BehaviorComponent.tree` is extracted and it's value is changed from "Pathfinding:Stray" to "WildAnimals:Flee". The exact reverse happens when the deer reaches a safe distance from the player, i.e. `BehaviorComponent.tree` is changed from "WildAnimals:Flee" to "Pathfinding:Stray".
Whenever a change is made, the entity's (deer's) new entry is added to the `entityInterpreter` defined in the `BehaviorSystem`. (Using the `onBehaviorActivated` and `onBehaviorDeactivated` methods)
Now the `BehaviorSystem` has a `update` running, which gets called from time to time. The `update` fetches values from the `entityInterpreter`.
`entityInterpreter` is a hashMap, mapping entityRefs to their behaviorComponent.
So the problem that surfaces on every second or third run is that I get a `java.util.ConcurrentModificationException`, which I reckon means that the `onBehaviorDeactivated` makes a change to `entityInterpreter` just while the `update` method tries to read it.

This PR changes the `Map` to `ConcurrentHashMap` in the `BehaviorSystem`, which solves the above issue.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
+++ b/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import org.terasology.assets.ResourceUrn;
@@ -73,7 +74,7 @@ public class BehaviorSystem extends BaseComponentSystem implements UpdateSubscri
     @In
     private AssetManager assetManager;
 
-    private Map<EntityRef, Interpreter> entityInterpreters = Maps.newHashMap();
+    private ConcurrentHashMap<EntityRef, Interpreter> entityInterpreters = new ConcurrentHashMap<EntityRef, Interpreter>();
     private List<BehaviorTree> trees = Lists.newArrayList();
 
     @Override
--- a/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
+++ b/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
@@ -15,17 +15,7 @@
  */
 package org.terasology.logic.behavior;
 
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


GooeyHub:

flo:
Have you checked via debugging why this is necessary? Won't it only be called from the same thread? The one that runs the main loop? e.g. have you figured out which thread and method is causing the concurrent modificaiton besides the main thread?

GooeyHub:

nihal111:
Leaving slack discussion here-
This has nothing to do with different threads. The entityInterpreter being iterated over in the BehaviorSystem loop is removed at some point when the UpdateBehaviorEvent is called (Everything happens in the main thread).

oniatus:
To summarize: `BehaviorSystem.update` is called which somehow ends in a call to `onBehaviorActivated` due to changes in the behavior strategy of an entity, causing an `ConcurrentModificationException`.

Can't we just decouple the update and remove operations and store the entities and behavior components until the next update?
At the beginning of the update, we could then drain the Maps? with the entities and components to add and to remove and proceed with a default update cycle.

This does pretty much the same because ConcurrentHashMap does not reflect add operations in the iterator, meaning that added entities are first visible in the next `update` call.
We have a little overhead for two extra collections but only if behaviors are activated and deactivated, compared to the likely more expensive calls to the ConcurrentHashMap (evaluation would be nice, I found only this one: http://stackoverflow.com/a/32141829).

In case we stick to the ConcurrentHashMap, it could make sense to use the advanced constructor and pass a concurrencyLevel less than 16 as the system is not designed for multithreading anyway :)

GooeyHub:

nihal111:
@oniatus Thanks for the performance tip.
After a discussion with @flo on Slack, we decided to come up with a new way to deal with this. Since the loop inside update was changing the Interpreters while iterating over it, the ConcurrentModificationException was raised. The new method deals with this by having a cachedInterpreters list which is set to empty every time a Component event handler (OnChangedComponent/ OnActivatedComponent/ BeforeDeactivateComponent) makes a change to any interpreter. Further, in the `update()` method we now fill the cachedInterpreters list with `entityInterpreters.values()` (values of the hash map). This list- `cachedInterpreters` is used for iteration in the subsequent for loop (where changes are made to the entityInterpreters).

I've tested this with around 10 deer, and it seems to work without any issues :)

GooeyHub:

Issue #2921, opened 2017-05-01
Small translation updates required


ar0ne:
Hi All,

This issue related PR #2920, where we decided to change part of greeting phrase ("enter-username-message") 
**From**: Would you like to use a non-default username?
**To**: Would you like to update your username?

You can see it when first visit of Join/Host game menu. 
At present we need small changes in our translations for next languages:

- [ ] cs (Czech)
- [ ] fa (Farsi)
- [ ] fr (French)
- [ ] hi (Hindi)
- [ ] id (Indonesian)
- [ ] ko (Korean)
- [ ] nl (Dutch)
- [ ] pr (Pirate) :D
- [x] Russian (recently checked by Translator5)
- [x] German (recently checked by Translator5)

I tried to mark this word as "fuzzy" in weblate (http://translate.terasology.org/projects/terasology/menu/), but it looks like we don't have all requested languages there. Correct me if I'm wrong, guys?

Translator5:
I now fully re-translated the German translation manually.
(With open Terasology on the one desk and Weblate on the other one.)
Now there are much more context-related phrases!
*As soon as GitHub will recognize them, I will create a new Pull request!

Based on the German translation I will try to do the Russian translation soon.
The current Russian translation is full of false-context translations or totally false ones!
*Note to all translators;
Do not use just the english-source, but also use other languages to get the best translation!

Cervator:
Bump - merged the first PR and Weblate has refreshed now. 

@Translator5 could you please double check on Russian? I had to resolve some conflicts manually and a few might have gone with a worse option. Hard for me to tell which is better :-)

There are other known issues where some languages aren't showing up in Weblate right. Not sure at present how to fix, need somebody able to spend some time poking at the server itself, maybe reset/refresh something else.

Changing the bullet list to a checkbox list and marking off German and Russian, hope that's accurate as of now.

Translator5:
Some issues, belonging to the translation;
- we need separate name-strings for the "Player-Name:"-input and the "Behaviour"- as well as the "NUI-Editors"-title-input - because of translation issues!
- we should change the Settings in the NUI-Editors from "disable XYZ" into "enable XYZ" or "toggle XYZ" cause of translation issues (it looks better if settings just take one text-line, instead of 2 lines)

Cervator:
[...]

PR #2937, opened 2017-05-06
german and russian translation update

Translator5:
just merge the german and russian file commits
#2921


oniatus:
German changes look good to me, I added a small update for the world-seed, as `Seed` or `Weltgenerator Seed` are common terms for games like this.
I think Minecraft changed it to `Startwert für den Weltgenerator` which is another option.

Cervator:
Grabbed latest from Weblate local (some uncommitted changes) and merged locally.

There were some conflicts in the Russian file - hopefully I took a fair mix but it might need another look sometime.

Pushed to both `develop` and `weblate` so everything should be in sync again. Thanks! :-)

PR #2545, opened 2016-10-15
Fix for background music not looping

njd5475:
### Contains

Fixes #2527 
### How to test

Start Game wait for main menu track to finish and start again. Approximately 3:31 seconds.
Start Game music should stop
Exit to menu background music should start again

PR #2542, opened 2016-10-14
Loop main menu background music

arpan98:
### Contains

Fix for #2527 
### How to test
- Open game
- Wait for music to stop 
- Wait for more than 3 minutes 29 seconds (Duration of Menu Theme Song)

PR #2945, opened 2017-05-08
Add loop functionality to AudioManager, loop menu theme

### LINKED ISSUE #2527: Main menu background track not looping

oniatus:
Contains snippets and changes from PRs #2846, #2545 and #2542 which are all inactive for more than a month. I would like to continue here and close the old PRs.

This fixes #2527. 
I was unable to reproduce #2364, so maybe @ratz or @Cervator can test this again?

Note that the change in the `AudioEndListener` will require a follow-up PR in the [MusicDirector](https://github.com/Terasology/MusicDirector) module.
If you want to keep the API untouched we could either introduce a new listener for the interrupt action on a music track or just omit the end listener entirely when music stops early.

### DIFF
--- a/engine/src/main/java/org/terasology/audio/AudioEndListener.java
+++ b/engine/src/main/java/org/terasology/audio/AudioEndListener.java
@@ -17,6 +17,7 @@ package org.terasology.audio;
 
 /**
  * Is called by AudioManager to inform about end of playing music or sound.
+ * The listener will only be called once. If the same track is played multiple times, a new listener has to be used each time.
  * <br><br>
  * Notice: The code in onAudioEnd is run in the update() method of the AudioManager once the sound/music ends playing.
  */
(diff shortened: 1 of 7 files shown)


Cervator:
Tests out fine, thanks :-)

For MusicDirector we can just fix it. I haven't looked but figure we just need to add a parameter to a method call or two? If you can note the details real quick I can fix it while merging, or you can throw a PR together for the module real quick, your call.

oniatus:
https://github.com/Terasology/MusicDirector/pull/1 there you go :)

Gave it a short test over 3 or 4 tracks in singleplayer and switched back and forth to the menu and ensured that the menu theme still looped. Looks good to me.

Cervator:
Thanks! Merged and am closing old related issues (with those authors also added to the credits for their earlier efforts)

PR #2947, opened 2017-05-12
Dependency injection tests

oniatus:
Adds some learning tests how the constructor injection works and javadoc for the helper method which documents the behavior.
Also fixes a case for a nullpointer, the original system was unable to fallback to a constructor with less parameters, even though it did all the sorting logic and would try each one. It's easy to reproduce, just undo the `of()`->`ofNullable()` change in the `InjectionHelper` and re-run the tests.

I was thinking about using it to initialize component systems but this is more complicated as these may have dependencies between each other which is handled different by the current system so I am going to leave it with the tests for now.

Should be ready to merge the way it is.


GooeyHub:

Cervator:
Nice and easy. Merged, ran the new tests, ran the game real quick, all appears well. Thanks :-)

Issue #2893, opened 2017-04-10
Serialization of FlexibleConfig


vampcat:
With PR #2757 (Flexible Config) almost being ready to merge, it's about time we start looking at the future of FlexibleConfig, and the opportunities available for it's immediate use.

The very first opportunity presents itself in the form of "Node Parameters". The rendering pipeline consists of various `Nodes`, the parameters of which can be tweaked by the user. Currently, these parameters are exposed via `RenderingConfig`, and that works for static Nodes. However, since we want to provide the user with the option to add/replace Nodes via Modules, this approach will no longer work, and we'll need an easier way to expose the options. Enter the FlexibleConfig (FC), and everything becomes simple.

However, the current PR only deals with the actual implementation of FC, and not on storing/retrieving data from the disk. This aspect of the class, the "Serialization of FC", is of utmost importance before this class can be used for practical purposes, since this is what makes the user settings persist across game sessions.

The basic approach to serialization of FC will closely resemble the approach followed by RenderingConfig, but ideally the storage/retrieval of settings for a particular Module will be independent of the storage/retrieval of settings for another Module, so that we don't have to load the settings for every Module when requesting the settings of one Module.

Ideally this would be done (at least in part) before the start of GSoC, so that this class can be used during the work period.

Pinging @eviltak @emanuele3d @Cervator for feedback, and suggestions.

[This issue is a WIP]

emanuele3d:
What follows is a proposal for the replacement of the current description of this issue. This comment and any following related to this meta-issue will eventually be deleted as to unclutter the space for the discussion on the issue proper.

----
With PR #2757 almost ready to merge, it's time we start looking at the future of FlexibleConfig.

The currently open PR only deals with the actual implementation of FC, and not on storing/retrieving data from the disk. This aspect is of utmost importance before this class can be used for practical purposes, since this is what makes settings persist across game sessions.

Here are the descriptions of a number of different, co-existing FCs:
1. **global engine settings**: these are settings currently stored in classes such as RenderingConfig and RenderingDebugConfig. An FC instance automatically stores the engine's hardcoded setting values and any value the user might have changed from the hardcoded ones. We'd want to store to disk only the user-customized ones in a *.cfg file in the user's home folder.
[...]

PR #2948, opened 2017-05-15
Input and Binds refactoring

oniatus:
Goal of this PR:

- Move all binding logic in a separate manager and out of the input system and the config class
- Get the input and binding system in a state where it can be tested without writing ~120 lines of mocking
- Prepare the input system for the flexible config
- Maybe split the input system in separate implementations for mouse/keyboard/controller

I'll expect this to become a bit bigger and will try to work in commits as small as possible with a testable state between committs. Current state seems to work, even as it excludes the logic to load and save bindings.

So far, I moved most of the logic from `BindsConfig` to a new `BindsManager` which is implemented as a separate engine subsystem. I found lot's of the code being close to redundant and will do a second iteration over the code later.

Feel free to leave early comments on the API, especially the `BindsManager`. I will add a separate comment when the PR would be ready for a deeper review before merging.


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/engine/subsystem/config/BindsManager.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import java.util.Map;
+
+public interface BindsManager {
+
+    BindsConfig getBindsConfig();
+    
+    BindsConfig createDefault(Context context);
+    
+    void updateForChangedMods(Context context);
```
Cervator: Just for consistency's sake should this be `Modules` rather than `Mods` ? Historically we've tried to make (container) modules distinct from (gameplay) mods. Probably a somewhat doomed battle, I admit.

engine/src/main/java/org/terasology/engine/subsystem/config/BindsSubsystem.java
```diff
@@ -0,0 +1,441 @@
...
+            }
+        }
+        return config;
+    }
+
+    /**
+     * Updates a config with any binds that it may be missing, through reflection over RegisterBindButton annotations
+     */
+    @Override
+    public void updateForChangedMods(Context context) {
```
Cervator: The `Mods` vs `Modules` thing again

Cervator:
Also pinging @eviltak directly, can't tag as reviewer, need to accept org invite first :-)

oniatus:
>I see a lot of redundancy in the BindsSubsystem that can be removed. The createDefault and updateForChangedMods methods are almost the same except a couple of lines.

True, the only difference between `update*For` and `add*DefaultsFor` is the check `!binds.hasBinds(bindUri)` from the update methods.

I think we can keep this check in both methods, maybe even log a warning if binds are added duplicate times. It makes little sense to add the same bind multiple times to the config, even for default values. I'll test this next.

GooeyHub:

oniatus:
I added a cached default binding config. The defaults are used in the ChangeBindingPopup which now opens instantly instead of taking ~1-2 seconds to resolve the entire module environment.
We update the bindings only during the `postInitialize` phase in the `LwjglInput`-system, so this should break nothing and also work if one downloads a module while joining a server.

Next step will be the connection from the BindManager to the inputsystem, the method ` public void applyBinds(InputSystem inputSystem, ModuleManager moduleManager)` can probably hide both parameters as the input system is not used at all and the moduleManager is available on the context.

GooeyHub:

GooeyHub:

oniatus:
I added save and load logic, similar to the previous config file which saves the BindsConfig to `bindsConfig.cfg`. We can replace these two methods as soon as #2893 is implemented.
The BindsConfig itself contains now little logic, so it should be easy to switch to the FlexibleConfig as baseclass.

Secondly, i added a separate loading state in the engine. The binds manager takes a second or so to initialize and this will now show `Post-Initializing Binds subsystem` instead of `Initializing Asset Management...`.

The BindsManager/BindsSubsystem is feature-complete from my side. 

Next step is a cleanup of the InputSystem itself.

oniatus:
First batch of cleanup in the InputSystem is done, I also added a minor fix for #2949, the input entities are now fetched once per update cycle instead of once per binding.

GooeyHub:

oniatus:
`MouseAxisEvent` is now a single class instead of a separate one for each axis.
There were no special cases where it made sense to have one event class per axis and we still send one event per axis which are distinguished by an enum as before.
I also changed the class to a singleton with a reset function, similar to `MouseButtonDownEvent`, so we no longer need ~200 event instances only to turn the character once by 360 degree 😄

GooeyHub:

GooeyHub:

GooeyHub:

oniatus:
InputSystem can now be tested. I added only tests for basic keyboard input and one binding so far and don't think i will cover the entire class with tests. However, this should leave enough space to implement enhancements like #2859 in a test-driven way. Mouse and Controller input or bindings can be tested similar, without architectural changes,

Last step are similar tests for the BindsSubsystem, then I'll concider this mergeable.

GooeyHub:

GooeyHub:

oniatus:
Gave it a test again, actually i wrote https://oniatus.github.io/developing/input.html on these changes, so input should work as before :)

Ready to merge from my side, TL;DR changes:

- Input system and config behave as before, except less delay when opening the binding popup
- State notification when game loads for the post-initialization of systems
- Input handling and binds are now handled in separate classes
- Unit Tests for InputSystem (Keyboard input -> Event or Binding update as output)
- Unit Tests for BindsSubsystem (register binding classes from module environment)
- Facades for InputConfig and BindsConfig, we can use them later for the FlexibleConfig migration
- Much cleanup, javadoc and small stuff

Cervator:
Poking to see if anybody can give this another look: @eviltak @skaldarnar @msteiger @flo @ThompsonTyler @emanuele3d @vampcat

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
```diff
@@ -394,7 +394,8 @@ public synchronized void run(GameState initialState) {
     private void mainLoop() {
         PerformanceMonitor.startActivity("Other");
         // MAIN GAME LOOP
-        while (tick()) { /* do nothing */ }
+        while (tick()) {
+            /* do nothing */ }
```
vampcat: Maybe have the bracket on a newline?

engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
```diff
@@ -292,14 +292,13 @@ public void onPlayerSpawn(OnPlayerSpawnedEvent event, EntityRef character) {
...
-    public void onMouseX(MouseXAxisEvent event, EntityRef entity) {
-        lookYawDelta = event.getValue();
-        event.consume();
-    }
-
-    @ReceiveEvent(components = CharacterComponent.class)
-    public void onMouseY(MouseYAxisEvent event, EntityRef entity) {
-        lookPitchDelta = event.getValue();
+    public void onMouseX(MouseAxisEvent event, EntityRef entity) {
+        MouseAxis axis = event.getMouseAxis();
```
vampcat: I might be missing something, but shouldn't this just be onMouse(), since we're handling both X and Y axis here?

engine/src/main/java/org/terasology/input/InputSystem.java
```diff
@@ -142,254 +113,83 @@ public void setControllerDevice(ControllerDevice controllerDevice) {
...
-    public void setCapturingMouse(boolean capturingMouse) {
-        this.capturingMouse = capturingMouse;
+    private void updateInputEntities() {
+        inputEntities = new EntityRef[] {localPlayer.getClientEntity(), localPlayer.getCharacterEntity()};
     }
 
     private void processMouseInput(float delta) {
-        if (!isCapturingMouse()) {
+        if (!display.hasFocus()) {
             return;
```
eviltak: I think that extracting the `display.hasFocus()` call to a `inputSystem.isCapturingMouse()` method is a better idea than simply calling `hasFocus` everywhere as it makes the intent much more clear.

engine/src/main/java/org/terasology/rendering/nui/internal/NUIManagerInternal.java
```diff
@@ -495,7 +496,7 @@ public void update(float delta) {
             widget.update(delta);
         }
         InputSystem inputSystem = context.get(InputSystem.class);
-        inputSystem.getMouseDevice().setGrabbed(inputSystem.isCapturingMouse() && !(this.isReleasingMouse()));
+        inputSystem.getMouseDevice().setGrabbed(display.hasFocus() && !(this.isReleasingMouse()));
```
eviltak: Same as above.

oniatus:
Thanks for the review, PR is finally up to date again :)

GooeyHub:

Cervator:
Bump - should we go ahead and merge this one? Does it bowl over #2859 or can that then be rebased on top and accepted after? Pinging @ThompsonTyler :-)

oniatus:
Ready to be merged from my side (maybe v2?). It will break #2859 but we can port most of it to a new pr if needed.
I think it is more easy to re-implement the feature and take the best ideas off the PR instead of rebasing it:

Compare https://github.com/ThompsonTyler/Terasology/blob/c5defdf55f338f0afb7b6ad0237430b71fdceec2/engine/src/main/java/org/terasology/input/InputSystem.java#L500
to
https://github.com/oniatus/Terasology/blob/fb2b9537ac0cdbf73d9a83d16e565e326124d6ab/engine/src/main/java/org/terasology/input/InputSystem.java#L277
There is no way I can rebase the first onto the second while preserving the feature and at least something close to readable code 😉

ThompsonTyler:
Yeah, I was not thinking that #2859 would be usable with ANY changes to the input system. But shouldn't be too hard to convert it to the new system with most of the heavy work already being completed. Will be able to work on it once the school year starts/GSOC finishes up if no one else has grabbed it by then

emanuele3d:
Just went through the diff to see if I can understand what's going on.... and I can't. This part of the codebase is too unfamiliar to me. So, nothing to do with your work @oniatus, I've just been in the pixel forge for too long. :wink: 

The only thing I'd say is: I've noticed a good number interfaces, classes and methods with no javadocs. I'd invest some time taking care of those. Of course many things are self-explanatory for you, and perhaps they are for many. But a class-level javadoc, giving a bit of context about what a class or interface is for, and how it fits in the bigger picture, would be useful, at least for people unfamiliar with this aspect of the engine.

GooeyHub:

oniatus:
Went over it once again and added more javadoc :)

engine/src/main/java/org/terasology/config/BindsConfig.java
```diff
@@ -133,15 +129,15 @@ public BindsConfig deserialize(JsonElement json, Type typeOfT, JsonDeserializati
...
+            for (SimpleUri key : src.bindUriToInput.keySet()) {
                 bindByModule.put(key.getModuleName(), key);
             }
             List<Name> sortedModules = Lists.newArrayList(bindByModule.keySet());
             Collections.sort(sortedModules);
             for (Name moduleId : sortedModules) {
                 SetMultimap<String, Input> moduleBinds = HashMultimap.create();
                 for (SimpleUri bindUri : bindByModule.get(moduleId)) {
-                    moduleBinds.putAll(bindUri.getObjectName().toString(), src.data.get(bindUri));
+                    moduleBinds.putAll(bindUri.getObjectName().toString(), src.bindUriToInput.get(bindUri));
```
emanuele3d: Similar concern here with the naming of variables and verbs vs nouns: I'd rename bindUri to bindingUri.

engine/src/main/java/org/terasology/config/facade/BindsConfiguration.java
```diff
@@ -10,24 +10,54 @@
 import java.util.List;
 
 /**
- * Facade for {@link Config#getBinds()}
+ * Facade for {@link Config#getBinds()}.
+ * The binds configuration holds the mapping from bind uris.
+ * The {@link SimpleUri} for a binding contains the module from the binding and the id from the binding annotation, 
+ * e.g. {@link RegisterBindButton.
```
emanuele3d: Missing curly bracket?

engine/src/main/java/org/terasology/config/facade/BindsConfiguration.java
```diff
@@ -10,24 +10,54 @@
 import java.util.List;
 
 /**
- * Facade for {@link Config#getBinds()}
+ * Facade for {@link Config#getBinds()}.
+ * The binds configuration holds the mapping from bind uris.
```
emanuele3d: ...the mapping from bind(ing?) uris.... to what? It feels like the sentence is cut short.

engine/src/main/java/org/terasology/config/facade/BindsConfiguration.java
```diff
@@ -10,24 +10,54 @@
...
  */
 public interface BindsConfiguration {
 
+    /**
+     * Returns true if the input is bound to a bind uri.
+     */
     boolean isBound(Input newInput);
 
+    /**
+     * Sets all bindings to the mappings from the given config.
```
emanuele3d: I'd write: 

` * Sets all bindings to the mappings contained in the given config.`

engine/src/main/java/org/terasology/config/facade/BindsConfiguration.java
```diff
@@ -10,24 +10,54 @@
...
+     */
     void setBinds(BindsConfig other);
 
+    /**
+     * Returns a list of all input, bound to the given bind uri.
```
emanuele3d: Returns a list of all input**s** bound to the given ~~bind~~ uri.

engine/src/main/java/org/terasology/config/facade/BindsConfiguration.java
```diff
@@ -10,24 +10,54 @@
...
+     */
     List<Input> getBinds(SimpleUri uri);
 
+    /**
+     * Returns the internal binds config.
```
emanuele3d: I take it binds and bindings are used interchangeably in this part of the codebase. I checked on dictionary.com and the term "bind" can be a noun. I just feel it's a bit ambiguous and here I'd recommend the term "bindings" instead, leaving "bind" to be used only as a verb. But I don't want to increase too much the scope of this PR - sorry for the diversion into the dictionary.

engine/src/main/java/org/terasology/config/facade/BindsConfiguration.java
```diff
@@ -10,24 +10,54 @@
...
     BindsConfig getBindsConfig();
 
+    /**
+     * Returns true if the configuration contains any bindings for the given uri.
+     */
     boolean hasBinds(SimpleUri uri);
 
+    /**
+     * Binds one or more inputs to one bind uri.
+     * If the inputs are used for other uris, they will be removed and used for the new bind uri.
```
emanuele3d: I'd write:
```
* Binds one or more inputs to one uri.
* If the inputs are bound to other uris they will be disassociated from them first.
```
And use the same for the next method.

engine/src/main/java/org/terasology/config/facade/BindsConfiguration.java
```diff
@@ -10,24 +10,54 @@
...
+     */
     void setBinds(SimpleUri bindUri, Iterable<Input> inputs);
 
-    Collection<Input> values();
+    /**
+     * Returns a collection with all bound input values.
```
emanuele3d: Returns a collection with all bound inputs ~~values~~.

The fact that the inputs are the values in a map is irrelevant here.

engine/src/main/java/org/terasology/config/facade/BindsConfiguration.java
```diff
@@ -10,24 +10,54 @@
...
+    /**
+     * Binds one or more inputs to one bind uri.
+     * If the inputs are used for other uris, they will be removed and used for the new bind uri.
+     */
     void setBinds(SimpleUri bindUri, Iterable<Input> inputs);
 
-    Collection<Input> values();
+    /**
+     * Returns a collection with all bound input values. 
+     * Changes to this collection may be visible to the underlying data, therefore it should be used for read-only.
```
emanuele3d: "underlying data"? What do you mean by that? Perhaps you meant something like "other consumers of the data"?

Also, you might want to look into using Collections.unmodifiableCollection() and enforce readonly-ness of the returned collection.

engine/src/main/java/org/terasology/config/facade/InputDeviceConfiguration.java
```diff
@@ -3,21 +3,38 @@
...
     float getMouseSensitivity();
 
+    /**
+     * Sets the mouse sensitivity, the value should be in [0,1].
+     */
     void setMouseSensitivity(float mouseSensitivity);
 
-    void reset(Context context);
+    /**
+     * Resets the the mouse sensitivity and axis settings from the default configuration.
```
emanuele3d: `* Resets the the mouse sensitivity and axis settings back to the default values.`

engine/src/main/java/org/terasology/config/facade/InputDeviceConfiguration.java
```diff
@@ -3,21 +3,38 @@
...
+     */
+    void reset();
 
+    /**
+     * Returns true if the y axis is inverted.
+     */
     boolean isMouseYAxisInverted();
 
+    /**
+     * Sets the y axis to be inverted.
```
emanuele3d: `* Sets the y-axis as inverted (true) or resets it back to normal (false).`

engine/src/main/java/org/terasology/engine/subsystem/config/BindsSubsystem.java
```diff
@@ -178,7 +178,7 @@ private void addDefaultBindings(SimpleUri bindUri, Class<?> event, BindsConfigur
 
     private List<Input> fetchDefaultBindings(Class<?> event, BindsConfiguration config) {
         List<Input> defaultInputs = Lists.newArrayList();
-        Collection<Input> values = config.values();
+        Collection<Input> values = config.getBoundInputs();
```
emanuele3d: `values` (generic map term) --> `boundInputs` (the actual content of the collection).

GooeyHub:

oniatus:
Nah, review is highly appreciated. If code is hard to understand for someone unfamiliar with the codebase it is not clean code, period 😎 
At the moment there is some duplication between `BindsConfig` (the old config) and `BindsConfiguration` (the new interface).
I want to remove the BindsConfig entirely when we implement the FlexibleConfig. Thats why i ripped all logic except data management out of it in this PR.
On a second though, we may rename the new configuration to `BindingConfiguration`, i like it much more than `BindsConfiguration`, same for all the interface and variables. For now `Binds` is okay for me but I want to keep it in mind when I touch it again in a follow-up PR.

engine/src/main/java/org/terasology/config/BindsConfig.java
```diff
@@ -31,23 +30,27 @@
...
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import static java.util.stream.Collectors.toList;
+
 /**
- * User binds configuration. This holds the key/mouse binding for Button Binds.
+ * The binds configuration holds the mapping from binding uris to Inputs.
+ * The {@link SimpleUri} for a binding contains the module from the binding and the id from the binding annotation,
```
emanuele3d: "contains the module from the binding": how can a module be _from_ a binding? Can this be clarified?

You also mention "the id from the binding annotation", I feel this and the previous line could use a little bit of context and further explanation.

engine/src/main/java/org/terasology/config/facade/BindsConfiguration.java
```diff
@@ -23,12 +25,12 @@
...
 
     /**
-     * Sets all bindings to the mappings from the given config.
+     * Sets all bindings to the mappings contained in the given config.
      */
     void setBinds(BindsConfig other);
 
     /**
-     * Returns a list of all input, bound to the given bind uri.
+     * Returns a list of all input, bound to the given uri.
```
emanuele3d: Returns a list of all input**S** bound to the given uri.

GooeyHub:

GooeyHub:

oniatus:
Rebased on latest develop, tested again and included the latest feedback.
Done from my side :)

engine/src/main/java/org/terasology/config/BindsConfig.java
```diff
@@ -45,8 +48,13 @@
...
  * The binds configuration holds the mapping from binding uris to Inputs.
- * The {@link SimpleUri} for a binding contains the module from the binding and the id from the binding annotation, 
- * e.g. from {@link RegisterBindButton}.
+ * The {@link SimpleUri} for a binding contains the module id where the binding is defined and the id from the binding annotation, 
+ * e.g. from {@link RegisterBindButton} as object name.
+ * <p>
+ * One example for a binding-input combination is the {@link ForwardsButton} 
+ * which is defined in the engine with the id <code>forwards</code>.
+ * The default input binding for the forward movement is the W-Key.
+ * Therefore the binds for <code>engine:forwards</code> would contain an {@link Input} object with type {@link InputType#KEY} and id {@link KeyId.W}.
```
emanuele3d: :+1:

Cervator:
Oh boy oh boy! This is a big one and I finally get to merge it. Woo! :D

Tested locally in whichever ways I could think of to challenge the input / binds systems, including hooking up a game controller. While said controller sucks and only half the stuff really works, I'm pretty sure that's the controller as that's about how I remember it working before ;-)

Yay merging! Thanks @oniatus and everybody else involved in this one :-)

emanuele3d:
Thanks Cervator for the merge, and thanks @oniatus for the work and the patience! =)

Issue #2555, opened 2016-10-20
Crouch follow-ups


nihal111:
As a follow up to #2544:
- [ ] Add falling off ledge prevention.
- [ ] Change key mappings as per discussion [here](http://forum.terasology.org/threads/crouch-and-prone-controls.1622/)
##

Cervator:
Some more:

- [ ] Switch back out of the temporary `x` key (covered by the key mapping note above but just to get an explicit reminder)
- [ ] Tiny / general: Clear out any empty javadoc headers when making changes anyway (like `/** */` leftover from a script removing author tags)

oniatus:
Closing as proposed in https://github.com/MovingBlocks/Terasology/pull/2961#issuecomment-341993601
Follow up issue is #1075

PR #2961, opened 2017-05-25
Remove hackish and not multiplayer ready crouch visuals.

flo:
### Contains

This removes the constant camera up/down movement in multiplayer when you crouch.

It also makes it so that while "default croaching" is enabled with key "x", presseing the temporary croach button "ctrl" will make the character temporarly stop crouching. This is analog to how it is for running.

This change removes all the hackish existing player height and eye height modificaitons during crouching. The old code had the following issues:
- the client ordered the server to change the eye height based on possibly outdated eye height values it had received previously from the server
- character height and eye height change got triggered by client instead of being a result of movement mode change
- client triggered redundant movement mode changes that would also
  automatically result from movement input
- movementDebugCommands were used to do eye height and height modification. The commands are hackish too and not really something that should be used in non debug code.
- the player height change triggered a teleportation of the player which messes up further with the movement prediction system

### How to test

Start a headless server and join as client. Then try crouching. Movement speed should be slightly slower while crouching.


GooeyHub:

GooeyHub:

nihal111:
I do agree that most of the code was 'hacky' and not ready for multiplayer. However, this change simply destroys the point of "crouching" which is reduction of height. The shift (Run/Walk toggle) button probably would do the exact same thing as crouching now.

If you want to remove this for the sake of removing the multiplayer bug or if you plan to initiate a movement refactoring rework soon, I don't mind going ahead with this. Otherwise, it is a step in the opposite direction and simply undoes things.

flo:
I think the main intention of crouching was to have a "do not fall of edge" feature which was not implemented yet anyway.
The size reduction is not something that will be needed often / at all. 

The current code just solves it in the wrong way. Having it there just gives other developers bad examples of how to do stuff.

Cervator:
I agree the code was problematic in multiplayer, but it did work really well in single player. The crouching for height differences is also desirable imho :-)

Could we live with a workaround for multiplayer (disable crouching entirely if needed) along with some big visible TODOs while keeping it functional as-is in single player? Or is that too "code smelly" ? I don't know who might have time to finish this since it isn't part of GSOC which will eat most of our time.

Or would it be better to park the code elsewhere entirely and wrap it into the overall movement refactoring whenever we get somebody active on that area?

portokaliu:
I think crouching actually allows you to pass through 1x1 areas (like if we had vents or small tunnels). It would be useful if we ever wanted to make adventure maps with secrets, traps or w/e else.

Completely removing it for now is ok, but the idea is good and we should get back to it when possible.

GooeyHub:

skaldarnar:
Could be a really cool issue to work on right after GSOC. Movement refactoring is a bigger topic, so we could try to get students, mentors, and anybody interested work together on this right after the coding period ended, to not loose the focus and drive we just gained.

jellysnake:
Seeing as it's raising bugs like #3108 I think it would be good to merge this and then make it some kind of gsoc task

GooeyHub:

oniatus:
Fixed the conflict and gave it a short test.
This change is more like a sneak than a crouch but it works and my personal bias is to have a working feature with clean code rather than a broken one with more bad code behind.

I suggest to merge this PR, then close #2555, #2651, #3108 and continue on #1075 (it is the only one with a bounty) with the requests from the named issues.

Full list for crouching for me would be:
- Multiplayer support (host & connected client)
- 1x1 holes enterable
- Visual Feedback in eye height
- Falling of ledges support
- Toggle/Untoggle + temporary button (as is with this PR)
- Leave 'LocalPlayerSystem' cleaner instead of making it worse.

For the last point it would make sense to clean up the system first (after this PR) which is something i can work on.
I will leave this PR open until next weekend for extra feedback (ping @Cervator @jellysnake @skaldarnar @nihal111 @flo) then merge and continue as proposed if no one raises an objection 😉

oniatus:
Left it open a bit longer :)
I will now update the issues. The follow-up will be #1075.

flo:
Thanks for merging

PR #2965, opened 2017-05-29
Client identity storage service

gianluca-nitti:
### Contains

The client to synchronize with a client identity storage service, as discussed [here in the forum](http://forum.terasology.org/threads/client-identity-cloud-storage-service.1846/).

### How to test

An implementation of the backend server is available [here](https://github.com/gianluca-nitti/terasology-key-server). You can install and configure it locally as described [here](https://github.com/gianluca-nitti/terasology-key-server/blob/master/docs/install.md), or use the public instance I'm hosting [here](http://165.227.140.7/).

To test, register an account on a server by navigating to it's address with a browser (or directly clicking the link above if you use the hosted instance). Then, start the game client (from this branch) and login by going to Settings > Player Settings. Now you can join a game server and as soon as you do, the new identity will be uploaded to the service; you can then start another client with empty/different configuration (e.g. use the 2nd or 3rd client run configurations in IntelliJ, or specify another data directory from the command line) and login into it too; your client identities will be synchronized and if you join the same game server, you will find your character in the same position you left it from the other client.

~~Marking as WIP since I think there is still some minor improvement that can be done, e.g. localization of some UI messages. Also, the conflict resolving feature (which allows you which identity you want to keep for a certain server in case the locally stored one conflicts with the one stored on the service) is not fully working but this is due to a server issue which doesn't allow to overwrite the identities (will fix it soon).~~

**Update:** I just added the localization feature (the strings are in the menu translations files; the values are filled for English only but can be easily be added for the other languages via editing the proper files or - I think - with Weblate if this branch is merged). Also fixed the [server](https://github.com/gianluca-nitti/terasology-key-server) to support replacing of already uploaded identities.

Tagging @msteiger since we talked about this recently on appear.in.


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/identity/ClientIdentity.java
```diff
@@ -49,4 +50,21 @@ public PrivateIdentityCertificate getPlayerPrivateCertificate() {
         }
         return playerPrivateCertificate;
     }
+
+    @Override
+    public int hashCode() {
+        return playerPrivateCertificate.hashCode() ^ playerPrivateCertificate.hashCode();
```
msteiger: I think that this will always be zero. Maybe just just `31 * playerPublicCertificate` instead?

engine/src/main/java/org/terasology/identity/storageServiceClient/Action.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.identity.storageServiceClient;
+
+/**
+ * Represents an interaction that can be made with the storage service server.
+ */
+abstract class Action {
```
msteiger: Either make it an interface or use `Consumer<StorageServiceWorker>` instead.

engine/src/main/java/org/terasology/identity/storageServiceClient/BigIntegerBase64Serializer.java
```diff
@@ -0,0 +1,53 @@
...
+    private static final Base64.Encoder ENCODER = Base64.getEncoder();
+    private static final BigIntegerBase64Serializer INSTANCE = new BigIntegerBase64Serializer();
+
+    public static BigIntegerBase64Serializer getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public BigInteger deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+        return new BigInteger(DECODER.decode(json.getAsString().replace("\n", "")));
```
msteiger: Why could there be newline chars. in the data?

engine/src/main/java/org/terasology/identity/storageServiceClient/HttpMethod.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.identity.storageServiceClient;
+
+/**
+ * Http methods; only the relevant ones for now but can easily be extended with the other standard ones.
+ */
+enum HttpMethod {
```
msteiger: Maybe complete the list and make it public? There are not so many HTTP Methods

engine/src/main/java/org/terasology/identity/storageServiceClient/ServiceAPIRequest.java
```diff
@@ -0,0 +1,89 @@
...
+import java.math.BigInteger;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+/**
+ * Utility class to perform requests to the service API.
+ */
+final class ServiceAPIRequest {
+
+    private static final Gson GSON = new GsonBuilder().registerTypeHierarchyAdapter(BigInteger.class, BigIntegerBase64Serializer.getInstance()).create();
```
msteiger: Can you reuse one of the existing instances of GSON? Not sure if it's possible or makes sense ..

engine/src/main/java/org/terasology/identity/storageServiceClient/ServiceAPIRequest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private static boolean isSuccessful(int code) {
+        return code >= 200 && code < 300;
+    }
+
+    private static void parseError(HttpURLConnection conn) throws IOException, StorageServiceException {
+        try (InputStream errResponse = conn.getErrorStream()) {
+            try {
+                throw new StorageServiceException(GSON.fromJson(new InputStreamReader(errResponse), ErrorResponseData.class).error);
+            } catch (RuntimeException e) {
```
msteiger: Can we be more specific here? Who could even throw during throwing an exception?

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceException.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.identity.storageServiceClient;
+
+class StorageServiceException extends Exception {
```
msteiger: I think that exceptions should always be public. If it's not, callers cannot know about it which conflicts with the idea of having a custom exception.

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceWorker.java
```diff
@@ -0,0 +1,183 @@
...
+public final class StorageServiceWorker {
+
+    private static final Logger logger = LoggerFactory.getLogger(StorageServiceWorker.class);
+
+    StorageServiceWorkerStatus status = StorageServiceWorkerStatus.LOGGED_OUT;
+    APISession sessionInstance;
+    String loginName;
+    Stack<IdentityBundle> conflictingRemoteIdentities;
+
+    final IdentityStorageServiceConfig storageConfig;
```
msteiger: Is this intended to be shared across the package? I'd rather avoid this, if possible.

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceWorker.java
```diff
@@ -0,0 +1,183 @@
...
+ * This class can be in two states, with the default being "logged out".
+ */
+public final class StorageServiceWorker {
+
+    private static final Logger logger = LoggerFactory.getLogger(StorageServiceWorker.class);
+
+    StorageServiceWorkerStatus status = StorageServiceWorkerStatus.LOGGED_OUT;
+    APISession sessionInstance;
+    String loginName;
+    Stack<IdentityBundle> conflictingRemoteIdentities;
```
msteiger: I'd use a Deque<T> implementation. Stack is from the dark ages of Java 1.0

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceWorker.java
```diff
@@ -0,0 +1,183 @@
...
+        console.addMessage(String.format(localizedMessage, args), CoreMessageType.NOTIFICATION);
+    }
+
+    private synchronized void performAction(Action action, StorageServiceWorkerStatus requiredStatus) {
+        if (status != requiredStatus) {
+            throw new RuntimeException("StorageServiceWorker is not in the required status");
+        }
+        status = StorageServiceWorkerStatus.WORKING;
+        logger.info("Performing action {}", action.getClass().getSimpleName());
+        new Thread(() -> {
```
msteiger: Please provide a Runnable to the Thread instead of subclassing it. I hope we have a shared thread pool executor soon, so you can just throw it there.

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceWorkerStatus.java
```diff
@@ -0,0 +1,46 @@
...
+        this.buttonEnabled = buttonEnabled;
+        this.statusMessageId = statusMessageId;
+        this.buttonMessageId = buttonMessageId;
+    }
+
+    public boolean isButtonEnabled() {
+        return buttonEnabled;
+    }
+
+    public String getLocalizedStatusMessage(TranslationSystem translationSystem, String loginName) {
```
msteiger: Those two methods look strange to me. I would move it either to a static helper class or - even better - to the (single?) caller class.

engine/src/main/java/org/terasology/logic/console/ConsoleSystem.java
```diff
@@ -24,22 +25,39 @@
...
 import org.terasology.logic.console.commandSystem.ConsoleCommand;
+import org.terasology.logic.console.ui.MiniChatOverlay;
 import org.terasology.network.ClientComponent;
 import org.terasology.registry.In;
 import org.terasology.rendering.nui.NUIManager;
 
 @RegisterSystem
 public class ConsoleSystem extends BaseComponentSystem {
+
+    private static final ResourceUrn MINICHAT_UI = new ResourceUrn("engine:minichatOverlay");
```
msteiger: Why is that here now?

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
```diff
@@ -82,6 +82,9 @@
     @LayoutConfig
     protected boolean readOnly;
 
+    @LayoutConfig
+    private boolean passwordMode;
```
msteiger: Many UI frameworks have a separate class (derived from the default text control). Have you thought about creating a UIPasswordText instead? It could be too much duplication, though ..

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
```diff
@@ -177,6 +180,15 @@ public UIText(String id) {
         cursorTexture = Assets.getTexture("engine:white").get();
     }
 
+    private String buildPasswordString() {
```
msteiger: You could use this snippet if you like
```        
char[] arr = new char[33];
Arrays.fill(arr, '*');
String str = String.valueOf(arr);
```

engine/src/main/java/org/terasology/identity/ClientIdentity.java
gianluca-nitti: Wanted to XOR the public with the private but accidentally typed private two times. I just found there is an [Objects.hash](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#hash-java.lang.Object...-) for this purpose of building hashcodes for objects composed by multiple fields, would that be OK? Will change again if you prefer the manual implementation.

engine/src/main/java/org/terasology/identity/storageServiceClient/BigIntegerBase64Serializer.java
gianluca-nitti: I know it looks ugly, the problem is that the PostgreSQL base64 encoder used on the server inserts some newlines every a certain numbers of chars. I'll see if I can strip them on the server instead.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
gianluca-nitti: Yes considered that, but avoided for
> too much duplication

exactly this reason. Can move to this approach if you prefer, though.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
gianluca-nitti: I was trying to figure out a better approach than using the StringBuilder but nothing came to my mind. Will now replace with your snippet, thanks.

engine/src/main/java/org/terasology/identity/storageServiceClient/Action.java
gianluca-nitti: Interesting point, making it an interface. In both cases I need to make the method public, however shouldn't be a big deal since both the interface and all the implementor classes are package-private and references to those classes never "leak" outside the package, right?

engine/src/main/java/org/terasology/identity/storageServiceClient/HttpMethod.java
gianluca-nitti: Ok, adding all the methods according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

engine/src/main/java/org/terasology/identity/storageServiceClient/ServiceAPIRequest.java
gianluca-nitti: I don't think so, I searched for all the usages of GsonBuilder in the project and looks like that all the classes which need JSON serialization make their own instance with the appropriate type adapters.

engine/src/main/java/org/terasology/identity/storageServiceClient/ServiceAPIRequest.java
gianluca-nitti: I agree that maybe it doesn't look very readable. However it should be correct since line 47:
- Tries to parse the server response, which, if in correct format, is a JSON object with a single "error" key with the error message as value.
- Throws an exception containing the message obtained from the server error response.

However, if the server returns something which isn't in the correct format or isn't JSON at all (it *shouldn't*, but I think its still better to take this possibility in consideration to avoid to crash the client), GSON.fromJson will throw.
Looks like [it throws `JsonSyntaxException`](https://google.github.io/gson/apidocs/com/google/gson/Gson.html#fromJson-java.lang.String-java.lang.Class-), so I'm replacing RuntimeException with it.

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceWorker.java
gianluca-nitti: Ok, will use an ArrayDeque. Thanks for the input, didn't know that Stack is now discouraged, I used that since conceptually I just need to access the items at one end so it seemed the simplest data structure. Maybe the queue is also a bit better since conflicts can be solved in the order they were discovered (though this probably isn't very important from a user perspective).

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceWorker.java
gianluca-nitti: Yes it's intended to be shared (like the other non-private fields `status`, `sessionInstance`, etc). Basically, all the `Action`s access the `StorageServiceWorker` by accessing and sometimes modifying it's fields; by using the default access modifier, this is only possible from the inside of the package, so it shouldn't be possible to bring the class to an inconsistent state from the outside.
I'm sorry if this doesn't look like a very good practice, but using the `Action`s is the only way I found to avoid making a huge `StorageServiceWorker` class. I didn't use getters and setters since they wouldn't make a big difference (each of these fields would be class-private and have a package-private getter and setter). I'm open to any suggestions to improve this.

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceWorker.java
gianluca-nitti: Please forgive me but I don't understand what you mean here, I'm providing an anonymous implementation of Runnable to the Thread's constructor, not subclassing Thread.

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceWorkerStatus.java
gianluca-nitti: Moving to static helper class since `getLocalizedStatusMessage` is called in two places (`PlayerSettingsScreen` and `MainMenuScreen`)

engine/src/main/java/org/terasology/logic/console/ConsoleSystem.java
gianluca-nitti: I used the MiniChatOverlay to show notifications from the storage service in the menus. If I understand correctly in the menu state there is no `ChatSystem` so I made some modifications to `ConsoleSystem` to show certain notifications in `MiniChatOverlay`. In my opinion it's a good way to notify user when, for example, the synchronization is completed - since it happens asynchronously, I thought that showing message boxes may not be a good idea since they would pop up at random moments possibly annoying the user.
Let me now if something needs to be changed.

engine/src/main/java/org/terasology/identity/ClientIdentity.java
gianluca-nitti: Also changed the hashCode() implementation in `PrivateIdentityCertificate` to use this method. I'm not sure if I should change it in `PlayerPublicCertificate` too since an implementation was already there (I didn't added it in this PR) which simply returns `signature.hashCode()`; combining with the other fields (id, modulus and exponent) will probably reduce the collision rate but could also slow down the computation of the hash, what do you think about it?

gianluca-nitti:
Thank you for the review! Left some comments and made a commit with various fixes.

GooeyHub:

engine/src/main/java/org/terasology/identity/storageServiceClient/StorageServiceWorker.java
msteiger: Ah sorry, I misread.

engine/src/main/java/org/terasology/identity/storageServiceClient/BigIntegerBase64Serializer.java
msteiger: Nevermind

engine/src/main/java/org/terasology/logic/console/ConsoleSystem.java
msteiger: That's an interesting idea - I like it. The word "chat" could be confusing though. Maybe we can find a better name... `NotificationOverlay` maybe? Does not need to happen in this PR though.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
msteiger: Ok, fair enough. Thanks for clarifying

GooeyHub:

engine/src/main/java/org/terasology/logic/console/ConsoleSystem.java
gianluca-nitti: I agree, the name "chat" would not be consistent anymore. I did the refactoring anyway in the latest commit since it was just a 5 minute task.

GooeyHub:

GooeyHub:

gianluca-nitti:
I added one more commit to fix a problem: trying to open the player's settings screen from the in-game pause menu was throwing a NullReferenceException due to the StorageServiceWorker instance (which the player settings screen polls) was being registered in the main menu's context, and thus not accessible from the in-game context.

I apologize if there is something bad in the fact that I added code that nobody asked for after the PR had been approved, possibly making the review/merge process longer, but I think this is still better than letting buggy code to propagate to other branches (develop/master). Unfortunately I didn't noticed this behavior when I tested before opening the PR because I didn't think about opening the settings from the pause menu.

Anyway, now it's fixed: the StorageServiceWorker instance is registered in the rootContext and thus is more decoupled from engine and user interface's states, and the player settings screen can be accessed both from the main menu and from the in-game pause menu.

Cervator:
Tried testing this out, but the server you're hosting seemingly took my registration (again I guess, tried earlier but the site looks much better now) then never sent me a confirmation key. Is it still working? :-)

Also wondering where we should host it. Since our Weblate instance runs out of Docker already maybe we could just add a couple more containers or something? Poking @qwc and @msteiger for ideas

Game itself seemed to run fine and I was able to test it successfully in the past!

gianluca-nitti:
@Cervator : it's OK that it allowed you to "register again" after [the previous time you registered](http://forum.terasology.org/threads/client-identity-cloud-storage-service.1846/#post-15070) because I rebuilt the database about a week ago when I added email verification and reCAPTCHA (the only accounts were mine and your tests, so I guessed it wasn't a big deal; anyway, I have a dump). I had a look at the database now and saw you made an account with email address which is waiting for confirmation.

I tried to register another test account with my email and it worked, I got the activation mail - **in the spam folder, however**. Did you check yours? The reason it goes into spam, I guess, is that it doesn't come from a well-known/trusted mail service, but from a local SMTP server installed on the VPS, so Gmail thinks it's suspicious.
P.S. even if you find the "old" verification email in spam, if you want to retry you probably need to redo the registration (same username and email is fine) because an unverified account expires after 1-2 hours (not sure about the exact value now).

msteiger:
Can we merge this one, @Cervator?

Cervator:
@msteiger certainly - checked it out again and rested, got the confirmation email via junk mail, confirmed, and was able to validate the identity carried over :-) You don't need to wait for me if you have approved and feel you've tested decently, in that case merge ahead and just assign the PR (and any related issues) to the current milestone

Great work @gianluca-nitti !

One small suggestion: Make the UI element a little more consistent between different screens - the little thing in the top right is only visible on the main menu? Maybe it should be there on the settings screen etc too?

We should probably also find some places to update the docs, and a good on-going way to host the service. Will bump the forum thread :-)

Thanks again!

qwc:
@Cervator Of course, spinning up docker containers is the easiest. That's the fancy thing with docker.
But we should think twice regarding resources usage, the weblate droplet may be to small for things like elastic search, etc.

PR #2966, opened 2017-06-01
changed CollideEvent to be consumable

0shine0:
Required for CombatSystem module where this event is consumed if the other entity the weapon collided with was the player who fired it, so that no action is taken on the player.
For eg. arrows that player fire are spawned in the bounding box of the player and shouldn't hurt him.

### DIFF
--- a/engine/src/main/java/org/terasology/physics/events/CollideEvent.java
+++ b/engine/src/main/java/org/terasology/physics/events/CollideEvent.java
@@ -17,13 +17,12 @@
 package org.terasology.physics.events;
 
 import org.terasology.entitySystem.entity.EntityRef;
-import org.terasology.entitySystem.event.Event;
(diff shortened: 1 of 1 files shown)


0shine0:
pinging @Cervator and @portokaliu for review and merge as soon as possible.

0shine0:
linking pr in CombatSystem module where it is required 
https://github.com/Terasology/CombatSystem/pull/11

GooeyHub:

Cervator:
Seems reasonable to me. Pinging a couple others for review just in case - may be wise to try poking them on Slack occasionally too until somebody else can confirm :-)

Cervator:
Tested it out minimally with GooeysQuests, just to make sure a skeleton could still hurt me and vice versa.

PR #2967, opened 2017-06-01
ContextMenu API Annotations

ThompsonTyler:
### Contains

Adds API annotations to the required ContextMenu Classes in order for the context menu of NUI to be used with modules.

### How to test

Create a module that uses the contextMenu from NUI and test it out.
Current version of Scenario module uses it with PR https://github.com/Terasology/Scenario/pull/13


GooeyHub:

PR #2970, opened 2017-06-04
Apply impulse, movement mode change and teleport only to alive charac…

flo:
### Contains

Just the addition of a few checks if a character is alive in locations where they were missing.

### How to test

* Try if impulse events work e.g. via the command `pushCharacter 0 20 0`
* Try if teleport works. e.g. via the command`teleport 0 50 0`
* Try if diving works `ghost` command works.


GooeyHub:

Cervator:
Checked it out real quick, thanks for the testing instructions :-)

I tried it on a headless server where I couldn't easily get debug permission since that only happens if you put the server in debug mode - which is harder when the server is headless. Took the opportunity to tweak at `usePermissionKey` to where it *will* give you debug permissions - for now. Left the old variant commented out so we can go back to it later like at Beta.

Also improved an unrelated exception message

PR #2973, opened 2017-06-07
Fixes `dumpEntities` console command

nihal111:
Added serializeSpecialFloatingPointValues to the Gson builder.
This solves the error which appears on using `dumpEntities` command in console:
```
An error occurred while executing command 'dumpEntities': NaN is not a valid double value as per JSON specification. To override this behavior, use GsonBuilder.serializeSpecialFloatingPointValues() method.
```


GooeyHub:

Cervator:
Tried it out, got a valid dump file. Although I'm not sure if you needed any special entities to trigger the error. In any case tests out so merging, thanks :-)

nihal111:
An entity with floating point arguments would trigger the error.
Something like this:
```
  "BoxShape" : {
    "extents" : [0.5, 2, 5]
  },
```

PR #2974, opened 2017-06-14
Small refactor in AbstractClient

gianluca-nitti:
### Contains

In the `AbstractClient` class, the private method `findClientEntityRef` is only called from `createEntity`, where we already have a handy reference to the `EntityManager`, so in my understanding there is no need to take it from the CoreRegistry.
Passing the reference as an argument also makes testing of the class and it's subclasses easier and cleaner (I needed to look at the `AbstractClient` class while writing tests for the `HeadlessClient` class I added in MovingBlocks/FacadeServer#2) since you don't have to set up a mocked context, add the entityManager to it, setting the CoreRegistry's context, etc.

### How to test

Not much to test, all the automated unit tests pass and if you can try to use the game normally (e.g. join a local world or multiplayer server) and you can see that everything is as before.

Nothing I can think of.


GooeyHub:

PR #2975, opened 2017-06-14
[WIP] Sectors

Vizaxo:
### Sectors WIP PR

The first stage of the implementation of sectors as part of my GSOC project, for mentor review.

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.entitySystem.entity;
+
+/**
+ */
+public interface EntityCache {
+
+    /**
+     * Create a new entity in this EntityCache
+     *
+     * @return the new entity
+     */
+    EntityRef create();
+
+}
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
@@ -162,4 +162,13 @@ public interface EntityCache {
      * @return the {@link EntityRef}, if it exists; {@link EntityRef#NULL} otherwise
      */
     EntityRef getExistingEntity(long id);
+
+    /**
+     * Fund out if a particular entity has a component of the given class.
+     *
+     * @param entityId the entity to check
+     * @param componentClass the class to check for
+     * @return whether the entity has the component
+     */
+    boolean hasComponent(long entityId, Class<? extends Component> componentClass);
 }
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
@@ -151,10 +151,6 @@ public interface EntityCache {
      */
     EntityRef createEntityRefWithId(long id);
 
-    void destroy(long entityId);
-
-    void destroyEntityWithoutEvents(EntityRef entity);
-
     Iterable<EntityRef> getAllEntities();
 
     /**
@@ -163,6 +159,15 @@ public interface EntityCache {
      */
     Iterable<EntityRef> getEntitiesWith(Class<? extends Component>... componentClasses);
 
+    /**
+     * @param componentClasses
+     * @return A count of entities with the provided component types
+     */
+    int getCountOfEntitiesWith(Class<? extends Component>... componentClasses);
+
+    /**
+     * @return A count of currently active entities
+     */
     int getActiveEntityCount();
 
     /**
@@ -173,12 +178,4 @@ public interface EntityCache {
      */
     EntityRef getExistingEntity(long id);
 
-    /**
-     * Fund out if a particular entity has a component of the given class.
-     *
-     * @param entityId the entity to check
-     * @param componentClass the class to check for
-     * @return whether the entity has the component
-     */
-    boolean hasComponent(long entityId, Class<? extends Component> componentClass);
 }
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
@@ -65,6 +65,16 @@ public interface EntityCache {
      */
     EntityRef create(Iterable<Component> components);
 
+
+    /**
+     * Creates a new entity from the given components.
+     *
+     * @param components the components to create this entity from
+     * @param sendLifecycleEvents will only send lifecycle events if this is true
+     * @return
+     */
+    EntityRef create(Iterable<Component> components, boolean sendLifecycleEvents);
+
     /**
      * @param prefabName The name of the prefab to create.
      * @return A new entity, based on the the prefab of the given name. If the prefab doesn't exist, just a new entity.
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
@@ -25,6 +25,11 @@ import org.terasology.math.geom.Vector3f;
 public interface EntityCache {
 
     /**
+     * Removes all entities from the cache.
+     */
+    void clear();
+
+    /**
      * Creates an EntityBuilder.
      *
      * @return A new entity builder
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
@@ -146,4 +146,12 @@ public interface EntityCache {
     void destroyEntityWithoutEvents(EntityRef entity);
 
     Iterable<EntityRef> getAllEntities();
+
+    /**
+     * @param componentClasses
+     * @return An iterable over all entities with the provided component types.
+     */
+    Iterable<EntityRef> getEntitiesWith(Class<? extends Component>... componentClasses);
+
+    int getActiveEntityCount();
 }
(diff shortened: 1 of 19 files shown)


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityManager.java
```diff
@@ -250,31 +213,17 @@ public EntityRef getEntity(long id) {
...
-        for (Component component : prefab.iterateComponents()) {
-            components.add(componentLibrary.copy(component));
-        }
-        components.add(new EntityInfoComponent(prefab, prefab.isPersisted(), prefab.isAlwaysRelevant()));
-        return create(components);
+        return globalCache.create(prefab);
     }
 
     @Override
+    //Todo: Depreciated, maybe remove? Not many uses
```
skaldarnar: What should be used instead? This just creates a _new_ EnityRef for an entity with the same set of components. Can we create a new builder from a give entity instead?

GooeyHub:

engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityManager.java
Vizaxo: `EntityManager` says that it's depreciated in favour of `EntityRef.copy()`. Creating a builder from an existing entity is an interesting idea, especially if entities need to be duplicated multiple times (e.g. maybe for duplicating inventory items).

engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityCache.java
```diff
@@ -0,0 +1,350 @@
...
+        }
+        return entity;
+    }
+
+    public EntityBuilder newBuilder() {
+        return new EntityBuilder(this);
+    }
+
+    @Override
+    public EntityBuilder newBuilder(String prefabName) {
```
Vizaxo: The builder class says that the whole entity can be setup before creation without any events being sent in-between. It also seems like a good way of creating lots of the same entity, so might be useful for spawning lots of cities, for example.

Implementing some of the create() methods in terms of the builder should reduce code duplication, especially the code to extract components from a prefab.

engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityCache.java
```diff
@@ -0,0 +1,350 @@
...
+            }
+            return create(prefab);
+        }
+        return create();
+    }
+
+    @Override
+    public EntityRef create(String prefabName, Vector3f position) {
+        if (prefabName != null && !prefabName.isEmpty()) {
+            Prefab prefab = entityManager.getPrefabManager().getPrefab(prefabName);
```
Vizaxo: Ah, I'll add the checks where they're needed.

Some of them are mostly safe because the associated create(String prefabName) does the null check, but I'll add the checks there as well, in case it is called with a null prefab from somewhere else (currently nowhere within the engine calls the create() methods with a Prefab directly, but it doesn't hurt to be safe).

engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityCache.java
```diff
@@ -0,0 +1,350 @@
...
+
+/**
+ */
+public class PojoEntityCache implements EntityCache {
+
+    private PojoEntityManager entityManager;
+
+    private static final Logger logger = LoggerFactory.getLogger(PojoEntityCache.class);
+
+    private Map<Long, BaseEntityRef> entityStore = new MapMaker().weakValues().concurrencyLevel(4).initialCapacity(1000).makeMap();
```
Vizaxo: I hadn't considered entities being in multiple caches. It might be useful (e.g. for entities near a sector boundary), but it could cause lots of problems with systems trying to update the same entity at the same time in two different caches (especially if we're thinking about using multi-threading to handle sectors in the future). So I think it's best to have each entity in a single cache.

I think that'll definitely be necessary when we have multiple sectors, so sectors can be moved or split (and entities can change sector if they move around, or find themselves interacting with lots of things outside of their sector).

engine/src/main/java/org/terasology/entitySystem/entity/EntityCache.java
```diff
@@ -0,0 +1,147 @@
...
+ * limitations under the License.
+ */
+package org.terasology.entitySystem.entity;
+
+import org.terasology.entitySystem.Component;
+import org.terasology.entitySystem.prefab.Prefab;
+import org.terasology.math.geom.Quat4f;
+import org.terasology.math.geom.Vector3f;
+
+/**
```
Vizaxo: Yep, I will. I didn't want to go documenting lots of stuff now when it's still rapidly changing.

engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityCache.java
```diff
@@ -0,0 +1,350 @@
...
+        if (prefab != null) {
+            for (Component component : prefab.iterateComponents()) {
+                builder.addComponent(entityManager.getComponentLibrary().copy(component));
+            }
+            builder.addComponent(new EntityInfoComponent(prefab, prefab.isPersisted(), prefab.isAlwaysRelevant()));
+        }
+        return builder;
+    }
+
+    public Map<Long, BaseEntityRef> getEntityStore() {
```
Vizaxo: I hadn't considered that; I think an unmodifiable map would be more sensible, and then have a `putEntity` method so that the EntityManager can put entities into the map (from `createEntityRef`). But perhaps that should be changed too: maybe the EntityManager should just call the cache's `createEntityRef`? I think restricting external changes to the internal map would be an improvement; I'll look into the best way of doing it.

GooeyHub:

engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityCache.java
```diff
@@ -0,0 +1,350 @@
...
+        entityManager.notifyComponentRemovalAndEntityDestruction(entityId, ref);
+        destroy(ref);
+    }
+
+    private void destroy(EntityRef ref) {
+        // Don't allow the destruction of unloaded entities.
+        long entityId = ref.getId();
+        entityStore.remove(entityId);
+        entityManager.remove(entityId);
+        if (ref instanceof PojoEntityRef) {
```
Vizaxo: I don't see why you shouldn't be able to invalidate any EntityRef, even if that doesn't do anything for some implementations of it. I've added that in 15ca85c; I think that getting rid of the `instanceof` makes it a lot cleaner.

engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityCache.java
```diff
@@ -0,0 +1,350 @@
...
+        if (entity.isActive()) {
+            entityManager.notifyComponentRemovalAndEntityDestruction(entity.getId(), entity);
+            destroy(entity);
+        }
+    }
+
+    @Override
+    public EntityRef createEntityWithId(long id, Iterable<Component> components) {
+        if (!entityManager.registerId(id)) {
+            return EntityRef.NULL;
```
Vizaxo: An error gets logged by `registerId`. I'm not sure if it happens often or if it's a problem; this is only currently called for storage/serialization tasks, which already know the ID they're trying to create for from the entity they already had (so presumably they know that it's not invalid).

Most uses would only use the normal create methods, which will guarantee a valid ID. Perhaps a note should be added about this to the documentation?

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/entitySystem/entity/EntityManager.java
```diff
@@ -18,88 +18,22 @@
...
-    EntityRef create(Prefab prefab, Vector3f position);
-
-    /**
-     * @param prefab
-     * @param position
-     * @param rotation
-     * @return
-     */
-    EntityRef create(Prefab prefab, Vector3f position, Quat4f rotation);
+    default EntityRef createSectorEntity() {
```
msteiger: I think that `null` should not be returned, if possible. What about an empty sector entity?

Vizaxo:
@msteiger Thanks for the feedback.

> Does the actual implementation of the entity manager change?

Vizaxo:
@msteiger Thanks for the feedback.

> Does the actual implementation of the entity manager change?

The implementation of most of the methods is moved into the EntityCache, and I have been refactoring those methods to reduce duplicate code, but their functionality remains the same. The EntityManager now calls the methods on the global cache for lots of its methods by default, so the interface remains the same if you are just dealing with entities in the global scope.

> Sorry, I'm not very familiar with the GSOC proposal. What exactly are you aiming for in this PR?

[This issue](https://github.com/MovingBlocks/Terasology/issues/2976) might help explain it, and I've just posted a [blog post](https://vizaxo.github.io/2017/06/28/google-summer-of-code-introduction.html) (skip to "What is my project?" section) which should outline what I'm aiming for. This first stage will just be to get a basic sector/cache system working; doing things like multi-threading and multi-world will have to come later (after things like re-integrating Gestalt-EntitySystem have been worked out (as mentioned in that issue thread)).

If that's still unclear, please let me know.

> I would try to use other components as fields rather than creating an inheritance tree. Also see https://en.wikipedia.org/wiki/Composition_over_inheritance for details.

The caches are implemented as fields; for example, the PojoEntityManager has a global cache, and a PojoSectorManager. The PojoSectorManager in turn has caches for each of its sectors (just one, at the moment).

But I did create an inheritance tree by making the EntityManager and SectorManager implement EntityCache. This is because the managers are caches in their own right, and can be used in place of individual caches when it's appropriate (so the SectorManager is the cache for all sector entities, and only the SectorManager needs to know about how each cache is represented internally).

I initially represented this with pure composition, but the interfaces had to be duplicated because many of the same operations have to be performed on the managers and caches (e.g. creating an entity in that cache).

I think this way of working works in this scenario, because the caches are separate fields, but the inheritance means the interfaces don't have to be duplicated. There might be a possibility for turning the inheritance of EntityManager -> LowLevelEntityManager -> EngineEntityManager (and the equivalent for caches) into a composition relationship, but I can't immediately see much advantage to that, as inheritance seems to fit their relationship well already.

Saying that, i tend to think about this kind of stuff in terms of inheritance, but can certainly see the advantage of favouring composition instead. I might be missing the point here or thinking about it in the wrong way, so I'd like to keep considering alternate ways of doing things if they lead to better code, even if it's not how I naturally think about these kind of problems. If you think this can be better represented without the inheritance in the interfaces, I'm all ears.

engine/src/main/java/org/terasology/entitySystem/entity/EntityManager.java
Vizaxo: Yep, you're right. `EntityRef.NULL` is the preferred way of returning a null entity, as it behaves sensibly by avoiding NullPointerExceptions.

I'm actually not sure why I implemented this as a default method; the only EntityManager in the engine is the Pojo one, for which I've implemented the method anyway (I expect my thinking was that I didn't want to have to go and implement it in multiple places while the interface is still changing). It's probably best to just remove the default implementation and leave it abstract.

GooeyHub:

GooeyHub:

Cervator:
Closing as superseded by #3009 which has been merged into the v2 branch :-)

PR #2916, opened 2017-04-26
[WIP] ShaderParameters Removal

vampcat:
Moves the functionality of ShaderParameters to the Nodes, making the system more transparent.

This is still a WIP, and this PR has been made to get feedback.

@emanuele3d What do you think? Should I proceed forward like this, or should I make some changes? Easier if you review one or two ShaderParameters, so that you don't have to type the same comments 20 times :D

PR #2981, opened 2017-06-17
Shader Parameter Removal (New)

vampcat:
Opened in lieu of #2916, to have a clean PR without all the messy rebasing history.


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
```diff
@@ -190,4 +206,35 @@ public void process() {
...
+            } else {
+                isParallaxMapping = renderingConfig.isParallaxMapping();
+                if (isNormalMapping) {
+                    if (isParallaxMapping) {
+                        addDesiredStateChange(setHeightTerrain);
+                    } else {
+                        removeDesiredStateChange(setHeightTerrain);
+                    }
+                }
+            }
```
emanuele3d: And here I'd just write:
`} // else: no other cases are possible - see subscribe operations in initialize().`

engine/src/main/java/org/terasology/rendering/dag/nodes/FinalPostProcessingNode.java
```diff
@@ -95,15 +95,17 @@ public void process() {
...
-    public void propertyChange(PropertyChangeEvent evt) {
-        // we assume here that this property change event is fired only if there has been a change
-        if (!renderingDebugConfig.isEnabled()) {
-            removeDesiredStateChange(enableDebugMaterial);
-            addDesiredStateChange(enablePostMaterial);
-        } else {
-            removeDesiredStateChange(enablePostMaterial);
-            addDesiredStateChange(enableDebugMaterial);
+    public void propertyChange(PropertyChangeEvent event) {
+        if (event.getOldValue() != event.getNewValue()) {
```
emanuele3d: See comment above regarding event being fired only if the value is indeed changed.

engine/src/main/java/org/terasology/rendering/dag/nodes/InitialPostProcessingNode.java
```diff
@@ -83,17 +95,24 @@ public InitialPostProcessingNode(Context context) {
 
         initialPostMaterial = getMaterial(INITIAL_POST_MATERIAL);
 
+        renderingConfig = context.get(Config.class).getRendering();
+        isBloom = renderingConfig.isBloom();
+        renderingConfig.subscribe(RenderingConfig.BLOOM, this);
+        isLightShafts = renderingConfig.isLightShafts();
```
emanuele3d: Let's rename these variables to:
- bloomIsEnabled
- lightShaftsAreEnabled

please.

engine/src/main/java/org/terasology/rendering/dag/nodes/InitialPostProcessingNode.java
```diff
@@ -120,4 +138,27 @@ public void process() {
 
         PerformanceMonitor.endActivity();
     }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent event) {
+        if (event.getOldValue() != event.getNewValue()) {
```
emanuele3d: See earlier comment.

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
```diff
@@ -66,6 +69,12 @@
 
     private SubmersibleCamera activeCamera;
 
+    private boolean isNormalMapping;
+    private boolean isParallaxMapping;
+
+    private StateChange setNormalTerrain;
+    private StateChange setHeightTerrain;
+
```
emanuele3d: Let's rename the four variables above in line with my previous comments please.

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
```diff
@@ -199,4 +215,35 @@ private void renderChunkBoundingBox(RenderableChunk chunk, Vector3f chunkPositio
 
         GL11.glPopMatrix(); // Resets the matrix stack after the rendering of a chunk.
     }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent event) {
```
emanuele3d: This block of code seems to be absolutely identical to one in a previous class in this commit. Can they inherit it from a common ancestor?

engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
```diff
@@ -56,17 +59,26 @@
...
+    private boolean isLocalReflections;
+    private boolean isSsao;
+    private boolean isOutline;
+    private boolean isInscattering;
+    private boolean isVolumetricFog;
+
+    private StateChange setTexSceneReflectiveRefractiveNormals;
+    private StateChange setTexSsao;
+    private StateChange setTexEdges;
+    private StateChange setTexSceneSkyBand;
```
emanuele3d: Please rename the 10 variables above in line with:
- localReflectionsAreEnabled
- setReflectiveRefractiveNormalsInputTexture

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
```diff
@@ -83,6 +87,13 @@
 
     private SubmersibleCamera activeCamera;
 
+    private boolean isNormalMapping;
+    private boolean isParallaxMapping;
+    private boolean isAnimateWater;
+
+    private StateChange setNormalTerrain;
+    private StateChange setHeightTerrain;
```
emanuele3d: Let's rename these 5 variables please.

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
```diff
@@ -267,4 +286,37 @@ public void update() {
 
         readOnlyGBufferFbo.attachDepthBufferTo(refractiveReflectiveFbo);
     }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent event) {
```
emanuele3d: This method is an _almost_ exact duplicate of other propertyChange methods. Perhaps, rather than inheriting the method from an ancestor class (and here overriding it because of the animated water aspect) like I suggest in a previous comment, we could have a couple of more specialized methods in an ancestor class, so that you could write:

```java
if (event.getPropertyName().equals(RenderingConfig.NORMAL_MAPPING)) {
    handleNormalMappingSettingChange();
} else if (event.getPropertyName().equals(RenderingConfig.PARALLAX_MAPPING)) {
    handleParallaxMappingSettingChange();
} else if (event.getPropertyName().equals(RenderingConfig.ANIMATE_WATER)) {
   animateWaterIsEnabled = renderingConfig.isAnimateWater();
} // else: no other case is possible. See subscribe operations in initialise().
```

engine/src/main/java/org/terasology/rendering/dag/nodes/WorldReflectionNode.java
```diff
@@ -199,4 +215,37 @@ public void process() {
 
         PerformanceMonitor.endActivity();
     }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent event) {
```
emanuele3d: This method seem to be an exact duplicate of the same method in another two classes: to be obtained via inheritance instead?

engine/src/main/resources/assets/shaders/lightGeometryPass_frag.glsl
```diff
@@ -20,7 +20,9 @@ uniform vec3 lightViewPos;
 
 uniform sampler2D texSceneOpaqueDepth;
 uniform sampler2D texSceneOpaqueNormals;
+#if defined (FEATURE_LIGHT_DIRECTIONAL)
```
vampcat: No, not *necessary*.

That being said, digging through the entire shaders to find out why a particular varaible was being reported as unused was a chore. Since I wanted to make the code clearer for everyone (and myself for later on when I dived into it), I went ahead with it.

engine/src/main/java/org/terasology/rendering/dag/nodes/InitialPostProcessingNode.java
vampcat: I re-used the variable names from the renderingConfig, as you can see. Would you like me to just rename them inside Nodes, or inside renderingConfig too?

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
```diff
@@ -151,9 +167,9 @@ public void process() {
         chunkMaterial.setInt("textureAtlas", 0, true);
         chunkMaterial.setInt("textureEffects", 1, true);
         chunkMaterial.setInt("textureLava", 2, true);
-        if (renderingConfig.isNormalMapping()) {
+        if (isNormalMapping) {
             chunkMaterial.setInt("textureAtlasNormal", 3, true);
-            if (renderingConfig.isParallaxMapping()) {
+            if (isParallaxMapping) {
```
vampcat: Yeah, I'm aware that Parallax Mapping doesn't work, and also of some of the glitches associated with it. I hope to dive into it as soon as I get past the preliminary work :D
( Hopefully I won't have to wait for ~4 years :P )

engine/src/main/java/org/terasology/rendering/dag/nodes/AlphaRejectBlocksNode.java
```diff
@@ -190,4 +206,35 @@ public void process() {
 
         PerformanceMonitor.endActivity();
     }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent event) {
+        if (event.getOldValue() != event.getNewValue()) {
```
vampcat: Yes, you are absolutely correct.

I've added a comment to all these functions implying that it only gets called when the property is changed, since we want to make things as explicit as possible. I hope that's not too redundant for your tastes?

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
vampcat: I'm holding off on this change until further discussion on Slack.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
```diff
@@ -69,16 +69,16 @@
...
 
-    private boolean isLocalReflections;
-    private boolean isSsao;
-    private boolean isOutline;
-    private boolean isInscattering;
-    private boolean isVolumetricFog;
+    private boolean localReflectionsAreEnabled;
+    private boolean ssaoIsEnabled;
+    private boolean outlineIsEnabled;
+    private boolean inscatteringIsEnabled;
```
emanuele3d: My bad: I just realized the following two variables should be renamed to be easier to understand and consistent with the related nodes:
- inscatteringIsEnabled   -->  hazeIsEnabled
- setSkyBandInputTexture   -->  setHazeInputTexture

I should have thought about it earlier.

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
emanuele3d: I was thinking about this option. Indeed if the method have no arguments the ancestor class would need all sorts of additional local variables. But what if the two handle*() methods above obtained everything they need as input arguments? Feasible?

emanuele3d:
Ta-dah! PR approved! @Cervator : from my perspective this PR is good to go. Please give vampcat another 24 hours or so to do some stress testing on changing rendering settings.

And @vampcat do not fix the camera projection matrix issue in this PR unless the problem was introduced by this PR please.

emanuele3d:
@Cervator: while testing we found that this PR breaks outline and haze (fog). We'll try and get them fixed before flagging this PR as good to go again.

GooeyHub:

emanuele3d:
I have reviewed the latest changes and I confirm we are good to go from my perspective.

@Cervator? It's GO for merge at your convenience.

emanuele3d:
Wow. Just 14 days and 33 conversations to get this thing merged! Congrats @vampcat! =)

PR #2958, opened 2017-05-23
Do not destroy player entity completely on death

nihal111:
### Contains

An approach to not destroy the player entity completely upon death. This is done so as to allow for components attached to the player entity to persist after respawn.
Added death handling for character entities, that takes care of not destroying the player entity by consuming the BeforeDestroyEvent. The CharacterSystem captures the BeforeDestroy event and sends out a DeathEvent for all character entities regardless. There is just an additional check to consume the BeforeDestroy event (so as to save the DoDestroy from happening) in the case that the character entity is a Player (i.e. has a PlayerCharacterComponent). In that case, it also sends out a PlayerDeathEvent.

This adds the **OnPlayerRespawnedEvent, PlayerDeathEvent, AliveCharacterComponent, PlayerCharacterComponent**.

### Outstanding before merging

- [x] Respawn logic.
- [x] Fix things that break.
- [ ] Follow up: Add AliveCharacterComponent to all the character entities out there.

### Details

A character lives. It has the AliveComponent.
It dies.
BeforeDestroyEvent receives all entities having AliveComponent
+ If the entity *also has a PlayerCharacterComponent*, the event is consumed to prevent destruction. PlayerDeathEvent sent along with DeathEvent
+ Else just a DeathEvent sent
DeathScreen launches on DeathEvent and sends the respawnRequestEvent
respawnRequestEvent is received by the PlayerSystem
PlayerSystem then respawns player, sets the location for clientEntity to spawn location. Sends onPlayerRespawnedEvent

Components that need to be removed on death are removed by receiving the PlayerDeathEvent.
Components that need to be reset are done by receiving the onPlayerRespawnedEvent,

PR #2982, opened 2017-06-18
Prevents player entity from being destroyed after death onDamage

nihal111:
Fixes the bug of a player entity being destroyed after death.

https://github.com/MovingBlocks/Terasology/pull/2958/ made the change to preserve the player entity even after death. 

Basically, the BeforeDestroyEvent for a player entity is consumed if it has the AliveCharacterComponent, the AliveCharacterComponent is stripped off and the DeathScreen is created ([here](https://github.com/MovingBlocks/Terasology/blob/c18b201a4d5ad678e6f1346b5e8727cc3367e7db/engine/src/main/java/org/terasology/logic/characters/CharacterSystem.java#L89-L89)). 
Once the player loses its AliveCharacterComponent, it is susceptible to destruction should it get damaged again (since the HealthAuthoritySystem would trigger a DestroyEvent again).

Since most damage inflictors like attacking skeletons were made to attack only players having AliveCharacterComponent and committing suicide by teleporting to a great height won't work once you're already dead, this went unnoticed.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
@@ -29,6 +29,7 @@ import org.terasology.entitySystem.systems.UpdateSubscriberSystem;
 import org.terasology.logic.characters.AliveCharacterComponent;
 import org.terasology.logic.characters.CharacterComponent;
 import org.terasology.logic.characters.CharacterTeleportEvent;
+import org.terasology.logic.health.BeforeDestroyEvent;
 import org.terasology.logic.location.Location;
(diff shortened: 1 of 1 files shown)


nihal111:
In addition, https://github.com/Terasology/DamagingBlocks/pull/5 was an attempt to make the DamageSystem for lava blocks similar to how the attacking system for skeleton works, i.e. do not damage dead characters.
It doesn't hurt having it in place, but can now be reverted since this PR would take care of it.

GooeyHub:

Cervator:
Tests out fine. Thanks! :-)

PR #2983, opened 2017-06-22
Separated spawnLocation calculation from respawnPlayer method

nihal111:
This PR is linked to https://github.com/Terasology/AdventureAssets/pull/11
Separates the existing RespawnRequestEvent handler into two handlers with different priorities. The RespawnRequestEvent is sent to the client entity once the "Respawn" button is clicked on the Death Screen.

Basically, this is what happens now:

- The spawn location for a player is calculated using the worldGenerator information at the earliest step (`CRITICAL` priority).
- Any other system which wants to change the respawn location can then intercept the RespawnRequestEvent and make change to the entity's (client's) location component. (priority between `TRIVIAL` and `CRITICAL`, both exclusive)
- Finally, the RespawnRequestEvent is received by the PlayerSystem again. This time the respawnPlayer method is called and the player respawns. (`TRIVIAL` priority).

Also, someone left the tap on in the HealthAuthoritySystem. Removed the heavy logging.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
@@ -21,6 +21,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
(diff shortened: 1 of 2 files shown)


GooeyHub:

SkySom:
It works wonderfully, thankfully I didn't find myself under the angel of nightmares at night. :D

PR #2985, opened 2017-06-22
changes in physics to enable trigger collisions with the world and to make triggers have their own collision groups

0shine0:
### Contains

changes in physics to enable trigger collisions with the world and to make triggers have their own collision groups. It also makes slight changes in PhysicsLiquidWrapper and PhysicsWorldWrapper so that the latest TeraBullet module can recompile with the engine.

Will need to merge https://github.com/MovingBlocks/TeraBullet/pull/2 before merging this.

NOTE: After merging this pr you will need to recompile the engine module with latest TeraBullet module.

### DIFF
--- a/engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
+++ b/engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
@@ -55,6 +55,8 @@ import org.terasology.world.WorldProvider;
 import org.terasology.world.block.Block;
 import org.terasology.world.block.BlockComponent;
 
+import com.google.common.collect.Lists;
+
 import java.util.Iterator;
 import java.util.List;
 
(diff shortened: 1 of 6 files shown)


GooeyHub:

0shine0:
Pinging @Cervator for reveiwing the pr, to merge it before coming saturday for proper testing of CombatSystem module.

engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
```diff
@@ -242,13 +244,55 @@ public void update(float delta) {
 
         for (CollisionPair pair : collisionPairs) {
             if (pair.b.exists()) {
-                pair.a.send(new CollideEvent(pair.b,pair.pointA,pair.pointB,pair.distance,pair.normal));
+                short bCollisionGroup = getCollisionGroupFlag(pair.b);
+                short aCollidesWith = getCollidesWithGroupFlag(pair.a);
+                if((bCollisionGroup & aCollidesWith) != 0 || 
+                        (pair.b.hasComponent(BlockComponent.class) && !pair.a.hasComponent(BlockComponent.class))){
```
portokaliu: This check would mean that no two items colliding would ever mention the collision took place. Is that intended?

0shine0: No, i check for blocks here in addition to the entities/items (in line 249). I just added this here for blocks because i don't know if they have collision filter or not and this bypasses filters for blocks and sends CollideEvent only to the other entity that is colliding with block and no event is sent to block entities.

portokaliu:
@0shine0 Could you maybe re-do the pr? Not exactly sure what's wrong with the checks

0shine0:
@portokaliu The checks have failed because this repo references to an old version of teraBullet. The methods that i defined in the PhysicsLiquidWrapper and the PhysicsWorldWrapper class are not introduced in this old version. This repo needs to include the latest teraBullet version before merging the pr and for the checks to work.

Cervator:
Yeah I'm working with the merge locally right now and figured that was it :-)

Cervator:
Released TeraBullet v1.3.0, bumped the engine's dependency to use it, tested locally with the CombatSystem and Rails (and lots of railgun), and merged! :-)

One quirk: Dropped physics objects now seem to keep going waaaay longer, as if they're not being drained by friction fast enough. Very bouncy and slippery, will keep sliding across the ground slowing down so incrementally that they go poof from "timing out" before they actually stop moving entirely.

Also added CombatSystem + AdventureAssets to the module lineup!

PR #2991, opened 2017-06-25
Reworking FBOs

vampcat:
- Rework FBOs regeneration: Instead of deleting and creating a new FBO whenever FBOs have to be regenerated (upon resizing), we now only regenerate the attachments of that FBO. Due to this, the fboId is now retained and hence most Nodes and StateChanges now don't need to subscribe to FBOManagers (since no update call is ever issued).
- Rework fbo swapping: Instead of maintaining two pre-defined read/write buffers, we now maintain two generic buffers and a logical distinction between the read/write buffers. Due to this, on every invocation of swap, instead of literally swapping two buffers (and therefore causing an update call to all subscribers), we alternate the buffer that we consider read (or write) from now on.
- A few minor changes, which have been explained in their own individual comments.

There should be no noticeable visual changes as a side-effect of this PR.


vampcat:
Tagging @emanuele3d for feedback.

GooeyHub:

engine/src/main/java/org/terasology/rendering/opengl/FBO.java
```diff
@@ -369,6 +370,78 @@ public static FBO create(ResourceUrn urn, Dimensions dimensions, Type type,
...
+            GL20.glDrawBuffers(GL11.GL_NONE);
+        } else {
+            GL20.glDrawBuffers(bufferIds);
+        }
+
+        verifyCompleteness(fboConfig.getName(), type, fbo);
+        glBindFramebufferEXT(GL_FRAMEBUFFER_EXT, 0);
+
+        /*
+        FBO fbo2 = create(fboConfig);
```
vampcat: Please ignore these comments for now 😓

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/AmbientOcclusionNode.java
```diff
@@ -60,7 +60,6 @@
     private float outputFboWidth;
     private float outputFboHeight;
 
-    @SuppressWarnings("FieldCanBeLocal")
```
vampcat: This warning longer applies.

engine/src/main/java/org/terasology/rendering/dag/nodes/ApplyDeferredLightingNode.java
```diff
@@ -77,8 +77,6 @@ public void process() {
 
         // Actual Node Processing
 
-        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT); // TODO: this is necessary - but why? Verify in the shader.
```
vampcat: This was only required because the shader wasn't being cleared even once, and that lead to problems with depth buffer. Moved this task to a more appropriate location (in WorldRendererImpl, in it's own BufferClearingNode)

engine/src/main/java/org/terasology/rendering/dag/nodes/BlurredAmbientOcclusionNode.java
```diff
@@ -60,7 +60,6 @@
     private float outputFboWidth;
     private float outputFboHeight;
 
-    @SuppressWarnings("FieldCanBeLocal")
```
vampcat: No longer applies.

engine/src/main/java/org/terasology/rendering/opengl/OpenGLUtils.java
```diff
@@ -18,17 +18,26 @@
...
+import static org.lwjgl.opengl.GL11.glEndList;
+import static org.lwjgl.opengl.GL11.glGenLists;
+import static org.lwjgl.opengl.GL11.glLoadIdentity;
+import static org.lwjgl.opengl.GL11.glMatrixMode;
+import static org.lwjgl.opengl.GL11.glNewList;
 import static org.lwjgl.opengl.GL11.glPopMatrix;
+import static org.lwjgl.opengl.GL11.glPushMatrix;
+import static org.lwjgl.opengl.GL11.glTexCoord2d;
+import static org.lwjgl.opengl.GL11.glVertex3i;
+import static org.lwjgl.opengl.GL11.glViewport;
```
vampcat: All these new imports are due to removal of old * import.

engine/src/main/java/org/terasology/rendering/opengl/OpenGLUtils.java
```diff
@@ -61,66 +70,6 @@ public static void applyBillboardOrientation() {
     }
 
     /**
-     * Sets the viewport of the currently bound FBO to the dimensions of the FBO
```
vampcat: The following methods have been converted to StateChanges and are unused here, so removing to eliminate confusion and code redundancy.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -242,7 +242,7 @@ private void initRenderGraph() {
         // (i.e. water) reflection generation
         FBOConfig reflectedBufferConfig = new FBOConfig(BackdropReflectionNode.REFLECTED_FBO, HALF_SCALE, FBO.Type.DEFAULT).useDepthBuffer();
         BufferClearingNode reflectedBufferClearingNode = new BufferClearingNode(reflectedBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
-        renderGraph.addNode(reflectedBufferClearingNode, "reflectedBufferClearingNode"); // TODO: verify this is necessary
```
vampcat: Yes, it is necessary.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -255,17 +255,21 @@ private void initRenderGraph() {
         BufferClearingNode reflectedRefractedClearingNode = new BufferClearingNode(reflectedRefractedBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
         renderGraph.addNode(reflectedRefractedClearingNode, "reflectedRefractedClearingNode");
 
-        FBOConfig sceneOpaqueFboConfig = displayResolutionDependentFBOs.getFboConfig(BUFFER1);
+        FBOConfig scenePrimaryBufferConfig = displayResolutionDependentFBOs.getFboConfig(BUFFER1);
```
vampcat: Done to make it match the functions getPrimaryBuffer and getSecondaryBuffer. I expect that (and subsequently this) to change after @emanuele3d's comments.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -255,17 +255,21 @@ private void initRenderGraph() {
...
+        FBOConfig sceneSecondaryBufferConfig = displayResolutionDependentFBOs.getFboConfig(BUFFER1);
 
-        BufferClearingNode readBufferClearingNode = new BufferClearingNode(sceneOpaqueFboConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+        BufferClearingNode readBufferClearingNode = new BufferClearingNode(scenePrimaryBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
         renderGraph.addNode(readBufferClearingNode, "readBufferClearingNode");
 
+        BufferClearingNode writeBufferClearingNode = new BufferClearingNode(sceneSecondaryBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
```
vampcat: Clearing both BUFFER1 and BUFFER2 is essential (probably depth or stencil issues) as mentioned earlier, so shifted from a random line in a random Node to a proper Node.

GooeyHub:

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -111,11 +123,19 @@ private void createFBOs() {
...
+
+    public FBO getPrimaryBuffer() {
+        return get(buffersSwapped ? BUFFER2 : BUFFER1);
+    }
+
+    public FBO getSecondaryBuffer() {
+        return get(buffersSwapped ? BUFFER1 : BUFFER2);
+    }
+
+    public FBO getFinalBuffer() {
```
emanuele3d: If you want to go down this road of making these FBOs very very special, then these methods should be renamed: getReadOnlyGBuffer(), getWriteOnlyGBuffer() and getFinalImageBuffer().

But I don't see the advantage of doing it this way versus doing it via URNs.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -33,35 +32,32 @@
...
 
     /**
      * The constructor, to be used in the initialise method of a node.
      *
      * Sample use:
      *      addDesiredStateChange(new SetViewportToSizeOf("engine:sceneOpaque", displayResolutionDependentFboManager);
      *
-     * @param fboName a URN identifying an FBO.
-     * @param frameBuffersManager the BaseFBOsManager instance that will send change notifications via the update() method of this class.
+     * @param fbo
```
emanuele3d: Please check all javadocs in which an fbo parameter has been added: please make sure to add some description to each.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFboWriteMask.java
```diff
@@ -54,70 +49,55 @@
      * @param renderToColorBuffer A boolean indicating whether the Color buffer of the given FBO should be written to.
      * @param renderToDepthBuffer A boolean indicating whether the DepthStencil buffer of the given FBO should be written to.
      * @param renderToLightBuffer A boolean indicating whether the Light Accumulation buffer of the given FBO should be written to.
-     * @param fboName A ResourceUrn identifying the FBO whose render masks have to be modified - usually only the writeOnlyGBuffer FBO.
-     * @param fboManager The FBOManager responsible for managing the given FBO.
+     * @param fbo
      */
-    public SetFboWriteMask(boolean renderToColorBuffer, boolean renderToDepthBuffer, boolean renderToLightBuffer, ResourceUrn fboName, BaseFBOsManager fboManager) {
+    public SetFboWriteMask(boolean renderToColorBuffer, boolean renderToDepthBuffer, boolean renderToLightBuffer, FBO fbo) {
```
emanuele3d: Can we put the fbo parameter as the first parameter please?

emanuele3d:
Please make sure to check the comments also in the outdated code blocks.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -127,11 +127,15 @@ public void swapReadWriteBuffers() {
         buffersSwapped = !buffersSwapped;
     }
 
-    public FBO getPrimaryBufferUrn() {
+    public FBO getPrimaryBuffer() {
```
vampcat: Okay, so I created a new SwappableFBO class, that should make this a bit more generic. However, since FBOManagers only store FBOs right now (and not SwappableFBOs), I *had* to introduce a `getGBuffer()` method that returns the only SwappableFBO (for now). That being said, I addressed your concerns of the names being more clear.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -30,14 +30,17 @@
  * TODO: Better naming
  */
 public class DisplayResolutionDependentFBOs extends AbstractFBOsManager {
-    public static final ResourceUrn READONLY_GBUFFER = new ResourceUrn("engine:sceneOpaque");
-    public static final ResourceUrn WRITEONLY_GBUFFER = new ResourceUrn("engine:sceneOpaquePingPong");
+    // TODO: Shift this kind of swapping functionality to a more generic SwappableFBO class
+    public static final ResourceUrn BUFFER1 = new ResourceUrn("engine:sceneOpaque1");
+    public static final ResourceUrn BUFFER2 = new ResourceUrn("engine:sceneOpaque2");
```
vampcat: Regarding the first part : Yeah, no more "scene" in any of the FBOs anymore, everything reflects gbuffer or final buffer.

I do have some use cases in mind, but they all rely on being able to add custom nodes to the renderer.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -120,11 +123,15 @@ private void createFBOs() {
         }
     }
 
-    // TODO: Pairing FBOs for swapping functionality
     public void swapReadWriteBuffers() {
-        FBO fbo = get(READONLY_GBUFFER);
-        fboLookup.put(READONLY_GBUFFER, get(WRITEONLY_GBUFFER));
-        fboLookup.put(WRITEONLY_GBUFFER, fbo);
-        notifySubscribers();
+        buffersSwapped = !buffersSwapped;
```
vampcat: Not sure how late into the review you wrote this comment.

My PR description now answers this question, but just to reiterate : It's not just about swapping the variables, it's the side effects caused due to the swap that I'm concerned about (update calls to every subscriber).

This might get a bit long, but..
`     CURRENT                    NEW     
sda
`

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -127,11 +127,15 @@ public void swapReadWriteBuffers() {
...
+    public FBO getPrimaryBuffer() {
         return get(buffersSwapped ? BUFFER2 : BUFFER1);
     }
 
-    public FBO getSecondaryBufferUrn() {
+    public FBO getSecondaryBuffer() {
         return get(buffersSwapped ? BUFFER1 : BUFFER2);
     }
+
+    public FBO getFinalBuffer() {
```
vampcat: Yeah, this was my bad. I changed it back.

engine/src/main/java/org/terasology/rendering/dag/nodes/InitialPostProcessingNode.java
```diff
@@ -100,11 +100,15 @@ public InitialPostProcessingNode(Context context) {
...
+        FBO one8thBloomFbo = requiresFBO(one8thScaleBloomConfig, displayResolutionDependentFBOs);
+
         int textureSlot = 0;
-        addDesiredStateChange(new SetInputTextureFromFbo(textureSlot++, READONLY_GBUFFER, ColorTexture, displayResolutionDependentFBOs, INITIAL_POST_MATERIAL, "texScene"));
+        addDesiredStateChange(new SetInputTextureFromFbo(textureSlot++, displayResolutionDependentFBOs.getPrimaryBuffer(), ColorTexture, displayResolutionDependentFBOs, INITIAL_POST_MATERIAL, "texScene"));
         addDesiredStateChange(new SetInputTexture(textureSlot++, "engine:vignette", INITIAL_POST_MATERIAL, "texVignette"));
-        setBloomInputTexture = new SetInputTextureFromFbo(textureSlot++, ONE_8TH_SCALE_FBO, ColorTexture, displayResolutionDependentFBOs, INITIAL_POST_MATERIAL, "texBloom");
-        setLightShaftsInputTexture = new SetInputTextureFromFbo(textureSlot, LIGHT_SHAFTS_FBO, ColorTexture, displayResolutionDependentFBOs, INITIAL_POST_MATERIAL, "texLightShafts");
+        setBloomInputTexture = new SetInputTextureFromFbo(textureSlot++, one8thBloomFbo, ColorTexture, displayResolutionDependentFBOs, INITIAL_POST_MATERIAL, "texBloom");
```
vampcat: ONE_8TH_SCALE_FBO is the urn of an FBO, and one8thBloomFbo is the fbo corresponding to that urn. No scaling factors involved anywhere. However, the change (from urn to fbo) is because I modified all stateChanges to accept raw FBOs (and not URBs) coz it's much cleaner (and more efficient).

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
vampcat: That is something I remember, I'm just waiting to see if you're fine with the other changes.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -251,7 +255,7 @@ private void initRenderGraph() {
         BufferClearingNode reflectedRefractedClearingNode = new BufferClearingNode(reflectedRefractedBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
         renderGraph.addNode(reflectedRefractedClearingNode, "reflectedRefractedClearingNode");
 
-        FBOConfig sceneOpaqueFboConfig = displayResolutionDependentFBOs.getFboConfig(READONLY_GBUFFER);
+        FBOConfig sceneOpaqueFboConfig = displayResolutionDependentFBOs.getFboConfig(BUFFER1);
```
vampcat: Ehh, I was low on coffee and imagination.

..Forgive me, senpai.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -86,7 +80,7 @@ public void update() {
 
     @Override
     public String toString() { // TODO: used for logging purposes at the moment, investigate different methods
-        return String.format("%30s: %s (%sx%s)", this.getClass().getSimpleName(), fboName, fboWidth, fboHeight);
+        return String.format("%30s: %s (%sx%s)", this.getClass().getSimpleName(), fbo.fboId, fboWidth, fboHeight);
     }
 
     public static void disposeDefaultInstance() {
```
vampcat: Yes, we can write it so. As you said, there *might* be a tiny loss in performance, and I'm not sure if JIT will optimize it.. but I don't think the performance loss will be anywhere near noticeable, even if we time it to microseconds.

I'll do this change in a separate commit, and we can decide if we want to keep it :D

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFboWriteMask.java
```diff
@@ -83,21 +83,21 @@ public StateChange getDefaultInstance() {
...
     public int hashCode() {
-        return Objects.hashCode(fbo.fboId);
+        return Objects.hashCode(fbo);
     }
 
     @Override
     public boolean equals(Object obj) {
         return (obj instanceof SetFboWriteMask)
-                && fbo.fboId == ((SetFboWriteMask) obj).fbo.fboId
+                && fbo == ((SetFboWriteMask) obj).fbo
```
vampcat: I might be (read: I probably am) misunderstanding you, but doesn't my change upon up the door to two FBOs having same ID ? Earlier, id was compared, so two FBOs having same id are the same thing. Now, since actual objects are being compared, two different objects might have same id, but they will still be different.

That being said, I think one fbo object can only be associated with one id and vice versa (strict one-on-one relation), so I simplified it down. What do you recommend I do?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFboWriteMask.java
vampcat: Ah, when I first created this class, I wanted to put the fboName and fboManager as first parameter, but you asked me to change it :P

engine/src/main/java/org/terasology/rendering/dag/nodes/CopyImageToScreenNode.java
```diff
@@ -47,17 +47,20 @@ public CopyImageToScreenNode(Context context) {
         WorldRenderer worldRenderer = context.get(WorldRenderer.class);
         requiresCondition(() -> worldRenderer.getCurrentRenderStage() == MONO || worldRenderer.getCurrentRenderStage() == LEFT_EYE);
 
-        displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
-        update(); // Cheeky way to initialise sceneFinalFbo
+        DisplayResolutionDependentFBOs displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
+        sceneFinalFbo = displayResolutionDependentFBOs.get(FINAL_BUFFER);
```
vampcat: Actually, I just removed the variable :P

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
emanuele3d: I will have a look.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
emanuele3d: I see.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
emanuele3d: I'll try and discuss this via slack rather than here. There might be misunderstandings that it's better to clear in an interactive session.

engine/src/main/java/org/terasology/rendering/dag/nodes/InitialPostProcessingNode.java
emanuele3d: Sorry, I got confused on that one. Indeed the concern is the change from urn as input to fbo as input.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFboWriteMask.java
emanuele3d: Hmmm. Confusing issue. 

Ok, for the way you changed it, two instances of SetFboWriteMask must hold a reference to the exact same FBO to be considered the same - never mind the mask for a moment.

_Theorethically_ however, there could be two FBO objects that are distinct but point to the same fbo GPU-side - they store the same fbo id. I don't think Terasology's code allows this so it's a very remote theory. But in that circumstance the two state changes would be considered different even though the FBOs could, in fact, be the same.

In short I agree that we should change the == fbo object comparison to fbo.equals(obj), if that's what you meant.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFboWriteMask.java
emanuele3d: Hmmm... I don't remember the rational behind that particular request. Admittedly it's not a big deal and we can leave it as it is. Somehow in this context it feels the fbo should go first: Set-Fbo-WriteMask provides the order somehow.

engine/src/main/java/org/terasology/rendering/dag/nodes/InitialPostProcessingNode.java
emanuele3d: Can you elaborate more on the rational behind this change?

emanuele3d: I mean, why do you feel it's cleaner and more efficient?

engine/src/main/java/org/terasology/rendering/dag/nodes/DeferredMainLightNode.java
```diff
@@ -101,18 +101,19 @@ public DeferredMainLightNode(Context context) {
...
+        addDesiredStateChange(new SetFboWriteMask(gBufferRead, false, false, true));
 
         initMainDirectionalLight();
 
         ShadowMapResolutionDependentFBOs shadowMapResolutionDependentFBOs = context.get(ShadowMapResolutionDependentFBOs.class);
         int textureSlot = 0;
-        addDesiredStateChange(new SetInputTextureFromFbo(textureSlot++, READONLY_GBUFFER, DepthStencilTexture, displayResolutionDependentFBOs, LIGHT_GEOMETRY_MATERIAL, "texSceneOpaqueDepth"));
-        addDesiredStateChange(new SetInputTextureFromFbo(textureSlot++, READONLY_GBUFFER, NormalsTexture, displayResolutionDependentFBOs, LIGHT_GEOMETRY_MATERIAL, "texSceneOpaqueNormals"));
-        addDesiredStateChange(new SetInputTextureFromFbo(textureSlot++, READONLY_GBUFFER, LightAccumulationTexture, displayResolutionDependentFBOs, LIGHT_GEOMETRY_MATERIAL, "texSceneOpaqueLightBuffer"));
+        addDesiredStateChange(new SetInputTextureFromFbo(textureSlot++, gBufferRead, DepthStencilTexture, displayResolutionDependentFBOs, LIGHT_GEOMETRY_MATERIAL, "texSceneOpaqueDepth"));
```
emanuele3d: Here and elsewhere, readOnlyGBuffer please.

engine/src/main/java/org/terasology/rendering/dag/nodes/ToneMappingNode.java
```diff
@@ -63,16 +63,16 @@ public ToneMappingNode(Context context) {
...
+        addDesiredStateChange(new BindFbo(toneMappingFbo));
+        addDesiredStateChange(new SetViewportToSizeOf(toneMappingFbo));
 
         addDesiredStateChange(new EnableMaterial(TONE_MAPPING_MATERIAL));
 
         toneMappingMaterial = getMaterial(TONE_MAPPING_MATERIAL);
 
         int textureSlot = 0;
-        addDesiredStateChange(new SetInputTextureFromFbo(textureSlot, INITIAL_POST_FBO, ColorTexture, displayResolutionDependentFBOs, TONE_MAPPING_MATERIAL, "texScene"));
+        addDesiredStateChange(new SetInputTextureFromFbo(textureSlot, displayResolutionDependentFBOs.get(INITIAL_POST_FBO), ColorTexture, displayResolutionDependentFBOs, TONE_MAPPING_MATERIAL, "texScene"));
```
emanuele3d: Given the FBO manager is passed in anyway, why not leave an additional constructor that takes the FBO's URN as input, retrieves the FBO from the manager and simply passes it to the constructor being used here? It would shorten this line and others like it.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFbo.java
```diff
@@ -61,22 +56,17 @@ public StateChange getDefaultInstance() {
...
-    @Override
-    public void update() {
-        fboId = fboManager.get(fboName).fboId;
+        return (obj instanceof BindFbo) && fboId == ((BindFbo) obj).fboId;
     }
 
     @Override
     public String toString() { // TODO: used for logging purposes at the moment, investigate different methods
-        return String.format("%30s: %s (fboId:%s)", this.getClass().getSimpleName(), fboName, fboId);
+        return String.format("%30s: fboId: %s", this.getClass().getSimpleName(), fboId);
```
emanuele3d: Hmmmm. I just noticed this change and it isn't good: from a human perspective having the fbo id alone in the logger output is of limited use. Having the name printed out is quite important. 

I can imagine a couple of solution:
1. we pass the fbo's URN on construction, solely for the purpose of printing it out in the log alongside the fboId
2. we store the URN in the FBO, but I'm not sure if this is going to come and bite us in the future.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetFboWriteMask.java
```diff
@@ -49,75 +44,58 @@
...
 
     @Override
-    public void update() {
-        fbo = fboManager.get(fboName);
-    }
-
-    @Override
     public String toString() {
-        return String.format("%30s: %s, %b, %b, %b", this.getClass().getSimpleName(), fboName, renderToColorBuffer, renderToDepthBuffer, renderToLightBuffer);
+        return String.format("%30s: %s, %b, %b, %b", this.getClass().getSimpleName(), fbo.getId(), renderToColorBuffer, renderToDepthBuffer, renderToLightBuffer);
```
emanuele3d: Same issue as earlier comment: the fbo name/urn is quite important for us mere mortals. =)

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFbo.java
```diff
@@ -138,15 +126,14 @@ private int fetchTextureId() {
...
         textureId = fetchTextureId();
     }
 
     @Override
     public String toString() {
         if (this != defaultInstance) {
-            return String.format("%30s: slot %s, fbo %s, textureType %s, fboManager %s, material %s, parameter '%s'", this.getClass().getSimpleName(),
-                    textureSlot, fboUrn.toString(), textureType.name(), fboManager.toString(), materialUrn.toString(), shaderParameterName);
+            return String.format("%30s: slot %s, fbo %s, textureType %s, material %s, parameter '%s'", this.getClass().getSimpleName(),
+                    textureSlot, fbo.getId(), textureType.name(), materialUrn.toString(), shaderParameterName);
```
emanuele3d: As mentioned earlier, we need the name/urn here.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -15,79 +15,59 @@
...
-        fboWidth = fbo.width();
-        fboHeight = fbo.height();
+        return (obj instanceof SetViewportToSizeOf) && (this.fbo.width() == ((SetViewportToSizeOf) obj).fbo.width())
+                                                    && (this.fbo.height() == ((SetViewportToSizeOf) obj).fbo.height());
     }
 
     @Override
     public String toString() { // TODO: used for logging purposes at the moment, investigate different methods
-        return String.format("%30s: %s (%sx%s)", this.getClass().getSimpleName(), fboName, fboWidth, fboHeight);
+        return String.format("%30s: fboId %s (%sx%s)", this.getClass().getSimpleName(), fbo.getId(), fbo.width(), fbo.height());
```
emanuele3d: Need the name/urn in the output.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -251,17 +256,21 @@ private void initRenderGraph() {
...
-        FBOConfig sceneOpaqueFboConfig = displayResolutionDependentFBOs.getFboConfig(READONLY_GBUFFER);
+        FBOConfig gBufferReadConfig = displayResolutionDependentFBOs.getFboConfig(GBUFFER_READ);
+        FBOConfig gBufferWriteConfig = displayResolutionDependentFBOs.getFboConfig(GBUFFER_WRITE);
+
+        BufferClearingNode gBufferReadClearingNode = new BufferClearingNode(gBufferReadConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+        renderGraph.addNode(gBufferReadClearingNode, "gBufferReadClearingNode");
 
-        BufferClearingNode readBufferClearingNode = new BufferClearingNode(sceneOpaqueFboConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
-        renderGraph.addNode(readBufferClearingNode, "readBufferClearingNode");
+        BufferClearingNode gBufferWriteClearingNode = new BufferClearingNode(gBufferWriteConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
```
emanuele3d: gBufferWriteClearingNode -> writeOnlyGBufferClearningNode

engine/src/main/java/org/terasology/rendering/dag/nodes/AmbientOcclusionNode.java
```diff
@@ -74,13 +72,15 @@ public AmbientOcclusionNode(Context context) {
...
-        update(); // Cheeky way to initialise ssaoFbo, outputFboWidth, outputFboHeight
+        DisplayResolutionDependentFBOs displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
+        ssaoFbo = requiresFBO(new FBOConfig(SSAO_FBO, FULL_SCALE, FBO.Type.DEFAULT), displayResolutionDependentFBOs);
+        addDesiredStateChange(new BindFbo(ssaoFbo));
+        addDesiredStateChange(new SetViewportToSizeOf(ssaoFbo));
+        update(); // Cheeky way to initialise outputFboWidth, outputFboHeight
         displayResolutionDependentFBOs.subscribe(this);
 
+        ShaderParametersSSAO.setGBufferRead(displayResolutionDependentFBOs.getGBuffer().getWriteFbo());
+
```
emanuele3d: 1. Is there an indentation issue on this line?
2. Why are you setting the writeOnlyFbo as the readOnlyFbo via setGBufferRead?
3. All instances of setGBufferRead()/setGBufferWrite() should be changed to setReadOnlyGBuffer/setWriteOnlyGBuffer.

engine/src/main/java/org/terasology/rendering/dag/nodes/OutlineNode.java
```diff
@@ -53,14 +52,12 @@
...
 
     private Material outlineMaterial;
 
-    @SuppressWarnings("FieldCanBeLocal")
-    private FBO sceneOpaqueFbo;
-    private float sceneOpaqueFboWidth;
-    private float sceneOpaqueFboHeight;
+    private FBO gBufferRead;
+    private float primaryBufferWidth;
+    private float primaryBufferHeight;
```
emanuele3d: gBufferWidth/gBufferHeight please.

engine/src/main/java/org/terasology/rendering/dag/nodes/OutlineNode.java
```diff
@@ -53,14 +52,12 @@
...
     private SubmersibleCamera activeCamera;
-    private DisplayResolutionDependentFBOs displayResolutionDependentFBOs;
 
     private Material outlineMaterial;
 
-    @SuppressWarnings("FieldCanBeLocal")
-    private FBO sceneOpaqueFbo;
-    private float sceneOpaqueFboWidth;
-    private float sceneOpaqueFboHeight;
+    private FBO gBufferRead;
```
emanuele3d: readOnlyGBuffer please.

engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
```diff
@@ -102,9 +101,13 @@ public PrePostCompositeNode(Context context) {
         worldRenderer = context.get(WorldRenderer.class);
         activeCamera = worldRenderer.getActiveCamera();
 
-        displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
+        DisplayResolutionDependentFBOs displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
+        SwappableFBO gBuffer = displayResolutionDependentFBOs.getGBuffer();
```
emanuele3d: gBuffer -> gBufferPair please.

engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
```diff
@@ -102,9 +101,13 @@ public PrePostCompositeNode(Context context) {
...
+
+        gBuffer.swap();
```
emanuele3d: Again, why is this done here rather than in process(), every frame?

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
```diff
@@ -155,9 +151,10 @@ public RefractiveReflectiveBlocksNode(Context context) {
...
         addDesiredStateChange(new LookThrough(activeCamera));
 
-        displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
-        requiresFBO(new FBOConfig(REFRACTIVE_REFLECTIVE_FBO, FULL_SCALE, FBO.Type.HDR).useNormalBuffer(), displayResolutionDependentFBOs);
-        addDesiredStateChange(new BindFbo(REFRACTIVE_REFLECTIVE_FBO, displayResolutionDependentFBOs));
+        DisplayResolutionDependentFBOs displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
+        gBufferRead = displayResolutionDependentFBOs.getGBuffer().getWriteFbo();
+        refractiveReflectiveFbo = requiresFBO(new FBOConfig(REFRACTIVE_REFLECTIVE_FBO, FULL_SCALE, FBO.Type.HDR).useNormalBuffer(), displayResolutionDependentFBOs);
+        addDesiredStateChange(new BindFbo(refractiveReflectiveFbo));
         update(); // Cheeky way to initialise readOnlyGBufferFbo, refractiveReflectiveFbo
```
emanuele3d: The comment no longer seems relevant here.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -251,17 +256,21 @@ private void initRenderGraph() {
...
+        renderGraph.addNode(gBufferWriteClearingNode, "gBufferWriteClearingNode");
 
         Node backdropNode = new BackdropNode(context);
         renderGraph.addNode(backdropNode, "backdropNode");
 
         String aLabel = "hazeIntermediateNode";
         FBOConfig hazeIntermediateConfig = new FBOConfig(HazeNode.INTERMEDIATE_HAZE_FBO, ONE_16TH_SCALE, FBO.Type.DEFAULT);
-        HazeNode hazeIntermediateNode = new HazeNode(context, sceneOpaqueFboConfig, hazeIntermediateConfig, aLabel);
+        HazeNode hazeIntermediateNode = new HazeNode(context, gBufferReadConfig, hazeIntermediateConfig, aLabel);
```
emanuele3d: gBufferReadConfig -> readOnlyGBufferConfig

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -24,23 +24,25 @@
...
  */
 public class DisplayResolutionDependentFBOs extends AbstractFBOsManager {
     // TODO: Shift this kind of swapping functionality to a more generic SwappableFBO class
-    public static final ResourceUrn BUFFER1 = new ResourceUrn("engine:sceneOpaque1");
-    public static final ResourceUrn BUFFER2 = new ResourceUrn("engine:sceneOpaque2");
-    public static final ResourceUrn FINAL_BUFFER = new ResourceUrn("engine:sceneFinal");
+    public static final ResourceUrn GBUFFER = new ResourceUrn("engine:gBuffer");
+    public static final ResourceUrn GBUFFER_READ = new ResourceUrn("engine:gBuffer");
+    public static final ResourceUrn GBUFFER_WRITE = new ResourceUrn("engine:gBuffer");
+    public static final ResourceUrn FINAL_BUFFER = new ResourceUrn("engine:final");
```
emanuele3d: And I also think the variable names should be: READONLY_GBUFFER and WRITEONLY_GBUFFER.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
emanuele3d: Had a look and I now understand better your line of thinking. Let's see where that leads us.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
emanuele3d: I still have some concerns about this, i.e. the .swap() method called in the constructor of nodes rather than in process. But overall I think I understand this better and it makes sense.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
emanuele3d: Meanwhile, I can now see one advantage of doing it this way: we don't have to pass the fboManager as input.

engine/src/main/java/org/terasology/rendering/dag/nodes/InitialPostProcessingNode.java
emanuele3d: Never mind, I got it.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
emanuele3d: Yep, I'm fine with those changes and thank you for completing the javadocs.

emanuele3d:
All good now, I reviewed the review (!) and added some more comments/reply in the the previous review, showing as outdated commits.

engine/src/main/java/org/terasology/rendering/dag/nodes/ApplyDeferredLightingNode.java
```diff
@@ -53,22 +45,26 @@
...
 
-    private DisplayResolutionDependentFBOs displayResolutionDependentFBOs;
-
     public ApplyDeferredLightingNode(Context context) {
-        displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
-        addDesiredStateChange(new BindFbo(WRITEONLY_GBUFFER, displayResolutionDependentFBOs));
+        DisplayResolutionDependentFBOs displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
+        SwappableFBO gBuffer = displayResolutionDependentFBOs.getGBuffer();
+
+        gBuffer.swap();
```
vampcat: No. I think it would be easier to explain the reasoning on Slack.

engine/src/main/java/org/terasology/rendering/dag/nodes/ToneMappingNode.java
vampcat: Okay, this is something I have to agree with - for now.

However, since we want to move towards more self-contained Nodes, all the shaders (materials), fbos and everything will be created by the Node itself, and will be directly passed to the StateChanges as required.

At least that's what I hope will happen ^^

engine/src/main/java/org/terasology/rendering/dag/nodes/AmbientOcclusionNode.java
vampcat: I can't seem to find any indentation issue here (neither on github, nor in IntelliJ). That being said, it did look weird on my mobile in GMail.. strange.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFbo.java
vampcat: Going with passing a string for now, will discuss in detail on slack.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/ToneMappingNode.java
emanuele3d: Actually, it all depends on how far we want to push the graph concept. 

One possible avenue is to have DataNodes (i.e. FBOs, cameras, perhaps shaders/materials) and RenderingNodes, the former being connected to the latter on graph construction. In this context it wouldn't be nodes that create everything but rather the objects that create the nodes. For example a module might inject its nodes into the DAG and also take care of creating all the connections between them and the existing DAG. 

I am happy to continue along the lines you mentioned though, leaving the concept above for the future, if ever.

engine/src/main/java/org/terasology/rendering/dag/nodes/BufferClearingNode.java
```diff
@@ -56,28 +56,40 @@ public BufferClearingNode(FBOConfig fboConfig, BaseFBOsManager fboManager, int c
...
+     * <p>
      * This method is executed within a NodeTask in the Render Tasklist.
      */
     @Override
     public void process() {
         glClear(clearingMask);
     }
 
-    private boolean validateArguments(FBOConfig anFboConfig, BaseFBOsManager anFboManager, int clearingMask) {
+    private boolean validateArguments(FBOConfig fboConfig, BaseFBOsManager fboManager, int clearingMask) {
```
emanuele3d: IF you must.

:smile:

engine/src/main/java/org/terasology/rendering/opengl/AbstractFBOsManager.java
```diff
@@ -44,8 +44,8 @@
  * rendering engine.
  * <p>
  * Default FBOs:
- * sceneOpaque:  Primary FBO: most visual information eventually ends up here
- * sceneOpaquePingPong:  The sceneOpaque FBOs are swapped every frame, to use one for reading and the other for writing
+ * writeOnlyGBuffer:  Primary FBO: most visual information eventually ends up here
+ * readOnlyGBuffer:  The gbuffer FBOs are swapped around when needed, to use one for reading and the other for writing
```
emanuele3d: I'd write:

```
* writeOnlyGBuffer: this is the primary FBO and its attachments are the target of a good number of rendering operations, eventually storing most information needed for the deferred rendering.
* readOnlyGBuffer: this FBO holds attachments that are used as input textures by a number of rendering stages.
* The two FBOs above are from time to time swapped. But this is handled transparently and getWriteOnlyFbo() will always return the FBO appropriate for writing.
```

emanuele3d: By the way, the sceneBloom# fbos below can be described as:

`A number of buffers used to repeatedly blur and downsample the content of the high pass fbo. The resulting blurred version of the high pass is then used for the bloom effect. `

And there is no more ocUndistorted fbo, right?

And we should probably remove all those "scene" terms.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -97,8 +95,8 @@ private void updateFullScale() {
     public void update() {
         updateFullScale();
 
-        FBO sceneOpaqueFbo = gBuffer.getReadFbo();
-        if (sceneOpaqueFbo.dimensions().areDifferentFrom(fullScale)) {
+        FBO readOnlyGBuffer = gBufferPair.getReadFbo();
+        if (readOnlyGBuffer.dimensions().areDifferentFrom(fullScale)) {
             // disposeAllFBOs();
             // createFBOs();
```
emanuele3d: Let's remove these if they are no longer in use.

engine/src/main/java/org/terasology/rendering/dag/nodes/BlurredAmbientOcclusionNode.java
```diff
@@ -54,7 +54,6 @@
     public static final ResourceUrn SSAO_BLURRED_FBO = new ResourceUrn("engine:ssaoBlurred");
     private static final ResourceUrn SSAO_FBO = new ResourceUrn("engine:ssao");
```
emanuele3d: Should we perhaps rename this as SSAO_BLURRED_URN and SSAO_URN? They are not FBOs! And please also check all other URNs elsewhere with the _FBO suffix.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFbo.java
```diff
@@ -83,6 +83,34 @@ public SetInputTextureFromFbo(int textureSlot, FBO fbo, FboTexturesTypes texture
         fboManager.subscribe(this);
     }
 
+    /**
+     * The constructor, to be used in the initialise method of a node.
+     *
+     * Sample use:
+     *      addDesiredStateChange(new SetInputTextureFromFbo(0, fbo, ColorTexture,
+     *                                  displayResolutionDependentFboManager, "engine:prog.chunk", "textureWater"));
```
emanuele3d: `fbo -> fboUrn` or `fbo -> FBO_URN`

?

GooeyHub:

emanuele3d:
I checked the last three commits and they are all good for me. We just have to catch up on any unfinished business.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SwapGBuffers.java
```diff
@@ -0,0 +1,79 @@
...
+import org.terasology.rendering.opengl.SwappableFBO;
+
+/**
+ * This StateChange is used to swap the readOnly and writeOnly buffers in the gBufferPair.
+ * This is useful for Nodes that want to update the value of the gBuffer, by passing it's contents through a shader.
+ *
+ * Note that this StateChange is special because unlike all the other StateChanges that modify the state of the
+ * system during the execution of the task list, this StateChange does it's job by indirectly changing what data gets
+ * sent to future StateChanges, by causing a change in the return value of gBufferPair.get*Buffer().
+ *
```
vampcat: I'm thinking of changing the middle paragraph to:

` * Note that this StateChange is special because unlike all the other StateChanges that modify the state of the
 * system during the execution of the task list, this StateChange does it's job during the initialization phase,
 * changing the final state of the system indirectly by changing what data gets sent to future StateChanges.
 * It does so by causing a change in the return value of gBufferPair.get*Buffer().`

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SwapGBuffers.java
```diff
@@ -0,0 +1,79 @@
...
+ * Due to this, this StateChange -has- to be added before any calls to gBuferPair.get*Buffer() in the Node.
+ */
+public class SwapGBuffers implements StateChange {
+    private static StateChange defaultInstance = new EmptyStateChange();
```
vampcat: I later changed this to null and deleted the EmptyStateChange class, I'll push that in next commit.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
```diff
@@ -15,79 +15,59 @@
...
+    public SetViewportToSizeOf(FBO fbo) {
+        this.fbo = fbo;
     }
 
     @Override
     public StateChange getDefaultInstance() {
         if (defaultInstance == null) {
-            defaultInstance = new SetViewportToSizeOf(READONLY_GBUFFER, CoreRegistry.get(DisplayResolutionDependentFBOs.class));
+            DisplayResolutionDependentFBOs displayResolutionDependentFBOs = CoreRegistry.get(DisplayResolutionDependentFBOs.class);
+            assert(displayResolutionDependentFBOs != null);
```
emanuele3d: Why all of a sudden are you worried this fboManager could be null? It is one of the first thing initialized in WorldRendererImpl.initRenderGraph(). Nodes and state changes only exist after that, I thought.

engine/src/main/java/org/terasology/rendering/opengl/AbstractFBOsManager.java
```diff
@@ -44,11 +44,11 @@
  * rendering engine.
  * <p>
  * Default FBOs:
- * sceneOpaque:  Primary FBO: most visual information eventually ends up here
- * sceneOpaquePingPong:  The sceneOpaque FBOs are swapped every frame, to use one for reading and the other for writing
+ * writeOnlyGBuffer: this is the primary FBO and its attachments are the target of a good number of rendering operations, eventually storing most information needed for the deferred rendering.
+ * readOnlyGBuffer: this FBO holds attachments that are used as input textures by a number of rendering stages.
+ * Note that these two FBOs might get swapped during the DAG creation stage, on a node's request. However, this is handled transparently and getWriteOnlyFbo() will always return the FBO appropriate for writing.
```
emanuele3d: Note that these two FBOs **may be swapped** on a node's request **(ping-pong technique). See the DisplayResolutionDependentFboManager class for more details.**

engine/src/main/java/org/terasology/rendering/opengl/SwappableFBO.java
```diff
@@ -0,0 +1,48 @@
...
+        this.readFbo = readFbo;
+        this.writeFbo = writeFbo;
+        isSwapped = false;
+    }
+
+    public FBO getReadFbo() {
+        return isSwapped ? writeFbo : readFbo;
+    }
+
+    public FBO getWriteFbo() {
```
emanuele3d: getReadOnlyFbo()/getWriteOnlyFbo() please.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -90,32 +95,30 @@ private void updateFullScale() {
...
         }
-        fboLookup.clear();
-    }
+   }
 
-    private void createFBOs() {
-        for (FBOConfig fboConfig : fboConfigs.values()) {
-            generateWithDimensions(fboConfig, fullScale.multiplyBy(fboConfig.getScale()));
+    private void disposeAllFbos() {
+        // TODO: This should be public, and should be called while disposing the DRDFboM to prevent leaks.
```
emanuele3d: Sounds dangerous given that third-party nodes might have access to it - to be discussed. How about package-private? Any why not do it right away?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -247,25 +252,29 @@ private void initRenderGraph() {
...
-        FBOConfig sceneOpaqueFboConfig = displayResolutionDependentFBOs.getFboConfig(READONLY_GBUFFER);
+        FBOConfig readOnlyGBufferConfig = displayResolutionDependentFBOs.getFboConfig(READONLY_GBUFFER);
+        FBOConfig writeOnlyGBufferConfig = displayResolutionDependentFBOs.getFboConfig(WRITEONLY_GBUFFER);
+
+        BufferClearingNode readOnlyGBufferClearingNode = new BufferClearingNode(readOnlyGBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+        renderGraph.addNode(readOnlyGBufferClearingNode, "readOnlyGBufferClearingNode");
 
-        BufferClearingNode readBufferClearingNode = new BufferClearingNode(sceneOpaqueFboConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
-        renderGraph.addNode(readBufferClearingNode, "readBufferClearingNode");
+        BufferClearingNode writeOnlyGBufferClearingNode = new BufferClearingNode(writeOnlyGBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
```
emanuele3d: As I look at this I wonder if it'd be worth to create a GBufferClearingNode that also clear the additional buffers (i.e. normals and light accumulation) , perhaps taking advantage of an FBO.clear(mask) method that wraps things neatly. Something to think about, not to solve in this PR.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -247,25 +252,29 @@ private void initRenderGraph() {
...
+        renderGraph.addNode(writeOnlyGBufferClearingNode, "writeOnlyGBufferClearingNode");
 
         Node backdropNode = new BackdropNode(context);
         renderGraph.addNode(backdropNode, "backdropNode");
 
         String aLabel = "hazeIntermediateNode";
-        FBOConfig hazeIntermediateConfig = new FBOConfig(HazeNode.INTERMEDIATE_HAZE_FBO, ONE_16TH_SCALE, FBO.Type.DEFAULT);
-        HazeNode hazeIntermediateNode = new HazeNode(context, sceneOpaqueFboConfig, hazeIntermediateConfig, aLabel);
+        FBOConfig hazeIntermediateConfig = new FBOConfig(HazeNode.INTERMEDIATE_HAZE_FBO_URI, ONE_16TH_SCALE, FBO.Type.DEFAULT);
+        HazeNode hazeIntermediateNode = new HazeNode(context, writeOnlyGBufferConfig, hazeIntermediateConfig, aLabel);
```
emanuele3d: Hmmm... I just checked and the first fboConfig argument is used to obtain an -input- fbo. So, here an in the other nodes extending the BlurNode class, it might be worth using the lastUpdatedBuffer naming.

That been said, my mind somewhat fogged by the late hour, shouldn't this be affected by the swaps in previous nodes? Not so much in these early nodes, but later in the node list.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -326,7 +335,7 @@ private void initRenderGraph() {
         renderGraph.addNode(initialPostProcessingNode, "initialPostProcessingNode");
 
         aLabel = "downSampling_gBuffer_to_16x16px_forExposure";
-        DownSamplerForExposureNode exposureDownSamplerTo16pixels = new DownSamplerForExposureNode(context, sceneOpaqueFboConfig, displayResolutionDependentFBOs, FBO_16X16_CONFIG, immutableFBOs, aLabel);
+        DownSamplerForExposureNode exposureDownSamplerTo16pixels = new DownSamplerForExposureNode(context, writeOnlyGBufferConfig, displayResolutionDependentFBOs, FBO_16X16_CONFIG, immutableFBOs, aLabel);
```
emanuele3d: Follow up from previous comment, a few lines above: for example who is to say that in here we don't need the content of the readOnlyGBuffer instead? And in fact in this case theoretically we should because this node only reads from the gBuffer, and it should have the latest content from it.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -262,7 +262,7 @@ private void initRenderGraph() {
         BufferClearingNode readOnlyGBufferClearingNode = new BufferClearingNode(readOnlyGBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
         renderGraph.addNode(readOnlyGBufferClearingNode, "readOnlyGBufferClearingNode");
 
-        BufferClearingNode writeOnlyGBufferClearingNode = new BufferClearingNode(writeOnlyGBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+        BufferClearingNode writeOnlyGBufferClearingNode = new BufferClearingNode(writeOnlyGBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
         renderGraph.addNode(writeOnlyGBufferClearingNode, "writeOnlyGBufferClearingNode");
```
vampcat: Before this PR, the only one GBuffer was being cleared here, and the second was being cleared in some random Node in the process method() (don't even remember which buffer it was), along with a comment asking why it was necessary. I removed that comment and that glClear from that random Node, and shifted it to a proper BufferClearingNode to make things a lot more explicit. I also mentioned this specifically via comment in the commit that I made this change.

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
vampcat: Mostly because I wanted to discuss the idea with you, and because I want to do this (and all other disposal tweaks) in one large PR - since I'm sure there are a lot of leaks :)

engine/src/main/java/org/terasology/rendering/opengl/fbms/DisplayResolutionDependentFBOs.java
```diff
@@ -97,20 +97,21 @@ public void update() {
...
         for (SimpleUri urn : fboConfigs.keySet()) {
             FBOConfig fboConfig = getFboConfig(urn);
             fboConfig.setDimensions(fullScale.multiplyBy(fboConfig.getScale()));
             FBO.recreate(get(urn), getFboConfig(urn));
         }
    }
 
-    private void disposeAllFBOs() {
+    private void disposeAllFbos() {
+        // TODO: This should be public, and should be called while disposing the DRDFboM to prevent leaks.
```
vampcat: YPotLF? I'm actually interested in knowing what that means xD

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetViewportToSizeOf.java
vampcat: I mostly did this to avoid warnings by IntelliJ.

However, now that I think about it, the warning is a sign of (possibly) bad code, and shouldn't be covered up.

I'll revert this!

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
vampcat: Yeah, you're correct, this will be affected later in the list.
I'll fix this! :)

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
vampcat: Okay, I have no clue what we should be doing - I'll have to look at the Nodes again to be certain.
However, as mentioned before, I've only renamed things in this area, and haven't actually changed anything here.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -227,49 +231,86 @@ private void initRenderingSupport() {
...
-        // sky rendering
-        FBOConfig reflectedRefractedBufferConfig = new FBOConfig(RefractiveReflectiveBlocksNode.REFRACTIVE_REFLECTIVE_FBO, FULL_SCALE, FBO.Type.HDR).useNormalBuffer();
+    private void addSkyNodes(RenderGraph renderGraph) {
+        FBOConfig reflectedRefractedBufferConfig = new FBOConfig(RefractiveReflectiveBlocksNode.REFRACTIVE_REFLECTIVE_FBO_URI, FULL_SCALE, FBO.Type.HDR).useNormalBuffer();
         BufferClearingNode reflectedRefractedClearingNode = new BufferClearingNode(reflectedRefractedBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
         renderGraph.addNode(reflectedRefractedClearingNode, "reflectedRefractedClearingNode");
 
-        FBOConfig sceneOpaqueFboConfig = displayResolutionDependentFBOs.getFboConfig(READONLY_GBUFFER);
+        FBOConfig readOnlyGBufferConfig = displayResolutionDependentFBOs.getFboConfig(READONLY_GBUFFER);
+        FBOConfig writeOnlyGBufferConfig = displayResolutionDependentFBOs.getFboConfig(WRITEONLY_GBUFFER);
```
emanuele3d: Let's add // TODO: change terminology to match the stale/lastUpdated dichotomy. We can do this in a separate PR as I know you look forward wrapping this PR up.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -309,41 +354,46 @@ private void initRenderGraph() {
...
-        renderGraph.addNode(exposureDownSamplerTo16pixels, aLabel);
+        FBOConfig writeOnlyGBufferConfig = displayResolutionDependentFBOs.getFboConfig(WRITEONLY_GBUFFER);
+        String label = "downSampling_gBuffer_to_16x16px_forExposure";
+        DownSamplerForExposureNode exposureDownSamplerTo16pixels = new DownSamplerForExposureNode(context, writeOnlyGBufferConfig, displayResolutionDependentFBOs, FBO_16X16_CONFIG, immutableFBOs, label);
+        renderGraph.addNode(exposureDownSamplerTo16pixels, label);
 
-        aLabel = "downSampling_16x16px_to_8x8px_forExposure";
-        DownSamplerForExposureNode exposureDownSamplerTo8pixels = new DownSamplerForExposureNode(context, FBO_16X16_CONFIG, immutableFBOs, FBO_8X8_CONFIG, immutableFBOs, aLabel);
-        renderGraph.addNode(exposureDownSamplerTo8pixels, aLabel);
+        label = "downSampling_16x16px_to_8x8px_forExposure";
```
emanuele3d: You are getting yourself distracted here... :smile:

engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
```diff
@@ -150,6 +148,8 @@ public PrePostCompositeNode(Context context) {
         if (hazeIsEnabled) {
             addDesiredStateChange(setHazeInputTexture);
         }
+
+        addDesiredStateChange(new SwapGBuffers(gBufferPair));
```
emanuele3d: I guess we'll need to add a warning in the code here, to refer to the SwapGBuffers class and on why this is counterintuitive but correctly placed here.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -232,15 +232,41 @@ private void initRenderingSupport() {
...
+
+        addRefractiveReflectiveNode(renderGraph);
+
+        add3dDecorationNodes(renderGraph);
+
+        addPrePostProcessingNodes(renderGraph);
+
+        addPostProcessingNodes(renderGraph);
+
+        addImageToSceeenNodes(renderGraph);
```
vampcat: Yes, I'm still aware of that issue - however, I'm waiting for shader overhaul before I start splitting shaders, and consequently, Nodes.

engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
vampcat: I'm not sure which commit you're looking at.

In the read/write terminology, I correctly swapped it before the rendering.
In the latest/stale terminology, I correctly swapped it after the rendering.

The latest commit should have the swap -after- the rendering.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
vampcat: I took care of this in one last commit for this pr :D

GooeyHub:

emanuele3d:
Even with the latest commits this PR remains approved and I believe it's ready for merge @Cervator. 

@vampcat, from the next PRs I'll try to be more strict with the scope and size of the changes. It is very difficult for both of us to deal with a big diff like the one of this PR. Even if a set of changes potentially affects many files, try to find a way to restrict the changes to a dozen or so. Once we work out the issues on those the others will be easier to handle. That's the theory at least.

GooeyHub:

GooeyHub:

Cervator:
Poked at it around a while and didn't see anything particularly weird - merged! :-)

emanuele3d:
Thanks @Cervator for the merge, thanks @vampcat for the hard work!

PR #2993, opened 2017-06-29
FloatingTextRenderer accepts multiple \n separated lines

nihal111:
This PR allows the FloatingTextRenderer to now split lines and separate them at "\n".
The FloatingTextComponent can now have the `text` parameter contain multiple "\n" separated lines.

The width for the mesh is calculated based on the longest length String after having broken the text. 

This is an example of how it looks:
terasology-170629201022-1152x720

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
@@ -41,6 +41,7 @@ import org.terasology.rendering.opengl.OpenGLUtils;
 import org.terasology.world.WorldProvider;
 
 import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 
 import static org.lwjgl.opengl.GL11.GL_DEPTH_TEST;
@@ -99,18 +100,24 @@ public class FloatingTextRenderer extends BaseComponentSystem implements  Render
 
             FloatingTextComponent floatingText = entity.getComponent(FloatingTextComponent.class);
 
-            String text = floatingText.text;
+            List<String> text = Arrays.asList(floatingText.text.split("\n"));
             Color baseColor = floatingText.textColor;
             Color shadowColor = floatingText.textShadowColor;
             boolean underline = false;
-            int textWidth = font.getWidth(text);
+
+            String longest = "";
+            for (String s : text) {
+                if (s.length() > longest.length())
+                    longest = s;
+            }
+            int textWidth = font.getWidth(longest);
 
             FontMeshBuilder meshBuilder = new FontMeshBuilder(underlineMaterial);
 
             Map<Material, Mesh> meshMap = entityMeshCache.get(entity);
             if (meshMap == null) {
                 meshMap = meshBuilder
-                        .createTextMesh(font, Arrays.asList(text), textWidth, HorizontalAlign.CENTER, baseColor,
+                        .createTextMesh(font, text, textWidth, HorizontalAlign.CENTER, baseColor,
                                 shadowColor, underline);
                 entityMeshCache.put(entity, meshMap);
             }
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
```diff
@@ -99,18 +100,24 @@ private void render(Iterable<EntityRef> floatingTextEntities) {
 
             FloatingTextComponent floatingText = entity.getComponent(FloatingTextComponent.class);
 
-            String text = floatingText.text;
+            List<String> text = Arrays.asList(floatingText.text.split("\n"));
```
emanuele3d: Can you check if you can simply write here:

`String[] text = floatingText.text.split("\n");`

GooeyHub:

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
emanuele3d: I'd also rename the variable to `linesOfText`.

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
```diff
@@ -99,18 +100,24 @@ private void render(Iterable<EntityRef> floatingTextEntities) {
...
             Color shadowColor = floatingText.textShadowColor;
             boolean underline = false;
-            int textWidth = font.getWidth(text);
+
+            String longest = "";
+            for (String s : text) {
+                if (s.length() > longest.length())
+                    longest = s;
+            }
+            int textWidth = font.getWidth(longest);
```
nihal111: Nice catch 🙂

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
```diff
@@ -100,24 +93,23 @@ private void render(Iterable<EntityRef> floatingTextEntities) {
...
+            String[] linesOfText = floatingText.text.split("\n");
             Color baseColor = floatingText.textColor;
             Color shadowColor = floatingText.textShadowColor;
             boolean underline = false;
 
-            String longest = "";
-            for (String s : text) {
-                if (s.length() > longest.length())
-                    longest = s;
+            int textWidth = font.getWidth("");
```
emanuele3d: You don't need a potentially expensive call to getWidth()! Just set it to 0. =)

GooeyHub:

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
```diff
@@ -100,24 +93,23 @@ private void render(Iterable<EntityRef> floatingTextEntities) {
...
             }
-            int textWidth = font.getWidth(longest);
 
             FontMeshBuilder meshBuilder = new FontMeshBuilder(underlineMaterial);
 
             Map<Material, Mesh> meshMap = entityMeshCache.get(entity);
             if (meshMap == null) {
                 meshMap = meshBuilder
-                        .createTextMesh(font, text, textWidth, HorizontalAlign.CENTER, baseColor,
+                        .createTextMesh(font, Arrays.asList(linesOfText), textWidth, HorizontalAlign.CENTER, baseColor,
```
emanuele3d: Ok, looking at the code I can't see a reason why the createTextMesh() method can't accept a string array instead. Can you take a look? It will cascade in a few other changes but they might still be safe to do.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
emanuele3d: We discussed this in slack and we agreed to leave it as it would impact code elsewhere.

emanuele3d:
@flo, @Cervator: GO for merge from my side.

Cervator:
Merged! :-)

emanuele3d:
Good job everybody! Especially @nihal111!

PR #3001, opened 2017-07-08
Fixes block particle effects

skaldarnar:
- particles when damaging a block now have the block's texture (randomly selected from all sides)
001-block-damage-particle-effect
002-block-damage-particle-effect
- smoke effect is only shown when block is destroyed and drops debris
003-block-damage-particle-effect

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
@@ -18,6 +18,7 @@ package org.terasology.world.block.entity;
 import org.terasology.audio.AudioManager;
 import org.terasology.audio.StaticSound;
 import org.terasology.audio.events.PlaySoundEvent;
+import org.terasology.entitySystem.entity.EntityBuilder;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.event.EventPriority;
@@ -155,6 +156,14 @@ public class BlockEntitySystem extends BaseComponentSystem {
         BlockDamageModifierComponent blockDamageModifierComponent = event.getDamageType().getComponent(BlockDamageModifierComponent.class);
         // TODO: Configurable via block definition
         if (blockDamageModifierComponent == null || !blockDamageModifierComponent.skipPerBlockEffects) {
+            // dust particle effect
+            if (entity.hasComponent(LocationComponent.class) && block.isDebrisOnDestroy()) {
+                EntityBuilder dustBuilder = entityManager.newBuilder("core:dustEffect");
+                dustBuilder.getComponent(LocationComponent.class).setWorldPosition(entity.getComponent(LocationComponent.class).getWorldPosition());
+                dustBuilder.build();
+            }
+
+            // sound to play for destroyed block
             BlockSounds sounds = block.getSounds();
             if (!sounds.getDestroySounds().isEmpty()) {
                 StaticSound sound = random.nextItem(sounds.getDestroySounds());
--- a/modules/Core/assets/prefabs/particleEffects/defaultBlockParticles.prefab
+++ b/modules/Core/assets/prefabs/particleEffects/defaultBlockParticles.prefab
@@ -12,7 +12,7 @@
         "maxVelocity": [1.5, 4, 1.5]
     },
     "scaleRangeGenerator": {
-        "minScale": [0.2, 0.2, 0.2],
+        "minScale": [0.1, 0.1, 0.1],
         "maxScale": [0.2, 0.2, 0.2]
     },
     "textureOffsetGenerator": {},
--- a/modules/Core/src/main/java/org/terasology/logic/health/BlockDamageAuthoritySystem.java
+++ b/modules/Core/src/main/java/org/terasology/logic/health/BlockDamageAuthoritySystem.java
@@ -37,6 +37,7 @@ import org.terasology.utilities.Assets;
 import org.terasology.utilities.random.FastRandom;
 import org.terasology.utilities.random.Random;
 import org.terasology.world.block.Block;
+import org.terasology.world.block.BlockAppearance;
 import org.terasology.world.block.BlockComponent;
 import org.terasology.world.block.BlockManager;
 import org.terasology.world.block.BlockPart;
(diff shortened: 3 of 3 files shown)


GooeyHub:

skaldarnar:
Should I make the particles smaller then and select only a part of the texture? The offset calculation seemed to be off before, when hitting a door you would get diamond particles instead... 
Is there an easy way to just create particle sprites of a specific color? We could just sample the color from a random position in the texture and apply that as flat colored sprite.

GooeyHub:

skaldarnar:
Updated the particle texture to be sampled randomly from the sides of the damaged block. The texture size is reduce to sample only a single pixel, i.e., a flat color from the texture. For each block part, 8 samples are taken.

In addition, the particle sizes vary slightly (0.1 to 0.2).

005-block-damage-particle-effect

@MaxBorsch is there an easy way to apply a custom function `Vector2f -> Vector2f` to the selected offset in the `TextureGeneratorFunction`? The sampling of `n` pixels in only necessary because I could not figure out how specify a general function/modifier to be applied to the selected offset.

MaxBorsch:
@skaldarnar I'm not sure of what you're asking for, but why not just increase the textureoffsetsize to move than a single pixel to get smaller chunks of the texture? Like a few pixels?

skaldarnar:
Ah, let me try to break it down. 

First, I was using the whole texture tile (as can be seen above) randomly selected from the block's parts. This basically gave me a list of 6 coordinate offsets (front/back, left/right, top/bottom) which were all added to the `TextureOffsetGeneratorComponent`. AFAIK the randomness of the displayed texture comes from the implementation of the texture offset generator function.
```Java
final List<Vector2f> offsets = 
  Arrays.stream(BlockPart.sideValues())
    .map(blockAppearance::getTextureAtlasPos)
    .collect(Collectors.toList());
textureOffsetGeneratorComponent.validOffsets.addAll(offsets);
```

Since we only want to display a random part of the texture, we need to add random local offset. Ideally, this is done every time a new texture is required, i.e., in the texture offset generator function. The modifier function should be executed each time an offset is queried. So, want I _want_ to write is:
```Java
final List<Vector2f> offsets = 
  Arrays.stream(BlockPart.sideValues())
    .map(blockAppearance::getTextureAtlasPos)
    .collect(Collectors.toList());
textureOffsetGeneratorComponent.validOffsets.addAll(offsets);

textureOffsetGeneratorComponent.offsetModifierFunction(coord -> 
  new Vector2f(coord).add(random.nextInt(absoluteTileSize) * pixelSize, random.nextInt(absoluteTileSize) * pixelSize))
```

However, I'm currently just generating a number of predefined (local) offsets for each block part offset. 
For small texture sizes, one could also iterate over all possible offsets...

PS: maybe I'm thinking too functional here, and passing around "functions" is still not a good idea in Java. It is not strictly necessary, but would help a lot in randomizing texture offsets in general.

skaldarnar:
Regardless of the discussion, I think this can be merged if there are no objections....

emanuele3d:
Quick note to say I'm having some difficulties loading the animated GIFs through the relatively slow WiFi connection I have here in holidays. 

I'd also prefer that a random sample of the block's texture is used rather than a single solid color or the whole texture is used. 

I will try and contribute more to this thread in a separate post as soon as I have caught up with everything else.

skaldarnar:
I'm sorry for the huge file sizes, was just trying out a tool and didn't realize it is not compressing very well...
Using a part of the texture sometimes looks totally off (at least that is my impression). That's why I went for the full texture of one of the sides as particle sprite. As alternative, I like the flat color solution because it matches the color distribution of the block texture itself. Tweaking these bits is quite easy, though. 

In mid- to long-term I'd like to give blocks some way to override the default particle effect (and dig sound). Maybe that would also allow to change the default behavior using a module or through a config entry (in the settings menu).

MaxBorsch:
@skaldarnar I agree with @emanuele3d that the particles need to be chunks of the block side textures. Instead of changing it to use a single pixel to avoid the few blocks in which it doesn't look good, it would make more sense to find a solution to make those few blocks look good. Maybe even create a component like `BlockBreakParticleDataComponent` to attach to special blocks that specifies what texture/chunks of the texture to use to make its particles look good, otherwise if no such component is attached use the default way.

skaldarnar:
Alright, the damage effect now uses a fraction of the block texture (0.25), particles are still randomly sized. I've attached two GIFs that show the effect. I did not look into setting up a `BlockBreakParticleDataComponent`, I would leave this for another issue.

@MaxBorsch @emanuele3d is this acceptable for you or should I change anything else?

out3

out2

GooeyHub:

MaxBorsch:
@skaldarnar It's a bit hard to see from the gifs, but the pixels in each of the fragments should be a uniform size & W/H ratio. So every particle should be a square and draw a square sample from the block texture. So if you have a 16x16 block texture, the particles should all be 2x2 / 4x4 / 6x6 or 8x8 regions sampled from the block texture. **Summarized: The particles themselves can be randomly sized, but they need to be square and sample a square area of the texture so that there is no "stretching".** If that's how it is then I have no problems with it.

emanuele3d:
Should every particle be square or have random sides? I don't mind either ways at this stage, but in the future I'd like for module developers to have the chance to scale the dimensions of the particles as they like. So, at this stage it would be ok for me to leave non-square particles as this PR produces.

However, if it was up to me I'd make sure that each particle has sides dimensioned to be multiples of the block's texels, so that the texels on the particles and the texels on the block are matching in size, for visual consistency. Indeed this also means that the texels on the particles should not be stretched or squashed, although from the GIFs it's unclear if this is happening.

MaxBorsch:
Manu's second paragraph is what I was mostly talking about. Having the particles be square is optional, but I personally think it would look more consistent if all the particles were square since the whole theme of this game is cubes, so artistically those uniform cubes should smash into perfect squares.

skaldarnar:
At this point I don't care anymore about the details. If anyone of you wants to grab it up, feel free to do so. I don't know if there is any easy way to limit the size of particles to squares,`minScale` and `maxScale` only allow for the two independent random side length. I can fix it at some value again, but I'd rather like @MaxBorsch to take it over because he knows the details of this way better.

As far as I can tell the pixels itself were not stretched on the sprites, only an uneven-sized rectangle may be selected.

Cervator:
Comments are fair - we could keep going a while with this, but current state is good enough, solid progress, anybody willing can tweak further later :-)

I've tested it out and all seems well, thanks all especially @skaldarnar!

skaldarnar:
I apologize if my comment sounded a bit harsh.

@MaxBorsch maybe [ScaleRangeGeneratorComponent](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/particles/components/generators/ScaleRangeGeneratorComponent.java) should have a boolean flag `preserveRatio` so that it only creates random sizes with the same side ration within the specified bounds. I'll create some follow up issues to gradually improve this further.

PR #3005, opened 2017-07-13
rollback the changes that were done to enable world collisions

0shine0:
turns out that those changes triggered multiple CollideEvent for same entity and for same collisions. Found another way to achieve what i needed, so please do merge this to rollback the changes.

PR #3008, opened 2017-07-16
prevents characters from colliding with sensors

0shine0:
Because the characters in the game use convex sweep on every input to check for collisions. The consuming of CollideEvent was useless in this case. Thats why the required changes have been introduced where a character ignores the collisions with a trigger that has Sensor collision group to avoid responding to invisible sensors.
Please note that if you require a Trigger for an entity that is not a Sensor and wants to collide with Characters than you are free to do so but do change the collisionGroup to something relevant in the TriggerComponent. The default collisionGroup is Sensor, so be sure to change it in that case.

### DIFF
--- a/engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
+++ b/engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
@@ -933,6 +933,7 @@ public class BulletPhysics implements PhysicsEngine {
             BulletSweepCallback callback = new BulletSweepCallback(collider, new org.terasology.math.geom.Vector3f(0, 1, 0), slopeFactor);
             callback.collisionFilterGroup = collider.getBroadphaseHandle().collisionFilterGroup;
             callback.collisionFilterMask = collider.getBroadphaseHandle().collisionFilterMask;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
@0shine0 I went ahead and merged #3005 but this one I take it has the potential to break modules (in gameplay, not compilation) if they rely on functionality as-is - can you give me some clues as what to look? A nice simple search phrase or two that should find everything?

Lots of physics going on all over like @nihal111's traps, @pollend's most recent tweaks to Rails, etc

0shine0:
@cervator the colliders that act as sensors are only used to sense entities. The other entities should not react to them. The sensors are differentiated from normal entities by `StandardCollisionGroup.SENSOR` collision group. The characters collide with sensors as if they are a physical barrier. This tweak resolves that. So that characters collide with everything they intend to but do not collide with entities that have their collision group as sensors. Though rigid body entities still collide with sensors as if they are a physical barrier.

portokaliu:
I've tried this out on rails + this change and everything seemed to work fine. Except of course for the fact that off the track carts were just sliding into the sunset because of lack of friction, everything was dandy

Cervator:
Thanks for the extra details and testing :-)

PR #3013, opened 2017-07-20
Added EditScreen as parameter to InteractionScreenComponent

nihal111:
Adds an `editScreen` parameter to the InteractionScreenComponent, which is *false* by default. Setting it to `true` would disallow opening of the screen, when the editMode (in CharacterComponent) is set to false.

editMode can be toggled by the new command from console `toggleEditMode`

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java
@@ -36,6 +36,10 @@ public final class CharacterComponent implements Component {
      */
     public float interactionRange = 5f;
     /**
+     * Specifies whether the character can interact with entities that have an edit screen.
+     */
+    public boolean editScreenInteraction = false;
+    /**
      * The current interaction target of a character which has been authorized by the authority (e.g. the server).
      * <br><br>
      * Modules should not modify this field directly.
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java
@@ -38,7 +38,7 @@ public final class CharacterComponent implements Component {
     /**
      * Specifies whether the character can interact with entities that have an edit screen.
      */
-    public boolean editScreenInteraction = false;
+    public boolean editMode = false;
     /**
      * The current interaction target of a character which has been authorized by the authority (e.g. the server).
      * <br><br>
--- a/engine/src/main/java/org/terasology/logic/characters/interactions/EditScreenComponent.java
+++ b/engine/src/main/java/org/terasology/logic/characters/interactions/EditScreenComponent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 MovingBlocks
+ * Copyright 2017 MovingBlocks
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java
```diff
@@ -36,6 +36,10 @@
      */
     public float interactionRange = 5f;
     /**
+     * Specifies whether the character can interact with entities that have an edit screen.
+     */
+    public boolean editMode = false;
```
Cervator: This name seems somewhat ambiguous, and the PR description confuses me a bit - when it is *true* disallows editing? :-)

Did the variable at first get added to `InteractionScreenComponent` then later moved into `CharacterComponent` ?

How about `canEditInteractionScreens` ? `false` means no, change it to `true` to allow the character to edit interaction screens?

engine/src/main/java/org/terasology/logic/characters/interactions/EditScreenComponent.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.terasology.entitySystem.Component;
+import org.terasology.module.sandbox.API;
+
+/**
+ * Entities with this component will show a different UI Screen during interactions when edit mode is on. <br/>
+ *
+ */
+@API
+public class EditScreenComponent implements Component {
```
Cervator: Rather than the new variable on `CharacterComponent` should having this component attached instead be the trigger for whether something can edit *at all* ? If no component then nope no edit for you. If there is a component variables here can customize it further if needed

engine/src/main/java/org/terasology/logic/characters/interactions/InteractionScreenComponent.java
```diff
@@ -25,5 +25,4 @@
 @API
 public class InteractionScreenComponent implements Component {
     public String screen;
```
Cervator: Guessing this got reverted, if you make further changes might just want to wipe out changes to this class at all :-)

engine/src/main/java/org/terasology/logic/characters/interactions/InteractionSystem.java
```diff
@@ -61,6 +62,36 @@ public void onActivate(ActivateEvent event, EntityRef target) {
 
     }
 
+    /**
+     * To consume the activation event in the edit mode if the target entity has an EditScreenComponent with no screen defined.
+     * @param event
```
Cervator: Maybe some javadoc for what's expected in the `@param`s beyond their name? Like a typical use case.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -520,6 +521,16 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
         return "Spawned block.";
     }
 
+    @Command(shortDescription = "Toggle edit mode", helpText = "Toggles the edit mode to allow interaction with edit screens",
+            runOnServer = true, requiredPermission = PermissionManager.CHEAT_PERMISSION)
+    public String toggleEditMode(@Sender EntityRef sender) {
+        EntityRef character = sender.getComponent(ClientComponent.class).character;
+        CharacterComponent characterComponent = character.getComponent(CharacterComponent.class);
+        characterComponent.editMode ^= true;
```
Cervator: `^=` - neat!

nihal111:
Sorry for wasting your time on this review. This already got implemented in a different way.

nihal111:
This was solved by an alternative approach in https://github.com/MovingBlocks/Terasology/pull/3020 andhttps://github.com/MovingBlocks/Terasology/pull/3020

PR #3019, opened 2017-07-24
add support for module jars passed via -classpath

kaen:
This is a fix for a fairly esoteric problem, but I'll do my best to explain:

The Problem
===

Specifying module `.jar` files in the classpath causes `ModuleEnvironment.getModuleProviding` to fail [while registering components](https://github.com/MovingBlocks/Terasology/blob/8d3173cad52fd0305f39013ee20cdea4ff19a42d/engine/src/main/java/org/terasology/engine/bootstrap/EntitySystemSetupUtil.java#L134) for any classes loaded from that jar. This occurs when trying to use the `ModuleManagerImpl` in a module's test environment within IDEA for the purpose of running integration tests using full `TerasologyEngine` instances.

In the module configurations, module dependencies are parsed by the gradle scripts and listed as compile time dependencies in the generated IDEA project. This means that they are passed in to the executable via `-classpath` when tests are run. When gradle downloads module jars they are stored in some place like `$HOME/.gradle/caches/`. However, the jar is copied to `modules` by some extra gradle logic, and registered with classpaths that live in `$projectRoot/modules` by the `ModulePathScanner`. This causes the equality check in `getModuleProviding` to fail because the [left side does not equal the right side here](https://github.com/MovingBlocks/gestalt/blob/bca6b16808c7bb77f69cec986529aa644a6c2eed/gestalt-module/src/main/java/org/terasology/module/ModuleEnvironment.java#L291). The code location as seen from the class itself will be in `$HOME/.gradle/caches/` while the module's classpaths will all be in `$projectRoot/modules`.

Semantically, this means the `ModuleEnvironment` can't find a module containing the class's self-reported code location, because it's been loaded from a jar in the gradle cache as directed by `-classpath` and not from the `modules` subdir.

The Solution
===

This PR looks for Terasology modules provided on the classpath. Any class path URL pointing to a jar containing a valid `module.txt` is assumed to be a Terasology module. When found, it parses the URL from the system's default `ClassLoader`, which is assumed to be a `URLClassLoader`. The jar path extracted from the URL is then loaded by way of a standard `ModuleLoader`, passing in the extracted file path. This creates an `ArchiveModule` entry in the registry that *does* match the reported source location of the loaded class, and allows `getModuleProviding` to work as expected.

Extra Thoughts
===

- It might be possible to solve this by use the `$projectRoot/modules` file path when specifying dependencies in gradle. However, I couldn't get this to work without breaking automatic module JAR downloads.
- In theory `gestalt-modules` could contain this logic, but semantically it feels more correct that Terasology supports this feature than gestalt-module,  because it's caused by an idiosyncrasy of the Terasology build system. 
- This may fix some heisenbugs where tests with module resolution sporadically fail depending on the gradle cache's state


GooeyHub:

GooeyHub:

GooeyHub:

OvermindDL1:
> Any class path URL pointing to a jar containing a valid module.txt is assumed to be a Terasology module. 

I would suggest parsing and confirming that the file really is a Terasology module.  I've seen such a file in jar's elsewhere and something could be linked in as a library that has one.

kaen:
> I would suggest parsing and confirming that the file really is a Terasology module
I had a similar concern, but I'm not sure how to check for that beyond parsing the `module.txt` as JSON (which is currently implemented). Is there a canonical way to test that a jar is a Terasology module?

OvermindDL1:
> I had a similar concern, but I'm not sure how to check for that beyond parsing the module.txt as JSON (which is currently implemented). Is there a canonical way to test that a jar is a Terasology module?

If not then maybe a special option *has* to be specified in the JSON to delineate it as such?

kaen:
Updated to check for both an id and display name. If a JAR has a `module.txt` in the root that is not only valid JSON but contains the keys `id` and `displayName` and can be processed successfully by `ModuleLoader.load`, then it is assumed to be a Terasology module :)

I should also note that even if a JAR were to haphazardly be loaded in this way without actually intending to be a Terasology module, it should still be harmless unless the module is manually activated by the user or its ID and version cause it to override a real module in `modules/` subdirectory, which would be a fairly astonishing chain of coincidences.

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Tested out along with https://github.com/kaen/ModuleTestingEnvironment - seems to do the trick! Neato topic, glad to see progress. Wonder if we can do something similar to fix up how test execution starts out of the Gradle cache (the engine jar) in module build jobs in Jenkins.

PR #3022, opened 2017-07-26
[v2.0.0] Add SectorRegionComponent to allow sector entities to span multiple chunks

Vizaxo:
This PR implements the `SectorRegionComponent`, which allows sector entities to watch a set of chunks, rather than just one chunk. This allows the entity to span a region, e.g. a city, or a big herd of deer. If any of the chunks in the region are loaded, the `LoadedSectorUpdateEvent` will be sent to the entity.

It also adds some helper methods in the `SectorUtil` class for dealing with the `SectorRegionComponent`.

There are also a few other small fixes/improvements, and the addition of the `blockInChunk()` method to `ChunkMath`, which I've used in my changes to DynamicCities.

I've updated the [TutorialSectors](https://github.com/terasology/tutorialsectors) module to demonstrate its use. The `postBegin()` method of the `TestSystem` adds the SectorRegionComponent. It can be tested in-game by going to position `0, 0, 0` and verifying that the log shows `Entity 47 loaded. Value is ...`, then moving away so that the chunk unloads and the loaded message stops appearing. Then teleporting to `1000, 0, 1000` should cause the loaded messages to return.

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityRef.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityRef.java
@@ -106,7 +106,7 @@ public abstract class EntityRef implements MutableComponentContainer {
      *
      * @param maxDelta the maxDelta for the sector-scope entity
      */
-    public void setSectorScope(float maxDelta) {
+    public void setSectorScope(long maxDelta) {
     }
 
     /**
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityRef.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityRef.java
@@ -106,7 +106,7 @@ public abstract class EntityRef implements MutableComponentContainer {
      *
      * @param maxDelta the maxDelta for the sector-scope entity
      */
-    public void setSectorScope(long maxDelta) {
+    public void setSectorScope(float maxDelta) {
     }
 
     /**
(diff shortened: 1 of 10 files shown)


GooeyHub:

engine/src/main/java/org/terasology/entitySystem/sectors/SectorSimulationSystem.java
```diff
@@ -45,15 +45,17 @@
  * {@link EntityManager#createSectorEntity(long)} or its scope was set by
  * {@link BaseEntityRef#setScope(EntityScope)}.
  *
+ * @see SectorUtil#getWatchedChunks(EntityRef) for the definition of wathed chunks.
+ *
  * It periodically sends a {@link SectorSimulationEvent} to all sector-scope entities, which should trigger any
- * simulation that needs to happen regardless of whether or not the entity's chunk is loaded.
+ * simulation that needs to happen regardless of the status of the enitty's watched chunks.
```
skaldarnar: ~enitty~ entity

GooeyHub:

engine/src/main/java/org/terasology/entitySystem/entity/EntityRef.java
```diff
@@ -106,7 +106,7 @@ public void setScope(EntityScope scope) {
      *
      * @param maxDelta the maxDelta for the sector-scope entity
      */
-    public void setSectorScope(long maxDelta) {
+    public void setSectorScope(float maxDelta) {
```
Cervator: Reminder to change this back to `long` as discussed on Slack :-)

GooeyHub:

skaldarnar:
I accidentally merged #3037 first, but I guess this is okay ... 🙄

Issue #3021, opened 2017-07-26
Can't start headless server due to UnsatisfiedLinkError exception in telemetry subsystem


gianluca-nitti:
### What you were trying to do
Running an headless server, from the current version of the `develop` branch (c7d50f045620d5f10f90f704250571d01f486eec). I tried both `gradlew server` and the *Terasology PC (Headless)* IntelliJ run configuration.

### What actually happened
During the server initialization I get this exception:
```
java.lang.UnsatisfiedLinkError: no lwjgl64 in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1864)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
	at org.lwjgl.Sys$1.run(Sys.java:72)
	at java.security.AccessController.doPrivileged(Native Method)
	at org.lwjgl.Sys.doLoadLibrary(Sys.java:66)
	at org.lwjgl.Sys.loadLibrary(Sys.java:87)
	at org.lwjgl.Sys.<clinit>(Sys.java:117)
	at org.lwjgl.opengl.GLContext.<clinit>(GLContext.java:112)
	at org.lwjgl.opengl.GL11.glGetString(GL11.java:1792)
	at org.terasology.telemetry.metrics.SystemContextMetric.<init>(SystemContextMetric.java:82)
	at org.terasology.telemetry.Metrics.initialise(Metrics.java:46)
	at org.terasology.engine.subsystem.common.TelemetrySubSystem.postInitialise(TelemetrySubSystem.java:68)
	at org.terasology.engine.TerasologyEngine.postInitSubsystems(TerasologyEngine.java:267)
	at org.terasology.engine.TerasologyEngine.initialize(TerasologyEngine.java:199)
	at org.terasology.engine.TerasologyEngine.run(TerasologyEngine.java:368)
	at org.terasology.engine.Terasology.main(Terasology.java:142)
```

The game client with graphics (`gradlew game` or the *Terasology PC* IntelliJ run configuration) work correctly.

### How to reproduce
Just run `gradlew server`, or alternatively in IntelliJ select the *Terasology PC (Headless)* run configuration and click run. After a while the exception is thrown and the initialization hangs (no game is loaded).

### Log details and game version
Full output here: https://pastebin.com/RKxNEdVn
[...]

PR #3024, opened 2017-07-27
Checking isHeadless from DisplayDevice

nihal111:
Quick Fix for #3021 

Fetches the context from `CoreRegistry.get()` (since `@In` wasn't working, maybe because it is not a registered system). Finds the `DisplayDevice` from the context and uses the built in method `isHeadless` to check if it is headless.

### DIFF
--- a/engine/src/main/java/org/terasology/telemetry/metrics/SystemContextMetric.java
+++ b/engine/src/main/java/org/terasology/telemetry/metrics/SystemContextMetric.java
@@ -18,6 +18,9 @@ package org.terasology.telemetry.metrics;
 import com.snowplowanalytics.snowplow.tracker.events.Unstructured;
 import com.snowplowanalytics.snowplow.tracker.payload.SelfDescribingJson;
 import org.lwjgl.opengl.GL11;
+import org.terasology.context.Context;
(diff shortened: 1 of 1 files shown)


GooeyHub:

GooeyHub:

Cervator:
Looks good to me, especially with the super critical optimization added in :D

@GabrielXia: general question - how do you think something like this would impact aggregated data? Imagine if we're trying to make a pie chart for video cards. For normal client runs you get a valid video card. For a headless run you get *nothing* now. And could a user selectively exclude stuff at some point? Probably there are better examples than video card for that.

Would we have something like:

* Total number of systems that have submitted metrics
* Number of systems with video card x (and y, z, etc)
* Number of systems that excluded this metric
* Number of systems that ran headless

That would make a chart pretty easily. But would the lack of a metric get in the way, or would setting *a different* metric be better? Or instead of skipping them like this PR does should we just set those keys to an artificial value to indicate headless?

kaen:
Tested and working locally, thanks and LGTM :+1:

Cervator:
Merged - thanks! :-)

PR #3028, opened 2017-07-30
Refactor module installation code to decouple UI (SelectModulesScreen) from logic

gianluca-nitti:
### Contains

The title pretty much summarizes the content, the code to determine the required dependencies according to the [meta server](http://meta.terasology.org/modules/show) and download a module has been extracted from the `SelectModulesScreen` class. I did this for two reasons:
- As part of my GSoC project I'm implementing a feature in [FacadeServer](https://github.com/MovingBlocks/FacadeServer) to allow server administrators to manage multiple games on a server (start/stop the saved ones and create new games), and having the ability to install missing modules from the web interface is useful when you want to create a new game with a module which is currently not present on the server. As agreed with my mentor @msteiger redoing the dependency resolution and download code in FacadeServer would add useless duplication, so I extracted the existing code from the only user interface which uses it at the moment.
- `SelectModulesScreen` was becoming a huge class with multiple not-so-related inner classes (like MultiFileDownloader for example - it could later become useful for other purposes too).

### How to test

Using the module selection screen (which you can reach from Single Player > Create > Modules... ) should work as before, including installing new modules or updating existing ones. Some additional information is provided during the download (total bytes, downloaded bytes, number of modules, etc.); a small downside, however, is that when dependency resolution fails a generic not-very-detaild error message is shown. The reason is that the dependency resolution code has been refactored to use the already existing DependencyResolver class (with a ModuleRegistry based on the module list obtainde from the meta server), and providing additional information is an open issue (MovingBlocks/gestalt#15).

I have some doubts about the decision of which modules must be downloaded. I simply used the same strategy present in the old code, which is redownloading a dependency only if the online version is newer ([link to the old code](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/SelectModulesScreen.java#L485) - [link to the new code](https://github.com/MovingBlocks/Terasology/compare/develop...gianluca-nitti:refactor/module-downloader#diff-b0f5937519800e1a39bcbb713998c424R94)).
- Module dependencies have a range of compatible versions
- The Dependencyresolver returns a set of exact module versions
- This is then filtered to download only the versions newer than the installed ones

Is this correct? I did some manual tests and everything was fine, but I'm not 100% sure this is correct in any case of dependencies combinations, so I'd suggest someone with a deeper knowledge of the module dependency system reviews this before it's merged.


GooeyHub:

GooeyHub:

GooeyHub:

gianluca-nitti:
After opening the PR I realized I still had some improvements to add, so I did them and also added some unit test. Now it's ready for review.

Also, in the previous comment I mentioned having some doubts about version management, so I'll try to explain my concern here. Both with the "old" (`develop` branch) code and with the "new" one (this branch), before installing a module or a set of modules, a dependency resolution algorithm is ran to produce a new list of modules comprehensive of all the dependencies of the required ones. Then, for each of the modules of this list, if the module is already installed it is re-downloaded only if the online version is newer than the local one (or both are snapshots of the same version number). However, each module dependency has a version range (See [DependencyInfo](https://github.com/MovingBlocks/gestalt/blob/a5def332a6e717612d5556520da3c308a032006c/gestalt-module/src/main/java/org/terasology/module/dependencyresolution/DependencyInfo.java#L34) class in Gestalt), so both lower and upper bounds. Consider this situation:
- Locally, we have module A at version 1.0.0, and B at version 1.0.0, which depends on A with lower bound 0.1.0 and upper bound 2.0.0. This is coherent because 0.1.0 < 1.0.0 < 2.0.0.
- We install C (it's version doesn't matter), which depends on A with lower bound 3.0.0. The version 3.0.0 of A is available online, so dependency resolution for C succeeds and A 3.0.0 is downloaded together with C.
- Now we should have two versions of A, 1.0.0 and 3.0.0.

First question is: is this OK? Can multiple versions of the same module coexist? From a ModuleRegistry and filesystem point of view I think so since they are saved and loaded with their version number. However I'm not sure if you could load both in the same game and how you could select the version in the SelectModulesScreen.

Second question: isn't it possible that a lower version needs to be downloaded? Suppose that in the previous example, C requires A between 0.1.0 and 0.2.0 (but 1.0.0 is installed) - in my understanding with current code (both this version and the pre-this-PR one), A 0.2.0 wouldn't be downloaded because 0.2.0 < 1.0.0, but then C would be forced to use an unsupported dependency version. A possible solution could be making the DependencyResolver (in Gestalt) return a supported version range for each module rather that the latest version in the registry. This way, the logic could become "download if the installed version is not in the range".

The "install only if newer method" worked before this PR, and here the same method is used; the difference is that the list of dependencies which is filtered is produced by a DependencyResolver and not by a separate "custom" method, which shouldn't be any difference (but reduces duplication). I'm just expressing my concerns about the correctness this strategy itself, maybe it always worked because the considered case is very rare. Sorry for wall of text and if this isn't 100% related to this pull request.

Cervator:
Ahh welcome to the wondrous world of [dependency hell](https://en.wikipedia.org/wiki/Dependency_hell) which is a frequently encountered issue in software development :-)

**In short:** we should only allow a single version of a dependency loaded. Imagine a user on the module select screen in your example. Enabling `A` is fine since it and `B` are happy together. Everything is green, dependencies get enabled and downloaded if needed.

If the user then enables `C` and the dependency resolution now realizes no single combination of dependencies exist that'll support everything it should outright fail and provide the user an error that those modules together result in an unsatisfied dependency tree. Much like how at present if the game is shipped with a bad set of modules any one module that is missing a dependency is highlighted in red. 

**In long:** *Can* you get around it? Yes. There are Java packaging schemes that cleverly repackage a dependency plus the stuff that depend on it into unique packages to avoid clashes. You can also force the issue as a user and say for sure "I'm fine with everything using `A` v3 - just do it and launch the game so I can try it"

However, both those paths are fraught with danger. There is a pretty big write-up in the forum at http://forum.terasology.org/threads/more-version-increments-to-avoid-dependency-issues.1270

My personal goal is to use our tools and infrastructure to minimize the barrier to updating modules to the latest at all times. That forum thread gets into it some - I want a ton of automated testing that seeks out modules that are used together and finds where one breaks another, then highlight it right away so authors (or module adopters / the general community) can spot an upcoming change and adjust for it. There shouldn't really be a reason for why anybody would want `A` way back at v1 in your example if we're really on top of the process, and we should end up with fewer major releases so they'll stay naturally compatible longer.

One remaining challenge is when a updated dependency `x` qualifies because it hasn't gone to a new major version (so the default range accepts it at newer minor release levels for use by `y`) *but* behind the scenes it updated one of *its* dependencies `z` to a new major, after validating that all is well. That's fine since if we follow SemVer correctly the external API of `x` remains unchanged, but a completely different module elsewhere in the tree could still have `z` required at an older major version (like in your A-B-C example). Everything works functionality but now you again have two different major versions in the same tree so stuff could explode. More things we can probably detect via automation.

Another old related forum thread that might be of interest: http://forum.terasology.org/threads/module-tracker-infrastructure.1075/

gianluca-nitti:
Okay, thank you for the very in-depth explanation.

engine/src/main/java/org/terasology/engine/module/ModuleManagerImpl.java
```diff
@@ -52,8 +53,9 @@
     private ModuleRegistry registry;
     private ModuleEnvironment environment;
     private ModuleMetadataJsonAdapter metadataReader;
+    private ModuleInstallManager installManager;
 
-    public ModuleManagerImpl() {
+    public ModuleManagerImpl(String masterServerAddress) {
```
msteiger: I would probably call this `metaServer..`, but this is really a minor issue.

engine/src/main/java/org/terasology/engine/module/ModuleManagerImpl.java
```diff
@@ -83,6 +85,11 @@ public ModuleManagerImpl() {
 
         setupSandbox();
         loadEnvironment(Sets.newHashSet(engineModule), true);
+        installManager = new ModuleInstallManager(this, masterServerAddress);
+    }
+
+    public ModuleManagerImpl(Config config) {
+        this(config.getNetwork().getMasterServer());
```
msteiger: Ah. That's why. I guess this should be changed, but might affect existing config JSON docs, right?

gianluca-nitti: In my understanding it could be changed without breaking config compatibility by replacing
`public String getMasterServer() { return masterServer; }`
with
`public String getMetaServer() { return masterServer; }`
 in `NetworkConfig` class, i.e. renaming the getter but not the field. Not sure if it's good practice though.

Or alternatively we could keep both fields (`masterServer` and `metaServer`), then in the getter returning `metaServer` by default but falling back to `masterServer` if null, and copying the value from the old field to the new one when the config is saved. This way we could keep compatibility with old config files for a certain amount of time, then one day in the future "deprecating" the old field name.

GooeyHub:

Cervator:
Did a little testing in advance of merging. One observation: you can update a module that's already at the latest version - is that on purpose? Not actually sure if it already worked like that or not.

In any case - merged! :-)

I'm not too worried about keeping JSON config compatibility front and center just yet (Alpha and all that), although when possible it is nice to have. I didn't hit any errors testing it locally though, was able to download a module just fine :-)

gianluca-nitti:
Thanks for merging.

> One observation: you can update a module that's already at the latest version - is that on purpose? Not actually sure if it already worked like that or not.

Yes, if I recall correctly you could already do that before this PR (though I'm not 100% sure) and it should be because they are snapshots. A snapshot is always considered to update (when you install a module which has a dependency on another module which you already have in a snapshot version, it's redownloaded); I got this information from this comment in the code:
> Multiple binaries get released as the same snapshot version, A version name match thus does not
> guarantee that we have the newest version already if it is a snapshot version.
> 
> Having the user redownload the same binary again is not ideal, but it is better then having the user
> being stuck on an outdated snapshot binary.

It was in the [`isOnlineVersionNewer()` method ](https://github.com/MovingBlocks/Terasology/pull/3028/files#diff-059ce3ab46f9e05f0a9e91e188d69952L762), previously in the `ModuleSelectionInfo` class (inner class of `SelectModulesScreen`) and now is in [`ModuleDownloadListGenerator`](https://github.com/MovingBlocks/Terasology/pull/3028/files#diff-21953cbd65e582ddc7e5eccac1fd3f49R58).

PR #3033, opened 2017-08-01
Dead players alive-d on join

nihal111:
If the last save happens when a player is dead (on the death screen) then the player entity is saved in the player store without the AliveCharacterComponent.

When a client rejoins, the player is spawned but without the AliveCharacterComponent. This PR ensures that all players that join a game have a AliveCharacterComponent.


GooeyHub:

PR #3037, opened 2017-08-01
[v2.0.0] Convert alwaysRelevant to use entity scope instead

Vizaxo:
### Contains

This removes the `alwaysRelevant` entity field, and makes the relevant methods look at the `alwaysRelevant` field on the entity's scope instead.

`GLOBAL` and `SECTOR` scope are `alwaysRelevant`, but `CHUNK` scope is not. The default scope is `CHUNK`, so the default for `alwaysRelevant` is still `false.`.

This also updates some entity-related tests, and fixes some problems found by them (now all of the entity creation code goes through the `EntityBuilder` (previously almost all of it did, but methods like `createEntityWithId()` didn't), so that the scope can be set properly without putting it in multiple places).

### How to test

Load the [TutorialSectors](https://github.com/Terasology/TutorialSectors/tree/remove-alwaysrelevant) module on the `remove-alwaysrelevant` branch, and observe that the entity doesn't need to be set to `alwaysRelevant`, because it has already been marked as sector-scope (it will still simulate when you have unloaded all of its chunks).

Note: this depends on #3022

Some internal methods on the `PojoEntityManager` had to be exposed for use by the `EntityBuilder`, so it might be worth having a look at how the classes are structured (e.g. extracting the `EntityBuilder` to be an interface, and having a `PojoEntityBuilder` to make the Pojo-specific entities. Or putting the `EntityBuilder`'s `bulid()` method back into the `EntityManager`, now that it handles all of the different scenarios (scope, id, etc.) in one method.).

(Also, while on the topic of the `EntityBuilder`, it's probably worth looking at whether the `EntityBuilder` and `EntityStore` can be combined. They both do the same thing, they just differ in that the `EntityBuilder` requires an `EntityManager` when it is created, but the `EntityStore` involves the `EntityManager` when the entities are being built.)

There's also a possible issue that the `OnAddedComponent` is still sent when the entity is re-created from storage. With the previous implementation, this wasn't sent when re-created with the `createEntityWithId()` method, but if `ignoringEntityId` is true in the `EntitySerializer`, it would still send that event (but I don't know if that is ever actually used).

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityBuilder.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityBuilder.java
@@ -21,11 +21,16 @@ import org.slf4j.LoggerFactory;
 import org.terasology.entitySystem.Component;
 import org.terasology.entitySystem.MutableComponentContainer;
 import org.terasology.entitySystem.entity.internal.EngineEntityManager;
+import org.terasology.entitySystem.entity.internal.EngineEntityPool;
 import org.terasology.entitySystem.entity.internal.EntityInfoComponent;
 import org.terasology.entitySystem.entity.internal.EntityScope;
+import org.terasology.entitySystem.entity.lifecycleEvents.OnActivatedComponent;
+import org.terasology.entitySystem.entity.lifecycleEvents.OnAddedComponent;
 import org.terasology.entitySystem.prefab.Prefab;
 
+import java.util.Collections;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * An entity builder provides the ability to set up an entity before creating it. This prevents events being sent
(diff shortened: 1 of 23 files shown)


GooeyHub:

engine/src/main/java/org/terasology/entitySystem/sectors/SectorSimulationComponent.java
```diff
@@ -30,15 +30,17 @@
 @API
 public class SectorSimulationComponent implements Component {
 
-    public static final float MAX_DELTA_DEFAULT = 10;
+    public static final long MAX_DELTA_DEFAULT = 10;
```
skaldarnar: Is the value of 10ms still correct? Or should it be (much) more per default?

engine/src/main/java/org/terasology/entitySystem/entity/EntityBuilder.java
```diff
@@ -159,11 +187,11 @@ public void setAlwaysRelevant(boolean alwaysRelevant) {
     }
 
     public void setScope(EntityScope scope) {
-        this.scope = scope;
+        this.scope = Optional.of(scope);
     }
 
     public EntityScope getScope() {
-        return scope;
+        return scope.orElse(null);
```
skaldarnar: Since you hold the scope internally as `Optional`, why not expose it? IMHO the internal interfaces looks nicer than the external one here...

engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityPool.java
```diff
@@ -81,23 +78,10 @@ public EntityRef create(Iterable<Component> components) {
...
-            }
-        }
-
-        //Retrieve the components again in case they were modified by the previous events
-        for (Component component : entityManager.iterateComponents(entity.getId())) {
-            entityManager.notifyComponentAdded(entity, component.getClass());
-        }
-
-        return entity;
+        EntityBuilder builder = newBuilder();
```
skaldarnar: Can't these calls on the builder be chained (maybe this is not so common in Java, I might got confused with Scala here)?

Vizaxo: I wanted to do that, as it seems much cleaner (and is how I naturally start to write it every time), but I wasn't sure if it was idiomatic Java or not (some quick searching suggests that it is idiomatic, especially for builders).

engine/src/main/java/org/terasology/entitySystem/sectors/SectorSimulationComponent.java
Vizaxo: It should be more by default; I changed it in #3041, but forgot to here.

PR #3041, opened 2017-08-02
[v2.0.0] Allow sector entities to simulate at a different rate when loaded

Vizaxo:
### Contains

This splits SectorEntityComponent's maxDelta field into unloadedMaxDelta and loadedMaxDelta, which specify the maximum simulation delta for when all of the entity's watched chunks are unloaded, or when at least one is loaded, respectively.

This the simulation rate to be adjusted depending on whether or not the entity is loaded, so the unloaded simulation rate can be set as low as possible to maximise performance, while the loaded simulation rate can be higher to allow the player to observe the changes more frequently.

### How to test

Load the [TutorialSectors](https://github.com/Terasology/TutorialSectors/tree/v2-branches/loaded-max-delta) module on the `v2-branches/loaded-max-delta` branch, and observe that the the entity simulates every second while the player is standing at (0, 0) (when a watched chunk is loaded), but simulates every 10 seconds when the player is at (-1000, -1000) (because none of the watched chunks are loaded).

Note: this depends on #3022

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/entity/EntityManager.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/EntityManager.java
@@ -30,11 +30,23 @@ public interface EntityManager extends EntityPool {
     /**
      * Creates a new EntityRef in sector-scope
      *
-     * @param maxDelta the maximum delta for the sector entity's simulations
-     *                 @see SectorSimulationComponent#maxDelta
+     * @param maxDelta the maximum delta for the sector entity's simulations (both unloaded and loaded)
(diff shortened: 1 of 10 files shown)


GooeyHub:

GooeyHub:

Vizaxo:
The merge conflicts were caused by the "Vizaxo committed with skaldarnar" commits from #3037, which git interprets as brand new commits so it can't easily build on top of them.

I've rebased this one (it automatically rebased fine with no conflict) and tested it, and it seems to work fine now with no conflicts.

PR #3042, opened 2017-08-02
Implemented libgdx native bullet wrapper

pollend:
# Contains

This implements libgdx's native bullet wrapper and removed Terra bullet along with all references to javax math and instead I've replaced these instances with libgdx's math library. For the most part of lot of the logic is unchanged and I've just rewritten the internal features of each function. A lot of the function endpoints for bullet have been slightly changed, but I'm still not sure how this will behave for the scope of the entire project. 

In a separate pull request I should provide the modified libgdx library. An Artifactory build of this will need to be made along with updating gradle. I've statically listed this out in gradle which is not ideal but should be usable for the time being.

# How to Test

You will have to include the modified libgdx library into lib and install the maven libraries from there (mvn install). Memory usage will have to be watched since I'm not sure that I've cleaned up all the resources. The behavior of this in regards to other mods will also have to be tested.


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

oniatus:
What is the current state of this one? Is there a way we can have this change on top of the refactoring in #3199 so that the framework change is less intrusive in the project?

pollend:
not sure what the plans are for this. @eviltak  is recommending https://github.com/tzaeschke/ode4j over libgdx. Might be easier to implement since compiling libgdx appears to be pretty troublesome. This branch depends on a patch to native portion of libgdx. I think @Cervator could comment more on the compiling aspect of libgdx.

Cervator:
Just merged #3199 and this isn't even conflicting with it, woo! :-)

I hope to poke around with the related build system stuff in the near future, although it isn't quite next on my list yet. With #3199 in we might actually be able to treat the different approaches like separate physics implementations, for easy comparisons and benchmarking? Could try both LibGDX physics as well as ode4j

pollend:
#3432 is just an improvement on this work. we can close this and just make a ref back to this.

PR #3049, opened 2017-08-05
Item persisting on the map after saving fix

portokaliu:
This should fix #2810 's fix

There shouldn't be a case for seeing other people's items (yet)

This fix adds `preAutoSave` and `postAutoSave` to the game so that items can be destroyed before they are saved by the system. The save method was just separated into 2 not so distinct ones.

This should prevent items from being despawned from your hand while playing as well (at least from the save PoV)

The problem was that the item held in your hand wasn't ever being destroyed, and was kept in the save as well. This created a floating mesh for the rest of eternity on the map.

The fix is that it's destroyed only during forced save and not auto.

I checked for the item remaining on the map in single player and multiplayer, and there doesn't seem to be any type of issue.

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/systems/BaseComponentSystem.java
+++ b/engine/src/main/java/org/terasology/entitySystem/systems/BaseComponentSystem.java
@@ -35,6 +35,14 @@ public abstract class BaseComponentSystem implements ComponentSystem {
     }
 
     @Override
+    public void preAutoSave() {
+    }
+
+    @Override
+    public void postAutoSave() {
+    }
+
+    @Override
     public void preSave() {
     }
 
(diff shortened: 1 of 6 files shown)


GooeyHub:

GooeyHub:

Cervator:
Tested this out - while I seemingly replicated the *old* issue (held item at 0 0 0 after sort sort of save/reload) once I couldn't do it again, and at no point did I see the held item persisted where I left a player. So maybe the one replication was a fluke / dirty workspace (happened on the first test, JoshariasSurvival in single player)

This PR is technically an API violation as it adds new required methods to the `ComponentSystem` interface. While related engine classes are tweaked there are usages in modules (like WorldPortals`) that would break. To avoid that I made the new methods `default` which also allowed me to remove the empty implementations in some implementors.

I tested out Inferno (depends on WorldPortals) while making sure my local WP hadn't rebuilt. It made it through an autosave cycle successfully so .. yay? Default method to the rescue? Hopefully I'm doing it right :-)

The actual architectural split between the two types of saves seem fair enough to me. Maybe something could still be improved further - I wonder if I triggered it by killing the game after/during an auto-save rather than exit gracefully triggering the non-auto-save that includes the held item kill?

Thanks @portokaliu!

PR #3054, opened 2017-08-08
[WIP] Zones

Vizaxo:
### Contains

This is the first implementation of zones (described in [this forum post](http://forum.terasology.org/threads/new-conceptual-layer-sector-plus-musings-on-multi-world-node.1420/#post-12708) and lower down in that thread).

### Notes

It implements zones at the world generation level, so they can be used to specify different zones for generation and preview (e.g. a surface zone, and an underground zone). In the world preview, the desired zone can be selected, and that one will be previewed. Each zone can determine its own preview image (via FacetLayers), which just uses the current preview system (so they can be created and customized in the same way).

The region selection is currently done with a function that takes the world position of the block, and the Region (so facets can be accessed), and returns a boolean of whether or not that block is in the region. This is very customizable, but I'm not sure about the performance of it (a more restrictive criteria for determining what is in the zone might allow for much greater optimization, at the chunk and the block level (e.g. being able to ignore whole chunks at a time)).

The current implementation does rasterization by saving and restoring all blocks that aren't in the given zone, which works for now but probably has lots of room for optimization.

I did get into a bit of difficulty with the current way some things are implemented with annotations. I wanted to make the zone itself the `FacetProvider` for all of the facets it contains, but because `FacetProvider`s have to be declared with annotations, I don't think that's possible at the moment (or, at least, I don't think that's how they're intended to be used, and it might be quite fragile). I got around it just fine by adding the zone's list of providers rather than the zone itself, but I'm not sure if that limits some things that the zone can do because it can't easily access the facets before they get added (for example, when zone connectors are added).

### How to test

The [PrototypeWorld](https://github.com/Vizaxo/PrototypeWorld/tree/v2/zones) `v2/zones` branch has the world generator (`ProtoGenerator.java`) set up to use zones. It has an underground zone, a surface zone, and an example zone which uses an odd region function (currently without a preview layer, so it should probably be removed from the preview zones list as it just appears black). 
In practice the example zone would be better off implemented with a facet, but this is just an example of how zones can work.

This is just a WIP, it's not ready for merging yet.


GooeyHub:

GooeyHub:

Vizaxo:
### Nested zones

I've allowed zones to be nested, so that a given inner zone will only affect areas that are also contained within its parent zone. I'm not sure what different uses this will have, but one idea is to re-implement the generation-level biomes of the `PrototypeWorld` as zones that are nested in the surface zone.

An example of this is the diagonal glass wall in the `ProtoGenerator`, which is a sub-zone of the surface zone, so it does not extend underground.

### Layered zones

I've also added layered zones, which are implemented with the regionFunction of standard zones. These will automatically calculate their positions based on the SurfaceHeightFacet and the other layered zones present, as well as each layered zone's ordering and width. I think this is more akin to the original surfaces concept, as these zones are stacked vertically on top of each other, and don't overlap at all.

Currently the layers' width is just their minWidth at all points, but it would be good to be able to have some control over varying the thickness in different places.

The highest and lowest layers will currently extend to infinity, but that might be customizable in the future by specifying a maximum width (the maximum width field doesn't do anything at the moment).

Currently the ordering is just done by specifying an integer, and higher integers are higher up in the world (with positive ones being aboveground, and negative ones being underground). This is similar to the ZOrder of the world preview system, or the priority of the events system. I'm not sure if this is the best solution though, especially in the case of two layers with the same priority (currently the ordering of them is undefined; I'm not sure what the best solution in that case is).

Layers will be a good candidate for world plugins (I haven't made them work as plugins yet) because they automatically calculate positions, so layers from different modules will just order themselves and won't affect each other. This will allow for, say, the underground generation and the surface generation to be defined in different modules, with no knowledge of each other.

I haven't added zone connectors yet, but they might only be applicable to layered zones, and they can probably be implemented by adding another zone at the overlap points of different layers.

The example module has 3 layered zones: a surface layer (which is where the SurfaceHeightFacet is set; setting the surface height in anything but the surface layer might cause weird behaviour), a 1-block-thick layer of glass right under the surface, and an underground stone layer which extends to the bottom of the world.

Vizaxo:
### More expressive layered zones

Another thing to look into is allowing you to express in more detail how you want the layered zone to be constructed. At the moment each zone is a set width, and each zone follows the contours of the surface exactly.

There should be some way of allowing the thickness of the zone to vary, and specifying how closely it follows the contours of the other layers (e.g. a zone for underground soil types would closely follow the contours of the surface, but lava pools and caves deep underground wouldn't be affected by the surface much).

There should also be options for specifying absolute depths, rather than just relative ones (as the depth could change considerably based on the number of installed modules adding their own layers). E.g. a zone for an underground hell-like area might want to always at about 1000 blocks below the surface, rather than just being placed below the next-highest zone.

Cervator:
Haven't tested yet, but did look overall the code and it looks pretty good to me :-)

GooeyHub:

Vizaxo:
### Interface cleanup

I've cleanup the interface of the region function and layered regions, so the LayeredZoneRegionManager isn't needed (that wasn't so much of a problem, but I think it would get messed up when the layered zones were nested). This also allows the world preview to directly get the target zone by name, rather than hunting through a map of all zones.
 
This was done by allowing the region function to have access to the zone it is in, which will also make implementing other kinds of zones easier (and might help with connectors, because now the function can easily access the parent zone and the other zones that it might connect to).

### Optimisations

I've got the VisualVM profiler working, and I've been using it to optimise the zones. The were pretty slow before, but now I think they don't have much overhead (it turns out adding hundreds of thousands of items a second to small dynamically resizing maps isn't that fast... Now that I think about it's absolutely incredible how much computers can do per second).

The layered zones might still need a bit of tweaking (and making sure they work concurrently, because there's sometimes an issue with the optimisation in 99c6166 causing a NPE and making some chunks not load because they're not properly synchronised across threads).

Given these optimisations I think it's safe to stick with the current region functions for zones. I initially thought they might be too slow (world generation was ~10x slower than normal), but these changes make the speed much more reasonable. I haven't tested with that many zones, so adding lots might still cause performance problems.

engine/src/main/java/org/terasology/world/zones/LayeredZoneRegionFunction.java
```diff
@@ -0,0 +1,200 @@
...
+
+            boolean aboveground = ordering >= 0;
+            int cumulativeDistanceSmall = 0;
+            int cumulativeDistanceLarge = 0;
+            LayerRange layerRange = null;
+
+            List<LayeredZoneRegionFunction> layers =
+                    aboveground ? getAbovegroundLayers(zone) : getUndergroundLayers(zone);
+
+            int i;
```
Cervator: Would be ideal with a better variable name here :-)

engine/src/main/java/org/terasology/world/zones/LayeredZoneRegionFunction.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+        return layerRangeMap.get(pos);
+    }
+
+    private List<LayeredZoneRegionFunction> getSiblings(Zone zone) {
+        if (siblings == null) {
+            siblings = getSiblingRegionFunctions(zone).stream()
+                    .filter(f -> f instanceof LayeredZoneRegionFunction)
+                    .map(l -> (LayeredZoneRegionFunction) l)
+                    .sorted((l1, l2) -> ((Integer) Math.abs(l1.getOrdering())).compareTo(Math.abs(l2.getOrdering())))
```
Cervator: Magic numbers? (11 and 12)

engine/src/main/java/org/terasology/world/zones/LayeredZoneRegionFunction.java
```diff
@@ -0,0 +1,200 @@
...
+                    .filter(LayeredZoneRegionFunction::isUnderground)
+                    .collect(Collectors.toList());
+        }
+         return undergroundLayers;
+    }
+
+    private List<LayeredZoneRegionFunction> getAbovegroundLayers(Zone zone) {
+        if (abovegroundLayers == null) {
+            abovegroundLayers = getSiblings(zone).stream()
+                    .filter(l -> !l.isUnderground())
```
Cervator: `l` maybe renamed to `layer` ?

engine/src/main/java/org/terasology/world/zones/ProviderStore.java
```diff
@@ -0,0 +1,54 @@
...
+package org.terasology.world.zones;
+
+import com.google.common.collect.ImmutableList;
+import org.terasology.world.generation.FacetProvider;
+import org.terasology.world.generation.WorldRasterizer;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class ProviderStore {
```
Cervator: Maybe some class level Javadoc here describing at a high level what the idea is with a provider?

engine/src/main/java/org/terasology/world/generator/WorldGenerator.java
```diff
@@ -39,6 +42,10 @@
 
     World getWorld();
 
+    List<Zone> getZones();
+
+    Zone getNamedZone(String name);
```
skaldarnar: Should this be `Optional<Zone>`? Also, are we going to identify by String or by Uri?

engine/src/main/java/org/terasology/world/zones/LayeredZoneRegionFunction.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * A function that can be used as a {@link Zone#regionFunction} to create zones that are layered on top of each other.
+ *
+ * These layers are ordered according to {@link #ordering}, and have a width of {@link #minWidth}.
```
skaldarnar: I'm not sure about "width" here (it could also be a height or depth). Maybe "extend" would fit better?

engine/src/main/java/org/terasology/world/zones/LayeredZoneRegionFunction.java
skaldarnar: it's layer1 and layer2 I guess, although it is confusing that you mix up `f` for function and `l` for layer here.

engine/src/main/java/org/terasology/world/zones/Zone.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private final List<FacetProvider> facetProviders = new ArrayList<>();
+    private final List<WorldRasterizer> rasterizers = new ArrayList<>();
+    //TODO: add entity providers
+    private final List<FacetLayer> facetLayers = new ArrayList<>();
+
+    private ProviderStore parent;
+
+    private final ZoneRegionFunction regionFunction;
+    private final String name;
```
skaldarnar: again, should this be string or an Uri?

skaldarnar: and probably a few empty lines too much afterwards :wink:

engine/src/main/java/org/terasology/world/zones/ZoneRegionFunction.java
```diff
@@ -0,0 +1,44 @@
...
+     * Calculates whether or not the given block is part of this layer.
+     *
+     * @param x the x position to check
+     * @param y the y position to check
+     * @param z the z position to check
+     * @param region the Region in the area
+     * @param zone the zone that contains this function
+     * @return true if the position is within this layer, false otherwise
+     */
+    boolean apply(int x, int y, int z, Region region, Zone zone);
```
skaldarnar: should there be an overloaded version which takes a `Vector3i` instead of single parameters?

engine/src/main/java/org/terasology/world/generator/WorldGenerator.java
Vizaxo: Yeah, this probably should be an optional. Currently the method's only use is where the name is known to correspond to a zone, but there are uses (specifically with using zones with world plugins) where there might not be a zone for a given name.

I have realised that the current only usage (in the `PreviewWorldScreen`) is redundant (it fetches the target zone's name, then retrieves the same zone again with this method).

Identifying with a URI is probably a good idea, especially if they need to be referenced between modules. I was just using the name as a display name for the preview, but I think there will need to be a nice way of accessing zones across modules.

engine/src/main/java/org/terasology/world/zones/LayeredZoneRegionFunction.java
Vizaxo: Yeah, I wasn't sure either. I think that depth might imply how far away from the surface it is, rather than how thick it is. Extent is closer, but that makes me think of the whole extent (considering all 3 dimensions), rather than just how wide it is. I think that width/height both have the same problem of referring to different things depending on where your axes are placed. I think thickness is the closest word I've found, and it seems like it fits nicely.

engine/src/main/java/org/terasology/world/zones/ZoneRegionFunction.java
Vizaxo: I don't think this method needs to be overloaded; it's only used in the `Zone`, where I think I swapped using a `Vector3i` for `int`s because the profiler showed a not-insignificant amount of time spent creating the vectors (this method gets called a lot).

There are alternative versions available in the constructors for the `Zone` which use vectors (though I'm not sure of the performance implications of that; I was mainly testing with the layered zones, which don't use the vector methods).

GooeyHub:

GooeyHub:

Vizaxo:
### Variable-thickness layered zones

* Note: I've used `width` in the code because I haven't updated it yet, but references to `width` will probably be replaced by `thickness` (unless a better name is found)

Previously, each layered zone was a constant thickness, which limited their use. Now, I've swapped out the min/max width parameters and replaced it with a `LayerThickness` parameter, which is a function that determines the desired thickness of the layer at a given `(x, z)` column. This is an arbitrary function, much like the `ZoneRegionFunction`, with some easy-to-use implementations (`ConstantLayerThickness` and `MinMaxLayerThickness`). These basic implementations will be suitable for most use-cases, but they can be customized or adapted to the module's needs. The `MinMaxLayerThickness` will use a noise function to pick a value between the minimum and maximum at each point (the noise function might need tweaking, or could even be exposed to the modules).

The same system will likely be applied for determining the ordering of the layers, so more parameters can be specified to find the ordering (such as specifying a desired depth/height).

### SingleBlockRasterizer

There's also a `SingleBlockRasterizer`, which will rasterize all of the blocks in the chunk to the given block. Previously this wouldn't've been very useful, but it is useful for block-filling zones.

Cervator:
Merged! Nice work :-)

PR #3056, opened 2017-08-09
[telemetry] More user options

GabrielXia:
### Contains

Now the user can choose which field he wants to send in the `metrics menu`. 
- He can enable/disable a single field, e.g. the "language" field.
- He can enable/disable a whole metric e.g. "System Context", then all fields in that metric will be enabled/disabled.
- Click the check box "Send metrics to server" will enable/disable all fields.
- If all fields in the metric is disabled, that metric will be disabled automatically. If one field is enabled, that metric will be enabled automatically.

Make all the string default to `no_analysed` in ELK, we can finally do some pie graphs :)  The utility.terasology.org:5601 is updated. Here is an example proposed in #3024 

For the headless server, we need to edit in `terasology-server/config.cfg`:
- `"telemetryEnabled": true` to enable telemetry
- `"errorReportingEnabled": true` to enable error reporting
-  copy `bindingMap` in `/config.cfg` to `terasology-server/config.cfg` to have more options :)

The telemetry in the headless server is too manual, if you have some more ideas, please let me know! Thanks :)

Thanks @rzats @oniatus @Cervator @skaldarnar @msteiger @qwc for review and comments!

### DIFF
--- a/engine/src/main/java/org/terasology/config/MetricsUserPermissionConfig.java
+++ b/engine/src/main/java/org/terasology/config/MetricsUserPermissionConfig.java
@@ -19,8 +19,12 @@ import java.util.HashMap;
 import java.util.Map;
 
 /**
+ * This config is used in Telemetry.
+ * It gives the user more options such as sending one part of the fields but not the other part.
  */
 public class MetricsUserPermissionConfig {
+
+    private Map<String, Boolean> bindingMap = new HashMap<>();
 
     public Map<String, Boolean> getBindingMap() {
         return bindingMap;
@@ -29,7 +33,4 @@ public class MetricsUserPermissionConfig {
     public void setBindingMap(Map<String, Boolean> bindingMap) {
         this.bindingMap = bindingMap;
     }
-
-    private Map<String, Boolean> bindingMap = new HashMap<>();
-
 }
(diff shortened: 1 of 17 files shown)


GooeyHub:

GooeyHub:

GabrielXia:
Thanks @Cervator @rzats so much for the test!

@Cervator thanks for the bug that you find. The nlp happens the first time you enable the telemetry, not open the metrics screen in the game and return to the main menu or quit. In this pr, I simply do the nlp check to fix the nlp. Ideally I should add all the telemetry field at the beginning. Because I'll do several changes in the next pr, could I solve this issue totally in the next pr? I could also update all the code in this pr :) That will make this pr more complicated

I think at the bottom of the metrics menu, if you click send metrics to server, it'll enable/diable all the fields. I like also the idea of limiting metrics, do you think it should be done in the client side? e.g. the same user isn't able to send 100 metrics to the server? But that might cause a problem when some devs are debugging the telemetry code

Cervator:
Retested - seems good, thanks :-) Merged! Always good to harden further but yeah that can be whenever.

Enabling all on actually selecting the metrics seems handy.

On throttling - it would have to be a server side setting and I wouldn't be surprised if logstash / snowplow / something already has related options. It would be something like throttling individual sources that are sending unrealistically high amounts of data (likely malicious or a bug)

PR #3061, opened 2017-08-12
Added removeAllWidgets method

nihal111:
Simply adds a method called `removeAllWidgets` that allows clearing all the Widgets that are stored inside a layout.


Cervator:
@nihal111 / @SkySom: Be careful with changes like these, specifically adding new non-default methods to interfaces covered by `@API` - they are technically API violations requiring a new major engine release :-)

In this case I found that it broke the EventualSkills module, which had a related UI class without the new method. Unsure if it might be in use elsewhere so I submitted #3063 as a general fix. I'm liking the utility of default methods although I fear I am growing to enjoy a dangerous tool ... :D

PR #3062, opened 2017-08-13
expose executor service in taskmaster

kaen:
### Contains

A getter method for the ExecutorService used by TaskMaster

There are useful Guava utilities that can make use of ExecutorServices, for example the [SimpleTimeLimiter](https://google.github.io/guava/releases/22.0/api/docs/com/google/common/util/concurrent/SimpleTimeLimiter.html)

I make use of this in FlexiblePathfinding to [manage timeouts](https://github.com/kaen/FlexiblePathfinding/blob/master/src/main/java/org/terasology/flexiblepathfinding/JPSImpl.java#L66-L85) for pathfinding executions. This way I can use google time limiting logic instead of writing my own.

Currently, we can instantiate an ExecutorService directly ourselves, but we can't properly manage its lifecycle. Calling `executorService.shutdown` results in the following permissions error:
```
11:57:17.881 [Pathfinder-0] ERROR o.t.u.concurrency.TaskProcessor - Error in thread Pathfinder-0
java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "modifyThread")
```
Without the ability to call that method in module code, the thread and executor are left quietly running in the background after Terasology shutdown. This is messy and requires use of the red "stop" button in IDEA to fix.

This getter allows use of a managed `ExecutorService` without changing the API whitelist. There is an idiosyncrasy with the behavior but I believe that's well documented in the getter's javadoc, and an example of its usage can be [seen here](https://github.com/kaen/FlexiblePathfinding/blob/master/src/main/java/org/terasology/flexiblepathfinding/PathfinderSystem.java)

### DIFF
--- a/engine/src/main/java/org/terasology/utilities/concurrency/Task.java
+++ b/engine/src/main/java/org/terasology/utilities/concurrency/Task.java
@@ -17,12 +17,27 @@
 package org.terasology.utilities.concurrency;
 
 /**
+ * Task interface for use with {@link TaskMaster}. Typically, this interface is implemented by a BaseTask class
+ * specific to your TaskMaster client code. For example, a pathfinder might create a PathfindingBaseTask, and
+ * extend that to make a PathfindingMapAnalysisTask.
+ * <p>
+ * The {@link #run()} method is called (generally on a separate task processing thread) when the task manager is
+ * ready to execute the task.
+ *
+ * @see TaskMaster
  */
 public interface Task {
 
     String getName();
 
+    /**
+     * Called when the Task is ready to be executed
+     */
     void run();
 
+    /**
+     * Return true if the {@link TaskProcessor} should shut down after performing this task
+     * @return
+     */
     boolean isTerminateSignal();
 }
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Oh so that's one of the causes for the game failing to dispose? That'd be nice to get fixed.

With this merged on its own it won't fix that issue I take it. Can we easily apply a module fix or two on top to then fix existing usages? I figure there must be one in the old school Pathfinding module.

Cervator:
Another small side note: If you understand some of this code better now, mind sprinkling in a little JavaDoc here and there? `TaskMaster` could totally use some class-level Javadoc and so on :D

kaen:
I've checked pathfinding, doesn't look like it uses an executor like this. This fixes a bug in some code I have, and opens up possibilities for using the executor service to do smarter worker handling. However I don't think it can be used to fix any existing bugs (if I've missed an ExecutorService in a module please just point the way).

I am happy to document the class a bit.

GooeyHub:

Cervator:
Thanks for the great Javadoc!

And aw on this maybe not being a smoking gun for current disposal issues. Oh well. We'll get them eventually :-)

Merged!

PR #3070, opened 2017-08-16
First commit of animation tools. Contains very experimental code.

kartikey0303:
Adds a new tool to test animation of entities. Right now only contains option to speed up the animation. 
To get the tool in game open console (by pressing '~') and type "give animTest".
The tool is very experimental right now and does not have a proper UI. The speed can be changed using the UISlider and then the entity needs to spawned again.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.nui.layers.ingame.metrics;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.Lists;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.terasology.assets.ResourceUrn;
+import org.terasology.assets.management.AssetManager;
+import org.terasology.entitySystem.entity.EntityManager;
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.prefab.Prefab;
+import org.terasology.logic.characters.CharacterMovementComponent;
+import org.terasology.logic.characters.StandComponent;
+import org.terasology.logic.characters.WalkComponent;
+import org.terasology.logic.common.InspectionToolComponent;
+import org.terasology.logic.location.LocationComponent;
+import org.terasology.logic.players.LocalPlayer;
+import org.terasology.math.geom.Quat4f;
+import org.terasology.math.geom.Vector3f;
+import org.terasology.registry.In;
+import org.terasology.rendering.assets.material.Material;
+import org.terasology.rendering.assets.skeletalmesh.SkeletalMesh;
+import org.terasology.rendering.logic.SkeletalMeshComponent;
+import org.terasology.rendering.nui.BaseInteractionScreen;
+import org.terasology.rendering.nui.CoreScreenLayer;
+import org.terasology.rendering.nui.widgets.*;
+
+import java.util.ArrayList;
+
+/**
+ */
+public class AnimationScreen extends CoreScreenLayer {
+    private UIText fullDescriptionLabel;
+    private UIText entityIdField;
+    private UIButton spawnEntityIdButton;
+    private UISlider animationSpeedSlider;
+    @In
+    private EntityRef entityRef;
+    @In
+    private LocalPlayer localPlayer;
+    @In
+    private EntityManager entityManager;
+    @In
+    private AssetManager assetManager;
+
+    private static final Logger logger = LoggerFactory.getLogger(AnimationScreen.class);
+
+
+    @Override
+    public void initialise() {
+        spawnEntityIdButton = find("spawnEntityIdButton", UIButton.class);
+        UIDropdown<ResourceUrn> entityDropdown = find("entityDropdown", UIDropdownScrollable.class);
+        logger.info("Number of available skeletal meshes: " + assetManager.getAvailableAssets(SkeletalMesh.class).size());
+        if (entityDropdown != null) {
+            entityDropdown.setOptions(new ArrayList(assetManager.getAvailableAssets(SkeletalMesh.class)));
+        }
+        animationSpeedSlider = find("entityAnimationSpeedSlider", UISlider.class);
+        if (animationSpeedSlider != null) {
+            animationSpeedSlider.setMinimum(-2.5f);
+            animationSpeedSlider.setIncrement(0.1f);
+            animationSpeedSlider.setRange(5.0f);
+            animationSpeedSlider.setPrecision(1);
+        }
+        spawnEntityIdButton.subscribe(widget -> {
+            Vector3f localPlayerPosition = localPlayer.getPosition();
+            Quat4f localPlayerRotation = localPlayer.getRotation();
+            java.util.Optional<Prefab> prefab = assetManager.getAsset(entityDropdown.getSelection(), Prefab.class);
+            if (prefab.isPresent() && prefab.get().getComponent(LocationComponent.class) != null)
+            entityRef = entityManager.create(prefab.get(), localPlayerPosition, localPlayerRotation);
+            SkeletalMeshComponent skeletalMeshComponent = entityRef.getComponent(SkeletalMeshComponent.class);
+            skeletalMeshComponent.animationRate = animationSpeedSlider.getValue();
+            entityRef.saveComponent(skeletalMeshComponent);
+//            StandComponent standComponent = entityRef.getComponent(StandComponent.class);
+//            standComponent.animationPool.clear();
+//            WalkComponent walk = entityRef.getComponent(WalkComponent.class);
+//            walk.animationPool.clear();
+            CharacterMovementComponent movementComponent = entityRef.getComponent(CharacterMovementComponent.class);
+            movementComponent.speedMultiplier = animationSpeedSlider.getValue();
+            entityRef.saveComponent(movementComponent);
+//            Vector3f offset = new Vector3f(localPlayer.getRotation());
+//            ClientComponent clientComponent = sender.getComponent(ClientComponent.class);
+//            LocationComponent characterLocation = clientComponent.character.getComponent(LocationComponent.class);
+//            Vector3f spawnPos = characterLocation.getWorldPosition();
+//            Vector3f offset = new Vector3f(characterLocation.getWorldDirection());
+//            offset.scale(2);
+//            spawnPos.add(offset);
+//            Vector3f dir = new Vector3f(characterLocation.getWorldDirection());
+//            dir.y = 0;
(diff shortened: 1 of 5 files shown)


GooeyHub:

Cervator:
Cool :-) Looking forward to reviewing when I get a chance

@GooeyHub: add to whitelist

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+import org.terasology.rendering.assets.skeletalmesh.SkeletalMesh;
+import org.terasology.rendering.logic.SkeletalMeshComponent;
+import org.terasology.rendering.nui.BaseInteractionScreen;
+import org.terasology.rendering.nui.CoreScreenLayer;
+import org.terasology.rendering.nui.widgets.*;
+
+import java.util.ArrayList;
+
+/**
+ */
```
Cervator: Keep an eye out for empty javadoc blocks like these. Either delete them or please write some minimal Javadoc :-)

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+public class AnimationScreen extends CoreScreenLayer {
+    private UIText fullDescriptionLabel;
```
Cervator: It would probably be better to have the non-`@In`-ed variables be defined below those. That isn't a formal rule - since they're all private fields any order satisfies the Checkstyle rules (although maybe alpha-sorted would be an option). But it seems sensible to consider `@In`-ed variables something more formal to leave in its own group above the others.

Some of the variables also aren't in use any longer so may as well clear those out

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+    @In
+    private EntityRef entityRef;
+    @In
+    private LocalPlayer localPlayer;
+    @In
+    private EntityManager entityManager;
+    @In
+    private AssetManager assetManager;
+
+    private static final Logger logger = LoggerFactory.getLogger(AnimationScreen.class);
```
Cervator: This var should be at the top (private static final comes before plain private)

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.nui.layers.ingame.metrics;
+
+import com.google.common.base.Optional;
```
Cervator: You might have accidentally gotten this `Optional` imported instead of `java.util.Optional` - that's why later on you end up with `java.util.Optional<Prefab> prefab` where if you take off the `java.util.` it results in an error.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(AnimationScreen.class);
+
+
+    @Override
+    public void initialise() {
+        spawnEntityIdButton = find("spawnEntityIdButton", UIButton.class);
+        UIDropdown<ResourceUrn> entityDropdown = find("entityDropdown", UIDropdownScrollable.class);
+        logger.info("Number of available skeletal meshes: " + assetManager.getAvailableAssets(SkeletalMesh.class).size());
+        if (entityDropdown != null) {
+            entityDropdown.setOptions(new ArrayList(assetManager.getAvailableAssets(SkeletalMesh.class)));
```
Cervator: You may want to split this onto multiple lines and do some validation - IntelliJ has two "unchecked" warnings. Also, what *if* it is null? There is no handling for that case. May be worth logging something at `warn` level

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+        if (animationSpeedSlider != null) {
+            animationSpeedSlider.setMinimum(-2.5f);
+            animationSpeedSlider.setIncrement(0.1f);
+            animationSpeedSlider.setRange(5.0f);
+            animationSpeedSlider.setPrecision(1);
+        }
+        spawnEntityIdButton.subscribe(widget -> {
+            Vector3f localPlayerPosition = localPlayer.getPosition();
+            Quat4f localPlayerRotation = localPlayer.getRotation();
+            java.util.Optional<Prefab> prefab = assetManager.getAsset(entityDropdown.getSelection(), Prefab.class);
```
Cervator: This is the line you can both clarify by fixing the import for `Optional` and probably getting the selection value first  - I think you were hitting issues with that, but maybe that was simply from the wrong `Optional` import?

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+            if (prefab.isPresent() && prefab.get().getComponent(LocationComponent.class) != null)
```
Cervator: Always always always use brackets `{ }` for blocks like this - it is exceedingly easy to produce bugs otherwise. For instance here literally only the next line is inside the if. Indentation doesn't indicate that and another reader likely has no idea if you meant for only the next line to count or all the following lines.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+            entityRef = entityManager.create(prefab.get(), localPlayerPosition, localPlayerRotation);
+            SkeletalMeshComponent skeletalMeshComponent = entityRef.getComponent(SkeletalMeshComponent.class);
+            skeletalMeshComponent.animationRate = animationSpeedSlider.getValue();
```
Cervator: You need some validation code here. Right now the slider offers negative values, yet if you pick one the game crashes:

```
java.lang.IndexOutOfBoundsException: index (-1) must not be negative
	at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:310)
	at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:293)
	at com.google.common.collect.RegularImmutableList.get(RegularImmutableList.java:67)
	at org.terasology.rendering.assets.animation.MeshAnimationImpl.getFrame(MeshAnimationImpl.java:64)
	at org.terasology.rendering.logic.SkeletonRenderer.updateSkeletalMeshOfEntity(SkeletonRenderer.java:169)
```

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
kartikey0303: Yeah the Optional import was wrong. I made a mistake there.

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/AnimationScreen.java
kartikey0303: I made a mistake here in thinking the `@Range` in the `skeletalMesh` was for `animationRate`. I changed the code for range from zero.

GooeyHub:

modules/Core/src/main/java/org/terasology/core/debug/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * A Tool for testing the animations of entities.
+ * Animation speed can be edited with this utility
+ */
+public class AnimationScreen extends CoreScreenLayer {
+
+    private static final Logger logger = LoggerFactory.getLogger(AnimationScreen.class);
+    @In
+    private EntityRef entityRef;
```
Cervator: This here doesn't make sense - `EntityRef` isn't any sort of Singleton service, so it can't be injected. That's for `System` type classes that are shared for use by other things.  You should just make it a regular variable in the method where it gets used

modules/Core/assets/ui/ingame/animationTest.ui
```diff
@@ -0,0 +1,77 @@
+{
+    "type": "Core:AnimationScreen",
```
Cervator: You don't need `Core:` in front here. It is a type definition not a reference to something else. Since the item is in Core it'll automatically work when referenced from elsewhere as `Core:animationTest`

modules/Core/assets/prefabs/animTest.prefab
```diff
@@ -0,0 +1,12 @@
+{
+    "parent": "engine:iconItem",
+    "DisplayName": {
+        "name": "Animation Test"
+    },
+    "Item": {
+        "icon": "engine:items#whiteRecipe"
+    },
+    "InteractionScreen": {
```
Cervator: If you add `"InteractionTarget": {},` above this line then the UI opens and works again (along with the other changes)

modules/Core/src/main/java/org/terasology/core/debug/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+import org.terasology.logic.location.LocationComponent;
+import org.terasology.logic.players.LocalPlayer;
+import org.terasology.math.geom.Quat4f;
+import org.terasology.math.geom.Vector3f;
+import org.terasology.registry.In;
+import org.terasology.rendering.assets.material.Material;
+import org.terasology.rendering.assets.skeletalmesh.SkeletalMesh;
+import org.terasology.rendering.logic.SkeletalMeshComponent;
+import org.terasology.rendering.nui.CoreScreenLayer;
+import org.terasology.rendering.nui.widgets.*;
```
Cervator: Avoid wildcard imports (the * bit) - IntelliJ will automatically go to wildcards after a set number of imports from the same package, you can change that so it doesn't by changing "Class count to use import with '*'" to "99" in Settings -> Editor -> Code Style -> Java -> Imports tab

modules/Core/src/main/java/org/terasology/core/debug/AnimationScreen.java
```diff
@@ -0,0 +1,127 @@
...
+            animationSpeedSlider.setMinimum(-0.0f);
+            animationSpeedSlider.setIncrement(0.1f);
+            animationSpeedSlider.setRange(10.0f);
+            animationSpeedSlider.setPrecision(1);
+        }
+        spawnEntityIdButton.subscribe(widget -> {
+            Vector3f localPlayerPosition = localPlayer.getPosition();
+            Quat4f localPlayerRotation = localPlayer.getRotation();
+            Optional<Prefab> prefab = assetManager.getAsset(entityDropdown.getSelection(), Prefab.class);
+            if (prefab.isPresent() && prefab.get().getComponent(LocationComponent.class) != null) {
```
Cervator: You may want to add an `else` leg here and/or put the remainder of the live code inside the if's body - in the else case `entityRef` wouldn't be initialized and likely the later lines would cause a crash.

GooeyHub:

GooeyHub:

Cervator:
Latest changes leave it pretty good and functional :-)

Some more hardening against nulls would be good in a next PR, such as validating that a given entity actually has a needed component or logging a warning that it doesn't then returning early.

Next up another dropdown for individual animations would be ideal (probably clearing out any other animations so only the selected one plays)

PR #3072, opened 2017-08-17
Add AdvancedFilter section to SelectModulesScreen

### LINKED ISSUE #2935: Allow filtering out library modules from module list
The user currently gets all modules listed by default to select from. A lot of those modules are however only intended to be used as libraries by other modules. It does not make sense for the user to select library modules manually as they do nothing by themselfs. For this reason it does not make sense to list those module in the module list by default. e.g. the Pathfinding module is such a library module that is an important dependency for other modules but does not provide user usable features by its own.

So I suggest that we hide library modules by default in the module list and allow via a "show libraries" check box to show them in case someone is interested in just looking at the dependencies.

In order to identify library modules I propose that we add a isLibrary flag to the module.txt

TheFlash98:
### Contains
Fixes #2935 
In this PR, I've created a new checkbox associated with a new button, `AdvancedFilter`. When you click the button, a new screen opens with checkboxes associated with different module types. I have put around 6 types from the forum discussion [here](http://forum.terasology.org/threads/module-categories.1068/). Adding/Deleting new types would be fairly simple. If the checkbox next to `Advanced Filter` button is ticked, the list of modules gets sorted according to the module types selected. The state of all the checkboxes have been saved in `config.cfg`. 

### How to test
To test I would suggest you add a field like `"isLibrary": true` to the `module.txt` of any of your modules and go ahead and try filtering it out. It would be the only one in the list. You can edit `module.txt` of multiple modules, add multiple filters and test if the desired modules are listed. The available options in this PR are `isLibrary`, `isAugmentation`, `isWorld`, `isSpecial`, `isAsset` and `isGameplay`. All these are from the same forum discussion linked above.


GooeyHub:

nihal111:
@GooeyHub ok to test

GooeyHub:

engine/src/main/java/org/terasology/config/SelectModulesConfig.java
```diff
@@ -25,4 +32,60 @@ public boolean isChecked() {
     public void setIsChecked(boolean value) {
         this.isChecked = value;
     }
+
+    public boolean isLibraryChecked() {
+        return isLibraryChecked;
+    }
+
+    public void setIsLibraryChecked(boolean value) {
+        this.isLibraryChecked =value;
```
gianluca-nitti: I think there should be a space after the `=`, here and on the other setters (not sure if it's reported by Checkstyle but I always saw spaces both before and after operators in the Terasology codebase)

engine/src/main/java/org/terasology/engine/module/StandardModuleExtension.java
```diff
@@ -56,6 +61,31 @@ public static boolean isGameplayModule(Module module) {
         return isGameplay != null && isGameplay;
     }
 
+    public static boolean isAssetModule(Module module) {
+        Boolean isAssetplay = module.getMetadata().getExtension(IS_ASSET.getKey(), Boolean.class);
```
gianluca-nitti: Personally, since it's used in multiple methods, I would have extracted the code in common between this and the other boolean getters to a separate private method, i.e. something like
```
private static boolean getBooleanExtension(Module module, StandardModuleExtension ext) {
    Boolean result = module.getMetadata().getExtension(ext.getKey(), Boolean.class);
    return result != null && result;
}
```
and then rewrite this and the other getters as
```
public static boolean isAssetModule(Module module) {
    return getBooleanExtension(module, IS_ASSET);
}
```
but I see that the similar accessor methods before your changes were like yours, so not a big deal probably.

Oh and by the way, you probably meant `isAsset` instead of `isAssetplay` here.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/AdvanceModuleFilter.java
```diff
@@ -0,0 +1,147 @@
...
+    private UICheckbox assetCheckBox;
+    private UICheckbox specialCheckBox;
+    private UICheckbox augmentationCheckBox;
+    private UICheckbox worldCheckBox;
+    private UICheckbox gameplayCheckBox;
+
+    @In
+    private Config config;
+    private SelectModulesConfig selectModulesConfig;
+    private SelectModulesScreen selectModulesScreen;
```
gianluca-nitti: This isn't used anywhere if I see correctly? (not sure because I'm looking from GitHub and not using an IDE right now)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/AdvanceModuleFilter.java
```diff
@@ -0,0 +1,147 @@
...
+
+    @Override
+    public void initialise() {
+
+        selectModulesConfig = config.getSelectModulesConfig();
+
+        libraryCheckBox = find("libraryCheckbox", UICheckbox.class);
+        libraryCheckBox.bindChecked(
```
gianluca-nitti: I don't see the point of declaring the controls as class fields, in my understanding they could be local variables in this method but most of them don't need to be declared at all since the binding system is being used, e.g. if I understand correctly here you could just do `find("libraryCheckbox", UICheckbox.class).bindCkecked(` and delete the `libraryCheckBox` field (same for the other checkboxes).

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -403,6 +433,32 @@ private void filterText(String newText) {
...
+
+    private void moduleWiseFilter() {
+        sortedModules.clear();
+        for (ModuleSelectionInfo m : allSortedModules) {
+            if (selectModulesConfig.isLibraryChecked() && StandardModuleExtension.isLibraryModule(moduleManager.getRegistry().getLatestModuleVersion(m.getMetadata().getId()))) {
+                sortedModules.add(m);
+            }
+            if (selectModulesConfig.isAssetChecked() && StandardModuleExtension.isAssetModule(moduleManager.getRegistry().getLatestModuleVersion(m.getMetadata().getId()))) {
+                sortedModules.add(m);
+
```
gianluca-nitti: Empty line

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -403,6 +433,32 @@ private void filterText(String newText) {
         }
     }
 
+
+    private void moduleWiseFilter() {
+        sortedModules.clear();
+        for (ModuleSelectionInfo m : allSortedModules) {
+            if (selectModulesConfig.isLibraryChecked() && StandardModuleExtension.isLibraryModule(moduleManager.getRegistry().getLatestModuleVersion(m.getMetadata().getId()))) {
```
gianluca-nitti: I see possible improvements here, for example you could extract `moduleManager.getRegistry().getLatestModuleVersion(m.getMetadata().getId())` to a variable instead of repeating it in each check.
A more elaborate refactoring could be extracting a method like `moduleMatchesFilter(Module module, StandardModuleExtension filter)` but it would require more changes, like extracting the possible filters (asset, gameplay, etc) to a custom enum, and saving a `java.util.Set` of said enum in the config and as a single extension rather than multiple boolean extensions and config fields.
But probably it's not necessary for now since it's quite readable anyway (in my understanding at least).

TheFlash98:
Hey @gianluca-nitti thank you so much for your review. I'm very new to the community and your review about code styles and structures would even help me with my future PRs. I've made most the changes you asked for. However on some debugging I noticed that when once goes to the AdvanceFilterScreen and tick/unticks some boxes and then comes back, the changes are implemented only when the AdvanceFilterCheckbox is unticked and then ticked again. This will happen as the method which sorts everything, the `filterModules` method is only triggered when there is a change to the box. Any suggestions how I can run that method everything `triggerBackAnimation()` is executed in the `AdvanceModuleFilter.java`

engine/src/main/java/org/terasology/config/SelectModulesConfig.java
TheFlash98: Fixed!

engine/src/main/java/org/terasology/engine/module/StandardModuleExtension.java
TheFlash98: Implemented the method. Thanks for the suggestion!

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/AdvanceModuleFilter.java
TheFlash98: Yeah, it wasn't being used. Removed it.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/AdvanceModuleFilter.java
TheFlash98: Actually there is a separate method, the `filterModules()` method which uses it now. So this would be needed **now**

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
TheFlash98: Removed!

GooeyHub:

Cervator:
Tested this out :-)

For some reason adding `"isLibrary": true,` to a `module.txt` then attempting a filter didn't seem to show any modules at all. Can you drop in a bigger snippet if I'm not getting it right?

At first I thought something was wrong with the UI - it seems very unnatural to have a checkbox next to a button that relates to what's on the other screen. Maybe we can find a better UI approach somehow?

I'm not sure if we have a good UI element or the right option for this, but rather than checkbox + button could we have a toggle for exposing advanced filter checkboxes off to the one side of the existing frame? So essentially just widen the frame to allow showing the extra elements while the advanced filter mode is active.

TheFlash98:
I'll check about why this is happening, I might have broken something in the last 3 commits. 

Are you suggesting that I should completely do away with the button and new screen, just add a checkbox? So whenever the checkbox is ticked, a new section should appear on the same screen showing the Advanced Filter setting, else it should be hidden. Sounds like a good idea, but don't you think that Select Modules Screen would be too crowded?

GooeyHub:

TheFlash98:
Hmmm, this has never happened before. Everything is working fine locally. What does this fail mean?

Cervator:
Figured out the PR builder issue - I merged a Gradle update which likely also changed the version of the unit test reporting, making the path to the results changed. Should be fixed now :-)

Can't remember if this works to re-test, but let us see:

@GooeyHub test this please

As for the UI ... yeah it might get a bit crowded. I was thinking some sort of toggle like a little side widget on the frame that you click to sort of pull out advanced options (not a checkbox). But I'm not sure how crowded it would get.

We need a module details screen at some point, but that's no good for filtering. Should we have a whole variant of the module selection screen that shows the advanced filters *and* the module list being filtered? It was odd to me having to pick filters then "go back" to the main module screen to see the effect.

GooeyHub:

TheFlash98:
I agree that is a bit odd to go back to see the sorted modules list. However we can lengthen the already existing screen and have 3 columns(Modules list, Module Details, Advance Filter). The 3rd column could be triggered by some widget, or be visible the entire time.

Cervator:
> I agree that is a bit odd to go back to see the sorted modules list. However we can lengthen the already existing screen and have 3 columns(Modules list, Module Details, Advance Filter). The 3rd column could be triggered by some widget, or be visible the entire time.

Yep I'd be curious to see that. I guess then the first column could be cleaned up some - the name filter and "local only" could also be moved to the new filter column :-)

TheFlash98:
Sounds like a good idea. I guess Module Details could be the last column towards the right, the middle column would be a general filter column and towards the right we display the list of filtered modules.

Cervator:
How about putting the module list in the middle? Then your eyes don't have to hop over the filter column to connect a given module with its details.

TheFlash98:
Okay, that's better!

GooeyHub:

TheFlash98:
@Cervator changed the Select Modules Screen to now have a new section for filtering on the left. Go ahead and tick whatever filter you want and press the "Activate Filter" button. You will see the modules list change accordingly. 🙂

oniatus:
I did a short test with these modules: 
[TestModules.zip](https://github.com/MovingBlocks/Terasology/files/1385280/TestModules.zip)


Notes while testing:
1. The filtering should happen when a checkbox changes, the separate button makes the UI less intuitive for me.
2. I would like to have all filters in one column below each other, left side of the module screen is okay though. Also sort them by priority for the user, e.g. Local Only, Gameplay, World, Augumentation, Library, Asset, Special.
3. I'd expect the filters to work like OR, not AND. For example when checking _World_ and _Special_ and _Local Only_, I would expect to see the modules TestWorld and TestSpecial.

Cervator:
@TheFlash98 - pinging for response after the comments from @oniatus :-)

TheFlash98:
Hey @cervator I spoke to @oniatus about this over slack, a little busy with college stuff right now. We will back to this soon. :)

jellysnake:
Hey, @TheFlash98. Would like to include this as a Google Code In task. Are you still busy with college stuff?

TheFlash98:
Hey @jellysnake, yeah. Sorry about the delay. Exams get over on Wednesday will get to this straight away. I know the GCI deadline is on 28th. I will try finishing is as soon as possible after 22nd. Is that okay?

jellysnake:
Wonderful. That will be fine. We can still publish and make tasks after GCI starts so it's not a hard deadline. So long as you're willing to have it as a task is fine :)
Good luck with your exams!

TheFlash98:
That's great. Thank you! 😄

GooeyHub:

oniatus:
@jellysnake Is this PR feature-complete for you? I think the filtering works as expected.
Some open points from my side:
- Gameplay does not show remote gameplay modules (may need some digging)
- Module download for remote modules fails (could be unreleated to the PR)

When the PR fulfils the features, I would like to go over the code again. At the moment there is a lot of duplicated or very similar code. It is probably easier to maintain all possible properties in a single enumeration and use a single field in the modules for it. We could then generate the comboboxes and filters instead of copying code.

TheFlash98:
> Gameplay does not show remote gameplay modules (may need some digging)

Can you tell me exactly what you did to encounter this?

> It is probably easier to maintain all possible properties in a single enumeration

Could you elaborate about this? Any idea about how to implement it?

TheFlash98:
Ping @oniatus for some guidance here 🙂

oniatus:
Gameplay modules: Start game with a clean directory, open select modules and check the gameplay filter.
All grey remote modules are filtered out. I would expect a module like `JoshariasSurvival` to be visible in the list and then have the option to download it. If the module is available locally in the workspace, the filter will keep it.

Will need some more time to think about the enumeration changes :)

TheFlash98:
I think the issue with `JoshariasSurvival` is that the field `"isGameplay"` has been given a string value `"true"` instead of a boolean value `true` as you can see [here](https://github.com/Terasology/JoshariasSurvival/blob/master/module.txt#L24) I can send a PR to JoshariasSurvival regarding the same, quick fix!

engine/src/main/java/org/terasology/config/SelectModulesConfig.java
```diff
@@ -17,6 +17,13 @@
 
 public class SelectModulesConfig {
     private boolean isChecked;
+    private boolean isLibraryChecked;
```
msteiger: I would rather make this an `EnumSet` instead.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -359,50 +362,141 @@ public String get() {
...
     }
 
+
+    private void advancedModuleFilter() {
+        Iterator<ModuleSelectionInfo> iter = sortedModules.iterator();
+        while (iter.hasNext()) {
+            ModuleSelectionInfo m = iter.next();
+            Module module;
+            if(m.isPresent()) {
+                module = moduleManager.getRegistry().getLatestModuleVersion(m.getMetadata().getId());//(m.getOnlineVersion() == null) ? m.getLatestVersion() : m.getOnlineVersion();
```
msteiger: Please remove the commented code.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -359,50 +362,141 @@ public String get() {
...
+            }
+            else {
+                module = (m.getOnlineVersion() == null) ? m.getLatestVersion() : m.getOnlineVersion();
+            }
+
+            if(selectModulesConfig.isLibraryChecked() && StandardModuleExtension.isLibraryModule(module)) {
```
msteiger: I think it would be better to get the different types of modules in a `Set` so you can do simply call `removeAll()` or `retainAll()` instead of manually iterating over all possible values. This is also somewhat error prone, in particular when a new type is added.

mkienenb:
@TheFlash98 Any progress on this?   This seems like it should be one of our higher-priority items.

PR #3016, opened 2017-07-22
[telemetry] game play metric

GabrielXia:
### Contains

Game play metric including:

- Block destroyed
- Block placed
- Monster killed
- Game configuration: word generator, network mode, language, player height, player eye height
- Game play stats: distance traveled, play time

### How to test

- In a game, the stats will show in the `Metrics Menu`
- If you enable `Send metrics to server`, the metrics will be sent to http://utility.terasology.org/. You can see on port 5601 http://utility.terasology.org:5601/

gameplay
gameconfiguration
blockdestroyed

@rzats @oniatus @skaldarnar @msteiger @Cervator @qwc thanks for review :-)

PR #3075, opened 2017-08-20
Telemetry API

GabrielXia:
### Contains

- A Telemetry API that could be used in the module. The test module is here: https://github.com/GabrielXia/TelemetryApiTest 
    - The new telemetry in module will also be shown in the `Metrics Menu` (after a game launch of cause, since in the `main menu` you have not chosen the module)
   - The option for the module field is also valid
- Follow-up to comments in #3016: mark ip address and timezone field as "anonyme"
- Follow-up to #3056: check and update the map which notes the user's telemetry options when game initialise() so the value could be null, that npe could not happen 
- Rewrite the `Metrics.class` which store all the `Metric.class` instances. Now if the new `Metric` instance is created using the constructor in parent class `Metric.class`, it will be added to `Metrics` automatically. 

### How to test

Fetch the test module in https://github.com/GabrielXia/TelemetryApiTest , enable the telemetry in the `Metrics Menu`, select the module and load the game. The `TelemetryApiTestMetric` will be sent to the server and you can see the fields in the `Metrics Menu` in game. You can see the metric sent in http://utility.terasology.org:5601/


GabrielXia:
@rzats @oniatus @Cervator @skaldarnar @msteiger @qwc thinks for review :)

GooeyHub:

engine/src/main/java/org/terasology/telemetry/Metrics.java
```diff
@@ -27,101 +28,50 @@
...
-    private BlockPlacedMetric blockPlacedMetric;
-
-    private GamePlayMetric gamePlayMetric;
-
-    private MonsterKilledMetric monsterKilledMetric;
-
-    public Metrics() {
-
-    }
+    private Map<String, Metric> metricsMap = new HashMap<>();
```
oniatus: This maps classnames to metrics, how about a name like `metricByClassName` or `classNameToMetric`? `metricsMap` is very generic and adds little extra information to the field definition.

engine/src/main/java/org/terasology/telemetry/TelemetrySystem.java
```diff
@@ -122,52 +151,76 @@ private void recordPlayTime() {
...
 
-            MonsterKilledMetric monsterKilledMetric = metrics.getMonsterKilledMetric();
-            Unstructured unstructuredMonsterKilled = monsterKilledMetric.getUnstructuredMetric();
-            TelemetryUtils.trackMetric(emitter, trackerNamespace, unstructuredMonsterKilled, monsterKilledMetric, bindingMap);
+            optional = metrics.getMetric(MonsterKilledMetric.class);
+            if (optional.isPresent()) {
+                Metric monsterKilledMetric = optional.get();
+                Unstructured unstructuredMonsterKilled = monsterKilledMetric.getUnstructuredMetric();
+                TelemetryUtils.trackMetric(emitter, trackerNamespace, unstructuredMonsterKilled, monsterKilledMetric, bindingMap);
+            }
```
oniatus: I see a pattern and a lot of copy-paste code here ;)

engine/src/main/java/org/terasology/telemetry/TelemetryUtils.java
```diff
@@ -47,30 +51,60 @@
...
+                    .ipAddress("anonymous")
+                    .timezone("anonymous")
+                    .build();
 
-        // initialise tracker
-        Tracker tracker = new Tracker.TrackerBuilder(emitter, nameSpace, TelemetryParams.APP_ID_TERASOLOGY)
-                .subject(subject)
-                .platform(TelemetryParams.PLATFORM_DESKTOP)
-                .build();
+            // initialise tracker
```
oniatus: Thats what someone would expect when reading `Tracker tracker = new Tracker....`.
No extra information from the comment -> delete.

engine/src/main/java/org/terasology/telemetry/metrics/Metric.java
```diff
@@ -21,24 +21,43 @@
...
+    protected Map metricMap = new HashMap();
+
+    /**
+     * One a new metric is created, it should be added to the {@link org.terasology.telemetry.Metrics}.
+     * Use this constructor (super(context);) will add metric automatically to Metrics class.
+     * @param context the game context.
+     */
+    public Metric(Context context) {
+        Metrics metrics = context.get(Metrics.class);
+        addToMetrics(metrics);
```
oniatus: I don't like this kind of API for two reasons:
1. The initialization in the Metrics class looks like floating and unused object instances and it is not clear where a instance to a metric is stored.
Also, you do the initialization in metrics, so Metrics schould know where to store a metric.
2. You can't use a metric in any other place without a valid context containing a Metrics class. There is no way to create an instance e.g. for tests.

If you want to force metrics to be on the metrics class and inside the context, why not use an annotation as marker and auto register all annotated metrics. This would also drive the Open-Closed Principle as you could add extra Metrics without changing the Metrics class again and again for each metric.

engine/src/main/java/org/terasology/telemetry/metrics/Metric.java
```diff
@@ -21,24 +21,43 @@
...
+ * All metric fields should be annotated {@link org.terasology.telemetry.TelemetryField}.
  * An example is {@link org.terasology.telemetry.metrics.SystemContextMetric}.
  */
+@API
 public abstract class Metric {
 
     private static final Logger logger = LoggerFactory.getLogger(Metric.class);
 
-    protected Map metricMap;
+    protected Map metricMap = new HashMap();
```
oniatus: add type parameteres (`<String,Object>`?) and a better name and/or javadoc.

engine/src/main/java/org/terasology/telemetry/metrics/Metric.java
```diff
@@ -51,20 +70,22 @@
      * @return a map with key (field's name) and value (field's value).
      */
     public Map<String, ?> getFieldValueMap() {
+        return AccessController.doPrivileged((PrivilegedAction<Map<String, ?>>) () -> {
```
oniatus: A get method with more code than one return statement has a bad name.
This is more like a `createTelemetryFieldValueMap`.

GabrielXia:
Thanks @oniatus for the review.

Appreciate so much for the comments on code style and I like so much the idea of fetching the metrics by the annotation! Now the pr is updated, thanks again for the review :) @rzats @oniatus @Cervator @skaldarnar @msteiger @qwc

GooeyHub:

engine/src/main/java/org/terasology/config/Config.java
```diff
@@ -65,6 +66,7 @@
  * Terasology user config. Holds the various global configuration information that the user can modify. It can be saved
  * and loaded in a JSON format.
  */
+@API
```
Cervator: I'm not sure if it is safe giving modules access to the root `Config` object? Wouldn't that in theory allow modules to read the "security" block which includes some certificate stuff? Something deeper like only the metrics subset seems fine, but I dunno if that's problematic

engine/src/main/java/org/terasology/telemetry/logstash/TelemetryLogstashAppender.java
```diff
@@ -32,12 +31,13 @@
...
 import org.slf4j.LoggerFactory;
 import org.terasology.context.Context;
-
+import org.terasology.module.sandbox.API;
 
 /**
  * This is a logback Logstash appender that enriches error logs and sent them to the Logstash in server.
  * The constructor has the default configuration of this appender. The destination of this appender will be set when error reporting is enabled.
  */
+@API
```
Cervator: Is this one safe? I'm not sure if this class extending `LogstashTcpSocketAppender` would allow a malicious module to start messing around with something network related. As an alternative you could perhaps provide a wrapper method somewhere that indirectly lets a user request a new appender or something.

GabrielXia:
@Cervator @oniatus thanks for the reviews and the suggestions :)

I removed some classes that might cause a security problem: those who have something to do with network (`Emitter`, `Logstash`), those who might change user's options on telemetry(`TelemetryConfig`). I used a wrapper `TelemetryConfiguration` to provide modules with necessary configuration for telemetry.

@Cervator I don't know what the module could do with `logstashAppender` now, so I just removed it for the moment.

Thanks for the review and let me know there is anything you don't like :)

GooeyHub:

Cervator:
Thanks for the security tweaks - I'm not sure either exactly what might be a risk where, but better safe than sorry :-)

Merged!

One remaining question: does it make sense to make time zone anonymous as well as IP? An IP you can do something with, not so much time zone or even country info, and those might be a bit interesting to see roughly where in the world we have players.

And when something does go fully anonymous (rather than scrambled like a user id that we need to make records distinct) - why are we even storing the field at all? Shouldn't we just exclude it entirely? There are also bunches of nulls when you dig around on the utility server. Are we expecting those things to occasionally be populated or could we exclude a bunch of them to cut down on fluff?

GabrielXia:
@Cervator thanks for the suggestions!

Set timezone to non anonymous via #3083 

Based on pr #3056, if all the telemetry fields are disabled, that metric will be disabled automatically, thus it will not be sent to the server. But for metric such as `BlockDestroyed`, if we don't have any blocks destroyed, it will be sent to the server. I'll fix this as soon as possible.

Having discussed with @rzats we don't find a way yet to exclude the null fields, but I'll difinately will keep an eye on that :)

PR #3079, opened 2017-08-24
Zones 2

Vizaxo:
### Contains

* Some small fixes for zones
* Allowing zones to be used as world plugins
* Updating the way the layers are arranged to be more useful (the top block of the surface is counted as underground, so the ground layer can have more than 1 layer of blocks)

### How to test

* Plugin zones can be tested by cloning my [Inferno fork](https://github.com/Vizaxo/Inferno) on the `v2/zone-plugin` branch. Enabling the Inferno module will enable Inferno in every world generator (that has plugins enabled), not just the inferno world generator.
* I've also made an example world generator using zones: [PerlinZonesExample](https://github.com/Vizaxo/PerlinZonesExample).

### Outstanding (but can be merged without these)

* I haven't updated the Zone name to be a URI yet. I'm not sure whether it is needed, because I don't know if getting certain zones by name is useful (currently the only place is in the tests, which can be rewritten to not use it, but it may be useful when adding plugin zones as nested zones on a given named zone). Also, the main reason for the name was to show it on the preview screen, and I don't think the Uri preserves the case of the name, which might be annoying for some zones.
* The plugin zones can currently only be added to the base WorldGenerator, not to other zones. Adding them as nested zones would be useful to, for example, add biomes to the surface layer.
* The layer ordering probably needs to be updated, to allow layers to be placed at a given depth/height.


GooeyHub:

engine/src/main/java/org/terasology/world/zones/LayerThickness.java
```diff
@@ -17,11 +17,24 @@
 
 import org.terasology.module.sandbox.API;
 
+/**
+ * This function is used to determine the thickness of a {@link LayeredZoneRegionFunction} at each point on the layer.
+ */
 @API
 @FunctionalInterface
-public interface LayerWidth {
+public interface LayerThickness {
```
skaldarnar: Thickness is a way better term than width... and we can haz Javadoc 😄

engine-tests/src/test/java/org/terasology/world/Zones/LayeredZoneRegionFunctionTest.java
```diff
@@ -51,11 +52,12 @@
 
     @Before
     public void setup() {
-        parent .addZone(new Zone("Surface", new LayeredZoneRegionFunction(new MinMaxLayerWidth(100, 100), SURFACE)))
-                .addZone(new Zone("Low sky", new LayeredZoneRegionFunction(new MinMaxLayerWidth(100, 100), LOW_SKY)))
-                .addZone(new Zone("Medium sky", new LayeredZoneRegionFunction(new MinMaxLayerWidth(100, 100), MEDIUM_SKY)))
-                .addZone(new Zone("Shallow underground", new LayeredZoneRegionFunction(new MinMaxLayerWidth(100, 100), SHALLOW_UNDERGROUND)))
-                .addZone(new Zone("Medium underground", new LayeredZoneRegionFunction(new MinMaxLayerWidth(100, 100), MEDIUM_UNDERGROUND)));
+        parent .addZone(new Zone("Above ground", new LayeredZoneRegionFunction(new MinMaxLayerThickness(100, 100), ABOVE_GROUND)))
```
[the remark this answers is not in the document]
Vizaxo: Yeah, good spot. I was playing around with different ways of formatting this and didn't properly revert the changes.

engine-tests/src/test/java/org/terasology/world/Zones/LayeredZoneRegionFunctionTest.java
```diff
@@ -38,11 +38,12 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.terasology.world.zones.LayeredZoneRegionFunction.LayeredZoneOrdering.ABOVE_GROUND;
```
Cervator: This block of static imports practically screams to be an exception to alpha sorting imports, so they could be listed in order based on how the layers would come one after another, but alas! I suspect trying to make an exception work would be more hassle than it would be worth seeing the "right" order :-)

engine-tests/src/test/java/org/terasology/world/Zones/LayeredZoneRegionFunctionTest.java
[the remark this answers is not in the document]
Cervator: Unlike with the import block, would it be fair to sort the addition of the zones here logically from top to bottom? So probably medium sky -> low sky -> above ground -> ground -> shallow underground -> medium underground? And generally trying to follow such approaches to make the overall setup feel more natural?

engine/src/main/java/org/terasology/world/zones/LayeredZoneRegionFunction.java
```diff
@@ -40,15 +40,16 @@
...
+    private LayerThickness layerThickness;
     private long seed;
     private Zone parent;
 
     public static final class LayeredZoneOrdering {
-        public static final int HIGH_SKY = 300;
-        public static final int MEDIUM_SKY = 200;
-        public static final int LOW_SKY = 100;
-        public static final int SURFACE = 0;
+        public static final int HIGH_SKY = 400;
```
Cervator: Yay! Logical ordering :D

engine-tests/src/test/java/org/terasology/world/Zones/LayeredZoneRegionFunctionTest.java
Vizaxo: I agree that would be nice, but like you say it would probably be more hassle than its worth. I (and I'd imagine most people) just use Intellij's auto-imports, so it would require overriding that for specific classes, making sure everyone uses that override (including people who don't use Intellij), and keeping it up-to-date if the names/values change.

It would probably break at some point, creating annoying merge conflicts, and the potential gain is rather small (especially given that imports already have a logical and consistent ordering scheme that works well enough).

engine-tests/src/test/java/org/terasology/world/Zones/LayeredZoneRegionFunctionTest.java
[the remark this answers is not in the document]
Vizaxo: @Cervator Yeah, definitely. I think I had it in a logical order at the start, but kept adding and changing stuff without updating it.

Vizaxo:
@Cervator I've made those small formatting/ordering changes, so I think it's ready to merge now :)

GooeyHub:

Cervator:
Thanks for tweaks - merged :-)

Issue #59, opened 2012-01-28
Review applet jars in source dir


Cervator:
The applet was added to the codebase with the standard jar lineup, but we also have a LWJGL jar in the main lib dir. Need to review which jar files we really need in the applet source dir (unique vs the main lib dir) and which we can simply copy from the main lib dir instead in the gant build target for the applet.

This item should be fairly friendly to a contributor with some familiarity with scripting (Gant is easy and very much like Ant) and Java library jars. Not a rush item.

PR #3080, opened 2017-08-25
Removing assembly BT functionality + Light and Shadow related tweaks

dkambersky:
As the bytecode/assembly parts of the new BT system are sort of obscure and with an unclear use case, they've been removed for now. If we ever need to re-instate them for some reason, we can find them in the git history.

Also contains minor tweaks for Light and Shadow / Work related functionality, accompanies [L&S #59](https://github.com/Terasology/LightAndShadow/pull/59) and related PRs.


GooeyHub:

PR #3083, opened 2017-08-27
uncover timezone and rename

GabrielXia:
### Contains

Follows-up to #3075, uncover timezone info in telemetry

### DIFF
--- a/engine/src/main/java/org/terasology/config/facade/TelemetryConfiguration.java
+++ b/engine/src/main/java/org/terasology/config/facade/TelemetryConfiguration.java
@@ -32,5 +32,5 @@ public interface TelemetryConfiguration {
 
     Boolean get(String telemetryField);
 
-    boolean containField(String telemetryField);
+    boolean containsField(String telemetryField);
 }
--- a/engine/src/main/java/org/terasology/config/facade/TelemetryConfigurationImpl.java
+++ b/engine/src/main/java/org/terasology/config/facade/TelemetryConfigurationImpl.java
@@ -49,7 +49,7 @@ public class TelemetryConfigurationImpl implements TelemetryConfiguration {
         return bindingMap.get(telemetryField);
     }
 
-    public boolean containField(String telemetryField) {
+    public boolean containsField(String telemetryField) {
         Map<String, Boolean> bindingMap = telemetryConfig.getMetricsUserPermissionConfig().getBindingMap();
         return bindingMap.containsKey(telemetryField);
     }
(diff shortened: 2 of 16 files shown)


GooeyHub:

GabrielXia:
As mentioned in #3016 Rename `MonsterKilled` to `CreatureKilled` :)

GooeyHub:

GooeyHub:

GabrielXia:
In this commit:https://github.com/MovingBlocks/Terasology/pull/3083/commits/8db1d3570bd2d7cfd6b23e4a62a553cc800e3df0 If a metric such as `BlockDestroyedMetric` has no data, this is to say, there is no block destroyed record, then we don't send this metric. Does it make sense?

How to test: Don't destroyed blocks in a game, `Metrics Menu` shows that there is not blocks destroyed. Then exit to the main menu, the `BlockDestroyedMetric` is not sent to the server. (In before, it will be sent)

GabrielXia:
Update to https://github.com/MovingBlocks/Terasology/pull/3083/commits/04154ee68a98c0bc5d3464cc5357456df53b5c4f

1. In `fetchMetricAndSend` method, found two args not needed, so removed them
2. Add a `filterMetricMap` method takes `TelemetryConfiguration` as a param in stead of `bindingMap`, so that an other module could use it, e.g. https://github.com/GabrielXia/TelemetryApiTest/blob/master/src/main/java/org/terasology/telemetry/metrics/TelemetryApiTestMetric.java

p.s. I wrote a draft doc about the tutorial module: https://github.com/GabrielXia/TelemetryApiTest/wiki thanks for suggestions :)

GooeyHub:

Cervator:
> In this commit:8db1d35 If a metric such as BlockDestroyedMetric has no data, this is to say, there is no block destroyed record, then we don't send this metric. Does it make sense?

> How to test: Don't destroyed blocks in a game, Metrics Menu shows that there is not blocks destroyed. Then exit to the main menu, the BlockDestroyedMetric is not sent to the server. (In before, it will be sent)

I'm thinking if 0 is a valid value (which it would be for blocks destroyed) then it should be set as a default value during initialization somewhere. Then it will also qualify for sending, right? Some metrics probably should just have good defaults like that.

Anyway: tested out real quick and merged! Thanks :-)

GabrielXia:
@Cervator Thanks for the tests and suggestions.

For example, the blocks destroyed metric includes block type -> block destroyed number. The metric will add a new block only if there is a new kind of block destroyed. Do you think we should set some default block type and set their value to 0? If so, which are those default blocks?

Cervator:
Ahh, gotcha, it is a map. Okay. Hmm. At that point I don't know if it matters enough to throw in something special, like 0 destroyed air blocks. Up to you / how it might help the graphs look better :-)

PR #3085, opened 2017-08-29
Implemented Tweaks to block families

pollend:
I thought I could simplify the process for adding block families to an existing block. I replaced the entire factory pattern with a block family inheritance scheme that takes advantage of dependency injection. Any additional systems that are needed by the family are passed in by injection. Any additional factors that would be covered by the family is handled through annotation. For instance a family might need access to a secondary system when the actual block is placed. This avoids all the fixed limitations of the factory pattern. A family has only access to what is available to the factory or what is handled in a given function. Access to anything outside of this scope requires adding the variable to the family and tweaking the call throughout the engine. I think this is problematic for the long term use of a family.  

This also reduces the additional boilerplate needed to get started with block families. In the old scheme you had to create a factory along with a family that returns that family. There is also the additional connectionCondition that is limited to the location the side that is being connected to the worldprovider and entityRegistry. You may need access to the owner of the placed block but that is not available for the class. You would need to also reimplement the UpdatesWithNeighboursFamily. I believe the factory pattern produces a lot of extra unnecessary boilerplate code. I think this scheme is somewhat fragile and hard to use. It shouldn't be necessary to reinvent the wheel because the scheme does not exactly fit.

I would like to see how this could be extended onto. I think it may be possible to also let BlockFamilies handle how a block are rendered, but this is out of the scope of this pull request.

### How to test

I'm not sure about the injection portion of this pull request in regards to families. There are probably something that are not accessible from the injection helper when the families are registered.


GooeyHub:
Build started sha1 is merged.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

pollend:
Here is an example use case that takes advantage of the tweaks I made to block families.

https://github.com/pollend/Signalling/commit/b348c98a8064731cefad92d91a91f4e68b66c09a

GooeyHub:

engine/src/main/java/org/terasology/world/block/family/AbstractBlockFamily.java
```diff
@@ -29,13 +32,23 @@
...
+    public AbstractBlockFamily(BlockFamilyDefinition blockFamilyDefinition, BlockBuilderHelper blockBuilderHelper) {
+    }
+
+    protected void setCategory(Iterable<String> categories) {
         for (String category : categories) {
             this.categories.add(category.toLowerCase(Locale.ENGLISH));
         }
     }
 
+    protected void setBlockUri(BlockUri uri) {
```
kaen: Seems odd that this property is mutable now, but I assume it's a plumbing thing

engine/src/main/java/org/terasology/world/block/family/AttachedToSurfaceFamily.java
```diff
@@ -15,47 +15,85 @@
...
-     */
-    public AttachedToSurfaceFamily(BlockUri uri, Map<Side, Block> blocks, Iterable<String> categories) {
-        super(uri, categories);
+    public AttachedToSurfaceFamily(BlockFamilyDefinition definition, BlockShape shape, BlockBuilderHelper blockBuilder) {
+        super(definition, shape, blockBuilder);
+        throw new UnsupportedOperationException("Freeform blocks not supported");
+    }
+
+    public AttachedToSurfaceFamily(BlockFamilyDefinition definition, BlockBuilderHelper blockBuilder) {
+        super(definition, blockBuilder);
```
kaen: Calls super but that method is empty in the base class

engine/src/main/java/org/terasology/world/block/family/AbstractBlockFamily.java
```diff
@@ -29,13 +32,23 @@
     private BlockUri uri;
     private Set<String> categories = Sets.newHashSet();
 
-    public AbstractBlockFamily(BlockUri uri, Iterable<String> categories) {
-        this.uri = uri;
+    public AbstractBlockFamily(BlockFamilyDefinition definition, BlockShape shape, BlockBuilderHelper blockBuilder) {
```
kaen: seems like these constructors should set `protected` properties on the base class. They're called via `super` below.

engine/src/main/java/org/terasology/world/block/family/AttachedToSurfaceFamily.java
```diff
@@ -15,47 +15,85 @@
...
 
-    /**
-     * @param uri    The uri for the block group.
-     * @param blocks The set of blocks that make up the group. Front, Back, Left and Right must be provided - the rest is ignored.
-     */
-    public AttachedToSurfaceFamily(BlockUri uri, Map<Side, Block> blocks, Iterable<String> categories) {
-        super(uri, categories);
+    public AttachedToSurfaceFamily(BlockFamilyDefinition definition, BlockShape shape, BlockBuilderHelper blockBuilder) {
+        super(definition, shape, blockBuilder);
+        throw new UnsupportedOperationException("Freeform blocks not supported");
```
kaen: Could you add a comment or reword this exception message to explain how a module developer could fix it? I'm not sure what would make this occur and would be very confused to receive this as written.

engine/src/main/java/org/terasology/world/block/family/AttachedToSurfaceFamily.java
```diff
@@ -15,47 +15,85 @@
...
                 this.blocks.put(side, block);
                 block.setBlockFamily(this);
-                block.setUri(new BlockUri(uri, new Name(side.name())));
+                block.setUri(new BlockUri(definition.getUrn(), new Name(side.name())));
             }
         }
         if (this.blocks.containsKey(Side.TOP)) {
             archetype = this.blocks.get(Side.TOP);
         } else {
             archetype = this.blocks.get(Side.FRONT);
```
kaen: not your code but `this` technically isn't required

engine/src/main/java/org/terasology/world/block/family/AttachedToSurfaceFamily.java
```diff
@@ -15,47 +15,85 @@
...
         }
+        this.setBlockUri(new BlockUri(definition.getUrn()));
```
kaen: shouldn't need `this` for method invocations

engine/src/main/java/org/terasology/engine/bootstrap/EnvironmentSwitchHandler.java
```diff
@@ -171,23 +168,17 @@ private void unregisterPrefabFormats(ModuleAwareAssetTypeManager assetTypeManage
 
     private static void loadFamilies(DefaultBlockFamilyFactoryRegistry registry, ModuleEnvironment environment) {
         registry.clear();
-        for (Class<?> blockFamilyFactory : environment.getTypesAnnotatedWith(RegisterBlockFamilyFactory.class)) {
-            if (!BlockFamilyFactory.class.isAssignableFrom(blockFamilyFactory)) {
-                logger.error("Cannot load {}, must be a subclass of BlockFamilyFactory", blockFamilyFactory.getSimpleName());
+        for(Class<?> blockFamily : environment.getTypesAnnotatedWith(RegisterBlockFamily.class)) {
+            if (!AbstractBlockFamily.class.isAssignableFrom(blockFamily)) {
+                logger.error("Cannot load {}, must be a subclass of BlockFamilyFactory", blockFamily.getSimpleName());
```
kaen: > must be a subclass of `BlockFamilyFactory`

should be 

> must be a subclass of `AbstractBlockFamily`

right?

engine/src/main/java/org/terasology/world/block/family/DefaultBlockFamilyFactoryRegistry.java
```diff
@@ -16,27 +16,125 @@
...
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Optional;
 
-public class DefaultBlockFamilyFactoryRegistry implements BlockFamilyFactoryRegistry {
-    private Map<String, BlockFamilyFactory> registryMap = Maps.newHashMap();
-    private BlockFamilyFactory defaultBlockFamilyFactory = new SymmetricBlockFamilyFactory();
+public class DefaultBlockFamilyFactoryRegistry implements BlockFamilyRegistry {
```
kaen: Think you can remove the `Factory` right?

engine/src/main/java/org/terasology/world/block/family/DefaultBlockFamilyFactoryRegistry.java
```diff
@@ -16,27 +16,125 @@
...
+        } catch (Exception e) {
+            logger.error("Failed to load blockFamily {}", blockFamily, e);
+            e.printStackTrace();
+        }
+        return null;
+    }
+
+    public static BlockFamily createFamily(Class<? extends AbstractBlockFamily> blockFamily, BlockFamilyDefinition blockFamilyDefinition, BlockShape shape, BlockBuilderHelper blockBuilderHelper) {
+        try {
+            SimpleClassFactory simpleClassFactory = new SimpleClassFactory(new ParameterProvider() {
```
kaen: This anonymous class smells funny. I think you could refactor this into a new class (with its own file) and take the parameters in this method (bound here by closure...) in a constructor and clean this up a bit.

engine/src/main/java/org/terasology/world/block/family/FreeformFamily.java
```diff
@@ -16,27 +16,118 @@
...
+    private Map<Side, Block> blocks = Maps.newEnumMap(Side.class);
+    private Block block = null;
+
+    public FreeformFamily(BlockFamilyDefinition definition, BlockShape shape, BlockBuilderHelper blockBuilder) {
+        super(definition, shape, blockBuilder);
+        BlockUri uri;
+        if (CUBE_SHAPE_URN.equals(shape.getUrn())) {
+            uri = new BlockUri(definition.getUrn());
+        } else {
+            uri = new BlockUri(definition.getUrn(), shape.getUrn());
```
kaen: Looks like you can get rid of this conditional and maybe even the constant above if you just call this signature in every case.

Unless there's a semantic difference in the two signatures?

engine/src/main/java/org/terasology/world/block/family/FreeformFamily.java
```diff
@@ -16,27 +16,118 @@
...
+            }
+        }
+
+        this.setBlockUri(uri);
+        this.setCategory(definition.getCategories());
+    }
+
+    public FreeformFamily(BlockFamilyDefinition blockFamilyDefinition, BlockBuilderHelper blockBuilderHelper) {
+        super(blockFamilyDefinition, blockBuilderHelper);
+        throw new UnsupportedOperationException("Shape expected");
```
kaen: Seems weird to implement this signature and always throw an exception.

engine/src/main/java/org/terasology/world/block/family/FreeformFamily.java
```diff
@@ -16,27 +16,118 @@
...
+
+    @Override
+    public Block getBlockFor(BlockUri blockUri) {
+        if (block == null) {
+            if (getURI().equals(blockUri.getFamilyUri())) {
+                try {
+                    Side side = Side.valueOf(blockUri.getIdentifier().toString().toUpperCase(Locale.ENGLISH));
+                    return blocks.get(side);
+                } catch (IllegalArgumentException e) {
+                    return null;
```
kaen: definitely need to log here at a minimum

engine/src/main/java/org/terasology/world/block/family/HorizontalBlockFamily.java
```diff
@@ -15,52 +15,85 @@
...
-    public HorizontalBlockFamily(BlockUri uri, Map<Side, Block> blocks, Iterable<String> categories) {
-        this(uri, Side.FRONT, blocks, categories);
+
+    public HorizontalBlockFamily(BlockFamilyDefinition definition, BlockShape shape, BlockBuilderHelper blockBuilder) {
+        super(definition, shape, blockBuilder);
+        BlockUri uri = null;
+        if (CUBE_SHAPE_URN.equals(shape.getUrn())) {
+            uri = new BlockUri(definition.getUrn());
+        } else {
+            uri = new BlockUri(definition.getUrn(), shape.getUrn());
```
kaen: same as above (hope I'm not missing something here)

engine/src/main/java/org/terasology/world/block/family/MultiConnectBlockFamily.java
```diff
@@ -175,9 +179,53 @@ private Rotation getRotationToAchieve(byte source, byte target) {
...
+    }
+
+    @Override
+    public Block getBlockFor(BlockUri blockUri) {
+        if (getURI().equals(blockUri.getFamilyUri())) {
+            try {
+                byte connections = Byte.parseByte(blockUri.getIdentifier().toString().toLowerCase(Locale.ENGLISH));
+                return blocks.get(connections);
+            } catch (IllegalArgumentException e) {
+                return null;
```
kaen: please log (throwing seems better but definitely don't silently eat the exception)

engine/src/main/java/org/terasology/world/block/family/SymmetricFamily.java
```diff
@@ -20,32 +20,52 @@
...
-    public SymmetricFamily(BlockUri uri, Block block) {
-        this(uri, block, Collections.<String>emptyList());
+    public SymmetricFamily(BlockFamilyDefinition definition, BlockShape shape, BlockBuilderHelper blockBuilder) {
+        super(definition, shape, blockBuilder);
+        block = blockBuilder.constructSimpleBlock(definition, shape);
+        BlockUri uri;
+        if (CUBE_SHAPE_URN.equals(shape.getUrn())) {
+            uri = new BlockUri(definition.getUrn());
+        } else {
+            uri = new BlockUri(definition.getUrn(), shape.getUrn());
```
kaen: as above

engine/src/main/java/org/terasology/world/block/internal/BlockManagerImpl.java
```diff
@@ -252,7 +254,12 @@ public BlockFamily getBlockFamily(BlockUri uri) {
                         block.setId(getNextId());
                     }
                     registerFamily(newFamily.get());
-                } finally {
+                }
+                catch (Exception ex){
+                    logger.error(ex.toString());
```
kaen: :+1:

engine/src/main/java/org/terasology/world/block/loader/BlockFamilyDefinitionFormat.java
```diff
@@ -70,22 +65,21 @@
...
         this.assetManager = assetManager;
+        this.blockFamilyRegistry = blockFamilyRegistry;
         gson = new GsonBuilder()
                 .registerTypeAdapterFactory(new CaseInsensitiveEnumTypeAdapterFactory())
                 .registerTypeAdapterFactory(new AssetTypeAdapterFactory(assetManager))
                 .registerTypeAdapter(BlockFamilyDefinitionData.class, new BlockFamilyDefinitionDataHandler())
                 .registerTypeAdapter(Vector3f.class, new Vector3fTypeAdapter())
                 .registerTypeAdapter(Vector4f.class, new Vector4fTypeAdapter())
-                .registerTypeAdapter(BlockFamilyFactory.class, new BlockFamilyFactoryHandler(blockFamilyFactoryRegistry))
+                .registerTypeAdapter(Class.class, new BlockFamilyHandler(blockFamilyRegistry))
```
kaen: This is throwing a big red flag for me -- is there a way to limit it to subtypes of a certain class? I guess it's not so bad if it's limited to  just block family definitions

engine/src/main/java/org/terasology/world/block/loader/BlockFamilyDefinitionFormat.java
```diff
@@ -141,16 +135,18 @@ public BlockFamilyDefinitionData deserialize(JsonElement json, Type typeOfT, Jso
             // Deserialize everything
             BlockFamilyDefinitionData result = new BlockFamilyDefinitionData(base);
             setBoolean(result::setTemplate, jsonObject, "template");
-            setObject(result::setFamilyFactory, jsonObject, "rotation", BlockFamilyFactory.class, context);
+//            setObject(result::setFamilyFactory, jsonObject, "rotation", BlockFamilyFactory.class, context);
```
kaen: dead

engine/src/main/java/org/terasology/world/block/loader/BlockFamilyDefinitionFormat.java
```diff
@@ -309,8 +304,8 @@ private BlockFamilyDefinitionData createBaseData(JsonObject jsonObject) {
                 Optional<BlockFamilyDefinition> baseDef = assetManager.getAsset(basedOn.getAsString(), BlockFamilyDefinition.class);
                 if (baseDef.isPresent()) {
                     BlockFamilyDefinitionData data = baseDef.get().getData();
-                    if (data.getFamilyFactory() instanceof FreeformBlockFamilyFactory) {
-                        data.setFamilyFactory(null);
+                    if (data.getBlockFamily() == FreeformFamily.class) {
```
kaen: This doesn't immediately make sense to me but maybe it would if you were debugging the code?

modules/Core/src/main/java/org/terasology/core/logic/door/DoorSystem.java
```diff
@@ -123,8 +123,8 @@ public void placeDoor(ActivateEvent event, EntityRef entity) {
             closedSide = attachSide.yawClockwise(1);
         }
 
-        Block newBottomBlock = door.bottomBlockFamily.getBlockForPlacement(worldProvider, blockEntityRegistry, bottomBlockPos, closedSide, Side.TOP);
-        Block newTopBlock = door.topBlockFamily.getBlockForPlacement(worldProvider, blockEntityRegistry, bottomBlockPos, closedSide, Side.TOP);
```
kaen: Odd that both lines use Side.TOP here, but I see it was like that before.

engine/src/main/java/org/terasology/engine/bootstrap/EnvironmentSwitchHandler.java
```diff
@@ -171,23 +168,17 @@ private void unregisterPrefabFormats(ModuleAwareAssetTypeManager assetTypeManage
...
                 continue;
             }
-
-            RegisterBlockFamilyFactory registerInfo = blockFamilyFactory.getAnnotation(RegisterBlockFamilyFactory.class);
+            RegisterBlockFamily registerInfo = blockFamily.getAnnotation(RegisterBlockFamily.class);
```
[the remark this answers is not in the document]
pollend: make sense when you look at how its bound to the family. 
```
@RegisterBlockFamily("freeform")
@FreeFormSupported(true)
public class FreeformFamily extends AbstractBlockFamily {
...
}
```

engine/src/main/java/org/terasology/engine/modes/loadProcesses/RegisterBlockFamilies.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public String getMessage() {
+        return "Registering Block Families ...";
+    }
+
+    @Override
+    public boolean step() {
+        BlockManager blockManager = context.get(BlockManager.class);
+        for(BlockFamily blockFamily : blockManager.listRegisteredBlockFamilies()) {
+            InjectionHelper.inject(blockFamily, context);
```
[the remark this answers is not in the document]
pollend: All this does is inject the dependencies into the family. you probably wouldn't even see any of the messages as the dependencies get injected. I can't imagine each iteration taking more then a couple ms.

engine/src/main/java/org/terasology/world/block/loader/BlockFamilyDefinitionFormat.java
pollend: Its based on changes from the old implementation. It does not make immediate sense to me. ummm

GooeyHub:

engine/src/main/java/org/terasology/engine/modes/loadProcesses/RegisterBlockFamilies.java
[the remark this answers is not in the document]
pollend: might be a way to move it forward so block families get bound after systems get instantiated. :?

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/engine/modes/loadProcesses/RegisterBlocks.java
```diff
@@ -27,6 +27,7 @@
 import org.terasology.world.block.Block;
 import org.terasology.world.block.BlockManager;
 import org.terasology.world.block.family.BlockFamily;
+import org.terasology.world.block.family.BlockFamilyRegistry;
```
skaldarnar: what is this import for (no related code changes)?

engine/src/main/java/org/terasology/world/block/BlockExplorer.java
```diff
@@ -16,6 +16,7 @@
 package org.terasology.world.block;
 
 import org.terasology.assets.management.AssetManager;
+import org.terasology.world.block.family.BlockFamilyRegistry;
```
skaldarnar: another stray import?

engine/src/main/java/org/terasology/world/block/entity/BlockCommands.java
```diff
@@ -49,6 +49,7 @@
 import org.terasology.world.block.BlockManager;
 import org.terasology.world.block.BlockUri;
 import org.terasology.world.block.family.BlockFamily;
+import org.terasology.world.block.family.BlockFamilyRegistry;
```
skaldarnar: .. and another one?

engine/src/main/java/org/terasology/world/block/family/AbstractBlockFamily.java
skaldarnar: ❓

engine/src/main/java/org/terasology/world/block/family/AttachedToSurfaceFamily.java
```diff
@@ -15,47 +15,85 @@
...
+import org.terasology.world.block.shapes.BlockShape;
 
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 
 /**
  */
+@RegisterBlockFamily("attachedToSurface")
+@BlockSections({"front", "left", "right", "back", "top", "bottom"})
```
skaldarnar: are block sections generic or what is the reason to handle them via plain strings?

engine/src/main/java/org/terasology/world/block/family/BlockFamily.java
```diff
@@ -15,12 +15,20 @@
...
 import org.terasology.world.block.BlockUri;
+import org.terasology.world.block.loader.BlockFamilyDefinition;
+import org.terasology.world.block.shapes.BlockShape;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
 
 /**
  * A collection of blocks that are all different rotations of the same core block.
```
skaldarnar: Are block families limited to block rotations or do/can they also include form variations (based on neighbors)?

engine/src/main/java/org/terasology/world/block/family/MultiSection.java
```diff
@@ -16,32 +16,19 @@
 package org.terasology.world.block.family;
 
 import com.google.common.collect.ImmutableSet;
+import org.terasology.input.DefaultBindings;
 import org.terasology.util.Varargs;
 
+import java.lang.annotation.*;
```
skaldarnar: mind your star imports 😉

engine/src/main/java/org/terasology/world/block/family/UpdatesWithNeighboursFamily.java
```diff
@@ -15,78 +15,15 @@
...
-    }
-
-    @Override
-    public Iterable<Block> getBlocks() {
-        return blocks.valueCollection();
-    }
+/**
+ * Interface for Block family that gets updated by a change in a neighbor block.
+ */
+public interface UpdatesWithNeighboursFamily {
```
skaldarnar: Should this interface already extend `BlockFamily`. We don't have an implementation here, but I think we can be pretty sure that `UpdatesWithNeighborsFamily` will always be a `BlockFamily` as well.

engine/src/main/java/org/terasology/world/block/internal/BlockManagerImpl.java
```diff
@@ -252,7 +254,11 @@ public BlockFamily getBlockFamily(BlockUri uri) {
                         block.setId(getNextId());
                     }
                     registerFamily(newFamily.get());
-                } finally {
+                }
+                catch (Exception ex){
```
skaldarnar: What exceptions can be thrown? Can we be more specific here?

pollend: I'll add a comment, but didn't like how the old implementation just skipped the family

engine/src/main/java/org/terasology/world/block/family/AttachedToSurfaceFamily.java
pollend: you can define your own sections in the family. This is carried over from the old implementation that used a factory.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/engine/bootstrap/EnvironmentSwitchHandler.java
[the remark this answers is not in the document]
mjuvekar7: I agree with @kaen. `RegisterBlockFamily` sounds more like a function name than a type name. `BlockFamilyRegistryHelper` would be much better for readability.

engine/src/main/java/org/terasology/world/block/family/AttachedToSurfaceFamily.java
mjuvekar7: Reword the exception. Be a bit more verbose.

engine/src/main/java/org/terasology/world/block/internal/BlockManagerImpl.java
mjuvekar7: Would it be possible to catch a more specific exception type rather than `Exception`?

pollend: the block family implementation from this point is unknown. It can basically fail in a lot of unexpected ways that are unforeseeable. since the family type can also be implemented on the module end.

GooeyHub:

pollend:
#2907 will need to reworked

GooeyHub:
Build triggered. sha1 is original commit.

GooeyHub:
Build started sha1 is original commit.

GooeyHub:
Build finished. 

GooeyHub:

GooeyHub:
Build triggered. sha1 is original commit.

GooeyHub:
Build started sha1 is original commit.

GooeyHub:
Build finished. 

GooeyHub:

GooeyHub:
Build triggered. sha1 is merged.

GooeyHub:
Build started sha1 is merged.

GooeyHub:
Build finished. 

GooeyHub:

Cervator:
Changed this to target its own dedicated branch and will merge it there for usage during GCI :-)

Tasks may target that branch for additional improvements, and eventually we'll merge it into `develop` or perhaps `v2.0.0` (target for merge shortly after GCI)

GooeyHub:

PR #3086, opened 2017-08-30
Prevent chests from losing items.

### LINKED ISSUE #3067: Goodie chest the player starts with in CoreSampleGameplay is empty

Do final testing for Alpha 8 - Core. I placed the chest the player starts with and expected it to contain stuff

### What actually happened

No stuff!

Probably a simple issue breaking how the starting player inventory is set up. See `PlayerStartingInventorySystem` and maybe the `InventoryAuthoritySystem` - might even be bite-sized for somebody able to step through code to see where the items get lost (probably set a breakpoint in `onPlayerSpawnedEvent` in `PlayerStartingInventorySystem`

### How to reproduce

* Start the game with CoreSampleGameplay selected and any world
* Place the chest you start with in your toolbar into the world
* Open the chest with `e` - no stuff!

High priority since that chest contains all kinds of useful goodies to help demo the game.

portokaliu:
Fixes  #3067 by preventing the held item component deletion. The mesh is removed when held but otherwise it is untouched. This is a temporary fix while a better solution will be underway.

How to test:
Plonk down the starting chest, you'll see that it now has items inside. 

I've tested it thoroughly before judging it as an ok fix, but I might miss something, so a secondary test would be neat to have before merging :)

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
@@ -39,6 +39,7 @@ import org.terasology.network.ClientComponent;
 import org.terasology.network.NetworkComponent;
 import org.terasology.physics.components.RigidBodyComponent;
 import org.terasology.registry.In;
+import org.terasology.rendering.logic.MeshComponent;
 import org.terasology.rendering.world.WorldRenderer;
 
 @RegisterSystem(RegisterMode.CLIENT)
@@ -155,10 +156,8 @@ public class FirstPersonClientSystem extends BaseComponentSystem implements Upda
             if (mountPointComponent != null) {
 
                 if (clientHeldItem.exists()) {
-                    clientHeldItem.destroy();
-                    clientHeldItem = EntityRef.NULL;
+                    clientHeldItem.removeComponent(MeshComponent.class);
                 }
-
 
                 // remove the location from the old item
                 if (oldItem != null && oldItem.exists()) {


GooeyHub:

Cervator:
Tested it out, found a couple sneaky issues:

1) Probably the reason the entire entity was previously destroyed is from not knowing what other interesting components could be active - with this change if you switch to a torch then something else the `LightComponent` from the torch remains active :D
2) There still seems to be a case where with two regular clients (no headless) player connected together the empty chest can still appear - I saw it in one test as player 2 seeing player 1's placed goodie chest as empty, although surprisingly the *inner* goodie chest also placed by player 1 was fine

I'm not worried very much about case number 2. The first issue can be resolved simply by adding `clientHeldItem.removeComponent(LightComponent.class);` as well, but I'm not sure what other sneaky components we might need to look for. Might be a good enough fix till we can rework this better :-)

Dunno what your ideas are for going forward, but the current code does seem awkward. It edits the sort of locally persisted held entity both before and after copying it in fresh from the newly selected held item. Fun things happen if you start messing with it though, like removing the pre-copy block entirely :D

2017-08-31 00_35_01-terasology - c__dev_terasology_git_terasology - engine - _engine_src_main

I'd think we'd just want to split that apart further to make it cleaner. Completely separate the held item from the selected item's entity - which is what I thought simply re-copying it would do. But nope that just made other fun things happen. I guess we need a copy but strictly with only the components needed to draw it on the local client - as for other clients in multiplayer maybe they just need a copy synchronized over when it comes time to show other people's held items (which isn't quite yet)

Anyway, you probably have had all those thoughts already. Check out the lighting thing and see if you can find any other creative components. Probably we can merge this with just that one additional component removal added in :-)

Cervator:
Bump. Out of 3 current blockers one got fixed, one has another fix ready, and this is the third, nearly fixed. So close! :-)

Cervator:
Went ahead and applied the minimal `LightComponent` fix on top and merged this. Couldn't replicate my second issue. So this might be good and properly fixed - at least before any sort of overhaul makes it a moot point anyway :-)

We should keep an eye on any quirky behavior around switching held items in the meantime, especially for unusual items that may have components making some sort of impact while held.

PR #3089, opened 2017-08-31
Clear ComponentSystem subscribers on shutdown - fixes #3087

### LINKED ISSUE #3087: Game crash on second world load due to odd particle initialization / lifecycle
Noticed by @portokaliu - after you leave the game world once `ParticleSystemManagerImpl.shutdown()` executes  `particleUpdater = null;` probably in an attempt to clean something up. However the `initialise()` method that sets it only runs on *game* start, not entering a new world. So on attempt two `particleUpdater` remains null and NPEs the game.

Blocker level critical for Alpha 8 but probably relatively straight forward fix: _don't do that_ - question is whether we should not null it out, instead favoring a better cleanup. Or if we just want a hack fix then add null checks and reset the variable to a valid new updater.

Needs a bit of poking around to see what the better option is. Submitting this so we don't forget. Pinging @MaxBorsch if interested (also: argh why doesn't your name auto-complete in a ping, I don't understaaaaand)

Unsure if we have a lifecycle step on entering a world to initialize this thing, or if we need one. Or if the particle system just needs a bit more tweaking - probably :-)

oniatus:
Fixes #3087 
See https://github.com/MovingBlocks/Terasology/issues/3087#issuecomment-326409756 for a detailed explanation.

->100% TDD 😎 *wohoo*

### DIFF
--- a/engine-tests/src/test/java/org/terasology/engine/ComponentSystemManagerTest.java
+++ b/engine-tests/src/test/java/org/terasology/engine/ComponentSystemManagerTest.java
@@ -1,0 +1,70 @@
+
+package org.terasology.engine;
+
+import com.google.common.collect.Iterables;
+import org.junit.Before;
+import org.junit.Test;
+import org.terasology.context.Context;
+import org.terasology.entitySystem.entity.EntityManager;
+import org.terasology.entitySystem.event.internal.EventSystem;
+import org.terasology.entitySystem.systems.RenderSystem;
+import org.terasology.entitySystem.systems.UpdateSubscriberSystem;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class ComponentSystemManagerTest {
+
+    private ComponentSystemManager sut;
+
+    @Before
+    public void setUp() {
+        Context context = mock(Context.class);
+        EntityManager entityManager = mock(EntityManager.class);
+        when(entityManager.getEventSystem()).thenReturn(mock(EventSystem.class));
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine-tests/src/test/java/org/terasology/engine/ComponentSystemManagerTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.terasology.entitySystem.systems.UpdateSubscriberSystem;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class ComponentSystemManagerTest {
+
+    private ComponentSystemManager sut;
```
emanuele3d: sut? System Under Testing? I (and most coding academics I imagine) advise against all but the most well known acronyms. I.e. NASA might be appropriate in some contexts, noDISBtET (1) is much less appropriate.

(1) obviously: number of Dogs I Scratched Behind the Ears Today

Cervator:
Tested it out and clarified the naming real quick since it is a tiny change - hopefully `systemUnderTest` is a valid option @emanuele3d & @oniatus? Let me know if not and we can change it :-)

Thanks for fix and review!

oniatus:
Welcome to conventions and habits from different projects 😅 and yeah, the intention is http://xunitpatterns.com/SUT.html
The advantage of such a name besides `componentSystemManager` is that it is refactoring robust against the classname + it is short in tests and makes clear which field is tested and which field is a dependent component.
However, we have zero occurances of such names and if it is unclear -> rename 😄  `systemUnderTest` or even `componentSystemManager` is totally fine for me.

emanuele3d:
systemUnderTest is good for me!

PR #3090, opened 2017-09-01
Somewhat hacky fix for #3066 - validate the particle prefab being available

### LINKED ISSUE #3066: Breaking blocks in SampleBuilderGameplay doesn't work right

Test Alpha 8 - SampleBuilderGameplay world

* Left click should remove a block without any drops and "select" that block type
* Right click should place a block of whatever selected type

### What actually happened

Right click is fine. But Left click fails to remove a block and instead a physics-cube pops out and flies off to the side, one per click.

Probably some change to how block drops are handled in the engine caused this?

### How to reproduce

* Run the game and pick BuilderSampleGameplay plus its world
* Try to break any block (left click)
* See the block not disappear, yet a new physics-enabled cube pops out and bounces around

High priority issue since it effectively breaks the BSG template.

Cervator:
### Contains

Fixes #3066 by adding a check to catch situations in which the particle prefab doesn't exist, such as when the Core module isn't enabled

### How to test

Run BuilderSampleGameplay. Break a block. Should not get a `NullPointerException` and the block should disappear.

Admittedly I think behavior a while ago didn't even include a physics cube dropping - just the active block you place changes to the last you broke.

Nothing really. There are better fixes, such as grabbing the PrefabManager to see if the prefab exists before trying to use it, but at that point you might get tempted to try to fix it deeper. Should we cache the dust prefab in the system so we don't need to find it every time? Would that cause trouble if the player modifies the prefab on disk and wants a hot reload? Etc.

Ultimate fix would be better packaging / organizing of the particle system so it doesn't have a "soft" dependency on Core despite being inside the engine.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
@@ -159,8 +159,11 @@ public class BlockEntitySystem extends BaseComponentSystem {
             // dust particle effect
             if (entity.hasComponent(LocationComponent.class) && block.isDebrisOnDestroy()) {
                 EntityBuilder dustBuilder = entityManager.newBuilder("core:dustEffect");
-                dustBuilder.getComponent(LocationComponent.class).setWorldPosition(entity.getComponent(LocationComponent.class).getWorldPosition());
-                dustBuilder.build();
+                // TODO: particle system stuff should be split out better - this is effectively a stealth dependency on Core from the engine
+                if (dustBuilder.hasComponent(LocationComponent.class)) {
+                    dustBuilder.getComponent(LocationComponent.class).setWorldPosition(entity.getComponent(LocationComponent.class).getWorldPosition());
+                    dustBuilder.build();
+                }
             }
 
             // sound to play for destroyed block


GooeyHub:

MaxBorsch:
@Cervator Like always sorry for my unparalleled inactivity, school resumed and it has thus far been a grind. I see nothing wrong with this as a temporary fix, however as you mentioned, the problem is more the structure of the block effects. The question I'm thinking of is:
1. Should the engine ever have dependencies on core? I'm leaning towards no.
2. I say no because people will want different particle effects onBlockBreak, so it would be better to make an event / api for a block break event (if it isn't made already), and then **in core** use that event to spawn the dust particles. The block particles (not dust) are a bit more engine-related so those should stay as part of the engine and we can just add properties to blocks that configure them.

pollend:
Is this for when the core module isn't enabled. I.e when attempting to break a block but the dust particle does not exist since the core module isn't loaded.

Cervator:
Correct and correct. We want to move some of this stuff out of the engine, leaving just whatever may be required to stay (for instance due to some rendering features not being exposed to modules yet). No dependency on Core and Core or any alternative module is responsible for hooking up the particles :-)

@MaxBorsch np on delays, glad to just hear from you! We'll be here whenever you have spare time.

Going to merge this for now since Alpha 8.

PR #3098, opened 2017-09-03
Permission level displayed for a command on using help

### LINKED ISSUE #3094: The console command "help" should also show the permission required to execute it
So for instance the `give` command requires `PermissionManager.CHEAT_PERMISSION` but current output from "help give" only describes it as such:

```
=====================================================================================================================
give <String prefabId or blockName> (Integer amount) (String blockShapeName)
=====================================================================================================================
Puts the desired number of the given item or block with the given shape into your inventory
=====================================================================================================================
```

We should add in the permission level somewhere. See `ItemCommands.give()` as an example. The "help" command is over in `CoreCommands`

Trickiest part might be figuring out how to query for the right annotation but that should be doable with a quick bit of research I hope :-)

TheFlash98:
### Contains

Fixes #3094 

### How to test

Open the console and type `help give` It will display `Required Permission Level - cheat`

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -601,6 +601,12 @@ public class CoreCommands extends BaseComponentSystem {
                     msg.append("=====================================================================================================================");
                     msg.append(Console.NEW_LINE);
                 }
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Oh hey that was easier than I had hoped - thanks! :D

@GooeyHub: add to whitelist

GooeyHub:

PR #3107, opened 2017-09-23
Add jdk.internal.reflect to ExternalAPIWhitelist for Java 9 compatibility (fixes #3106)

### LINKED ISSUE #3106: Compatibility issue with Java 9
Terasology crashes when trying to create a new game giving [this error](https://github.com/MovingBlocks/Terasology/files/1326193/terasology.log). 
After discussing it with @Cervator on IRC he found out the crash is related to the Java version installed on my pc.
```
$ java -version
openjdk version "9"
OpenJDK Runtime Environment (build 9+181)
OpenJDK 64-Bit Server VM (build 9+181, mixed mode)
```
Java 9 is now in [general availability](http://openjdk.java.net/projects/jdk9/) and has already landed in the [Arch Linux repository](https://www.archlinux.org/packages/extra/x86_64/jdk9-openjdk/).

gianluca-nitti:
### Contains

Adds the `jdk.internal.reflect` to the API whitelist for modules to fix the compatibility issue with Java 9 as reported in #3106. I tested with OpenJDK 9 and appears to work correctly.

As you can see in the crash log reported by @Dariost in the issue, right before throwing the exception there is a line which says `02:40:14.218 [main] ERROR o.t.module.sandbox.ModuleClassLoader - Denied access to class (not allowed with this module's permissions): jdk.internal.reflect.ConstructorAccessorImpl`. So in my understanding there are modules which use code that relies on reflection, and the new JVM implementation uses additional internal classes for reflection and the custom filtering classloader denies access to said classes from sandboxed modules. So "unlocking" the mentioned package seems to solve the issue.

### How to test

Run the game with a Java 9 JVM and create a new game with Core Gameplay or Josharias Survival gameplay model and start it. On this branch should load and work correctly without crashing as in the current `master`/`develop` versions.

It probably is a good idea to find some documentation about the added package to make sure it does not 
introduce security issues, i.e. it does not allow modules to break out from the sandbox to access filesystem or doing anything the shouldn't do. Looks like this isn't an easy task since these are probably classes meant for internal use by other JDK parts only, in fact I couldn't find the javadoc (and in fact we aren't using them directly; it's reached through `ReflectionReflectFactory` which uses the `java.lang.reflect.Constructor` class, which in turn, in the new Java 9 implementations, appears to use the discussed package).

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
+++ b/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
@@ -46,6 +46,7 @@ public final class ExternalApiWhitelist {
             .add("java.awt")
             .add("java.awt.geom")
             .add("java.awt.image")
+            .add("jdk.internal.reflect")
             .add("com.google.common.annotations")
             .add("com.google.common.cache")
             .add("com.google.common.collect")


GooeyHub:

Cervator:
Tested Java 9 some with this in place. Found a few things we need to stay on top of when trying to use Java 9 - not related to this PR though :-)

* Need a fairly recent IntelliJ to work with Java 9 or you might get some weird invalid release flag
* Going to a newer IntelliJ will likely trigger #3113 - we need to update to newer Checkstyle
* The CrashReporter needs additional work for Java 9 - see https://github.com/MovingBlocks/CrashReporter/issues/41

None of that blocks this PR, so I'm merging it - thanks! :-)

As for the security bit we have an outstanding issue to review the whitelist anyway so we'll just roll that into there

PR #3111, opened 2017-09-24
Update menu_ar.lang

Omaranwa:
arabic update thats all :)


GooeyHub:

jellysnake:
As per #2677 we don't have support for displaying the Arabic language along with others, due to not having the fonts yet. As such Arabic is currently disabled.

I'm thinking that this pull should be merged anyway (Everything is in order) and hopefully we can fix the issue with displaying them.
Pinging some folks.
@Cervator @msteiger @MarkDerosier

Omaranwa:
blz marge this

Cervator:
Hi @Omaranwa thanks for the PR :-)

I don't mind merging this in advance of the later font changes we need to actually be able to display it in-game. But how come you also changed the flag?

We had some trouble and discussion coming up with an ideal flag for a language used across a great many states each with their own flag(s). In the end it seemed a plain green flag might be the safest and most generic option. What's your reasoning behind picking a specific flag, and which is it meant to be?

From looking real quick it might be Yemen, Iraq, or Egypt?

Omaranwa:
Its egypt flag as my translations in arabic egypt format.

jellysnake:
So in the interest of trying to keep things neutral I think we should either revert the flag back to the generic option or rename the language to Arabic (Egyptian) to make it clear. @Cervator what do you think?

Omaranwa:
Rename it to  Arabic (Egypt) >

Cervator:
I'm fine with renaming, am curious how other projects that support Arabian dialects do it (how much do the dialects vary? Like Spanish vs Portuguese or so?). I simply have no knowledge of the topic so it would be good to know more about best practices and approaches :-)

Omaranwa:
sample project is luckey patcher wtich have a 3 arabic formats (arabic eg. arabic ksa arabic normal)

jellysnake:
Okay, so it appears we are going to be renaming it. If you could rename this to "Arabic (Egyptian)" then we should be able to go ahead and merge it.
You'll need to change it in the list of "disabled" languages for now until we fix fonts.
(Or if @Cervator is fine with it we can add it to the list after merging this)

Omaranwa:
i donot know how to renam it
can show my how
thanks

Cervator:
Hmm, I think we have another challenge.

So the latest idea here is to split out Arabian dialects, such as Egyptian Arabic. That would mean keeping a language file per dialect. So instead of `menu_ar.lang` we would rename to ... what, exactly?

As per https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes there appears to only be one language code for Arabic - "ar" - since we're using a language code lookup system, I wonder how we'd go about "nesting" one code into dialects.

Sorry for the delays @Omaranwa - this is proving trickier than expected. Your translations are fine and we'll find a way to use them, we just need to figure out how to organize the different languages :-)

Pinging @jellysnake @Halamix2 @msteiger for reference and ideas

Halamix2:
In ISO 639-3 there is `arz` code for Egyptian Arabic (and over 20 codes for various other Arabic languages) but I don't know if we have ISO 639-3 support

msteiger:
I think we need to differentiate between language and countries. Flags are usually for countries, so I picked the country which seemed to be the most natural choice. So I would go for Saudi-Arabia maybe?

The Noto Font supports Arabic, so the chars. are not the problem. When I worked on the implementation, I realized that many chars. are composed of several others (also true for Hindi and a few others), which does not work without some serious effort on the text rendering engine.

Omaranwa:
i think i will revert the flag back to ksa flag

Omaranwa:
hi can you guide me on adding arabic support
thanks

flo:
@Omaranwa Since no one replied yet I assume that there is no one that knows a quick and easy way of adding arabic support. So if you want to have it you would have to investigate for yourself how to do it.

@jellysnake @msteiger @Cervator if you don't object I would suggest we merge this as has been previously proposed.

skaldarnar:
I like this feature of updating someone else's branch 😄 I just changed the language name to "Arabic (Egyptian)" to be clear for future translators (hopefully avoids some conflict/discussion on different variants), and also reduces irritations in-game ;-)

Omaranwa:
ok looks good but can you merge this bull

PR #3118, opened 2017-10-04
Added FCManager and support for FC serialization

vampcat:
Closes issue #2893 

This PR primarily deals with adding a new `FlexibleConfigManager` class, that is responsible for managing all the FCs, as well as saving/loading them.


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -54,4 +57,10 @@
      * Returns a boolean stating whether the config contains a {@link Setting} with the given id.
      */
     boolean contains(SimpleUri id);
+
+    void setUnusedSettings(Map<SimpleUri, String> unusedSettings);
+
+    Map<SimpleUri, String> getUnusedSettings();
+
+    Map<SimpleUri, Setting> getActiveSettings();
```
oniatus: I don't understand the difference between an unused and an active setting from the api, maybe add some javadoc to explain the concept?

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -52,7 +55,14 @@ public boolean add(Setting setting) {
             return false;
         }
 
+        if (unusedSettings != null) {
+            if (unusedSettings.containsKey(id)) {
+                setting.setValueFromString(unusedSettings.remove(id));
+            }
+        }
```
oniatus: This could be an unexpected sideeffect because a caller of `fc.add(setting)` could expect that an add call keeps the value of the setting the same.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -91,4 +101,37 @@ public boolean remove(SimpleUri id) {
...
+                Setting setting = entry.getValue();
+                if (!setting.getValue().equals(setting.getDefaultValue())) {
+                    jsonObject.addProperty(entry.getKey().toString(), setting.getValue().toString());
+                }
+            }
+
+            return jsonObject;
+        }
+    }
+    */
```
oniatus: dead lines, can be removed if not needed.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.config.flexible;
+
+import org.terasology.engine.SimpleUri;
+
+public interface FlexibleConfigManager {
+    void addFlexibleConfig(SimpleUri flexibleConfigUri, FlexibleConfig fc);
```
oniatus: - use flexibleConfig for the second parameter, similar to the first one.
- the interface could define the behavior when two configs with the same uri are added (throw RTE)

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
```diff
@@ -0,0 +1,151 @@
...
+        ensureDirectoryExists(filePath);
+        return filePath;
+    }
+
+    private Gson getGson() {
+        if (gson == null) {
+            gson = new GsonBuilder().setPrettyPrinting().create();
+        }
+
+        return gson;
```
oniatus: I don't think lazy initialization provides a benefit here. Gson is not that expensive to build and also not that expensive to keep.
I would inline the initialization logic to the field and use direct field access instead.

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
```diff
@@ -209,4 +214,19 @@ public String getHumanReadableName() {
...
+    public void setValueFromString(String valueString) {
+        if (value instanceof Integer) {
+            value = (T)(Integer)Integer.parseInt(valueString);
+        } else if (value instanceof Float) {
+            value = (T)(Float)Float.parseFloat(valueString);
+        } else if (value instanceof Double) {
+            value = (T)(Double)Double.parseDouble(valueString);
+        } else if (value instanceof String) {
+            value = (T)valueString;
+        } else {
```
oniatus: I don't like this block for three reasons:
1. It violates OCP because every new type needs a code change
2. The double cast looks odd
3. It supports only int, float, double and string but this is only visible in the code

I don't have a better alternative right now but I am sure we will find one :)

engine/src/main/java/org/terasology/engine/subsystem/common/ConfigurationSubsystem.java
```diff
@@ -46,6 +52,15 @@ public void preInitialise(Context rootContext) {
         config = new Config();
         config.load();
 
+        FlexibleConfigManager flexibleConfigManager = new FlexibleConfigManagerImpl();
+        rootContext.put(FlexibleConfigManager.class, flexibleConfigManager);
+
+        FlexibleConfig renderingFlexibleConfig = new FlexibleConfigImpl();
+        flexibleConfigManager.addFlexibleConfig(new SimpleUri("engine:rendering"), renderingFlexibleConfig);
+
+        flexibleConfigManager.load();
```
oniatus: can't this be part of `config.load()`? We already have an integration to `config.save()`

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
vampcat: Yeah, I don't like this part of the code either - one of the first that I plan on reworking, as soon as i get a bit of free time.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
vampcat: I was planning on doing that after the code stabilizes a bit, but I'll explain some comments explaining the basic logic for reviewers' benifit.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
vampcat: When you say "the config can define the behaviour", you mean I should add a "throws RuntimeException" here?

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
vampcat: I'm not sure how else we can handle the saving/loading part.

We can make it clear in the javadocs that adding a settings will loads the non-default value that was saved earlier, and that if you want the default value, you must call `.resetValue()` right after?

engine/src/main/java/org/terasology/engine/subsystem/common/ConfigurationSubsystem.java
vampcat: One thing we'll have to be careful about is that FCs get initialized -before- config gets loaded. Not sure if this is possible for every single config.

Not a big deal right now, but just wanted to bring it up.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
```diff
@@ -0,0 +1,151 @@
...
+    private void ensureDirectoryExists(Path filePath) {
+        try {
+            Files.createDirectories(filePath.getParent());
+        } catch (Exception e) {
+            throw new RuntimeException("Cannot create directory for FC!");
+        }
+    }
+
+    private Path getPathForFlexibleConfig(SimpleUri flexibleConfigUri) {
+        Path filePath = PathManager.getInstance().getHomePath().resolve("configs").resolve(flexibleConfigUri.getModuleName().toString()).resolve(flexibleConfigUri.getObjectName().toString() + ".cfg");
```
emanuele3d: Wowowowow. This line is too short. Can you make it any longer please? :smile: 

At least let's split it over multiple lines, i.e.

```
...getHomePath.resolve(blah)
              .resolve(blahblah)
              .resolve(blahblahblah)
```

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
emanuele3d: I agree with @oniatus assessment. 

I now have the even stronger feeling we might want to go back to a Setting class that is abstract, -not- generic-based, and is the base for classes such as FloatSetting, IntSetting etc.

I guess one alternative would be delegate the parsing to type-specific parsing objects which can be registered somewhere at runtime and retrieved/used in this method.

Any other possibility?

engine/src/main/java/org/terasology/engine/subsystem/common/ConfigurationSubsystem.java
emanuele3d: I think the aim is to eventually eliminate Config entirely and leave only the FCmanager to fill those shoes. So, I think it makes sense to keep them in parallel for the time being.

engine/src/main/java/org/terasology/engine/subsystem/common/ConfigurationSubsystem.java
```diff
@@ -46,6 +52,15 @@ public void preInitialise(Context rootContext) {
         config = new Config();
         config.load();
 
+        FlexibleConfigManager flexibleConfigManager = new FlexibleConfigManagerImpl();
+        rootContext.put(FlexibleConfigManager.class, flexibleConfigManager);
+
+        FlexibleConfig renderingFlexibleConfig = new FlexibleConfigImpl();
```
emanuele3d: Hmmm. I'm thinking that there might be reasons to split the existing rendering config in two separate configs, one for generic, always needed settings, i.e. resolution, and another for all those settings that kick in only in-game. 

The only problem I see is that this would mean some settings become visible to the user only in-game, which is not quite the typical behavior of settings: you can normally tweak them before you enter the game proper. 

On the other hand, given what @vampcat and I have been discussing about render nodes providing their own settings to the RenderingFC, at this stage we only instantiate those nodes when the renderer is alive, which leads anyway to this behavior.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
vampcat: Okay, so based on a quick tour of the rest of the FC code, the `throws` clause is added in the javadocs everywhere, and nowhere in the declaration.

Adding it in the code here makes things inconsistent - although we can can obviously go back and add it to the code everywhere.
Thoughts?

engine/src/main/java/org/terasology/config/flexible/SettingImpl.java
vampcat: "delegate the parsing to type-specific parsing objects which can be registered somewhere at runtime"
Sounds a lot like gson adapters - can't say I'm the biggest fan of that.

I've implemented a hybrid generic-and-concerete solution which I feel is a decent compromise. Have a look at it.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
```diff
@@ -0,0 +1,151 @@
...
+        for (Entry<SimpleUri, FlexibleConfig> entry : flexibleConfigs.entrySet()) {
+            SimpleUri flexibleConfigUri = entry.getKey();
+            FlexibleConfig flexibleConfig = entry.getValue();
+
+            try (Reader reader = Files.newBufferedReader(getPathForFlexibleConfig(flexibleConfigUri), TerasologyConstants.CHARSET)) {
+                JsonReader jsonReader = new JsonReader(reader);
+
+                Map<SimpleUri, String> unusedSettings = Maps.newHashMap();
+
+                if (jsonReader.hasNext()) {
```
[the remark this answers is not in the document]
vampcat: I was considering an empty file - but to be fair, that's not a valid config file in the first place.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
```diff
@@ -0,0 +1,151 @@
...
+                Map<SimpleUri, String> unusedSettings = flexibleConfig.getUnusedSettings();
+
+                for (Entry<SimpleUri, Setting> activeSetting : settingMap.entrySet()) {
+                    Setting setting = activeSetting.getValue();
+                    if (!setting.getValue().equals(setting.getDefaultValue())) {
+                        jsonObject.addProperty(activeSetting.getKey().toString(), setting.getValue().toString());
+                    }
+                }
+
+                // Add all the non-default settings that were not used in this session
```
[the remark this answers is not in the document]
vampcat: This is like old save files - we can't delete them based on suspicion of being a zombie, since they might very well be a vampire instead.

I think we should leave this untouched, since I'd rather live with 10 extra settings than lose all my settings because I did not load a module for (say) 10 consecutive games.

engine/src/main/java/org/terasology/engine/subsystem/common/ConfigurationSubsystem.java
vampcat: I'm definitely up for the split.

And I do think the behaviour of not showing settings in the main menu is the desired one, at least in the long run.

GooeyHub:

emanuele3d:
@vampcat, please let me know when you think it's time to review the latest changes.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -0,0 +1,30 @@
...
+
+    void removeFlexibleConfig(SimpleUri flexibleConfigUri);
```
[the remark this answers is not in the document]
vampcat: I could have sworn that I replied to this comment.. Ah well.

By a similar logic, every single object in Context is also vulnerable. And if we do decide to go down this path, this will increase the amount of book-keeping that every class using an FC has to do (imagine the engine - it has to store the URIs of the FCs somewhere, right? What if someone uses that map (or list) to retrieve the receipt and then deletes the FC?

I think a better solution would be to restrict deletion (maybe even access?) of FCs to the creating Module. To be fair, I haven't thought about this *too* hard, so I might be missing some obvious things.

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -36,16 +36,8 @@
...
-
-    /**
-     * Creates a new {@link FlexibleConfigImpl} instance.
-     */
-    public FlexibleConfigImpl() {
-        this.settings = Maps.newHashMap();
-        this.temporarilyParkedSettings = Maps.newHashMap();
-    }
+    private Map<SimpleUri, Setting> settings = Maps.newHashMap();
+    private Map<SimpleUri, String> temporarilyParkedSettings = Maps.newHashMap();
```
vampcat: Morale behind shifting this here: When something is initialized like this, it's clear at the first glance that it can never be null. However, if it's initialized in the constructor, that raises the question of whether this can ever be null, and whether checks for the same should be added.

(P.S. I could not find any real reason to prefer one style over the other.)

vampcat:
I have no clue how I broke the build system :sweat:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
emanuele3d: ...hmmm... the code could still assign null (intentionally or by mistake) to such variable. 

It's no big deal though, let's not worry about this anymore.

vampcat: True. I was sorely tempted to make it final as well in the same line, but I refrained.

Not a big deal, but can't deny that I am tempted :D

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
emanuele3d: I guess you can, but I don't think we should encourage anybody to override a user's choice without his or her consent.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
emanuele3d: Sorry for the delay about this.  

I think in this case it is warranted for the interface to show the exception thrown as it is part of the intended behavior of this method AND it diverges from the behavior of a typical map that would just overwrite the previously existing value for a given key.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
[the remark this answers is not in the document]
emanuele3d: It's a fair point you are making about Context objects also being vulnerable. And I don't know how to secure the receipts from reflection if that's what you meant. Is reflection capable of getting absolutely anywhere, including obtaining private variables from objects?

Restricting deletion of FCs might be a better alternative, perhaps with the engine cleaning up by deleting module configs when a game is stopped and the user goes back to the main menu. That been said, some modules, i.e. terrain generators (I think), might be active earlier than a game's startup, and I don't know if there is a way to differentiate between the two. But again, wouldn't reflection make any deletion method ultimately available to any sophisticated enough developer?

Finally, I think restricting access is perhaps overkill. We might theoretically want to restrict a config access so that only dependent modules can access the config of a module they are dependent on. In practice I feel even restricting deletion might be overkill.

So, I'd say let's leave this aspect just for discussion for now. We might revisit it in a later PR.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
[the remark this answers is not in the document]
emanuele3d: Hmmmm. I wonder if only the engine should add settings to an engine config and only modules should add settings to their very own module config. This should eliminate zombie settings as configs are atomically loaded and atomically discarded.

Programmatically this might not have consequences, but when it's time to group settings, potentially from different configs, into a single UI, that's where the uiPath idea would become a necessity. 

What do you think?

engine/src/main/java/org/terasology/engine/subsystem/common/ConfigurationSubsystem.java
emanuele3d: <sigh> I don't see how else we can do it admittedly. But it will be a bit unusual I feel. Perhaps on loading/creating a game we'll need a brief alert, fading in/out near the bottom of the screen, informing the user that additional settings have become available.

emanuele3d:
@Cervator? Can you have a look at the outcome of the build? It doesn't look like something this PR should be capable of generating.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
emanuele3d: Admittedly, why not final?

engine/src/main/java/org/terasology/config/flexible/settings/StringSetting.java
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.config.flexible.settings;
+
+import org.terasology.config.flexible.validators.SettingValueValidator;
+import org.terasology.engine.SimpleUri;
+
+public class StringSetting extends SettingImpl<String> {
```
emanuele3d: I see you made String, Float, Integer and Double settings. I would suggest to add BooleanSetting and perhaps LongSetting. Byte, Char and Short would complete the list but I feel they are probably a bit overkill for our needs. I wouldn't say no if you added them though... :wink:

engine/src/main/java/org/terasology/config/flexible/settings/DoubleSetting.java
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.config.flexible.settings;
+
+import org.terasology.config.flexible.validators.SettingValueValidator;
+import org.terasology.engine.SimpleUri;
+
+public class DoubleSetting extends SettingImpl<Double> {
```
emanuele3d: I'm guessing there is no way to eliminate the SettingImpl<T> class without a lot of almost identical code (save for the type), right?

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -58,9 +60,9 @@
...
     boolean contains(SimpleUri id);
 
-    void setUnusedSettings(Map<SimpleUri, String> unusedSettings);
+    Map<SimpleUri, Setting> getSettings();
 
-    Map<SimpleUri, String> getUnusedSettings();
+    void save(Writer writer);
 
-    Map<SimpleUri, Setting> getActiveSettings();
+    void load(Reader reader);
```
emanuele3d: Are the reader/writer passed here because it is not possible to change their path once they have been instantiated?

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
emanuele3d: My previous comment wasn't quite concrete enough. I'd say: yes document the behaviour in the javadoc, no do not suggest to call resetValue() if necessary - it shouldn't be necessary unless the user wants it.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
```diff
@@ -72,25 +70,9 @@ public void load() {
...
-                    }
-
-                    jsonReader.endObject();
-                }
-
-                flexibleConfig.setUnusedSettings(unusedSettings);
+                flexibleConfig.load(reader);
             } catch (IOException e) {
-                logger.error("Failed to load config file {}, falling back on default config");
+                // TODO: Handle exception
```
emanuele3d: Yes please. :slightly_smiling_face:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -90,29 +98,55 @@ public boolean remove(SimpleUri id) {
...
+
+            while (jsonReader.hasNext()) {
+                SimpleUri id = new SimpleUri(jsonReader.nextName());
+                String value = jsonReader.nextString();
+                temporarilyParkedSettings.put(id, value);
+            }
+
+            jsonReader.endObject();
+        } catch (Exception e) {
+            // TODO: Handle exception
```
emanuele3d: Yes please. :slightly_smiling_face:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -18,11 +18,11 @@
 import org.terasology.engine.SimpleUri;
 
 public interface FlexibleConfigManager {
-    void addFlexibleConfig(SimpleUri flexibleConfigUri, FlexibleConfig flexibleConfig);
+    void addConfig(SimpleUri configUri, FlexibleConfig config) throws RuntimeException;
```
emanuele3d: Sorry if I said otherwise, but this configURIs are really configIDs, so I'd rename all instances of configUri to configId, to stick to what it is rather than its type.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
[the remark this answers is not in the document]
vampcat: > This should eliminate zombie settings as configs are atomically loaded and atomically discarded.
I'm sure I'm missing something, but I can't see how different config files can prevent zombie settings from piling up. (The only possible way I can see is deleting config associated with myModule when user deletes myModule, but that's not a normal scenario, since I don't expect user to be deleting modules all the time.)

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
vampcat: I had not considered the second part - the divergence. Definitely makes way more sense to me now :D

engine/src/main/java/org/terasology/engine/subsystem/common/ConfigurationSubsystem.java
vampcat: Yeah, the alert seems like a logical idea. That should be easily doable via NUI I think.

( To be fair, since I haven't dived into it too much, I might be totally and horribly wrong :D )

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
vampcat: The configManager is responsible for deciding the path of serialization, creating a reader/writer for it and passing that to the FC.

This way, the FC does not have to concern itself with *where* the settings are being stored, and can focus only on the *how*.

engine/src/main/java/org/terasology/config/flexible/settings/DoubleSetting.java
vampcat: As far as I know, that seems to be the case.

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
emanuele3d: I understand that. But could the writer/reader be passed on Construction? If a path change requires re-instantiation, I guess the answer is no.

engine/src/main/java/org/terasology/config/flexible/settings/DoubleSetting.java
emanuele3d: test

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
```diff
@@ -138,7 +136,7 @@ public void load(Reader reader) {
 
             jsonReader.endObject();
         } catch (Exception e) {
-            // TODO: Handle exception
+            throw new RuntimeException("Error parsing config file!");
```
emanuele3d: Hmmm... I imagine this prints out the whole stack trace normally. But will this include any hint of where the parsing went wrong? It would be important to have that information printed out.

And can the catch any more precise? I.e. see:

https://www.javaworld.com/article/2077500/testing-debugging/java-tip-134--when-catching-exceptions--don-t-cast-your-net-too-wide.html

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -60,9 +60,32 @@
...
+     */
     void save(Writer writer);
 
+    /**
+     * Loads the values of the settings having non-default values, to enable persistence across sessions.
+     *
+     * All the non-default values are loaded and "parked" till the actual setting is added to the config,
+     * and which point they are "unparked" and parsed, replacing the default value of the setting.
+     *
+     * Note that this function should be called before adding any settings to the FlexibleConfig.
```
emanuele3d: I'd expand on this a little and clarify what happens if this isn't the case, i.e. settings are overwritten.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -60,9 +60,32 @@
...
+     * saved as-is, to be used later.
+     *
+     * @param writer A writer that will serve as the destination of settings.
+     */
     void save(Writer writer);
 
+    /**
+     * Loads the values of the settings having non-default values, to enable persistence across sessions.
+     *
+     * All the non-default values are loaded and "parked" till the actual setting is added to the config,
```
emanuele3d:
```
* All the non-default values are loaded and "parked", initially remaining inaccessible. 
* Once a Setting object is added to the config, a corresponding non-default value is sought 
* among the parked values. If one is found it is parsed and stored in the Setting object.
```

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -17,14 +17,46 @@
 
 import org.terasology.engine.SimpleUri;
 
+/**
+ * Stores multiple {@link FlexibleConfig} instances that can be retrieved using their id.
+ * Also responsible for overseeing their serialization - to and from - disk.
```
emanuele3d: I'd maybe write "coordinating" instead of "overseeing".

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -17,14 +17,46 @@
...
+ */
 public interface FlexibleConfigManager {
+    /**
+     * Adds the given {@link FlexibleConfig} to this manager.
+     * This method throws a RuntimeException if a FlexibleConfig with the given id already exists.
```
emanuele3d: I think you can move the content of this line and put it besides the @throws line below, isn't it?

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -17,14 +17,46 @@
...
+     * @param config The config that is to be added.
+     * @throws RuntimeException
+     */
     void addConfig(SimpleUri configId, FlexibleConfig config) throws RuntimeException;
 
+    /**
+     * Removes the config associated with the given id, and returns it.
+     *
+     * @param configId The id of the config to remove.
+     * @return The config associated with the given id, and null if no config is associated with the given id.
```
emanuele3d: The config associated with the given id, **or** null if no config is associated with the given id.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -17,14 +17,46 @@
...
+     */
     FlexibleConfig removeConfig(SimpleUri configId);
 
+    /**
+     * Retrieves the config associated with the given id.
+     *
+     * @param configId The id of the config to retrieve.
+     * @return The config associated with the given id.
```
emanuele3d: We should clarify what happens if there is no config associated with the given id.

Cervator:
Here's the failure from Jenkins:

```
:modules:Core:processResources UP-TO-DATE
:modules:Core:classes
:modules:Core:checkstyleMain
:modules:Core:compileTestJava/home/jenkins/workspace/TerasologyPRs/modules/Core/src/test/java/org/terasology/world/generator/TreeTests.java:60: error: constructor Config in class Config cannot be applied to given types;
        context.put(Config.class, new Config());
                                  ^
  required: Context
  found: no arguments
  reason: actual and formal argument lists differ in length
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
1 error
 FAILED
```

So in other words it sounds like a *test* class is failing to compile due to the changes. Which is a common issue and easy to miss since the test classes may not auto-compile when tinkering with the regular Java source.

Doesn't look like it would be a very difficult fix either. Try it out and force-compile the test source and/or try running all the engine tests to see if they're okay :-)

vampcat:
@emanuele3d Passing the Context into Config to avoid the use of CoreRegistry (which is temporary, and will be removed once we're done finding a proper place to put `FC.save()` instead of inside `Config.save()`) has caused me to change code in a lot of places. I personally don't mind too much, but I think it's worth reconsidering once.

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigImpl.java
vampcat: I restructured the code slightly to print the name of the config that failed to parse, and printed the stack trace of that exception as well - that should cover the "where" part.

emanuele3d:
@vampcat wrote:

> Passing the Context into Config to avoid the use of CoreRegistry (which is temporary, and will be removed once we're done finding a proper place to put FC.save() instead of inside Config.save()) has caused me to change code in a lot of places. I personally don't mind too much, but I think it's worth reconsidering once.

Are you talking about changes that I haven't seen yet? Because so far the changes related to the context being passed in didn't seem so bad. And what would a proper place for FC.save() be?

vampcat:
No, I just meant the changes I have pushed (not sure if you've viewed them all, when saying "the ones i've viewed"). However, that does mean that a lot of tests have to pass `null` to the config, which feels weird..

Although now that I think about it, it would have been better to simply add a new constructor that takes no arguments, that automatically sets the variable to null.

Your thoughts?

emanuele3d:
Tests are meant to also provide a hint of how things are used in practice. If in practice we should always pass a Context to the constructor then we should do it that way in the tests too. So, I'd avoid a new constructor in this case.

Indeed passing null when we should pass an object is not ideal though. Ideally we'd pass a MockContext object. 

You didn't respond about where else you'd put FC.save() by the way.

vampcat:
For the time being, nowhere. However, since FC will replace Config either way, we'll soon replace all `config.save()` calls with `fcManager.save()` calls.

emanuele3d:
I'm skeptical about that "soon". :wink: But ok for the rest.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -54,4 +59,33 @@
...
+     * and which point they are "unparked" and parsed, replacing the default value of the setting.
+     *
+     * Note that this function should be called before adding any settings to the FlexibleConfig.
+     *
+     * @param reader A reader that will serve as the source of the settings.
+     */
+    void load(Reader reader);
```
oniatus: I would separate loading and saving from managing settings. Changing and reading setting values and load/save are two different responsibilities for different clients and the strategy how to save a setting (io.Reader and io.Writer) are implementation details which should not be part of this interface.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.config.flexible;
+
+import org.terasology.engine.SimpleUri;
+
+/**
+ * Stores multiple {@link FlexibleConfig} instances that can be retrieved using their id.
+ * Also responsible for overseeing their serialization - to and from - disk.
```
oniatus: In addition to my previos comment about save/load, I'd expect a method like `save(FlexibleConfig)` and include the reader/writer logic here.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -0,0 +1,62 @@
...
+     * @param configId The id of the config to retrieve.
+     * @return The config associated with the given id.
+     */
+    FlexibleConfig getConfig(SimpleUri configId);
+
+    /**
+     * Iterates over all the configs registered with this manager, and loads the
+     * settings stored in them.
+     */
+    void load();
```
oniatus: Just by reading `FlexibleConfigManager.load()` -> what will it do?
How about `FlexibleConfigManager.loadAllRegisteredConfigs()`?

Also naming consistency -> javadoc calls it "registered", the method to register is called "add".

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Iterates over all the configs registered with this manager, and saves the
+     * settings stored in them.
+     */
+    void save();
```
oniatus: same as above

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.IOException;
+import java.io.Reader;
+import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
+import java.util.Map;
+import java.util.Map.Entry;
+
+public class FlexibleConfigManagerImpl implements FlexibleConfigManager {
+    private static final Logger logger = LoggerFactory.getLogger(FlexibleConfigManager.class);
```
oniatus: This should be a logger for the -Impl class

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public void load() {
+        for (Entry<SimpleUri, FlexibleConfig> entry : flexibleConfigs.entrySet()) {
+            SimpleUri flexibleConfigId = entry.getKey();
+            FlexibleConfig flexibleConfig = entry.getValue();
+
+            try (Reader reader = Files.newBufferedReader(getPathForFlexibleConfig(flexibleConfigId), TerasologyConstants.CHARSET)) {
+                flexibleConfig.load(reader);
+            } catch (NoSuchFileException e) {
+                // The config does not exist, so the default values will be used.
```
oniatus: Exceptions should not be part of the expected controlflow.
Use `java.nio.files.Files.exists(Path)` to check if the file at the path exists and ignore this case explicitly

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
emanuele3d: I'd recommend to rename load to loadAllConfigs().

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
emanuele3d: Hmmm... I'm not sure about the dogma implied in the first sentence. :wink: 
But the fix sounds reasonable enough. =)

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
emanuele3d: Hmmmm... I'm not sure I agree with this and the related comment above. 

Can you do a comparison between the two different ways of doing things and show why the changes you suggest would be better?

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
emanuele3d: Fair point about consistency. I'd retain the add method and change the javadoc.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
vampcat: Actually, I had the loading/saving login in FCManager before, but @emanuele3d commented the following:
> I am thinking that in terms of load/save functionality, it makes sense that the manager orchestrates it, i.e. iterating over the FCs as it happens here and providing the file or even the reader. But I'd want the FCs themselves to handle the actual reading/writing, to separate the concerns and mantain some of the functionality private to the FC class.
Which is why I put this here.

I'm fine with either, so can you two decide which feels better? :D

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
emanuele3d: I'd keep it as it is for the time being. Let's get this PR moving again and merged in. We can always refactor after we have gained some experience with FCs in general.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
emanuele3d: Reminder about this one.

emanuele3d: More specifically we could write:

* Note that this function should be called before adding any settings to the FlexibleConfig, or any corresponding parked setting will never be unparked.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManager.java
emanuele3d: Reminder about this one.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

emanuele3d:
@vampcat: I've dealt with a few outstanding javadoc issues. Can you please fix the conflicts?

@Cervator:  stand-by for merge plz. =)

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
```diff
@@ -80,10 +80,12 @@
...
-     * All the non-default values are loaded and "parked" till the actual setting is added to the config,
-     * and which point they are "unparked" and parsed, replacing the default value of the setting.
+     * All the non-default values are loaded and "parked", initially remaining inaccessible. 
+     * Once a Setting object is added to the config, a corresponding non-default value is sought 
+     * among the parked values. If one is found it is parsed and stored in the Setting object.
      *
-     * Note that this function should be called before adding any settings to the FlexibleConfig.
+     * Note that this function should be called -before- adding any settings to the FlexibleConfig. 
+     * Otherwise any corresponding parked value will never be loaded.
      *
```
vampcat: Kinda nitpicky, but shoudn't it be a comma before the otherwise? Something like `the FlexibleConfig, otherwise any`
Not a show-stopper in any case.

vampcat:
Btw, how do you feel about creating a new MockContext object? I really dislike the `Context(null)` constructor i've had to use, so I'd be glad to replace it with MockContext in a small PR.

GooeyHub:

engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java
emanuele3d: I made the conscious decision to add a period and make the sentence shorter, but I don't have a problem either ways.

emanuele3d:
Yes, to use a MockContext would be better, as I suggested in an earlier comment. I'd do it in this PR though: it is in this PR that a context is introduced as input to the constructor, therefore it is in this PR that ideally the tests should be modified accordingly. It's not a big deal though: if you feel strongly about doing it in a separate PR, ok.

GooeyHub:

oniatus:
Approved, next change should use the new system for at least one of our config concepts to put some usage on the new Api 👍

emanuele3d:
@oniatus: indeed the idea is to start using FCs in the context of the Renderer, at least. As the two Rendering configs have a lot of parameters we are unlikely to migrate everything to FCs in one step, but that's the direction we are aiming for.

emanuele3d:

emanuele3d:
Poking @Cervator for merge. Unless there is something in this PR that doesn't quite fully convince him?

vampcat:
Cervator is currently flying to Europe or something, so I don't think he'll be available a whole lot for a few days. That being said, I'll try poking him on slack (since he tends to ignore all github pings).

oniatus:
Went for a second review and gave it another playtest to make sure everything works.
Good work, I am looking forward to upcoming changes in order to improve the unflexible configs 👍

emanuele3d:
@oniatus, I know you have merge privileges (so do I) and I appreciate the good intention. However, please do not merge PRs in which I am involved unless explicitly requested. There are some type of PR that @Cervator prefers to handle himself.

oniatus:
@emanuele3d Sorry for that, we should clarify the review/merge process then because I did not know this till now.

emanuele3d:
No worries, I do realize this has never been thoroughly discussed and responsibilities within the merge-privileged have not been clearly defined. 

There is no doubt Cervator could use some help. But perhaps we can help with the least controversial, least consequential PRs first, leaving to him the more significant ones, especially in light of the fact that he tests them against his "mega-workspace", an IJ project he has that includes all existing modules, to spot conflicts that normally would go unnoticed. 

That been said, it is certainly good we have this conversation so that we can clarify things. Maybe he'd be happy to relax the workflow and let a number of other people deal with any PR.

Cervator:
Belated response to this here. Interesting situation with the merging - I *really* would love more help merging, and honestly with this amount of PRs some bumps and bruises might be worth it. So I'm thrilled @oniatus stepped in :-)

At the same time @emanuele3d couldn't have picked better timing with this PR as an example for the usefulness of a mega-workspace as it *did* in fact break something super subtle, a test class in PolyWorld that had `Config config = new Config();`

I only found it from Alpha 9 prep after sorting out an API violation that slipped through and is documented at #3184. That was keeping my mega-workspace from doing full builds, so I might not have caught this anyway. To make things more perfect-storm'ish I had built every single module as part of release prep just 3 days before this got merged, making me puzzled as Jenkins was still happy ... :D

Had me confused for a bit until I finally realized commit https://github.com/MovingBlocks/Terasology/commit/ee0d74a6e6da7783cbe3b04a2fb0e140921f7289#diff-dbff4de92b8b73995ac93deb86683778 from here added a new constructor for `Config` meaning the default one went away, which is in fact also an API violation, woo! :-) That is a tricky one to catch

Both API violations seemingly only caused one tiny module break each, fixing both right now. A fancier setup in Jenkins would have caught both and done away with the need for mega workspaces. In the meantime with the new Groovy utility making such a workspace is easier so maybe a few beyond me can keep such a workspace around for testing tricky PRs (but nothing else - keep a separate dir for it and don't do any work there, aiiee can they get tricky to keep clean)

Thanks for everybody's effort and initiative here, please keep at it, even if the road may get bumpy from time to time. So long as we test before doing an actual *release* we can still fix issues like this and stay in compliance with SemVer, so even if we get a few PR merges with slight issues that still keeps us moving and OK. 

Onward and upwards!

PR #3121, opened 2017-10-12
Implemented libgdx bullet wrapper Tera Bullet 2 #Discussion 

pollend:
@vampcat  mentioned some performance issues with converting between math implementations between Terasology and libgdx. So I've extracted the bullet portion from libgdx and refactored it to use TerMath. This is not quite complete and does not quite work at the moment. I have some problems with dependencies in the bullet implementation that need to be worked out.

ref #3042 
https://github.com/pollend/TeraBullet-2


GooeyHub:

Cervator:
After merging #3199 this one is conflicting quite a bit, unlike #3042 - should we still consider this approach, or is #3042 really a superior option we should focus on instead?

eviltak:
@Cervator From my last discussion with @pollend, this PR has libgdx build problems whereas #3042 has JNI problems and segfaults. As of then, the bullet physics wrapper introduced #3042 was not working (correct me if I'm wrong, @pollend), but this one was. However, the one in #3042 should, in theory, be quite more performant. Michael should be able to throw more light on this.

PR #3122, opened 2017-10-15
Log an error when all modules aren't found locally in the /modules folder when running from source

### LINKED ISSUE #2381: No support for auto-downloading modules if server if running from source
I expect this is because the auto-download in the server expects to find module jar files under `/modules` rather than source directories. I would like for us to catch that situation better and log it, then maybe attempt to find built jars (`modules/[module]/build/libs/[module-[version].jar]`) if present.

The logging would need to include a disclaimer that if the goal is to have a server running from source successfully serving modules the developer would have to run `gradlew jar` first to make sure there are actually built jar files present.

TheFlash98:
### Contains
Fixes #2381 
Added a simple logger which runs when there is a server timeout and the game is running from source.

### DIFF
--- a/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
+++ b/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
@@ -81,6 +81,9 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
                             && joinStatus.getStatus() != JoinStatus.Status.FAILED) {
                         joinStatus.setErrorMessage("Server stopped responding.");
                         channel.close();
+                        if(server.getRemoteAddress().startsWith("localhost")) {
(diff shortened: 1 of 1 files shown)


GooeyHub:

jellysnake:
So, I don't feel like this properly addresses the issue.  
The logging in this PR occurs in the client whilst #2381 is proposing to add logging for the server.

What I would expect this PR to contain is logging for the server when there is no jar files under `/modules`. The logging should say that if one is attempting to run a server from source then you need to run `gradlew jar` first (Or the logging should link to a wiki page that says the same thing)

TheFlash98:
Ah, I see. Sorry about this. Do you want me to close this and make a new one?

jellysnake:
You can either make a new commit undoing these changes or make a new PR it's up to you.
Thanks for picking up the issue either way 👍.

Issue #1433, opened 2014-12-09
Easy option to export player identity


Cervator:
Currently the game generates an identity for the player so in a client-server setting the server knows who is who. If you clear your local game config you lose said identity and cannot get it back.

We should add a simple "Export Identity" to the player settings menu that'll write out some simple player identity stuff, along with an "Import Identity" to overwrite current with a saved version.

Later we could probably get into outright profiles, but for now this would be a nice little usability win
##

Cervator:
Pinging @flo on this one for awareness - thought it might be somewhat related to your user identity work for ownership and such. Would be nice if when players can "own" or protect blocks that they'd also be able to safeguard their identity somewhat :-)

Cervator:
Bump - thanks to @gianluca-nitti we now have an option to run a central identity server that can let you use your identity from different places (like multiple PCs) and log in to a game server as the same player.

That's more or less this issue taken to the next step beyond it, but I suppose some might still like the manual option. Probably wouldn't be hard to add at this point.

gianluca-nitti:
> I suppose some might still like the manual option. Probably wouldn't be hard to add at this point.

I think I could implement this, however I'm not sure what exactly is desired.
Do you mean that a file picker dialog should open when clicking export/import so that the user can choose the location of the file (sorry if it's obvious, I just want to make sure)? This could be done with the Swing file chooser or by implementing a custom NUI file picker (which might become useful for other purposes too in the future).

Also, the identities on each server the client has joined are just a part of the configuration. Maybe we could reconsider what exactly should be exported. I see some options:
* Identity certificates only (probably what was meant when this issue was opened)
* Whole config (which wouldn't even be a real data export, just a file copy from the data directory to the user-specified location)
* Identities + Player config (+ keybindings? not sure), similar to first option but we also include player name and color; IMO an advantage over the whole config is that we exclude the video settings, which a user may want to keep different between two clients on different computers with different graphic capabilities
* Have a list of checkboxes to let the user choose what to export, and possibly what to import from an exported config, with reasonable defaults (probably the best option for "power users" but may confuse beginners).

Any comments?

PS. I probably won't be able to do it right now because I have a university exam in mid September which I'm revising a lot for, but I'll be happy to contribute on this issue after it.

Cervator:
Forgot to close this - it was completed by #3123

PR #3123, opened 2017-10-15
Add feature to export/import multiplayer identities

gianluca-nitti:
### Contains

Implements the feature suggested in #1433; in other words, adds two buttons in the player settings screen that allow to export the identities used to authenticate to multiplayer server to a file, and to import a file of previously exported identities.
A NUI-based file picker is also included to allow the user to choose the file to save to or load from (screenshot attached).
screenshot_20171015_191840

### How to test

* Make sure to have some identities in your configuration (if not sure, just join a server to automatically generate one). For testing purposes, also make sure that at least one identity is associated with something meaningful; for example, join a server and move your character to a certain place and/or pick up or drop some items.
* From the main menu, go to Settings > Player, and click the `Export...` button next to the `Multiplayer identities` label. the file picker popup will open; navigate to the directory where you want the file containing the identities to be saved, then enter a name for the file in the text box at the bottom of the popup and click OK.
* Exit Terasology and delete or rename the configuration file (`config.cfg`); obviously before deleting it, back it up if it contains something you want to preserve beside identities.
* (optional) Start the game again and join the same server; you will be placed at the spawn point and have the default item set because the identity has been deleted when you deleted the configuration file.
* Go back to the main menu, then go to Settings > Player and click `Import...`; in the file picker, select the file you previously exported to and click OK. If you did the previous step, you will be prompted to choose if you want to replace the locally saved identity; choose to import it anyway, overwriting the one which has been generated while performing the previous step with the one present in the file.
* Join the same server; you will be in the same position, have the same items, etc. as before deleting the configuration file.

Nothing I can think of, but obviously any feedback/suggestion is welcome.


GooeyHub:

Cervator:
Very nice work - tested it out and all was well first try even detecting that I was importing with a different homedir that had a different identity. I picked overwrite and was able to connect to a test server and resume where I left off from the original client :-)

The file picker was slightly confusing at first just from its appearance, but worked fine otherwise. Seems a useful widget if we don't just spin up native look & feel popup file chooser windows for everything (which I guess would be a possibility)

PR #3124, opened 2017-10-16
Add logger for missing .jar files when loading modules as a server

### LINKED ISSUE #2381: No support for auto-downloading modules if server if running from source

TheFlash98:
### Contains

Fixes #2381 

### How to test

Delete jar files of some modules you have locally. Run the game as a server and add modules whose jar files you just deleted. You will notice a logger which reads 
"The module moduleId could not be loaded from /modules. Try running "gradlew jar" "

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
@@ -28,11 +28,13 @@ import org.terasology.engine.TerasologyConstants;
 import org.terasology.engine.modes.StateLoading;
 import org.terasology.engine.module.ModuleManager;
 import org.terasology.engine.module.StandardModuleExtension;
+import org.terasology.engine.paths.PathManager;
 import org.terasology.game.GameManifest;
 import org.terasology.i18n.TranslationSystem;
 import org.terasology.module.DependencyInfo;
 import org.terasology.module.DependencyResolver;
 import org.terasology.module.Module;
+import org.terasology.module.ModuleLoader;
 import org.terasology.module.ResolutionResult;
 import org.terasology.naming.Name;
 import org.terasology.network.NetworkMode;
@@ -58,6 +60,9 @@ import org.terasology.world.generator.internal.WorldGeneratorManager;
 import org.terasology.world.internal.WorldInfo;
 import org.terasology.world.time.WorldTime;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Thanks for this and finding the right area :-)

I tried to test this but am hitting an issue that might be environment related:

```
21:22:11.150 [New I/O worker #3] ERROR o.t.n.i.ServerConnectionHandler - Error sending module
java.nio.file.AccessDeniedException: C:\Dev\Terasology\Git\Terasology\modules\WildAnimalsGenome
	at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:83)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:97)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:102)
	at sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:230)
	at java.nio.file.Files.newByteChannel(Files.java:361)
	at java.nio.file.Files.newByteChannel(Files.java:407)
	at java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:384)
	at java.nio.file.Files.newInputStream(Files.java:152)
	at org.terasology.network.internal.ServerConnectionHandler.sendModules(ServerConnectionHandler.java:105)
	at org.terasology.network.internal.ServerConnectionHandler.messageReceived(ServerConnectionHandler.java:80)
...
```

My setup is as follows:

1) Source workspace in which I've deleted the `build` dir under "WildAnimalsGenome" so no jar file is available (the class files rebuilt automatically, but didn't get jarred)
2) Binary download where I've deleted the jar file as well
3) Run headless server from the workspace I have open in IntelliJ
4) Run client from the binary download dir, try to connect to the local server

The server and client both start fine, and I can see on the join screen that the client sees the server running WildAnimalsGenome yet it isn't present locally. On trying to connect the client attempts to download but the server fails with the above message.

Oddly it seems like it manages to entirely avoid the new code (I had a breakpoint set - it didn't get hit). Maybe it is trying to even *see* what's in `/modules/WildAnimalsGenome` yet hits a permission denied error on the directory because Win10 is weird?

Maybe that's an OS issue and it works fine for you and some others. Not sure whether somebody else could test it and validate or if we can find a good solution to catch this case as well.

TheFlash98:
@Cervator thanks for testing this out. Although why did you delete the build dir under "WildAnimalsGenome"?
I'm on macOSX. I run the game through IntelliJ, host a new game with the modules whose jars I delete from the `/modules` dir and not the build dir under a particular module. 
Then I ran another game and tried to join the game hosted earlier. The "Join Game" screen _does not_ show the modules whose jars were missing and which did not actually load.

Cervator:
I deleted `build` under an example module as a short cut for making sure that source module didn't have a jar inside its build dir - that's the jar I'd expect the server would send to connected clients when running from source. If that jar doesn't exist then the log entry would be served instead and the user encouraged to run `gradlew jar` :-)

Which jars are you deleting? Versions living directly under `/modules/*.jar` in a source workspace? Those make up a subtly different case as they don't exist in source format so you wouldn't be able to successfully start a server without them. That's unlike modules present in source, which the server should be able to use directly from class files, but those can't be sent to a remote client.

Does that help? Were you running the server from source and if so did you have a module set configured that included one or more missing jars (but existing class files)?

I'm not entirely sure exactly how the logic for sending jars to clients works when it comes to running from source in the first place where you might have *both* jar files from built dirs (source modules) and jar files from `/modules` (dependencies without source present). That is probably a bigger issue than #2381 covers and may involve OS-specific quirks like the permission error I get.

TheFlash98:
Hey @Cervator sorry about the delay. But as written here https://github.com/MovingBlocks/Terasology/issues/2381 
> I expect this is because the auto-download in the server expects to find module jar files under `/modules` rather than source directories"

Which is exactly what I tried to handle here, if there is no jar file under the /module folder, I log it.

skaldarnar:
More logging is always nice, but code has change too much over time - `CreateGameScreen.java` does not even exist anymore. If anybody wants to add more logging you can take this as a starting point 🙃

Issue #3099, opened 2017-09-07
Further options for new Groovy-based utility commands


Cervator:
With #2979 merged we now have the new [Groovy Wrapper](https://github.com/MovingBlocks/groovy-wrapper) in place for making easier and more powerful utility scripts than the somewhat awkward old `gradlew fetchModuleX` approach. Now we can has `groovyw module get X` instead!

I only prepared two initial options for fetching individual modules or all modules in a dependency tree. With that example in place I expect it will be fairly easy to go beyond that to include module creation in `module.groovy` and adding new *stub* scripts for facades, libs, and meta modules (each type would live in its own script like `facade.groovy`)

Likely there will be shared functionality between the scripts that could be centralized in a single include script of some sort. So here are some tasks:

- [x] Finish the module creation part of `module.groovy` (there is just a placeholder entry in the `switch` right now)
- [x] Create `facade.groovy` and make the new versions of retrieving and creating facades (shouldn't need to worry about dependencies so recursing isn't useful there). Good example to retrieve: https://github.com/MovingBlocks/FacadeServer (`gradlew fetchFacadeServer` -> `groovyw facade get Server`)
- [x] Create `lib.groovy` and retrieval and creation there. Also no dependencies/recursing there. Good example: https://github.com/MovingBlocks/CrashReporter (`gradlew fetchLibCrashReporter` -> `groovyw lib get CrashReporter`)
- [x] Create `meta.groovy` and retrieval plus creation (just an empty dir really) for meta modules. No dependencies or recursing needed, they're just storage repos. Good example: https://github.com/MetaTerasology/WildAnimals (`gradlew fetchMetaWildAnimals` -> `groovyw meta get WildAnimals`)
- [x] Centralize any shared logic resulting from the above tasks
- [x] Add entirely new option to *update* one, a set (dependencies), or all modules with a single command (so something like `groovyw module update X`, `groovyw module updateTree X`, and `groovyw module updateAll`)
- [x] Come up with some additional suggestions? Unsure if there is any point in bulk updating facades or libs. Maybe meta modules? Seems unlikely though.

Look in `config/gradle/utility.gradle` for how the old approach worked, which files to include or create, and so on. See also #2979 for more details about how the new system works.

SkySom:
So does the new module get allow for telling it to pull all dependencies as repos?

Cervator:
@SkySom `groovyw module recurse X` will :-) X plus all its dependency modules will be retrieved as source modules.

[...]

PR #3127, opened 2017-10-19
Implement new module interface in groovy

mjuvekar7:
Port the 'new module' Gradle interface to Groovy.

Ref: https://github.com/MovingBlocks/Terasology/issues/3099

### DIFF
--- a/module.groovy
+++ b/module.groovy
@@ -180,8 +180,44 @@ if (args.length == 0) {
             break
         case "create":
             println "We're doing a create"
-            String someParam = getUserString("Please enter a thing")
-            println "User return was: $someParam"
+            String name = ""
+
+            // Get new module's name
+            if (args.length > 2) {
+              println "Received more than one argument. Aborting."
+              break
+            } else if (args.length == 2) {
+              name = args[1]
+            } else {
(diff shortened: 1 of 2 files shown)


Cervator:
Thanks for getting this started! There are a few more bits needed:

- [ ] Wrap the majority of the new logic in a `def createModule(String name) {` leaving just the input logic in the case statement (getting the name). I'd put the new method between `readModuleDependencies` (relates to the methods above) and `getUserString` (general utility).
- [ ] Also copy in the `build.gradle` file from the template (the fetch does that too) - it'll immediately get git ignored which is fine (Gradle still depends on that file to determine if something is a valid module)
- [ ] Test that `gradlew idea` will correctly register the new module, which would have exposed the bug in the prior item ... :D
- [ ] Add a tiny bit of javadoc to the new method to go with the existing ones
- [ ] Add an entry for the new method to the `printUsage()` info

mjuvekar7:
Done. Also added build.gradle to the templates folder (currently the script was just copying from `modules/Core`, which isn't good. so I copied the build.gradle into the templates folder)

Cervator:
Tests out great now, thanks! :-)

Issue #2939, opened 2017-05-06
Dying with the player online UI visible makes it get stuck, player cannot work any other UI element or make it go away


Cervator:

GabrielXia:
When player dies with the `Online Player` UI visible (holding `tab`), the `Online Player`UI remains and covers other buttons. So the player can neither play nor exit.
![2017-05-06 8 01 26](https://cloud.githubusercontent.com/assets/17897358/25774993/0d6ce17a-329b-11e7-8eb9-3e2a60b49d38.png)
- code of this UI: engine/src/main/java/org/terasology/rendering/nui/layers/ingame/OnlinePlayersOverlay.java
- maybe related pr(the ping system): #2799

GabrielXia:
Upload logs after drying
~~~
20:46:08.399 [main] INFO  o.t.network.internal.ServerImpl - Destroying entity: EntityRef{id = 0}
20:46:08.400 [main] INFO  o.t.network.internal.ServerImpl - Destroying entity: EntityRef{id = 3240, netId = 3124, prefab = 'Equipment:EquipmentInventory'}
20:46:08.401 [main] INFO  o.t.network.internal.ServerImpl - Destroying entity: EntityRef{id = 3239, netId = 3123, prefab = 'engine:gaze'}
20:46:08.401 [main] INFO  o.t.network.internal.ServerImpl - Destroying entity: EntityRef{id = 3237, netId = 3122, prefab = 'engine:player'}
20:46:08.402 [main] INFO  o.t.network.internal.ServerImpl - Destroying entity: EntityRef{id = 3236, netId = 3121, prefab = 'Core:transferSlot'}
20:46:08.714 [main] ERROR o.t.l.debug.ChunkEventErrorLogger - Multiple loads of chunk (0, -2, 0)
20:46:10.165 [main] ERROR o.t.l.debug.ChunkEventErrorLogger - Multiple loads of chunk (-1, -2, 0)
20:46:10.531 [main] ERROR o.t.l.debug.ChunkEventErrorLogger - Multiple loads of chunk (0, -2, 1)
20:46:10.598 [main] ERROR o.t.l.debug.ChunkEventErrorLogger - Multiple loads of chunk (0, -2, -1)
20:46:10.662 [main] ERROR o.t.l.debug.ChunkEventErrorLogger - Multiple loads of chunk (1, -2, 0)
[...]

PR #3128, opened 2017-10-20
Close online player overlay before showing death screen

ThompsonTyler:
### Contains
Solves #2939 by closing the overlay prior to pushing the death screen. 

Problem was happening because the ping/online players is an overlay and therefore is not actually in the screen stack and won't be closed when the death screen is pushed (death screen is set to not have lower screens visible).

I think functionally overlays and screens work as intended so no need to change their functionality, just this specific case, as overlays are "above" other elements and therefore should still show with screens displayed.

### How to test

Go into game, hit tab to show online players then go into console and kill with online players screen open, screen should close and allow access to death screen.

Open/close online players to make sure that still functions. 

Kill with out player screen open to verify it still works.

Nothing that I know of, unless we want to instead actually check for any open overlays on displaying a screen with `isLowerLayerVisible return false`

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/MenuControlSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/MenuControlSystem.java
@@ -73,6 +73,7 @@ public class MenuControlSystem extends BaseComponentSystem {
     public void onPlayerDeath(PlayerDeathEvent event, EntityRef character) {
         EntityRef client = character.getComponent(CharacterComponent.class).controller;
         if (client.getComponent(ClientComponent.class).local) {
+            nuiManager.removeOverlay("engine:onlinePlayersOverlay");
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Thanks for the fix!

Sure seems like a pretty straight forward one-liner. I wonder if it does point to a need for improvement to the system, making it aware of different UI contexts. In this case the players online overlay is in the engine and pretty fundamental, but what if an author adds a new overlay in a module - we couldn't explicitly close that the same way. 

We'd have to rely on the author to either add explicit handling to close in the module itself or make UI stuff default to something sensible, like close everything on death/reset except elements requesting otherwise. Maybe that's a bigger issue surrounding game states / fundamental events.

Pinging @rzats for reference

PR #3129, opened 2017-10-20
Move `TimeRange` class from `SimpleFarming` module; rename it `Interval`

aaron-harris:
### Contains

Moves the `TimeRange` class from the `SimpleFarming` module into the `Core` module.  (I used the new package `org.terasology.core.logic.random`; let me know if there's a better place to put this.)  

Also renames this class `Interval` to make it more clear that there's nothing about the class that mandates viewing the interval as a time range; it's just a way to get a random number in a certain range, and allow components to have that range as a field type.  Similarly, renames the sampling method from `getTimeRange` to `sample`.

The goal here is just to make this functionality available to other modules; I'm working on Terasology/WildAnimals#25 and am trying not to reinvent this particular wheel.

### How to test

Not much testing is required at this stage, since the new class is essentially a copy of the original with no references to it.  I did add a basic unit test that verifies that the samples are in the correct range, but I don't see any way that this could break anything.

A second PR Terasology/SimpleFarming#60 handles updating the usages within `SimpleFarming` to point at the new class.

### DIFF
--- a/modules/Core/src/main/java/org/terasology/core/logic/random/Interval.java
+++ b/modules/Core/src/main/java/org/terasology/core/logic/random/Interval.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.core.logic.random;
+
+import org.terasology.reflection.MappedContainer;
+import org.terasology.utilities.random.FastRandom;
+import org.terasology.utilities.random.Random;
+
+/**
+ * A representation of a uniformly-distributed integer range, for use as a datatype by {@code Component}s
+ *
+ * For example, suppose that {@code ExampleComponent} is a {@link org.terasology.entitySystem.Component} whose {@code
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Cool, thanks, will be checking this out soon :-)

@GooeyHub: ok to test

GooeyHub:

aaron-harris:
As @jellysnake pointed out in PR Terasology/SimpleFarming#60, upcoming revisions to SimpleFarming do away with TimeRange anyway.  Looking them over, the idea seems to be that components would store the interval bounds as a pair of bare of integers and systems would be responsible for using them to generate a time randomly.  That's simpler, but means that multiple independent systems might duplicate the (very simple) sampling code.  Probably not a big deal, since that method is very simple.

In any event, I'm closing this PR because there's no point in making code more generally available if existing modules are moving away from it.  I'll redo PR Terasology/Wildanimals#26 to be independent.

Cervator:
Thanks still for the work @aaron-harris ! I just merged that PR so if you want to adjust :-)

Also, hop on IRC sometime or let me know an email address somehow so I can send you a Slack invite? Easier to coordinate on there

jellysnake:
Wanted to say that the sentiment I wanted to get across was not that this was made obselete by my pull, but rather that it should be added after my pull. Now that has been merged if you want feel free to re-open it and simply alter what's there to fit the new systems.
It may be simple but it's already duplicated twice in simple farming alone so it's probably worth it

aaron-harris:
@jellysnake: Understood.  After looking at your new setup, though, I've come to the conclusion that the `Interval` class wouldn't do enough to justify its own existence.  Presumably you thought the same thing, since you deleted `TimeRange`.  For one thing, I like the cleaner JSON that results from using the bare integers instead.

For another, I notice that your duplicated method essentially boils down to error handling for the `max < min` edge case; otherwise, any `generateRandom` method call could be replaced with a call to `random.nextLong`, which I find perfectly intelligible.  That suggests if this is worth deduplicating, then maybe `nextLong` (and `nextInt`, etc.) should just handle backwards intervals in some more sensible manner.  Do we know if there's any code in the project that relies on the current undocumented behavior in the case of a backwards interval?

PR #3133, opened 2017-10-25
Added support for PRE and POST FBO_REGENERATION events in FboM

vampcat:
Divided the current `update()` event fired by FboManagers into two parts, the PRE_FBO_REGENERATION and POST_FBO_REGENERATION events that allow for finer control.
Also replaced the `FBOManagerSubscriber` interface with the generic `Subscribable` interface, allowing support for subsribing to specific proprties.


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/AmbientOcclusionNode.java
```diff
@@ -112,8 +113,8 @@ public AmbientOcclusionNode(Context context) {
         ssaoFbo = requiresFBO(new FBOConfig(SSAO_FBO_URI, FULL_SCALE, FBO.Type.DEFAULT), displayResolutionDependentFBOs);
         addDesiredStateChange(new BindFbo(ssaoFbo));
         addDesiredStateChange(new SetViewportToSizeOf(ssaoFbo));
-        update(); // Cheeky way to initialise outputFboWidth, outputFboHeight
-        displayResolutionDependentFBOs.subscribe(this);
+        propertyChange(null); // Cheeky way to initialise outputFboWidth, outputFboHeight
+        displayResolutionDependentFBOs.subscribe(DisplayResolutionDependentFBOs.POST_FBO_REGENERATION, this);
```
emanuele3d: Let's remove the `DisplayResolutionDependentFBOs.` part please. Had it been a short string we could have left it, like we do with FBO.HDR for example. But in this case the most important information is the event type and POST_FBO_REGENERATION gives a good amount of information already.

engine/src/main/java/org/terasology/rendering/dag/nodes/AmbientOcclusionNode.java
```diff
@@ -156,7 +157,8 @@ public void process() {
     }
 
     @Override
-    public void update() {
+    public void propertyChange(PropertyChangeEvent event) {
+        // The only property we are subscribing to is DisplayResolutionDependentFBOs.POST_FBO_REGENERATION
```
emanuele3d:
```
// The only property we are subscribing to is DisplayResolutionDependentFBOs.POST_FBO_REGENERATION,
// which means there is no need to check or process the event object.
```

engine/src/main/java/org/terasology/rendering/dag/nodes/BlurredAmbientOcclusionNode.java
```diff
@@ -103,7 +105,8 @@ public void process() {
     }
 
     @Override
-    public void update() {
+    public void propertyChange(PropertyChangeEvent event) {
+        // The only property we are subscribing to is DisplayResolutionDependentFBOs.POST_FBO_REGENERATION
```
emanuele3d: Same comment as above...

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
```diff
@@ -278,39 +278,40 @@ public void process() {
...
-                if (parallaxMappingIsEnabled) {
-                    removeDesiredStateChange(setTerrainHeightInputTexture);
-                }
-            }
-        } else if (event.getPropertyName().equals(RenderingConfig.PARALLAX_MAPPING)) {
-            parallaxMappingIsEnabled = renderingConfig.isParallaxMapping();
-            if (normalMappingIsEnabled) {
-                if (parallaxMappingIsEnabled) {
-                    addDesiredStateChange(setTerrainHeightInputTexture);
+        if (event.getPropertyName().equals(DisplayResolutionDependentFBOs.PRE_FBO_REGENERATION)) {
```
emanuele3d: 1. Let's store getEventPropertyName() once in a string and reuse that.
2. Let's remove the DRSDFBOs string, as mentioned earlier, importing the constants directly.

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
```diff
@@ -278,39 +278,40 @@ public void process() {
...
+            refractiveReflectiveFbo.detachDepthBuffer();
+        } else if (event.getPropertyName().equals(DisplayResolutionDependentFBOs.POST_FBO_REGENERATION)) {
+            lastUpdatedGBuffer.attachDepthBufferTo(refractiveReflectiveFbo);
+        } else { // Else: One of the settings was changed, handle the event and refresh the task list.
+            // This method is only called when oldValue != newValue.
```
emanuele3d: Except there are no oldValue/newValue variables in use anywhere in the method. In this case I'd remove this line, unless we can make it more meaningful in this context.

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
```diff
@@ -278,39 +278,40 @@ public void process() {
...
+                    if (parallaxMappingIsEnabled) {
+                        removeDesiredStateChange(setTerrainHeightInputTexture);
+                    }
                 }
-            }
-        } else if (event.getPropertyName().equals(RenderingConfig.ANIMATE_WATER)) {
-            animatedWaterIsEnabled = renderingConfig.isAnimateWater();
-        } // else: no other cases are possible - see subscribe operations in initialize().
+            } else if (event.getPropertyName().equals(RenderingConfig.PARALLAX_MAPPING)) {
+                parallaxMappingIsEnabled = renderingConfig.isParallaxMapping();
```
emanuele3d: This is an issue inherited from the previous version, but I think this line can go one line below, after the `if(normalMappingIsEnabled)` line.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFbo.java
```diff
@@ -80,8 +82,8 @@ public SetInputTextureFromFbo(int textureSlot, FBO fbo, FboTexturesTypes texture
 
         this.material = getMaterial(materialUrn);
 
-        update(); // Cheeky way to initialise textureId
-        fboManager.subscribe(this);
+        propertyChange(null); // Cheeky way to initialise textureId
+        fboManager.subscribe(DisplayResolutionDependentFBOs.POST_FBO_REGENERATION, this);
```
emanuele3d: Same comment as above.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFbo.java
```diff
@@ -108,7 +110,7 @@ public SetInputTextureFromFbo(int textureSlot, SimpleUri fboUri, FboTexturesType
 
         this.material = getMaterial(materialUrn);
 
-        update(); // Cheeky way to initialise textureId
+        propertyChange(null); // Cheeky way to initialise textureId
         fboManager.subscribe(this);
```
emanuele3d: Shouldn't we subscribe to the POST_FBO_REGENERATION event here too?

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFbo.java
```diff
@@ -154,7 +156,8 @@ private int fetchTextureId() {
      * This method refreshes the task's reference to the FBO attachment, so that they are always up to date.
      */
     @Override
-    public void update() {
+    public void propertyChange(PropertyChangeEvent event) {
+        // The only property we are subscribing to is DisplayResolutionDependentFBOs.POST_FBO_REGENERATION
```
emanuele3d: Same comment as above.

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFbo.java
vampcat: This is weird.
In my workspace, I already subscribe to POST_FBO_REGENERATION.

engine/src/main/java/org/terasology/rendering/dag/nodes/RefractiveReflectiveBlocksNode.java
vampcat: Nice catch :D

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFbo.java
emanuele3d: Strange. It seems ok now as I review your last commit. Never mind then.

emanuele3d: Oooh. This is weird again. If I look at the diff of the individual commit it seems ok. But if I see all the changes at once, the problem is still there. Weird. Let me do some additional checks..

emanuele3d: Ok, I think we are getting confused because there are two nearly identical constructors, which also points at why duplicated code is bad. Specifically, by the look of it, the second constructor should be calling the first instead of duplicating all the work.

vampcat: Oh boy, I feel ashamed.

Either way, it should be fixed now :)

GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/stateChanges/SetInputTextureFromFbo.java
```diff
@@ -103,16 +103,7 @@ public SetInputTextureFromFbo(int textureSlot, FBO fbo, FboTexturesTypes texture
...
-        this.textureType = textureType;
-        this.fbo = fboManager.get(fboUri);
-        this.materialUrn = materialUrn;
-        this.shaderParameterName = shaderParameterName;
-
-        this.material = getMaterial(materialUrn);
-
-        propertyChange(null); // Cheeky way to initialise textureId
-        fboManager.subscribe(this);
+        this(textureSlot, fboManager.get(fboUri), textureType, fboManager, materialUrn, shaderParameterName);
```
emanuele3d: As I look at this, and the lack of checks on the arguments, I wonder how defensive would be good to be when we open the API to modules. Let's cross that bridge when we get there, but my point here is that a custom node might use this and other state change objects, and if they are used incorrectly, i.e. passing wrong or null arguments, we might want to make sure the developer gets some good error logging about it.

emanuele3d:

Cervator:
Tested this out, but caught an issue! 

2017-10-29 23_59_48-terasology - c__dev_terasology_git_terasology - terasology - _module groo

Seems to be specific to something enabled on Ultra. Switch to something lower and it goes away. Back to Ultra and it comes back. Reset back to before this PR and Ultra looks fine :-)

Probably something simple.

vampcat:
@Cervator I'm not entire sure what you wanted to highlight with that image, so it's possible I might have missed it :D However, I did run the game on ultra, and came across the weird horizontal lines. Fixing them was easy enough - one line.

However, I also took the opportunity to fix one or two more things, so I would like to get another review from @emanuele3d before merging this.

GooeyHub:

emanuele3d:
In c3470d0 I was able to reproduce the bug @Cervator reported. 

However, d2d1c77 or one of the other two latest commits seems to have fixed it. @Cervator, can you confirm? 

Other than that I have reviewed the changes in the latest three commits and I'm still GO for merge.

Cervator:
Can confirm the fix - merged! :-)

For a bit more info that earlier screenie I posted wasn't in nighttime- that was full daylight :D

Looked normal again now, although SSAO still seems odd at times to me. Probably not related to this PR though.

Ultra + SSAO:

terasology-171030122223-576x360

Ultra - SSAO

terasology-171030122254-576x360

emanuele3d:
Actually, @vampcat I just pulled the latest develop (post-merge) and SSAO is broken again. Same behaviour as the first screenshot @Cervator posted.

emanuele3d:
To be clearer: just enabling SSAO shows a scene with LightShafts and water surface but black landscape.

vampcat:
Can you please let me know what your exact settings are?

I cannot get the black-landscape bug thing to show up.

Cervator:
It is bizarre - reminds me of that past issue where something would go wonky with rendering after changing graphic settings, but only *sometimes*

Makes me wonder if I had something in my `config.cfg` that was bypassing it, yet now that I've wiped my settings it shows up each time and I haven't been able to find the combo that looks right.

I'll keep an eye on it while testing unrelated stuff.

PR #3134, opened 2017-10-28
Add update and update-all functionalities to modules.groovy

mjuvekar7:
Ref: #3099

### DIFF
--- a/module.groovy
+++ b/module.groovy
@@ -171,7 +171,7 @@ def updateModule(String name) {
     }
 
     println "Updating module $name"
-    moduleGit.pull remote: "origin", branch: "master"
+    moduleGit.pull remote: "origin"
 }
 
 /**
--- a/module.groovy
+++ b/module.groovy
@@ -151,6 +151,30 @@ def createModule(String name) {
 }
 
 /**
+ * Update a given module.
+ * @param name the name of the module to update
+ */
+def updateModule(String name) {
+    println "Attempting to update module $name"
+    File targetDir = new File("modules/$name")
+    if (!targetDir.exists()) {
+        println "Module \"$name\" not found"
+        return
+    }
+
+    def moduleGit = Grgit.open(dir: targetDir)
+    def clean = moduleGit.status().clean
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

@mjuvekar7: nice! I was hoping it would be easy like that :-)

GooeyHub:

module.groovy
```diff
@@ -151,6 +151,30 @@ def createModule(String name) {
...
+    def moduleGit = Grgit.open(dir: targetDir)
+    def clean = moduleGit.status().clean
+    println "Is \"$name\" clean? $clean"
+    if (!clean) {
+        println "Module has uncommitted changes. Aborting."
+        return
+    }
+
+    println "Updating module $name"
+    moduleGit.pull remote: "origin", branch: "master"
```
Cervator: A thought here: it isn't guaranteed that the user has the `master` branch checked out. Plenty of modules will work out of the `develop` branch and I would think if we hard coded `master` and a user is on `develop` they'll end up pulling from a branch they haven't got checked out.

IIRC you can use `git pull` without a branch argument and that'll simply pull whatever you have checked out. Maybe there is a variant like that we can use instead? Try to test it out  - what happens if the user has `develop` checked out and how can we better support that? :-)

mjuvekar7: Oh I thought most modules are developed directly on master. Will change it.

GooeyHub:

GooeyHub:

Cervator:
Awesome! Tested this by running `goovyw module update-all` in my mega-workspace, ran though about a hundred modules catching the few with local changes and so on. In less than a minute!

One fun flaw it pointed out: the few old module repos that still have an out of date `.gitignore` may fail the update because it thinks the `build` dir being present (not in the older `.gitignores`) means the workspace is dirty. So we should fix those soonish :D

I added in the two new usage entries while merging, easy enough. Thanks again :-)

PR #3136, opened 2017-10-29
Enhance run_linux.sh to pass forward command line arguments

### LINKED ISSUE #3068: Enhance game executable scripts to pass forward command line arguments
As @nihal111 just noticed it doesn't actually appear that our executable scripts accept command line arguments, say if you're trying to launch the game with `-homedir` on Linux. The Windows exe works fine as does running via command line using `java ...`

When I was working on #2979 to get the Groovy Wrapper working I noticed the `gradlew` script (which my `groovyw` is just a slightly modified copy of) has extensive handling for command line arguments, environment settings, etc. Maybe we can just copy a small part of that into the `run_linux.sh` (and `run_macosx.command` if somebody can test on a Mac) so any command line arguments passed will be passed forward and placed `... -jar libs/Terasology.jar <HERE>`

Probably not very hard since that is a very common need so finding examples be it inside `gradlew` or on Google should be easy. However it does require access to a Linux and/or Mac system for testing.

Naman-sopho:
Fixes #3068

### DIFF
--- a/engine-tests/src/test/java/org/terasology/entitySystem/PojoEventSystemTests.java
+++ b/engine-tests/src/test/java/org/terasology/entitySystem/PojoEventSystemTests.java
@@ -23,8 +23,10 @@ import org.terasology.engine.SimpleUri;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.entity.internal.PojoEntityManager;
 import org.terasology.entitySystem.event.AbstractConsumableEvent;
+import org.terasology.entitySystem.event.Event;
 import org.terasology.entitySystem.event.EventPriority;
 import org.terasology.entitySystem.event.ReceiveEvent;
+import org.terasology.entitySystem.event.internal.EventReceiver;
 import org.terasology.entitySystem.event.internal.EventSystemImpl;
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Huh, that little? Neat! Thanks @Naman-sopho :-)

Can anybody with a Linux system handy please validate this? And maybe even see if the same fix works on the Mac executable as well? I forget who has a Mac.

OvermindDL1:
The `"$@"` bit is standard posix so will work on linux and mac both (as well as about anything except windows).  The `$@` variable is an 'array of arguments' passed in the current scope (whether at script-file level or a function or so), and you cannot just do `$@` in the file or it would expand things like `blah "one two" bleep` to become `blah one two bleep` in a string context, but if you wrap an array expansion with double-quotes like `"$@"` then it expands it to `"blah" "one two" "bleep"`, thus always doing what you expect.

So yes, this is correct.  :-)

Cervator:
@OvermindDL1 thanks! :-)

@Naman-sopho could you also apply that tiny change to the Mac script as well please?

Naman-sopho:
@Cervator I am on it.

Issue #3126, opened 2017-10-18
"Dropped items glitch!"


noxarivis:
### What you were trying to do
Play the game
### What actually happened
after loading i had less than 1 fps,
i saw this glitch
![bildschirmfoto von 2017-10-18 12-30-56](https://user-images.githubusercontent.com/30774451/31715128-e7535606-b402-11e7-92ba-9b94fa347fe7.png)
![bildschirmfoto von 2017-10-18 12-32-50](https://user-images.githubusercontent.com/30774451/31715180-0f2d5e2e-b403-11e7-9d92-0dd521936345.png)

### How to reproduce

* Drop several irems
* close the game
* load the game

### Log details and game version
no logs yet,
Dev 2060
Stbl 72
### Computer details
/dev/Acer Aspire 3
RAM 8GB, Swap 8GB
Intel Core i3-6006U
HD-Graphics 520   

OS: Fedora 26 WS (Gnome 3.25.2 | Linux: 4.13.5-200.fc26.x86_64)   
Java: jdk1.8.0_141 (The default one of Netbeans)

pollend:
Can you explain what is happening. Is this a result of poor framerate or is there something wrong with the physics component of the game. Can you expand on exactly what is happening.

jellysnake:
Ah, I know this issue. It seems like #3043 has returned from the grave.

noxarivis:
@pollend @jellysnake
The framerate is like 0,01FPS and I confirm that thease where items i held, so it might be the reincarnation of #3043

noxarivis:
but it might also be the dropped items

noxarivis:
BTW, i might help fixing bugs if i can, i kinda just started with java.
I just wait for Apache Netbeans 9 to be released.

emanuele3d:
@portokaliu, this seems to be identical to the bug you fixed in #3049. Can you take a look?

portokaliu:
@emanuele3d I will look into this tomorrow evening

noxarivis:
any news?

portokaliu:
I did look into it, but no solution as of yet. Still on it though. It might be the reworking of the "in hand" item dealio

noxarivis:
its the only bug i found so fare, else its pretty stable

jellysnake:
Hah, you've been lucky with your module choices it seems.

noxarivis:
i use the default setting

bbarker:
[...]

PR #3137, opened 2017-10-31
Create a new item that contains only the necessary visual components …

portokaliu:
This should be a fix to #3126

I've reworked a bit of the system to use as few new entities as really needed. No more items will be lost and no objects floating in the world upon loading.

The downsides are:
 - There's always gonna be a null warning when switching to a block type item because of the way it's constructed from the prefab.
 - Any new visual stuff that we might want later to be visible first person will have to be manually implemented here (via adding the component) (This can be avoided by making the visual related components inherit a "Visual Component", but currently the system doesn't store components properly for such items (usually blocks), so that implementation will be postponed (unless there's time to look into that as well)

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
@@ -206,7 +206,7 @@ public class FirstPersonClientSystem extends BaseComponentSystem implements Upda
         // ensure that there are no lingering items that are marked as still held. This situation happens with client side predicted items
         for (EntityRef entityRef : entityManager.getEntitiesWith(ItemIsHeldComponent.class)) {
             if (!entityRef.equals(currentHeldItem) && !entityRef.equals(handEntity)) {
-                entityRef.removeComponent(ItemIsHeldComponent.class);
+                entityRef.destroy();
             }
         }
 
--- a/engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
@@ -62,9 +62,6 @@ public class FirstPersonClientSystem extends BaseComponentSystem implements Upda
     // the item from the inventory synchronized with the server
     private EntityRef currentHeldItem = EntityRef.NULL;
 
-    // the client side entity representing the hold item
-    private EntityRef clientHeldItem = EntityRef.NULL;
-
     private EntityRef getHandEntity() {
         if (handEntity == null) {
             // create the hand entity
@@ -103,7 +100,7 @@ public class FirstPersonClientSystem extends BaseComponentSystem implements Upda
             CharacterHeldItemComponent characterHeldItemComponent = localPlayer.getCharacterEntity().getComponent(CharacterHeldItemComponent.class);
             if (characterHeldItemComponent != null) {
                 // special case of sending in null so that the initial load works
-                linkHeldItemLocationForLocalPlayer(characterHeldItemComponent.selectedItem, null);
+                linkHeldItemLocationForLocalPlayer(characterHeldItemComponent.selectedItem);
             }
         }
     }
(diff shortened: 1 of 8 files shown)


GooeyHub:

engine/src/main/java/org/terasology/logic/players/FirstPersonClientSystem.java
```diff
@@ -103,7 +100,7 @@ public void ensureHeldItemIsMountedOnLoad(OnChangedComponent event, EntityRef en
             CharacterHeldItemComponent characterHeldItemComponent = localPlayer.getCharacterEntity().getComponent(CharacterHeldItemComponent.class);
             if (characterHeldItemComponent != null) {
                 // special case of sending in null so that the initial load works
-                linkHeldItemLocationForLocalPlayer(characterHeldItemComponent.selectedItem, null);
+                linkHeldItemLocationForLocalPlayer(characterHeldItemComponent.selectedItem);
```
aaron-harris: The comment immediately above this now appears to be out of date.

noxarivis:
merge pls

jellysnake:
@bbarker, this is the fix in question.

bbarker:
@jellysnake thanks!

I was able to cherry pick and merge this on top of 1.5.0 and it seems to run. I'd like to drop it in as a replacement to my existing stable/Launcher install of Tersalogy if possible for further testing (and fun). As I'm on Linux this looks to be under `~/.terasology`. However, I'm not sure exactly which jars should be replaced. In my build directory I have just a few jars:

```
$ find ./ -name '*.jar'
./engine/build/libs/engine-1.4.1-SNAPSHOT.jar
./engine/libs/jopenvr.jar
./engine-tests/build/libs/engine-tests-1.4.1-SNAPSHOT.jar
./facades/PC/build/libs/PC-1.4.1-SNAPSHOT.jar
./facades/TeraEd/build/libs/TeraEd-1.4.1-SNAPSHOT.jar
./gradle/wrapper/gradle-wrapper.jar
./modules/BuilderSampleGameplay/build/libs/BuilderSampleGameplay-1.4.1-SNAPSHOT.jar
./modules/Core/build/libs/Core-1.4.1-SNAPSHOT.jar
./modules/CoreSampleGameplay/build/libs/CoreSampleGameplay-1.4.1-SNAPSHOT.jar
```
They appear to be mislabeled in their filename; this was checked out from 1.5.0, but anyway, I suspect I should just copy `./engine/build/libs/engine-1.4.1-SNAPSHOT.jar` to `./RELEASE/TerasologyStable/72/libs/engine-1.5.0.jar` in this case? Maybe also `./facades/TeraEd/build/libs/TeraEd-1.4.1-SNAPSHOT.jar` to `./RELEASE/TerasologyStable/72/libs/Terasology.jar`?

portokaliu:
Fixed it according to the mentioned issues. Now there shouldn't be any null warnings present and any module can add visual stuff to the item as long as the component implements `VisualComponent`

GooeyHub:

jellysnake:
@bbarker I'm actually not sure. @Cervator might know. Not sure who else. I presume that copying those should work but I'm not fully sure.

bbarker:
@jellysnake @Cervator Cool, not a big deal but was thinking that might be the way to go about testing (and then making) a 1.5.1 release - correct me if I'm wrong.

Cervator:
So we haven't really needed true hot fixes to past releases before, but maybe this would be an interesting experiment :-)

I'm not sure what happened with those `v1.4.1-SNAPSHOT` jars - those would have come from a develop build before Alpha 8 / v1.5.0 was released. So some point before https://github.com/MovingBlocks/Terasology/commit/e5d5b7355ebc00e2270cf157c03370124a8ebb99

The game Launcher maintains two dirs: installs and data. Each version of the game it maintains gets its own, or at least its own install dir, I forget.

The jars in the v1.5.0 dir there should be updated to make the game executed by the launcher use a hot fix like that. The install + data dir defaults are probably both under your home dir.

To do a proper hot fix we'd branch off master or the v1.5.0 tag (same thing at the moment) to add a single commit to fix this on top. Interestingly that would produce v1.5.1 - which is also the current *next* release that would come out of `develop` if it were released (realistically it would become v1.6.0 as we've added more than just fixes by now, but that determination is usually not made until we release)

I'd guess we would make such a branch then build, test, and release v1.5.1 (again pushing to `master` and making a tag), then immediately after merge it to `develop` and bump it to `v1.5.2-SNAPSHOT` or even `v1.6.0-SNAPSHOT` as we know by now the next release will have new features and not just patch fixes. We just haven't needed to keep the version number updated between releases before.

Adhering to SemVer is fun!

As to the actual bug at hand yay simplifying code. Although is introducing `VisualComponent` as a child interface of `Component` that other components sometimes implement a risk to the trait-based architecture? Since the component hierarchy is meant to be very flat. I don't know if it makes sense to have some special cases like this. It might? I'd be curious to hear from others that are more architecture minded :-)

For the testing I can say things are better (again) - but unless my workspace is dusty I was still able to provoke it, just back down at 0,0,0 :-)

After one reload, single player:

terasology-171106173503-1152x720

And a second:

terasology-171106173539-1152x720

I could also replicate it for the *host* in a server-client + client multiplayer session, but not for the client, and I expect headless would be entirely free from the issue. So we're back to about 99% fixed + merge-worthy IMHO. Unsure if it is worth doing a full round of hotfix release experimenting, although we rarely get such opportunities so maybe :-)

TL;DR: Nearly fixed, happy to merge it, but wouldn't mind hearing a bit more feedback first

portokaliu:
Right, about that. Was too focused on the "not appear where you load" and not breaking inventory and stuff that I managed to forget about the core issue that brought this all up.

If making a child interface isn't a good idea, we can always go back to manually doing everything, or if anyone has a better idea I'm open to suggestions. Will also think about what other cases might there be for child interfaces and how it might impact code. Certainly found my fair share of anomalies.

Should have another update today to fix that, think I've found a reason for it existing but I'll investigate why the save system still wants to keep the item...

portokaliu:
I've submitted a fix that bypasses the issue with FP items still being saved. Still need to look into why they're saved in the first place, but that may be in an unrelated issue

GooeyHub:

Cervator:
Poked about some more and couldn't find any issues. Yay! Maybe this one is finally done for good :-)

Thanks @portokaliu and others!

Issue #3055, opened 2017-08-09
Accidentally changing the view distance can cause bad performance and a crash


Vizaxo:
### The problem

For a while I've had an intermittent problem where sometimes the game would get slow, start stuttering, and then eventually crash with an OutOfMemoryError or other GC problem. I thought it was something to do with the changes I was making to entities or something with Dynamic Cities, but I reproduced it on a fresh build on the flat world today

Then I realised that I'd accidentally hit `Home` a couple of times, setting my view distance to extreme. This was causing the memory to spike very high, up to the maximum, and the stuttering was caused by the GC going crazy trying to keep the game alive. When the memory got too high the game couldn't cope, and crashed.

### The solution(s)

I think there are really 2 problems here: one is that I could change the view distance and not notice, and the other is that setting the view distance too high caused a crash, rather than gracefully degrading.

I think that a message should pop up when the view distance changes to inform the player of what happened (and/or maybe changing the view distance without going into the menu should be disabled by default, to prevent confusion if the button is accidentally pressed).

There should also probably also be a way of detecting that the current view distance is unsustainable, warning the player that they haven't allocated enough memory, and automatically degrading to a lower view distance instead of crashing.

Cervator:
I would consider the following a decent definition of done:

* To increase visibility log a change of the view distance to the chat bar area. That happens already when a player enters or leaves a multiplayer session (along with a few other cases). Should be easy to run two instances of the game, look for the exact string that having player two join causes, then search the code base to find where it is to get a good example :-)

As a bonus make the message include a warning if the "MEGA" or "EXTREME" view distance is selected (neither is really expected to be viable for the vast majority of PCs)

I had posted about an issue in the past about better marking the view distances / marking some programatically to be unusually heavy, but am having trouble finding that issue.

Cervator:
Closing this as completed by #3140 - the harsh impact to performance still remains but is now vastly more visible. The performance issue is a bigger piece I think we'll naturally improve over time.

PR #3140, opened 2017-11-11
Fix for issue #3055.

floppypanda:
### Contains
This is a fix for issue #3055. 

### How to Test
The fix may be tested using the following steps:

1. Enter a world in Terasology.
2. Increase the view distance by pressing the appropriate key binding (by default the "home" key).
3. A message should pop up in the chat window stating that the view distance has been increased.
4. Continue to increase the view distance until it reaches the "Mega" setting.
5. A warning message should pop up in the chat window regarding possible performance issues.
6. Increase the viewing distance to the "Extreme" setting.
7. Another warning message regarding possible performance issues should appear.
8. Decrease the view distance by pressing the appropriate key binding (by default the "end" key).
9. If all tests pass, the change should be safe to merge.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
@@ -21,12 +21,15 @@ import java.util.List;
 
 import org.lwjgl.input.Mouse;
 import org.terasology.config.Config;
+import org.terasology.context.Context;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.event.EventPriority;
 import org.terasology.entitySystem.event.ReceiveEvent;
 import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.entitySystem.systems.RegisterMode;
 import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.i18n.TranslationSystem;
+import org.terasology.i18n.TranslationSystemImpl;
 import org.terasology.input.Keyboard;
 import org.terasology.input.binds.general.HideHUDButton;
 import org.terasology.input.events.KeyDownEvent;
@@ -49,20 +52,26 @@ public class DebugControlSystem extends BaseComponentSystem {
 
     @In
     private WorldProvider world;
-
+    
     @In
     private Config config;
+    
+    @In
+    private Context context;
 
     @In
     private NUIManager nuiManager;
     
-    private DebugOverlay overlay;    
-
+    private DebugOverlay overlay;
+    
+    private TranslationSystem translationSystem;
+    
     private boolean mouseGrabbed = true;
 
     @Override
     public void initialise() {
         overlay = nuiManager.addOverlay("engine:debugOverlay", DebugOverlay.class);
+        translationSystem = new TranslationSystemImpl(context);
     }
 
     @ReceiveEvent(components = ClientComponent.class)
@@ -81,7 +90,7 @@ public class DebugControlSystem extends BaseComponentSystem {
     /**
      * Increases view distance upon receiving an increase view distance event.
      * @param button The button or key pressed to increase view distance.
-     * @param entity The entity corresponding to event.
+     * @param entity The player entity that triggered the view distance increase.
      */
     @ReceiveEvent(components = ClientComponent.class)
     public void onIncreaseViewDistance(IncreaseViewDistanceButton button, EntityRef entity) {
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Hi @floppypanda and thanks! :-)

Will submit a review momentarily, need to trigger our build bot in a normal comment:

@GooeyHub ok to test

engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
```diff
@@ -51,8 +55,8 @@
 
     @In
     private NUIManager nuiManager;
-
-    private DebugOverlay overlay;
+
```
Cervator: Might want to check for stray whitespace like this, since it doesn't really serve a difference other than making the diff bigger :-)

engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
```diff
@@ -74,24 +78,45 @@ public void onHideHUD(HideHUDButton event, EntityRef entity) {
         }
     }
 
+    /**
+     * Increases view distance upon receiving an increase view distance event.
+     * @param button The button or key pressed to increase view distance.
+     * @param entity The entity corresponding to event.
```
Cervator: Hooray for adding javadoc! For the `entity` parameter I'm guessing it is the player/client that triggered the event, maybe that would read better in the text? Like "Player entity that triggered the increase" (do confirm that's actually the case - I"m just speculating)

engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
```diff
@@ -74,24 +78,45 @@ public void onHideHUD(HideHUDButton event, EntityRef entity) {
...
     public void onIncreaseViewDistance(IncreaseViewDistanceButton button, EntityRef entity) {
         int viewDistance = config.getRendering().getViewDistance().getIndex();
         int maxViewDistance = ViewDistance.values().length - 1;
-
+        
+        //Ensuring that the view distance does not exceed its maximum value.
         if (viewDistance != maxViewDistance) {
-            config.getRendering().setViewDistance(ViewDistance.forIndex((viewDistance + 1)));
+            ViewDistance greaterViewDistance = ViewDistance.forIndex(viewDistance + 1);
+            config.getRendering().setViewDistance(greaterViewDistance);
```
Cervator: Maybe move the actual setting of the view distance below the logging? That way if anything fails during the view distance change the chat panel should show what's about to happen, even if briefly (for the case where a player might have hit the key accidentally and didn't realize how they crashed the game)

You can tweak the notification message slightly to something like "Increasing view distance to:"

engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
```diff
@@ -74,24 +78,45 @@ public void onHideHUD(HideHUDButton event, EntityRef entity) {
...
+            fireChangeEvent("View distance increased to " + greaterViewDistance.toString() + ".", Arrays.asList(entity));	
+            //Presenting user with a warning if the view distance is set higher than recommended.
+            if (greaterViewDistance == ViewDistance.MEGA || greaterViewDistance == ViewDistance.EXTREME) {
+            	fireChangeEvent("Warning: Setting view distance to " + greaterViewDistance.toString() 
+            	                    + " may result in performance issues.", Arrays.asList(entity));
```
Cervator: Somewhat puzzled here why you wrap the entity in a List? :-)

It can only ever be one entity I'm thinking, and later on each element in the list is used for the event sending. Is there anything to it I'm missing, or maybe you started that way for some reason that later went away?

GooeyHub:

Cervator:
Screenie for the internationalization text:

2017-11-12 19_53_43-terasology _ alpha

GooeyHub:

floppypanda:
I have made changes which update the Javadoc entity parameter descriptions, log view distance changes before they are applied, and translate the view distance strings. I originally decided to have the fireChangeEvent method take a list of entities instead of a single entity in anticipation of other methods within the DebugControlSystem needing to send notifications to multiple entities in the future. The case that I had in mind was debugging a server instance in which we would like to confirm that a parameter change is picked up by the server's clients. If you believe that passing a single entity would be more appropriate, I would be happy to oblige.

Cervator:
Thanks for the changes!

Good thought on the list of entities that way, if there were other users. You got me looking at the class a bit more which made me realize that I have no idea what the view distance changing events are doing in there in the first place - they don't require debug mode (F3) enabled at all. Maybe an oversight from the past. I've submitted separate issue #3141 for that to get some attention. Can also try to determine if we really need that utility method or not via that issue :-)

Merged this and added you to the credits! Feel free to jump on the new issue if interested since you've already worked in that class, or find anything else that looks interesting

PR #3142, opened 2017-11-15
Disable Load and Delete buttons until a game has been selected

jameshiew:
### Contains

Disable the 'Load/Delete' buttons on the select game screen until a savegame has been selected from the menu to make the UI a bit clearer.

### How to test

The buttons should be greyed out if there are no savegames present, or on first transitioning to the select game screen before a savegame has been selected.

It could be merged as is but if a user selects a savegame then deletes it, then transitions away from the select game screen then back to it, the 'Load/Delete' buttons are still enabled.


GooeyHub:

Cervator:
Nice little change @jameshiew - thanks!

Will trigger our build bot for review and check this out soon :-)

@GooeyHub: ok to test

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/SelectGameScreen.java
```diff
@@ -83,6 +84,11 @@ public String get() {
 
         refreshList(gameList);
         gameList.subscribe((widget, item) -> loadGame(item));
+        gameList.subscribeSelection((widget, item) -> {
+            for (String targetId : new String[]{"load", "delete"}) {
+                find(targetId, UIButton.class).setEnabled(true);
```
msteiger: Does that still work if you close the menu and open it again? I would probably tie the selection explicitly to the question whether an item in the list is selected or not. This will take care of disabling them at the right time, too.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/SelectGameScreen.java
```diff
@@ -83,6 +84,11 @@ public String get() {
 
         refreshList(gameList);
         gameList.subscribe((widget, item) -> loadGame(item));
+        gameList.subscribeSelection((widget, item) -> {
+            for (String targetId : new String[]{"load", "delete"}) {
```
msteiger: I would try to keep it simple: assign both to a member variable and use them later.

jameshiew:
Thanks for the feedback, unfortunately I can't finish off this PR but someone else might want to implement the last changes

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/SelectGameScreen.java
skaldarnar: `.setEnable(item != null)` is probably all change needed here.

oniatus:
Added the requested change and did some smaller cleanups :)
Tested moving back and forth from the screen (item will stay selected if was before), deleting a game (button gets disabled) and re-opening screen without selecting a game (buttons stay disabled)
@skaldarnar ready to merge from my side.

GooeyHub:

mkienenb:
I have fixed the merge conflicts.  However, there is still a bug with this PR.

- Create a new world
- Exit to Main Menu
- Go to SinglePlayer (not actually one word!)

Note that the new world's game is selected, but the Load and Delete buttons are disabled.

GooeyHub:

Cervator:
I'm guessing the final bug here relates to #3275 which auto-selects the first available save (if present) to speed up the UI flow a bit.  That somewhat clashes with this PR then since if would only ever trigger if there are no saves at all. Which might be fair? 

This might be fixable with another line added to the one-line PR #3275 that triggers whatever check this PR uses to see if something has been selected.

Still just a guess but I wanted to link the other PR. Maybe also related is #3301 which adds a confirmation popup for deletion (might have caused the conflict?)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/SelectGameScreen.java
```diff
@@ -85,26 +88,43 @@ public String get() {
         refreshList(gameList);
         gameList.select(0);
         gameList.subscribe((widget, item) -> loadGame(item));
+        gameList.subscribeSelection((widget, selectedGame) -> {
```
skaldarnar: My first guess for the problem described by @mkienenb is that the `select(0)` happens too early, i.e., before the subscription is set up. Moving `gameList.select(0);` a few lines lower might already fix the problem, as `select(...)` will notifiy listeners (see https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/widgets/UIList.java#L203-L204)

Cervator:
Closing this as superseded by #3361 as part of @ar0ne's GSOC project

Thanks again to everybody involved :-)

PR #3147, opened 2017-11-18
fix for unregistering a "general" event receiver

kaen:
### Contains

`unregisterEventReceiver` did not unregister receivers that had no components associated with them (general vs specific receivers). Adds a clause for handling this case (and a test to prove it)

### How to test

Run the test suite :3

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/event/internal/EventSystemImpl.java
+++ b/engine/src/main/java/org/terasology/entitySystem/event/internal/EventSystemImpl.java
@@ -241,6 +241,16 @@ public class EventSystemImpl implements EventSystem {
                 }
             }
         }
+
+        if (0 == componentTypes.length) {
+            Iterator<EventHandlerInfo> eventHandlerIterator = generalHandlers.values().iterator();
+            while (eventHandlerIterator.hasNext()) {
+                EventHandlerInfo eventHandler = eventHandlerIterator.next();
+                if (eventHandler.getHandler().equals(eventReceiver)) {
+                    eventHandlerIterator.remove();
+                }
+            }
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/entitySystem/event/internal/EventSystemImpl.java
```diff
@@ -241,6 +241,16 @@ private void addToComponentSpecificHandlers(Class<? extends Event> type, EventHa
                 }
             }
         }
+
+        if (0 == componentTypes.length) {
```
Cervator: Not a complaint at all but more an observation: we haven't really done [Yoda Conditions](https://en.wikipedia.org/wiki/Yoda_conditions) much before, but that might be more from just not having thought about them much. It seems like they have some decent pros. I'm unsure what I would go for if we try to think about what sort of convention we'd want to go with

engine/src/main/java/org/terasology/entitySystem/event/internal/EventSystemImpl.java
```diff
@@ -241,6 +241,16 @@ private void addToComponentSpecificHandlers(Class<? extends Event> type, EventHa
...
+            Iterator<EventHandlerInfo> eventHandlerIterator = generalHandlers.values().iterator();
+            while (eventHandlerIterator.hasNext()) {
```
Cervator: A little bird named IntelliJ whispered in my ear that this could be replaced with a magic incantation going `Collection.removeIf ...` - neat!

That doesn't get in the way of merging this though, but I thought it was a cool suggestion. Feel free to look for things like that if you're in a class anyway. Same goes for updating those annoying empty javadoc blocks, especially since you doc so well! :D

PR #3150, opened 2017-11-28
Enhance game executable scripts to pass forward command line arguments

### LINKED ISSUE #3068: Enhance game executable scripts to pass forward command line arguments

Naman-sopho:
Fixes #3068

### DIFF
--- a/facades/PC/launchScripts/run_linux.sh
+++ b/facades/PC/launchScripts/run_linux.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 cd "$(dirname "$0")"
-java -Xms128m -Xmx1536m -jar libs/Terasology.jar
+java -Xms128m -Xmx1536m -jar libs/Terasology.jar "$@"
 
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Merged this and added you to the game credits :-)

PR #3151, opened 2017-11-28
Tidy up / refactor DebugControlSystem.

### LINKED ISSUE #3141: Tidy up / refactor DebugControlSystem
Came up while reviewing #3140 - the DebugControlSystem (DCS) has some rather weird stuff in it that doesn't seem to have anything to do with debug mode.

It is a very old class with lots of history so it probably slowly morphed into what it is today. It lacks Javadoc so I'm not really sure what the intent was but as I interpret it the only thing it should be doing is related to unique actions only available when debug mode (F3) is enabled.

Things to review/change to complete this issue:

- [ ] Add class-level Javadoc and any other javadoc possible when fully sure what it all does
- [ ] Should `onHideHUD` really be there? It does getting into something with `getDebug()` but is that really valid? The player should be able to hide the HUD regardless of whether debug mode is on
- [ ] `onIncreaseViewDistance` and `onDecreaseViewDistance` should be moved somewhere more sensible. Definitely not something that requires debug mode and it doesn't reference it at all
- [ ] `fireChangedEvent` was added with #3140 including support for sending change events to a series of interested parties in case that became relevant in the class - but with it already doing more than it is supposed to maybe it'll never need it. Do we have any such utility somewhere else already?

I'm not sure *where* the view distance stuff should get moved to. May be relevant to @vampcat or @emanuele3d's interests since rendering config is involved.

The translation bits may no longer be needed in this class with the view distance changing moved out. They should be added if not already present in a suitable target system. Feel free to look for similar issues there - for instance changing the view distance at present also results in a simple log statement in `RenderableWorldImpl` that at present isn't using the translated value, just the key. Fix fix fix! :-)

Another maybe extra: Should the view distance warning notification for MEGA and EXTREME just be merged with the regular notification so there's only one? Just with the warning added at the end when appropriate.

Naman-sopho:
Fixes #3141 
 - [x] Add class-level Javadoc and any other javadoc possible when fully sure what it all does
- [x] Should onHideHUD really be there? It does getting into something with getDebug() but is that really valid? The player should be able to hide the HUD regardless of whether debug mode is on
(Discussed with @Cervator could not reach a conclusion on the right place for this to be moved)
 - [x] onIncreaseViewDistance and onDecreaseViewDistance should be moved somewhere more sensible. Definitely not something that requires debug mode and it doesn't reference it at all
- [ ] fireChangedEvent was added with #3140 including support for sending change events to a series of interested parties in case that became relevant in the class - but with it already doing more than it is supposed to maybe it'll never need it. Do we have any such utility somewhere else already?

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
@@ -87,51 +87,14 @@ public class DebugControlSystem extends BaseComponentSystem {
         }
     }
 
-    /**
-     * Increases view distance upon receiving an increase view distance event.
-     * @param button The button or key pressed to increase view distance.
-     * @param entity The player entity that triggered the view distance increase.
-     */
-    @ReceiveEvent(components = ClientComponent.class)
-    public void onIncreaseViewDistance(IncreaseViewDistanceButton button, EntityRef entity) {
-        int viewDistance = config.getRendering().getViewDistance().getIndex();
-        int maxViewDistance = ViewDistance.values().length - 1;
-        
-        //Ensuring that the view distance does not exceed its maximum value.
-        if (viewDistance != maxViewDistance) {
-            ViewDistance greaterViewDistance = ViewDistance.forIndex(viewDistance + 1);
-            String greaterViewDistanceStr = translationSystem.translate(greaterViewDistance.toString());
-            fireChangeEvent("Increasing view distance to " + greaterViewDistanceStr + ".", Arrays.asList(entity));
-            //Presenting user with a warning if the view distance is set higher than recommended.
-            if (greaterViewDistance == ViewDistance.MEGA || greaterViewDistance == ViewDistance.EXTREME) {
-                fireChangeEvent("Warning: Increasing view distance to " + greaterViewDistanceStr
-            	                    + " may result in performance issues.", Arrays.asList(entity));
-            }
-            config.getRendering().setViewDistance(greaterViewDistance);
-        }
-        button.consume();
-    }
 
     /**
-     * Decreases view distance upon receiving a decrease view distance event.
-     * @param button The button or key pressed to decrease view distance.
-     * @param entity The player entity that triggered the view distance decrease.
+     * Creates illusion of time flying by if correspondig key is held down.
+     * Up / Down : Increases / Decreases time of day by 0.005 per keystroke.
+     * Right / left : Increases / Decreases time of day by 0.02 per keystroke.
+     * @param event
+     * @param entity
      */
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Hey @Naman-sopho and thanks for this PR and the additional work you're doing :-)

One question though - did you try to test this locally by running from source? You have rather severe compilation problems including a stray `s` in the middle of nowhere plus excess/missing imports between the two files.

The `fireChangedEvent` also wasn't moved, yet is used by the code that was. 

You could also clear out otherwise empty javadoc + update parameter args within javadoc (one is named wrong). That's a general recommendation whenever you're working on something anyway and sticks out nicely with warnings and such in IntelliJ

Try it out and I'll check again when you've made updates :-)

Naman-sopho:
@Cervator Sorry about that.
I completely forgot about the pending work on this PR will get back right on it.

Naman-sopho:
@Cervator Please check it now and suggest any further changes if required.

Cervator:
So it compiles now, but it breaks when you try to change the view distance :-(

```
14:45:13.824 [Chunk-Generator-4] INFO  o.t.w.b.internal.BlockManagerImpl - Registered Block core:RedLeaf with id 40
14:45:32.293 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
	at org.terasology.rendering.world.viewDistance.ClientViewDistanceSystem.onDecreaseViewDistance(ClientViewDistanceSystem.java:128)
	at org.terasology.rendering.world.viewDistance.ClientViewDistanceSystemMethodAccess.invoke(Unknown Source)
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:516)
```

Are you testing the new code? That part is pretty critical :-)

I merged the branch, started a game, and pressed the `end` key to reduce the view distance by one and it triggered the above error. This is likely because you don't have the step in the new class the code lives in that initializes the `translationSystem` variable. But that should have shown up really easily during testing.

There are also still a few stray bits of unnecessary stuff, like unused imports in the debug system class, the `translationSystem` there is no longer used, empty javadoc, etc

Naman-sopho:
@Cervator 
I had something wrong with my IntelliJ config the Auto import was not working, so I had to manually figure out all the imports from the errors. So, some unused ones might have crept in.
Regarding the testing I did test by running the game and changing view distance worked fine. I might have tested from a wrong branch(currently working with three of them simultaneously :) ), my bad.

Cervator:
Tests out great now, thanks! And yep we have all been there. Branches are fun :D

PR #3153, opened 2017-11-30
generalized event handler for  OnBlockItemPlaced

pollend:
just realized a potential problem for the case when the player is holding onto an item that isn't a BlockItemComponent. Full block items do not have the BlockItemComponent component.


GooeyHub:

PR #3154, opened 2017-12-01
Hide FloatingTextComponent behind objects

eviltak:
Adds a `boolean` to the `FloatingTextComponent` called `isOccluded` that occludes the floating text if enabled. This patch simply enables/disables depth testing based on the value of `isOccluded`.

### How to test

Place a `FloatingTextComponent` behind an obstacle with `isOccluded` set to `true`. The text will not be visible. It will, however, be visible behind an obstacle with `isOccluded` set to `true`.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/logic/FloatingTextComponent.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/FloatingTextComponent.java
@@ -25,5 +25,5 @@ public class FloatingTextComponent implements VisualComponent {
     public Color textColor = Color.WHITE;
     public Color textShadowColor = Color.BLACK;
     public float scale = 1f;
-    public boolean isOccluded;
+    public boolean isOverlay;
 }
--- a/engine/src/main/java/org/terasology/rendering/logic/FloatingTextComponent.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/FloatingTextComponent.java
@@ -25,4 +25,5 @@ public class FloatingTextComponent implements VisualComponent {
     public Color textColor = Color.WHITE;
     public Color textShadowColor = Color.BLACK;
     public float scale = 1f;
+    public boolean isOccluded;
 }
--- a/engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
@@ -94,7 +94,6 @@ public class FloatingTextRenderer extends BaseComponentSystem implements RenderS
             // If the FloatingTextComponent is meant to be occluded, enable depth tests
             if (floatingText.isOccluded) {
                 glEnable(GL_DEPTH_TEST);
-                glDepthFunc(GL_LESS);
             } else {
                 glDisable(GL_DEPTH_TEST);
             }
--- a/engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
@@ -97,13 +97,6 @@ public class FloatingTextRenderer extends BaseComponentSystem implements RenderS
 
             FloatingTextComponent floatingText = entity.getComponent(FloatingTextComponent.class);
 
-            // If the FloatingTextComponent is meant to be occluded, enable depth tests
-            if (floatingText.isOverlay) {
-                glEnable(GL_DEPTH_TEST);
-            } else {
-                glDisable(GL_DEPTH_TEST);
-            }
-
             String[] linesOfText = floatingText.text.split("\n");
             Color baseColor = floatingText.textColor;
             Color shadowColor = floatingText.textShadowColor;
@@ -125,6 +118,11 @@ public class FloatingTextRenderer extends BaseComponentSystem implements RenderS
                                 shadowColor, underline);
                 entityMeshCache.put(entity, meshMap);
             }
+
+            if (floatingText.isOverlay) {
+                glDisable(GL_DEPTH_TEST);
+            }
+
             glPushMatrix();
 
             float scale = METER_PER_PIXEL * floatingText.scale;
@@ -146,8 +144,12 @@ public class FloatingTextRenderer extends BaseComponentSystem implements RenderS
             }
 
             glPopMatrix();
+
+            // Revert to default state
+            if (floatingText.isOverlay) {
+                glEnable(GL_DEPTH_TEST);
+            }
         }
-        glEnable(GL_DEPTH_TEST);
     }
 
     private void diposeMeshMap(Map<Material, Mesh> meshMap) {
--- a/engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
@@ -43,7 +43,13 @@ import org.terasology.world.WorldProvider;
 import java.util.Arrays;
 import java.util.Map;
 
-import static org.lwjgl.opengl.GL11.*;
+import static org.lwjgl.opengl.GL11.GL_DEPTH_TEST;
+import static org.lwjgl.opengl.GL11.glDisable;
+import static org.lwjgl.opengl.GL11.glEnable;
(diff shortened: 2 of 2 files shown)


GooeyHub:

Cervator:
@GooeyHub add to whitelist

@eviltak yay more stuff thanks! :-)

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
```diff
@@ -43,13 +43,7 @@
...
 import java.util.Map;
 
-import static org.lwjgl.opengl.GL11.GL_DEPTH_TEST;
-import static org.lwjgl.opengl.GL11.glDisable;
-import static org.lwjgl.opengl.GL11.glEnable;
-import static org.lwjgl.opengl.GL11.glPopMatrix;
-import static org.lwjgl.opengl.GL11.glPushMatrix;
-import static org.lwjgl.opengl.GL11.glScaled;
-import static org.lwjgl.opengl.GL11.glTranslated;
+import static org.lwjgl.opengl.GL11.*;
```
vampcat: Please, no star imports. Revert this back to the way it was.

You can change the lines of imports before it gets converted to * in intelliJ. Set that to a high value, like 999.

engine/src/main/java/org/terasology/rendering/logic/FloatingTextComponent.java
```diff
@@ -25,4 +25,5 @@
     public Color textColor = Color.WHITE;
     public Color textShadowColor = Color.BLACK;
     public float scale = 1f;
+    public boolean isOccluded;
```
vampcat: I'm not sure, but I feel like `isOccluded` is not the best name, because it feels like it is referring to the current state of the text. 'shouldBeOccluded' sounds better, but I'm not sold on the "occluded" part either.
@emanuele3d your thoughts?

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
```diff
@@ -99,6 +91,14 @@ private void render(Iterable<EntityRef> floatingTextEntities) {
...
             FloatingTextComponent floatingText = entity.getComponent(FloatingTextComponent.class);
 
+            // If the FloatingTextComponent is meant to be occluded, enable depth tests
+            if (floatingText.isOccluded) {
+                glEnable(GL_DEPTH_TEST);
+                glDepthFunc(GL_LESS);
+            } else {
+                glDisable(GL_DEPTH_TEST);
+            }
+
```
vampcat: Hmm.. you are enabling it, but where are you disabling it back?
Always try to revert system back to the state you found it in.

GooeyHub:

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
eviltak: Hmm, IntelliJ did this automatically for me, so I thought it must be in line with the Terasology code style. I will get rid of them.

engine/src/main/java/org/terasology/rendering/logic/FloatingTextComponent.java
eviltak: I was under the same impression, but the Google Code-in task required me to name the field `isOccluded`. A better option IMO would be to use a name like `isDepthTested` which directly relates to the field's function.

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
eviltak: Depth testing, by default, is enabled, otherwise _everything_ would be rendered on the screen. The method in its _previous_ and current state enables depth testing before it returns.

GooeyHub:

eviltak:
@Cervator Can you please approve the corresponding Google Code-in task so that I can move on to others? Thanks!

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
vampcat: > Depth testing, by default, is enabled, otherwise everything would be rendered on the screen.
Depth testing in OpenGL is not enabled by default.

That being said, you are correct, it does have a line to disable it back - it didn't appear in the diff, because it was untouched by you.

engine/src/main/java/org/terasology/rendering/logic/FloatingTextComponent.java
vampcat: I realize that the task mentioned `isOccluded` specifically.. but not having written that task myself and after having reviewed this, I'm fairly certain I'm in favour of renaming this.

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
eviltak: I meant that in the context of the Terasology rendering engine, where depth testing enabled should be the "default" state.

GooeyHub:

Cervator:
> @Cervator Can you please approve the corresponding Google Code-in task so that I can move on to others? Thanks!

Yeah that seems fair enough so long as you keep up with the review process here for any additional tweaks :-)

eviltak:
@vampcat approved the task, thanks.

engine/src/main/java/org/terasology/rendering/logic/FloatingTextComponent.java
emanuele3d: There are two approaches to this. 
1. rename it after what it does from a technical standpoint (depth testing-related)
2. rename it after what it does from a visual standpoint (acts as an overlay or is in-scene)
I'm in favor of option 2 simply because it's quite clear from the code what it does technically. 

So, if it was up to me I'd probably go for something like isOverlay, or isRenderedAsOverlay, effectively flipping the meaning of the variable. That been said, I'm not overly fussed about this.

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
emanuele3d: Depth Testing is indeed enabled by default in LWJGLGraphics.initOpenGLParams(): it is a Terasology default and is the opposite of the OpenGL default. 

That doesn't mean that's the way it "should" be, as only a limited number of nodes in the renderer actually requires depth testing enabled. Properly evaluating this aspect is not in the scope of this discussion albeit it is one of the things I'd like to do at some point.

More relevant is the way depth testing is enabled and disabled in this method. 

Right now depth testing is enabled or disabled in the first part of the method, within the outer for loop, and then "universally" re-enabled outside of it. I have a few comments:

1. rely on the default (depth testing enabled) and do nothing if the default is enough.
2. if depth testing must be disabled, re-enable it on a per-entity basis (two IF blocks)
2b. if you wanted to be really fancy, sort the entities first, to do the least enabling/disabling
3. move the code to enable/disable things as close as possible to the rendering code, i.e. just outside or inside the two glPush/PopMatrix statements.
4. If glDepthFunc(GL_LESS) is really needed (Terasology default is GL_LEQUAL) that can be set and reset to GL_LEQUAL outside of the for(entity) loop, just once. 

That been said, isn't LEQUAL enough?

emanuele3d: Oh. I just noticed the removal of the glDepthFunc(GL_LESS) line in the next commit. Never mind point 4 then!

engine/src/main/java/org/terasology/rendering/logic/FloatingTextComponent.java
eviltak: `isOverlay` is a nice choice, much better than `isOccluded`.

engine/src/main/java/org/terasology/rendering/logic/FloatingTextRenderer.java
eviltak: Sorting will be overkill, but I agree with the other two points. Changes incorporated.

GooeyHub:

GooeyHub:

Cervator:
Tested out fine - thanks! Merged :-)

PR #3155, opened 2017-12-02
adding scala-lib packages to whitelist

bbarker:
This is simply a partial fix for getting Scala working; we've established that Scala libraries need to be whitelisted. However, the libraries also need to be loaded, which is not fixed in this PR.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
+++ b/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
@@ -46,6 +46,14 @@ public final class ExternalApiWhitelist {
             .add("java.awt")
             .add("java.awt.geom")
             .add("java.awt.image")
+            // sub-packages of root for scala-lib: https://www.scala-lang.org/api/current/ 
+            .add("scala.collection")
+            .add("scala.concurrent")
+            .add("scala.io")
+            .add("scala.math")
+            .add("scala.sys")
+            .add("scala.util")
+            // end of scala-lib packages
             .add("jdk.internal.reflect")
             .add("com.google.common.annotations")
             .add("com.google.common.cache")
--- a/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
+++ b/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
@@ -49,7 +49,6 @@ public final class ExternalApiWhitelist {
             // sub-packages of root for scala-lib: https://www.scala-lang.org/api/current/ 
             .add("scala.collection")
             .add("scala.concurrent")
-            .add("scala.io")
             .add("scala.math")
             .add("scala.sys")
             .add("scala.util")


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

GooeyHub:

Cervator:
Curiously, would adding stuff like [Scala IO](http://www.scala-lang.org/api/2.12.3/scala/io/index.html) to the whitelist indirectly expose options for file access and such? Or would that still break (as intended) if you tried to reach it through some Scala class?

bbarker:
In my experience it will typically rely on java libs to do the heavy
lifting, in which case it will break, but I can remove scala.io from the
list for now to be safe.

bbarker:
@Cervator I removed scala.io for now

GooeyHub:

Cervator:
Hmm, on that note there might be more things to consider ... :D

I just looked at `scala.sys` real quick and there is an `exit()` in there and other rather fundamental sounding things (like hooking up on-exit threads to run). Sounds like that might also be a bit of a wasp's nest of potential Fun :-)

For instance right now if I place:

`System.exit(0);`

In a module the sandbox defeats it gracefully:

`java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "exitVM.0")`

And the game keeps running. That may also actually be a great way to test this. If you try a Scala style system exit, will the sandbox catch it when it defers to `System.exit` - if it does? If that's the case it may be safe - as long as we indirectly block things like thread manipulation and so on.

Not saying a system quit is necessarily a great example of a thing to be scared of. It may be somewhat of a question what we're actually trying to guard against. A modder could sneak in a system exit triggered by some obscure thing, get a module accepted in a variety of larger collections, then go crash servers? That could potentially corrupt worlds, even if the server OS itself remains safe.

bbarker:
@Cervator I wanted to test out your suggestions regarding security, but am running into further surprises that need addressing first (using Nanoware now, but not sure it matters). It seems that there is an issue somewhere with sandboxing in loading scala objects (similar to static classes in Java in usage, but maybe not in implementation under the hood - not sure), which end in a `$` [when used from Java](https://stackoverflow.com/questions/3845737/how-can-i-pass-a-scala-object-reference-around-in-java). E.g., to even use a Seq I got a `java.lang.NoClassDefFoundError: scala/Predef$` - this was fixed by adding `scala.Predef$.class` to the class whitelist on my hacked branch of Nanoware that loads Scala libraries directly in the engine, but since this class whitelist is directly referencing an object that is known statically, we may need to think of another workaround - the classes whitelisted use static references, but package references only use strings (dynamic whitelisting).

Another problem is that if I use a scala `List` instead of a scala `Seq`, I get this similar error:

```
07:47:37.055 [main] ERROR o.t.engine.ComponentSystemManager - Failed to load system ScalaLib:ScalaCommands
java.lang.NoClassDefFoundError: scala/collection/immutable/List$
        at org.terasology.ScalaCommands.<init>(ScalaCommands.scala:9)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
        at java.lang.Class.newInstance(Class.java:442)
        at org.terasology.engine.ComponentSystemManager.loadSystems(ComponentSystemManager.java:96)
        at org.terasology.engine.modes.loadProcesses.RegisterSystems.step(RegisterSystems.java:53)
        at org.terasology.engine.modes.StateLoading.update(StateLoading.java:242)
        at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:413)
        at org.terasology.engine.Terasology.main(Terasology.java:163)
```

However, it *should* be covered by the fact that `scala.collection` is already whitelisted, so what I suspect is happening is that somewhere the sandboxing code isn't accounting for "classes" ending in `$`. I can try to dig into this further but am a bit hampered by even my normal (non Nanoware) workspace being a bit wacky. I ran `gradlew idea` and opened the project, but still many things appear missing:

image

Cervator:
Gotcha. Yep, may need to leave this to simmer for a bit until the Gradle overhaul is further along and we've played with the sandbox more :-)

skaldarnar:
Not strictly bound to the scala part of this PR, but still related: with our current approach on whitelisting is it possible at all for modules to add additional dependencies, or is this only an issue on the "language" level (io, system, …).
And, are we facing a similar problem with the Kotlin PoC?

PR #3160, opened 2017-12-03
Add functionality to take server MOTD from a file motd.md.

Naman-sopho:
### Contains
Add feature for server MOTD input from file as mentioned in #2855 

### How to test
- Create a motd.md file in the root directory(temporary placement of the file for development)
- Spin up a server using `./gradlew server`
- Stop the server and edit the `config.cfg` file in the server directory as below
     `"upstreamBandwidth": 1024,`
     `"serverPort": 25777,`
     `"serverMOTD": "[Text would appear as MOTD in absence of the motd.md file]",`
     `"masterServer": "meta.terasology.org"`
- Now start the server again and start a client.
- Join the server (localhost if on the same machine).

### DIFF
--- a/engine/src/main/java/org/terasology/config/NetworkConfig.java
+++ b/engine/src/main/java/org/terasology/config/NetworkConfig.java
@@ -20,9 +20,7 @@ import com.google.common.collect.Lists;
 
 import org.terasology.engine.TerasologyConstants;
 
-import java.io.BufferedReader;
-import java.io.FileReader;
-import java.io.IOException;
+import java.io.*;
 import java.util.Collections;
 import java.util.List;
 
@@ -75,13 +73,15 @@ public class NetworkConfig {
         try {
             StringBuffer mOTD = new StringBuffer();
 
-            FileReader reader = new FileReader("motd.md");
+            InputStreamReader reader = new InputStreamReader(new FileInputStream("motd.md"), "UTF8");
             BufferedReader bufferedReader = new BufferedReader(reader);
 
             String line;
             while ((line = bufferedReader.readLine()) != null) {
                 mOTD.append(line);
             }
+
+            reader.close();
 
             serverMOTD =  mOTD.toString();
 
--- a/engine/src/main/java/org/terasology/config/NetworkConfig.java
+++ b/engine/src/main/java/org/terasology/config/NetworkConfig.java
@@ -26,8 +26,7 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 
-/**
- */
+
 public class NetworkConfig {
 
     private List<ServerInfo> servers = Lists.newArrayList(new ServerInfo("localhost", "localhost", TerasologyConstants.DEFAULT_PORT));
@@ -73,14 +72,14 @@ public class NetworkConfig {
     }
 
     public String getServerMOTD() {
-        try{
+        try {
             StringBuffer mOTD = new StringBuffer();
 
             FileReader reader = new FileReader("motd.md");
             BufferedReader bufferedReader = new BufferedReader(reader);
 
             String line;
-            while((line = bufferedReader.readLine()) != null){
+            while ((line = bufferedReader.readLine()) != null) {
                 mOTD.append(line);
             }
 
@@ -88,8 +87,7 @@ public class NetworkConfig {
 
             return serverMOTD;
 
-        }
-        catch(IOException e) {
+        } catch (IOException e) {
             return serverMOTD;
         }
     }
--- a/engine/src/main/java/org/terasology/config/NetworkConfig.java
+++ b/engine/src/main/java/org/terasology/config/NetworkConfig.java
@@ -20,6 +20,9 @@ import com.google.common.collect.Lists;
 
 import org.terasology.engine.TerasologyConstants;
 
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

Hopefully we can start merging more of these PRs :-)

GooeyHub:

Cervator:
On second thought / review of this - is this adding the check in the right place? The message of the day option is already fully defined, the idea is just to load it either from `config.cfg` or a new file. Shouldn't it be possible to load it from an alternative location but then use the same property to store it so we don't need to regenerate the `NetData` class?

Naman-sopho:
@Cervator I reverted the changes made in the Network layer and tested the feature.
One bug (not sure if it can be considered a bug) which I detected was that the MOTD was displayed from the `motd.md` file even if the `config.cfg` has no `serverMOTD` property defined. Is this a desired behaviour or should this be handled as no MOTD defined?

GooeyHub:

jellysnake:
I would be inclined to treat that as correct behaviour. I would say that the fallback should be something like

`motd.md` file. Then try the `serverMOTD` property. Then display a default MOTD

engine/src/main/java/org/terasology/config/NetworkConfig.java
```diff
@@ -70,7 +72,24 @@ public void setServerPort(int serverPort) {
     }
 
     public String getServerMOTD() {
-        return serverMOTD;
+        try {
+            StringBuffer mOTD = new StringBuffer();
+
+            FileReader reader = new FileReader("motd.md");
```
flo: This file reader gets never closed. Also no enclude gets specified which is bad as it is then platform specific. Should be UTF-8

GooeyHub:

engine/src/main/java/org/terasology/config/NetworkConfig.java
```diff
@@ -75,14 +73,16 @@ public String getServerMOTD() {
...
+            InputStreamReader reader = new InputStreamReader(new FileInputStream("motd.md"), "UTF8");
             BufferedReader bufferedReader = new BufferedReader(reader);
 
             String line;
             while ((line = bufferedReader.readLine()) != null) {
                 mOTD.append(line);
             }
 
+            reader.close();
+
```
flo: If an exception occurs between reader creation and this line then it won't be closed.

There is the autoclosable functionallity you can go for. E.g.
```try(InputStream inputStream = Files.newInputStream) {}```
or old style:
```
InputStream inputStream = ...;
try {

} finally {
   inputStream.close();
}
```
However since you are just reading all lines you can use `Files.readAllLines`.

I also noticed that the motd file has the extension `md`. Do we have highltighting for that already? If not do we have support for simple html? Maybe we should go for html then. Or even better put it in a module and then we are not limited.

flo:
@qwc As you created the issue. What do you think about the idea of having it as module?

engine/src/main/java/org/terasology/config/NetworkConfig.java
Naman-sopho: - I moved the `reader.close()` in a `finally` block.
- If the file is relatively larger in size won't the `Files.readAllLines()` cause problems? We can't be sure about the size of the message to be displayed afaik.
- Had a discussion with @Cervator regarding the` .html` or `.md` issue but currently there is support for neither of them so it did not matter when I implemented this.
- The MOTD should be displayed to all users when they join the game, so if this is moved to a module a user with that module deactivated would have no idea about the MOTD unless there is a feature present to activate this module by default for every user.

qwc:
Well yes, I think it's a great idea to put it into a module, it extends the possibilities for everything. :)

I requested it back in the day, to have something to give the players of my two contribution servers some information about the fact that their buildings and characters may not last forever on those servers due to the update automatisms, I'm of course fine with modularization and multiple possibilities to bring new approaches for this feature into existence. :+1:

vampcat:
@Naman-sopho: The enabled modules are decided by the server, and not the player. That means if I, as a server host, enable a MOTD module, every single player will see it.

Naman-sopho:
@vampcat Ok.
In that case the MOTD module would be better than the current engine implementation to keep the engine lean.

Cervator:
MOTD in a module sounds great to me :-)

Primary reason I figured engine was the desire to have a MOTD loaded from a file, like is often done with other games. Could be the same with a whitelist file for which users can connect. I don't think we can support that from a module at present? In the short term sticking with `config.cfg` is probably sufficient, although that has some limitations.

Maybe we should look at adding support for a safe user-provided data/config directory that modules are allowed to have some minimal temp space in, both read and write (maybe add a throttle to avoid abuse like a module overwriting tiny files to cause disk I/O spam). Maybe hook that up to a new asset type that indirectly allows interacting with a thing on disk? If even needed. Then you could have a module try to look for a MOTD file that a server admin could still provide pretty easily.

So instead of `<Terasology root>/motd.whatever`  you could have `<Terasology root>/sandbox/modules/MOTD/motd.whatever` (available only to the "MOTD" module - modules could perhaps also request shared space somewhere)

Cervator:
Summarizing some chat just now on Slack:

* May close this PR as it is fairly minor and include the functionality in a redo that's module based (should submit an issue with updated details)
* Can simply use a new `MOTD` module that has a UI showing the MOTD with a default message. If the module is enabled, pop the message via `showScreen` in the `postBegin` of a MOTD System class. The default message could simply contain instructions for how to configure the MOTD, including setting it empty somewhere to disable the MOTD
* Later on can draw in an actual message to display via config. This is where it gets interesting
  * Could we still read it from a module specific part of `config.cfg` just to have a first working version? Key bindings can work that way but are somewhat special
  * A FlexibleConfig could probably provide it (module specific FC I'm sure that's a thing. Where would it store to?)
  * A file with the message stored in a new module sandbox file storage location could be read if present (think `sandbox/modules/MOTD/motd.txt` with new engine functionality to allow reading stuff from there). This is an alternative to leaving a file in the data root (shove more stuff into modulespace)
  * A console command used in-game could update the MOTD if the user has the appropriate permissions
* Later on more advanced alternatives could be implemented like one supporting graphical elements in the MOTD. These could be either a new variant in the MOTD module or a whole new module (server admins would just enable one)
* Other functionality like score boards or tutorials could also display on initial join, but aren't they just similar from using the same `postBegin()` to show the message? They're more like gameplay while a MOTD is a system type thing

I might not have caught everything so anybody feel free to chime in with extras or questions :-)

Cervator:
Alright, MOTD in a module now exists! https://github.com/Terasology/ServerMOTD - thanks @Naman-sopho :-)

I created a few follow-up issues in that repo for further improvements, like the file based approach listed here, and am linking back here for reference. With that done I'll close this PR and let work continue in the module

Issue #3113, opened 2017-09-25
Checkstyle Error - Property `maxLineLength` in module `LeftCurly` does not exist


aaron-harris:
When running Checkstyle (through IntelliJ, with plugin version 5.10.1 and therefore Checkstyle version 8.2), I get the following error:

    The Checkstyle rules file could not be parsed.
    cannot initialize module TreeWalker - Property 'maxLineLength' in module LeftCurly does not exist, please check the documentation
    The file has been blacklisted for 60s.

It looks like the referenced property was first [deprecated](https://github.com/checkstyle/checkstyle/commit/289ef67e9e9ee424723b8aa22ae7a6d31f126d23) and then [removed](https://github.com/checkstyle/checkstyle/issues/3671).  In my local copy of the config file, I just blew that property away, and everything works.  I mention it here so that the script generating that file can be updated (by someone with better understanding than I of how that works).

Cervator:
Thanks for the note! Yeah this happens from time to time as versions of things get upgraded, appreciate the detective work :-)

The rules file isn't generated so much as copied in from a central location: https://github.com/MovingBlocks/TeraConfig/tree/master/checkstyle

Keeping it there lets us use the same rules across a bunch of different repositories. We "build" that repo and publish a zip with the rules and such to our Artifactory instance. From there the rules are retrieved like any other dependency and unzipped for use by Gradle and IDEs :-)

We should tweak that file (or both) to avoid this error, after checking on the version configured in Gradle as well (including a few other projects just in case)

theobisproject:
I also stumbled over this error while working on #3112. Since we are using an old checkstyle version in the gradle build this is no problem but it is a much easier to correct the issues while they are highlighted by Intellij.
I think we should remove the property from the checkstyle file and also upgrade checkstyle to the latest version. @Cervator how do modules include the checkstyle? Is it inherited from the main project?

oniatus:
The checkstyle version is defined here: https://github.com/MovingBlocks/Terasology/blob/develop/config/gradle/common.gradle
The used config is from here: https://github.com/MovingBlocks/TeraConfig/tree/master/checkstyle

The root project defines the dependency here: https://github.com/MovingBlocks/Terasology/blob/develop/build.gradle#L80 this should apply to modules because they are build as sub-projects.

Cervator:
Finally fixed via #3162 - huzzah!

Needs a `gradlew extractConfig` to get the new rules file.

PR #3162, opened 2017-12-03
Slight update to the code metrics / scanning stuff.

Cervator:
Brings in https://github.com/MovingBlocks/TeraConfig/releases/tag/v1.2.0 to the engine - two tiny tweaks for Checkstyle, including the fix for #3113

Does enable nagging over empty Javadoc so it'll add some new warnings.

Requires updating the metrics config files via `gradlew extractConfig` or it won't make a difference. On the plus side no IntelliJ restart needed, just hit the button that reloads Checkstyle rules!


GooeyHub:

Cervator:
721 new Checkstyle warnings - woo!

PR #3163, opened 2017-12-04
Multiconnect family rework

pollend:
I tried reworking how multiconnect family works. I removed all the specific elements and made the implementation more general. One concern is that this creates a bit more boilerplate code when defining a new family. The gist below shows how this may look with the rework. 

https://gist.github.com/pollend/f5f9870303e4e3f68d1c5debcc8a8184

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/family/MultiConnectFamily.java
+++ b/engine/src/main/java/org/terasology/world/block/family/MultiConnectFamily.java
@@ -83,8 +83,6 @@ public abstract class MultiConnectFamily extends AbstractBlockFamily implements 
      */
     public abstract byte getConnectionSides();
 
-    public abstract boolean horizontalOnly();
-
     @Override
     public abstract Block getArchetypeBlock();
 
--- a/engine/src/main/java/org/terasology/world/block/family/MultiConnectFamily.java
+++ b/engine/src/main/java/org/terasology/world/block/family/MultiConnectFamily.java
@@ -37,6 +37,14 @@ import org.terasology.world.block.shapes.BlockShape;
 import java.util.Locale;
 import java.util.Set;
 
+/**
+ * Multi-Connect family describes a block family that will connect to other neighboring blocks.
+ *
+ * examples:
+ * - Rail Segments
+ * - Cables
+ * - Fence
+ */
 public abstract class MultiConnectFamily extends AbstractBlockFamily implements UpdatesWithNeighboursFamily {
     private static final Logger logger = LoggerFactory.getLogger(FreeformFamily.class);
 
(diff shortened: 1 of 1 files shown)


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/world/block/family/MultiConnectFamily.java
```diff
@@ -36,160 +34,62 @@
...
-        addConnections(basicBlocks, 5, FIVE_CONNECTIONS);
-        addConnections(basicBlocks, 6, SIX_CONNECTIONS);
-    }
-
+    public abstract byte getConnectionSides();
 
-    public boolean horizontalOnly() {
-        return false;
-    }
+    public abstract boolean horizontalOnly();
```
kaen: Doesn't look like this is used (at least in this or the example code). Can it be removed? If not, is there reason for not having verticalOnly?

Feels like a leftover or a weird thing used by something else in the engine. In that case I'd like to see something like `shouldRotateVertically` or whatever the appropriate predicate is so that implementations of this class can own the logic. That may be unclear, but as an example `animal.canSwim()` instead of `animal.isFish()`

engine/src/main/java/org/terasology/world/block/family/MultiConnectFamily.java
```diff
@@ -36,160 +34,62 @@
...
 import org.terasology.world.block.shapes.BlockShape;
 
-import java.util.Collection;
 import java.util.Locale;
-import java.util.Map;
 import java.util.Set;
-import java.util.stream.Collectors;
 
-@BlockSections({"no_connections", "one_connection", "line_connection", "2d_corner", "3d_corner", "2d_t", "cross", "3d_side", "five_connections", "all"})
 public abstract class MultiConnectFamily extends AbstractBlockFamily implements UpdatesWithNeighboursFamily {
```
kaen: Missing JavaDocs :)

This class is pretty esoteric so a bit of documentation about why it's abstract and the contracts for its abstract methods could go a long way. I think generalizing it is the right way to go, however.

Otherwise I think it needs a generously commented concrete implementation.

Cervator:
Bump - @pollend want to add a little javadoc or an implementation example / unit test so we can merge this? :-)

I know there was some discussion on Slack from Soundwave and Ghost but it didn't make it here so it is hard to act on it / review :-)

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/world/block/family/MultiConnectFamily.java
pollend: There is a lot of legacy stuff I'm unsure about.

GooeyHub:

Cervator:
From the sound of it on Slack this is ready now - or anything else left to consider?

pollend:
@Cervator I believe so. I made the couple of changes that @kaen requested.

GooeyHub:

Cervator:
Merging this - onward to the next round of improvements, if any remain :-)

PR #3164, opened 2017-12-05
Extended the get functionality

digitalripperynr:
NOT FINAL... ONLY FOR TEST
I've added '-remote' detection and two more testing commands for listing and adding remotes to a module.
There is still a lot of cleaning left. Also I need to add MyUser/MyModule feature.


GooeyHub:

Cervator:
Some initial thoughts and feedback

* A bit better clarity for variable names would be good - no real need to abbreviate these days
* The `add-remote` is great, but I think you could skip or make the URL entirely optional: deduce that from the remote name. If I enter "Cervator" plug in `https://github.com/Cervator/[moduleName]` - just allow the user to add a URL at the end if it differs from the deduced URL
* Maybe add a git fetch operation to the end of adding a new remote? I dunno if that would lead to more errors (if a user is adding a remote that doesn't exist yet - but would that happen a lot?) than the added convenience is worth
* Avoid using the GitHub website file uploader :D It is usable in a pinch and for rare cases but usually looks like something is wrong or rushed

digitalripperynr:
Done!
These are the things I added -

- A list-remotes function to list all the remotes of a particular module.

- An add-remote function to add a remote to a module.

- The get command now has a '-remote' detection mechanism to check for a separately specified remote name. The name specified with remote is added to all the modules cloned by the get command.

- If no name is given after '-remote' then the user is asked to enter the name.

- The add-remote function now automatically generates a URL for the remote and gives the option to change the module name for the URL.

- The info column is now updated, containing brief overview of the 3 functions.

- Small adjustments and bug fixes to my code and clean up.


Hope you like my work :D 
-Arpit Kamboj

jellysnake:
Done some testing and this is the result of that:

list-remotes and add-remote both error if the module being added doesn't exist. Would be better to fail gracefully and tell the user to use `groovyw module get` before using these commands

With add-remote specifically I think that user interface could be a little better, I'd rather something like

    > groovyw module add-remote SimpleFarming jellysnake
    Adding Remotes for SimpleFarming module.
    Using remote "http://github.com/jellysnake/SimpleFarming" (Y/n)

If the uses inputs `y` or `enter` then proceed as normal. Else prompt them for the new module name

    Using remote "http://github.com/jellysnake/SimpleFarming" (Y/n)
    > n
    Enter repository name to use:
    > SimpleFarmingFork
    Successfully added remote "http://github.com/jellysnake/SimpleFarmingFork" for SimpleFarming

It would be nice if it could detect if the user has entered an invalid url combo but that's not required for now. Other than those two things it looks to be good

Cervator:
I love the work and this is giving me more ideas still for additional and faster utility :-)

Agreed with Jelly for a check that a local module does exist before allowing `list-remotes` or `add-remote` to do any sort of work. Fewer ugly errors == good plus it avoids a GitHub lookup.

IMHO (and this from somebody who spends waaaay too much time adding remotes) I'd rather the short version remains short with no additional prompting at all and the long version simply takes an additional URL for when you want a variant :-)

So in that case:

`groovyw module add-remote SimpleFarming jellysnake` 

-> adds a remote `jellysnake` to the module `SimpleFarming` with standard URL http://github.com/jellysnake/SimpleFarming and additionally does a git fetch of it immediately - one command, no added keystrokes or anything needed

`groovyw module add-remote SimpleFarming jellysnake http://github.com/jellier/SimpleFarmingFork` 

-> Adds a remote `jellysnake` to the module `SimpleFarming` but uses a non-standard URL including both a different username and module name

I'm having a hard time thinking of examples where I needed *either* a custom remote name or a module name variant. I started using copy paste for user account names a long time ago rather than use made up abbreviations. It is nice and consistent to always have `jellysnake` wherever I've got a link there rather than stumble a moment because in one place I made it `jelly`. So I don't imagine I'd ever use the second variant of the command, making it a pure obstacle to have a prompt for whether to accept the default or not :-)

Adding the fetch (and maybe even a sub-command to just fetch all remotes for a given module) would cut my setup time further plus be one way to catch bad URLs since then you'd get a valid error indicating either you have a bad URL or somebody still needs to create the repo.

I tested the `-remote` as well and it worked, cool stuff. It did make me think of the case in which you're trying to add a module that doesn't yet have a Terasology repo though. Dunno if it is intuitive that `-remote` expects both Terasology + the given user fork to exist.

Right now I think it is fine as-is but maybe we should consider one of two additional options to address:

* The `Account/Module` version and let that be the way to get a module *only* from that location (no matching `Terasology/module`)
* Make both remotes but if the Terasology URL doesn't resolve instead clone from the variant account (I guess the clone would have to happen first - in which case name that after the account instead of `origin` then set that on the Terasology remote)

Maybe it is best for consistency to use the approach in the second bullet regardless so you *always* get two remotes but won't have the script fail if the Terasology one doesn't exist yet.

Cervator:
Oh, also, the diff is also still wrong somehow - the whole file is marked as changed, every single line :/

I tried copy pasting it locally which correctly only marked some of the lines as changed. I don't know why.

pollend:
@Cervator did the line endings change?

Cervator:
> did the line endings change?

Possibly. I think the default option these days goes check out OS specific check in Linux style - but maybe if check in Windows style / retain existing got set instead this could happen. When I copy paste or pull it locally everything looks fine - perhaps since I'm on Windows.

Cervator:
Merged the other PR that included this - although since additional commits were added here this PR ended up looking closed instead of merged.

@digitalripperynr added you to the credits too. In the future keep a watch out for which branch you're adding commits to and look up how to use topic branches - although based on your new PR you might have figured out that already :-)

digitalripperynr:
Thanks @Cervator

PR #3165, opened 2017-12-07
Fix typos and wildcard error in EnvironmentSwitchHandler

PS-Soundwave:
Fixes a compile error in specification-compliant Java compilers, of which javac is not one.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/bootstrap/EnvironmentSwitchHandler.java
+++ b/engine/src/main/java/org/terasology/engine/bootstrap/EnvironmentSwitchHandler.java
@@ -64,6 +64,7 @@ public final class EnvironmentSwitchHandler {
     public EnvironmentSwitchHandler() {
     }
 
+    @SuppressWarnings("unchecked")
     public void handleSwitchToGameEnvironment(Context context) {
         ModuleManager moduleManager = context.get(ModuleManager.class);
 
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/engine/bootstrap/EnvironmentSwitchHandler.java
```diff
@@ -123,6 +119,15 @@ public void handleSwitchToGameEnvironment(Context context) {
 
     }
 
+    @SuppressWarnings("unchecked")
+    private <T, U extends CopyStrategy<?>> void registerCopyStrategy(CopyStrategyLibrary copyStrategyLibrary, Class<T> type, Class<U> strategy) {
```
pollend: Any reason why this needs to be moved to its own function?

PS-Soundwave: The call to CopyStrategyLibrary.register had two different wildcards captured by the same type argument. This is not supposed to compile successfully (and IDEA reports this), but does currently for whatever reason. The fix is to capture each wildcard in a separate argument, which I've done with this function.

oniatus: Good catch but change the second type to `U extends CopyStrategy<T>`. This will save you the cast in line 125 and fits the signature of CopyStrategyLibrary.register: `public <T> void register(Class<T> type, CopyStrategy<T> strategy)`. Then this PR is good to go =)

PS-Soundwave: It just makes the cast implied in the call, but it cleans up the signature of `registerCopyStrategy` so I went ahead and did it.

GooeyHub:

oniatus:
Approved and merged, thank you 👍

PR #3166, opened 2017-12-07
Fixed Various Checkstyle Issues

FieryPhoenix909:
### Contains

Removes empty Javadoc blocks in some files and fixes a TypeNameCheck.

### DIFF
--- a/engine-tests/src/test/java/org/terasology/registry/InjectionHelperTest.java
+++ b/engine-tests/src/test/java/org/terasology/registry/InjectionHelperTest.java
@@ -196,7 +196,7 @@ public class InjectionHelperTest {
         }
     }
 
-    public static class ConstructorA_AB {
+    public static class ConstructorAAB {
         private ServiceA serviceA;
 
         private ServiceB serviceB;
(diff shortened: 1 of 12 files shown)


GooeyHub:

engine-tests/src/test/java/org/terasology/registry/InjectionHelperTest.java
```diff
@@ -196,7 +196,7 @@ public ServiceB getServiceB() {
         }
     }
 
-    public static class ConstructorA_AB {
+    public static class ConstructorAAB {
```
Cervator: Erm, this may fix a Checkstyle warning but it breaks the code in doing so :-)

engine/src/dev/java/org/terasology/benchmark/entitySystem/EntityCreateBenchmark.java
```diff
@@ -27,9 +27,6 @@
 
 import java.util.List;
 
-/**
```
Cervator: While empty javadoc removal is fine for the GCI task as noted in the task each instance only counts for 1/10th to the requirements for the task - since it is trivial to delete a few lines.

PR #3167, opened 2017-12-07
Fixes checkstyle issues in WorldRendererLmpl

NicholasBatesNZ:
### Contains

- 19 checkstyle error fixes
- several warning fixes

in class engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java

- [ ] Need to consider rGraph variable name
- [ ] Need to consider displayResolutionDependentFBOs1 variable name

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -109,11 +109,11 @@ import static org.terasology.rendering.opengl.ScalingFactors.QUARTER_SCALE;
 
 /**
  * Renders the 3D world, including background, overlays and first person/in hand objects. 2D UI elements are dealt with elsewhere.
- *
+ * <p>
  * This implementation includes support for OpenVR, through which HTC Vive and Oculus Rift is supported.
- *
+ * <p>
  * This implementation works closely with a number of support objects, in particular:
- *
+ * <p>
  * TODO: update this section to include new, relevant objects
  * - a RenderableWorld instance, providing acceleration structures caching blocks requiring different rendering treatments<br/>
  */
@@ -177,18 +177,18 @@ public final class WorldRendererImpl implements WorldRenderer, ComponentSystem {
 
     /**
      * Instantiates a WorldRenderer implementation.
-     *
+     * <p>
      * This particular implementation works as deferred shader. The scene is rendered multiple times per frame
      * in a number of separate passes (each stored in GPU buffers) and the passes are combined throughout the
      * rendering pipeline to calculate per-pixel lighting and other effects.
-     *
+     * <p>
      * Transparencies are handled through alpha rejection (i.e. ground plants) and alpha-based blending.
      * An exception to this is water, which is handled separately to allow for reflections and refractions, if enabled.
-     *
+     * <p>
      * By the time it is fully instantiated this implementation is already connected to all the support objects
      * it requires and is ready to render via the render(RenderingStage) method.
      *
-     * @param context a context object, to obtain instances of classes such as the rendering config.
+     * @param context    a context object, to obtain instances of classes such as the rendering config.
      * @param bufferPool a GLBufferPool, to be passed to the RenderableWorld instance used by this implementation.
      */
     public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
@@ -212,7 +212,7 @@ public final class WorldRendererImpl implements WorldRenderer, ComponentSystem {
                 * in match.
                  */
                 vrProvider.getState().setGroundPlaneYOffset(
-                        GROUND_PLANE_HEIGHT_DISPARITY  - context.get(Config.class).getPlayer().getEyeHeight());
+                        GROUND_PLANE_HEIGHT_DISPARITY - context.get(Config.class).getPlayer().getEyeHeight());
                 currentRenderingStage = RenderingStage.LEFT_EYE;
             } else {
                 playerCamera = new PerspectiveCamera(worldProvider, renderingConfig, context.get(DisplayDevice.class));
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -278,204 +278,210 @@ private void initRenderGraph() {
         renderPipelineTaskList = renderTaskListGenerator.generateFrom(orderedNodes);
     }
 
-    private void addShadowMapNodes(RenderGraph renderGraph) {
+    private void addShadowMapNodes(RenderGraph rGraph) {
```
vampcat: I *definitely* don't like this change.

Rather than rename all variables to rGraph, you can simply remove them from the parameters.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -278,204 +278,210 @@ private void initRenderGraph() {
...
+        BufferClearingNode staleGBufferClearingNode = new BufferClearingNode(gBuffer2Config, displayResolutionDependentFBOs,
+                GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+        rGraph.addNode(staleGBufferClearingNode, "gBuffer2ClearingNode");
 
         Node backdropNode = new BackdropNode(context);
-        renderGraph.addNode(backdropNode, "backdropNode");
+        rGraph.addNode(backdropNode, "backdropNode");
 
-        DisplayResolutionDependentFBOs displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
+        DisplayResolutionDependentFBOs displayResolutionDependentFBOs1 = context.get(DisplayResolutionDependentFBOs.class);
```
vampcat: Same as above.

Rather than rename all variables to `displayResolutionDependentFBOs1`, you can simply remove this line and use the existing class variable `displayResolutionDependentFBOs`.

You need to identify the *cause* of the problem when trying to fix warning, rather than just making random changes that make the warnings go away.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -278,204 +278,210 @@ private void initRenderGraph() {
...
 
         FBOConfig halfScaleBloomConfig = new FBOConfig(BloomBlurNode.HALF_SCALE_FBO_URI, HALF_SCALE, FBO.Type.DEFAULT);
-        FBO halfScaleBloomFbo = displayResolutionDependentFBOs.request(halfScaleBloomConfig);
+        FBO halfScaleBloomFbo = displayResolutionDependentFBOs1.request(halfScaleBloomConfig);
 
         label = "halfScaleBlurredBloom";
-        BloomBlurNode halfScaleBlurredBloom = new BloomBlurNode(context, displayResolutionDependentFBOs.get(HighPassNode.HIGH_PASS_FBO_URI), halfScaleBloomFbo, label);
-        renderGraph.addNode(halfScaleBlurredBloom, label + "Node");
+        BloomBlurNode halfScaleBlurredBloom = new BloomBlurNode(context, displayResolutionDependentFBOs1.get(HighPassNode.HIGH_PASS_FBO_URI), halfScaleBloomFbo, label);
```
vampcat: Same as above.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: @NicholasBatesNZ: what's the feedback checkstyle gives you regarding this variable?

A general rule of thumb in programming nowadays is to -not- abbreviate. It made sense once when we had to type everything, but nowadays with autocompletion and easy replacement available through the IDE softwares it no longer apply and clarity trumps brevity.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -701,25 +707,32 @@ public void recompileShaders() {
     }
 
     @Override
-    public void initialise() { }
+    public void initialise() {
```
emanuele3d: @NicholasBatesNZ: which checkstyle are you using here, the one in IntelliJ or the one from the builds?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
vampcat: Unless I'm mistaken, Checkstyle was complaining about variable hiding.

NicholasBatesNZ: Yes, it was variable hiding. I was not entirely sure what to do about it so I'll remove the parameter entirely as you suggest

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
NicholasBatesNZ: Yes, I am running from source. I am using the Checkstyle-IDEA plugin configured to the terasology checkstlye.xml

NicholasBatesNZ:
Hidden variable changes have been made

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -287,9 +287,10 @@ private void addShadowMapNodes(RenderGraph renderGraph) {
         renderGraph.addNode(shadowMapNode, "shadowMapNode");
     }
 
-    private void addReflectionNodes(RenderGraph renderGraph) {
+    private void addReflectionNodes() {
         FBOConfig reflectedBufferConfig = new FBOConfig(BackdropReflectionNode.REFLECTED_FBO_URI, HALF_SCALE, FBO.Type.DEFAULT).useDepthBuffer();
-        BufferClearingNode reflectedBufferClearingNode = new BufferClearingNode(reflectedBufferConfig, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+        BufferClearingNode reflectedBufferClearingNode = new BufferClearingNode(reflectedBufferConfig, displayResolutionDependentFBOs,
```
emanuele3d: Is this a line length problem? If so, I'd start the second line with the "new" keyword, so that each side of the assignment is on one line. Other lines below have similar issues.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -299,41 +300,44 @@ private void addReflectionNodes(RenderGraph renderGraph) {
...
-        BufferClearingNode staleGBufferClearingNode = new BufferClearingNode(gBuffer2Config, displayResolutionDependentFBOs, GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+        BufferClearingNode staleGBufferClearingNode = new BufferClearingNode(gBuffer2Config, displayResolutionDependentFBOs,
+                GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
         renderGraph.addNode(staleGBufferClearingNode, "gBuffer2ClearingNode");
 
         Node backdropNode = new BackdropNode(context);
         renderGraph.addNode(backdropNode, "backdropNode");
 
-        DisplayResolutionDependentFBOs displayResolutionDependentFBOs = context.get(DisplayResolutionDependentFBOs.class);
+        DisplayResolutionDependentFBOs displayResolutionDependentFBOs1 = context.get(DisplayResolutionDependentFBOs.class);
```
emanuele3d: Let's not forget about this one please. You should be able to use the instance-level variable.

NicholasBatesNZ: Ah yes, I forget it happened twice :)

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
NicholasBatesNZ: Yes, it goes past the 175 character limit and I have just adjusted the line breaks. Also I have play tested it to confirm that it is indeed functional.

NicholasBatesNZ:
Changes have been made and playtest confirms proper functionality

Cervator:
Thanks for the fixes! Also added you to the credits :-)

PR #3169, opened 2017-12-08
Refactor groovy module script - builds on #3164

Cervator:
This enhances PR #3164 from @digitalripperynr some - contains a variety of code quality and clarity improvements I think would be a bit much to expect from a GCI task, since you can kinda keep moving and changing things around a while before settling on something that seems satisfactory ;-)

I used fairly granular commits to point out various good practices and to keep the changes easier to see.

One extra enhancement is favoring the user-supplied remote over the default Git clone target under the Terasology org. The module repo will now be cloned from the user-supplied remote *but* that will be named accordingly while `origin` gets added as well, pointing to the Terasology org version of the module.

Sample output (tweaked a little for clarity):

```
C:\Dev\Terasology\Git\Terasology>groovyw module get Compass -remote Nanoware
Preparing to get one or more modules
Before processing custom remote: [get, Compass, -remote, Nanoware]
Going to look for and process remotes out of args if present - current: [get, Compass, -remote, Nanoware]
The remote arg was in position 2 and the value is -remote
Dropped two args from arguments which is now: [get, Compass]
Done with -remote if it was there, removing the first argument to leave only modules
After processing custom remote: [Compass]
Now inside retrieve, user (recursively? false) wants: [Compass]
Starting loop for module Compass, are we recursing? false
Modules retrieved so far: []
Request to retrieve module Compass would store it at modules\Compass - exists? false
Retrieving module Compass - if it doesn't appear to exist (typo for instance) you'll get an auth prompt (in case it is private)
Doing a retrieve from a custom remote: Nanoware - will name it as such plus add the Terasology remote as 'origin'
Successfully added remote origin for Compass

C:\Dev\Terasology\Git\Terasology>groovyw module list-remotes Compass
Listing Remotes for module Compass
1 Nanoware (https://github.com/Nanoware/Compass.git) <-- Clone happened from here
2 origin (https://github.com/Terasology/Compass.git)
```

Thanks @digitalripperynr for getting this kicked off and please review this in turn and let me know if you can find any problems with it and if you like the changes :-)

With this I'm somewhat walking away from the possibility of support for a command in the style of `Account/Module1 Account/Module2 Module3` - right now the cloning logic works off a class level variable so it is all or nothing. First it is set to either "Terasology" or the value loaded from `gradle.properties` (existing support, shared with various Gradle commands). Then if `-remote` is included that'll overwrite that variable.

I think adding a single automatic fetch to run after `add-remote` would be a nice touch, may look into how to do that real quick but not tonight :-)

Merging this should auto-close #3164 as merged, so nobody close that separately - just wait for this to get merged

Edit: Yay this also fixed the line endings so the overall diff is clean now

I'll also getting curious if the update commands would need to run a fetch first or if that's automatic ...

### DIFF
--- a/module.groovy
+++ b/module.groovy
@@ -1,7 +1,7 @@
 // We use GrGit for interacting with Git. This gets a hold of it as a dependency like Gradle would
 // TODO: Consider if we should do something to fix/suppress the SLF4J warning that gets logged on first usage?
 @GrabResolver(name = 'jcenter', root = 'http://jcenter.bintray.com/')
-@Grab(group='org.ajoberstar', module='grgit', version='1.9.3')
+@Grab(group = 'org.ajoberstar', module = 'grgit', version = '1.9.3')
 import org.ajoberstar.grgit.Grgit
 import org.ajoberstar.grgit.Remote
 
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
I ended up checking out this branch locally just to make testing some modules easier - loving the new functionality :-)

Trying it out with a fetch attached in the final `addRemote` that does the actual work. Seems as simple as:

`remoteGit.fetch remote: remoteName`

Still want to see how it feels when you end up creating a custom remote where the Terasology one doesn't exist yet. Should report back gracefully somehow rather than flip out and barf all over the log :-)

@digitalripperynr ping! Also pinging @mjuvekar7 for an opinion

GooeyHub:

Cervator:
Am happy with it now. I added a fetch step that runs after adding a remote. That in turn would cause trouble if the URL isn't right, either from a typo or not being created yet since GitHub considers the possible existence of a private repo by asking the user for credentials. That's always just confusing in our case as the reality is the user just made a typo or the target repo hasn't been created yet.

To avoid that entirely I also added URL validation, using a simple HTTP HEAD request (yay how easy Groovy makes this!) to test the URL without trying to retrieve anything. That doesn't trigger the auth step if the URL is invalid, so there is no longer any concern or confusion there.

When using `-remote` the actual remote is what gets cloned, but named uniquely like mentioned above. Then the remote add happens to subsequently bind `origin` to the Terasology org version of the module.   A `git fetch` then happens, but only if the URL is valid. Or that's how it was supposed to work, but whoops, I forgot that step, one more commit coming :p 

(in other words: always add the remote, but only fetch if it is valid. Also only clone if valid)

GooeyHub:

PR #3172, opened 2017-12-09
Fix explode sound

### LINKED ISSUE #3171: Item explodeTool plays sound even when there is no target block
The item explodeTool in the core module plays an explode sound every time it is used, doesn't really check if there is a block in range when the tool is used.
This is caused by the PlaySoundAction triggered every time without any condition, in the explode tool prefab.

How to reproduce, with the explodeTool in hand, right click while the crosshair is pointing at the sky or at a block that is far away. An explode sound will sound every time.

The expected behavior is to only play the sound when a block is in range.

pamtdoh:
### Contains

Fixes #3171 , also fixes the railgun which has the same problem.

### How to test

The explodeTool and the railgun should not play an explode sound when there's no block destroyed.

### DIFF
--- a/modules/Core/src/main/java/org/terasology/logic/actions/TunnelAction.java
+++ b/modules/Core/src/main/java/org/terasology/logic/actions/TunnelAction.java
@@ -15,6 +15,9 @@
  */
 package org.terasology.logic.actions;
 
+import com.google.common.collect.Lists;
+import org.terasology.audio.StaticSound;
+import org.terasology.audio.events.PlaySoundEvent;
 import org.terasology.entitySystem.entity.EntityBuilder;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
@@ -29,11 +32,15 @@ import org.terasology.math.geom.Vector3f;
 import org.terasology.math.geom.Vector3i;
 import org.terasology.physics.Physics;
 import org.terasology.registry.In;
+import org.terasology.utilities.Assets;
 import org.terasology.utilities.random.FastRandom;
 import org.terasology.utilities.random.Random;
 import org.terasology.world.BlockEntityRegistry;
 import org.terasology.world.WorldProvider;
 import org.terasology.world.block.Block;
+
+import java.util.List;
+import java.util.Optional;
 
 @RegisterSystem(RegisterMode.AUTHORITY)
 public class TunnelAction extends BaseComponentSystem {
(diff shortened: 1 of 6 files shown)


GooeyHub:

Cervator:
Since you've put some thought into this and the initial issue detection I"m going to approve the GCI task for it, but on thinking more closely I'm not sure this is quite the right solution.

Going back I think the issue has been that the explode tool is meant to target a block and if there is one then trigger an explosion. So it makes sense if there's no actively selected block then don't explode. In other words the explode tool could be said to transform the target block into TNT and then trigger it.

The railgun on the other hand for sure is meant to be a directed blast of destructive energy - it doesn't matter if it actually encounters blocks as it isn't turning them into TNT or destructive energy, the destruction comes from the weapon itself and so should the SFX.

Does that make sense? I should've noticed and started a discussion on this when the issue was raised. 

If you are up for tweaking this PR to fit that approach better or have a counter-argument to present feel free :-)

Also:

@GooeyHub: add to whitelist

GooeyHub:

pamtdoh:
How about adding a new "charging" sound when the railgun is used, and play an explode sound when it hits something. That way the player can know if he or she missed.

modules/Core/src/main/java/org/terasology/logic/actions/TunnelAction.java
```diff
@@ -50,15 +57,25 @@
     private EntityManager entityManager;
 
     private Random random = new FastRandom();
+    private List<Optional<StaticSound>> explosionSounds = Lists.newArrayList();
```
flo: A list of optionals is strange. If an asset isn't there it should not be in the list or there should be actually an exception when the asset could not be found.

modules/Core/src/main/java/org/terasology/logic/actions/TunnelAction.java
```diff
@@ -50,15 +57,25 @@
...
+        explosionSounds.add(Assets.getSound("core:explode4"));
+        explosionSounds.add(Assets.getSound("core:explode5"));
     }
 
     @Override
     public void shutdown() {
     }
 
+    private StaticSound getRandomExplosionSound() {
+        return explosionSounds.get(random.nextInt(0, explosionSounds.size() - 1)).get();
```
flo: This would fail if the optional is empty.

modules/Core/src/main/java/org/terasology/logic/actions/TunnelAction.java
```diff
@@ -50,15 +57,25 @@
...
     private Random random = new FastRandom();
+    private List<Optional<StaticSound>> explosionSounds = Lists.newArrayList();
 
     @Override
     public void initialise() {
+        explosionSounds.add(Assets.getSound("core:explode1"));
+        explosionSounds.add(Assets.getSound("core:explode2"));
+        explosionSounds.add(Assets.getSound("core:explode3"));
+        explosionSounds.add(Assets.getSound("core:explode4"));
+        explosionSounds.add(Assets.getSound("core:explode5"));
```
flo: my personal prefered way of getting assets is via @In AssetManager assetManager  as it is not a static access and could thus be mocked by for example a test.

jellysnake:
@pamtdoh
Are you currently still working on these changes? I'd like to be able to merge this as it's a relatively small fix

Cervator:
Closing this from age - additionally I think somebody recently proposed a different way to fix this, although there's no PR yet.

Hope you're doing well @pamtdoh 👋

PR #3177, opened 2017-12-10
Make spawn logic accept optional 2D fields

PS-Soundwave:
### Contains

Work-in-progress changes to the spawn logic to allow for world types where the concept of "surface height" or "sea level" doesn't make sense everywhere or at all. It will reject spawn locations with no surface height and accept locations regardless of height when there is no sea level (not yet implemented). See PS-Soundwave/ProjSndwvWorldGeneratorTest for a demonstration module.

### How to test

Ensure that properly updated world generators don't break, as well as that functionality works as intended for new generators. Perlin and my generator have been my test cases thus far.

- [ ] Optional sea level

### DIFF
--- a/engine/src/main/java/org/terasology/logic/spawner/AbstractSpawner.java
+++ b/engine/src/main/java/org/terasology/logic/spawner/AbstractSpawner.java
@@ -26,7 +26,11 @@ import org.terasology.math.geom.Vector3i;
 import org.terasology.world.generation.Region;
 import org.terasology.world.generation.World;
 import org.terasology.world.generation.facets.SeaLevelFacet;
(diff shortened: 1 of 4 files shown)


GooeyHub:

mkienenb:
Are you still working on this PR?
It seems reasonable enough to me, although I'm not an expert on world-gen.

I would recommend that you include javadocs for at least the SpawnHeightFacet and OptionalFacet2D classes, possibly re-iterating why these classes are useful like you did in your initial PR comments.

And these seem like decent candidates for real unit tests.

PS-Soundwave:
Yep, it's still in progress. I'm currently on 4 projects and a bit thin on time, but I'll prioritize this so that it doesn't freeze over.

PR #3187, opened 2017-12-17
Add error-prone to engine

vorburger:
see http://errorprone.info

Signed-off-by: Michael Vorburger <mike@vorburger.ch>

### DIFF
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -3,6 +3,7 @@
 // Get the gradle versioneye plugin from the plugins.gradle.org repo
 plugins {
     id "org.standardout.versioneye" version "1.1.0"
+    id "net.ltgt.errorprone" version "0.0.13"
 }
 
 // Grab all the common stuff like plugins to use, artifact repositories, code analysis config, etc


GooeyHub:

Cervator:
Uh - thanks? But it would be great with a bit more reasoning here than a link :-)

There are a bunch of those sites and services we could sign up for and we definitely should, but it needs a bit more digging. 

For instance with that enabled, how does the engine build differ? Do any errors/warnings trigger? Do we need any additional custom config to ignore any false positives? Would the tool likely be eclipsed by a different but similar service we might consider instead? Did you pick this one for any particular reason, like a specific warning you found useful, or have you used it for a different project?

jellysnake:
@vorburger pinging to see how this is going?

vorburger:
@jellysnake @Cervator hey there - thanks for replying, sorry I was AFK.

So http://errorprone.info is one of those Code Quality type tools, like Checkstyle, SpotBugs (formerly known as FindBugs), PMD et al.

I'm reasonably familiar with this space, and have done digging and find error-prone to be useful. Note that AFAIK it is by and extensively used inside Google for their HUGE Java code base, including all of Google's Android apps.

FYI this one is NOT a site or service you need to sign up for or create any server - contrary to something like say Sonar, or others like that. Personally I do not like those "parallel" post-merge "metrics Dashboard" thingies - nobody ever looks at that. The only think that really works in my experience over the years is simple stuff you integrate in your build, which makes even local builds fails for devs. And this error-prone thing is just something that runs during your build and emits warnings or errors (like Checkstyle and SpotBugs).

I have (started to) use this in a large code base I work on "by day", and have configuration tweaks there which you could look at later if you want to add the base tool first, [see the opendaylight/infrautils/common/parent/pom.xml](https://github.com/opendaylight/infrautils/blob/c0b33dcaaae708b17a2feef1ce868c7b6c297611/common/parent/pom.xml).

I've absolutely no problem to Close and Abandon this - it was just a suggestion.

jellysnake:
I'm going to close this and move it to an issue as it's probably better suited for there.

vorburger:
> I'm going to close this and move it to an issue as it's probably better suited for there.

==> #3362

PR #3188, opened 2017-12-20
WIP: Introduce engine libs

oniatus:
This PR introduces the concept for engine-libs as discussed in http://forum.terasology.org/threads/architecture-discussion-for-v3-engine-split-engine-libs.1897

I started with the entity-system and excluded only 3 interfaces for a demonstration.
Actual PR includes:
- easy configurable subproject `engine-libs`. A new lib requires a build.gradle and is ready to go.
- no API breaking, everything should work the way it is
- high quality bar for all engine-libs:
  - Build fails on low test coverage
  - Build fails on low semantic test coverage (uses http://pitest.org/)
  - Build fails on checkstyle violations (no warnings)
  - Lower visibility of classes is enforced via a custom checkstyle check (currently living at https://github.com/oniatus/custom-checks)
  - No impact on existing code! Development on engine can continue as-is

Outstanding changes:

- [x] Project dependency does not work with the engine dev sourceset. An easy fix could be to move the dev stuff to a benchmark project or the test sourceset. Or maybe some gradle magic from @Cervator  😉?
_Edit: This now works by including the buildscript in each library and adding the project as compile dependency. Subproject configuration seems not to play well with our subprojects script. Would be okay for now until we have too many libs._
- [ ] The checkstyle extension is included as local jar. A cleaner solution would be to move the project to the org and publish the artifact to artifactory. Could also be used for further extension if we want more custom checks for terasology (no getters/setters on components would be just one of my further ideas).

Open for discussion 😄


GooeyHub:

GooeyHub:

emanuele3d:
I have reviewed the whole forum thread to try and understand this issue as best as I can. 

I will defer to a discussion between you, @Cervator, @msteiger, @skaldarnar and ideally also @flo and @immortius,  as I feel from a technical standpoint this is not something I am knowledgeable enough, especially how the build process works and the new and more advanced features here on github.

Question. What do these lines mean exactly:

- Build fails on low test coverage
- Build fails on low semantic test coverage 

as in: say we have an existing package with little or no testing, how much would it have to change for the build to succeed? How would we know what needs to be tested and what doesn't?

My concern is that while in theory stricter code quality requirements would produce higher quality, more maintainable code, in practice the higher barrier of entry my deter people from contributing, resulting in no code at all. 

If we had a community with hundreds or even dozens of regular contributors, we might be able to risk raising the formal standards and have such automated approach. Given the small size of our core crew however, I wonder if it would be too great of a risk, as deterring even one or two core contributors would greatly impact the rate of progress.  

Besides, if moving one of the existing packages into engine-libs would require hundreds or thousands of tests to be written, what are the chances of this actually happening?

In this context I wonder if, at least at this stage, a more pragmatic approach would be to improve our developement-oriented documentation. This could take the form of a wiki page detailing broader development and code-quality goals that people with merge privileges should uphold when approving and merging a PR. 

I look forward to your reply @oniatus.

oniatus:
Thanks for the fast reply 😃 

Questions first:
> What do these lines mean exactly:
> - Build fails on low test coverage
> - Build fails on low semantic test coverage

I added the gradle plugin for http://pitest.org/ to the build, this allows to set a threshold for test coverage and mutation coverage (see the tool page as a better reference). In summary it allows to measure if a test fails given a line of code is modified. This is a far stronger metric than simple line coverage because line coverage does not take assert statements into account - you can have 100% coverage without a single assert statement. The build will not pass if the coverage is too low.
We can adjust these values but I like to have them as high as possible at the begin, at best both close to 99-100.

> say we have an existing package with little or no testing, how much would it have to change for the build to succeed?

All code which moves to engine-libs has to be put under test with the described constraints and must fit checkstyle. Good thing is this can be done class by class or even piece by piece. No need for a big huge monster PR which does all at once.

> How would we know what needs to be tested and what doesn't?

Counter question: Which code would you test under no circumstances? 😉 
I would like to have tests for all core features of the engine to gain more confidence when checking PRs.
Practically there will be a lot of messy classes which can't be tested easily (e.g. the Input System: https://github.com/MovingBlocks/Terasology/pull/2948/files#diff-5c6813d9bc3fc50155edc545aa5ceb8c
and https://github.com/MovingBlocks/Terasology/pull/2948/files#diff-3c6e32858f7c4c95be14a46d737c3eb0). But we can still extract testable pieces and move them to the libs in a better environment where further changes have less chances to break something.

> Besides, if moving one of the existing packages into engine-libs would require hundreds or thousands of tests to be written, what are the chances of this actually happening?

See the two linked diffs. It takes a huge amount of effort to clean up such classes and even start to write tests for them - and this is nothing which fits a gci task ;) I already did some changes like this on the engine and I already started to move pieces from the entity system in the libs folder in a second branch based on this one (not ready for a PR yet). Good thing is we don't have to go fast here and realistically i don't think we will ever finish such a big cleanup but that is no reason not to start it and make some pieces better =) One small step at a time.

>My concern is that while in theory stricter code quality requirements would produce higher quality, more maintainable code, in practice the higher barrier of entry my deter people from contributing, resulting in no code at all.

It depends. One advantage of automated validation is that the need for review is lower. Instead we may have to explain people how to write tests and where to look at but then they are able to go their own way. On the other hand there is a lot of code in the engine which will stay there as-is. The first blocks of code to move would be the lowest level ones, aka module and entity-system. Then maybe config, engine-lifecycle or injection/util stuff. Especially on these modules should the barrier for new code very high because regressions or bugs are likely to affect a lot of functionality on top. At the moment we would ensure this via deep reviews and lots of manual testing but this puts more effort on the low number of reviewers. 

This is also a side-effect I would like to achieve: It takes far less review time to check if new code is covered by tests, check that all old tests are untouched or review their modifications, then check the new tests and if they are good and green -> ship it. Extra time can the be spent on refactoring and design cleanup and less time should be needed for tons of manual experimenting and browsing code to check for side-effects.
For the same reason my confidence in quality goals, definition of done or review protocols is not that high :) most of the time everything which is not automated or does not fit on a credit-card size of paper gets lost sometime in the future.

emanuele3d:
> One advantage of automated validation is that the need for review is lower. Instead we may have to explain people how to write tests and where to look at but then they are able to go their own way. 

I think this part is quite crucial: first I'd like to make sure those explanations are provided, then I'd introduce stricter, automated requirements. In Real Life governments often announce new rules and detailed reasoning behind them well in advance of actually enforcing them, so people have time to take them onboard. In this particular circumstance I can see how enforcement only happens when things gets moved into engine-libs, but I don't think you have quite "sold" the concept to the crowds just yet. For example I do not see this connection between automated validation and lower review burden: validation and reviews just catch different problems and are complementary. Compared to the current state of affair however, any move to engine-libs would require tons of new tests (?) and I wonder how many have the will to to do that.

Still, maybe I'm overestimating the actual work that would have to be put in. Rather than providing the refactoring of the InputSystem classes as example, can you provide some examples of the work required to move much smaller, simpler classes into engine-libs? Something that acts as a tutorial for those who are unfamiliar with how the code should eventually looks like to conform to the stricter requirements?

Another thing I remain unclear about is the context in which this testing effort fits. If I look at the wikipedia page on [Software Testing](https://en.wikipedia.org/wiki/Software_testing) there are lots of different types of testing listed. Is this effort all and only about Unit Testing? Integration testing? Interface testing?

oniatus:
Closing this for now after some external discussions.
Summary for other people: Changing parts of the codebase would work but overall takes probably too much effort which we can spend to make more smaller changes.
Likely the better approach is to add small improvements in the active areas und maybe refactor things out once needed or once the change has less impact. At the moment it takes way too much refactoring effort to extract useful entities out of the engine project. We can still use snippets from here if we ever need subprojects or extra gradle magic :) Also thanks @emanuele3d for the feedback and the follow-up discussions at Zurich =)

PR #3195, opened 2017-12-23
Allow non-block entities to explode

anuar2k:
### Contains

A modification of the ExplosionAuthoritySystem.
onDelayedExplosion event handler required the exploding entities to be a block. Now, with this modification, every entity which does have LocationComponent should be able to explode.

I have also made the DELAYED_EXPLOSION_ACTION_ID public to allow other systems use it, instead of statically defining it again in other places (Minesweeper module is the perfect example).

### DIFF
--- a/modules/Core/src/main/java/org/terasology/logic/actions/ExplosionAuthoritySystem.java
+++ b/modules/Core/src/main/java/org/terasology/logic/actions/ExplosionAuthoritySystem.java
@@ -16,6 +16,8 @@
 package org.terasology.logic.actions;
 
 import com.google.common.collect.Lists;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


GooeyHub:

modules/Core/src/main/java/org/terasology/logic/actions/ExplosionAuthoritySystem.java
```diff
@@ -16,6 +16,8 @@
 package org.terasology.logic.actions;
 
 import com.google.common.collect.Lists;
+import org.slf4j.Logger;
```
skaldarnar: These imports are unused and can be removed.

GooeyHub:

skaldarnar:
We discussed some further changes to the explosion authority system on slack. Will keep those for later after some more discussion on the interfaces and components that should be available.

PR #3197, opened 2017-12-23
Add blockDamageModifierComponent null check

### LINKED ISSUE #3196: Dereferencing possible null value causes a NPE in BlockEntitySystem.commonDefaultDropsHandling
NPE occurs when accessing `blockDamageModifierComponent.impulsePower` in `commonDefaultDropsHandling` when `blockDamageModifierComponent` is missing (`null`).

https://github.com/MovingBlocks/Terasology/blob/ee0da21d523fd67ca951a1c600d80c3b85a17564/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java#L134

## Reproduction
See Terasology/Rails#32.

eviltak:
Fixes #3196 and closes Terasology/Rails#32.

If `blockDamageModifierComponent` is null, uses `0` as the `impulsePower`.

## Testing
Problem detailed in Terasology/Rails#32 does not occur anymore, needs confirmation.
@anuar2k

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
@@ -131,7 +131,14 @@ public class BlockEntitySystem extends BaseComponentSystem {
             entity.send(new OnBlockToItem(item));
 
             if (shouldDropToWorld(event, block, blockDamageModifierComponent, item)) {
-                processDropping(item, location, blockDamageModifierComponent.impulsePower);
+                float impulsePower;
+                if (blockDamageModifierComponent != null) {
+                    impulsePower = blockDamageModifierComponent.impulsePower;
+                } else {
+                    impulsePower = 0;
+                }
+                
+                processDropping(item, location, impulsePower);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
```diff
@@ -131,7 +131,14 @@ public void commonDefaultDropsHandling(CreateBlockDropsEvent event, EntityRef en
             entity.send(new OnBlockToItem(item));
 
             if (shouldDropToWorld(event, block, blockDamageModifierComponent, item)) {
-                processDropping(item, location, blockDamageModifierComponent.impulsePower);
+                float impulsePower;
+                if (blockDamageModifierComponent != null) {
+                    impulsePower = blockDamageModifierComponent.impulsePower;
+                } else {
+                    impulsePower = 0;
+                }
```
jellysnake: I feel like this would be more readable if the else was removed and `impulsePower` was initialised to `0` at the same time it is declared.

eviltak: Great catch. Must have missed it at the time :)

GooeyHub:

jellysnake:
Thanks, merging now :)

PR #3199, opened 2017-12-26
Physics refactor

eviltak:
## Goal
This refactor aims to reduce the dependency of engine code on the physics engine library used. Engine items currently use items in the `com.bulletphysics` ([MovingBlocks/TeraBullet](https://github.com/MovingBlocks/TeraBullet)) package, which can cause problems while refactoring (moving to a different engine) and necessitates adding an otherwise unrelated package to the module whitelist. The main cuplrits were references to `com.bulletphysics.collision.shapes.CollisionShape` and its subclasses, and `com.bulletphysics.linearmath.Transform`. All these classes have been wrapped around engine classes and interfaces, which can be easily used in both engine as well as module code.

These changes also allow for easily changing the underlying physics engine used without having to change _any_ engine or module code (except the contents of the `PhysicsEngineManager` class). All physics engine integration code will reside in a subpackage of the `physics` subpackage, and will only be interacted with via the relevant interfaces. Theoretically, this also allows multiple physics engine integrations to coexist, with the ability to change the physics engine used at _runtime_ (or in the worst case, a couple of lines of code).

### Introduces
- Wrapper around TeraBullet's `CollisionShape` (and derivatives) by way of interfaces of the same name under the `physics.shapes` subpackage
- `CollisionShapeFactory` interface for creating `CollisionShape`s for entities and blocks, primarily used in the prefab and block loading system
- Functional equivalent of TeraBullet's `Transform` in the `math` subpackage
- A static helper class `PhysicsEngineManager` containing a helper method which creates a new `PhysicsEngine` instance and a constant reference to a `CollisionShapeFactory` instance

### Changes
- Removes reliance on `com.bulletphysics.linearmath.AabbUtil2` by inlining functionally equivalent implementations of the methods used.
- Moves the collision shape component classes from `physics.shapes` to the more suited `physics.components.shapes` package
- Adds a custom `EPSILON` value to `Vector3fUtil` instead of referring to `BulletGlobals.SIMD_EPSILON`

### Removals
- Removes `MeshRenderer.addToBatch` (unrelated to the goals of this PR)

## Considerations

1. ~Introduce a `CompoundShapeFactory` that helps create  a `CompoundShape` and add child shapes to it in order to restrict the `CollisionShape` interface API to read-only methods, further increasing API security~ Not required, as discussed below.

## Future Work

1. Introduce more methods to the `CollisionShape` interface family that may be required in the future
2. Improve upon and widely adopt or fully remove the `Transform` class 
3. Add an `EPSILON` constant to the `PhysicsEngineManager` class, thereby making `PhysicsEngine.getEpsilon` redundant, or find a way to make available the value returned by `getEpsilon` everywhere without needing a `PhysicsEngine` instance (using a static method or a constant). This makes it easier to refer to the epsilon value used by the physics engine without needing to refer to the `PhysicsEngine` instance (like in the case of `Vector3fUtil.safeNormalize`).  
Related: #3204

## TODO
- Document 
  - [x] `CollisionShape` interface family
  - [x] `CollisionShapeFactory`
  - [x] `Transform`

Comments on the future considerations, general suggestions and improvements are welcome.


GooeyHub:

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletCollisionShapeFactory.java
```diff
@@ -0,0 +1,47 @@
...
+import org.terasology.math.geom.Vector3f;
+import org.terasology.physics.shapes.BoxShape;
+import org.terasology.physics.shapes.CollisionShapeFactory;
+import org.terasology.physics.shapes.CompoundShape;
+import org.terasology.physics.shapes.ConvexHullShape;
+import org.terasology.physics.shapes.SphereShape;
+
+import java.util.List;
+
+public class BulletCollisionShapeFactory implements CollisionShapeFactory {
```
pollend: What's the value of using this type of implementation? This just adds an additional layer of complexity to the implementation. What is the value of this if you are basically going to use this to just declare the object. This implementation just introduces complexity.

oniatus: This applies dependency inversion, the factory hides the implementation details from the used object/interface. You could switch this to any other instance of a CollisionShapeFactory and migrate the entire codebase to a different framework.

pollend: I understand a factory is used to generalize the implementation so it can be exchanged for something else. The entire model is already tied to how jbullet works. You can interface out the rest of the shapes and generalize it in the fashion or even wrap the objects themselves.

emanuele3d:
@pollend: Michael, the feedback you just provided is dismissive, destructive. 

If you wish to provide feedback make sure it is constructive. That is: make sure to understand what's the aim of the proposed changes and if you think there are better ways forward spend some time making alternative suggestions, respectful of the time @eviltak obviously already put in.

pollend:
@emanuele3d Sorry, i reacted poorly. The implementation looks fine the closer I looked.

GooeyHub:

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletCompoundShape.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private BulletCompoundShape(CompoundShape compoundShape) {
+        this.compoundShape = compoundShape;
+        underlyingShape = compoundShape;
+
+        childList = new ArrayList<>();
+    }
+
+    @Override
+    public void addChildShape(org.terasology.math.Transform transform, CollisionShape collisionShape) {
```
emanuele3d: Can't we provide a com.bulletphysics.linearmath.Transform in input too? Avoiding the toBulletTransform line?

engine/src/main/java/org/terasology/world/block/shapes/JsonBlockShapeLoader.java
```diff
@@ -120,7 +117,7 @@ public BlockShapeData deserialize(JsonElement json, Type typeOfT, JsonDeserializ
                 JsonObject collisionInfo = shapeObj.get(COLLISION).getAsJsonObject();
                 processCollision(context, shape, collisionInfo);
             } else {
-                shape.setCollisionShape(CUBE_SHAPE);
+                shape.setCollisionShape(PhysicsEngineManager.COLLISION_SHAPE_FACTORY.unitCube());
```
emanuele3d: Can the PhysicEngineManager qualifier be removed?

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -146,15 +144,56 @@ public AABB move(Vector3f offset) {
     }
 
     public AABB transform(Quat4f rotation, Vector3f offset, float scale) {
-        Transform transform = new Transform(new Matrix4f(VecMath.to(rotation), VecMath.to(offset), scale));
+        Transform transform = new Transform(offset, rotation, scale);
         return transform(transform);
```
emanuele3d: This is a bit ugly. I do not understand, by looking at this code, what the transform does to itself.

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
```diff
@@ -203,8 +202,7 @@ private void renderEntitiesByMaterial(SetMultimap<Material, EntityRef> meshByMat
                     location.getWorldPosition(worldPos);
                     float worldScale = location.getWorldScale();
 
-                    javax.vecmath.Matrix4f matrixWorldSpace = new javax.vecmath.Matrix4f(VecMath.to(worldRot), VecMath.to(worldPos), worldScale);
-                    transWorldSpace.set(matrixWorldSpace);
+                    Transform transWorldSpace = new Transform(worldPos, worldRot, worldScale);
```
emanuele3d: Could this variable be renamed `toWorldSpace`?

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
```diff
@@ -203,8 +202,7 @@ private void renderEntitiesByMaterial(SetMultimap<Material, EntityRef> meshByMat
...
 
                     Vector3f worldPositionCameraSpace = new Vector3f();
```
emanuele3d: I am aware you are inheriting this, but could this be renamed as `offsetFromCamera`?

engine/src/main/java/org/terasology/world/block/shapes/JsonBlockShapeLoader.java
eviltak: Not directly, but I can introduce a class level constant that maps to the constant in the `PhysicsEngineManager`.

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletCompoundShape.java
eviltak: Unfortunately, we cannot. The whole point of this refactor is to remove usages of physics engine-specific classes in the engine, and using the Bullet `Transform` class in an engine facing API would defeat that purpose. The whole engine is modified to use `org.terasology.math.Transform`, and it wouldn't make sense to change the accepted type. Furthermore, future inheritors may want to use their own `Transform` representation. It is thus the implementors responsibility to convert the `org.terasology.math.Transform` object into an object of its own liking.

engine/src/main/java/org/terasology/math/Vector3fUtil.java
```diff
@@ -56,9 +56,11 @@ public static Vector3f getPerpendicularComponent(Vector3f direction, Vector3f no
     }
 
     public static Vector3f safeNormalize(Vector3f v, Vector3f out) {
+        final float EPSILON = 0.000001f;
```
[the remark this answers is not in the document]
eviltak: I can make this constant global, but it wouldn't make sense because it isn't used anywhere else. If we can find a solution for consideration 4, we wouldn't need this constant at all.

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -146,15 +144,56 @@ public AABB move(Vector3f offset) {
...
     public AABB transform(Transform transform) {
-        javax.vecmath.Vector3f newMin = new javax.vecmath.Vector3f();
-        javax.vecmath.Vector3f newMax = new javax.vecmath.Vector3f();
-        AabbUtil2.transformAabb(VecMath.to(min), VecMath.to(max), 0.01f, transform, newMin, newMax);
-        return new AABB(VecMath.from(newMin), VecMath.from(newMax));
+        return transform(transform, 0.01f);
+    }
+
+    public AABB transform(Transform transform, float margin) {
+        // Courtesy of TeraBullet
```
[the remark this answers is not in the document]
eviltak: Yes, cut&paste&adapt of the code in the `AabbUtil2.transform` method in TeraBullet.

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -146,15 +144,56 @@ public AABB move(Vector3f offset) {
...
+
+        localHalfExtents.x += margin;
+        localHalfExtents.y += margin;
+        localHalfExtents.z += margin;
+
+        Vector3f localCenter = new Vector3f(max);
+        localCenter.add(min);
+        localCenter.mul(0.5f);
+
+        Matrix3f absBasis = transform.getBasis();
```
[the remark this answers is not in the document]
eviltak: This method is almost entirely (including the variable names) a copy of the  `AabbUtil2.transform` method in TeraBullet. This variable specifically is the absolute value matrix of the basis matrix. In other words, each matrix element is the absolute value of the corresponding basis matrix element.

engine/src/main/java/org/terasology/math/AABB.java
eviltak: Renaming the `transform` method to `transformWith` might help. Arguably, moving this method into the `Transform` class itself (`AABB Transform.transformAabb(AABB aabb)`) would be even better, but that would bring about even more changes not directly related to the physics engine. I think this change would be better off as another PR once this is merged, but if you don't think so I can do it here.

engine/src/main/java/org/terasology/math/AABB.java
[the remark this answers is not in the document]
emanuele3d: Then please let's write that, as Courtesy of TeraBuller is a bit vague.

engine/src/main/java/org/terasology/math/AABB.java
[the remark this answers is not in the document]
emanuele3d: That's not plain english, that's calculus speak. :wink: 

Why are the absolute values needed rather than the values already in the basis matrix?

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
eviltak: I see what you are getting at, but I think `worldSpaceTransform` would arguably be a better name, since transforms usually represent a single space (in this case, world space) to transform to/from.

engine/src/main/java/org/terasology/math/Vector3fUtil.java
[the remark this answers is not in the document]
emanuele3d: Actually, if you do a project-wide search for the string "epsil" you'll find a few other places where an epsilon is used to name a small number.

engine/src/main/java/org/terasology/world/block/shapes/JsonBlockShapeLoader.java
emanuele3d: You mean you can't write an import in the form:

import [...]PhysicsEngineManager.COLLISION_SHAPE_FACTORY

so that you can just write:

COLLISION_SHAPE_FACTORY.getUnitCube()

?

engine/src/main/java/org/terasology/math/AABB.java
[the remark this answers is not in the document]
eviltak: I am not too sure of the implementation myself, but my best guess is that the absolute values are taken to ensure that the rotation and scaling done is in the first octant and hence all elements of `extents` will always be positive.

engine/src/main/java/org/terasology/world/block/shapes/JsonBlockShapeLoader.java
eviltak: Yes, that may be possible. I keep forgetting Java has static imports 😄

engine/src/main/java/org/terasology/math/AABB.java
eviltak: Just for clarity, the transform does not do anything to itself. The `this` AABB is transformed (rotated, translated and scaled) according to the `transform` object and the transformed AABB is returned.

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletCollisionShapeFactory.java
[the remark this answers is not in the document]
eviltak: Done.

GooeyHub:

engine/src/main/java/org/terasology/math/Vector3fUtil.java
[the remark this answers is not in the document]
eviltak: True. Making a single `EPSILON` constant which is used everywhere is a good a idea, but is out of scope for this PR (at best, loosely related with consideration 4). The only reason I made this change was to remove the usage of  `BulletGlobals.SIMD_EPSILON` (in the TeraBullet package) in the `safeNormalize` method.

GooeyHub:

GooeyHub:

GooeyHub:

eviltak:
Thank you for the review @emanuele3d. Do you have any thoughts or suggestions on the considerations?

eviltak:
On further thought, I have decided to include consideration 1 (_Introduce a `CompoundShapeFactory` that helps create and a `CompoundShape` add child shapes to it in order to restrict the `CollisionShape` interface API to read-only methods, further increasing API security_) into the scope of this PR. Please do throw in comments and suggestions.

GooeyHub:

engine/src/main/java/org/terasology/math/AABB.java
emanuele3d: I know what you mean. It's just that this code seems to say that a transform is being transformed. When instead it is being passed through a chain of "transform" methods that (I'm guessing) are all needed at one point or another, down to the one actually doing the work. 

Hmmm. I can't think of a way to write it differently though. So, let's leave it as it is for now.

engine/src/main/java/org/terasology/math/AABB.java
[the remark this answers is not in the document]
emanuele3d: Ok, let's leave it as it is.

engine/src/main/java/org/terasology/math/Vector3fUtil.java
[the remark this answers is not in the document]
emanuele3d: Ok, the out-of-scope is a good argument. Can you please write a brief issue on the idea of having a global epsilon constant, so that the concept doesn't get lost?

engine/src/main/java/org/terasology/math/AABB.java
eviltak: I agree that it can be made more readable. Moving the method into the `Transform` class as I mentioned before would, in my opinion, make things much better, but this method must have been used in other places outside the physics engine, making a refactor out of scope for this PR.

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
eviltak: I can. Nice catch.

engine/src/main/java/org/terasology/math/Vector3fUtil.java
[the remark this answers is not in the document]
eviltak: Definitely. Will make one soon enough. I like the idea of centralizing the global epsilon, as part of the physics engine system or otherwise.

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
emanuele3d: Hmmmm... I don't think I agree with that description. 

A transformation is a mapping from one space to another. Sure, you can -interpret- the information stored in the transformation as creating a new space relative to some other frame of reference, i.e. the world space. But the way we use it in our context is to transform vectors from one space to another. Crucially, a specific transformation brings you from space A to space B but it then needs to be inverted to go the other way around. So, the name of a variable storing a transformation in my opinion should reflect the direction of the transformation.

eviltak: Your description does make more sense the more I think about it. However, I still feel that `toWorldSpace` is a bit lacking as a variable name - to me it sounds more like a method. How about `toWorldSpaceTransform`? I don't normally agree with the practice of including the type in the variable name, but I feel it is justified in this case.

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -146,15 +146,56 @@ public AABB move(Vector3f offset) {
...
     public AABB transform(Transform transform) {
-        javax.vecmath.Vector3f newMin = new javax.vecmath.Vector3f();
-        javax.vecmath.Vector3f newMax = new javax.vecmath.Vector3f();
-        AabbUtil2.transformAabb(VecMath.to(min), VecMath.to(max), 0.01f, transform, newMin, newMax);
-        return new AABB(VecMath.from(newMin), VecMath.from(newMax));
+        return transform(transform, DEFAULT_MARGIN);
+    }
+
+    public AABB transform(Transform transform, float margin) {
+        // Adaptation of AabbUtil2.transformAabb in the TeraBullet
```
emanuele3d: `// Adaptation of method AabbUtil2.transformAabb from the TeraBullet library. `

(remember the people reading your code in the future might not yet know what kind of animal TeraBullet is)

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
emanuele3d: Sure, I can live with that.

engine/src/main/java/org/terasology/math/AABB.java
eviltak: I understand. Would `from MovingBlocks' fork of JBullet` be better?

emanuele3d: Nah. In the context of Terasology project we talk about TeraBullet. In the context of the TeraBullet project we can talk about it being a fork.

eviltak: What do you recommend?

emanuele3d:
I did notice you asked me to look specifically in some of the considerations you list in the description of this PR. It's 2:22am however and I better go to sleep. I shall try and get back to you on those tomorrow evening.

engine/src/main/java/org/terasology/math/AABB.java
emanuele3d: // Adaptation of method AabbUtil2.transformAabb from the TeraBullet library.

as stated above. :wink:

eviltak: Lol. I thought you were simply restating what I wrote but didn't realize that you were suggesting a change 🙄

GooeyHub:

emanuele3d:
@eviltak: I have gone through the considerations (a section I seem to understand would be better named as "Future Work") but I don't have enough knowledge of the physics engine's architecture to comment. Worse, I don't know who does! Perhaps @flo and @msteiger, but I'm not sure. 

Nevertheless, I'd recommend to spend some time adding javadocs to the various public classes/methods introduced or modified by this PR. 

I'd also like to commend the very well written description introducing this PR. Thank you for spending time writing it in such detail: I wish all PRs had such in-depth introduction. I also appreciate the very well named commits. I suspect some could be grouped into a single commit - something to think about for the next PR. 

So, let's add the icing on the cake (the javadocs) and if nobody has objection let's get it all merged in.

eviltak:
@emanuele3d thanks. The javadocs and a couple of rename changes are still left from your previous review before this PR is ready for merging. I will finish them by this week, RL is a bit hectic.

From an architectural viewpoint, would you rather:
- have creation and modification methods (specifically `CompoundShape.addChildShape`) exposed via the _interface itself_ to a public module-facing API, or 
- abstract the creation of objects out to an engine specific helper class/factory/builder? 

Your answer to the above question will help me figure out what to do with consideration 1.

In context, I don't see why modules should be allowed to modify a compound shape (unless they are creating new ones programmatically, in which case the creation helper can be used), nor have I encountered any use cases of the `addChildShape` method for _modifying_ (not creating) a `CompoundShape`. 

Choosing the latter option will a) force the creation of yet another `CompoundShapeBuilder` interface for physics engine integrators to implement, and b) prevent modification of a `CompoundShape`, as is the case with other collision shapes. 

To be honest, I quite fancy the idea of immutable collision shapes. They do not cause any problems in the current code base, but they might cause future problems that I haven't foreseen.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

eviltak:
All pending changes have been pushed. Once a decision is reached on the only consideration, this PR can be merged.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

emanuele3d:
Documentation review in progress, please standby.

engine/src/main/java/org/terasology/physics/shapes/CollisionShapeFactory.java
```diff
@@ -0,0 +1,60 @@
...
+public interface CollisionShapeFactory {
+    /**
+     * Creates a new box shape with the given extents.
+     * @param extents The full extents of the box shape.
+     * @return The box shape.
+     */
+    BoxShape getNewBox(Vector3f extents);
+
+    /**
+     * Creates a new convex hull shape with the given vertices.
```
emanuele3d: Do you mean, creates a new convex hull shape _enveloping_ the given vertices or are the vertices meant to be a valid (if unconnected) convex hull already? That is, can this method cope with a generic point cloud or does it require just the outer points of the cloud? 

I'm guessing it is the first option. If that's the case the following line should also be changed in something like: "The vertices to be wrapped by the convex hull."

engine/src/main/java/org/terasology/physics/shapes/CollisionShapeFactory.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+
+/**
+ * Factory interface to create new collision shapes.
+ */
+public interface CollisionShapeFactory {
+    /**
+     * Creates a new box shape with the given extents.
+     * @param extents The full extents of the box shape.
+     * @return The box shape.
```
emanuele3d: This comment is valid for the other methods in this interface: it can't return -the- box shape if it returns a new one every time. I'd rewrite as "A box shape".

engine/src/main/java/org/terasology/physics/shapes/CompoundShape.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.physics.shapes;
+
+import org.terasology.math.Transform;
+
+/**
+ * Represents a compound collision entity composed of other {@link CollisionShape}s in the physics engine.
```
emanuele3d: I'd remove "in the physics engine" - it's clear enough without it.

engine/src/main/java/org/terasology/physics/shapes/CompoundShape.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public interface CompoundShape extends CollisionShape {
+    /**
+     * Adds a child shape to the compound shape.
```
emanuele3d: I'd add an empty line after this one.

engine/src/main/java/org/terasology/physics/shapes/ConvexHullShape.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.physics.shapes;
+
+/**
+ * Represents a convex hull collision entity in the physics engine.
```
emanuele3d: I'd remove "in the physics engine". As it's part of the "physics" package, I think it's clear enough. Other shapes' javadoc have the same problem.

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
```diff
@@ -203,14 +202,13 @@ private void renderEntitiesByMaterial(SetMultimap<Material, EntityRef> meshByMat
...
 
-                    Vector3f worldPositionCameraSpace = new Vector3f();
-                    worldPositionCameraSpace.sub(worldPos, cameraPosition);
-                    Matrix4f matrixCameraSpace = new Matrix4f(worldRot, worldPositionCameraSpace, worldScale);
+                    Vector3f offsetFromCamera = new Vector3f();
+                    offsetFromCamera.sub(worldPos, cameraPosition);
+                    Matrix4f matrixCameraSpace = new Matrix4f(worldRot, offsetFromCamera, worldScale);
 
-                    AABB aabb = meshComp.mesh.getAABB().transform(transWorldSpace);
+                    AABB aabb = meshComp.mesh.getAABB().transform(toWorldSpaceTransform);
```
emanuele3d: One of my coding style aims is to write code that is as close as possible to plain English. In this case, I realise I OK'd the variable "toWorldSpaceTransform", but "toWorldSpace"  would have allowed to read this line as: "get the AABB of the mesh and transform it to world space". Right now it reads as "transform the transform". It's no big deal though and you can leave it. It's just an example of what I aim style-wise.

emanuele3d:
Regarding your question, I am certainly capable of thinking use cases where modules should be able to add children shapes. I.e. say we have a module providing spaceships (but could also be vehicles, airplanes or robots) and then a second module, developed subsequently and independently, adding all sorts of visible and voluminous components to said spaceships. Unless the first module developer had the foresight to allow for the addition of spaceship components, the second module needs to add its spaceship components onto the existing spaceships and needs to modify their collision shapes accordingly. 

Going the immutable way would require the second module to replace the original collision shape of a ship with a copy including the original and the new pieces. 

What do you think?

emanuele3d:
P.S. I'll be traveling back to Europe and I might be offline for 24-36 hours.

engine/src/main/java/org/terasology/math/AABB.java
eviltak: Similar to how the `Transform.transform` method (which, I think should be renamed to `transformPosition`) is:
```
public AABB transformAabb(AABB aabb)
```

engine/src/main/java/org/terasology/math/Transform.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+package org.terasology.math;
+
+import org.terasology.math.geom.Matrix3f;
+import org.terasology.math.geom.Quat4f;
+import org.terasology.math.geom.Vector3f;
+
+/**
+ * Represents a space transformation. Used for transforming points from one space to another space and
+ * vice versa.
```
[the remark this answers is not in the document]
eviltak: The current implementation does not include a vector transformation, and so I decided against generalizing the documentation for now. 

Truth be told, the `Transform` implementation is incomplete - it is missing inverse transformation methods, direction (only rotation) and vector (only rotation and scale) transformation methods, and a few helper accessors. I planned on completing the implementation as a separate PR after proper discussion, as this one is already pretty big and mainly focuses on getting rid of TeraBullet dependencies in the engine.

engine/src/main/java/org/terasology/math/Transform.java
```diff
@@ -0,0 +1,77 @@
...
+        Matrix3f basis = new Matrix3f();
+
+        basis.set(rotation);
+        basis.mul(scale);
+
+        return basis;
+    }
+
+    /**
+     * Transforms the point {@code v} into the space represented by this transform. The result is placed
```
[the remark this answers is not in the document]
eviltak: `v` is meant to be a position in this case. As I said in a previous comment, `Transform` is in its current state incomplete and not at all how I would want it, but I intend on addressing the improvements in a PR right after this one is merged.

engine/src/main/java/org/terasology/physics/shapes/BoxShape.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.physics.shapes;
+
+import org.terasology.math.geom.Vector3f;
+
+/**
+ * Represents a box shaped collision entity in the physics engine.
```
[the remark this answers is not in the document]
eviltak: The collision shape components are different (you can find them under `physics.components.shapes`). The `PhysicsEngine` implementation is responsible for creating a `CollisionShape` from the collision shape component on an entity.

As such, `CollisionShape` and its sub-interfaces are meant to represent collision entities (i.e. a physical object that takes up volume and can collide with other physical objects, not Terasology entities) in the physics engine. I put in the `in the physics engine` qualifier to help explain that these interfaces represent an object in the physics engine _only_.

engine/src/main/java/org/terasology/physics/shapes/BoxShape.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public interface BoxShape extends CollisionShape {
+    /**
+     * Returns the half extents of the box shape without adding a margin.
```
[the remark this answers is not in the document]
eviltak: As per the goals of this PR, `BoxShape` currently serves as a wrapper to the `com.bulletphysics.collision.shapes.BoxShape` class in TeraBullet. `getHalfExtentsWithoutMargin` is simply a wrapper around the `com.bulletphysics.collision.shapes.BoxShape.getHalfExtentsWithoutMargin` class. However, I do agree that changing this to `getExtents` will simplify the API. I am not quite sure whether to include this change in the current PR or take it up separately like `Transform` as another PR. I have few other tweaks to the `CollisionShape` family of interfaces in mind too, but I think it would be best to take them up in a separate PR.

engine/src/main/java/org/terasology/physics/shapes/CollisionShape.java
```diff
@@ -0,0 +1,41 @@
...
+import org.terasology.math.Transform;
+import org.terasology.math.geom.Quat4f;
+
+/**
+ * The base type representing a collision entity in the physics engine.
+ */
+public interface CollisionShape {
+    /**
+     * Returns the axis-aligned bounding box ({@link AABB}) for the shape in the space represented
+     * by {@code transform}.
```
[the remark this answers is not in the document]
eviltak: It returns an AABB encompassing the translated, rotated, and scaled `CollisionShape` aligned with the global axes.

engine/src/main/java/org/terasology/physics/shapes/CollisionShapeFactory.java
eviltak: I too support the first option, but will have to check if the `BulletCollisionShapeFactory` implementation of this method follows this norm. I shall update the docs now, but if possible, I shall address the actual implementation in the separate PR dealing with only `CollisionShapes`.

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
eviltak: I do understand what you are getting at and how it would be more readable in this case, but I feel that an even better option would be to move the method entirely into the `Transform` class as discussed above.

eviltak:
@emanuele3d I appreciate and agree with your assessment. I prefer mutability to avoid breaking references, if that is ever going to be a problem. That's how it is done in other engines too, and I don't see why it should be any different.

I posed this question because in the current state, `CompoundShape` is the only _mutable_ `CollisionShape`. However, as I mentioned before, I feel we should leave this as-is for now, merge in the changes, and take up all further `CollisionShape` API improvements in a near-future PR. The PR shall deal with addition of getters for each `CollisionShape` (eg. `BoxShape.getExtents` and `SphereShape.getRadius`), and if mutability is decided upon, the corresponding setters too.

engine/src/main/java/org/terasology/physics/shapes/CollisionShapeFactory.java
eviltak: I understand. How about "The created box shape" for retaining context?

GooeyHub:

GooeyHub:

emanuele3d:
Sorry for the long delay @eviltak, I have a couple of meeting this week (Tue and Wed) and preparing for them is taking all my time. Please stand-by.

eviltak:
Waiting on clarification on a couple of points raised by @emanuele3d, otherwise this PR should be good to go.

GooeyHub:

oniatus:
@eviltak and @emanuele3d I think there are some open review points left, besides that I assigned me and @flo for further review and to have some more eyes on this :)
Can you point out where you need a second opinion or what is not clear right now? =)

eviltak:
@oniatus I just need clarification on @emanuele3d's comments on the docs in light of my replies to the review. Just waiting for this PR to be merged (or to be approved at least) so that I can open the two other PRs (`Transform` improvements and `CollisionShape` API extensions) that depend on this one.

engine/src/main/java/org/terasology/physics/shapes/BoxShape.java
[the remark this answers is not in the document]
emanuele3d: Ok, just make sure to write it somewhere!

engine/src/main/java/org/terasology/physics/shapes/CollisionShape.java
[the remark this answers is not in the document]
emanuele3d: Ok, then let's write:

`* Returns the axis-aligned bounding box ({@link AABB}) of the transformed shape.`

engine/src/main/java/org/terasology/physics/shapes/CollisionShape.java
```diff
@@ -26,6 +26,7 @@
     /**
      * Returns the axis-aligned bounding box ({@link AABB}) for the shape in the space represented
      * by {@code transform}.
+     *
      * @param transform The {@link Transform} pertaining to the space in which the AABB is to be calculated.
      * @return The {@link AABB} bounding the shape.
      */
```
emanuele3d:
```
* @param transform The {@link Transform} to apply to the shape, after which the AABB is calculated.
* @return The {@link AABB} bounding the transformed shape.
```

Question that pops in my mind just by reading this javadoc: is the shape itself actually, permanently transformed or is this method meant to transform the shape only for the purpose of returning an AABB, with the shape stored in the object remaining untouched? I guess whatever the answer we need to get that across in this javadoc.

engine/src/main/java/org/terasology/math/Transform.java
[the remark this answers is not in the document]
emanuele3d: Ok.

engine/src/main/java/org/terasology/math/Transform.java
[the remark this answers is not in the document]
emanuele3d: Ok.

engine/src/main/java/org/terasology/physics/shapes/CompoundShape.java
emanuele3d: Ok, after a comment above I better understand what you mean now. We need to disambiguate the ambiguity though. Despite the three repetitions of the word "shape" I'd write:

`* Represents a compound collision shape, a shape composed of other {@link CollisionShape}s, within the physics engine. `

emanuele3d: The alternative would be to keep the term "entity", but then we'd have to clarify with a note that we are referring to entities within the physics engine rather Terasology's entities.

engine/src/main/java/org/terasology/physics/shapes/ConvexHullShape.java
emanuele3d: See comment above for similar issue.

engine/src/main/java/org/terasology/physics/shapes/SphereShape.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.physics.shapes;
+
+/**
+ * Represents a sphere shaped collision entity in the physics engine.
```
emanuele3d: See comment above. I guess another way you could write this is:

`* Represents a sphere shaped collision entity within the physics engine. Notice that the physics engine has its own entity concept/object, and this is not related to Terasology's entities.`

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
emanuele3d: I think I addressed this issue in another comment.

engine/src/main/java/org/terasology/rendering/opengl/OpenGLMesh.java
```diff
@@ -187,7 +188,7 @@ public int addToBatch(Transform transform, Transform normalTransform, TFloatList
         int n = 0;
         int c = 0;
         for (int v = 0; v < data.getVertices().size(); v += VERTEX_SIZE) {
-            javax.vecmath.Vector3f vert = new javax.vecmath.Vector3f(data.getVertices().get(v), data.getVertices().get(v + 1), data.getVertices().get(v + 2));
+            Vector3f vert = new Vector3f(data.getVertices().get(v), data.getVertices().get(v + 1), data.getVertices().get(v + 2));
```
emanuele3d: Please wrap data.getVertices() and data.getNormals() (below) in variables, to shorten the lines and improve readability.

engine/src/main/java/org/terasology/physics/shapes/CollisionShape.java
eviltak: The latter: the shape is itself not modified, it is transformed only for the purpose of returning the AABB.

engine/src/main/java/org/terasology/physics/shapes/SphereShape.java
eviltak: Like you suggested above, renaming the `entity` references to `shape` (`Represents a box/sphere/convex hull/compound collision shape`) shall work in my opinion.

engine/src/main/java/org/terasology/math/AABB.java
eviltak: Now that I think about this with a fresh mind, I have to agree with your assessment. It makes sense to rename this method to `transformWith` or `transformTo` or similar (or something entirely different), but I think this change (and any further deliberations) will be better suited to the subsequent Transform improvement PR.

engine/src/main/java/org/terasology/rendering/opengl/OpenGLMesh.java
eviltak: Bigger but potentially out of scope problem - the `addToBatch` method is not used anywhere (as reported by IntelliJ). Would removing it altogether be a better idea? If so, I'd rather not make the change here since the method is going to be removed anyway.

GooeyHub:

engine/src/main/java/org/terasology/rendering/opengl/OpenGLMesh.java
emanuele3d: Ok, let's remove it then. Let's see if anything breaks.

engine/src/main/java/org/terasology/math/AABB.java
emanuele3d: Ok.

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
emanuele3d: So, to be clear, I think `toWorldSpaceTransform` should still become `toWorldSpace`.

engine/src/main/java/org/terasology/math/AABB.java
emanuele3d: Just write it down somewhere!

vampcat:
@eviltak a small suggestion for future: Is possible, avoid creating new vector objects inside methods that are called frequently. For instance, in the `transform` method, you create a new vector for `localHalfExtents` on every invocation. That leads to GC spikes, and can be easily avoided by having a temp vector object in the class.

I know this is not an immediate priority by any means, but I just thought of throwing it out here because you seem inclined to work on physics, and I would rather not increase the amount of classes I have to rework in my soon-to-arrive optimization PR.

eviltak:
@vampcat Thank you, I always seem to forget that Java does not have stack allocated objects! I shall keep that in mind in the future. The optimization will be made as part of the upcoming Transform PR.

GooeyHub:

emanuele3d:
Ok, I helped best I can given how unfamiliar I am with this part of the codebase. 

@Cervator: from what I can see this PR is good to go!

That was a lot of work for you @eviltak: hats off for the persistence.

eviltak:
@Cervator and all those concerned: This PR is expected to break anything that currently accesses TeraBullet collision shape instances of a block (or any other item), like in

 https://github.com/MovingBlocks/Terasology/blob/6d1b75c89212fcd39f32e277e6fbd5b34c0fa4b8/engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java#L102-L103

Prepare for mismatched type errors across modules!

Cervator:
Took this for a spin locally. Just three modules seemed to be affected, and that was from package moves, not mismatched type errors. Should I be worried that no more stuff broke? :-)

* https://github.com/Terasology/CombatSystem/pull/34
* https://github.com/Terasology/Sensors/pull/3
* https://github.com/Terasology/ItemPipes/pull/5

I'll merge those in a bit after the engine build finishes. Yay API free change period!

I tested all affected modules and a bit more in general, as far as I can tell everything works normally.

One additional change I had to apply with https://github.com/MovingBlocks/Terasology/commit/2a9460052515ccfa0dfe470bed0ef32e2fa63984 - since there was a package move from a previously API flagged package (via a `package-info.java`) it was missing in the destination package that those three modules did rely on. So I added the package file figuring that to be a sensible fix.

Nice work! Looking forward to the next stage, and the other physics PRs - those are just a bit harder to test, validate, and merge right :-)

PR #3201, opened 2017-12-27
Added overloads for methods in ChunkImpl

Adrijaned:
### Contains

Added overload for setBiome() and getBiome() with BaseVector3i, like set/getBlock() and set/getLiquid() already had.

### DIFF
--- a/engine/src/main/java/org/terasology/world/chunks/CoreChunk.java
+++ b/engine/src/main/java/org/terasology/world/chunks/CoreChunk.java
@@ -49,7 +49,11 @@ public interface CoreChunk {
 
     Biome setBiome(int x, int y, int z, Biome biome);
 
+    Biome setBiome(BaseVector3i pos, Biome biome);
+
     Biome getBiome(int x, int y, int z);
+
+    Biome getBiome(BaseVector3i pos);
 
     void setLiquid(BaseVector3i pos, LiquidData state);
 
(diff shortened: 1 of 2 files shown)


GooeyHub:

emanuele3d:
@Adrijaned: is this part of a GCI task? Is it the answer to an existing issue? It would be good if the description had some context.

Adrijaned:
@emanuele3d nope, I am doing the world generator series of GCI and the absence of these methods has just been bugging me for some time now. Also, I'm not aware of any issues related to these.

emanuele3d:
The general rule is that we add things when there is a "business case" for them. Right now these two methods are unused: no other part of the codebase takes advantage of them. 

Are you planning to take advantage of the new methods somehow?

Adrijaned:
I am. The module I am writing now actually uses the new methods. In WorldRasterizers, when you are generating blocks or setting liquid data, you may just write:
```
chunk.setBlock(pos, water);
chunk.setLiquid(pos, waterLiquid);
```
However, when setting the biome, which works basically the same, you have to write:
```
chunk.setBiome(pos.x, pos.y, pos.z, biome);
```
Both of these examples were copied from [org.terasology.core.world.generator.rasterizers.SolidRasterizer](https://github.com/MovingBlocks/Terasology/blob/develop/modules/Core/src/main/java/org/terasology/core/world/generator/rasterizers/SolidRasterizer.java).

Cervator:
Checked this out and am merging - it is a nice little bit of utility, thanks!

Without a default implementation in the base interface it may count as an API violation as `CoreChunk` is `@API` so there could be additional child classes. But as far as I can tell there are only two extending interfaces and both those are then used for the one final implementor. I'm fairly confident (tho not entirely sure) there are no other implementations that would be broken by this.

Default methods *could* be added fairly easily, but I don't think it is worth it this time just for the exercise :-)

Cervator:
Oh: one additional comment: if at all possible try avoiding merge commits in PRs. They cloud up revision history and can be problematic in some cases. Use Git Rebase instead or if it makes no big difference just leave your branch a few commits behind. Using topic branches is also an additional nicety :-)

PR #3206, opened 2017-12-30
Unified Utility Script for Groovy Wrapper

digitalripperynr:
Adds a unified script for all utility commands for the Groovy Wrapper.

General Syntax:
`groovyw util (type) (sub-command)`

Check more information using `groovyw util usage`


GooeyHub:

Cervator:
Looking around some - it would be useful in general if you could elaborate a bit on the changes made and any non-obvious design decisions made :-)

From what I can tell initially it looks pretty good, nicely done. The typed-item approach should work. Some notes:

* The handling of the GitHub home seems a little suspect. Have you tested that it respects an alternative org home set in `gradle.properties` ? What about meta modules since they read a different property?
* There are some code style issues, mostly minor like whitespace. Missing spaces, sometimes indentation is 2 spaces sometimes 4, etc. It might be possible to scan it with Checkstyle with a bit of tweaking
* There are some javadoc/comment bits that could still be updated some, like recurse only working for modules (only indicated in some places at present)
* The create item isn't handling the types quite right, but is close. The modules have `module.txt` and `build.gradle` at present, but the meta modules doesn't have any Gradle at all and the facades should get the template file for facades, not modules
* `updateItem` seems to have been renamed away from module, but its content still seems to be specific to modules? At least in naming/logging
* The Groovy fall through warning override isn't in place for the case switch, so there's a warning about a missing break (which is intentional and why there was an override there)
* One of the `getUserString` calls uses `String` and the other `GString` (`"` vs `'`) - that should mean the `$displayName` variable won't expand in one case. Did you test for that?
* Could `displayName` be renamed to something clearer?
* I think the `module update-all` could use a URL check - I managed to get it to break as I had a module where the remote was no longer valid
* Trying to test `meta` leads to trouble as I guess it requires the rename of the "meta" to "metas" - right? That's what I think we should do. If you could confirm with the rename and add the rename to the PR I'd appreciate it :-)

util.groovy
```diff
@@ -0,0 +1,502 @@
...
+/**
+ * Default settings for modules.
+ * githubRepo stores the Repository name.
+ * targetDirectory stores the working directory.
+ * displayName is used to generalize the script.
+ */
+githubRepo="Terasology"
+targetDirectory="modules"
+displayName="Module"
+//Main Logic for changing githubRepo, targetDirectory & displayName
```
mjuvekar7: add blank line before this, and make this a Javadoc-style comment (to maintain consistency throughout)

digitalripperynr: Okay!

digitalripperynr:
A list of chanmges that I made :
- The `displayName` variable is renamed to `itemType` for more code clarity.
- The `createItem` now acts differently for modules, meta and facades as intended.
- The `updateItem` is no longer specific to modules in logs.
- The rename of meta folder was suggested by you for meta.groovy. So I renamed it to metas. I will add it in the next PR.

digitalripperynr:
I fixed all the issues listed by Cervator.
The only thing left now is the implementation of URL validation for update command.

Please review the changes.

Cervator:
Thanks for all the changes and keeping at it. Task has been approved already but I wanted to leave a note here that I'll get to this PR (and other Groovy ones) whenever I get a chance to spend some time on final organizing and docs :-)

digitalripperynr:
The '-remote' option should have an example in the help section. It is a bit confusing for new users.
I'll add that when this get approved and is ready to merge :)

jellysnake:
I'm not convinced that this is a needed change. It does increase code reuse but it also adds an unnecessary `util` to the command and makes the single file rather bloated. Is there a way to have a central file containing the utility functions but otherwise leave it intact?

mjuvekar7:
I agree with jelly. The current way is much more extensible, in that you can add more commands by just adding a new file. Maybe a better way to organize things would be to just put all those files (module.groovy, etc) in a single directory

Cervator:
Superseded / merged via #3263 - thanks! :-)

PR #3213, opened 2018-01-06
Add a logger when @Sender annotation is missing in EntityRef parameters

### LINKED ISSUE #2679: Log error for commands with missing @Sender parameters
Got a NPE when running the [hungerCheck](https://github.com/Terasology/Hunger/blob/master/src/main/java/org/terasology/hunger/HungerAuthoritySystem.java#L158) command, the client entity was null (tested in singleplayer and local hosted server). This could be easily fixed by adding the `@Sender` annotation to the method parameter. However there is no indication that the annotation is missing.

I think there is no way for a player to submit an EntityRef by the console (correct me if i am wrong), so any EntityRef parameter in a command method will always be null if there is no `@Sender` added.
We should log a better error in this case, preferred when such a command is registered for the first time, instead of silently registering it and passing null for the argument.

[The wiki](https://github.com/MovingBlocks/Terasology/wiki/Developing-Commands#commands-and-multiplayer) seems to be outdated too.

TheFlash98:
### Contains

Fixes #2679 

### How to test

As mentioned by @oniatus in that issue, download the Hunger module and remove `@Sender` annotation from `hungerCheck` command. Run the game as a host and you will see a log saying `Command hungerCheck contains a EntityRef without @Sender annotation, may cause a NullPointerException`. Already pinged oniatus for potential testing. 🙂

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/commandSystem/MethodCommand.java
+++ b/engine/src/main/java/org/terasology/logic/console/commandSystem/MethodCommand.java
@@ -41,6 +41,7 @@ import java.util.Set;
  */
 public final class MethodCommand extends AbstractCommand {
     private static final Logger logger = LoggerFactory.getLogger(MethodCommand.class);
+    private static final String ENTITY_REF_NAME = "org.terasology.entitySystem.entity.EntityRef";
 
     private MethodCommand(Name name, String requiredPermission, boolean runOnServer, String description, String helpText,
                           SpecificAccessibleObject<Method> executionMethod, Context context) {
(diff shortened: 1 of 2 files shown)


GooeyHub:

oniatus:
Good start :)

I would like to add the providing class to the log message so that the command source is easy to find from the logs.
Something like
`logger.error("Command {} provided by {} contains a EntityRef without @Sender annotation, may cause a NullPointerException", method.getName(), provider.getClass().getSimpleName())` as log output should work.

I added an extra commit with tests for the fix to your branch, you can use them as orientation if you want.

There is also a small bug in your inner loop: If EntityRef has multiple Annotations the code would log once for every annotation which is not `@Sender`. Instead the code should try to find the annotation for the entityref and log once if it can not be found. This separation (find annotation -> then log) would also allow to eliminate the duplicate log statements.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/logic/console/commandSystem/MethodCommand.java
```diff
@@ -89,17 +89,20 @@ public static void registerAvailable(Object provider, Console console, Context c
         for (Method method : commandMethods) {
             final Class[] paramTypes = method.getParameterTypes();
             final Annotation[][] paramAnnotations = method.getParameterAnnotations();
+            boolean hasSenderAnnotation = false;
             for (int i = 0; i < paramAnnotations.length; i++) {
                 if(paramTypes[i].getTypeName().equals(ENTITY_REF_NAME)) {
                     if(paramAnnotations[i].length == 0) {
-                        logger.info("Command {} contains a EntityRef without @Sender annotation, may cause a NullPointerException", method.getName());
+                        logger.error("Command {} provided by {} contains a EntityRef without @Sender annotation, may cause a NullPointerException", method.getName(), provider.getClass().getSimpleName());
```
oniatus: You could replace this line with `hasSenderAnnotation=false`.
Then you can extract Line 90 to 102 to a method `hasSenderAnnotation0.`
Inside this method the `hasSenderAnnotation` can be replaced with return statements.

Then this should be good to go 👍

TheFlash98:
Hey @oniatus thank you for your valuable review. What changes would you like to see in here https://metaterasology.github.io/docs/developing/entitySystem/commands.html I think I could add a `Note` about what this PR has made/changed? If there's anything else outstanding please feel free to tell me :-)

GooeyHub:

TheFlash98:
@oniatus made a PR MetaTerasology/metaterasology.github.io#1 adding a few lines about about what this PR addresses. 🙂

GooeyHub:

GooeyHub:

oniatus:
😅 Maybe I should have provided a better documentation how to edit the documentation =)
https://github.com/MovingBlocks/TerasologyDocs contains the sources and as I noticed, also an issue for the sender annotation: https://github.com/MovingBlocks/TerasologyDocs/issues/6

https://github.com/MetaTerasology/metaterasology.github.io is a target repository with the build output from these sources but the build overrides the repository each time the documentation is build :)

TheFlash98:
Hey @oniatus I didn't know that. Quite n00by of me haha. Anyways, I've made the PR to the docs. 
https://github.com/MovingBlocks/TerasologyDocs/pull/11
Closed the one to metaterasology.github.io 🙂

oniatus:
Looks good to me, I will wait a bit longer with the documentation PR as we are currently discussing if we move the documentation somewhere more visible and/or to a different format but that is something I will take care of :)
Besides that this PR should be good to go, nice small improvement 👍

PR #3214, opened 2018-01-07
Bouncy Objects

SurajDatta:
# Description
Hi, I have added the following codes:
- command for prepare 10 Bowling pins
- command for dropping many blocks
# How to test
- to activate bulk dropper type `bulkDrop (your_block_name) (drop_amount)`
- to activate bowling pin type `bowlingPrep (your_block_name)`

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -148,7 +148,7 @@ public class CoreCommands extends BaseComponentSystem {
      */
     @Command(shortDescription = "Search commands/prefabs/assets",
             helpText = "Displays commands, prefabs, and assets with matching name, description, "
-                    + "help text, usage or required permission")
+                + "help text, usage or required permission")
     public String search(@CommandParam("searched") String searched) {
         String searchLowercase = searched.toLowerCase();
 
@@ -158,7 +158,7 @@ public class CoreCommands extends BaseComponentSystem {
 
         // String containing numbers of commands, prefabs and block that match searched string
         String result = "Found " + commands.size() + " command matches, " + prefabs.size() +
-                " prefab matches and " + blocks.size() + " block matches when searching for '" + searched + "'.";
+            " prefab matches and " + blocks.size() + " block matches when searching for '" + searched + "'.";
 
         // iterate through commands adding them to result
         if (commands.size() > 0) {
@@ -188,7 +188,7 @@ public class CoreCommands extends BaseComponentSystem {
      */
     private List<String> findCommandMatches(String searchLowercase) {
         return console.getCommands().stream().filter(command -> matchesSearch(searchLowercase, command))
-                .map(ConsoleCommand::getUsage).collect(Collectors.toList());
+            .map(ConsoleCommand::getUsage).collect(Collectors.toList());
     }
 
     /**
@@ -199,10 +199,10 @@ public class CoreCommands extends BaseComponentSystem {
      */
     private static boolean matchesSearch(String searchLowercase, ConsoleCommand command) {
         return command.getName().toLowerCase().contains(searchLowercase)
-                || command.getDescription().toLowerCase().contains(searchLowercase)
-                || command.getHelpText().toLowerCase().contains(searchLowercase)
-                || command.getUsage().toLowerCase().contains(searchLowercase)
-                || command.getRequiredPermission().toLowerCase().contains(searchLowercase);
+            || command.getDescription().toLowerCase().contains(searchLowercase)
+            || command.getHelpText().toLowerCase().contains(searchLowercase)
+            || command.getUsage().toLowerCase().contains(searchLowercase)
+            || command.getRequiredPermission().toLowerCase().contains(searchLowercase);
     }
 
     /**
@@ -212,8 +212,8 @@ public class CoreCommands extends BaseComponentSystem {
      */
     private List<String> findPrefabMatches(String searchLowercase) {
         return StreamSupport.stream(prefabManager.listPrefabs().spliterator(), false)
-                .filter(prefab -> matchesSearch(searchLowercase, prefab))
-                .map(prefab -> prefab.getUrn().toString()).collect(Collectors.toList());
+            .filter(prefab -> matchesSearch(searchLowercase, prefab))
+            .map(prefab -> prefab.getUrn().toString()).collect(Collectors.toList());
     }
 
     /**
@@ -224,7 +224,7 @@ public class CoreCommands extends BaseComponentSystem {
      */
     private static boolean matchesSearch(String searchLowercase, Prefab prefab) {
         return prefab.getName().toLowerCase().contains(searchLowercase)
-                || prefab.getUrn().toString().toLowerCase().contains(searchLowercase);
+            || prefab.getUrn().toString().toLowerCase().contains(searchLowercase);
     }
 
     /**
@@ -234,9 +234,9 @@ public class CoreCommands extends BaseComponentSystem {
      */
     private List<String> findBlockMatches(String searchLowercase) {
         return assetManager.getAvailableAssets(BlockFamilyDefinition.class)
-                .stream().<Optional<BlockFamilyDefinition>>map(urn -> assetManager.getAsset(urn, BlockFamilyDefinition.class))
-                .filter(def -> def.isPresent() && def.get().isLoadable() && matchesSearch(searchLowercase, def.get()))
-                .map(r -> new BlockUri(r.get().getUrn()).toString()).collect(Collectors.toList());
+            .stream().<Optional<BlockFamilyDefinition>>map(urn -> assetManager.getAsset(urn, BlockFamilyDefinition.class))
+            .filter(def -> def.isPresent() && def.get().isLoadable() && matchesSearch(searchLowercase, def.get()))
+            .map(r -> new BlockUri(r.get().getUrn()).toString()).collect(Collectors.toList());
     }
 
     /**
@@ -494,7 +494,7 @@ public class CoreCommands extends BaseComponentSystem {
      * @return String containg final message
      */
     @Command(shortDescription = "Spawns a block in front of the player", helpText = "Spawns the specified block as a " +
-            "item in front of the player. You can simply pick it up.", runOnServer = true, requiredPermission = PermissionManager.CHEAT_PERMISSION)
+        "item in front of the player. You can simply pick it up.", runOnServer = true, requiredPermission = PermissionManager.CHEAT_PERMISSION)
     public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") String blockName) {
         ClientComponent clientComponent = sender.getComponent(ClientComponent.class);
         LocationComponent characterLocation = clientComponent.character.getComponent(LocationComponent.class);
@@ -594,7 +594,7 @@ public class CoreCommands extends BaseComponentSystem {
      * @return String containing ping or error message
      */
     @Command(shortDescription = "Your ping to the server", helpText = "The time it takes the packet " +
-            "to reach the server and back", requiredPermission = PermissionManager.NO_PERMISSION)
+        "to reach the server and back", requiredPermission = PermissionManager.NO_PERMISSION)
     public String ping(@Sender EntityRef sender) {
         Server server = networkSystem.getServer();
         if (server == null) {
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -373,7 +373,6 @@ public class CoreCommands extends BaseComponentSystem {
         } else {
             return "Switched to windowed mode";
         }
-
     }
 
     /**
@@ -619,7 +618,6 @@ public class CoreCommands extends BaseComponentSystem {
             return String.format("Error: Failed to ping server \"%s\" at %s:%s -- %s", remoteAddress[0], remoteAddress[1], remoteAddress[2], e);
         }
     }
-
 
     /**
      * Prints out short descriptions for all available commands, or a longer help text if a command is provided.
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -561,7 +561,7 @@ public class CoreCommands extends BaseComponentSystem {
             return "Value exceeds the maximum limit of 5000 blocks. your value: " + value + " blocks";
         }
 
-        for (int i = 0; i <= value; i++) {
+        for (int i = 0; i < value; i++) {
 
                 EntityRef blockItem = blockItemFactory.newInstance(block);
                 blockItem.send(new DropItemEvent(spawnPos));
@@ -591,9 +591,9 @@ public class CoreCommands extends BaseComponentSystem {
         BlockItemFactory blockItemFactory = new BlockItemFactory(entityManager);
         Vector3f startPos = new Vector3f(spawnPos);
 
-        float deltax = .2f; // delta x is the distance between the pins in the rows.
-        float deltaz = .3f; //delta z is the distance between the rows.
-        float vectorY = 1.0f; //the height of the drop (to be modified to keep the bowlingPin upright)
+        float deltax = 0.5f; // delta x is the distance between the pins in the rows.
+        float deltaz = 1.0f; //delta z is the distance between the rows.
+        float vectorY = 0.0f; //the height of the drop (to be modified to keep the bowlingPin upright)
         //rownumber loop is for selecting row
         for (int rownumber = 0; rownumber < 4; rownumber++) {
             startPos.add(deltax * (4 - rownumber), vectorY, deltaz); //Spawn starting position for Rownumber
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -586,7 +586,7 @@ public class CoreCommands extends BaseComponentSystem {
             // a is the distance betweeen the pins and coordinate difference in z vector
             //xyz are the normal coordinates
 
-            float y = 2.0f; float x = 2.0f; float z =  2.0f; float a = 0.3f; float b = 0.5f;
+            float y = 6.0f; float x = 2.0f; float z =  2.0f; float a = 0.3f; float b = 0.5f;
             // there is no guarantee that the pins will stand straight, but the range for ir to stand up right is from the vector y between (5.8 - 6.1)
             float x1 = 0; float x2 = 0; float x3 = 0; float x4 = 0; float x5 = 0; float x6 = 0; float x7 = 0; float x8 = 0; float x9 = 0; float x10 = 0; float z1 = 0; float z2 = 0; float z3 = 0; float z4 = 0; float z5 = 0; float z6 = 0; float z7 = 0; float z8 = 0; float z9 = 0; float z10 = 0;
 
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -16,6 +16,7 @@
 package org.terasology.logic.console.commands;
 
 import com.google.common.collect.Ordering;
+import com.sun.jna.platform.unix.X11;
 import org.terasology.assets.ResourceUrn;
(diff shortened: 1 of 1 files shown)


GooeyHub:

SurajDatta:
bowling pin* 😅

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -76,18 +78,16 @@
...
 import org.terasology.world.block.loader.BlockFamilyDefinition;
 
 import java.io.IOException;
 import java.net.UnknownHostException;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Locale;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
jellysnake: Please change your settings in IntelliJ so this doesn't happen. We require all imports to not use the `*`

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +536,139 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
         blockItem.send(new DropItemEvent(spawnPos));
         return "Spawned block.";
     }
+    @Command(shortDescription = "spawns many block together (value up to you prefer)", helpText = "you can select the particular block you want to spawn" +
```
jellysnake: I'd prefer this description to be re-worded to fit better with the other commands. Eg, correct grammar (If English isn't your first language feel free to ask someone to check it before you make the PR), similar third person type description as the other commands. Eg, no "you"

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +536,139 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+            Vector3f spawnPos = characterLocation.getWorldPosition();
+            Vector3f offset = characterLocation.getWorldDirection();
+            offset.scale(3);
+            spawnPos.add(5,10,0);
+
+            BlockFamily block = blockManager.getBlockFamily(blockName);
+            if (block == null) {
+                return "";
+            }
+            BlockItemFactory blockItemFactory = new BlockItemFactory(entityManager);
```
jellysnake: Even though the compiler will move some of it outside the for loop, I'd like to see you move it all outside. Only the last two lines should be in the loop.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +536,139 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+
+            ClientComponent clientComponent = sender.getComponent(ClientComponent.class);
+            LocationComponent characterLocation = clientComponent.character.getComponent(LocationComponent.class);
+
+            Vector3f spawnPos = characterLocation.getWorldPosition();
+            Vector3f offset = characterLocation.getWorldDirection();
+
+
+            offset.scale(0);
+            // a is the distance betweeen the pins and coordinate difference in z vector
```
jellysnake: Again, like for the other command please move what doesn't need to be in the loop to outside

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +536,139 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+            //xyz are the normal coordinates
+
+            float y = 6.0f; float x = 2.0f; float z =  2.0f; float a = 0.3f; float b = 0.5f;
+            // there is no guarantee that the pins will stand straight, but the range for ir to stand up right is from the vector y between (5.8 - 6.1)
+            float x1 = 0; float x2 = 0; float x3 = 0; float x4 = 0; float x5 = 0; float x6 = 0; float x7 = 0; float x8 = 0; float x9 = 0; float x10 = 0; float z1 = 0; float z2 = 0; float z3 = 0; float z4 = 0; float z5 = 0; float z6 = 0; float z7 = 0; float z8 = 0; float z9 = 0; float z10 = 0;
```
jellysnake: Please use informative names, in this particular case you probably should have used a a loop.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +536,139 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+            x1 = x; z1 = z;
+            x2 = x + a; z2 = z+b;
+            x3 = x-a; z3 = z2;
+            x4 = x1+a; z4 = z3+b;
+            x5 = x1; z5 = z4;
+            x6 = x3-a; z6 = z5;
+            x7 = x4+a; z7 = z6+b;
+            x8 = x2; z8 = z7;
+            x9 = x3; z9 = z8;
+            x10 = x6-a; z10 = z8;
```
jellysnake: See above, incredibly hard to read. I have literally no idea what this section does without spending a lot of time reading it.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +536,139 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+            }
+            if (m == 7) {
+                spawnPos.add(x8, y, z8);
+            }
+            if (m == 8) {
+                spawnPos.add(x9, y, z9);
+            }
+            if (m == 9) {
+                spawnPos.add(x10, y, z10);
+            }
```
jellysnake: Please use a switch-case here instead of chaining if statements.

However this is pointing to you using loops incorrectly. This whole section could be simplified drastically

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +536,139 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+            //spawnPos.add(-x, -2, -z);
+            try {
+                Thread.sleep(10);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        }
```
jellysnake: Rather than sleep twice, just sleep once for longer.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +536,139 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+
+
+        return "Dropped 10 Bowlingpin :)";
```
jellysnake: See previous comment on grammar

SurajDatta:
Hi, I have ade the following changes:
- Changed the import setting, so now there is no `*`
- Changes all the gramatical mistakes
- from the bulk dropper I have moved all the unneccesary code outside the loop, but in Bowling Pin I could not move the top 5 lines oustide. Because it is behaving wierdly; if I take those lines outside, the pins are not dropping in a regular manner. I agreewith you that this could be taken outside the loop but since I am facing the problem, I am unable to do so.
- I have changed the variables and given them a proper understandable name.
- I have added a switch case in place of multiple if statements.
- I have removed a sleep timer and have only one.

hope everything is fine now :)

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
jellysnake: These should still be outside the loop. You only need two lines inside the loop

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +541,126 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+                return "";
+            }
+            BlockItemFactory blockItemFactory = new BlockItemFactory(entityManager);
+            EntityRef blockItem = blockItemFactory.newInstance(block);
+
+            blockItem.send(new DropItemEvent(spawnPos));
+        }
+
+        // this returns the block you have spawned and the amount
+        return "Dropped " + value + " Blocks :)";
```
jellysnake: Please include the name of the block here

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +541,126 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+    }
+
+    @Command(shortDescription = "Drops bowling pin", helpText = " `prepare` command to be entered along with the `block name` ",
```
jellysnake: This description is incorrect, it doesn't drop a bowling pin. It arranges the blocks in a bowling pin format.
See the other commands for how the `helpText` section should be written as well.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +541,126 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+            }
+            BlockItemFactory blockItemFactory = new BlockItemFactory(entityManager);
+            EntityRef blockItem = blockItemFactory.newInstance(block);
+
+            blockItem.send(new DropItemEvent(spawnPos));
+            try {
+                Thread.sleep(100);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
```
jellysnake: Is it needed to sleep?

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -516,14 +541,126 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+
+        }
+
 
+        return "Dropped 10 Bowlingpin :)";
```
jellysnake: Please make this more informative, like in other commands.

SurajDatta:
Hi, I have made the following changes:

- Removed all the unused imports
- Rebuild the whole `prepare`
- Removed the sleep statements
- Changes the descriptions and the help text to be more informative
- Changed the return statements and added more information

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -506,24 +526,91 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
-            return "";
-        }
 
         BlockItemFactory blockItemFactory = new BlockItemFactory(entityManager);
         EntityRef blockItem = blockItemFactory.newInstance(block);
 
         blockItem.send(new DropItemEvent(spawnPos));
         return "Spawned block.";
     }
+    @Command(shortDescription = "Drops many blocks as per value choosen by the player", helpText = "the particular block can be seleted which the player wants to drop"
```
Cervator: Help text here is a bit weird and has a typo. Also doesn't touch the `value` param (doesn't need a full independent description or anything but could be something like "You can indicate which block to drop with the first parameter and how many with the second")
  
Could the `value` param be optional with a default value of maybe 100? Then the player could just use `bulkDrop dirt`. And on a similar note should `value` be capped? Have you tried to see what a reasonable upper bound would be? If 2000 blocks lags even a good system we might as well make maybe 10000 the max and return a message to the player if a higher number was requested.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -506,24 +526,91 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+                , runOnServer = true, requiredPermission = PermissionManager.CHEAT_PERMISSION)
+    public String bulkDropper(@Sender EntityRef sender, @CommandParam("blockName") String blockName, @CommandParam("value") int value) {
```
Cervator: Maybe a better name: `bulkDrop`

Seems more natural to `bulkDrop dirt 250` than `bulkDropper dirt 250` and it is slightly shorter

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -506,24 +526,91 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+        ClientComponent clientComponent = sender.getComponent(ClientComponent.class);
+        LocationComponent characterLocation = clientComponent.character.getComponent(LocationComponent.class);
+
+        Vector3f spawnPos = characterLocation.getWorldPosition();
+        Vector3f offset = characterLocation.getWorldDirection();
+        offset.scale(3);
+        spawnPos.add(5,10,0);
+        BlockFamily block = blockManager.getBlockFamily(blockName);
+        if (block == null) {
+            return "";
```
Cervator: In case of an error please return a useful message to the user, that'll show up in the console to indicate something went wrong, hopefully with some details :-)

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -506,24 +526,91 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+        }
+
+        // this returns the block you have spawned and the amount
+        return "Dropped " + value + " " + blockName + " Blocks :)";
+    }
 
+    @Command(shortDescription = "Arranges the blocks in a bowling pin format.", helpText = "Spawns a specific blocks in a regular bowling pin pattern, The item " +
+            " front of the player can simply picked up",
+            runOnServer = true, requiredPermission = PermissionManager.CHEAT_PERMISSION)
+    public String prepare  (@Sender EntityRef sender, @CommandParam("blockName") String blockName) {
```
Cervator: Requesting a command rename here: `prepare` is very generic and suggests nothing about what is being prepared. Maybe instead "bowlingPrep` or `bowlingPins` or something in that vein. Look for a command that's unique enough so auto-complete will have no trouble with ambiguous leading letters. 

Text could also be improved a little bit ("Arranges the blocks" is bad grammar - which blocks? Could as well be "Sets up a typical bowling pin arrangement in front of the player" or so)

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -506,24 +526,91 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+        LocationComponent characterLocation = clientComponent.character.getComponent(LocationComponent.class);
+
+        Vector3f spawnPos = characterLocation.getWorldPosition();
+        Vector3f offset = characterLocation.getWorldDirection();
+        offset.scale(5);
+        spawnPos.add(offset);
+
+        BlockFamily block = blockManager.getBlockFamily(blockName);
+        if (block == null) {
+            return "";
```
Cervator: Like the other command it would be nice to tell the user something helpful if there is an error :-)

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -506,24 +526,91 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+        }
+        BlockItemFactory blockItemFactory = new BlockItemFactory(entityManager);
+
+        Vector3f startPos = new Vector3f(spawnPos);
+
+        float deltax = .2f;
```
Cervator: A quick comment describing these three might be handy, maybe also making the `y` variable name more descriptive (isn't that actually triggering a Checkstyle warning?)

SurajDatta:
made*

SurajDatta:
Hi, I have made the following changes:
- changed the help texts and descriptions
- caped to 5000 blocks
- made the commands shorter 
- improved the return statements
-  I have also check with the checkStyle and everything looks fine :-)
- I have tried to implement the default command, which was setting `value` to 100 when just `bulkDrop stone` is called, but it seems it is very hard to do so.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -517,13 +537,87 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+        LocationComponent characterLocation = clientComponent.character.getComponent(LocationComponent.class);
+
+        Vector3f spawnPos = characterLocation.getWorldPosition();
+        Vector3f offset = characterLocation.getWorldDirection();
+
+        offset.scale(3);
+        spawnPos.add(5,10,0);
+        BlockFamily block = blockManager.getBlockFamily(blockName);
+        if (block == null)
+        {
```
Cervator: Are you sure you've got Checkstyle set up within IntelliJ? Braces like this on the next line is one of the very common violations it catches. Starting braces should be on the same line as the statement that begins the new block :)

Also, you can make the return more informative by including the actual block name in the message

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -517,13 +537,87 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+            return "Sorry, your block is not found";
+        }
+
+        BlockItemFactory blockItemFactory = new BlockItemFactory(entityManager);
+        if(value > 5000)
```
Cervator: More Checkstyle: Should have caught missing space between `if(`

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -517,13 +537,87 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+        {
+            return "sorry the value you have entered exceeds the limit of 5000";
```
Cervator: Easy to tell the user what the number was (rarely can help if a bug results in a different value than the user expected)

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -517,13 +537,87 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+        }
+
+        // this returns the block you have spawned and the amount
+        return "Dropped " + value + " " + blockName + " Blocks :)";
+    }
+
+    @Command(shortDescription = "Sets up a typical bowling pin arrangement in front of the player. ", helpText = "Spawns a specific blocks in a regular bowling pin pattern, The item " +
+            " front of the player can simply picked up",
+            runOnServer = true, requiredPermission = PermissionManager.CHEAT_PERMISSION)
+    public String bowlingPrep  (@Sender EntityRef sender, @CommandParam("blockName") String blockName) {
```
Cervator: You still have some mystery extra spaces or lines like here between `bowlingPrep  (@...`

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -517,13 +537,87 @@ public String spawnBlock(@Sender EntityRef sender, @CommandParam("blockName") St
...
+        BlockFamily block = blockManager.getBlockFamily(blockName);
+        if (block == null) {
+            return "Sorry, your block is not found";
+        }
+
+        BlockItemFactory blockItemFactory = new BlockItemFactory(entityManager);
+        Vector3f startPos = new Vector3f(spawnPos);
+
+        float deltax = .2f; // delta x is the distance between the pins in vectorX coordinate
+        float deltaz = .3f; //delta z is the distance between the pins in vectorZ coordinate
```
Cervator: For clarity it may help to indicate what dimension `z` is rather than just saying delta z is the dist between pins in vectorZ - that somewhat just rephrases things :)

SurajDatta:
Hi, I have made the following changes:
- corrected all the checkStyle issues
- rephrased some explanations
- add the value typed by the user
- Also in the return statement, I am not sure what it does hence I cant write any info on that until I understand when does that return execute

Cervator:
Finally had a chance to get back to this. Works pretty well, good work!

At the moment the bowling pins don't manage to stay standing, and the command seems to place stuff off to the side. But physics will change soon so maybe we can improve this some more later. I'd also be curious to be able to vary the spread on the bulk drop one day :-)

Thanks @SurajDatta !

I'm merging this via web editor with squash since the commits are a bit messy, then pushing another commit myself with slight tweaks to the help text :-)

PR #3217, opened 2018-01-10
Added some javadocs

Adrijaned:
### Contains

Adds or edits a bunch of javadocs for a GCI task.

### How to test
Since there are no code related changes, there is no testing required/possible.

### DIFF
--- a/engine/src/main/java/org/terasology/utilities/subscribables/AbstractGeneralSubscribable.java
+++ b/engine/src/main/java/org/terasology/utilities/subscribables/AbstractGeneralSubscribable.java
@@ -18,17 +18,14 @@ package org.terasology.utilities.subscribables;
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
 
-
 public class AbstractGeneralSubscribable implements GeneralSubscribable {
 
     protected transient PropertyChangeSupport propertyChangeSupport = new PropertyChangeSupport(this);
-
 
     @Override
     public void subscribe(PropertyChangeListener changeListener) {
         this.propertyChangeSupport.addPropertyChangeListener(changeListener);
     }
-
 
     @Override
     public void unsubscribe(PropertyChangeListener changeListener) {
--- a/engine/src/main/java/org/terasology/utilities/subscribables/AbstractGeneralSubscribable.java
+++ b/engine/src/main/java/org/terasology/utilities/subscribables/AbstractGeneralSubscribable.java
@@ -18,24 +18,18 @@ package org.terasology.utilities.subscribables;
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
 
-/**
- * {@inheritDoc}
- */
+
 public class AbstractGeneralSubscribable implements GeneralSubscribable {
 
     protected transient PropertyChangeSupport propertyChangeSupport = new PropertyChangeSupport(this);
 
-    /**
-     * {@inheritDoc}
-     */
+
     @Override
     public void subscribe(PropertyChangeListener changeListener) {
         this.propertyChangeSupport.addPropertyChangeListener(changeListener);
     }
 
-    /**
-     * {@inheritDoc}
-     */
+
     @Override
     public void unsubscribe(PropertyChangeListener changeListener) {
         this.propertyChangeSupport.removePropertyChangeListener(changeListener);
--- a/engine/src/main/java/org/terasology/world/chunks/Chunk.java
+++ b/engine/src/main/java/org/terasology/world/chunks/Chunk.java
@@ -19,10 +19,12 @@ package org.terasology.world.chunks;
 import org.terasology.module.sandbox.API;
 
 /**
- * Chunks are the smallest logical units of the worlds.
+ * Chunks are a box-shaped logical grouping of Terasology's blocks, for performance reasons. 
  *
- * Chunks are the default unit passed to world generator during world generation.
- * For details on dimensions and other parameters of chunk, check {@link ChunkConstants}.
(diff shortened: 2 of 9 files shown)


GooeyHub:

engine/src/main/java/org/terasology/utilities/subscribables/AbstractGeneralSubscribable.java
```diff
@@ -18,24 +18,18 @@
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
 
-/**
- * {@inheritDoc}
- */
+
```
vampcat: It's great that you decided to remove these redundant javadocs.

However, note that every method now has 2 newlines (in this case, 20 and 21) perceding it. We usually like 1 newline. Can you please remove the extra newline for all the javadocs you deleted?

engine/src/main/java/org/terasology/world/chunks/Chunk.java
```diff
@@ -19,6 +19,7 @@
 import org.terasology.module.sandbox.API;
 
 /**
+ * This interface bundles together all the methods chunk contains.
```
vampcat: The name of the interface is Chunk, so obviously it will bundle the methods of chunk - doesn't really help the reader. I would like to see a more concrete description of this (explain what a chunk is).

engine/src/main/java/org/terasology/world/chunks/CoreChunk.java
```diff
@@ -37,56 +37,211 @@
  */
 @API
 public interface CoreChunk {
+    /**
+     * @return Position of the chunk in blocks
```
vampcat: "position of the chunk in blocks"
This is not entirely clear. What do you mean? Do you mean it returns the local position of the chunk within a block, or blocks is the unit (like 3 blocks away from origin) ?

engine/src/main/java/org/terasology/world/generation/BaseFacetedWorldGenerator.java
```diff
@@ -20,6 +20,9 @@
 import org.terasology.world.generator.WorldConfigurator;
 import org.terasology.world.generator.WorldGenerator;
 
+/**
+ * The most common implementation of {@link WorldGenerator} based on the idea of Facets
```
vampcat: "The most common implementation"
The word common seems strange in this context. Common in what sense? Most commonly implemented (as in multiple people implement it, and most people implement this) or most commonly used implementation (implemented by one person, and used by most users) or what? Can you replace it with something else?
`A basic implementation` or something.

engine/src/main/java/org/terasology/world/generation/BaseFacetedWorldGenerator.java
```diff
@@ -83,6 +89,10 @@ public World getWorld() {
         return world;
     }
 
+    /**
+     * Returns current {@link WorldBuilder} or creates new if none exists
```
vampcat: The way your sentence is structured, it will
- Return existing one if it exists
- Or create a new one (and not return anything).
Rephrase it. The job of a javadocs writer is not just to convert `getWorldBuilder` to `Returns world builder`.

For reference, this is how I would personally write it:
`Returns the current WorldBuilder used by this WorldGenerator. If no WorldBuilder has been initialized so far, a new one is created and then returned.`
Kinda verbose, but much more clear in meaning.

vampcat: Also, don't leave the `@return` blank.
`@return The WorldBuilder used by this WorldGenerator` or something.
I realize there is some duplication, but it's inevitable.

engine/src/main/java/org/terasology/world/generator/WorldGenerator.java
```diff
@@ -24,24 +24,56 @@
...
 
+    /**
+     * @return Seed used for creating this world generator
+     */
     String getWorldSeed();
 
+    /**
+     * Sets the seed to use for creating of the world made by this world generator.
+     * <p>
+     * NOTE: this is a String value. The long value used in the most common implementation of this interface
```
vampcat: Instead of writing "in the most common implementation" link to the implementation that you're talking about.

engine/src/main/java/org/terasology/world/chunks/Chunk.java
```diff
@@ -19,6 +19,9 @@
 import org.terasology.module.sandbox.API;
 
 /**
+ * Chunks are the smallest logical units of the worlds.
+ *
+ * Each chunk is 32x32x64 in dimensions, and is the default unit passed to world generator during world generation.
```
pollend: and is the default unit passed to world generator during world generation. I'm not sure what this means exactly. I think you are talking about chunks as if they were a unit. The basic concept of a chunk is to allow the world to be partitioned into manageable blocks. maybe talk about a chunk in those terms rather than as a cubic unit.

oniatus: I think the size of a chunk is defined in `ChunkConstants`. Probably it is better to add a reference to ChunkConstants instead of duplicating the values in the documentation here.

emanuele3d:
@oniatus can you have a look at the latest changes and see if they are satisfactory? It'd be good to get this small PR out of the way.

oniatus:
Looks okay. We may update the javadoc once we have sectors because then chunks have an impact on the entity-system too, at least as far as I know :) For now this is okay.

emanuele3d:
Merged! Thanks @Adrijaned! Another PR bites the dust!

PR #3222, opened 2018-01-13
AABB improvements

eviltak:
## Goal
Adds documentation to the undocumented methods in the `AABB` class, and performs some refactoring.

## Documents
- `createMinMax`
- `createCenterExtent`
- `createEmpty`
- `createEncompassing` (vertices variant)
- `getCenter`
- `getMin`
- `getMax`
- `transform` (both variants)
- `getFirstHitPlane`
- `intersectRectangle`
- `minX`, `Y`, `Z`
- `maxX`, `Y`, `Z`

## Removes
- `AABB.getFirstHitPlane`

### DIFF
--- a/engine/src/main/java/org/terasology/math/AABB.java
+++ b/engine/src/main/java/org/terasology/math/AABB.java
@@ -108,7 +108,7 @@ public final class AABB {
     }
 
     /**
-     * Creates a new AABB that contain the vertices as represented by a {@link TFloatList}.
+     * Creates a new AABB that contains the vertices as represented by a {@link TFloatList}.
      *
      * @param vertices The vertices to encompass.
      * @return The created AABB.
--- a/engine/src/main/java/org/terasology/math/AABB.java
+++ b/engine/src/main/java/org/terasology/math/AABB.java
@@ -73,7 +73,7 @@ public final class AABB {
     }
 
     /**
-     * Creates an empty AABB that does not encompass any points.
+     * Creates an empty AABB that does not contain any points.
      *
      * @return The created AABB.
      */
@@ -108,7 +108,7 @@ public final class AABB {
     }
 
     /**
-     * Creates a new AABB that encapsulates the vertices as represented by a {@link TFloatList}.
+     * Creates a new AABB that contain the vertices as represented by a {@link TFloatList}.
      *
      * @param vertices The vertices to encompass.
      * @return The created AABB.
@@ -143,6 +143,8 @@ public final class AABB {
     }
 
     /**
+     * Returns the center of the AABB.
+     * 
      * @return The center of the AABB.
      */
     public Vector3f getCenter() {
@@ -153,6 +155,8 @@ public final class AABB {
     }
 
     /**
+     * Returns the minimum of the AABB.
+     * 
      * @return The minimum of the AABB.
      */
     public Vector3f getMin() {
@@ -160,6 +164,8 @@ public final class AABB {
     }
 
     /**
+     * Returns the maximum of the AABB.
+     * 
      * @return The maximum of the AABB.
      */
     public Vector3f getMax() {
@@ -185,7 +191,7 @@ public final class AABB {
      * @param rotation The rotation from the current AABB to the new AABB.
      * @param offset The offset between the current AABB and the new AABB.
      * @param scale The scale of the new AABB with respect to the old AABB.
-     * @return the new transformed AABB.
+     * @return The new transformed AABB.
      */
     public AABB transform(Quat4f rotation, Vector3f offset, float scale) {
         Transform transform = new Transform(new Matrix4f(VecMath.to(rotation), VecMath.to(offset), scale));
@@ -405,6 +411,8 @@ public final class AABB {
     }
 
     /**
+     * Returns the x-coordinate of the minimum of the AABB.
+     * 
      * @return The x-coordinate of the minimum of the AABB.
      */
     public float minX() {
@@ -412,6 +420,8 @@ public final class AABB {
     }
 
     /**
+     * Returns the y-coordinate of the minimum of the AABB.
+     * 
      * @return The y-coordinate of the minimum of the AABB.
      */
     public float minY() {
@@ -419,6 +429,8 @@ public final class AABB {
     }
 
     /**
+     * Returns the z-coordinate of the minimum of the AABB.
+     * 
      * @return The z-coordinate of the minimum of the AABB.
      */
     public float minZ() {
@@ -426,6 +438,8 @@ public final class AABB {
     }
 
     /**
+     * Returns the x-coordinate of the maximum of the AABB.
+     * 
      * @return The x-coordinate of the maximum of the AABB.
      */
     public float maxX() {
@@ -433,6 +447,8 @@ public final class AABB {
     }
 
     /**
+     * Returns the y-coordinate of the maximum of the AABB.
+     * 
      * @return The y-coordinate of the maximum of the AABB.
      */
     public float maxY() {
@@ -440,6 +456,8 @@ public final class AABB {
     }
 
     /**
+     * Returns the z-coordinate of the maximum of the AABB.
+     * 
      * @return The z-coordinate of the maximum of the AABB.
      */
     public float maxZ() {
--- a/engine/src/main/java/org/terasology/math/AABB.java
+++ b/engine/src/main/java/org/terasology/math/AABB.java
@@ -71,7 +71,7 @@ public final class AABB {
         max.add(extent);
         return new AABB(min, max);
     }
-    
+
     /**
      * Creates an empty AABB that does not encompass any points.
      *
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -58,6 +72,11 @@ public static AABB createCenterExtent(Vector3f center, Vector3f extent) {
         return new AABB(min, max);
     }
 
+    /**
+     * Creates an empty AABB that does not encompass any points.
```
mjuvekar7: ~encompass~ contain

mjuvekar7: Do this everywhere. Sounds much more natural

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -88,6 +107,12 @@ public static AABB createEncompassing(Iterable<AABB> others) {
         return new AABB(min, max);
     }
 
+    /**
+     * Creates a new AABB that encapsulates the vertices as represented by a {@link TFloatList}.
```
mjuvekar7: ~encapsulates~ contains

mjuvekar7: Also ~the~ and ~as~

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -117,17 +142,26 @@ public Vector3f getExtents() {
         return dimensions;
     }
 
+    /**
```
mjuvekar7: Description!

mjuvekar7: Even the most trivial methods must have a description

mjuvekar7: Same for all other get functions

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -145,11 +179,25 @@ public AABB move(Vector3f offset) {
         return new AABB(newMin, newMax);
     }
 
+    /**
+     * Transform this AABB into a new AABB with the given rotation, offset and scale.
+     *
+     * @param rotation The rotation from the current AABB to the new AABB.
+     * @param offset The offset between the current AABB and the new AABB.
+     * @param scale The scale of the new AABB with respect to the old AABB.
+     * @return the new transformed AABB.
```
mjuvekar7: Capital T

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -345,26 +404,44 @@ public Vector3f centerPointForNormal(Vector3f normal) {
         return new Vector3f();
     }
 
+    /**
```
mjuvekar7: Description. Same for all other min functions

GooeyHub:

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -88,6 +107,12 @@ public static AABB createEncompassing(Iterable<AABB> others) {
         return new AABB(min, max);
     }
 
+    /**
+     * Creates a new AABB that contain the vertices as represented by a {@link TFloatList}.
```
mjuvekar7: contain**s** (small grammatical mistake)

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -46,10 +46,24 @@ private AABB(Vector3f min, Vector3f max) {
         this.max = max;
     }
 
+    /**
+     * Creates a new AABB from the given minimum and maximum points.
+     *
```
oniatus: Can you add an example like if min or max are inclusive or exclusive?

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -46,10 +46,24 @@ private AABB(Vector3f min, Vector3f max) {
...
+     * @param max The maximum of the AABB.
+     * @return The created AABB.
+     */
     public static AABB createMinMax(Vector3f min, Vector3f max) {
         return new AABB(new Vector3f(min), new Vector3f(max));
     }
 
+    /**
+     * Creates a new AABB with the given center and extents.
+     *
```
oniatus: Can you provide an example? Like creating an AABB around 0,0,0 with extent like 1,2,3. What would be the result?. At least that is something I would like to know when reading the documentation.
The current sentence adds no more information than the method and parameter names which already exist.

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -88,6 +107,12 @@ public static AABB createEncompassing(Iterable<AABB> others) {
         return new AABB(min, max);
     }
 
+    /**
+     * Creates a new AABB that contains the vertices as represented by a {@link TFloatList}.
+     *
```
oniatus: Important information for a caller would be the format for vertices and what would be the result. An example would be nice. E.g. this method spans a bounding box which contains all vertices.

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -88,6 +107,12 @@ public static AABB createEncompassing(Iterable<AABB> others) {
...
+     * @param vertices The vertices to encompass.
```
oniatus: Vertices like x,y,z,x2,y2,z2, .... Describe the expected format here.

engine/src/main/java/org/terasology/math/AABB.java
oniatus: I would like to see javadoc where it adds more information which is not visible in the API, for example what elements are expected to be in a passed list, if a method can handle null, if a method has side-effects, all these kind of things.
Javadoc which repeats the method name like
```
/**
* Getter for foo.
* @return the foo.
**/
public Foo getFoo();
```
are just 4 lines of wasted space and 10 seconds wasted of developer time, they add zero new information.

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -117,17 +142,32 @@ public Vector3f getExtents() {
...
         dimensions.add(min);
         dimensions.scale(0.5f);
         return dimensions;
     }
 
+    /**
+     * Returns the minimum of the AABB.
+     * 
+     * @return The minimum of the AABB.
+     */
```
oniatus: I think we can remove these, there is 0 logic here to explain.

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -117,17 +142,32 @@ public Vector3f getExtents() {
...
     public Vector3f getMin() {
         return new Vector3f(min);
     }
 
+    /**
+     * Returns the maximum of the AABB.
+     * 
+     * @return The maximum of the AABB.
+     */
```
oniatus: Same for this one, get rid of it :)

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -224,6 +278,17 @@ public Vector3f closestPointOnAABBToPoint(Vector3f p) {
         return r;
     }
 
+    /**
+     * Sweeps an AABB on to the AABB in the given direction and returns the normal of the first plane of intersection.
+     *
```
oniatus: @flo and I double-checked this method, it includes some weird behavior, e.g. if all flags are set the normal will be overridden and only the value for Z will be returned.
Besides that there is no single use of the method in the engine and also none in any module.

Please remove the entire method ;)

engine/src/main/java/org/terasology/math/AABB.java
oniatus: We should keep either the description or the return statement (my favorite would be the return statement and an empty method description).
Redundancy is always a bad thing ;) keep it DRY.

Same for all further min.. and max.. methods.

engine/src/main/java/org/terasology/math/AABB.java
eviltak: @oniatus Excellent point. Should I remove the docs in all such cases?

engine/src/main/java/org/terasology/math/AABB.java
eviltak: Definitely have to agree. Will keep that in mind.

engine/src/main/java/org/terasology/math/AABB.java
eviltak: Min and max should be inclusive. However, I do not understand what you mean by adding an example for this. What kind of example?

Cervator:
Bump - @oniatus & @eviltak you might both have comments to respond to here :-)

GooeyHub:

GooeyHub:

jellysnake:
I've just had a read over this, apart from the comments left by oniatus, it looks good to me :)

If we can get those few things tweak I think we should merge this

mjuvekar7:
Bump @oniatus @eviltak. Maybe it's time to take care of this? :)

Cervator:
Merging this with a few added trivial tweaks - can usually always improve docs further, but this is a step forward and we need to chew through some PRs already :-)

PR #3228, opened 2018-01-16
Remove FirstPersonViewNode

eviltak:
As per the discussion on Slack, this PR goes through with the impending deletion of the unnecessary `FirstPersonViewNode` class and `RenderSystem.renderFirstPerson` method. However, this may cause some modules implementing `RenderSystem` to break - @Cervator will be able to help pinpoint the affected modules and contain the fallout.

## Changes
No additions, only deletions; a rare situation indeed!
- Deletes `FirstPersonViewNode`
- Deletes `RenderSystem.renderFirstPerson` and all implementations

cc @vampcat @emanuele3d


GooeyHub:

emanuele3d:
@eviltak and @vampcat: 

I checked the code and tested it: didn't smoke nor explode - good sign. 

HOWEVER... du-du-du-dummmm....

It breaks the API. The package info file in the same folder as RenderSystem has the API annotation in it which means everything in the folder is part of the API. 

Which means we can't break it. YET. 

BUT... alpha 09 will be released shortly. After that everything that gets into develop will be released as alpha 10 which is actually v2.0.0. And as that's a major version change, we can break the API.

So, I'm going to approve the PR, but we'll keep it in stand-by until alpha 09 is out of the way.

Cervator:
And the API breakage festival has started woo!

Merged this and checked every module in the lineup: found two had dummy implementations of the method so those are now also being removed via:

* https://github.com/Terasology/Pathfinding/pull/35
* https://github.com/Terasology/IRLCorp/pull/17

Piece of cake just can't push those until the engine build is done :-)

Thanks!

PR #3229, opened 2018-01-17
Fixes NeighbourBlockFamilyUpdatesSystem not receiving OnChangedBlock events

Steampunkery:
## Contains
NeighbourBlockFamilyUpdateSystem did not correctly register with the world provider to be a `WorldListsener` and as such did not receive notifications from the `WorldProviderImpl.java#setBlock` and `WorldProviderImpl.java#setBlocks` methods.

### How to test
This issue was discovered when testing StructureTemplates. If you place a structure template that has fences without this PR in a orientation other than that it was constructed in, the fences will be oriented wrong. With this PR, the fences will automatically orient themselves correctly based on the rules in `FencesFamily.java`.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/neighbourUpdate/NeighbourBlockFamilyUpdateSystem.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/neighbourUpdate/NeighbourBlockFamilyUpdateSystem.java
@@ -15,7 +15,8 @@
  */
 package org.terasology.world.block.entity.neighbourUpdate;
 
-import com.google.common.collect.Sets;
+import java.util.Set;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.entitySystem.entity.EntityRef;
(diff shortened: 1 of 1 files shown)


GooeyHub:

flo:
Thanks for your efforts to make the game better.

Your PR contains also a lot of whitespace changes (mainly 4 spaces to tab conversion) which is bad for multiple reasons. One being that it makes it hard to see what has actually changed.

About the actual change: Can you describe in more detail what you are trying to fix?

It is also a sensitive area: Any further code that does additional stuff on block placement might result in performance issues when big structures get placed.

Steampunkery:
@flo When I’m next on my desktop, I’ll change the white space back to spaces. 

About the change: NeighborBlockFamilyUpdateSystem was not properly receiving OnChangedBlock events. The fix just adds NeighborBlockFamilyUpdateSystem as a WorldListener so on every setBlock call, it gets told that something was changed. Before this fix, things were only updated OnBlockPlaced which is not good enough, because ST uses setBlock. This resulted in fences being placed with the wrong orientation.

Cervator:
@gooeyhub: add to whitelist

Pinging @pollend :-)

GooeyHub:

flo:
When a structure template gets placed, it's blocks get rotated to match the rotation in which you placed the structure template. Maybe that is just not working for your new block family and we have to find a solution for that.

What I might see as issue is that structure templates place huge amount of blocks in some kind of any order. e.g. it might place a torch before the wall it is attached to. If the "torch" would have been smarter at placement it could have determined that it was an invalid placement and prevented it or something. That is why I think blocks set with setBlocks should not be smart.

Also when you update a neighboring blocks on block setting then it causes a lot of uncessary block changes coming from the placement not being atomic. e.g. fence gets placed when there are no neighboring blocks, second fence block gets placed which changes previous block, ground below fence gets placed, fence changes again etc.

Also a block change theoretically potentially recursivly trickering a block change does not seem right to me.

flo:
Oh and in the structure template branch animated-structure-placements (that is not yet merged) you can find in the class BlockRegionTransform a method transformBlock with this code:

```
    public Block transformBlock(Block block) {
        BlockFamily blockFamily = block.getBlockFamily();
        if (blockFamily instanceof SideDefinedBlockFamily) {
            SideDefinedBlockFamily sideDefinedBlockFamily = (SideDefinedBlockFamily) blockFamily;
            return sideDefinedBlockFamily.getBlockForSide(transformSide(block.getDirection()));
        } else if (blockFamily instanceof AttachedToSurfaceFamily) {
            // TODO add some proper method to block famility to not have to do this hack
            return blockFamily.getBlockForPlacement(null, null, null, transformSide(block.getDirection()), null);
        }
        return block;
    }
```

So if block families would have a general method for rotating them then this method could used by transformBlock to rotate also your fence block.

Steampunkery:
Well when a structure template is about to be placed, you could send a StartLargeBlockUpdate event then after send an EndLargeBlockUpdate event so prevent lots of unnecessary corrections.

flo:
I don't think we have those events already. I am also not sure if the concept is good that one events kind of sets a temporary mode till another event comes. When there is no second event then we would get stuck in an infinite large block placement.

What do you think about the idea with a new BlockFamily method?

Steampunkery:
It would definitely be useful, and probably the most elegant solution. Feel free to close this if that's all :)

flo:
Ok, closing it as suggested. Thanks for investigating in how to solve it.

PR #3230, opened 2018-01-17
Particle System Gen./Aff. Registration & refactor

MaxBorsch:
Replaces the old hard-coded generator and affector function registration with automated annotation based registration (also moves the registration logic to ParticleUpdater as that's where it's used).

**Usage:**
Mark `GeneratorFunction` / `AffectorFunction` classes with:
`@RegisterParticleSystemFunction(componentClass = AssociatedAffectorComponent.class)`
and they will be auto-registered when a game starts. This allows developers to add their own generators / affectors in their modules.

Ex.)
```java
@RegisterParticleSystemFunction(componentClass = ColorRangeGeneratorComponent.class)
public final class ColorRangeGeneratorFunction extends GeneratorFunction<ColorRangeGeneratorComponent> {
```


GooeyHub:

.gitignore
```diff
@@ -92,6 +92,6 @@ nuiEditorAutosave.json
 # Ignore stuff we extract for use with Terasology
 /natives/
 config/metrics/
-
+configs/engine/rendering.cfg
```
eviltak: Must have sneaked in through a commit 😄 We do need this, but I am not sure if this PR is the right place for the change.

MaxBorsch: @eviltak Yeah I noticed that. It came with my earlier particle javadocs commit that this was based on, a mentor requested that I add that in from an earlier task I think.

eviltak: This change has been incorporated in #3227 by eviltak@f393d2f, so it might be better to remove this change here to prevent merge conflicts.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/particles/functions/ParticleSystemFunction.java
```diff
@@ -22,30 +22,9 @@
...
 public abstract class ParticleSystemFunction<T> {
     private final int rawDataMask;
-    private final Class<T> component;
 
-    public ParticleSystemFunction(Class<T> component, ParticleDataMask dataMask, ParticleDataMask... dataMasks) {
+    public ParticleSystemFunction(ParticleDataMask dataMask, ParticleDataMask... dataMasks) {
         this.rawDataMask = ParticleDataMask.toInt(dataMask, dataMasks);
-        this.component = component;
-    }
-
```
oniatus: I would double-check if it is save to remove hashCode and Equals completely or if we can stick to something like the rawDataMask field. This may affect how these functions are stored in Lists and Maps.

engine/src/main/java/org/terasology/particles/updating/ParticleUpdaterImpl.java
```diff
@@ -58,12 +79,13 @@
 
     //== public ========================================================================================================
```
oniatus: if we touch the class anyway, can we git rid of this line of nonsense? :)

engine/src/main/java/org/terasology/particles/updating/ParticleUpdaterImpl.java
```diff
@@ -238,7 +318,7 @@ private void updateParticleSystem(final ParticleEmitterComponent partSys, final
             partSys.collisionUpdateIteration = (partSys.collisionUpdateIteration + 1) % PHYSICS_SKIP_NR;
         }
 
-        // System ran out of lifetime -> stop emission -> dispose
+        // System ran out of lifetime -> stop emission -> deregister
```
oniatus: why using a comment if you can just write `if(systemRanOutOfLifetime(...){deregister/unregister(...)}`?

MaxBorsch:
@oniatus Thanks for the review, #3230 includes these changes and is based on them, so further reviews could go there, I’ll fix whatever was mentioned here in that PR.

MaxBorsch:
@oniatus That last comment was meant for the other particles PR, my bad.

GooeyHub:

engine/src/main/java/org/terasology/particles/functions/ParticleSystemFunction.java
MaxBorsch: I added those implementations myself and they are no longer needed with the new system, and not possible because I removed the component field. I think we want default behavior for them for now.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/particles/functions/RegisterParticleSystemFunction.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * This annotation is used to mark a {@link ParticleSystemFunction} to be registered in by the {@link org.terasology.particles.updating.ParticleUpdater}.
+ */
+@API
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RegisterParticleSystemFunction {
+    Class<? extends Component> componentClass();
```
eviltak: We should be able to retrieve the component type from the generic parameter of `ParticleSystemFunction` by using `ReflectionUtil.getTypeParameterFromSuper`, making this annotation property redundant and making registration less verbose.

engine/src/main/java/org/terasology/particles/updating/ParticleUpdaterImpl.java
```diff
@@ -107,9 +125,37 @@ public void update(final float delta) {
...
+                    ParticleSystemFunction function = (ParticleSystemFunction) type.newInstance();
+
+                    if (function instanceof GeneratorFunction) {
+                        mapGeneratorFunction((GeneratorFunction) function, annotation.componentClass());
+
+                    } else if (function instanceof AffectorFunction) {
+                        mapAffectorFunction((AffectorFunction) function, annotation.componentClass());
+                    }
+                } catch (InstantiationException | IllegalAccessException e) {
+                    e.printStackTrace();
```
eviltak: Logging a warning here would be more appropriate.

pollend:
PR #3918 implements the changes in this PR.

PR #3231, opened 2018-01-17
fixed enum list handler

pollend:
I added a test case for a list of enums and also resolved an issue where a component that uses a list of enums deserializes to an empty list. I was trying to use a list of Sides with Rails and the set/list always deserializes to an empty list.


GooeyHub:

GooeyHub:

PR #3233, opened 2018-01-19
Add friction and restitution to block definitions

Josharias:
### Contains

Fixes thrown blocks that slide and never come to rest

### How to test

Throw a block with horizontal velocity on to the ground

### Known Issues
- Restitution doesnt quite seem to act the way I would have expected.  There is something funky going on with rotational energy that I cant quite put a finger on.  You can see with a rubber pickup item bouncing on a rubber block that it seems to gain energy and then never get to a rest state.
- Friction is still highly untested in how modifying the value affects the result.  The main win here is that it does apply some friction so that objects dont slide for forever.


GooeyHub:

modules/Core/assets/blocks/Ice.block
```diff
@@ -4,5 +4,6 @@
 {
     "ice": true,
     "translucent": true,
-    "hardness": 1
+    "hardness": 1,
+    "friction": 0.00001
 }
```
emanuele3d: Whoops. Inherited issue: missing newline.

engine/src/main/java/org/terasology/physics/components/RigidBodyComponent.java
```diff
@@ -48,6 +48,9 @@
     @Replicate
     public float friction = 0.5f;
 
+    @Replicate
+    public float restitution = 0f;
```
Cervator: Wish list: simple javadoc on the two fields that spell out the difference :D

Friction is pretty easy, other one a bit less purely on the name (figuring bouncy)

Josharias: I also had to search the internets to understand the difference.  Ill find a good explanation.

modules/Core/assets/blocks/soil/Dirt.block
```diff
@@ -2,5 +2,6 @@
  * Dirt is pretty basic
  */
 {
-    "basedOn": "core:soil"
+    "basedOn": "core:soil",
+    "friction": 100
```
Josharias: Oops,  this is testing data that slipped through

GooeyHub:

Cervator:
Thanks! I played with this locally for a bit. Realized rubber didn't have a unique texture so I gave it one real quick roughly approximated to natural rubber (latex) but with a bit of impurities and a hint of yellow/orange/brown. Hopefully it looks decent! :D

I also noticed rubber alone doesn't throw very far as it was set to the same mass as stone, so I dropped it down to where you can throw it around more. Checked on ice and such too. Very nice now!

Interesting how much more a difference it makes to be dropping stuff on rubber blocks (try rubber dropped on rubber) rather than dropping rubber on other stuff.

PR #3234, opened 2018-01-19
adds check for local source module to module downloader

skaldarnar:
A local source module will take precedence over a remote module with the same (snaphot) version. 
If the remote module has a higher version number than local source module it will still be downloaded.

> I tested it, it's safe!

_famous last words_


GooeyHub:

PR #3236, opened 2018-01-19
Support for optional dependencies

jellysnake:
Provides functionality for having code/events run only if specific modules are enabled. This allows for modules to implement content that will only be activated and run when some optional dependencies are included.

Making a PR now to get some eyes on the code I have so far, I want to implement a method for having optional assets so this shouldn't be merged until that's sorted.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ModuleEnvironmentInfo.java
+++ b/engine/src/main/java/org/terasology/engine/module/ModuleEnvironmentInfo.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.engine.module;
+
+import org.terasology.module.sandbox.API;
+import org.terasology.naming.Name;
+
+@API
+public interface ModuleEnvironmentInfo {
+
+    boolean moduleLoaded(String name);
+    boolean moduleLoaded(Name name);
+}
--- a/engine/src/main/java/org/terasology/engine/module/ModuleEnvironmentInfoImpl.java
+++ b/engine/src/main/java/org/terasology/engine/module/ModuleEnvironmentInfoImpl.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.engine.module;
+
+import org.terasology.entitySystem.systems.BaseComponentSystem;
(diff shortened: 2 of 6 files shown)


GooeyHub:

oniatus:
I think the `RegisterSystem` annotation contains some logic/documentation about systems being registered depending if modules are present in the game. Is this change aiming in the same direction or does it something different?

jellysnake:
Okay, I've added support for having optional prefabs.

@oniatus this is intended to supplement that feature.
This includes more features that will help developers better control how they implement a module optionally depending on others.

The three main features in this pull are as follows:
 - The class `ModuleEnvironmentInfo` which can be injected.
    - This has the method `moduleLoaded` which checks if the given module is currently loaded
- The ability to specify a list of modules in the `@RegisterEvent` annotation.
    - The event handler will only be registered if the modules in the list are enabled.
- The `ModuleDependencies` Component for prefabs, which contains a list of modules
    - The prefab will only be loaded if all the modules in the list are enabled.

GooeyHub:

engine/src/main/java/org/terasology/entitySystem/prefab/internal/PrefabFormat.java
```diff
@@ -38,10 +41,12 @@
 
     private ComponentLibrary componentLibrary;
     private TypeSerializationLibrary typeSerializationLibrary;
+    private ModuleEnvironment environment;
 
-    public PrefabFormat(ComponentLibrary componentLibrary, TypeSerializationLibrary typeSerializationLibrary) {
+    public PrefabFormat(ComponentLibrary componentLibrary, TypeSerializationLibrary typeSerializationLibrary, ModuleEnvironment envrionment) {
```
mkienenb: Fix the envrionment misspelling.

mkienenb:
This seems reasonable to me, and could be committed.

I do wonder if this would make more sense if it worked by registered component systems instead of named modules.

If the module is looking for a FooBarSystem, you don't really care what module it is found in, only that one is registered.   That might be a future project down the road worth investigating.

Cervator:
[Optional module dependencies](https://trello.com/c/P43Zi0WB/64-optional-module-dependencies)

jellysnake:
A few conflicts are caused by both this PR and Replay&Record adding new things to a constructor. It is easily fixed by manually merging the two small changes.

Others are just things like removing redundant imports and strange clashes

GooeyHub:

engine/src/main/java/org/terasology/engine/module/ModuleEnvironmentInfoImpl.java
```diff
@@ -0,0 +1,42 @@
...
+import org.terasology.entitySystem.systems.BaseComponentSystem;
+import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.module.sandbox.API;
+import org.terasology.naming.Name;
+import org.terasology.registry.In;
+import org.terasology.registry.Share;
+
+
+@Share(ModuleEnvironmentInfo.class)
+@API
```
eviltak: Should we mark this non-API and keep it to the engine itself? I'd say there's no need for modules to access the class itself since they already have access to the interface.

engine-tests/src/test/java/org/terasology/entitySystem/PrefabTest.java
```diff
@@ -75,7 +75,7 @@ public void setup() throws Exception {
                 (AssetFactory<Prefab, PrefabData>) PojoPrefab::new, "prefabs");
         ComponentLibrary componentLibrary = context.get(ComponentLibrary.class);
         TypeSerializationLibrary typeSerializationLibrary = context.get(TypeSerializationLibrary.class);
-        PrefabFormat prefabFormat = new PrefabFormat(componentLibrary, typeSerializationLibrary);
+        PrefabFormat prefabFormat = new PrefabFormat(componentLibrary, typeSerializationLibrary, null);
```
eviltak: Can/should we use `moduleManager.getEnvironment()` here for completeness?

engine/src/main/java/org/terasology/engine/module/ModuleEnvironmentInfo.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.engine.module;
+
+import org.terasology.module.sandbox.API;
+import org.terasology.naming.Name;
+
+@API
+public interface ModuleEnvironmentInfo {
```
eviltak: Documentation please :)

jdrueckert:
[Dependency Injection](https://trello.com/c/nqIadpY8/71-dependency-injection)

PR #3240, opened 2018-01-22
Changed REGEX format of FLOAT_PATTERN for boundary of mesh

kartikey0303:
The earlier REGEX format did not recognize longer numbers as the boundary for the mesh. It caused problems with the MawGooey model. The new pattern has been tested with all the models that have animation and is working in all cases.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/md5/MD5AnimationLoader.java
+++ b/engine/src/main/java/org/terasology/rendering/md5/MD5AnimationLoader.java
@@ -47,7 +47,8 @@ import java.util.stream.Collectors;
 public class MD5AnimationLoader extends AbstractAssetFileFormat<MeshAnimationData> {
 
     private static final String INTEGER_PATTERN = "((?:[\\+-]?\\d+)(?:[eE][\\+-]?\\d+)?)";
-    private static final String FLOAT_PATTERN = "((?:[\\+-]?\\d(?:\\.\\d*)?|\\.\\d+)(?:[eE][\\+-]?(?:\\d(?:\\.\\d*)?|\\.\\d+))?)";
+    //private static final String FLOAT_PATTERN = "((?:[\\+-]?\\d(?:\\.\\d*)?|\\.\\d+)(?:[eE][\\+-]?(?:\\d(?:\\.\\d*)?|\\.\\d+))?)";
+    private static final String FLOAT_PATTERN = "([-+]?[0-9]*\\.?[0-9]+)";
     private static final String VECTOR3_PATTERN = "\\(\\s*" + FLOAT_PATTERN + "\\s+" + FLOAT_PATTERN + "\\s+" + FLOAT_PATTERN + "\\s+\\)";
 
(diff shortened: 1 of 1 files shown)


GooeyHub:

flo:
Good job in finding out the cause for the model not loading!

Your change is a good proof of concept that the float parsing needs to be fixed. However the new pattern is much simpler than the previous one and does not seem to cover some cases like 2E10 of which I am not sure if we need them or not.

I think it would be good if you could write unit tests for the parsing of this regular expression. It should be pretty easy and quick to do as you can just check a series of patterns if they match properly.

The unit tests should contain the case you want to try to fix and the normal cases. Further cases are welcome but I think we should have at least those to say we are one step further.

J-Dill:
@flo @kartikey0303 I am willing to write some unit tests for this. What sort of cases would you like tested? It'd be nice to have some examples for testing, as I am not very familiar with what was causing problems before.

kartikey0303:
Hello @J-Dill. So earlier as I mentioned that any number with more than one digit causes a problem. That is because the current REGEX format `((?:[\\+-]?\\d(?:\\.\\d*)?|\\.\\d+)(?:[eE][\\+-]?(?:\\d(?:\\.\\d*)?|\\.\\d+))?)`  is according to the format that actually decimal numbers should be represented. [More here](https://mathsdoctor.co.uk/revision-help/gcse/number/standard-form/). But if we change the REGEX to `((?:[\+-]?\d*(?:\.\d*)?|\.\d+)(?:[eE][\+-]?(?:\d*(?:\.\d*)?|\.\d+))?)` it accepts the two digit numbers given to us by blender which is not following the standard. As @Cervator mentioned, test would involve writing something like [this](https://github.com/MovingBlocks/DestinationSol/pull/248/commits/72afb5cf30314c0b67f1cccc587b01493cc388d2#diff-f8611a35b156ddb4dac780e044e368e4). So the test would involve cases like 10.1, 10.0, 99E1E, 99.0E1.0, 99.0011E1.22 and more. Although blender would not give us numbers in exponential format but it good to cover all the cases, just in "case"..

flo:
Also the same issue exist for MD5 models. So ideally the pattern could be moved  as constant in a separate class used by both MD5 and MD5 animation loader.

J-Dill:
Are you talking about the patterns in MD5SkeletonLoader and MD5AnimationLoader? I'm guessing so, since the patterns in the two are exactly the same. I could throw the patterns from both classes into one class that they can just pull from. I'll then write the unit tests based on that combined class (I'll call it something like MD5Patterns)

flo:
@J-Dill yes, MD5Patterns sounds good.

Cervator:
Have tested this and #3292 out some. The regex there may need a bit more tweaking so I'm merging this to begin with to go with https://github.com/Terasology/MawGooey for more testing :-)

However, something went funny with the origin of this PR (branch deleted?) so I actually have to recreate the commit and just link it to here the best I can. So this won't auto-close as merged but *will* actually have been.

PR #3242, opened 2018-01-23
Replace visual character and allow for replacement by modules

flo:
### Contains

This PR:
* This PR introduces a new event (CreateVisualCharacterEvent) that modules can handle to create their own character visualization 
  * The visualization is completely up to the module. It could be a static model, a animated one or theoreticlly just a particle system.
  * The model can but does not need to be colored like the player. It is up to the event handler
  * For the own character no CreateVisualCharacterEvent gets fired so that "hideFromOwner" flag is needed for the SkeletalMeshComponent 
  * If the event does not get handled, a floating cube will be shown as character model (instead of the moneky head)
* The existing monkey head has been removed as it has been replaced by the floating cube.

The reason for replacing the default monkey head with a default floating cube is that the monkey head is not so clearly a place holder and people would just wonder why there is a monkey head. Also it is better to show of an animated model instead of a static one.

### How to test
Unit tests:
* Run VisualCharacterSystemTest

Manual testing:
1. Start the game as listen server.
2. Start another instance of the game and connect to server
3. Expected result: The own character is invisible in each instance while the other is a cube floating slightly and down constantly.


GooeyHub:

engine/src/main/java/org/terasology/logic/characters/VisualCharacterComponent.java
```diff
@@ -0,0 +1,32 @@
...
+package org.terasology.logic.characters;
+
+import org.terasology.entitySystem.Component;
+import org.terasology.logic.characters.events.CreateVisualCharacterEvent;
+import org.terasology.network.Replicate;
+
+/**
+ * Add this component to characters to give them a visual appearance.
+ *
+ * This component does not specifcy however how the character look like. Instead
```
Josharias: specifcy == "specify"

engine/src/main/java/org/terasology/logic/characters/VisualCharacterSystem.java
```diff
@@ -0,0 +1,128 @@
...
+            /*
+             * Before character has spawned localPlayer is not properly initialized
+             * and can not be used to test if character is visible.
+             */
+            return;
+        }
+        sendCreateVisualCharacterEventIfNotOwnCharacter(entity);
+    }
+
+    void sendCreateVisualCharacterEventIfNotOwnCharacter(EntityRef entity) {
```
Josharias: I wonder,  if in some future land we have a 3rd person camera,  we may likely want to have this event still sent. Maybe with a flag that it is for the local player,  and then let the mesh adding system do the logic whether or not it wants to add it for the local player's character?

Alternatively,  cancel the event if the player is in first person mode?  (not sure if this would be a good idea)

engine/src/main/java/org/terasology/logic/characters/VisualCharacterSystem.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+        entity.send(new CreateVisualCharacterEvent());
+    }
+
+    /**
+     * Handles the local character spawn  event by doing the work that had to be delayed till then:
+     * The CreateVisualCharacterEvent events that could not be sent previously will be sent.
+     */
+    @ReceiveEvent
+    public void onAwaitedLocalCharacterSpawnEvent(AwaitedLocalCharacterSpawnEvent event, EntityRef entity) {
```
Josharias: I am curious,  what happens if you add meshes and such before spawn?

Josharias:
Also, changing color didnt seem to affect the mesh color on rejoining a game. My chat color would change,  but not the mesh color.

emanuele3d:
I just wanted to say I did see this PR and plan to review it! Bear with me!

Cervator:
Bump, particularly for @flo :-) Would like to see this finishing up soon so we can start looking at additional characters.

flo:
Ah didn't test it with 3 instances, just with 2. I added the cleanup of the visual character entity.

Also I added some more explanation.

An open issue is still the players color: Rejoining players are not properly colored as the character gets loaded (and thus the VisualComponent activated) before the character got linked to the player. So once you rejoin a game your character is white. I will have to think about how to address it.

flo:
@Josharias  and thanks for the review btw

GooeyHub:

Cervator:
Thanks for the updates :-)

Build problem seems minor, a single unit test broke: http://jenkins.terasology.org/job/TerasologyPRs/2094/testReport/junit/org.terasology.logic.characters/VisualCharacterSystemTest/testSendingOfCreateVisualCharacterEvent/

flo:
Module for testing it out: https://github.com/Terasology/HumanoidCharacters

flo:
When testing it out with character module please not that the character is wrongly turned.

GooeyHub:

engine/src/main/resources/assets/materials/floatingCubeSkin.mat
```diff
@@ -0,0 +1,12 @@
...
+  "params": {
+    "diffuse": "white",
+    "colorOffset": [
+      1.0,
+      1.0,
+      1.0
+    ],
+    "textured": true
+  }
+}
```
emanuele3d: Something funny with the EOF here. Add a new line perhaps?

flo:
@emanuele3d thanks for review, I worked it in. I also changed the map to a field in the component to allow systems to get from character to visual character.

GooeyHub:

GooeyHub:

GooeyHub:

engine-tests/src/test/java/org/terasology/logic/characters/VisualCharacterSystemTest.java
```diff
@@ -0,0 +1,183 @@
...
+        List<Event> otherCharacterEntityEvents = new ArrayList<>();
+        recordEntityEventsToList(otherCharacterEntity, otherCharacterEntityEvents);
+
+        clientEntityReturnedByLocalPlayer = EntityRef.NULL;
+
+        /*
+         * Simulate activation before entity is done
+         * since the character is not properly linked yet nothing should happen
+         */
+        system.onActivatedVisualCharacter(OnActivatedComponent.newInstance(), otherCharacterEntity);
```
Cervator: Syntax error here and two further instance of the same situation - no `VisualCharacterComponent` is supplied to the method despite being required. I suspect the target method was changed after the test was written? Easy to miss since `engine-tests` won't force recompile when just trying to re-run the game for play testing.

This is what is marking the Jenkins build as broken. I tried just throwing in a `new VisualCharacterComponent()` in the three spots but then the test hit an similar sounding NPE so it probably needs a quick bit of tweaking

engine-tests/src/test/java/org/terasology/logic/characters/VisualCharacterSystemTest.java
```diff
@@ -0,0 +1,183 @@
...
+        });
+    }
+
+    /**
+     * This test verifies that:
+     * <ul>
+     *     <li>No visual character gets created (via event) for the own character (as it is first person)</li>
+     *     <li>That the system can deal with LocalPlayer and characters not being properly linked when
+     *     the character entities get loaded/created</li>
+     *     <li>A visual character gets created (via event) for charcters that were already present when the player
```
Cervator: Typo: "charcters"

engine-tests/src/test/java/org/terasology/logic/characters/VisualCharacterSystemTest.java
```diff
@@ -0,0 +1,183 @@
+/*
+ * Copyright 2017 MovingBlocks
```
Cervator: 2018 :D

engine-tests/src/test/java/org/terasology/logic/characters/VisualCharacterSystemTest.java
```diff
@@ -0,0 +1,183 @@
...
+        recordEntityEventsToList(laterJoiningCharacterEntity, laterJoiningCharacterEntityEvents);
+
+        // Joined player is not properly linked but it should not matter:
+        Mockito.when(laterJoiningCharacterEntity.getOwner()).thenReturn(EntityRef.NULL);
+
+        system.onActivatedVisualCharacter(OnActivatedComponent.newInstance(), laterJoiningCharacterEntity);
+
+
+        /*
+         * There is no second AwaitedLocalCharacterSpawnEvent event,
```
Cervator: This is probably short enough to just be a regular single line comment?

engine/src/main/java/org/terasology/logic/characters/VisualCharacterComponent.java
```diff
@@ -0,0 +1,41 @@
...
+import org.terasology.entitySystem.Component;
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.logic.characters.events.CreateVisualCharacterEvent;
+import org.terasology.network.NoReplicate;
+import org.terasology.network.Replicate;
+
+/**
+ * Add this component to characters to give them a visual appearance.
+ *
+ * This component does not specify however how the character look like.
```
Cervator: Lines could probably be a bit longer + excess blank lines

GooeyHub:

GooeyHub:

J-Dill:
I was able to fix the Null Pointer Exception in the test in my commit that I referenced above. I am not sure on how to get the changes to you, however, @flo. How do I go about doing that?

flo:
@J-Dill oh, that sounds like the changes I made. Sorry that this turned out in redundant work but I didn't expect someone to do a PR for a small PR like this.

J-Dill:
Ahh yeah, no problem! For future reference though, is that what I am supposed to do when suggesting a fix: create my own fix and then reference this PR when I merge into my own? Did that work alright? @flo

flo:
I would suggest to just ping the owner of the PR first and ask if you can help out. That way double work is avoided. Otherwise you can create a PR on my repository for merging your branch into mine.

J-Dill:
I see, thanks!

PR #3246, opened 2018-02-02
Redesigned Groovyw Setup

digitalripperynr:
This system makes the current groovyw script system much more modular and improves the reusability of functions.
All the functions are stored in a 'functions.groovy' script in the 'configs/groovy/' directory and the base scripts are located in the root directory. 

There is only one problem with this approach. The base scripts are entirely same except for the initialization code in the start.
I wonder if we could make a 'commands.groovy' script in the configs folder and the base scripts would just initialize the type of command and pass on the arguments to the commands script. 
That would make the system even more modular.

Suggestions Please !!!


GooeyHub:

PR #3219, opened 2018-01-11
Fixes #3184

anuar2k:
### Contains
A fix for #3184 (also a pr for Dialogs module is coming).

Brings `getInputsForBindButton(SimpleUri bindId)` back, probably missed somewhere during a merge or something.

Issue #3184, opened 2017-12-17
API violation detected: inputSystem.getInputsForBindButton


Cervator:
While rebuilding _all_ the modules for Alpha 9 final testing I found that the [Dialogs](https://github.com/Terasology/Dialogs) module failed to build, with `InputSystem.getInputsForBindButton` missing. Bit of research pointed to it having been removed in https://github.com/MovingBlocks/Terasology/commit/b91b832 which was part of #2948

That PR was biiiig and a nice overhaul of some input stuff so it isn't really a huge surprise that it managed to break something. And that's why it helps to build all the things - which I'm still antsy to get robots to do for me but oh well not yet :-)

Of possible interest to @oniatus @msteiger @rzats @vampcat @ThompsonTyler @eviltak ... pretty much anybody with some related exposure who can update the module real quick so it builds again :D

anuar2k:
https://github.com/MovingBlocks/Terasology/pull/3219 fixed with this pr.

oniatus:
Can't test it right now but `BindsConfig` has a method `List<Input> getBinds(SimpleUri uri)` which looks promising :) Maybe delegating to this method is simpler than implementing new logic.

anuar2k:
@oniatus Well, usage of both BindsConfig and BindsSubsystem cause ModuleClassLoader denying access to these classes so I gotta find a way around the moduleloader to at least check if both methods return the same values.

Cervator:
Closing as fixed by #3247 whether that's the cleanest fix or not. Stuff should build at least :-)

Thanks for the input and code fix ideas, can still apply a more proper fix in the future.

PR #3247, opened 2018-02-02
API fix for #3184 and minor code quality tweaks

Cervator:
More or less stole the fix @anuar2k put in #3219 (thanks! Even if it is just a restored method) but with a method in the old class that's flagged API that just calls the restored method in a new non-API place. So as far as the Dialogs module is concerned everything is back to normal and we're in API compliance again.

That is probably not ideal from a code perspective, since #2948 was doing some input binding overhaul activity, and this just puts one of the old methods back. I left TODOs in both places for future review and hopefully later on in some other input related effort we can sort it out. Maybe even in v2 since then we can break API in a valid fashion :D

Again may be interesting to @oniatus @msteiger @rzats @vampcat @ThompsonTyler @eviltak - but I'll merge this soonish just so work on Alpha 9 can continue. 

Note that I haven't actually *tested* the fix to Dialogs, as the one example I know of that uses it is in the `develop` branch for LightAndShadow, which is 2 years out of date. Tried to get it up to date but too much effort right now :-(

On a related note but not directly I found another API violation in #3118 while digging around for this issue, and am fixing that in Terasology/PolyWorld#15 (only module affected I've found so far, so fixing in module land with the engine violation OKed)

### DIFF
--- a/engine/src/main/java/org/terasology/engine/subsystem/config/BindsSubsystem.java
+++ b/engine/src/main/java/org/terasology/engine/subsystem/config/BindsSubsystem.java
@@ -1,4 +1,18 @@
-
+/*
+ * Copyright 2018 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


Cervator:
Self merged so I can continue building things, but anybody feel free to still provide feedback :-)

We can make an alternative fix within the next few days (maybe) if a non-API breaking solution is cleaner and easier, or an API violating fix after v2 is merged in.

PR #3250, opened 2018-02-03
Show chunks faster and reduce amount of invisible chunks (Issue #1886).

flo:
### Contains

Improves the situation described in issue #1886.  In single player chunks become now visible when the nearby chunks got loaded and the chunk in the current location got light merged.

In multiplayer the amount of needed chunks before the first chunk becomes visible should be reduced from 7x7x7 to 5x5x5. It is an improvement but there is still room for further optimization.

### How to test

Change the view distance to "Legally Bild (5x5x5)" and then start a single player game. Do this once with and without the changes of this PR. Without this PR only a single chunk is visible. With this PR 9 chunks are visible (3x3x3).

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
@@ -377,7 +377,14 @@ class RenderableWorldImpl implements RenderableWorld {
     }
 
     private boolean isChunkValidForRender(RenderableChunk chunk) {
-        return chunk.isReady() && chunk.areAdjacentChunksReady();
+        if (!chunk.isReady()) {
+            return false;
+        }
+        boolean surroundingChunksLoaded = worldProvider.getWorldViewAround(chunk.getPosition()) == null;
+        if (surroundingChunksLoaded) {
+            return false;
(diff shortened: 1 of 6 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
```diff
@@ -377,7 +377,14 @@ public void dispose() {
     }
 
     private boolean isChunkValidForRender(RenderableChunk chunk) {
-        return chunk.isReady() && chunk.areAdjacentChunksReady();
+        if (!chunk.isReady()) {
+            return false;
+        }
+        boolean surroundingChunksLoaded = worldProvider.getWorldViewAround(chunk.getPosition()) == null;
```
vampcat: Doesn't this variable actually represent `surroundingChunksNotLoaded` ?

GooeyHub:

Cervator:
Checked this out locally, can confirm legally blind now feels substantially less blindy :-)

Also threw in a quick commit to add `configs` as an excluded dir in IntelliJ, forgot to do that when merging another thing.

PR #3252, opened 2018-02-06
Rework path management to be more explicit

vampcat:
Adds the `configs` folder to gitignore, and makes the path selection for the same more organized.

Potentially relevant to @emanuele3d @oniatus @eviltak

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,6 +76,7 @@ Thumbs.db
 /saves/
 /screenshots/
 /logs/
+/configs/
 /terasology-server/
 /terasology-2ndclient/
 /terasology-3rdclient/
--- a/engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
+++ b/engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
@@ -87,21 +87,10 @@ public class FlexibleConfigManagerImpl implements FlexibleConfigManager {
         }
     }
 
-    private void ensureDirectoryExists(Path filePath) {
-        try {
-            Files.createDirectories(filePath.getParent());
-        } catch (Exception e) {
-            throw new RuntimeException("Cannot create directory for FC!");
-        }
-    }
-
     private Path getPathForFlexibleConfig(SimpleUri flexibleConfigId) {
-        Path filePath = PathManager.getInstance()
-                                        .getHomePath()
-                                        .resolve("configs")
-                                        .resolve(flexibleConfigId.getModuleName().toString())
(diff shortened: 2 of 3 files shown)


GooeyHub:

.gitignore
```diff
@@ -76,6 +76,7 @@ Thumbs.db
 /saves/
 /screenshots/
 /logs/
+/configs/
```
eviltak: This change has already been made in #3227 - if you think it is better here, I shall revert the change in that PR.

vampcat:
I made the change here because it makes the most sense with this PR (oniatus wanted that change to go in a separate PR either way). So yeah, it would be great if you could revert.

engine/src/main/java/org/terasology/config/flexible/FlexibleConfigManagerImpl.java
```diff
@@ -87,21 +87,10 @@ public void saveAllConfigs() {
...
-        }
-    }
-
     private Path getPathForFlexibleConfig(SimpleUri flexibleConfigId) {
-        Path filePath = PathManager.getInstance()
-                                        .getHomePath()
-                                        .resolve("configs")
-                                        .resolve(flexibleConfigId.getModuleName().toString())
-                                        .resolve(flexibleConfigId.getObjectName().toString() + ".cfg");
-        ensureDirectoryExists(filePath);
```
emanuele3d: What's the reasoning behind the removal of this check? Is the directory guaranteed to exist from some other parts of the code?

vampcat: Yes.

https://github.com/MovingBlocks/Terasology/pull/3252/files#diff-5bbcd1d375372f183a3afd8dafa16318R274

Cervator:
For the `configs` thing in #3227 watch out for Git getting any funny ideas. If it is added here, then also added + deleted in that other PR, when it gets merged it may try to delete it instead. Rebase fixes but brings in its own hassle. Yay Git! :-)

eviltak:
Great point, @Cervator, but I do not think it should be a problem in this case as the change in #3227 is on a separate line. Nonetheless, to avoid any problems the _best_ way out, in my opinion, is handling this through the merge conflict which will result on trying to merge #3227 after this PR.

Cervator:
Added in the merge on top of #3227 and manually resolved the one thing in the `.gitignore` - no technical conflict just had to remove one of the `configs` lines :-)

emanuele3d:
Yay!!! Congrats everybody! And in particular @eviltak! Thank you for sticking to it!

PR #3254, opened 2018-02-09
Basic groundwork for exposing render API (WiP)

vampcat:
The basic aim of this PR is to open up the RenderGraph and associated classes to modules, and enable them to add their own Nodes.

# Completed
Basic changes to expose RenderGraph to modules

# Remaining
- [ ] Add handler functions (onAdd, onRemove and so on) for Nodes
- [ ] Add `aka` attribute to Nodes
- [ ] Automatic module detection for Nodes, when setting their UID

### DIFF
--- a/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
+++ b/engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
@@ -26,6 +26,7 @@ import org.terasology.monitoring.PerformanceMonitor;
 import org.terasology.rendering.assets.material.Material;
 import org.terasology.rendering.cameras.Camera;
 import org.terasology.rendering.cameras.SubmersibleCamera;
+import org.terasology.rendering.dag.RenderGraph;
 import org.terasology.rendering.world.viewDistance.ViewDistance;
 import org.terasology.rendering.world.WorldRenderer;
 import org.terasology.world.WorldProvider;
@@ -190,6 +191,11 @@ public class HeadlessWorldRenderer implements WorldRenderer {
     @Override
     public String getMetrics() {
         return "";
+    }
+
+    @Override
+    public RenderGraph getRenderGraph() {
+        return null;
     }
 
     /**
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
@@ -17,9 +17,11 @@ package org.terasology.rendering.world;
 
 import org.terasology.math.geom.Vector3f;
 import org.terasology.math.geom.Vector3i;
+import org.terasology.module.sandbox.API;
 import org.terasology.rendering.assets.material.Material;
 import org.terasology.rendering.cameras.Camera;
 import org.terasology.rendering.cameras.SubmersibleCamera;
+import org.terasology.rendering.dag.RenderGraph;
 import org.terasology.rendering.world.viewDistance.ViewDistance;
 
 /**
@@ -218,4 +220,11 @@ public interface WorldRenderer {
      */
     // TODO: transform this to return an object or a map. Consumers would then take care of formatting.
     String getMetrics();
+
+    /**
+     * Returns the RenderGraph used to organize the nodes. This can be used to Modules to add/remove Nodes to the game.
+     *
+     * @return a RenderGraph used to organize the nodes.
+     */
+    RenderGraph getRenderGraph();
 }
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -31,6 +31,7 @@ import org.terasology.logic.players.LocalPlayerSystem;
 import org.terasology.math.TeraMath;
 import org.terasology.math.geom.Vector3f;
 import org.terasology.math.geom.Vector3i;
+import org.terasology.registry.CoreRegistry;
 import org.terasology.rendering.ShaderManager;
 import org.terasology.rendering.assets.material.Material;
 import org.terasology.rendering.backdrop.BackdropProvider;
@@ -126,7 +127,7 @@ public final class WorldRendererImpl implements WorldRenderer, ComponentSystem {
      * It's not, so for now, we use this factor to adjust for the disparity.
      */
     private static final float GROUND_PLANE_HEIGHT_DISPARITY = -0.7f;
-    private static RenderGraph renderGraph = new RenderGraph(); // TODO: Try making this non-static
+    private RenderGraph renderGraph = new RenderGraph();
 
     private boolean isFirstRenderingStageForCurrentFrame;
     private final RenderQueuesHelper renderQueues;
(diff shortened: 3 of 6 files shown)


engine/src/main/java/org/terasology/engine/subsystem/headless/renderer/HeadlessWorldRenderer.java
```diff
@@ -192,6 +193,11 @@ public String getMetrics() {
         return "";
     }
 
+    @Override
```
vampcat: I decided to go with the easier option for now - exposing the RenderGraph via WorldRenderer.

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
```diff
@@ -218,4 +220,11 @@
      */
     // TODO: transform this to return an object or a map. Consumers would then take care of formatting.
     String getMetrics();
+
+    /**
+     * Returns the RenderGraph used to organize the nodes. This can be used to Modules to add/remove Nodes to the game.
```
vampcat: This needs more work, I'm aware of that.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -126,7 +127,7 @@
      * It's not, so for now, we use this factor to adjust for the disparity.
      */
     private static final float GROUND_PLANE_HEIGHT_DISPARITY = -0.7f;
-    private static RenderGraph renderGraph = new RenderGraph(); // TODO: Try making this non-static
+    private RenderGraph renderGraph = new RenderGraph();
```
vampcat: Okay, so this was a tricky thing. Not *strictly* necessary for exposing the API, but good to tackle nonetheless.
The way @RegisterSystem works is, it creates a new instance of the class (which is also used for the commands registered via @Command). Which means we effectively have 2 WorldRenederer instances - one that is used to draw the world, and one that is used for handling commands. Which is why i had to create the variable static. Definitely not ideal, and I think we should tackle it in this PR (since I'm tackling other render graph related things as well).

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -641,7 +642,8 @@ private void preRenderUpdate(RenderingStage renderingStage) {
         renderableWorld.queueVisibleChunks(isFirstRenderingStageForCurrentFrame);
 
         if (requestedTaskListRefresh) {
-            renderTaskListGenerator.refresh();
+            List<Node> orderedNodes = renderGraph.getNodesInTopologicalOrder();
```
vampcat: Another problematic area in the past. `refresh()` basically uses the exact same Nodes as last time. Which means even though if we update the renderGraph, `refresh()` won't actually include the new Nodes in the final taskList. Since we were dealing with only 42 Nodes till now (with some being enabled/disabled), it did not matter. Now that we're dealing with 42 plus/minus some Nodes, we have to generate the taskList from the new set of Nodes (which might or might not be the same as last time).

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -808,9 +810,15 @@ public void postSave() {
...
 
+    @Override
+    public RenderGraph getRenderGraph() {
+        return renderGraph;
+    }
+
     @Command(shortDescription = "Debugging command for DAG.", requiredPermission = PermissionManager.NO_PERMISSION)
     public void dagNodeCommand(@CommandParam("nodeUri") final String nodeUri, @CommandParam("command") final String command, @CommandParam(value= "arguments") final String... arguments) {
-        Node node = renderGraph.findNode(nodeUri);
+        WorldRenderer worldRenderer = CoreRegistry.get(WorldRenderer.class);
```
vampcat: Since renderGraph is no longer a static variable, the instance of WR which is handling the commands (which was created by RegisterSystem) does not have the correct renderGraph variable. So, I'm obtaining it via CoreRegistry (rather than have a `@In WorldRenderer worldRenderer` inside WorldRenderer class - no need to confuse people so much. Might I recommend reconsidering the creation of a dedicated NodeCommands class? A `@In WorldRenderer worldRenderer` won't confuse people in that class.

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRenderer.java
emanuele3d:
```
* Returns the RenderGraph.
*
* This object is used by Engine and Modules to add/remove Nodes to/from the rendering process.
*
* @return the RenderGraph containing the nodes used by the rendering process.
```

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: WHAAAT??? So, we also have two render graphs? And which WorldRenderer instance is actually used for rendering? I mean, why is this not a massive problem?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: I noticed the change to the enabling/disabling of conditional nodes. The explanation above is not clear to me. Can you please rephrase everything in a new explanation?

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: I need a bit more context on why there are two separate instances of WorldRenderer, see earlier comment.

vampcat:
Regarding the nodeList refresh: Here is what `RenderTaskListGenerator.refresh()` looks like:
```
    public void refresh() {
        generateFrom(nodeList);
    }
```
where nodeList is the topologically sorted list we passed to `generateFrom(List<Node> orderedNodes)` in the first place. Now, so far we've been dealing with 43 Nodes, meaning any changes we make to the Nodes (but not their connections) will not actually change nodeList (which is essentially the topologically sorted list), and hence refresh will work as expected. But what if I add a new Node to the renderGraph, and call `refresh()`? `nodeList` won't contain the new node that I just added, and it will generate the exact same taskList as before - which is a problem. Hence, rather than *just* refresh the *current* list, we generate a new topolically sorted list from our renderGraph.

Cervator:
The double instance makes sense to me from reading the code - even if it is somewhat unexpected.

`WorldRendererImpl` implements both `WorldRenderer` and `ComponentSystem` and its two roles are instantiated independently. Poof, two separate WRs.

Seems sensible to take off the `@RegisterSystem` - I remember there was some discomfort over that in the first place anyway, and it has a bunch of empty ES override methods.

To access it through the `Context` we could probably just use a wrapper system of some sort that has a reference to the one true WR. That might also help clear up the code and shrink WR a bit by splitting out the two roles.

vampcat:
The only role "WR as System" plays is to process the node commands. I feel like we can add that to separate class, since we're exposing the renderGraph either way.

emanuele3d:
It seems to me removing @RegisterSystem is the only "easy way" to address this. I am now wondering if there are other situations in which, inadvertently, we create two instances of the same thing. I'm also wondering if we shouldn't give up on this and try and find another way, to strive and keep the renderer more aligned with the general ES architecture where systems play a big part. 

Given we have to look at how to do consolle commands differently, can we look into how arbitrary objects that aren't systems can register consolle commands on their own? Nodes will likely need that too at some point. 

I don't remember, why are we registering the dagNodeCommand via the @Command annotation rather than using the Console.registerCommand(ConsoleCommand) method?

vampcat:
> strive and keep the renderer more aligned with the general ES architecture
Currently, the WR is part of the "basic framework" that is used to support the ES architecture. Moving it from "part of framework" to "a system running on the framework" is not trivial. Certain possible, but it'll take time.

Regarding the `Console.registerCommand()` : I'll have a crack at it later today :D

GooeyHub:

vampcat:
Closing, in favour of smaller PRs.

PR #3258, opened 2018-02-11
Sample direction that FCs are going towards [RFR]

vampcat:
A small test to see how well FCs replace our current settings.

### DIFF
--- a/engine/src/main/java/org/terasology/config/RenderingConfig.java
+++ b/engine/src/main/java/org/terasology/config/RenderingConfig.java
@@ -18,6 +18,9 @@ package org.terasology.config;
 
 import org.lwjgl.opengl.DisplayMode;
 import org.lwjgl.opengl.PixelFormat;
+import org.terasology.config.flexible.FlexibleConfig;
+import org.terasology.config.flexible.SettingImpl;
+import org.terasology.engine.SimpleUri;
 import org.terasology.rendering.cameras.PerspectiveCameraSettings;
 import org.terasology.rendering.nui.layers.mainMenu.videoSettings.DisplayModeSetting;
 import org.terasology.rendering.nui.layers.mainMenu.videoSettings.ScreenshotSize;
@@ -28,13 +31,9 @@ import java.beans.PropertyChangeListener;
 
 /**
  */
+@SuppressWarnings("WeakerAccess")
 public class RenderingConfig extends AbstractSubscribable {
-
     public static final String PIXEL_FORMAT = "PixelFormat";
-    public static final String WINDOW_POS_X = "WindowPosX";
-    public static final String WINDOW_POS_Y = "WindowPosY";
-    public static final String WINDOW_WIDTH = "WindowWidth";
-    public static final String WINDOW_HEIGHT = "WindowHeight";
     public static final String DISPLAY_MODE_SETTING = "DisplayModeSetting";
     public static final String ANIMATED_MENU = "AnimatedMenu";
     public static final String VIEW_DISTANCE = "viewDistance";
@@ -76,6 +75,11 @@ public class RenderingConfig extends AbstractSubscribable {
     public static final String SCREENSHOT_FORMAT = "ScreenshotFormat";
     public static final String DUMP_SHADERS = "DumpShaders";
     public static final String VOLUMETRIC_FOG = "VolumetricFog";
+
+    public static final SimpleUri WINDOW_POS_X_ = new SimpleUri("engine:windowPosX");
+    public static final SimpleUri WINDOW_POS_Y_ = new SimpleUri("engine:windowPosY");
+    public static final SimpleUri WINDOW_WIDTH_ = new SimpleUri("engine:windowWidth");
+    public static final SimpleUri WINDOW_HEIGHT_ = new SimpleUri("engine:windowHeight");
 
     private PixelFormat pixelFormat;
     private int windowPosX;
@@ -125,6 +129,8 @@ public class RenderingConfig extends AbstractSubscribable {
     private String screenshotFormat;
     private PerspectiveCameraSettings cameraSettings;
 
+    private FlexibleConfig flexibleConfig;
+
     private RenderingDebugConfig debug = new RenderingDebugConfig();
 
     public PerspectiveCameraSettings getCameraSettings() {
@@ -143,46 +149,6 @@ public class RenderingConfig extends AbstractSubscribable {
         // This method could theoretically use a better equality check then. In practice
         // it's unlikely a new PixelFormat instance will ever be value-per-value identical
         // to the previous one.
-    }
-
-    public int getWindowPosX() {
-        return windowPosX;
-    }
-
-    public void setWindowPosX(int windowPosX) {
-        int oldValue = this.windowPosX;
-        this.windowPosX = windowPosX;
-        propertyChangeSupport.firePropertyChange(WINDOW_POS_X, oldValue, this.windowPosX);
-    }
-
-    public int getWindowPosY() {
-        return windowPosY;
-    }
-
-    public void setWindowPosY(int windowPosY) {
-        int oldValue = this.windowPosY;
-        this.windowPosY = windowPosY;
-        propertyChangeSupport.firePropertyChange(WINDOW_POS_Y, oldValue, this.windowPosY);
-    }
-
-    public int getWindowWidth() {
-        return windowWidth;
-    }
-
-    public void setWindowWidth(int windowWidth) {
-        int oldValue = this.windowWidth;
-        this.windowWidth = windowWidth;
-        propertyChangeSupport.firePropertyChange(WINDOW_WIDTH, oldValue, this.windowWidth);
-    }
-
-    public int getWindowHeight() {
-        return windowHeight;
-    }
-
-    public void setWindowHeight(int windowHeight) {
-            int oldValue = this.windowHeight;
-            this.windowHeight = windowHeight;
-            propertyChangeSupport.firePropertyChange(WINDOW_HEIGHT, oldValue, this.windowHeight);
     }
 
     public DisplayMode getDisplayMode() {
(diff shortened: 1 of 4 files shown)


GooeyHub:

engine/src/main/java/org/terasology/config/RenderingConfig.java
```diff
@@ -77,6 +76,11 @@
     public static final String DUMP_SHADERS = "DumpShaders";
     public static final String VOLUMETRIC_FOG = "VolumetricFog";
 
+    public static final SimpleUri WINDOW_POS_X_ = new SimpleUri("engine:windowPosX");
+    public static final SimpleUri WINDOW_POS_Y_ = new SimpleUri("engine:windowPosY");
```
eviltak: I think both of these can be put together into a single `Setting<Vector2i>` for easier access. However, this change will break things, so it may be better off in a separate PR.

eviltak:
I find the usage of `new SimpleUri` everywhere quite unwieldy. I think we can get around this by creating overloads of the `SettingImpl` constructor and `FlexibleConfig.get` that take a string parameter, convert it to a `SimpleUri`, and pass it to the `SimpleUri` overload.

engine/src/main/java/org/terasology/config/RenderingConfig.java
vampcat: I actually favour this change, for reasons other than easier access. Rather than subscribe to either width *and* height, you can simply subscribe to one property. So I think let's make it a `Vector2i` in this PR, unless someone has a problem. This PR will be kinda messy and break-all-the-things either way.

engine/src/main/java/org/terasology/engine/subsystem/lwjgl/LwjglDisplayDevice.java
```diff
@@ -78,7 +81,7 @@ public void setDisplayModeSetting(DisplayModeSetting displayModeSetting, boolean
             switch (displayModeSetting) {
                 case FULLSCREEN:
                     Display.setDisplayMode(Display.getDesktopDisplayMode());
-                    Display.setLocation(config.getWindowPosX(), config.getWindowPosY());
+                    Display.setLocation((Integer)config.getFlexibleConfig().get(WINDOW_POS_X_).getValue(), (Integer)config.getFlexibleConfig().get(WINDOW_POS_Y_).getValue());
```
vampcat: I definitely don't like this cast, but I can't see any way around it for now.

vampcat:
Also, @eviltak I totally agree about the overloaded String counterparts, I was about to make that change either way :D

Regarding the wrapper (point 2) : Not just the constants, the casting as well. The difference between `(Integer)context.get(FlexibleConfigManager.class).get("engine:rendering").get("engine:windowPosX").getValue()` and `context.get(RenderingConfig.class).getWindowPosX()` is huge, imo. That being said, it does introduce some extra classes, so..

Not sure what you mean by the `Key` class, since I really don't want to look at every commit and figure out which one has the Key class (it's not present in the net changes).

eviltak:
@vampcat So you mean a wrapper class with wrapper methods for each setting like it currently is? While I can see the aesthetic appeal, it kind of defeats the purpose of using a FlexibleConfig, IMHO. 

As for the casts, I see no way of avoiding them unless:
- The setting is first extracted to a local variable: 
```
Setting<Vector2i> windowPosSetting = config.getFlexibleConfig().get(RenderingConfig.WINDOW_POS);
Display.setLocation(windowPosSetting.getValue().x, windowPosSetting.getValue().y);
```
- The setting type is passed to the method either via the generic type parameter or via a `.class` argument:
```
config.getFlexibleConfig().get(WINDOW_POS_X_, Integer.class).getValue()
// OR
config.getFlexibleConfig().<Integer>get(WINDOW_POS_X_).getValue()
```

   or
- The key method is used. Keys were disproved for being over-engineered and unwieldy (an assessment I largely agree with) and are effective only when _all_ setting keys are stored for easy retrieval (like a constant or a static field). For the implementation, please take a look at this [old FlexibleConfig implementation](https://github.com/MovingBlocks/Terasology/blob/772413a5857359eedcab23a679f8cfd8c5e458ea/engine/src/main/java/org/terasology/config/flexible/FlexibleConfig.java). If there is renewed interest in this approach, I shall elucidate the usage of keys.

vampcat:
Yes, it kinda defeats the purpose of FCs. Which is why I've tried to avoid that approach in this PR. However, as you can see, it feels way too verbose. Hence this PR - to get more eyes to look at it :D

engine/src/main/java/org/terasology/config/RenderingConfig.java
```diff
@@ -657,4 +623,74 @@ public void setVolumetricFog(boolean volumetricFog) {
         propertyChangeSupport.firePropertyChange(VOLUMETRIC_FOG, oldValue, this.volumetricFog);
     }
 
+    public void loadDefaultRenderingConfig(FlexibleConfig flexibleConfig) {
+        this.flexibleConfig = flexibleConfig;
+
+        // flexibleConfig.add(new SettingImpl<>(new SimpleUri("engine:pixelFormat"), 24));
+        flexibleConfig.add(new SettingImpl<>(new SimpleUri("engine:windowPosX"), 0));
```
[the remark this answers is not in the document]
eviltak: Yes, it will be possible to do that. I proposed something similar in the last couple of comments on my PR. This change will also have the added benefit of not requiring the `SettingImpl` class being exposed to the API, satiating @oniatus's concerns. I also think that the `add(Setting setting)` method should entirely be removed to prevent setting publishers from putting in their own potentially malicious `Setting` implementations, but that may be something to discuss at another time.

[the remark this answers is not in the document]
emanuele3d: @eviltak: what do you have in mind when you think about a malicious setting implementation?

Regarding the SettingImpl class, I'm not bought on the idea that it shouldn't be public.

emanuele3d:
@vampcat, @eviltak: the three renderer-related PRs have just been merged and we can focus once more on this one. Please let's get back on this and let's see what we can do about it.

eviltak:
Readdressing something I briefly mentioned in a review comment (should address your clarification requests @emanuele3d):

I believe that we should change the current `FlexibleConfig` interface (and potentially the `FlexibleConfigManager` API too) to _not_ allow setting/FC publishes to add their own instances of `Setting`/`FlexibleConfig`, for the reasons that follow.

- I do not see the need to include more than one implementation of `Setting` in the future. Consider the possibilities of future implementations of `Setting`. Apart from `SettingImpl`, the only other implementation we have is `MockSetting` which I believe can also be removed. As such, I do not see the need to store in a FC an instance of a `Setting` implementation other than `SettingImpl`. Besides, even if we do add a couple of implementations of `Setting`, we can easily create overloads of the `add` method (or new methods entirely) that add a different `Setting` implementation.

- Modules will _not_ be allowed to add instances of their own `Setting` implementations to a FC. A module author may pass in a malicious `Setting` implementation like `MockSetting` that simply returns `null` in `getValue` and `getDefaultValue`. For this reason, I think it is better to prevent the setting publisher from creating a `Setting` instance and passing it to the FC - instead, the FC itself should be responsible for the creation of the `SettingImpl` object.

- External consumers will not have to deal with any _implementations_ of `Setting` - all their work can be done through the interface. This is in line with the usage patterns of other `Interface` -- `InterfaceImpl` pairs in the code base (`Chunk` -- `ChunkImpl`, `BlockManager` -- `BlockManagerImpl`, et cetera).

In code, `boolean FlexibleConfig.add(Setting setting)` should be replaced with `Setting<V> FlexibleConfig.add<V>(SimpleUri id, V defaultValue)` and `Setting<V> FlexibleConfig.add<V>(SimpleUri id, V defaultValue, SettingValueValidator<V> valueValidator)`. Both of these methods internally create a `SettingImpl<V>`, add it to the settings map with the already existing rules, and return the created `Setting`. This ensures that the setting publisher has nothing to do with the instantiation of the setting object (and, in turn, does not have to deal with `SettingImpl` or a `SettingFactory`).

A similar approach for similar reasons can be taken towards the `FlexibleConfigManager.addConfig` method, which also requires that the config publisher create a new `FlexibleConfigImpl` instance and pass it to the method. Since no other `FlexibleConfig` implementation will ever be used, it is better to change the signature of `addConfig` to `FlexibleConfig addConfig(SimpleUri id)` and let the method create the `FlexibleConfigImpl` instance.

@vampcat @oniatus @emanuele3d please put forward your thoughts and suggestions.

emanuele3d:
@eviltak, you need to look at your own communication and reasoning style as this really doesn't work. 

Let me clarify:

> I believe that we should change the current FlexibleConfig interface (and potentially the FlexibleConfigManager API too) to not allow setting/FC publishes to add their own instances of Setting/FlexibleConfig, for the reasons that follow.

This ignores and is in direct contradiction of my intention to get some experience first and then reassess. So, generally speaking it's a no.

> I do not see the need to include more than one implementation of Setting in the future.

This is your personal judgement and I disagree. So, it's another no.

> Besides, even if we do add a couple of implementations of Setting, we can easily create overloads of the add 

This is another no, as we don't want to change the engine source to allow for additional setting types. 

> Modules will not be allowed to add instances of their own Setting implementations to a FC.

This imperious tone is absolutely out of place: you do not get to decide, at most suggest. Fourth no. Furthermore, I was curious about what you intended with malicious code and the example you provide doesn't sound malicious at all, just bad programming. If a module behaves like that it will be easily forgotten, never to be loaded again. 

Summary: lots of circular reasoning, poor arguments and unacceptable communication style. 

We work with what we have, we get some experience with it and we reassess later. Perhaps there will be scope for some of the changes you suggest @eviltak, but for the time being I don't want to discuss them again as it's a waste of time.

emanuele3d:
@vampcat, please address my comments and let's see where we end up. 

Also, we should keep in mind is that what looks verbose now, in a single FC, with lots of repetition, will eventually be decentralized into many parts of the code, i.e. individual nodes adding or obtaining their own settings. 

So, we are going through a lot of cut&pasting right now, but eventually, a node developer or a module developers will only add or obtain the setting he or she is interested in: the whole list will be visible only UI side, and not necessarily all at once.

eviltak:
@emanuele3d while I am otherwise ok with your judgement of my comment, you may have mistook the statement you called imperious. In no way am I domineering or making decisions for others through that statement -- it is simply a direct consequence of my proposal.

eviltak:
Closing this PR as FCs have been replaced with AutoConfigs in #3723.

Issue #483, opened 2013-04-10
Add 'tab' order to GUI components


Cervator:
Just realized we don't have it. Primarily thinking about the main menu like the create world screen or various settings
##

Cervator:
Closing this as completed at long last by #3561 - woohoo!

PR #3260, opened 2018-02-14
Automatically follow HTTP/HTTPS redirects for identity storage service

gianluca-nitti:
### Contains

The identity storage service client now correctly follows HTTP redirects (301/302). Particularly useful if the server forces HTTP requests to move to HTTPS through 301/302 redirects, like the official server (http://utility.terasology.org) is now doing thanks to @qwc.

### How to test

Try to use the identity storage service on the official server (http://utility.terasology.org), by clicking (from the main menu) Settings > Player > Login. On the current master/develop branches, it will only work if you enter the HTTPS URL (https://utility.terasology.org); the login will fail if you put the plain HTTP URL (http://utility.terasology.org), because it redirects to the HTTPS one but the client does not follow this redirect.
On this branch, both the HTTP and HTTPS URLs work (HTTP can be a bit slower due to the larger number of requests involved) because when entering the HTTP one the redirect to the HTTPS one is correctly followed.

Not sure how much it's related but maybe we could populate by default the "Service URL" field in the login dialog with https://utility.terasology.org if this can help new users. Obviously, one would still be allowed to modify the field to use an alternative server if they want, anyway I'm not sure if populating it by default with our server is appropriate, just dropping it here as an idea.

### DIFF
--- a/engine/src/main/java/org/terasology/identity/storageServiceClient/ServiceApiRequest.java
+++ b/engine/src/main/java/org/terasology/identity/storageServiceClient/ServiceApiRequest.java
@@ -27,6 +27,7 @@ import java.io.OutputStream;
 import java.math.BigInteger;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.net.URLDecoder;
 
 /**
  * Utility class to perform requests to the service API.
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Thanks for that - tested out well :-)

Although, I tried signing up for a new account and after several minutes don't have the confirmation email so I can't actually log in. Was able to still validate it anyway. I have a bad feeling one or more of our servers are stuck in an IP group that got used for spamming and blacklisted recently. Getting lots of email bounces for emails sent by the forum :/

I don't know if there is a return email for the confirmation emails, to indicate a failure to send? Wonder how we can improve that setup.

I added in `https://utility.terasology.org` as default text for the URL, that seems perfectly fair and a good convenience. Also reminded me how badly we need the UI tabbing support #483 :D

PR #3261, opened 2018-02-15
Converted WorldRenderer to a non-system class

vampcat:
As discussed in #3254, WR should not be a system, since that leads to 2 instances of this singleton class. This PR addresses the issue by removing the System status of the class, and instead registering the commands in it through `MethodCommand.registerAvailable`.

This PR contains a second smaller commit, that moves the `recompileShaders` command from CoreCommands to WorldRenderer.

As a side note, I still feel like WorldRenderer is becoming a monolith class, and we should consider moving commands to a different class. Any feedback on this idea would be great.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -776,9 +776,9 @@ public final class WorldRendererImpl implements WorldRenderer {
      */
     @Command(shortDescription = "Forces a recompilation of shaders.", requiredPermission = PermissionManager.NO_PERMISSION)
     public void recompileShaders() {
-        console.addMessage("Recompiling shaders..");
+        console.addMessage("Recompiling shaders... ");
         shaderManager.recompileAllShaders();
-        console.addMessage("Done!");
+        console.addMessage("done!");
     }
 
     /**
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -769,6 +769,9 @@ public final class WorldRendererImpl implements WorldRenderer {
     /**
      * Forces a recompilation of all shaders. This command, backed by Gestalt's monitoring feature,
      * allows developers to hot-swap shaders for easy development.
+     *
+     * Usage:
+     *      recompileShaders
      */
     @Command(shortDescription = "Forces a recompilation of shaders.", requiredPermission = PermissionManager.NO_PERMISSION)
     public void recompileShaders() {
@@ -778,6 +781,12 @@ public final class WorldRendererImpl implements WorldRenderer {
     /**
      * Acts as an interface between the console and the Nodes. All parameters passed to command are redirected to the
      * concerned Nodes, which in turn take care of executing them.
+     *
+     * Usage:
+     *      dagCommandNode <nodeUri> <command> <parameters>
+     *
+     * Sample Usage:
+     *      dagNodeCommand engine:outputToScreenNode setFbo engine:fbo.ssao
      */
     @Command(shortDescription = "Debugging command for DAG.", requiredPermission = PermissionManager.NO_PERMISSION)
     public void dagNodeCommand(@CommandParam("nodeUri") final String nodeUri, @CommandParam("command") final String command, @CommandParam(value= "arguments") final String... arguments) {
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -780,34 +766,19 @@ public RenderingStage getCurrentRenderStage() {
         return currentRenderingStage;
     }
 
+    /**
+     * Forces a recompilation of all shaders. This command, backed by Gestalt's monitoring feature,
+     * allows developers to hot-swap shaders for easy development.
+     */
```
emanuele3d: How about we get in the habit of providing examples of the command?
I.e. in this case we can add in the comment:

`* To run the command simply type "recompileShaders" + Enter in the consolle. `

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -780,34 +766,19 @@ public RenderingStage getCurrentRenderStage() {
...
-    }
-
-    @Override
-    public void shutdown() {
-    }
-
+    /**
+     * Acts as an interface between the console and the Nodes. All parameters passed to command are redirected to the
+     * concerned Nodes, which in turn take care of executing them.
+     */
```
emanuele3d: Similarly to the comment above, let's add an example usage here to. I.e.:

```
* Usage:
*     dagCommandNode <nodeUri> <command> <parameters>
*
* Example:
*     dagCommandNode (what's the command to look at the content of FBOs again?)
```

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -198,6 +181,7 @@ public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
         this.backdropProvider = context.get(BackdropProvider.class);
         this.renderingConfig = context.get(Config.class).getRendering();
         this.shaderManager = context.get(ShaderManager.class);
+        // TODO: Instantiate the VR provider at a more reasonable location, and just obtain it via context here.
```
emanuele3d: Hehe, big can of worms there... :wink:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -231,6 +215,8 @@ public WorldRendererImpl(Context context, GLBufferPool bufferPool) {
         renderQueues = renderableWorld.getRenderQueues();
 
         initRenderingSupport();
+
+        MethodCommand.registerAvailable(this, context.get(Console.class), context);
```
emanuele3d: This is neat! We'll have to remember to use it elsewhere!! :smiley:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -770,8 +770,7 @@ public RenderingStage getCurrentRenderStage() {
      * Forces a recompilation of all shaders. This command, backed by Gestalt's monitoring feature,
      * allows developers to hot-swap shaders for easy development.
      *
-     * Usage:
-     *      recompileShaders
+     * To run the command simply type "recompileShaders" and then press Enter in the console.
```
vampcat: So picky :D

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -780,34 +766,27 @@ public RenderingStage getCurrentRenderStage() {
         return currentRenderingStage;
     }
 
+    /**
+     * Forces a recompilation of all shaders. This command, backed by Gestalt's monitoring feature,
+     * allows developers to hot-swap shaders for easy development.
+     *
+     * To run the command simply type "recompileShaders" and then press Enter in the console.
```
emanuele3d: By the way, I noticed that the command recompileShaders gives absolutely no feedback. There is no sense if the command actually did anything. Any chance we can print out to the console:

`Recompiling shaders.... (pause) done!`

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
emanuele3d: Aaaaaaaand (mostly) proud of it!

GooeyHub:

emanuele3d:
Hmmm. I just noticed that `Console.addMessage` adds a newline after each message. 

It's out of the scope of this PR, so let's no worry about it. I'll write a small issue on the matter.

GooeyHub:

PR #3262, opened 2018-02-16
ConditionDependentNode rework

vampcat:
This PR is somewhat of an experimental PR, since I'm not *entirely* sold on adding changeListener for every single condition dependent node, even if it's barely a single line. The only good work-around I see is to retain PropertyChangeListener in CDN, but that kinda makes it confusing for new readers to understand, since the Node related code is split across places.

Feedback on PropertyChangeListener notwithstanding, the rest of the rework (related to conditions) is still something I'd like to go ahead with.

Further explanation of the new system, copied from a comment below:

Whenever a property was changed, CDN called checkConditions(), which updated the "status" of the node (whether it's enabled or disabled), followed by which it called refreshTaskList, which only created tasks from enabled Nodes. There were two glaring issues with this:

If a method decides to override the propertyChange method, and does not explicitly call checkConditions(), enabling/disabling any property won't actually affect the Node's status, even if it leads to some requirements not being met.
The only way to enable/disable a CDN was through settings. That means if I have CDN with requirement bloomEnabled, I have to toggle bloomEnabled in renderingConfig to toggle the Node's status. I can't just call setEnabled() on the Node. Obviously a problem if we want to go for a renderGraph.
My solution:

Instead of changing the enabled/disabled method statically, we simply check the conditions when we are generating the task list.
We separate the status of the node from the conditions of the node. That means we can toggle a node on/off independently of the conditions, but the final status of the node depends on the user-set state and conditions.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
@@ -16,7 +16,10 @@
 package org.terasology.rendering.dag;
 
 import com.google.common.collect.Lists;
+import org.terasology.context.Context;
+import org.terasology.rendering.world.WorldRenderer;
 
+import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.util.List;
 import java.util.function.Supplier;
@@ -24,8 +27,14 @@ import java.util.function.Supplier;
 /**
  * TODO: Add javadocs
  */
-public abstract class ConditionDependentNode extends AbstractNode {
+public abstract class ConditionDependentNode extends AbstractNode implements PropertyChangeListener {
     private List<Supplier<Boolean>> conditions = Lists.newArrayList();
+
+    protected WorldRenderer worldRenderer;
+
+    protected ConditionDependentNode(Context context) {
+        worldRenderer = context.get(WorldRenderer.class);
+    }
 
     protected void requiresCondition(Supplier<Boolean> condition) {
         conditions.add(condition);
@@ -42,5 +51,10 @@ public abstract class ConditionDependentNode extends AbstractNode {
     @Override
     public boolean isEnabled() {
         return enabled && checkConditions();
+    }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent event) {
+        worldRenderer.requestTaskListRefresh();
     }
 }
--- a/engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
@@ -16,10 +16,7 @@
 package org.terasology.rendering.dag;
 
 import com.google.common.collect.Lists;
-import org.terasology.context.Context;
-import org.terasology.rendering.world.WorldRenderer;
 
-import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.util.List;
 import java.util.function.Supplier;
@@ -27,18 +24,11 @@ import java.util.function.Supplier;
 /**
  * TODO: Add javadocs
  */
-public abstract class ConditionDependentNode extends AbstractNode implements PropertyChangeListener {
+public abstract class ConditionDependentNode extends AbstractNode {
     private List<Supplier<Boolean>> conditions = Lists.newArrayList();
-
-    private WorldRenderer worldRenderer;
-
-    protected ConditionDependentNode(Context context) {
-        worldRenderer = context.get(WorldRenderer.class);
-    }
 
     protected void requiresCondition(Supplier<Boolean> condition) {
         conditions.add(condition);
-        checkConditions(); // TODO: better to remove this in near feature
     }
 
     private boolean checkConditions() {
@@ -46,23 +36,11 @@ public abstract class ConditionDependentNode extends AbstractNode implements Pro
         for (Supplier<Boolean> condition : conditions) {
             conditionsAreSatisfied = conditionsAreSatisfied && condition.get();
         }
-
-        if (conditionsAreSatisfied != isEnabled()) {
-            setEnabled(conditionsAreSatisfied);
-            // enabling/disabling here means we cannot enable/disable nodes directly:
-            // we must always go through the settings.
-            return true;
-        } else {
-            return false;
-        }
-
+        return conditionsAreSatisfied;
     }
 
     @Override
-    public void propertyChange(PropertyChangeEvent evt) {
-        boolean conditionsChanged = checkConditions();
-        if (conditionsChanged) {
-            worldRenderer.requestTaskListRefresh();
-        }
+    public boolean isEnabled() {
+        return enabled && checkConditions();
     }
 }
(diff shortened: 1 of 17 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/BloomBlurNode.java
```diff
@@ -57,5 +63,11 @@ protected void setupConditions(Context context) {
         RenderingConfig renderingConfig = context.get(Config.class).getRendering();
         renderingConfig.subscribe(RenderingConfig.BLOOM, this);
         requiresCondition(renderingConfig::isBloom);
+        worldRenderer = context.get(WorldRenderer.class);
+    }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent event) {
```
emanuele3d: These identical methods obviously represent a lot of code duplication. Couldn't we keep them in ConditionDependentNode and then override them only when strictly necessary?

engine/src/main/java/org/terasology/rendering/dag/nodes/LateBlurNode.java
```diff
@@ -67,11 +69,14 @@ protected void setupConditions(Context context) {
         renderingConfig = context.get(Config.class).getRendering();
         renderingConfig.subscribe(RenderingConfig.BLUR_INTENSITY, this);
         requiresCondition(() -> renderingConfig.getBlurIntensity() != 0);
+        worldRenderer = context.get(WorldRenderer.class);
     }
 
     @Override
     public void propertyChange(PropertyChangeEvent event) {
         updateBlurRadius();
+        // Changing the blurRadius can potentially enable/disable the Node, meaning we have to refresh the taskList.
```
emanuele3d: This is not obvious from the code in this method and in updateBlurRadius(). Can you be more explicit on why this is possible? Where exactly is the node disabled?

engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
```diff
@@ -16,53 +16,31 @@
...
         }
-
-        if (conditionsAreSatisfied != isEnabled()) {
-            setEnabled(conditionsAreSatisfied);
-            // enabling/disabling here means we cannot enable/disable nodes directly:
-            // we must always go through the settings.
-            return true;
-        } else {
-            return false;
-        }
```
emanuele3d: I don't remember why @tdgunes and I did it this way. I suspect if we did there was a reason, but if you @vampcat think perhaps that reason is no longer there and this code can be simplified so dramatically, why not!

So, how are nodes enabled/disabled now?

engine/src/main/java/org/terasology/rendering/dag/nodes/BloomBlurNode.java
vampcat: Yes, I was concerned about the code duplication as well. Let's do as you suggest!

vampcat: Oh wait, I just remember why I did not make that change myself.

So, here is the "default" propertyChange:
`worldRenderer.requestTaskListRefresh();`
Obviously, this requires the `worldRenderer` variable. If we store the variable in the class, one of two things happens:

- Unknown to the developer, he inherits CDN *and* creates a worldRenderer variable in his class, meaning two worldRenderer variables exist in the class. Not a big deal functionally, but I can't say I like the redundancy. This is what currently happens, btw.
- The developer is well versed with the system and knows that the `worldRenderer` variable is available to him. However, a new-comer looks at the code and goes "where did this variable come from?", or, he assumes it's a part of Node and tries to use it in non-CD Nodes, causing irritation when it doesn't work.

I'll revert back to having `worldRenderer` in the base class if you want me to, but I feel like we need a more elegant third solution for this.

vampcat:
Okay, so let me explain the old system properly. I'll add this to the PR description as well, to make it clearer to posterity.

Whenever a property was changed, CDN called `checkConditions()`, which updated the "status" of the node (whether it's enabled or disabled), followed by which it called `refreshTaskList`, which only created tasks from enabled Nodes. There were two glaring issues with this:
- If a method decides to override the propertyChange method, and does not explicitly call `checkConditions()`, enabling/disabling any property won't *actually* affect the Node's status, even if it leads to some requirements not being met.
- The only way to enable/disable a CDN was through settings. That means if I have CDN with requirement `bloomEnabled`, I have to toggle bloomEnabled in renderingConfig to toggle the Node's status. I can't just call `setEnabled()` on the Node. Obviously a problem if we want to go for a renderGraph.

My solution:
- Instead of changing the enabled/disabled method statically, we simply check the conditions when we 
- We separate the status of the node from the conditions of the node. That means we can toggle a node on/off independently of the conditions, *but* the final status of the node depends on the user-set state and conditions.

engine/src/main/java/org/terasology/rendering/dag/nodes/LateBlurNode.java
vampcat: Based on the explanation I gave below, I only update the value of blur here. When refreshing task list, the WR will check if this Node is enabled(), which will in turn call checkConditions to ensure all conditions are met, which is where the magic happens.

engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
vampcat: When generating the task list, it already checks `if (node.isEnabled())`. My tweak just makes `isEnabled()` return `is enabled by user AND has all conditions met` for CDN. So, Nodes are enabled/disabled by the user, and their final status is decided by user-set-state and conditions-being-met.

engine/src/main/java/org/terasology/rendering/dag/nodes/BloomBlurNode.java
emanuele3d: I don't think that reasoning is particularly reasonable: if you want to extend a class rather than, say, use an instance of it as a black box, through an interface, it will make sense for you to look into how the original class is written and what fields/methods it provides. 

Even if you decide to "go solo" and replicate some of the private/protected variables, given we'll be dealing with hundreds of nodes -at most-, it will be an example of poor code quality but it won't slow down the application dramatically.

engine/src/main/java/org/terasology/rendering/dag/ConditionDependentNode.java
emanuele3d: Understood. For the time being it's fine. Please let's keep in mind that in the future we'll probably want disabled nodes and/or nodes whose conditions are not met to be completely out of the graph.   This concept is far from fleshed out but it's good to keep it on the backburner I think.

emanuele3d:
Thanks for the clarification @vampcat, very useful. Please deal with my first comment and this PR will be good to go. Speaking of that first comment, let's also write in the javadoc of the `propertyChange` method that overriding classes should eventually run the base class method unless, for whatever reason, they explicitly do not want the tasklist to be refreshed.

GooeyHub:

GooeyHub:

emanuele3d:
Smoke-tested, including some setting changes (i.e. disabling/enabling SSAO) and noticed no problems due to this PR. 

One thing I noticed however is that DoF is broken, also in develop. :disappointed:

PR #3264, opened 2018-02-20
Remove setupConditions

vampcat:
This PR removes the setupConditions used by BlurNode and DownsamplerNode. This was probably required with the NodeFactory approach, but as things stand, it's very easy to move the contents of setupConditions to the constructor itself, which in turn makes the code cleaner and simpler.

This is an experimental move, and feedback would be appreciated.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/BloomBlurNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/BloomBlurNode.java
@@ -45,17 +45,9 @@ public class BloomBlurNode extends BlurNode {
      */
     public BloomBlurNode(Context context, FBO inputFbo, FBO outputFbo) {
         super(context, inputFbo, outputFbo, BLUR_RADIUS);
-    }
 
-    /**
-     * This method establishes the conditions in which the blur will take place, by enabling or disabling the node.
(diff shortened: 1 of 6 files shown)


GooeyHub:

emanuele3d:
Ok, generally speaking this change looks good as it is. You might be right that it might have made sense in the context of the NodeFactory. I don't have the time to smoke-test it right now but I will ASAP. This PR is completely independent from the other two, right?

vampcat:
Yup, completely independent.

vampcat:
The usual way to test this would be to resize the game a ton, change settings and so on. However, if you change settings, it won't work properly for some nodes. This issue is actually coverered in my CND rework PR - since some nodes override the propertyChange and don't explicitly call `checkConditions()`, the "status" of nodes never changes. If you started with node A enabled, it will remain enabled even if you disable the corresponding setting.

So, I would request that while testing this, you either only use resize, or exit to menu and then enter the world again if you change any settings.

emanuele3d:
Smoke-tested, including resizing: it works. Given I just merged the ConditionDependentNodes PR, the issues you mentioned about some nodes not working should no longer take place. It might be worth if you double-check though.

Issue #3131, opened 2017-10-22
Improve user feedback on module dependency resolution issues (Gestalt dependency)


Cervator:
This is somewhat of a companion issue to https://github.com/MovingBlocks/gestalt/issues/15 as Gestalt presently doesn't offer a lot of details on what went wrong when a module ends up lacking some dependency.

One example came up recently via support forum: http://forum.terasology.org/threads/module-dependency-resolution-failed.1908 - that particular case seems to have been a bug in Alpha 8 that have since been fixed or side-stepped, but resulted in a "Module dependency resolution failed" error on *any* attempted download of a module available online but not locally.

At present you can see the minimal handling in `ModuleDownloadListGenerator` for method `processResolutionResult` - all we have at present is a simple true/false to react to. To improve that we'd first need to figure out in Gestalt how to return something more informative, which is what https://github.com/MovingBlocks/gestalt/issues/15 is for, then on the Terasology side present the extended information in various ways depending on the scenario.

Screenie from the earlier mentioned example (can be replicated with the Alpha 8 release):

![2017-10-22 17_19_34-terasology _ alpha](https://user-images.githubusercontent.com/1063833/31866432-3f6f0d4c-b74d-11e7-9211-530c22a2ac48.png)

(Same happens even for modules with no actual dependencies)

PR #3265, opened 2018-02-22
"Create Game" Dependency Checking (#3159)

### LINKED ISSUE #3159: Default / Previous gameplay template remains selected even if dependencies are invalid

Play something like JoshariasSurvival which depends on a bunch of modules and is the default gameplay template selected if you run the game fresh via the Omega zip.

### What actually happened

While JS was selected and you could start the game that way it silently was in fact *not* activated as would be visible if you actually check the Modules section. Only JS itself was highlighted and one or more dependencies were listed as red.

### How to reproduce

* Have a series of modules available with one or more invalid dependencies. For instance increase the current version of a module to the next major version (2.0.0 -> 3.0.0) then anything that depends on it will show as red (cannot satisfy dependencies as a major new release is too new)
* Run the game while the previous gameplay template selection (or default on a first run) is affected by a 
broken dependency. It will have been set without considering the dependency state of the involved modules.
* You'll be allowed to start a game without realizing your selection is invalid. Making it really confusing to be dumped into a seemingly working Perlin world yet none of the JS stuff is active :-)

Instead if the selected gameplay template *cannot* resolve all its dependencies then disable the button that would start the game and perhaps color the selected gameplay template in red text. You could also highlight the Modules button's text in red. The user should be able to pick a different gameplay template and if that one is valid let the start game button activate.

As an extra we could do something similar if there is no valid world generator in that dropdown.

This could probably be turned into a GCI task if there is interest.

morrilet:
### Contains

Close #3159 

If a gameplay module in the "Create Game" screen has invalid dependencies, it will be displayed as red in the dropdown and the play button will be disabled (if it is selected).

### How to test

Change the dependencies of an existing gameplay module such as JoshariasSurvival such that they are no longer valid (change the version, for instance) and try to use the template to create a new game.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
@@ -139,20 +139,18 @@ public class CreateGameScreen extends CoreScreenLayer {
         gameplay.setOptionRenderer(new StringTextRenderer<Module>() {
             @Override
             public String getString(Module value) {
-            	return value.getMetadata().getDisplayName().value();
+                return value.getMetadata().getDisplayName().value();
             }
-            
+
             @Override
-            public void draw(Module value, Canvas canvas)
-            {
-        		canvas.getCurrentStyle().setTextColor(
-        			validateModuleDependencies(value.getId()) 
-        				? Color.WHITE : Color.RED);
-            	super.draw(value, canvas);
-            	canvas.getCurrentStyle().setTextColor(Color.WHITE);
+            public void draw(Module value, Canvas canvas) {
+                canvas.getCurrentStyle()
+                        .setTextColor(validateModuleDependencies(value.getId()) ? Color.WHITE : Color.RED);
+                super.draw(value, canvas);
+                canvas.getCurrentStyle().setTextColor(Color.WHITE);
             }
         });
-        
+
         UILabel gameplayDescription = find("gameplayDescription", UILabel.class);
         gameplayDescription.bindText(new ReadOnlyBinding<String>() {
             @Override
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
```diff
@@ -138,10 +139,20 @@ public void set(Module value) {
...
+            {
+        		canvas.getCurrentStyle().setTextColor(
+        			validateModuleDependencies(value.getId()) 
+        				? Color.WHITE : Color.RED);
+            	super.draw(value, canvas);
+            	canvas.getCurrentStyle().setTextColor(Color.WHITE);
             }
         });
-
+
```
vampcat: Can we please remove the space on this line (and on all others like it)?

Which IDE/text editor do you use? There should be a "trim trailing spaces" or something similar that you can enable in most softwares, to prevent this.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
```diff
@@ -138,10 +139,20 @@ public void set(Module value) {
         gameplay.setOptionRenderer(new StringTextRenderer<Module>() {
             @Override
             public String getString(Module value) {
-                return value.getMetadata().getDisplayName().value();
+            	return value.getMetadata().getDisplayName().value();
```
vampcat: Uhh.. you replaced the spaces just before `return` with a tab.

We strictly use 4 spaces over tabs.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
```diff
@@ -138,10 +139,20 @@ public void set(Module value) {
...
+            }
+            
+            @Override
+            public void draw(Module value, Canvas canvas)
+            {
+        		canvas.getCurrentStyle().setTextColor(
```
vampcat: Again, tab.

Change the default indentation mode in your IDE/text editor to 4 spaces.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/CreateGameScreen.java
```diff
@@ -138,10 +139,20 @@ public void set(Module value) {
...
+        			validateModuleDependencies(value.getId()) 
+        				? Color.WHITE : Color.RED);
+            	super.draw(value, canvas);
```
vampcat: You'll have to reformat the entire file, since it's mixed with tabs and spaces.

morrilet:
@vampcat I think this last commit fixed all the style issues. Set tabs to 4 spaces and formatted the entire file in Eclipse. Thanks for pointing out the IDE settings for this; I was totally unaware.

Cervator:
Hey @morrilet and welcome :-) Going to take this for a spin after triggering our bot:

@GooeyHub: add to whitelist

Yeah IDE settings and code style stuff is an easy one to bump into, there are usually differences between project preferences. I see some comments that got broken to multi-line now for instance, but that might be another IDE setting that hasn't picked up our max line length (far longer than what did get split). Also we try to avoid merges on active PRs, instead favoring rebase. Always lots to learn one way or the other but glad to help you get set up! :-)

GooeyHub:

Cervator:
Works wonderfully! Nice work :-)

I'm adding in a tiny commit just to revert a few line length fixes that weren't needed (often IDE defaults to 120 or even 80, ours is 175 at the moment), that cuts down on the overall diff a little bit more too to make it easier to see the true changes.

Will also add you to the credits and such :-)

If you're curious to take one step deeper in this area you can check out #3131 - with that fixed you'd be able to pinpoint exactly which dependencies are wrong and color those as well.

Issue #3267, opened 2018-02-24
Character can jump once under water


Maximetinu:
### What you were trying to do
While diving, I was trying to surface by pressing spacebar.

### What actually happened
The first time I press the spacebar, a normal jump happens as if I was grounded.

### How to reproduce
* Step 1 - Get into the water
* Step 2 - Press spacebar once to see jump happens

### Log details and game version
Current latest stable: 74 alpha 2018-02-05T04:45:49Z

### Computer details
OS: Win 10 x64 CPU: i5-4690 RAM: 8GB Graphic Card: GeForce GTX 970 but irrelevant as it happens in other computers too

### Video
[Vimeo video showing the bug](https://vimeo.com/257262232)

Maximetinu:
I've already had a look on this, it seems easy to solve.

[This](https://github.com/Maximetinu/Terasology/commit/ea6815ef8502df14da8ef17288d079b2de0ecda5) is the bugfix commit on my fork. As soon as I have your clearance I'll make the PR.

mjuvekar7:
You should go ahead and make a PR. Much easier for us to test if it's a PR. No need to ask for clearance (it's open source!) :)

PR #3269, opened 2018-02-24
Bug Character can jump once under water fixed

Maximetinu:
### Contains
Fixes [#3267](https://github.com/MovingBlocks/Terasology/issues/3267)

### How to test
Jump into the water and check that you're not able to jump, neither diving underwater nor swimming above the surface

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
+++ b/engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
@@ -612,6 +612,11 @@ public class KinematicCharacterMover implements CharacterMover {
             climb(state, input, desiredVelocity);
         }
 
+        // If swimming or diving, cancel double jump to avoid jumping underwater and on the surface
+        if (movementComp.mode == MovementMode.SWIMMING || movementComp.mode == MovementMode.DIVING) {
+            movementComp.numberOfJumpsLeft = 0;
+        }
+
         // Modify velocity towards desired, up to the maximum rate determined by friction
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Hi @Maximetinu and thanks so much for the fix! Just to clarify though, are there are special conditions at all needed to reproduce it? I can't actually replicate #3267 reliably when I'm *in* water, only when I'm at the very top and get the little "help hop" meant to make you have an easier time to get out from water onto solid land. Which admittedly doesn't make sense when you're not up against a block you can jump on to.

I do recall occasional weird jumping-like behavior, I just can't reproduce it to really test this well before applying the fix. After the fix can you still easily get out of water on to dry land with the "help hop" or does that get disabled too?

By chance are you one of the GSOC student candidates and/or do you have a forum account/intro yet?

@GooeyHub: ok to test

GooeyHub:

Maximetinu:
Hi @Cervator, glad to be of help! 

I've been testing and yes, there are a special condition that I have not mentioned. To reproduce it, you must get in the water without jumping, because if you jump just before it starts the swimming/diving mode, then the number of jumps left is set to 0 and it's not possible to jump. But, if you hit the water without jumping, whether by falling or by walking forward, the variable "numbersOfJumpsLeft"is still 1 and you're able to make a jump.

About the "help hop", I am aware that it's not a bug and after the fix it isn't disabled, of course.

And yes, I'm one of the GSoC students interested in working with MovingBlocks. I still don't have a forum account, thus far I've been diving into the code and thinking about the projects, but soon you'll see me over there.

Issue #3272, opened 2018-02-27
Select a game by default if it is the only one in a list.


Naman-sopho:
This is more of an enhancement for better user experience.
If there is only one game present in the saved games list while creating a game `Select Game` screen, this game must be selected by default, so that the game can be started by directly clicking on the `Load` button. For reference: 

![screenshot from 2018-02-26 14-31-13](https://user-images.githubusercontent.com/23008871/36743377-2d9e18f6-1c10-11e8-9bbd-575ee5a733a8.png)
![screenshot from 2018-02-27 23-29-20](https://user-images.githubusercontent.com/23008871/36746261-dc63c2b2-1c17-11e8-95d3-5c70371d3034.png)

This should be resolved for all such lists like single player / multiplayer game selection. 
In case there are multiple games in the list the game at the top must be selected by default, so that the arrow keys can be used to make the desired game selection.

dylanJech:
Hey! I would like to give this enhancement a try as my first contribution to this project. 

After forking my own version, and then creating a branch for this, I looked to see where to create a unit test that tests a method I created that selects the first item. I do not see any tests regarding the UI, however. Is the UI something that has testing written for it for Terasology?

The best place seems to be "engine-tests\src\test\java\org\terasology\rendering\nui\layers\ingame". Would this be an appropriate place to put a class that contains unit tests for this addition?

manas96:
I fixed this in the commit above using an pre-existing method. However, as @J-Dill asked, I would also like to know if a unit test is needed.

Cervator:
Hey both :-)

Generally any time a unit test is *offered* we will be absolutely thrilled! Unit test all the things!

As for required: No, not really, at least not yet. Some things are vastly easier to unit test than others and while it would be nice in those cases to have good test coverage we aren't in super strict code coverage checking land just yet.

Any unit test anybody wants to contribute anywhere even without a related issue would be hugely appreciated and reviewed for sure :-)

Also, I mistakenly referenced this issue in a commit while merging #3271. Off-by-one errors happen to us all ;-)

Closing as fixed by #3275 - thanks @manas96 ! I'm trying to add some more beginner friendly issues so hopefully you can find another interesting one to try @J-Dill, if not or even if feel free to also reach out on chat for any advice or other stuff :-)

PR #3275, opened 2018-03-01
Fix for (beginner) issue #3272.

manas96:
### Contains
My attempt at fixing issue #3272 .
I added a select(0) method to the SelectGameScreen class. It now selects the first saved game(if it exists).

There's no need to verify if index 0 is valid, since index checking is already performed in the select() method.
### How to test
1. Launch Terasology and create a saved world.
2. Go to the SelectGame screen from either 'Singleplayer' or 'Host Game' options in the main menu.
3. The first entry in the list is selected by default.
image


GooeyHub:

Cervator:
Nice! Thanks for adding this. Seems like a small enough change :-)

@GooeyHub: add to whitelist

GooeyHub:

PR #3277, opened 2018-03-02
Fix scrolling when there's already enough space.

4Denthusiast:
### Contains

Scrollbars for ScrollableAreas in spaces large enough that they don't need to scroll are now fixed at 0, rather than flickering between the top and bottom when the scrollwheel is used.

### How to test

Open the video config menu. Resize the window so that there is space left below the menu. Use the mouse scrollwheel.


GooeyHub:

Cervator:
Thanks for the quick fix! :-)

@GooeyHub: add to whitelist

GooeyHub:

Cervator:
Checked it out locally although I had to hack at that settings page a bit before I could even fit it on one screen ... how tall is your resolution :D

Issue #3268, opened 2018-02-24
Remove newline character from the output of the Console.addMessage method.


emanuele3d:
Occasionally, a command executed in the console might take some time, it might not be executed instantaneously. In this situation, it's useful for the console to output a small string before and after the execution of the bulk of the command, i.e.:

`Recompiling shaders... (pause) done!`

The (pause) is a real pause where the software is doing something and the word "done!" hasn't appeared yet. 

This is easy to achieve by framing the bulk of a command with two `Console.addMessage` lines, i.e.:

```
Console.addMessage("Recompiling shaders...");
ShadersManager.recompileShaders();
Console.addMessage("done!");
```

However, the addMessage method also adds a newline character at the end of the output, resulting in:

```
RecompilingShaders...
done!
```

Ideally, any newline character is explicitly added by the caller of the method rather than by default.

vampcat:
Closing, as fixed  by #3268.

PR #3278, opened 2018-03-02
Add method for inline messages (Issue #3268)

ar0ne:
### Contains

Hi All,
I've added new method which we can use to solve an issue #3268. 

### How to test

1. Open Console
2. Write "recompileShaders". So, at present we don't have new line between messages.
=> Recompiling shaders... done.
3. PROFIT :)

### Notes

I'm not sure, will it be useful to add this functionality to ChatScreen widget. If so, let me know...

### DIFF
--- a/engine-tests/src/test/java/org/terasology/logic/console/ConsoleTest.java
+++ b/engine-tests/src/test/java/org/terasology/logic/console/ConsoleTest.java
@@ -24,6 +24,8 @@ import java.util.Iterator;
 
 public class ConsoleTest extends TerasologyTestingEnvironment {
 
+    private final String MSG_TXT = "Test message";
+
     @Test
     public void testClearCommand() {
         for (int i = 0; i < 10; i++) {
@@ -33,6 +35,43 @@ public class ConsoleTest extends TerasologyTestingEnvironment {
         getConsole().clear();
 
         Iterator<Message> it = getConsole().getMessages().iterator();
+        Assert.assertFalse(it.hasNext());
+    }
+
+    @Test
+    public void testAddMessage() {
+        getConsole().addMessage(MSG_TXT);
+
+        checkMessage(getConsole().getMessages().iterator(), true);
+    }
+
+    @Test
+    public void testAddConsoleMessage() {
+        getConsole().addMessage(new Message(MSG_TXT));
+
+        checkMessage(getConsole().getMessages().iterator(), true);
+    }
+
+    @Test
+    public void testAddInlineMessage() {
+        getConsole().addInlineMessage(MSG_TXT);
+
+        checkMessage(getConsole().getMessages().iterator(), false);
+    }
+
+    @Test
+    public void testAddInlineMessage2() {
+        getConsole().addInlineMessage(new Message(MSG_TXT));
+
+        checkMessage(getConsole().getMessages().iterator(), false);
+    }
+
+    private void checkMessage(Iterator<Message> it, boolean hasNewLine) {
+        Assert.assertNotNull(it);
+        Assert.assertTrue(it.hasNext());
+        final Message message = it.next();
+        Assert.assertEquals(MSG_TXT, message.getMessage());
+        Assert.assertEquals(hasNewLine, message.hasNewLine());
         Assert.assertFalse(it.hasNext());
     }
 
--- a/engine-tests/src/test/java/org/terasology/logic/console/ConsoleTest.java
+++ b/engine-tests/src/test/java/org/terasology/logic/console/ConsoleTest.java
@@ -54,14 +54,14 @@ public class ConsoleTest extends TerasologyTestingEnvironment {
 
     @Test
     public void testAddInlineMessage() {
-        getConsole().addInlineMessage(MSG_TXT);
+        getConsole().addMessage(MSG_TXT, false);
 
         checkMessage(getConsole().getMessages().iterator(), false);
     }
 
     @Test
     public void testAddInlineMessage2() {
-        getConsole().addInlineMessage(new Message(MSG_TXT));
+        getConsole().addMessage(new Message(MSG_TXT, false));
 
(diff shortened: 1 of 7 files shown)


GooeyHub:

emanuele3d:
@ar0ne: Serj, welcome aboard and thank you for taking on this task. 

I see that you have added a whole parallel series of methods to handle this issue. Have you considered  adding instead an optional boolean `addNewline` parameter to the existing methods? This way we'd have, for example:


the base method, to be called if addNewline needs to be "false":
```addMessage(String message, boolean addNewline);```

the backward compatible method, internally calling the previous one with addNewline = true:
```addMessage(String message);```

**Comment by ar0ne** [comment]
Hi @emanuele3d , 
You're right, additional methods are excess.

**Comment by GooeyHub** [comment]

**Comment by Cervator** [comment]
Hey @ar0ne welcome back! Bit better timing than last year if you're interested in GSOC again? :-)

Thanks for the extra change on this PR, can confirm the tests work and the console change functions as expected. Merged!

As for the chat widget (and/or notification overlay?) - unsure. Since there generally you're just responding to a single point-in-time event, and could have multiple unrelated messages come in any moment (chat from other players etc). So leaving one message without a newline with the expectation that a follow-up of some sort will finish it isn't immune to interruptions from elsewhere.

**Comment by emanuele3d** [comment]
@Cervator: I realize this is a small PR on a small issue, but I was going to provide some further feedback, especially in terms of documentation. I.e. ConsolImpl now has the following javadoc line:

@param newLine    Flag indicates about new line at the of message

which is not only grammatically incorrect but also not particularly clear - does it add a newline character when this is false or true?

May I suggest that, no matter the size of the PR, you first give a 24 hours warning before merge, so that the people originally involved have the chance to support the merge (even tacitly) or ask for a delay? 

It would be appreciated.

File: engine-tests/src/test/java/org/terasology/logic/console/ConsoleTest.java [hunk 1]
```diff
@@ -24,6 +24,8 @@
 
 public class ConsoleTest extends TerasologyTestingEnvironment {
 
+    private final String MSG_TXT = "Test message";
```
emanuele3d: Let's not abbreviate please: 

`private final String MESSAGE_TEXT = "Test message";`

engine/src/main/java/org/terasology/logic/console/Console.java
```diff
@@ -59,6 +59,22 @@
      */
     void addMessage(Message message);
 
+    /**
+     * Adds a message to the console (as a CoreMessageType.CONSOLE message)
+     *
+     * @param message
+     * @param newLine Flag indicates about new line at the end of message
```
emanuele3d: I'd write:

```
@param message A message as a string
@param newLine A boolean: True causes a newline character to be appended at the end of the message. False doesn't.
```

engine/src/main/java/org/terasology/logic/console/Console.java
```diff
@@ -59,6 +59,22 @@
...
+     */
+    void addMessage(String message, boolean newLine);
+
+    /**
+     * Adds a message to the console (as a CoreMessageType.CONSOLE message)
+     *
+     * @param message
+     * @param type
+     * @param newLine Flag indicates about new line at the end of message
```
emanuele3d: Same as earlier comment. Also, the parameter "type" could use some explanation. What's it for?

engine/src/main/java/org/terasology/logic/console/ConsoleImpl.java
```diff
@@ -117,6 +117,30 @@ private void addErrorMessage(String message) {
         addMessage(new Message(message, CoreMessageType.ERROR));
     }
 
+    /**
+     * Adds a message to the console
+     *
+     * @param message    The message to be added
+     * @param newLine    Flag indicates about new line at the of message
```
emanuele3d: Same as previous comments. I'd also write:

`@param message    The message to be added, as a string.`

engine/src/main/java/org/terasology/logic/console/ConsoleImpl.java
```diff
@@ -117,6 +117,30 @@ private void addErrorMessage(String message) {
...
+    {
+        addMessage(new Message(message, newLine));
+    }
+
+    /**
+     * Adds a message to the console
+     *
+     * @param message    The message to be added
+     * @param type       The type of the message
+     * @param newLine    Flag indicates about new line at the of message
```
emanuele3d: See earlier comments.

engine-tests/src/test/java/org/terasology/logic/console/ConsoleTest.java
ar0ne: I could be wrong, but it's fine to abbreviate constants, e.g. MAX_SPEED. But probably I'm not so close to local code styles. Let's fix it.

emanuele3d: The general idea is to make sure people can read the code without having to de-crypt the code in their head first. Of course plenty of acronyms and abbreviations are widely used, i.e. min/max. But if, as a habit, we avoid abbreviation, our code is closer to plain language and requires fewer brain cycles to deal with, leaving room for more important issues to think about.

ar0ne:
Hi @emanuele3d,
Unfortunately, it's always difficult for me to stop thinking that somebody's doing the same work as me. So, I always try to provide my solution as soon as possible. But it's not a reason to produce bad code. Next time I'll try to find extra time to review my own changes, before ask about review other folks. Sorry for wasting your time.

emanuele3d:
Oh, no worries Serj. Even if we spend time reviewing our own code (which we certainly should always do, carefully), we would miss things. And that's what reviews are for: additional pair of eyes see things from different angles and catch things the author of a piece of code didn't see - not because he or she is a bad programmer, just because two or more people working as a team are much, much better than one person.

PR #3289, opened 2018-03-11
Fix #2246 Volumetric Fog broken

### LINKED ISSUE #2246: Volumetric Fog broken
Terasology Alpha 64
Linux Mint 17-3
Java JDK 8
Installed using the launcher

I played a bit with the graphics, and recognized the following issues.
If you open the graphic-setting while in a server/map, your unpicked/dropped items will not show any texture.
The way to fix is to reload the map, reconnect to the server.
[Screenshot](http://s7.postimg.org/e2ts2bqix/Terasology_160320010345_1920x1080.png)
[Mirror2](http://postimg.org/image/u12hsgkqv/full/)

If you allow Volumetric Fog, the following happens, even after game restart it do not work.
[Screenshot](http://s27.postimg.org/m6g92wcpt/Terasology_160320012657_1920x1080.png)
[Mirror2](http://postimg.org/image/h7sqod8wv/ef2ca543/)
##

4Denthusiast:
I don't fully understand the intention behind volumetric fog or how it's meant to interact with haze, but the obvious issue is fixed and it now looks like fog.

### Contains

Fix #2246. Re-enable menu item for volumetric fog. Add default setting for fog density and things. Remove volFogDensity, which looked like it was intended to make the fog never obscure things more than a certain amount, no matter how far away they are, as that seems to make it less useful for obscuring unloaded chunks.

The main part that resolves the original issue and the other issue this fix reveals is in globalFunctionsFragIncl.glsl .

### How to test

Turn on volumetric fog and turn off inscattering/fog from the video menu. Look around a large open area in-game.

fogDensityAtViewer is fixed at 1, rather than being used.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
@@ -49,7 +49,7 @@ import static org.terasology.rendering.opengl.OpenGLUtils.renderFullscreenQuad;
 
 /**
  * An instance of this class takes advantage of the content of a number of previously filled buffers
- * to add screen-space ambient occlusion (SSAO), outlines, reflections [1], atmospheric haze and volumetric fog [2]
+ * to add screen-space ambient occlusion (SSAO), outlines, reflections [1], atmospheric haze and volumetric fog
  *
  * As this node does not quite use 3D geometry and only relies on 2D sources and a 2D output buffer, it
  * could be argued that, despite its name, it represents the first step of the PostProcessing portion
@@ -58,7 +58,6 @@ import static org.terasology.rendering.opengl.OpenGLUtils.renderFullscreenQuad;
  * has been shot on stage or on location.
  *
  * [1] And refractions? To be verified.
- * [2] Currently not working: the code is there but it is never enabled.
  */
 public class PrePostCompositeNode extends AbstractNode implements PropertyChangeListener {
     private static final ResourceUrn PRE_POST_MATERIAL_URN = new ResourceUrn("engine:prog.prePostComposite");
@@ -97,6 +96,14 @@ public class PrePostCompositeNode extends AbstractNode implements PropertyChange
     @SuppressWarnings("FieldCanBeLocal")
     @Range(min = 0.0f, max = 1.0f)
     private float hazeThreshold = 0.8f;
+    
+    @SuppressWarnings("FieldCanBeLocal")
+    @Range(min = 0.0f, max = 0.1f)
+    private float volFogGlobalDensity = 0.005f;
+    @SuppressWarnings("FieldCanBeLocal")
+    @Range(min = -0.1f, max = 0.1f)
+    private float volFogHeightFalloff = -0.01f;
+    
 
     public PrePostCompositeNode(Context context) {
         worldRenderer = context.get(WorldRenderer.class);
@@ -177,7 +184,7 @@ public class PrePostCompositeNode extends AbstractNode implements PropertyChange
 
         if (volumetricFogIsEnabled) {
             prePostMaterial.setMatrix4("invViewProjMatrix", activeCamera.getInverseViewProjectionMatrix(), true);
-            //TODO: Other parameters and volumetric fog test case is needed
+            prePostMaterial.setFloat3("volumetricFogSettings", 1f, volFogGlobalDensity, volFogHeightFalloff, true);
         }
 
         if (hazeIsEnabled) {
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
@@ -99,11 +99,10 @@ public class PrePostCompositeNode extends AbstractNode implements PropertyChange
     
     @SuppressWarnings("FieldCanBeLocal")
     @Range(min = 0.0f, max = 0.1f)
-    private float volFogGlobalDensity = 0.005f;
+    private float volumetricFogGlobalDensity = 0.005f;
     @SuppressWarnings("FieldCanBeLocal")
     @Range(min = -0.1f, max = 0.1f)
-    private float volFogHeightFalloff = -0.01f;
-    
+    private float volumetricFogHeightFalloff = -0.01f;
 
     public PrePostCompositeNode(Context context) {
         worldRenderer = context.get(WorldRenderer.class);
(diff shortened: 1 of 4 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/PrePostCompositeNode.java
```diff
@@ -97,6 +96,14 @@
...
     @Range(min = 0.0f, max = 1.0f)
     private float hazeThreshold = 0.8f;
+    
+    @SuppressWarnings("FieldCanBeLocal")
+    @Range(min = 0.0f, max = 0.1f)
+    private float volFogGlobalDensity = 0.005f;
+    @SuppressWarnings("FieldCanBeLocal")
+    @Range(min = -0.1f, max = 0.1f)
+    private float volFogHeightFalloff = -0.01f;
+
```
vampcat: Do you mind getting rid of this extra newline? :)

4Denthusiast:
Okay, those variable names were all just the ones already present, but I've renamed them. As worldPosition and worldPositionViewSpace in the shader have the same value anyway, I merged them.

4Denthusiast:
It turns out those variables weren't the same, and I forgot to test before pushing. Oops. That's undone now.

GooeyHub:

GooeyHub:

engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
```diff
@@ -147,16 +147,15 @@ float calcPcfShadowTerm(sampler2D shadowMap, float lightDepth, vec2 texCoord, fl
 float calcVolumetricFog(vec3 fogWorldPosition, float volumetricHeightDensityAtViewer, float globalDensity, float heightFalloff) {
     vec3 cameraToFogWorldPosition = -fogWorldPosition;
 
-    float fogInt = length(cameraToFogWorldPosition) * volumetricHeightDensityAtViewer;
+    float fogTotal = length(cameraToFogWorldPosition) * volumetricHeightDensityAtViewer;
```
vampcat: `fogTotal` sounds a bit strange, especially since a big part of the code-philosophy here is "english-like code". `totalFog` sounds kinda better, but I'm wondering if we can do better.

4Denthusiast: What I mean by it is something like "total amount of fog in that line of sight", although that would also imply globalDensity is included (although actually, maybe it should be). Would "fogAmount" be any better?

vampcat: Think about it this way:
"the fog total is equal to.." vs "the total fog is equal to.."

Yes, fogTotal does indeed sound better.
Ykw? let's wait for manu, he often has better variable name suggestions than me :P

emanuele3d: A few of ideas:
- `totalFogInLineOfSight ` (more colloquial but less precise, my least favorite)
- `totalFogUntilSample` (compact, more precise)
- `totalFogBetweenCameraAndSample` (very precise, verbose, perhaps too much)
- `totalFogFromCameraToSample` (a good compromise?)

We could also drop the total and use something like `fogValueFromCameraToSample`.

vampcat: `fogBetweenCameraAndSample` ?
I kinda agree it's verbose, but it's not the 90s anymore ;)

4Denthusiast: Even if it's not the 90's, I (at least) am likely to gloss over excessively long variable names and find them more effort to read (this probably contributed to me confusing invProjMatrix and invViewProjMatrix). Unless there are further objections, I'm going with "totalFogToSample" ("until" sound too much like something time-related to me). I don't think there are any good options for many of these variable names, but the code is sufficiently simple that it should make their purpose clear anyway.

GooeyHub:

engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
```diff
@@ -147,16 +147,15 @@ float calcPcfShadowTerm(sampler2D shadowMap, float lightDepth, vec2 texCoord, fl
...
-    {
-        float t = heightFalloff * cameraToFogWorldPosition.y;
-        fogInt *= (1.0 - exp(-t)) / t;
+    float heightDensityFactor = heightFalloff * cameraToFogWorldPosition.y;
+    if (abs(heightDensityFactor) > slopeThreshold) {
+        totalFogToSample *= (1.0 - exp(-heightDensityFactor)) / heightDensityFactor;
     }
 
-    return exp(-globalDensity * fogInt);
+    return 1 - exp(-totalFogToSample);
```
emanuele3d: I'm feeling this part of the code could use some tender-loving-care: the math of the individual formulas is easy enough, but the ensemble is not really easy to follow for me, starting from the slopeThreshold constant until the return line.

I mean: I can run Terasology and see that it now does something it didn't do before, but I have no idea if what I see is the intended effect. It almost seems like the intended effect here is some kind of falloff of the fog effect over altitude for example. But I'm not seeing that right now when I toggle the effect. Do I have to go much higher perhaps? I don't know.

This of course is not to say it's your fault @4Denthusiast, it's just code we inherited from previous developers and you just brought it into the spotlight. It's just something the "rendering department" will have to deal with at some point, in the context of the Shaderland Refactoring. 

For the time being I think your PR gets the job done and after some last checks it should be mergeable.

engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
emanuele3d: Over the long variable names: it's certainly a style thing. I do encourage them as long as they make sense to limit ambiguities - obviously they don't always make sense and when possible shorter names are perfectly valid.

emanuele3d:
Ok, I found a bit of a problem testing this: if haze (fog) is enabled in the video settings, all is well, both with and without volumetric fog, as shown in the first screenshot. If haze is disabled, with and without volumetric fog you get the strange effect you see in the second picture - which was obviously taken from the same position and looking in the same direction. 

withhazeandvolumetricfog
withouthazewithvolmetricfog

engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
4Denthusiast: The fog falloff over altitude does work, it's just set to a small value so it's not very noticeable, and the fog density at the viewer's position is fixed even if you move up or down. (That's what fogDensityAtViewer is for.)

4Denthusiast:
I found the same thing previously but assumed it was an existing problem with the reflections (as everything is upside-down). I can't find the right settings to replicate that again though. I'm reasonably confident that if volumetric fog is off, the only difference should be that the checkbox is enabled. Have you tried using the same settings in the previous commit?

emanuele3d:
Doh. You are right. The problem is actually already there in the current develop. I guess I keep haze always on and I never noticed. Ok, then this PR adds something new and doesn't make anything worse: merge ahead!

Issue #3044, opened 2017-08-03
Enable face culling in RenderNodes where it would make sense.


emanuele3d:
It appears that Face Culling is currently disabled for a number of render nodes that should probably have it, i.e. OpaqueBlocksNode - which renders most of the landscape. 

We need to investigate which nodes could take advantage of face culling and add the line

`addDesiredStateChange(new EnableFaceCulling());`

in the constructor of those nodes for which it makes sense.

emanuele3d:
This was tackled by readv in #3291.

PR #3291, opened 2018-03-16
Added back face culling to several nodes

readv:
### Contains
This pull request adds back face culling into nodes that can take advantage of it (#3044).

Back face culling can make rendering more efficient by not rendering triangles that are facing away from the camera. This is very useful for opaque models. The camera can't see through them, which means there's no point in rendering the away-facing triangles. As most of the blocks and objects are opaque, this change should speed up the rendering process.

### Changes:
- Added back face culling to OpaqueBlocksNode, OpaqueObjectsNode and ShadowMapNode.
- Wireframe mode disables back face culling in OpaqueBlocksNode and OpaqueObjectsNode.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
@@ -65,6 +65,7 @@ public class OpaqueBlocksNode extends AbstractNode implements WireframeCapable, 
 
     private Material chunkMaterial;
     private SetWireframe wireframeStateChange;
+    private EnableFaceCulling faceCullingStateChange;
     private RenderingDebugConfig renderingDebugConfig;
 
     private SubmersibleCamera activeCamera;
@@ -86,6 +87,10 @@ public class OpaqueBlocksNode extends AbstractNode implements WireframeCapable, 
         renderQueues = context.get(RenderQueuesHelper.class);
         worldProvider = context.get(WorldProvider.class);
 
+        // Added before instantiating WireframeTrigger so it can remove this state change if needed - see enableWireframe()
+        faceCullingStateChange = new EnableFaceCulling();
+        addDesiredStateChange(faceCullingStateChange);
+
         wireframeStateChange = new SetWireframe(true);
         renderingDebugConfig = context.get(Config.class).getRendering().getDebug();
         new WireframeTrigger(renderingDebugConfig, this);
@@ -95,8 +100,6 @@ public class OpaqueBlocksNode extends AbstractNode implements WireframeCapable, 
         addDesiredStateChange(new LookThrough(activeCamera));
 
         addDesiredStateChange(new BindFbo(context.get(DisplayResolutionDependentFBOs.class).getGBufferPair().getLastUpdatedFbo()));
-
-        addDesiredStateChange(new EnableFaceCulling());
 
         addDesiredStateChange(new EnableMaterial(CHUNK_MATERIAL_URN));
 
@@ -125,39 +128,17 @@ public class OpaqueBlocksNode extends AbstractNode implements WireframeCapable, 
     }
 
     public void enableWireframe() {
-        boolean refreshTaskList = false;
-
-        EnableFaceCulling faceCullingStateChange = new EnableFaceCulling();
-        if (getDesiredStateChanges().contains(faceCullingStateChange)) {
-            removeDesiredStateChange(faceCullingStateChange);
-            refreshTaskList = true;
-        }
-
         if (!getDesiredStateChanges().contains(wireframeStateChange)) {
+            removeDesiredStateChange(faceCullingStateChange);
             addDesiredStateChange(wireframeStateChange);
-            refreshTaskList = true;
-        }
-
-        if (refreshTaskList) {
             worldRenderer.requestTaskListRefresh();
         }
     }
 
     public void disableWireframe() {
-        boolean refreshTaskList = false;
-
-        EnableFaceCulling faceCullingStateChange = new EnableFaceCulling();
-        if (!getDesiredStateChanges().contains(faceCullingStateChange)) {
-            addDesiredStateChange(faceCullingStateChange);
-            refreshTaskList = true;
-        }
-
         if (getDesiredStateChanges().contains(wireframeStateChange)) {
+            addDesiredStateChange(faceCullingStateChange);
             removeDesiredStateChange(wireframeStateChange);
-            refreshTaskList = true;
-        }
-
-        if (refreshTaskList) {
             worldRenderer.requestTaskListRefresh();
         }
     }
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
@@ -31,6 +31,7 @@ import org.terasology.rendering.dag.StateChange;
 import org.terasology.rendering.dag.WireframeCapable;
 import org.terasology.rendering.dag.WireframeTrigger;
 import org.terasology.rendering.dag.stateChanges.BindFbo;
+import org.terasology.rendering.dag.stateChanges.EnableFaceCulling;
 import org.terasology.rendering.dag.stateChanges.EnableMaterial;
 import org.terasology.rendering.dag.stateChanges.LookThrough;
 import org.terasology.rendering.dag.stateChanges.SetInputTexture2D;
@@ -95,6 +96,8 @@ public class OpaqueBlocksNode extends AbstractNode implements WireframeCapable, 
 
         addDesiredStateChange(new BindFbo(context.get(DisplayResolutionDependentFBOs.class).getGBufferPair().getLastUpdatedFbo()));
 
+        addDesiredStateChange(new EnableFaceCulling());
+
         addDesiredStateChange(new EnableMaterial(CHUNK_MATERIAL_URN));
 
         chunkMaterial = getMaterial(CHUNK_MATERIAL_URN);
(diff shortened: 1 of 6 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
```diff
@@ -122,15 +125,39 @@ public OpaqueBlocksNode(Context context) {
     }
 
     public void enableWireframe() {
+        boolean refreshTaskList = false;
+
+        EnableFaceCulling faceCullingStateChange = new EnableFaceCulling();
+        if (getDesiredStateChanges().contains(faceCullingStateChange)) {
```
emanuele3d: Because EnableFaceCulling is added as a desired state in the constructor you can rely on the fact that when you enable wireframe it will be there. So, you don't need the IF statement to check for its presence. 

Also notice that in this case the IF statement would always evaluate to false, as you are creating a new EnableFaceCulling state change every time.

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
```diff
@@ -122,15 +125,39 @@ public OpaqueBlocksNode(Context context) {
...
+        if (refreshTaskList) {
             worldRenderer.requestTaskListRefresh();
         }
     }
 
     public void disableWireframe() {
+        boolean refreshTaskList = false;
+
+        EnableFaceCulling faceCullingStateChange = new EnableFaceCulling();
+        if (!getDesiredStateChanges().contains(faceCullingStateChange)) {
```
emanuele3d: Similar comment to the one above: when you disable wireframe you can rely on the EnableFaceCulling state change to NOT be in the desired set because it has been removed by the enableWireframe method.

The only potential reason to do this here is if the wireframe mode is persisted between sessions. But in that case I'd put the IF in the constructor, not here. Can you verify?

Also similarly to the comment above, the IF block will always evaluate to true because the content of `faceCullingStateChange` is regenerated every time this method is invoked.

emanuele3d: Also notice OpaqueObjectsNode.java also needs the same treatment.

readv:
Thanks for the review @manu3D!

In the constructor, the `EnableFaceCulling` is added as a desired state change only after `WireframeTrigger` has already been instantiated. This means that when wireframe mode is enabled from the last session, the `enableWireframe()` will be called before desired state changes were added and try to remove non existent state change [*]. 

1. Is removing a non existent state change considered problematic? ([Set.remove() expects it can happen](https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#remove(java.lang.Object)))
If it is problematic, WireframeTrigger instantiation can be moved so that it's called after adding EnableFaceCulling state change.
2. Since `desiredStateChanges` is a `Set`, is adding existing state changes problematic?
   Since `new EnableFaceCulling()` is a new object, I'd expect inserting a second new `EnableFaceCulling` to the set would result in a "duplicate" (set would contain both, as they're not the same object), but after debugging the code, that doesn't happen.
   The same goes for the IF statements, I understand why you would expect them to be always true/false, but that does not happen when I run the code.

   Am I overlooking something? Would it be preferable to make a `faceCullingStateChange` member of the node (same as `wireframeStateChange`) and use it to add/remove state changes?

If removing non existent state changes and adding already exiting ones is alright, could we drop all the conditions (even the `wireframeStateChange` one) in `enable/disableWireframe`?

> The only potential reason to do this here is if the wireframe mode is persisted between sessions. But in that case I'd put the IF in the constructor, not here. Can you verify?

There seem to be two options (both in the constructor):
- Enable culling only if wireframe mode is off in the rendering debug config.
- Move WireframeTrigger instantiation below (all) state changes, so that enableWireframe() has the final say over what will be the desired culling state, if it is supposed to be enabled since last session.

[*] I found a bug that will crash the game as a world is loading when wireframe mode is enabled from the last session. The cause is calling `enableWireframe()` (because of `WireframeTrigger`) before worldRenderer is initialized, resulting in Null Pointer Exception. Moving `worldRenderer` initialization above the `WireframeTrigger` instantiation should fix it.

emanuele3d:
@readv, thank you for your feedback. I recommend you respond directly to the comment I write rather than back on the main thread of this PR. Also try to split the comments so that they are as close as possible to the lines of code they relate to. 

For example, I can see the logic of the bug you found: in this circumstances it'd be good if you write a comment below the line that causes trouble, so that we can discuss each issue in isolation - if possible.

emanuele3d:
Regarding the main issues in your reply, @vampcat and I have had some difficulties in fully understanding what you write. 

For example, you write:
> This means that when wireframe mode is enabled from the last session, the enableWireframe() will be called before desired state changes were added and try to remove non existent state change [*].

But enableWireframe() doesn't try to remove anything. It -adds- a state change and only if it's not there already. So, I'm not sure what you meant there. 

You also mention the issue of removing states that are not there and adding duplicate states: both things will not crash the program as the Set class indeed doesn't return an exception in either cases. Set simply returns false if you attempt to remove a state that isn't there and adds any object that isn't there already. 

However, I don't see why the code should get into those situation. Basically, this is the way I think it should work:

In the contructor:
- initialize a faceCullingStateChange member variable
- add faceCullingStateChange to the desired set IF wireframeStateChange is not in the desired set.

In enableWireframe: 
- if wireframeStateChange is *not* in the desired set, add it and remove faceCullingStateChange 

In disableWireframe:
- if wireframeStateChange is in the desired set, remove it and add faceCullingStateChange back in. 

Does this makes sense?

readv:
Yes it does, thank you.

When I mentioned `enableWireframe()` removing state change, I meant my commited version of it, which removed `EnableFaceCulling`.

As we've talked about this on IRC, I'll commit the new agreed upon version.

readv:
There's an issue I've noticed.

Sometimes, after the game is loaded, only one of `OpaqueObjectsNode`/`OpaqueBlocksNode` has correctly disabled culling when in wireframe mode. I've tried to debug this. The problem appears even though `enableWireframe()` and `process()` work as they should - `EnableFaceCulling` is NOT present and `SetWireframe` is.

The newly introduced code does not link `OpaqueObjectsNode` and `OpaqueBlocksNode` in any way, and they both handle wireframe mode equally.

I'll try to work this out soon.

engine/src/main/java/org/terasology/rendering/dag/nodes/OpaqueBlocksNode.java
```diff
@@ -86,6 +87,10 @@ public OpaqueBlocksNode(Context context) {
         renderQueues = context.get(RenderQueuesHelper.class);
         worldProvider = context.get(WorldProvider.class);
 
+        // Added before instantiating WireframeTrigger so it can remove this state change if needed - see enableWireframe()
```
emanuele3d: Let's rewrite this as:

```
// IF wireframe is enabled the WireframeTrigger will remove the face culling state change 
// from the set of desired state changes.
// The alternative would have been to check here first if wireframe mode is enabled and *if not*
// add the face culling state change. However, if wireframe *is* enabled, the WireframeTrigger
// would attempt to remove the face culling state even though it isn't there, relying on the
// quiet behaviour of Set.remove(nonExistentItem). We therefore favored the first solution.
```

And please remember to add the same text to `OpaqueObjectNode`!

emanuele3d:
The changes from commit #93c05ed look good! Let's sort out the issue with the face culling state change in the wrong place and then we'll be ready to merge!

readv:
The problem was `ChunkMesh` explicitly calling `glDisable(GL_CULL_FACE)`, then `glEnable(GL_CULL_FACE)` in `ChunkMesh.render(ALPHA_REJECT)`. This function was called from `AlphaRejectBlocksNode`, so nodes that followed it had their culling enabled even though they didn't request face culling state change.

Since `AlphaRejectBlocksNode` guarantees that face culling is disabled by its choice of state changes, these explicit calls are pointless, so I will remove them in the next commit.

This fixes the problem.

engine/src/main/java/org/terasology/rendering/primitives/ChunkMesh.java
```diff
@@ -275,9 +275,7 @@ private void render(RenderPhase type) {
                 break;
             case ALPHA_REJECT:
                 renderVbo(1);
-                glDisable(GL_CULL_FACE);
                 renderVbo(2);
-                glEnable(GL_CULL_FACE);
```
emanuele3d: Ok, first of all, I can see how this is causing the problem you mentioned @readv. This is the kind of thing that should be done only via StateChange objects rather than direct glCalls. And if glCalls are really needed they should leave the opengl state as they found it.

Now, from the code it seems that renderVbo(1) (translucent blocks I think) is meant to have face culling ENabled. So, _theoretically_, rather than removing those GL_CULL_FACE lines, you could move the glEnable line over the renderVbo(1) line. 

In practice, I think translucent blocks are for things such as leaves. And if that's the case we should see the back face of a volume of leaves through the front face. So, even in that case I feel face culling should be disabled. 

In short, I think you did the right thing. :smile: 

I'll do some testing tomorrow morning and if all is well we should be able to merge it.

readv: That was my line of thinking as well. I had a suspicition that leaves would be rendered incorrectly with the old code, but I couldn't find any translucent leaves block. Today I downloaded JoshariasSurvival module and indeed, the berry bushes were rendered incorrectly. With the new code, they are rendered correctly. Glad to see this helped fix more than just the wireframe mode :)

emanuele3d:
Ok, I tested this and on one hand everything works just like in develop. On the other hand I was expecting to see translucent blocks, i.e. ice and water -without- back culling. Instead the back faces do not seem to be visible at all, as if they are culled.  

You can try this by opening the in-game console (F1) and then typing "give ice". Just build up a small structure and go around it to see that 1) the blocks are semi-transparent, yes, but 2) the back faces are not visible. 

Do feel like doing some additional detective work @readv?

readv:
I'll definitely look into it, but I don't think it has to do with back face culling.

As you said, when looking through ice/water, the back faces are occluded. But it's not just the back faces. Any transparent block, viewed through another transparent block, is completely occluded. For example, you can't see ice below water, or water through ice.

My guess would be some undesirable depth checking when rendering transparent blocks, but I haven't gone through the code yet.

Apparently this problem has been present for a long time (before this PR), so it seems back face culling isn't the problem. Shouldn't this be treated as a separate issue?

emanuele3d:
Good points @readv. Let's treat it as separate then: this PR already improves things (although I haven't checked for performance differences) and doesn't make anything worse - let's merge it!

Issue #3050, opened 2017-08-05
Unexpected behavior on a filtered module screen + activating a module there


Cervator:
### What you were trying to do

Activate the InGameHelpAPI module on top of the standard content of the JoshariasSurvival gameplay template, from a filtered module view for "help" that showed other already active modules. I expected to simply see my additional module showing as activated.

### What actually happened

When I activated "In Game Help API" with two other visible modules already activated as standard dependencies of JoshariasSurvival the other two modules toggled to display as if they weren't activated (whether or not they actually deactivated is another question - it may well just be a display issue)

### How to reproduce

* Launch the game with the JoshariasSurvival modules present along with "InGameHelpAPI" - _or_ possibly any other set of modules where you can filter to a subset and see both active and inactive modules (untested).
* Select the JoshariasSurvival gameplay template on the Create Game screen
* Enter "help" in the filter box on the modules screen - should limited the list to "In Game Help", "In Game Help API", and "WorkstationInGameHelp" with the API module inactive and the other two active
* Double click "In Game Help API" and witness the other two modules seemingly turn themselves off.

### Suspicion

Somewhere in the logic/state behind the "Select Modules" screen something isn't flipped or cleaned up right when a filter is active. May always be the case or be specific to certain modules / dependency setups.

### Now you see it ...

![2017-08-05 02_32_05-terasology _ alpha](https://user-images.githubusercontent.com/1063833/28993552-807b84ea-7987-11e7-9b47-d9b107994151.png)

### Now you don't!

![2017-08-05 02_39_17-terasology _ alpha](https://user-images.githubusercontent.com/1063833/28993554-87469b84-7987-11e7-918e-bdb8e037f367.png)

grubdragon:
I'd like to work on this issue

grubdragon:
@Cervator I can't seem to reproduce this error
On the create game screen, I filtered "help" and deactivated the other two. Proceeded to double click In Game Help API, but nothing seemed to happen

ar0ne:
![ezgif-2-4e1c668623](https://user-images.githubusercontent.com/3996860/37602607-bf6a6a4a-2b9d-11e8-9d10-89f671d03f42.gif)
Hey @Cervator, I guess I was able to reproduce the issue. Will dig in into that.

grubdragon:
Hey, I was working on this issue a long time back, but lost track of the
project because of my academics. As far as I remember, @Cervator speculated
that the issue might be something related to clearing the preferences. All
[...]

PR #3296, opened 2018-03-19
Fix for filtered modules in SelectModulesScreen [Issue #3050]

ar0ne:
### Contains

Small fix for the Issue #3050. 
The issue was that _getExplicitlySelectedModules_() method use _sortedModules_, and _filterText()_ method do the same. In this case, when we call _refreshSelection()_ it works with "filtered/searched" modules which might be are not explicitly selected. 

### How to test

1. Create new game
2. Activate **JoshariasSurvival** module.
3. Search "help". You should get list from 3 items (In Game Help, In Game Help API, and WorkstationInGameHelp)
4. Activate and deactivate "In Game Help API". 
5. Check that search result is still yellow and presented.


GooeyHub:

Cervator:
Tests out! Thanks so much :-)

PR #3297, opened 2018-03-19
Block Sound Effects Overhaul

skate54345:
### Contains

- Created appropriate sound effects for all "Core" blocks for digging and destroying
- Implemented blocksounds for implementation of these effects in game
- Fixes latency and level issue with land.ogg sound under "engine"

### How to test

Spawn various blocks and dig/destroy with audio output set up in testing environment.  Some destroy sounds are very subtly different than the dig, but gravel, leaf, sand, snow, stone and wood will all be noticeably upgraded for both files.


GooeyHub:

vampcat:
@GooeyHub: ok to test

vampcat:
@GooeyHub: ok to test

vampcat:
@GooeyHub ok to test

GooeyHub:

Cervator:
Hey @skate54345 thanks for submitting this :-)

One important question we always need to ask: where did the sfx actually come from? I figure you might have made them yourself, just need to know for sure for attribution and such (any artwork will end up under CC BY 4.0 here, so we need to be sure it is license compatible / approved)

For your outstanding checklist - is it just not fully edited/removed? I don't see how any translation work is needed, for instance. Even if footstep noises aren't exhaustively completed by this PR isn't it still mergeable before adding more? 

Usually those kinds of checks *must* be done before merge, or should be considered, but if they don't block the merge they just might be nice extras :-)

Also: I think you might have two different emails associated with your commits / GitHub account somehow (your commits are authored and committed by two different Austins)

skate54345:
Hi!  I have created the sounds myself.  Most of them have the first few
milliseconds of the original dig noise overlayed to have some consistency
with the impact part.  A couple, such as the wood chopping, also implement
sound effects downloaded from freesound.com but those are almost completely
unrecognizable from different pitch and length changes and the fact that
theyre on top of other sounds.

This is actually my first time contributing to a large project on github so
i misunderstood the default checklist and added the footsteps thing almost
as something I would "like to do next" The outstanding list can mostly be
ignored, but I will go into it now and change what I need beforehand. Would
the wiki need to be changed in any way for this contribution?

As for the naming issue, I noticed that too and both users are me.  I
usually use the bash shell within windows but switched to the normal cmd
when dealing with some of the bash issues loading linux graphics through
it.  I have a feeling that has something to do with it, as both have
different local account names.

skate54345:
@Cervator this should be good to go now

modules/Core/assets/blockSounds/gravel.blocksounds
```diff
@@ -1,5 +1,5 @@
 {
     "stepSounds": [],
-    "destroySounds": ["engine:RemoveBlock"],
-    "digSounds": ["engine:Dig"]
+    "destroySounds": ["engine:GravelDestroy"],
```
Cervator: Forgot to switch to "Core" twice here :-)

Sounds fine when fixed tho!

modules/Core/assets/blockSounds/leaf.blocksounds
```diff
@@ -0,0 +1,5 @@
+{
+    "stepSounds": [],
+    "destroySounds": ["core:Leaves"],
```
Cervator: Should this (and the associated ogg) be named "leaf" instead of "leaves" ? I"m not sure what would be more consistent

modules/Core/assets/blockSounds/snow.blocksounds
```diff
@@ -1,5 +1,5 @@
 {
     "stepSounds": [],
-    "destroySounds": ["engine:RemoveBlock"],
-    "digSounds": ["engine:Dig"]
+    "destroySounds": ["core:Snow"],
+    "digSounds": ["core:SnowDestroy"]
```
Cervator: There is no "SnowDestroy" file?

Cervator:
Oh, and as far as the wiki being updated, what did you have in mind?

skate54345:
Great. I will get right to that when I get a chance! Should be some really
quick fixes.  I didn't change the sound for Land.ogg i just edited the
overall volume level and fixed the 20ish millisecond delay of it playing
after the player hits the ground.  It's also worth noting that the plant
sounds don't register when using the Advanced Plant Pack module (forget the
name but something like that). This is likely because it doesn't assign
each new block to the right categories in the json file to be qualified
under "Plants" for blocksounds.  Thats probably one of the ones you had
issues with and its another thing im hoping to get working on soon.
Thanks for the feedback!

GooeyHub:

PR #3299, opened 2018-03-23
Re-activate removal of underwater surfaces from reflections.

4Denthusiast:
The code for doing this all already exists, but was inactive, leading to surfaces below the water's surface being reflected above it. This was disabled in 296558, for unclear reasons, which makes me suspicious there's something I'm missing, but the result does seem to look better.

### How to test

Activate global water reflections. View a water surface from a reasonably high angle, with well-lit surfaces beneath the water. Before this commit, they're reflected and obscure the correct reflections. After, they are omitted.
Before:
incorrect reflections
After:
correct reflections

~~In Builder-sample mode, this causes nothing below sea-level to render.~~

### DIFF
--- a/engine/src/main/java/org/terasology/engine/modes/loadProcesses/InitialiseWorldGenerator.java
+++ b/engine/src/main/java/org/terasology/engine/modes/loadProcesses/InitialiseWorldGenerator.java
@@ -46,7 +46,7 @@ public class InitialiseWorldGenerator extends SingleStepLoadProcess {
         worldGenerator.initialize();
 
         WorldRenderer worldRenderer = context.get(WorldRenderer.class);
-        worldRenderer.getActiveCamera().setReflectionHeight(worldGenerator.getWorld().getSeaLevel());
+        worldRenderer.getActiveCamera().setReflectionHeight(worldGenerator.getWorld().getSeaLevel() + 0.5f);
 
         return true;
     }
--- a/engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
+++ b/engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
@@ -847,7 +847,7 @@ public class NetworkSystemImpl implements EntityChangeSubscriber, NetworkSystem 
         }
         WorldGenerator worldGen = context.get(WorldGenerator.class);
         if (worldGen != null) {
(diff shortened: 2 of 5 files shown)


4Denthusiast:
Okay, now I've worked out why this doesn't work: if reflections aren't being used, nothing below sea-level renders. I'll try fixing it later.

emanuele3d:
@4Denthusiast: just the fact that you managed to 1) reconnect the problem of strange reflections to those two lines of code and 2) connect the same lines of code to the disappearance of the underwater landscape is an EXCELLENT contribution. 

Those were issues that have been bothering vampcat and I for a while but we just postponed to the bigger shaderland refactoring work.

I'd suggest to keep this PR open (or re-open it when you are ready) and just flag up in its description that it's not for merge. Perhaps as soon as any of us three have a moment to really dig into it we'll find how to make everything work, both above and underwater.

4Denthusiast:
Okay, I think all that needs to be done is separately tell the chunk shader whether it's currently rendering reflections separately from FEATURE_REFRACTIVE_PASS, or possibly just fix builder-sample mode rendering so that FEATURE_REFRACTIVE_PASS is set (I'm unsure what exactly it's for), then use that to determine whether to apply clipping. I'll be away for a while from tomorrow, so probably won't have time to do this myself by then, but you're welcome to do so of course.

4Denthusiast:
While trying to work our stuff related to this, I also discovered that the reflection level was incorrect, being at the origin (centre) of the water blocks at sea-level, rather than the top. While this is not entirely related,  it felt relevant enough to include. It's easier to test if you turn off water ripples (e.g. by setting normalWaterOffset to 0 around line 139 of chunk_frag.glsl).

GooeyHub:

4Denthusiast:
The bug appeared in builder-sample mode simply because that world generator has parts of the surface below height 0. It also occurred elsewhere, but was less apparent. Now the clipping should only occur for the WorldReflectionNode, and not the OpaqueBlocksNode. This should now be ready for merging, if there isn't anything I've missed.

GooeyHub:

Cervator:
Thanks for the quick double-check @vampcat whom I totally didn't coerce into checking real quick after noticing online so I could have something to merge along with a credits update!

In other totally unrelated news I've added you to the game credits @4Denthusiast along with @Shade254 for a different PR. Thanks all for the good work :-)

Issue #3295, opened 2018-03-17
Terasology crashes when trying to connect to server with no loaded game


Shade254:
### What you were trying to do
I was running Terasology server at local instance of docker.  I launched game client and tried to join in "Custom Game" menu.

### What actually happened
After clicking localhost option in "Custom game menu" game instantely crashed, with standard error form.

### How to reproduce
* Step 1
Launch game server with no running game.
* Step 2
Try to connect in Terasology client.

### Log details and game version
https://pastebin.com/JQfMhPPJ 74-alpha

### Computer details
64-bit "Linux Mint 18 Sarah" (Ubuntu 16.04)

PR #3302, opened 2018-03-25
Null checks on server response on JoinGameScreen

Shade254:
### Contains
Few lines of code in JoinGameScreen.java, checking for null response from server, fixing #3295 "

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
@@ -416,7 +416,12 @@ public class JoinGameScreen extends CoreScreenLayer {
     private String getWorldText(Future<ServerInfoMessage> info) {
         try {
             List<String> codedWorldInfo = new ArrayList<>();
-            for (WorldInfo wi : info.get().getWorldInfoList()) {
+            ServerInfoMessage serverInfoMessage = info.get();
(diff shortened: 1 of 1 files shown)


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
```diff
@@ -416,7 +416,12 @@ public String get() {
     private String getWorldText(Future<ServerInfoMessage> info) {
         try {
             List<String> codedWorldInfo = new ArrayList<>();
-            for (WorldInfo wi : info.get().getWorldInfoList()) {
+            ServerInfoMessage serverInfoMessage = info.get();
+
+            if(serverInfoMessage == null)
+                return FontColor.getColored(translationSystem.translate("${engine:menu#connection-failed}"), Color.RED);
```
vampcat: Please put brackets around the body of if.

Here at Terasology, we use brackets around if/while even if it's a single line.

Shade254: should be fine now
I also didnt write catch clause for this, because it could catch some more general exceptions. Hope it is right like that

GooeyHub:

GooeyHub:

PR #3308, opened 2018-04-05
Fix parallax and normal mapping

### LINKED ISSUE #2418: Graphic artifacts possible after changing video settings in-game

4Denthusiast:
### Contains

This PR:
 1. Corrects the formulae used in the shaders to calculate parallax and normal maps.
 1. Restricts parallax maps to not fetch colours from beyond the current texture.
 1. Alters the default setting for parallax maps to improve their appearance.
 1. Enables the use of parallax mapping without normal mapping.

Parts 2 and 3 resolve #2418, and part 4 deals with some of the confusingness of the settings discussed in that issue.

The branch 4Denthusiast/Terasology:parallax uses a more realistic version but looks worse due to parallax mapping being unable to extend the block's rendering beyond its original size. This branch is a compromise which makes the block appear flatter at shallow angles.

### How to test

Activate parallax mapping in the video menu. Activating extra lighting as well should no longer be necessary.

Place some core:depthTest or core:ironBar blocks, and look at them from various angles.

To see that the parallax calculations are now more correct, it's easier to compare a094555 to develop, rather than the final version. In that version, it looks as though every block is a little window through onto the texture atlas. (This is more apparent in worlds with large-ish texture atlases i.e. many blocks. I used coreSampleGameplay.) This behaviour is correct (except the issues fixed in later commits), as it corresponds to realistic parallax. The current version in develop, by contrast, changes its appearance depending on camera angle, which is clearly incorrect.

- [ ] The effect of normal maps are more subtle, so I haven't managed to test that they're actually working properly.
- [ ] I'm not sure if it's appropriate to leave gaps of unused texture indices, or whether it would be better to reduce the index of textureAtlasHeight if textureAtlasNormal isn't used (e.g. WorldRelfectionNode line 179).
- [ ] I'd like some more opinions on whether 466a1d7 or 034b678 (realistic but more distorted, or compromise) is better.
- [ ] Is there any way to get linear interpolation for textures that's actually linear everywhere, rather than flat near the defined points? Setting textureAtlasHeight to use that would be better if possible.

### DIFF
--- a/engine/src/main/resources/assets/shaders/chunk_frag.glsl
+++ b/engine/src/main/resources/assets/shaders/chunk_frag.glsl
@@ -76,6 +76,12 @@ uniform sampler2D textureLava;
 
 uniform float clip;
 
+//inverse is not available in GLSL 1.20, so calculate it manually
+mat2 inverse2(mat2 m) {
+    float det = m[0][0] * m[1][1] - m[1][0] * m[0][1];
+    return mat2(m[1][1], -m[0][1], -m[1][0], m[0][0]) / det;
+}
+
 void main() {
 
 // Active for worldReflectionNode only.
@@ -95,34 +101,29 @@ void main() {
 #if defined (NORMAL_MAPPING) || defined (PARALLAX_MAPPING)
     // TODO: Calculates the tangent frame on the fly - this is absurdly costly... But storing
     // the tangent for each vertex in the chunk VBO might be not the best idea either.
-    // Screen-space coordinates happen to be a coordinate system to which it's possible to relate both view-space and texture coordinates, but are themselves irrelevant.
+    // The specific relationship between screen coordinates and view coordinates is irrelevant here. Screen coordinates just happen to be a basis that it's possible to relate both view and UV space coordinates to.
     mat2x3 screenToView = mat2x3(dFdx(vertexViewPos.xyz), dFdy(vertexViewPos.xyz));
     mat2   screenToUv   = mat2  (dFdx(gl_TexCoord[0].xy), dFdy(gl_TexCoord[0].xy)) / TEXTURE_OFFSET;
-    //mat2 uvToScreem = inverse(screenToUv);
-    //inverse is not available in GLSL 1.20, so calculate it manually:
-        float det = screenToUv[0][0] * screenToUv[1][1] - screenToUv[1][0] * screenToUv[0][1];
-        mat2 uvToScreen = mat2(vec2(screenToUv[1][1],-screenToUv[0][1]),vec2(-screenToUv[1][0],screenToUv[0][0])) / det;
+    mat2 uvToScreen = inverse2(screenToUv);
     mat2x3 uvToView = screenToView * uvToScreen;
-    mat3 tbn = mat3(uvToView[0], uvToView[1], normal);
 
 #if defined (PARALLAX_MAPPING)
-    vec3 eyeTangentSpace = normalizedVPos * tbn;
+    vec2 viewDirectionTextureOffset = normalizedVPos * uvToView;
 
-    float height =  parallaxScale * texture2D(textureAtlasHeight, texCoord).r - parallaxBias;
-    //Ideally this should be divided by eyeTangentSpace.z, but in practice this looks better at
-    //low angles, as it can't extend beyond its original space.
-	texCoord += height * eyeTangentSpace.xy * TEXTURE_OFFSET;
+    float height = parallaxScale * texture2D(textureAtlasHeight, texCoord).r - parallaxBias;
+    //Ideally this should be divided by dot(normal, normalizedVPos), as the texture offset is the amount the light-ray travels in across the texture in the time it takes to traverse the distance "height" perpendicular to the surface, but in practice this way looks better at low angles, as the height-map can't make the triangle extend beyond its normal size.
+	texCoord += height * viewDirectionTextureOffset * TEXTURE_OFFSET;
 	
 	//Crudely prevent the parallax from extending to other textures in the same atlas.
 	vec2 texCorner = floor(gl_TexCoord[0].xy/TEXTURE_OFFSET)*TEXTURE_OFFSET;
-	vec2 texSize = vec2(1,1)*TEXTURE_OFFSET*0.9999; //Remain strictly this side of the edge.
+	vec2 texSize = vec2(1,1)*TEXTURE_OFFSET*0.9999; //Remain strictly this side of the edge of the texture.
 	texCoord = clamp(texCoord, texCorner, texCorner + texSize);
 #endif
 #if defined (NORMAL_MAPPING)
-    //Normalised but not orthonormalised. It should usually be orthogonal anyway, but it's not obvious what's the best thing to do when it isn't.
-    mat3 tbnNormalised = mat3(normalize(uvToView[0]), normalize(uvToView[1]), normal);
+    //Normalised but not orthonormalised. It should be orthogonal anyway (except for some non-rectangular block shapes like torches), but it's not obvious what's the best thing to do when it isn't.
+    mat3 extendedTangentSpaceToViewSpace = mat3(normalize(uvToView[0]), normalize(uvToView[1]), normal);
     normalOpaque = normalize(texture2D(textureAtlasNormal, texCoord).xyz * 2.0 - 1.0);
-    normalOpaque = normalize(tbnNormalised * normalOpaque);
+    normalOpaque = normalize(extendedTangentSpaceToViewSpace * normalOpaque);
 
     shininess = texture2D(textureAtlasNormal, texCoord).w;
 #endif
--- a/engine/src/main/resources/assets/shaders/chunk_frag.glsl
+++ b/engine/src/main/resources/assets/shaders/chunk_frag.glsl
@@ -86,9 +86,9 @@ void main() {
 
 // Active for worldReflectionNode only.
 #if defined FEATURE_USE_FORWARD_LIGHTING
-	if (vertexWorldPos.y < clip) {
+    if (vertexWorldPos.y < clip) {
         discard;
-	}
+    }
 #endif
 
     vec2 texCoord = gl_TexCoord[0].xy;
@@ -113,13 +113,17 @@ void main() {
     vec2 viewDirectionUvProjection = -normalizedViewPos * uvToView;
 
     float height = parallaxScale * texture2D(textureAtlasHeight, texCoord).r - parallaxBias;
-    //Ideally this should be divided by dot(normal, normalizedViewPos), as the offset for texCoord is the component parallel to the surface of a vector along the view's forward axis, the other component being a vector perpendicular to the surface and having magnitude "height". In practice the current way looks better at low angles, as the height-map can't make the triangle protrude beyond its boundaries like displacement mapping would.
-	texCoord += height * viewDirectionUvProjection * TEXTURE_OFFSET;
+    // Ideally this should be divided by dot(normal, normalizedViewPos), as the offset for texCoord 
+    // is the component parallel to the surface of a vector along the view's forward axis, 
+    // the other component being a vector perpendicular to the surface and having magnitude "height". 
+    // In practice the current way looks better at low angles, as the height-map can't make the triangle 
+    // protrude beyond its boundaries like displacement mapping would.
+    texCoord += height * viewDirectionUvProjection * TEXTURE_OFFSET;
 	
-	//Crudely prevent the parallax from extending to other textures in the same atlas.
-	vec2 texCorner = floor(gl_TexCoord[0].xy/TEXTURE_OFFSET)*TEXTURE_OFFSET;
-	vec2 texSize = vec2(1,1)*TEXTURE_OFFSET*0.9999; //Remain strictly this side of the edge of the texture.
-	texCoord = clamp(texCoord, texCorner, texCorner + texSize);
+    //Crudely prevent the parallax from extending to other textures in the same atlas.
+    vec2 texCorner = floor(gl_TexCoord[0].xy/TEXTURE_OFFSET)*TEXTURE_OFFSET;
+    vec2 texSize = vec2(1,1)*TEXTURE_OFFSET*0.9999; //Remain strictly this side of the edge of the texture.
+    texCoord = clamp(texCoord, texCorner, texCorner + texSize);
 #endif
 #if defined (NORMAL_MAPPING)
     // Normalised but not orthonormalised. It should be orthogonal anyway (except for some non-rectangular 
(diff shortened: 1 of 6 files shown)


GooeyHub:

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -92,34 +92,41 @@ void main() {
...
-    vec3 dp2perp = cross(dp2, normal);
-    vec3 dp1perp = cross(normal, dp1);
-    vec3 tangent = dp2perp * duv1.x + dp1perp * duv2.x;
-    vec3 binormal = dp2perp * duv1.y + dp1perp * duv2.y;
-
-    float invMax = inversesqrt(max(dot(tangent,tangent), dot(binormal,binormal)));
-    mat3 tbn = mat3(tangent * invMax, binormal * invMax, normal);
+    // Screen-space coordinates happen to be a coordinate system to which it's possible to relate both view-space and texture coordinates, but are themselves irrelevant.
+    mat2x3 screenToView = mat2x3(dFdx(vertexViewPos.xyz), dFdy(vertexViewPos.xyz));
+    mat2   screenToUv   = mat2  (dFdx(gl_TexCoord[0].xy), dFdy(gl_TexCoord[0].xy)) / TEXTURE_OFFSET;
```
emanuele3d: These two variables store matrices to transform from one space to another: I'd rename them accordingly: `screenSpaceToViewSpace` and `screenSpaceToUvSpace`. I'd also rename vertexViewPos to the less ambiguous vertexPositionInViewSpace - assuming that's what it means.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -92,34 +92,41 @@ void main() {
...
+    //mat2 uvToScreem = inverse(screenToUv);
+    //inverse is not available in GLSL 1.20, so calculate it manually:
```
emanuele3d: How about you wrap this into a small function at the beginning of the shader file? This way you don't have to resort to indentation to separate it from the rest of the code and we don't get distracted by it.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -92,34 +92,41 @@ void main() {
...
+        float det = screenToUv[0][0] * screenToUv[1][1] - screenToUv[1][0] * screenToUv[0][1];
+        mat2 uvToScreen = mat2(vec2(screenToUv[1][1],-screenToUv[0][1]),vec2(-screenToUv[1][0],screenToUv[0][0])) / det;
+    mat2x3 uvToView = screenToView * uvToScreen;
+    mat3 tbn = mat3(uvToView[0], uvToView[1], normal);
```
emanuele3d: Hmmmm. I know what the three vectors in this matrix are and I know "tbn" is a frequently used variable name in the field of computer graphics and shaders. Nevertheless, I wonder if we could find a more meaningful name for it. 

I guess this matrix alone is not meant as a tool to convert from one space to another, it just represent a space, right? If that's the case I guess I'd be happy simply with "tbnSpace". Or does it have other purposes?

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -92,34 +92,41 @@ void main() {
...
 
 #if defined (PARALLAX_MAPPING)
-    vec3 eyeTangentSpace = tbn * vertexViewPos.xyz;
+    vec3 eyeTangentSpace = normalizedVPos * tbn;
```
emanuele3d: Two questions:
1. What's eyeTangentSpace?
2. What's normalized-V-Pos? Did you mean normalizedVertexPosition? And in what space is it?

And isn't the result something related to the vertex? I.e. the vertexPositionInEyeTangentSpace?

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -92,34 +92,41 @@ void main() {
...
 
     float height =  parallaxScale * texture2D(textureAtlasHeight, texCoord).r - parallaxBias;
```
emanuele3d: Height... in what frame of reference? I realize this is an inherited name and you didn't touch it...

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -92,34 +92,41 @@ void main() {
...
-	texCoord += height * normalize(eyeTangentSpace).xy * TEXTURE_OFFSET;
+    //Ideally this should be divided by eyeTangentSpace.z, but in practice this looks better at
+    //low angles, as it can't extend beyond its original space.
```
emanuele3d: This is an interesting insight. Can you clarify the first part of the sentence? I.e. by saying "Ideally this should be divided by eyeTangentSpace.z, for this and that reason, but in practice...."

It'd be good to also clarify what you mean with "as it can't extend beyond its original space." I mean: what's the subject and what do you mean with "extend" in this context?

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -92,34 +92,41 @@ void main() {
...
+	texCoord += height * eyeTangentSpace.xy * TEXTURE_OFFSET;
+	
+	//Crudely prevent the parallax from extending to other textures in the same atlas.
+	vec2 texCorner = floor(gl_TexCoord[0].xy/TEXTURE_OFFSET)*TEXTURE_OFFSET;
+	vec2 texSize = vec2(1,1)*TEXTURE_OFFSET*0.9999; //Remain strictly this side of the edge.
```
emanuele3d: "Remain strictly this side of the edge." Which edge are you referring to? The edge of the texture of a block face? I'd be more explicit about it.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -92,34 +92,41 @@ void main() {
...
+	texCoord = clamp(texCoord, texCorner, texCorner + texSize);
 #endif
-
+#if defined (NORMAL_MAPPING)
+    //Normalised but not orthonormalised. It should usually be orthogonal anyway, but it's not obvious what's the best thing to do when it isn't.
```
emanuele3d: In what circumstance would this matrix not be orthonormal?

engine/src/main/resources/assets/shaders/chunk_frag.glsl
4Denthusiast: I suppose I've referred to them like that myself in the comments too, but it's probably better to call them change-of-basis matrices than referring to them as converting between different spaces. In this case though some of the coordinate systems used aren't actually bases (or are only bases for the tangent space to the mesh, not to view space), as they have too few elements.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
4Denthusiast: As it's an extension of uvToView, it maps vectors from the UV and normal basis to view-space coordinates, but that first basis doesn't have a suitable descriptive name I can think of. The name tangent-binormal-normal is definitely inappropriate here though.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
4Denthusiast: In this case I would have thought the code itself should be sufficiently self-explanatory.

GooeyHub:

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -95,34 +101,29 @@ void main() {
...
-    //inverse is not available in GLSL 1.20, so calculate it manually:
-        float det = screenToUv[0][0] * screenToUv[1][1] - screenToUv[1][0] * screenToUv[0][1];
-        mat2 uvToScreen = mat2(vec2(screenToUv[1][1],-screenToUv[0][1]),vec2(-screenToUv[1][0],screenToUv[0][0])) / det;
+    mat2 uvToScreen = inverse2(screenToUv);
     mat2x3 uvToView = screenToView * uvToScreen;
-    mat3 tbn = mat3(uvToView[0], uvToView[1], normal);
 
 #if defined (PARALLAX_MAPPING)
-    vec3 eyeTangentSpace = normalizedVPos * tbn;
+    vec2 viewDirectionTextureOffset = normalizedVPos * uvToView;
```
emanuele3d: What's normalized-V-Pos? Is it normalized-Vertex-Position? And "normalized" in the sense of  Normalized Device Coordinates?

And can you help me understand viewDirectionTextureOffset? When I think about the view I think about a camera/eye space, and when I think about viewDirection I think about the single vector going out of the lens of the camera - neither, I feel, are the correct interpretation here, which hints to an ambiguity in the naming. Can we try and resolve this?

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -95,34 +101,29 @@ void main() {
...
 
-    float height =  parallaxScale * texture2D(textureAtlasHeight, texCoord).r - parallaxBias;
-    //Ideally this should be divided by eyeTangentSpace.z, but in practice this looks better at
-    //low angles, as it can't extend beyond its original space.
-	texCoord += height * eyeTangentSpace.xy * TEXTURE_OFFSET;
+    float height = parallaxScale * texture2D(textureAtlasHeight, texCoord).r - parallaxBias;
+    //Ideally this should be divided by dot(normal, normalizedVPos), as the texture offset is the amount the light-ray travels in across the texture in the time it takes to traverse the distance "height" perpendicular to the surface, but in practice this way looks better at low angles, as the height-map can't make the triangle extend beyond its normal size.
+	texCoord += height * viewDirectionTextureOffset * TEXTURE_OFFSET;
```
emanuele3d: Hmmm... I find the concept/metaphor of "light-ray traveling" and "the time it takes" somewhat confusing, as in the context of rendering a single frame there usually is no time (except perhaps when motion blur is involved) and no moving light rays. =)

Is this typical Parallax Mapping terminology or is it a noble attempt to clarify an otherwise tricky concept? I mean, aren't you talking about two components of a vector? One along the normal from the surface and the other along the surface? Can it be rephrased in this light?

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -95,34 +101,29 @@ void main() {
...
 	vec2 texCorner = floor(gl_TexCoord[0].xy/TEXTURE_OFFSET)*TEXTURE_OFFSET;
-	vec2 texSize = vec2(1,1)*TEXTURE_OFFSET*0.9999; //Remain strictly this side of the edge.
+	vec2 texSize = vec2(1,1)*TEXTURE_OFFSET*0.9999; //Remain strictly this side of the edge of the texture.
 	texCoord = clamp(texCoord, texCorner, texCorner + texSize);
 #endif
 #if defined (NORMAL_MAPPING)
-    //Normalised but not orthonormalised. It should usually be orthogonal anyway, but it's not obvious what's the best thing to do when it isn't.
-    mat3 tbnNormalised = mat3(normalize(uvToView[0]), normalize(uvToView[1]), normal);
+    //Normalised but not orthonormalised. It should be orthogonal anyway (except for some non-rectangular block shapes like torches), but it's not obvious what's the best thing to do when it isn't.
+    mat3 extendedTangentSpaceToViewSpace = mat3(normalize(uvToView[0]), normalize(uvToView[1]), normal);
```
emanuele3d: Could this be `normalizedTangentSpaceToViewSpace`? When I read "extended" I wonder: extended in what way?

engine/src/main/resources/assets/shaders/chunk_frag.glsl
```diff
@@ -95,34 +101,29 @@ void main() {
 #if defined (NORMAL_MAPPING) || defined (PARALLAX_MAPPING)
     // TODO: Calculates the tangent frame on the fly - this is absurdly costly... But storing
     // the tangent for each vertex in the chunk VBO might be not the best idea either.
-    // Screen-space coordinates happen to be a coordinate system to which it's possible to relate both view-space and texture coordinates, but are themselves irrelevant.
+    // The specific relationship between screen coordinates and view coordinates is irrelevant here. Screen coordinates just happen to be a basis that it's possible to relate both view and UV space coordinates to.
```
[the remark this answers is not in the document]
4Denthusiast: When I first read the previous version of this code, I assumed that as the screen coordinates were used, they were a part of the definition of the tbn frame, but it turns out that they aren't, and so in this version I get rid of the relationship to them as quickly as possible. It turns out that in the previous version, it *did* make a difference, but I'm pretty sure that wasn't intended. I just wanted to help other people avoid the same confusion.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
4Denthusiast: That second interpretation, the single vector going out of the lens of the camera, is exactly what I meant by that, unless I'm not understanding what you mean by that phrase. Would viewDirectionUvProjection be better? I can't quite call it the view direction in UV coordinates, as UV coordinates don't span the whole view-space, but it is the perpendicular projection onto that.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
4Denthusiast: I agree that referring to time here is a little inappropriate, I just couldn't think of a better way of phrasing it. I have no idea if there's standard parallax mapping terminology, but I assume that if someone knows the meaning of specific jargon for this technique, they don't need it explained anyway. What about `Ideally this should be divided by dot(normal, normalizedVPos), as the texture offset is the component parallel to the surface of a vector along the view direction whose coordinate perpendicular to the surface has value "height", but in practice this way looks better at low angles, as the height-map can't make the triangle extend beyond its normal size.`? Personally I find the metaphor of light physically travelling easier to understand than this version.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
4Denthusiast: I meant extended from 2D to 3D by the addition of another basis vector. Although the basis vectors are also normalised, that seems less important to me than that this isn't the same space.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
4Denthusiast: In this case these are mostly just different bases for the same space (or a subspace of it for the 2D ones), as there are natural maps between these coordinate systems that make it seem sensible to consider them as different representations of the same underlying vector space.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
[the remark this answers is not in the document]
4Denthusiast: What I'm trying to justify is the use of dFdx and dFdy.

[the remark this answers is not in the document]
emanuele3d: > What I'm trying to justify is the use of dFdx and dFdy.

That's the missing information then. Put it in the comment and it will work!

engine/src/main/resources/assets/shaders/chunk_frag.glsl
emanuele3d: So it's the forward axis of the view space projected onto the UV space? If that's the case, and I know you don't like this kind of mouthfuls, I'd call it viewForwardAxisProjectedOntoUVspace. Sadly it's the only way to be unambiguous about it.

And I remind you my question: _What's normalized-V-Pos? Is it normalized-Vertex-Position? And "normalized" in the sense of Normalized Device Coordinates?_

engine/src/main/resources/assets/shaders/chunk_frag.glsl
emanuele3d: > but I assume that if someone knows the meaning of specific jargon for this technique, they don't need it explained anyway.

True, but there are also people who will look at this code to learn rather than because they already know it and want to improve it. And then there are people that just need to understand it enough to fit it into a bigger context instead of wanting to know the specific technique  inside-out. Hence my push to make everything as clear and unambiguous as possible. I'm not writing this as a criticism in any way, it's just to describe where I come from and the other types of people that might go through this code. 

In regard to the wording, and starting from your latest example, let me counter propose:

_Ideally this should be divided by dot(normal, normalizedVPos), as TEXTURE_OFFSET is the component parallel to the surface of a vector along the view's forward axis, the other component being a vector perpendicular to the surface and having magnitude "height". In practice this way looks better at low angles, as the height-map can't make the triangle extend beyond its boundaries._

I'm hoping this keeps the meaning while breaking the long sentence a little and adding some further structure to it.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
emanuele3d: Ah. I now understand what you mean with extended. 

Can the purpose of this matrix be summarized as converting a vector from uvnSpace to view space? If that's the case I'd simply name this variable `uvnSpaceToViewSpace`. What do you think?

engine/src/main/resources/assets/shaders/chunk_frag.glsl
4Denthusiast: It's normalized as in same direction but unit length.

engine/src/main/resources/assets/shaders/chunk_frag.glsl
4Denthusiast: What I'm calling "texure offset" is different from TEXTURE_OFFSET, which I suppose is unhelpful, and it's compounded by the fact that TEXTURE_OFFSET is a confusing name to start with.

4Denthusiast:
I'm not sure why normalizedVPos (now normalizedViewPos) is antiparallel to vertexViewPos, but I'd rather just assume there's a sensible reason for it and compensate by negating it when I use it. That's all the requested changes (except viewDirectionUvProjection instead of viewForwardAxisProjectedOntoUVspace).

GooeyHub:

GooeyHub:

GooeyHub:

Josharias:
I ran this locally with much winning! I tried a couple years ago to fix this,  but couldnt figure out what part of that eyeTangentSpace was broken.  Good work!

These are my results from my local with extra lighting and parallax mapping to also confirm this working much better than before.  You can see the normal maps casting great internal shadows on the parallax-ed block face.
terasology-180417105941-1152x720


Oh my local, I also turned off extra lighting and still got east and west block faces to darken when not facing the sun.  Not quite sure how to reproduce what Manu found.



I am excited to some day also get the held blocks and other meshes to also go through the same rendering process to get these effects!

Josharias:
@4Denthusiast,  I think all the things noted have been addressed,  is there anything else to add before merging this in?

4Denthusiast:
@emanuele3d About the linear-interpolation-that's-actually-linear thing, it looks like it doesn't have that problem for you. I get ugly zig-zags, probably just something particular to my graphics driver (GeForce GT 430 with Nvidia's driver 384.111):
screenshot from 2018-04-17 21 36 30
(near the lower edge of the height-test block). I think that when the sample point is sufficiently near the centre of one of the texture's pixels, it doesn't bother to use interpolation.

The lack of shadows for heightmapped blocks is probably just you having the Extra Lighting option turned off, as that's normally working. I'm not following a particular published implementation of parallax mapping, just general linear algebra knowledge.

I'm fine with this being merged as-is.

PR #3309, opened 2018-04-07
Changes to enable FlowingLiquids module

4Denthusiast:
The block data assigned to liquid information is currently only accessible via LiquidData, which restricts its uses in alternative systems. This PR adds getRawLiquid and setRawLiquid methods corresponding to all of the existing getLiquid and setLiquid methods. It also exposes as APIs 2 enums used for rendering that are necessary for the implementation of custom instances of BlockMeshGenerator.

### API Changes
This PR only adds things to the engine API, so the only backwards-compatibility issue would be if modules implemented the extended interfaces, and these implementations became incomplete. This seems unlikely in these cases, but if it's a problem then it can just wait for v2.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/primitives/ChunkMesh.java
+++ b/engine/src/main/java/org/terasology/rendering/primitives/ChunkMesh.java
@@ -25,6 +25,7 @@ import org.lwjgl.opengl.GL13;
 import org.lwjgl.opengl.GL15;
 import org.terasology.engine.subsystem.lwjgl.GLBufferPool;
 import org.terasology.math.geom.Vector3f;
+import org.terasology.module.sandbox.API;
 import org.terasology.rendering.VertexBufferObjectUtil;
 import org.terasology.rendering.assets.material.Material;
 import org.terasology.world.chunks.ChunkConstants;
@@ -56,6 +57,7 @@ public class ChunkMesh {
     /**
      * Possible rendering types.
      */
+    @API
     public enum RenderType {
         OPAQUE(0),
         TRANSLUCENT(1),
--- a/engine/src/main/java/org/terasology/rendering/primitives/ChunkVertexFlag.java
+++ b/engine/src/main/java/org/terasology/rendering/primitives/ChunkVertexFlag.java
@@ -15,8 +15,11 @@
  */
 package org.terasology.rendering.primitives;
 
+import org.terasology.module.sandbox.API;
+
 /**
  */
+@API
 public enum ChunkVertexFlag {
     NORMAL(0, "BLOCK_HINT_NORMAL"),
     WATER(1, "BLOCK_HINT_WATER"),
--- a/engine-tests/src/main/java/org/terasology/MapWorldProvider.java
+++ b/engine-tests/src/main/java/org/terasology/MapWorldProvider.java
@@ -160,6 +160,16 @@ public class MapWorldProvider implements WorldProviderCore {
     }
 
     @Override
+    public boolean setRawLiquid(int x, int y, int z, byte newData, byte oldData) {
+        return false;
+    }
+
+    @Override
+    public byte getRawLiquid(int x, int y, int z) {
+        return 0;
+    }
+
+    @Override
     public byte getLight(int x, int y, int z) {
         return 0;
     }
--- a/engine-tests/src/main/java/org/terasology/testUtil/WorldProviderCoreStub.java
+++ b/engine-tests/src/main/java/org/terasology/testUtil/WorldProviderCoreStub.java
@@ -132,6 +132,16 @@ public class WorldProviderCoreStub implements WorldProviderCore {
     }
 
     @Override
+    public boolean setRawLiquid(int x, int y, int z, byte newData, byte oldData) {
+        return false;  //To change body of implemented methods use File | Settings | File Templates.
+    }
+
+    @Override
+    public byte getRawLiquid(int x, int y, int z) {
+        return 0;  //To change body of implemented methods use File | Settings | File Templates.
+    }
+
+    @Override
     public Block getBlock(int x, int y, int z) {
         Block result = blocks.get(new Vector3i(x, y, z));
         if (result == null) {
--- a/engine/src/main/java/org/terasology/world/ChunkView.java
+++ b/engine/src/main/java/org/terasology/world/ChunkView.java
@@ -187,6 +187,38 @@ public interface ChunkView {
     void setLiquid(int x, int y, int z, LiquidData newState);
 
     /**
+     * @param pos
+     * @return The state of liquid at the given position. This will be no liquid outside the view.
+     */
+    byte getRawLiquid(Vector3i pos);
+
+    /**
+     * @param x
+     * @param y
+     * @param z
+     * @return The state of liquid at the given position. This will be no liquid outside the view.
+     */
+    byte getRawLiquid(int x, int y, int z);
+
+    /**
+     * Sets the liquid state at the given position, if it is within the view
+     *
+     * @param pos
+     * @param newState
+     */
+    void setRawLiquid(Vector3i pos, byte newState);
+
+    /**
+     * Sets the liquid state at the given position, if it is within the view
+     *
+     * @param x
+     * @param y
+     * @param z
+     * @param newState
+     */
+    void setRawLiquid(int x, int y, int z, byte newState);
+
+    /**
      * Converts a coordinate from view-space to world space.
      *
      * @param localPos
--- a/engine/src/main/java/org/terasology/world/ChunkView.java
+++ b/engine/src/main/java/org/terasology/world/ChunkView.java
@@ -188,7 +188,7 @@ public interface ChunkView {
 
     /**
      * @param pos
-     * @return The state of liquid at the given position. This will be no liquid outside the view.
+     * @return The state of liquid at the given position. This will be 0 outside the view.
      */
     byte getRawLiquid(Vector3i pos);
 
@@ -196,7 +196,7 @@ public interface ChunkView {
      * @param x
      * @param y
      * @param z
-     * @return The state of liquid at the given position. This will be no liquid outside the view.
+     * @return The state of liquid at the given position. This will be 0 outside the view.
      */
     byte getRawLiquid(int x, int y, int z);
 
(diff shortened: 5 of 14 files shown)


GooeyHub:

Cervator:
Poking some potentially interested reviewers. Linked with https://github.com/Terasology/FlowingLiquids :-)

Looks really cool, thanks for the work @4Denthusiast. Good heads-up on the API concern, luckily we *are* in v2 development so this is the perfect timing anyway. I ran a build test in a mega-workspace and nothing came up as problematic 👍

4Denthusiast:
As justification for changing how chunks are marked dirty in WorldProviderCoreImpl.setBlocks, collecting them all into a list to avoid duplicates probably made it less efficient in the first place, as calling Chunk.setDirty multiple times is probably no more expensive than checking whether the updated chunk is already in the list. Also the code is neater this way.

GooeyHub:

4Denthusiast:
Another potential change which would be handy but also break the API more would be to not require acknowledgement of the previous liquid state in order to set a new one. That check doesn't actually seem that useful and it's mostly just a pain.

engine/src/main/java/org/terasology/rendering/primitives/ChunkMesh.java
```diff
@@ -56,6 +57,7 @@
     /**
      * Possible rendering types.
      */
+    @API
```
skaldarnar: I don't understand enough about the rendering pipeline to decide whether this should be exposed as `API` or not… the render types seem a bit arbitrary, especially `WATER_AND_ICE`.

engine/src/main/java/org/terasology/rendering/primitives/ChunkVertexFlag.java
```diff
@@ -15,8 +15,11 @@
  */
 package org.terasology.rendering.primitives;
 
+import org.terasology.module.sandbox.API;
+
 /**
  */
+@API
```
skaldarnar: Same here - should this be exposed?

engine/src/main/java/org/terasology/rendering/primitives/ChunkMesh.java
mkienenb: I'm also wondering what the use-case would be for exposing the render types.

4Denthusiast:
@skaldarnar @mkienenb The rendering-related APIs are exposed because FlowingLiquids replaces [the normal BlockMeshGenerator](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/primitives/BlockMeshGeneratorSingleShape.java) on liquids with [a custom one](https://github.com/Terasology/FlowingLiquids/blob/master/src/main/java/org/terasology/flowingliquids/rendering/primitives/BlockMeshGeneratorLiquid.java) that deals with varying the height of the block according to the amount of liquid in it. In order to do this, it's necessary to refer to ChunkVertexFlags and ChunkMesh.RenderTypes in order to pass them to BlockMeshPart.appendTo. I agree that the current values seem quite ad-hoc, but I don't think that's enough of a reason to prevent their actual use in a module.

What do you mean by rawLiquidData being a hard-coded field on blocks? I'd be perfectly happy with it being renamed to something more generic (i.e. not referring to liquids, such as "extraData", which is what it's called internally in ChunkImpl).

4Denthusiast:
As BlockMeshGenerator is itself API, it seems as though custom implementations of this are meant to be possible, but they currently aren't.

mkienenb:
> What do you mean by rawLiquidData being a hard-coded field on blocks? I'd be perfectly happy with it being renamed to something more generic (i.e. not referring to liquids, such as "extraData", which is what it's called internally in ChunkImpl).

I suspect that @skaldarnar means that it doesn't need to be part of the Chunk data but instead should be part of a component system. (stored in a Component).

4Denthusiast:
@mkinenenb The problem with that would be that as water is so common, there could easily be over 10^5 liquid blocks loaded at once, which I expect would end up often using more memory from all those extra block entities, as well as taking considerably more time to access for liquid calculations. It might be worth benchmarking as I'm not really sure, but having a more lightweight way to store a small amount of per-block data seems sensible.

mkienenb:
@skaldarnar may have made the same mistake that I made, thinking you were adding new data to chunk.   I see now that you are just changing how the byte used to construct LiquidData is made accessible.   There's only half-a-byte left free. 

It seems to me that you have two options.   Either make the other half-a-byte available as extra-data for any system to use, or make it available as extra storage in LiquidData.

if you're going to make it generally available, then your getExtraData method (formerly getRawLiquid) has to prevent reading-from and changes-to the first 4 bits of the byte

Othewise, move the methods from Chunk to LiquidData and bit-shift access as appropriate.

I think either approach is fine, depending on how generous you want to be with potentially sharing the data storage space.  I'd probably put it in LiquidData myself.

4Denthusiast:
@mkienenb The newest version (not yet pushed) of FlowingLiquids uses all of the bits available (and a few more would be useful, but that would mean moving up to short which would at least be its own discussion). When you say "LiquidData", do you mean org.terasology.world.liquid.LiquidData or org.terasology.flowingliquids.world.block.LiquidData? The problem I found with Terasology's LiquidData is that the encoding it uses makes only some bytes valid, losing usable space, and its encoding is non-trivial and unrelated to what I actually want to use it for, so going via LiquidData instead of directly is pointlessly inefficient. I also don't want to route things through LiquidData that would make it more essential, given that the feature it was clearly intended to support isn't actually implemented (and would in any case be itself incompatible with FlowingLiquids).

Trying to split up the available data into  portions for different systems to use on every block seems possible, but impractical, as the allocation of data would need to be managed to avoid overlaps between modules. A more sensible way to split the data would be to have each module that uses the extra data deal with some particular set of blocks (e.g. growth & nutrition information for plants or mineral composition of stones & ores), in which case it's reasonable to assume that liquids generally won't have additional functionality beyond flowing.

Anyway, I'll do the renaming & things.

mkienenb:
I've tried to speak in general terms as I've never looked into the specifics of liquid support.  And anything I would have known about it would be 4-years out-of-date.

> FlowingLiquids uses all of the bits available (and a few more would be useful

It was not possible to tell what you needed since your code wasn't committed :).   At the time I looked at the FlowingLiquids module yesterday, it only stored (more precise?) depth info in extraData.

> given that the feature it was clearly intended to support isn't actually implemented

I was unaware that LiquidData/LiquidType was unused in the engine (and I'm guessing unused in modules).   I did wonder why it had a bit indicating water/lava since that seemed to duplicate info elsewhere.   If that's the case, you should rewrite the existing LiquidData (drop LiquidType?) in the engine instead of adding a second view of the data.    I think identifying what the data is used for is important for clarity rather than extraData.   Or instead of calling it liquid, call it FlowData if that more accurately describes what you are doing with it.   Liquid blocks could use flow-data one way, and air blocks could use flow-data a different way.   (wind simulation module!)

I wouldn't worry about adding API for some other kind of system to reuse the same data area at this time.

In summary:  if we're not using the extraData byte, use it but get rid of any of the unused code that previously referenced it.  And indicate what you're using it for.

mkienenb:
And because intentions are not always clear in written communication, let me clarify that my intention is to enable you do this the best possible way, not to make your life difficult by putting unnecessary roadblocks in your path.   I apology that my misunderstandings and lack of knowledge in certain areas has dragged this out.

mkienenb:
Unrelated side note:  I wonder if liquid flows could be more effectively computed using a gpu shader?

jellysnake:
So I'm going to perhaps also recommend another change.
Currently the only two bits of data you cam store on a block are Biome and Liquid data. I'm proposing to change this to allow modules to set additional data points into them. For instance, if I wanted to do a heat simulation I could set Heat Data onto the block.
In doing this the liquid and biome data would be merged into this new field.

If we still keep the existing Liquid data and Biome data methods then the api would be backwards compatible as well.

mkienenb:
More discussion on slack about providing per-block storage.
Reviewing https://forum.terasology.org/threads/modding-per-block-storage-extensions.806/ and https://github.com/mbrotz/Terasology/tree/per-block-storage proposal

Cervator:
Yeah this is becoming a fun little brainstorming topic :-) Sorry to drag out your PR @4Denthusiast !

The big question in the short term would be whether to continue on this path with a right-now improvement and sidelining a potential extra-block-data project for sometime down the road or wait to do both at once. I'm not sure. Might depend a bit on how well @mbrotz's code has aged and how good a fit it would be for current needs. I miss that guy, he should totally come back! ;-)

For a bit more background info: way back in the early ages the engine-level liquid block data was used for the very first round of propagating water. That was eventually removed/broken and the SimpleLiquids module was the most recent attempt at re-implementing before this PR, and didn't use the engine system at all.

For a little while @LinusVanElswijk (DizzyDragon) also had a really cool finite water project going, also led to a "game of life " simulation-in-a-chunk. That was sometime between the loss of the original engine support and SimpleLiquids. I *think* that might have used the engine system but it has been so long I'm really not sure.

mkienenb:
Here's manually-cleaned up diffs of @mbrotz's work.
Four files.

per-block-storage-patch.txt is a patch file of changes against currently-existing files.  Some of them looked like they would merge without issues.  Others would need more work.

[per-block-storage-patch.txt](https://github.com/MovingBlocks/Terasology/files/1973218/per-block-storage-patch.txt)

ModInfo-patch.txt contains changes that need to be applied to gestalt or reworked to be supported by gestalt.

[ModInfo-patch.txt](https://github.com/MovingBlocks/Terasology/files/1973219/ModInfo-patch.txt)

persistence-patch.txt shows changes made to the old way of persisting data.   These would have to be rewritten to the new way to persist data.   Might be trivial.  Might be difficult.  The old files were replaced in this commit:

[persistence-patch.txt](https://github.com/MovingBlocks/Terasology/files/1973223/persistence-patch.txt)

AdvancedConfig-patch.txt allows you to configure per-block-storage.   Might not be needed.  If it were needed, it'd need to be folded into the new FlexibleConfig.

[AdvancedConfig-patch.txt](https://github.com/MovingBlocks/Terasology/files/1973225/AdvancedConfig-patch.txt)

mkienenb:
After sleeping on it, I realized there had to be a better way to make the changes available.

Here it is:

https://github.com/MovingBlocks/Terasology/compare/a7584d11bd94fc60d399424f6b813699fe1b2b6d...mkienenb:per-block-storage

4Denthusiast:
Assuming #3476 is merged, this is no longer relevant. The remaining small parts not relating to data access have been made a separate PR: #3477.

PR #3313, opened 2018-04-12
Update Credits.md

skate54345:
### Contains

Adds credits for new block sound effects that were added

### How to test

Spawn new blocks and hit/destroy them with audio output set up in your testing environment


GooeyHub:

jellysnake:
Is there a particular reason why this is not part of #3297? It seems like this should be simply part of that PR.
Unless I'm missing something here.

Cervator:
Yeah same PR is fine. If I were to wager a guess then this PR might have come about from a separate edit via the website rather than continuing in a local workspace with the branch from the other PR :-)

skate54345:
@Cervator Yeah, I did this one through the site since it was so quick.  My mistake

PR #3315, opened 2018-04-12
Close online players overlay when pause menu opened

ar0ne:
### Contains

I'm not sure will it be useful, but from my perspective it could be useful. 
Well, to avoid mixing of layers I removed OnlinePlayers overlay when we open PauseMenu.
Also I think debug overlay is fine as is. Don't it?
screen shot 2018-04-12 at 23 15 54

### How to test

1. Load/create game
2. Pressing TAB 
3. And press Esc together


GooeyHub:

Cervator:
Agreed! Tested out and merged. Thanks :-)

Yeah debug overlay is fine

PR #3328, opened 2018-04-17
Fixed : Time remains paused even when the back button is clicked in the pause menu.

l0ftyWhizZ:
On toggling the pause menu, if the player chooses to click on the `Back` button to return to the game, it results in the time still being paused (Only pressing `Escape` to return would unpause the time correctly). So, this fix unpauses the time state when the button is clicked.

### How to test

- Create a game, and toggle pause menu (Pressing Esc)
- Click on the `Back` button to return to the game.

Nothing as of now.


GooeyHub:

Cervator:
Thanks for the quick fix :-)

This relates to #3303

Issue #1608, opened 2015-02-22
White-lists and Black-lists


minnesotags:
A useful feature for a game server would be to have configuration file white-lists and black-lists for player names (and ip addresses). These lists could have stored, and/or "pre-seeded" configurations for player names.
##

Cervator:
Closing as completed by #3332 :-)

Still could use a bit of documentation so I submitted #3364 for that

PR #3332, opened 2018-04-20
Add Server Blacklist and Whitelist

Inei1:
### Contains

fix issue #1608.

#### Summary

Implementation for server whitelisting/blacklisting. Netty doesn’t allow refusing connections, but it is possible to immediately close a connection after opening one. I do this by making the network system force disconnect the client if the client ID is found on the blacklist. The client is notified when they are blocked by a new message in the scheduleTimeout function. To make it easier to communicate if the client is on a blacklist or not, protobuf ServerInfoMessage has the blacklist/whitelist added.

#### Details

NetworkSystemImpl is the location where the server rejects or accepts the client. The blacklist file is found and if the client’s ID is found on the list, the client is immediately force disconnected.

For the protobuf changes, the only new thing is that an error message is sent across. Changes in NetworkSystemImpl getServerInfoMessage method provide the data for the new protobuf change.

ClientConnectionHandler is changed so that the user can receive feedback about whether or not they have been blacklisted. The scheduleTimeout function was moved to after the NetMessage was declared. It looks to me that everything still works correctly, and we need it there to access the blacklist. The NetMessage needs to be initialized so we can tell the client that they have been blacklisted, otherwise it just tells them that the server timed out, which is a confusing error message.

The new class doesn’t do much at the moment, but it could be accessed easily from some other source, such as the server facade.

### How to test

At the moment, the blacklist appears to work.

1. Run the game once to generate the blacklist/whitelist files.

2. Add your client ID to the blacklist in JSON format. For example, my blacklist.json contains the line:

<code> ["6a5f11f7-4038-4ef0-91ac-86cb957588b1","01264d12-27cf-4699-b8e2-bdc92ac8ef73"] </code>

These two IDs are for my first and second clients. To find your own client ID, you can run the code with a normal client and headless client, and look at the testing print statement that says “ID: \<client ID\>.”

3. Try to connect to a server and observe how it says that the client is on the blacklist. Also try to connect with another client and notice how nothing is changed. A similar process can be done for testing the whitelist

- [x] Implementation for the whitelist needs to be completed.
- [ ] Allow users to be blacklisted/whitelisted by IP address.
- [ ] Need to add some unit tests.
- [ ] Add javadocs

### DIFF
--- a/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
+++ b/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
@@ -16,6 +16,7 @@
 package org.terasology.network.internal;
 
 import com.google.common.collect.Sets;
+import com.google.gson.Gson;
 import org.jboss.netty.channel.Channel;
 import org.jboss.netty.channel.ChannelHandlerContext;
 import org.jboss.netty.channel.MessageEvent;
@@ -28,6 +29,8 @@ import org.terasology.engine.TerasologyConstants;
 import org.terasology.engine.Time;
 import org.terasology.engine.module.ModuleManager;
 import org.terasology.engine.paths.PathManager;
+import org.terasology.identity.ClientIdentity;
+import org.terasology.identity.PublicIdentityCertificate;
 import org.terasology.module.ModuleLoader;
 import org.terasology.naming.Name;
 import org.terasology.naming.Version;
@@ -69,13 +72,23 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
         this.moduleManager = CoreRegistry.get(ModuleManager.class);
     }
 
-    private void scheduleTimeout(Channel inputChannel) {
+    private void scheduleTimeout(Channel inputChannel, NetData.NetMessage message) {
         channel = inputChannel;
         timeoutPoint = System.currentTimeMillis() + timeoutThreshold;
         timeoutTimer.schedule(new java.util.TimerTask() {
             @Override
             public void run() {
                 synchronized (joinStatus) {
+                    if (clientOnBlacklist(message)) {
+                        joinStatus.setErrorMessage("Client on blacklist");
+                        channel.close();
+                    }
+
+                    if (!clientOnWhitelist(message)) {
+                        joinStatus.setErrorMessage("Client not on whitelist");
+                        channel.close();
+                    }
+
                     if (System.currentTimeMillis() > timeoutPoint
                             && joinStatus.getStatus() != JoinStatus.Status.COMPLETE
                             && joinStatus.getStatus() != JoinStatus.Status.FAILED) {
@@ -94,10 +107,10 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
         if (joinStatus.getStatus() == JoinStatus.Status.FAILED) {
             return;
         }
-        scheduleTimeout(ctx.getChannel());
 
         // Handle message
         NetData.NetMessage message = (NetData.NetMessage) e.getMessage();
+        scheduleTimeout(ctx.getChannel(), message);
         synchronized (joinStatus) {
             timeoutPoint = System.currentTimeMillis() + timeoutThreshold;
             if (message.hasServerInfo()) {
@@ -261,6 +274,26 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
         bldr.setColor(clrbldr.setRgba(config.getPlayer().getColor().rgba()).build());
 
         channelHandlerContext.getChannel().write(NetData.NetMessage.newBuilder().setJoin(bldr).build());
+    }
+
+    private boolean clientOnBlacklist(NetData.NetMessage netMessage) {
+        Config config = CoreRegistry.get(Config.class);
+        if (config != null) {
+            for (ClientIdentity identity : config.getSecurity().getAllIdentities().values()) {
+                String blacklistString = netMessage.getServerInfo().getBlackListList().toString().substring(1,
+                        netMessage.getServerInfo().getBlackListList().toString().length() - 1);
+                Set blacklist = new Gson().fromJson(blacklistString, Set.class);
+                if (blacklist.contains(identity.getPlayerPublicCertificate().toString())) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    // TODO: add a way to enable whitelist and do these checks
+    private boolean clientOnWhitelist(NetData.NetMessage netMessage) {
+        return true;
     }
 
     public JoinStatus getJoinStatus() {
--- a/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
+++ b/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
@@ -16,7 +16,6 @@
 package org.terasology.network.internal;
 
 import com.google.common.collect.Sets;
-import com.google.gson.Gson;
 import org.jboss.netty.channel.Channel;
 import org.jboss.netty.channel.ChannelHandlerContext;
 import org.jboss.netty.channel.MessageEvent;
@@ -29,8 +28,6 @@ import org.terasology.engine.TerasologyConstants;
 import org.terasology.engine.Time;
 import org.terasology.engine.module.ModuleManager;
 import org.terasology.engine.paths.PathManager;
-import org.terasology.identity.ClientIdentity;
-import org.terasology.identity.PublicIdentityCertificate;
 import org.terasology.module.ModuleLoader;
 import org.terasology.naming.Name;
 import org.terasology.naming.Version;
--- a/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
+++ b/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
@@ -66,6 +66,8 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
     public ClientConnectionHandler(JoinStatusImpl joinStatus, NetworkSystemImpl networkSystem) {
         this.networkSystem = networkSystem;
         this.joinStatus = joinStatus;
+        // TODO: implement translation of errorMessage in messageReceived once context is available
+        // See https://github.com/MovingBlocks/Terasology/pull/3332#discussion_r187081375
         this.moduleManager = CoreRegistry.get(ModuleManager.class);
     }
 
--- a/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
+++ b/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
@@ -101,14 +101,14 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
 
         // Handle message
         NetData.NetMessage message = (NetData.NetMessage) e.getMessage();
-        /*String errorMessage = message.getServerInfo().getErrorMessage();
+        String errorMessage = message.getServerInfo().getErrorMessage();
         if (errorMessage != null && !errorMessage.isEmpty()) {
             synchronized (joinStatus) {
                 joinStatus.setErrorMessage(errorMessage);
                 ctx.getChannel().close();
                 return;
             }
-        }*/
+        }
 
         synchronized (joinStatus) {
             timeoutPoint = System.currentTimeMillis() + timeoutThreshold;
--- a/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
+++ b/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
@@ -72,24 +72,13 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
         this.moduleManager = CoreRegistry.get(ModuleManager.class);
     }
 
-    private void scheduleTimeout(Channel inputChannel, NetData.NetMessage message) {
+    private void scheduleTimeout(Channel inputChannel) {
         channel = inputChannel;
         timeoutPoint = System.currentTimeMillis() + timeoutThreshold;
         timeoutTimer.schedule(new java.util.TimerTask() {
             @Override
             public void run() {
-                System.out.println("AAAAAAAAAAAAAAAAAAAAAAAAAA");
                 synchronized (joinStatus) {
-                    if (clientOnBlacklist(message)) {
-                        joinStatus.setErrorMessage("Client on blacklist");
-                        channel.close();
-                    }
-
-                    if (!clientOnWhitelist(message)) {
-                        joinStatus.setErrorMessage("Client not on whitelist");
-                        channel.close();
-                    }
-
                     if (System.currentTimeMillis() > timeoutPoint
                             && joinStatus.getStatus() != JoinStatus.Status.COMPLETE
                             && joinStatus.getStatus() != JoinStatus.Status.FAILED) {
@@ -108,10 +97,19 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
         if (joinStatus.getStatus() == JoinStatus.Status.FAILED) {
             return;
         }
+        scheduleTimeout(ctx.getChannel());
 
         // Handle message
         NetData.NetMessage message = (NetData.NetMessage) e.getMessage();
-        scheduleTimeout(ctx.getChannel(), message);
+        /*String errorMessage = message.getServerInfo().getErrorMessage();
+        if (errorMessage != null && !errorMessage.isEmpty()) {
+            synchronized (joinStatus) {
+                joinStatus.setErrorMessage(errorMessage);
+                ctx.getChannel().close();
+                return;
+            }
+        }*/
+
         synchronized (joinStatus) {
             timeoutPoint = System.currentTimeMillis() + timeoutThreshold;
             if (message.hasServerInfo()) {
@@ -275,41 +273,6 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
         bldr.setColor(clrbldr.setRgba(config.getPlayer().getColor().rgba()).build());
 
         channelHandlerContext.getChannel().write(NetData.NetMessage.newBuilder().setJoin(bldr).build());
-    }
-
-    private boolean clientOnBlacklist(NetData.NetMessage netMessage) {
-        Config config = CoreRegistry.get(Config.class);
-        if (config != null) {
-            for (ClientIdentity identity : config.getSecurity().getAllIdentities().values()) {
-                String blacklistString = netMessage.getServerInfo().getBlackListList().toString().substring(1,
-                        netMessage.getServerInfo().getBlackListList().toString().length() - 1);
-                Set blacklist = new Gson().fromJson(blacklistString, Set.class);
-                if (blacklist == null){
-                    return false;
-                } else if (blacklist.contains(identity.getPlayerPublicCertificate().toString())) {
-                    return true;
-                }
-            }
-        }
-        return false;
-    }
-
-    private boolean clientOnWhitelist(NetData.NetMessage netMessage) {
-        Config config = CoreRegistry.get(Config.class);
-        if (config != null) {
-            for (ClientIdentity identity : config.getSecurity().getAllIdentities().values()) {
-                String whitelistString = netMessage.getServerInfo().getWhiteListList().toString().substring(1,
-                        netMessage.getServerInfo().getWhiteListList().toString().length() - 1);
-                Set whitelist = new Gson().fromJson(whitelistString, Set.class);
-                if (whitelist == null) {
-                    return true;
-                }
-                if (whitelist.contains(identity.getPlayerPublicCertificate().toString()) || whitelist.isEmpty()) {
-                    return true;
-                }
-            }
-        }
-        return false;
     }
 
     public JoinStatus getJoinStatus() {
--- a/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
+++ b/engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
@@ -78,6 +78,7 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
         timeoutTimer.schedule(new java.util.TimerTask() {
             @Override
             public void run() {
+                System.out.println("AAAAAAAAAAAAAAAAAAAAAAAAAA");
                 synchronized (joinStatus) {
                     if (clientOnBlacklist(message)) {
                         joinStatus.setErrorMessage("Client on blacklist");
@@ -283,7 +284,9 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
                 String blacklistString = netMessage.getServerInfo().getBlackListList().toString().substring(1,
                         netMessage.getServerInfo().getBlackListList().toString().length() - 1);
                 Set blacklist = new Gson().fromJson(blacklistString, Set.class);
-                if (blacklist.contains(identity.getPlayerPublicCertificate().toString())) {
+                if (blacklist == null){
+                    return false;
+                } else if (blacklist.contains(identity.getPlayerPublicCertificate().toString())) {
                     return true;
                 }
             }
@@ -291,9 +294,22 @@ public class ClientConnectionHandler extends SimpleChannelUpstreamHandler {
         return false;
     }
 
-    // TODO: add a way to enable whitelist and do these checks
     private boolean clientOnWhitelist(NetData.NetMessage netMessage) {
-        return true;
+        Config config = CoreRegistry.get(Config.class);
+        if (config != null) {
+            for (ClientIdentity identity : config.getSecurity().getAllIdentities().values()) {
+                String whitelistString = netMessage.getServerInfo().getWhiteListList().toString().substring(1,
+                        netMessage.getServerInfo().getWhiteListList().toString().length() - 1);
+                Set whitelist = new Gson().fromJson(whitelistString, Set.class);
+                if (whitelist == null) {
+                    return true;
+                }
+                if (whitelist.contains(identity.getPlayerPublicCertificate().toString()) || whitelist.isEmpty()) {
+                    return true;
+                }
+            }
+        }
+        return false;
     }
 
     public JoinStatus getJoinStatus() {
--- a/engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
+++ b/engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
@@ -23,6 +23,7 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Queues;
 import com.google.common.collect.SetMultimap;
 import com.google.common.collect.Sets;
+import com.google.gson.Gson;
 import com.google.protobuf.ByteString;
 import gnu.trove.map.TIntLongMap;
 import gnu.trove.map.hash.TIntLongHashMap;
@@ -46,6 +47,7 @@ import org.terasology.engine.SimpleUri;
 import org.terasology.engine.Time;
 import org.terasology.engine.module.ModuleManager;
 import org.terasology.engine.module.StandardModuleExtension;
+import org.terasology.engine.paths.PathManager;
 import org.terasology.engine.subsystem.common.hibernation.HibernationManager;
 import org.terasology.entitySystem.Component;
 import org.terasology.entitySystem.entity.EntityRef;
@@ -92,12 +94,15 @@ import org.terasology.world.block.family.BlockFamily;
 import org.terasology.world.chunks.remoteChunkProvider.RemoteChunkProvider;
 import org.terasology.world.generator.WorldGenerator;
 
+import java.io.IOException;
 import java.net.BindException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.InterfaceAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.List;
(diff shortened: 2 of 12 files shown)


GooeyHub:

Inei1:
All of the tests seem to be failing on this line with an NPE:
<code> if (blacklistedIDs.contains(client.getId())){ </code>
Looks like I might need to add a null check or two.

GooeyHub:

Cervator:
Poking @engiValk for review of this one and any related security concerns :-)

Inei1:
Just a few changes for this most recent commit. The new whitelist functionality is essentially just the blacklist functionality with a few minor logic and variable changes. The blacklist/whitelist files are generated on first run with the change to StateHeadlessSetup. There are also a bunch of new null checks. Also, the previous commit is actually not temporary.

I am concerned about using the CoreRegistry class since it isn't encouraged to do so. The ClientConnectionHandler class is the only one where I added it, and CoreRegistry is already used in an existing method. I figure that when the time comes to refactor CoreRegistry out of the class, my usages could be changed without too much difficulty.

GooeyHub:

engine/src/main/protobuf/NetMessage.proto
```diff
@@ -146,6 +146,8 @@ message ServerInfoMessage {
     repeated WorldInfo worldInfo = 9;
     repeated sint32 biomeShortId = 10 [packed=true];
     repeated string biomeId = 11;
+    repeated string blackList = 12;
```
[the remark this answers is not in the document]
mkienenb: Not IpAddress but clientId.   So "blacklistedClientId" and "whitelistedClientId".

Although I'm not certain what the point of blacklisting a client id would be since it'd be trivial to generate a new one.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
```diff
@@ -797,8 +802,34 @@ private void processRemovedClient(Client client) {
...
-        // log after connect so that the name has been set:
-        logger.info("New client connected: {}", client.getName());
+        ServerConnectListManager serverConnectListManager = ServerConnectListManager.getInstance();
+        serverConnectListManager.loadLists();
+        Path blacklistPath = ServerConnectListManager.getInstance().getBlacklistPath();
+        Path whitelistPath = ServerConnectListManager.getInstance().getWhitelistPath();
+        try {
+            Set blacklistedIDs = new Gson().fromJson(Files.newBufferedReader(blacklistPath), Set.class);
+            Set whitelistedIDs = new Gson().fromJson(Files.newBufferedReader(whitelistPath), Set.class);
+            System.out.println("BLIDS: " + blacklistedIDs);
```
mkienenb: use logger.

engine/src/main/java/org/terasology/network/internal/ServerConnectListManager.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Set;
+
+/**
+ *
+ */
+
+public class ServerConnectListManager {
+
+    private static final Logger logger = LoggerFactory.getLogger(ServerConnectListManager.class);
+    private static final Path defaultBlacklistPath = PathManager.getInstance().getHomePath().resolve("blacklist.json");
```
mkienenb: It's a bad idea to execute this during class loading.  homePath will be changed while the application is starting up, but this caches the homePath at the time of PathManager classloading before homePath is properly configured.  Seems like a design flaw in PathManager as well, but that's another issue.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
```diff
@@ -797,8 +802,34 @@ private void processRemovedClient(Client client) {
 
     private void processNewClient(NetClient client) {
         client.connected(entityManager, entitySerializer, eventSerializer, eventLibrary);
-        // log after connect so that the name has been set:
-        logger.info("New client connected: {}", client.getName());
+        ServerConnectListManager serverConnectListManager = ServerConnectListManager.getInstance();
+        serverConnectListManager.loadLists();
```
mkienenb: This section from line 806 through 831 is duplicating code and functionality with ServerConnectListManager.   Let ServerConnectListManager do the set up and checking for validity.

Something along the lines of the following should be sufficient for NetworkSystemImpl.
`
if (!serverConnectListManager.isClientAllowedToConnect(client)) {
    forceDisconnect(client);
    return;
}
`
By doing it this way, you encapsulate the validity of a client connection outside of this class.   Later on (probably sooner rather than later), we might want to allow access based on ip address in addition to client id.   Or by a public/private key.   Having all these checks in a single class makes it easy to extend and makes this class easier to read.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
```diff
@@ -797,8 +802,34 @@ private void processRemovedClient(Client client) {
...
+            if (whitelistedIDs != null) {
+                if (!whitelistedIDs.contains(client.getId())) {
+                    System.out.println("WLID: " + client.getId());
+                    forceDisconnect(client);
+                    return;
+                }
+            }
+            System.out.println("ID: " + client.getId());
+        } catch (IOException e) {
+            e.printStackTrace();
```
mkienenb: Use logger.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
```diff
@@ -847,7 +879,7 @@ private void connectClient(Client client) {
         }
         WorldGenerator worldGen = context.get(WorldGenerator.class);
         if (worldGen != null) {
-            serverInfoMessageBuilder.setReflectionHeight(worldGen.getWorld().getSeaLevel() + 0.5f);
+            serverInfoMessageBuilder.setReflectionHeight(worldGen.getWorld().getSeaLevel());
```
mkienenb: Why are you changing reflection height as part of this PR?

Sometimes it is reasonable to make a quick-fix for an unrelated issue while working on a file, but whenever you do so, clearly indicate this in the main PR description, and since GitHub makes it easy to do so, add a GitHub code comment to the line to make it obvious to any reviewers or mergers of your code.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
```diff
@@ -868,6 +900,18 @@ private void connectClient(Client client) {
         for (BlockFamily registeredBlockFamily : blockManager.listRegisteredBlockFamilies()) {
             serverInfoMessageBuilder.addRegisterBlockFamily(registeredBlockFamily.getURI().toString());
         }
+        Path homePath = PathManager.getInstance().getHomePath();
```
mkienenb: Note that this homePath is probably going to be different from the homePath cached in ServerConnectListManager on class load.   I'm surprised you haven't seen problems due to this.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
```diff
@@ -868,6 +900,18 @@ private void connectClient(Client client) {
...
+        try {
+            for (String line : Files.readAllLines(homePath.resolve("blacklist.json"))) {
```
mkienenb: It seems odd to me that we have class ServerConnectListManager for encapsulating blacklists and whitelists, but we're reading the data from the files rather than from ServerConnectListManager.   I get it that reading the data from the files is easier than reading it from the list collections, but moving this code into a method of ServerConnectListManager can hide that fact from the caller.   So you're going to want a method like
`
for (String line : serverConnectListManager.getWhiteListLines()) {
`
 instead of 
`
for (String line : Files.readAllLines(homePath.resolve("whitelist.json"))) {
`

When all is said and done, there shouldn't be any references to the whitelist and blacklist files outside of ServerConnectListManager.   Someone should be able to swap in a different implementation of ServerConnectListManager such as reading from a public REST service instead of local files without changing any other code.

engine/src/main/protobuf/NetMessage.proto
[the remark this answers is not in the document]
mkienenb: Now that I've looked at all the code, I am wondering why we are passing the whitelist and blacklist to the client.   If it's just to indicate what the error message states, I recommend sending an error message instead.   It's just asking for trouble to send the list of whitelisted client ids to a client who was rejected.   A clever person would just copy a whitelisted client id into their client.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
Inei1: It looks like it was a fairly recent change to the main repo and I haven't pulled it into my local workspace. I'll change it back.

engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
```diff
@@ -69,13 +72,24 @@ public ClientConnectionHandler(JoinStatusImpl joinStatus, NetworkSystemImpl netw
...
     }
 
-    private void scheduleTimeout(Channel inputChannel) {
+    private void scheduleTimeout(Channel inputChannel, NetData.NetMessage message) {
         channel = inputChannel;
         timeoutPoint = System.currentTimeMillis() + timeoutThreshold;
         timeoutTimer.schedule(new java.util.TimerTask() {
             @Override
             public void run() {
+                System.out.println("AAAAAAAAAAAAAAAAAAAAAAAAAA");
```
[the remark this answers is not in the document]
Inei1: I forgot to remove that debug statement before I committed. I'm using System,out.println() for debug statements because it is easier to notice and remove them before commiting (which I failed to do this time...).

engine/src/main/protobuf/NetMessage.proto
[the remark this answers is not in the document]
Inei1: Now that I think about it, you are right. I should probably be sending an error message across instead.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
Inei1: I just did this locally and should commit it later today.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
Inei1: N/A because of the change above.

engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
```diff
@@ -263,6 +277,41 @@ private void sendJoin(ChannelHandlerContext channelHandlerContext) {
         channelHandlerContext.getChannel().write(NetData.NetMessage.newBuilder().setJoin(bldr).build());
     }
 
+    private boolean clientOnBlacklist(NetData.NetMessage netMessage) {
+        Config config = CoreRegistry.get(Config.class);
```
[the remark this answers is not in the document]
Inei1: I'm probably going to delete this as part of not exposing the blacklist/whitelist and instead an error message.

engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
```diff
@@ -263,6 +277,41 @@ private void sendJoin(ChannelHandlerContext channelHandlerContext) {
...
+        if (config != null) {
+            for (ClientIdentity identity : config.getSecurity().getAllIdentities().values()) {
+                String blacklistString = netMessage.getServerInfo().getBlackListList().toString().substring(1,
```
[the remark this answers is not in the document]
Inei1: I'm probably going to delete this for the same reason as above.

engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
```diff
@@ -263,6 +277,41 @@ private void sendJoin(ChannelHandlerContext channelHandlerContext) {
...
+                        netMessage.getServerInfo().getBlackListList().toString().length() - 1);
+                Set blacklist = new Gson().fromJson(blacklistString, Set.class);
+                if (blacklist == null){
+                    return false;
+                } else if (blacklist.contains(identity.getPlayerPublicCertificate().toString())) {
```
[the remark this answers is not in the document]
Inei1: Same as above.

engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
```diff
@@ -263,6 +277,41 @@ private void sendJoin(ChannelHandlerContext channelHandlerContext) {
...
+            }
+        }
+        return false;
+    }
+
+    private boolean clientOnWhitelist(NetData.NetMessage netMessage) {
+        Config config = CoreRegistry.get(Config.class);
+        if (config != null) {
+            for (ClientIdentity identity : config.getSecurity().getAllIdentities().values()) {
+                String whitelistString = netMessage.getServerInfo().getWhiteListList().toString().substring(1,
```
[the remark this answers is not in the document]
Inei1: Deleting this as well.

engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
```diff
@@ -263,6 +277,41 @@ private void sendJoin(ChannelHandlerContext channelHandlerContext) {
...
+                        netMessage.getServerInfo().getWhiteListList().toString().length() - 1);
+                Set whitelist = new Gson().fromJson(whitelistString, Set.class);
+                if (whitelist == null) {
+                    return true;
+                }
+                if (whitelist.contains(identity.getPlayerPublicCertificate().toString()) || whitelist.isEmpty()) {
```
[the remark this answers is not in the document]
Inei1: And deleting this one too.

engine/src/main/java/org/terasology/network/internal/ServerConnectListManager.java
```diff
@@ -0,0 +1,101 @@
...
+                Files.createFile(blacklistPath);
+            }
+            if (!Files.exists(whitelistPath)) {
+                Files.createFile(whitelistPath);
+            }
+            blacklistedIDs = gson.fromJson(Files.newBufferedReader(blacklistPath), Set.class);
+            whitelistedIDs = gson.fromJson(Files.newBufferedReader(whitelistPath), Set.class);
+        } catch (IOException e) {
+            logger.error("Whitelist or blacklist files not found:");
+            e.printStackTrace();
```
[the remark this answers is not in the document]
Inei1: Good to know, I'll do that from now on.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
Inei1: This is removed as part of refactoring.

Cervator:
Thanks to you both for the feature, the nice review work, and the adjustments. Keep it coming and we'll get this merged soon :-)

Inei1:
Some of the requested changes have been done in the newest commit. There are still some bugs though, and one of them is serious. If you click join game and quickly switch to custom and join the server, you can have the error message say "Disconnected From Server" when it should say something about the blacklist. The blacklist blocks and accepts connections as appropriate, but for some reason connecting as an authorized user causes the following NPE:

<pre>20:59:36.320 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
	at org.terasology.logic.characters.CharacterMovementSystemUtility.setToState(CharacterMovementSystemUtility.java:75)
	at org.terasology.logic.characters.ClientCharacterPredictionSystem.onCharacterStateReceived(ClientCharacterPredictionSystem.java:117)
	at org.terasology.logic.characters.ClientCharacterPredictionSystemMethodAccess.invoke(Unknown Source)
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:516)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.sendStandardEvent(EventSystemImpl.java:279)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:270)
	at org.terasology.entitySystem.entity.internal.BaseEntityRef.send(BaseEntityRef.java:204)
	at org.terasology.network.internal.ServerImpl.processEvent(ServerImpl.java:244)
	at org.terasology.network.internal.ServerImpl.processMessages(ServerImpl.java:276)
	at org.terasology.network.internal.NetworkSystemImpl.update(NetworkSystemImpl.java:324)
	at org.terasology.engine.TerasologyEngine.tick(TerasologyEngine.java:430)
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:400)
	at org.terasology.engine.TerasologyEngine.run(TerasologyEngine.java:376)
	at org.terasology.engine.Terasology.main(Terasology.java:154) </pre>

GooeyHub:

mkienenb:
Well, the error is saying that there is no LocationComponent for the gaze entity returned by `GazeAuthoritySystem.getGazeEntityForCharacter(entity)`.
`
            LocationComponent gazeLocation = gazeEntity.getComponent(LocationComponent.class);
            gazeLocation.setLocalRotation(rotation);
`

That's a new area to me.   Maybe @Josharias can help as he committed the component.
It doesn't seem likely that this problem is caused by this patch, but maybe I've overlooked something.

Inei1:
It turns out that the serious bug was caused by two things:
1. processRemovedClient function was trying to remove a client that wasn't even connected, causing problems with the server. This also generated a problematic player file which caused crashes.
2. Kicked status was not being reset correctly. This made it so that whenever a blacklisted client tried to connect, the next client was said to have been kicked, and the function that normally cleans up a client requires that the client isn't clicked.

Now that these are taken care of, I can work on the rest of the feedback and try to finish this up.

Also, a few of the null checks I added are no longer needed with these recent changes, so I removed them.

GooeyHub:

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
```diff
@@ -797,8 +802,17 @@ private void processRemovedClient(Client client) {
 
     private void processNewClient(NetClient client) {
         client.connected(entityManager, entitySerializer, eventSerializer, eventLibrary);
-        // log after connect so that the name has been set:
-        logger.info("New client connected: {}", client.getName());
```
mkienenb: Note that there's nothing wrong with leaving logging in to help debug problems later.  Loggers can be enabled/disabled while the program is running.
If calculating the arguments are expensive, prefix with `if (logger.isInfoEnabled())`
"

engine/src/main/java/org/terasology/network/internal/ServerConnectListManager.java
```diff
@@ -0,0 +1,118 @@
...
+    public void addToWhitelist(String clientID) {
+        whitelistedIDs.add(clientID);
+    }
+
+    public void removeFromWhitelist(String clientID) {
+        whitelistedIDs.remove(clientID);
+    }
+
+    public Set<String> getWhitelist() {
+        return whitelistedIDs;
```
mkienenb: You still need to return either a copy of the data, or more appropriately in this case, an unmodifyableSet() wrapper of the set.

engine/src/main/java/org/terasology/network/internal/ServerConnectListManager.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public void addToBlacklist(String clientID) {
+        blacklistedIDs.add(clientID);
+    }
+
+    public void removeFromBlacklist(String clientID) {
+        blacklistedIDs.remove(clientID);
+    }
+
+    public Set<String> getBlacklist() {
```
mkienenb: You still need to return either a copy of the data, or more appropriately in this case, an unmodifyableSet() wrapper of the set.

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
```diff
@@ -784,6 +785,10 @@ void removeClient(NetClient client) {
 
     private void processRemovedClient(Client client) {
         if (client instanceof NetClient) {
+            ServerConnectListManager serverConnectListManager = context.get(ServerConnectListManager.class);
```
gianluca-nitti: Is the purpose of this part to avoid doing the cleanup performed by `processRemovedClient` in case the client was immediately disconnected due to not being allowed, and thus was never actually fully connected (i.e. not added to `clientList`, `netClientList`, `clientPlayerLookup`, etc)?

(not implying I disagree with anything, just asking because I don't have much experience with the NetworkSystem and I'm curios whether I'm understanding it correctly)

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
```diff
@@ -150,6 +150,7 @@
     private Map<EntityRef, EntityRef> ownerLookup = Maps.newHashMap();
     private SetMultimap<EntityRef, EntityRef> ownedLookup = HashMultimap.create();
     private StorageManager storageManager;
+    private String errorMessage;
```
gianluca-nitti: I'm not 100% sure but it seems like it's not strictly necessary to have this as a global field rather than a local variable passed around when needed. I think it would need to be passed just to `NetData.ServerInfoMessage getServerInfoMessage()`. So it may become a `private NetData.ServerInfoMessage getServerInfoMessage(String errorMessage)`, then to avoid breaking the interface of the class we could add an overloaded version with the same signature as before, like this:
```
NetData.ServerInfoMessage getServerInfoMessage() {
    return getServerInfoMessage(null);
}
```

engine/src/main/java/org/terasology/network/internal/ServerConnectListManager.java
```diff
@@ -0,0 +1,110 @@
...
+            blacklistedIDs = gson.fromJson(Files.newBufferedReader(blacklistPath), Set.class);
+            whitelistedIDs = gson.fromJson(Files.newBufferedReader(whitelistPath), Set.class);
+        } catch (IOException e) {
+            logger.error("Whitelist or blacklist files not found:", e);
+        }
+    }
+
+    public String getErrorMessage(String clientID) {
+        if (isClientBlacklisted(clientID)) {
+            return "client on blacklist";
```
gianluca-nitti: If you decide to localize, you could send the identifier of a string to take from the translation system (for example `${engine:menu#client-on-blacklist}`), instead of the human-readable string, and put the English value in `menu_en.lang`

engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
```diff
@@ -98,6 +101,15 @@ public void messageReceived(ChannelHandlerContext ctx, MessageEvent e) {
 
         // Handle message
         NetData.NetMessage message = (NetData.NetMessage) e.getMessage();
+        String errorMessage = message.getServerInfo().getErrorMessage();
+        if (errorMessage != null && !errorMessage.isEmpty()) {
+            synchronized (joinStatus) {
+                joinStatus.setErrorMessage(errorMessage);
```
gianluca-nitti: About localization: you may send the id of a string to take from the translation system instead of the human-readable English-only message. Then here you would need to access an instance of `TranslationSystem`, probably via `context`, to actually load the translated string.

However it may not be trivial to get access to a `Context` from `ClientConnectionHandler`... you may want to evaluate if it's worth it, in my opinion localizing such a short string would be nice if it's simple but not much of a big deal if it takes effort - that would be the job for another patch/PR (as you notice there are some other strings around which are not localized)

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
Inei1: Good point, this isn't too hard to fix so I will do so.

engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
Inei1: It would be nice to have translations here, but it's not trivial like you mentioned, and I think it belongs in a different PR.

engine/src/main/java/org/terasology/network/internal/NetworkSystemImpl.java
gianluca-nitti: Good, thanks for the explanation

engine/src/main/java/org/terasology/network/internal/ServerConnectListManager.java
```diff
@@ -0,0 +1,146 @@
...
+ * based on the blacklist and whitelist files.
+ */
+
+public class ServerConnectListManager {
+
+    private static final Logger logger = LoggerFactory.getLogger(ServerConnectListManager.class);
+    private static final Gson gson = new Gson();
+
+    private Context context;
+    private Set blacklistedIDs;
```
gianluca-nitti: Just noticed [raw types](https://docs.oracle.com/javase/tutorial/java/generics/rawTypes.html) are used here and `@SuppressWarnings("unckecked")` is used when manipulating the sets. I think the reason you did it this way is because if you make them `Set<String>`, then you get the warning when loading from json.
However I would solve the problem the other way round, i.e. using `Set<String>` as the type for `blacklistedIDs` and `whitelistedIDs`, because in my understanding these sets will always contain `String` objects. Then we could happily remove the `SuppressWarnings` from `addToBlacklist` and `addToWhitelist`.

We would then get warnings in `loadLists()`. I can think of a couple of solutions:
* Simply adding `@SuppressWarnings("unckecked")` to `loadLists()` - in my opinion if we need to do something unchecked, then the proper moment to do it is when deserializing, and once we are done with this, the data in memory is properly typed
* (this may be an overkill) Instead of serializing two `Set`s, we could serialize a class which contains them. This way, the storage format of the lists would be a single json file with a structure like `{"blacklistedIDs": [...], "whitelistedIDs": [...]}`. The class could be
    * this class itself (`ServerConnectListManager`), however the name would not be very appropriate from a semantic point of view - maybe it could be renamed to something like `ServerConnectLists` - and proper measures should be taken to make sure other non-relevant fields are not serialized; I think the paths could be made static (should be the same for any instance, correct?), and if we decide to keep `context` we have to somehow mark it as non-serialized (I think it can be made with the `transient` modifier, but I'm not sure).
    * a separate class, to which the two `Set` fields and `isClientOnWhitelist`, `isClientOnBlacklist`, `getWhitelist` and `getBlacklist` methods could be moved

(sorry if we are bothering you with all these changes request, I think this one is quite important from a code quality point of view but didn't notice it yesterday - my bad)

engine/src/main/java/org/terasology/network/internal/ServerConnectListManager.java
```diff
@@ -0,0 +1,146 @@
...
+        if (!Files.exists(whitelistPath)) {
+            Files.createFile(whitelistPath);
+        }
+        return true;
+    }
+
+    public String getErrorMessage(String clientID) {
+        TranslationSystem translationSystem = context.get(TranslationSystem.class);
+        if (isClientBlacklisted(clientID)) {
+            return translationSystem.translate("${engine:menu#client-on-blacklist}");
```
gianluca-nitti: If I understand it correctly, this way the string is translated *by the server*, but the language set in the client and the one set in the server may differ so the translation in my understanding would be quite pointless, and possibly confusing if for example the user has their client set in English but the server is set to another language the user doesn't understand.

For now I would just send the English string (basically I'd reverse the changes you made to access `TranslationSystem`, including the part which brings here the `Context`), then if there is time / anybody else wants to do it, we could work on a separate "improve localization" PR which makes the translation on the client (the "non-trivial" stuff we were discussing yesterday)

Inei1: I think you are right about how the server would translate the string based on its own settings. I suppose that when the time comes to do the translation for ClientConnectionHandler, we could change the value of getErrorMessage to return a string such as "client-on-blacklist" and use that with the translationSystem on the client side.

I still need the context to get the display so that the code will not generate blacklist and whitelist files whenever a game client starts up.

Inei1:
I think the changes I have made cover everything up to this point. We could merge this now and it would work, although I would like to try to block by IP addresses as well if I don't run out of time before GSOC coding starts.

@gianluca-nitti I'm fine with this taking a long time to merge, it is a fairly big change after all. I get to learn more about the code review process as well.

GooeyHub:

Inei1:
For blacklisting and whitelisting IP addresses, I'm not sure if it will be easily possible to do. I'm going to leave a few notes here to explain my process and provide an approach for if someone else (or me) decides to add this in later. My understanding is that the IP address of the connecting client is first available in the join() method of NetworkSystemImpl through the line ```connectCheck.await();```

You can then get the IP address of the client by calling ```connectCheck.getChannel().getLocalAddress()```
Note that the function is getLocalAddress because the join() function is called by the client; getRemoteAddress returns the address of the server. The problem is that this code is client side, and it isn't safe to do client side authentication.

The server does get the client IP address in ServerConnectionHandler. Maybe it also gets it in other places. The messageReceived() and receivedConnect() methods appear to be the last places where the client's IP address is visible through ```channelHandlerContext.getChannel().getRemoteAddress()```. receivedConnect() calls some functions that add the connecting client to the newClients variable in NetworkSystemImpl, where it gets added by update() and goes through my blacklisting code. What we need to have done is we need to make the IP address available in the same place as a context so that ServerConnectListManager can check to see if the IP address is blocked.

I think it would be easiest to send the IP address to the networkSystem in ServerConnectionHandler. We could make a new function ```setIPAddress()``` and a new global variable ```IPAddress``` in NetworkSystemImpl, then call ```networkSystem.setIPAddress(channelHandlerContext.getChannel().getRemoteAddress())``` inside ```receivedConnect()```. I don't know if this is acceptable from a code quality standpoint.

Either way, with GSoC coding starting in a few days I don't want to take on any more work than I have to. So if there aren't any other changes I need to make, this is ready for merging.

engine/src/main/java/org/terasology/network/internal/ServerConnectListManager.java
gianluca-nitti: Ah okay, I didn't notice the context is also needed for the display

gianluca-nitti:
Thank you @Inei1, I had a final look, tested everything out again and for me it's ok to merge.

Ping @mkienenb to see if he approves as well and @Cervator for a possible merge

engine/src/main/java/org/terasology/network/internal/ClientConnectionHandler.java
```diff
@@ -98,6 +98,15 @@ public void messageReceived(ChannelHandlerContext ctx, MessageEvent e) {
...
```
mkienenb: At some point context is going to have to be made available to this class to fix this line in the constructor, probably by passing a context into the constructor.
`
        this.moduleManager = CoreRegistry.get(ModuleManager.class);
`
For now, I think it's sufficient to add a todo to the effect of:
`
// TODO: implement translation of error messages once context is available
`
Probably right above the line that calls CoreRegistry.

mkienenb:
Yes, I don't think you should be trying to add in ip address checking as part of this PR.   I was going to say that earlier but didn't get around to it.

You can remove the work-in-progress tag as soon as you are no longer planning to make changes.   (ie, as soon as you want it to be reviewed and merged).

GooeyHub:

PR #3335, opened 2018-04-21
Show error if module loading fails.

### LINKED ISSUE #3333: No error message displayed on module update failure
Note, this Bug is strictly for fixing the lack of error message. #3334 is for fixing the module loading itself. This patch is to make it so the failure is not silent.

Update a module (on windows) that had been loaded. It failed, but I didn't get an error message.

### What actually happened

The loading began, but when it attempted to load the module, an exception was thrown because the file was being accessed by another process. However, no error message was displayed.

### How to reproduce

The following steps assume that a bug related to module updating hasn't been fixed.

* Step 1
Install a module like basic crafting.
* Step 2
Start A game and exit.
* Step 3
Click Create a new game.
* Step 4
Click modules
* Step 5
Attempt to update a module that depends on Basic Crafting
* Step 6
Wait for dialog to disappear without completing updating.

cwgreene:
Currently, if something goes wrong during module loading, the module loading dialog goes away without explanation.

This change displays an error message informing the user that something went wrong, possibly with enough information to debug.

Fixes #3333

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WaitPopup.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WaitPopup.java
@@ -97,6 +97,7 @@ public class WaitPopup<T> extends CoreScreenLayer {
         } catch (InterruptedException | ExecutionException e) {
             logger.warn("An error occurred during execution", e);
             getManager().popScreen();
+            getManager().pushScreen(MessagePopup.ASSET_URI, MessagePopup.class).setMessage("Error", e.getMessage());
         }
     }
 


GooeyHub:

Cervator:
Heya @cwgreene and thanks for the PR :-)

Do you have a preferred way to reproduce a situation in which this occurs? That makes it easier to test it

@GooeyHub: ok to test

GooeyHub:

cwgreene:
@Cervator, oops, sorry I missed that message. The most reliable way of testing this at the moment is unfortunately via #3333, though it may be possible to reliably repro it in future (in an integ test) by controlling the network connection and forcing a failure. I'll take a look at our existing integ test infra.

mkienenb:
Can be tested by starting Terasology, creating a new game, clicking the Modules button, disconnecting your internet connection, then attempting to download a module.   Without the patch, you'll only see a generic error message about the module failing to load.  With the patch, you'll see "java.net.UnknownHostException: artifactory.terasology.org"

mkienenb:
Ha!  I didn't see your comment an hour ago when I started working on this.   Yes, I came to the same conclusion that disconnecting the network was the easiest way to test it.   Thanks for the patch.

PR #3340, opened 2018-04-24
Direction class unit tests

J-Dill:
### Contains

Adds unit tests for the two of the methods inside of the Direction enum class.

### How to test

Run the tests yourself, along with the entire suite if you like : )

If you have suggestions as far as anything goes let me know! I am also open
to suggestions when it comes to adding more tests to this class.


GooeyHub:

Cervator:
Much <3 for unit tests! They look quite sensible and test out fine :-)

I don't know what else might be useful to test here. Maybe @pollend would have ideas related to block families, rotations, and such, but it may be easier to ask on Slack

PR #3149, opened 2017-11-27
New block families by pollend - merge to develop post-GCI

Cervator:
### Contains

New block family system by @pollend  - see [this Gist](https://gist.github.com/pollend/6f106889f199d1a30736c3beddaa9e26) for some documentation and philosophy on the change as well as original PR #3085

### How to test

Will need to experiment some with an existing block family or make a new one. Some directions in the Gist but more details for here would be good. Probably Rails would be the ultimate test case.

### Outstanding before merging

This breaks backwards compatibility and is part of the v2.0.0 engine release. Some modules are impacted and should be updated with PRs that make the associated changes required by this PR. A task chain for GCI is based on this PR's branch and should be allowed to finish with additional improvements PRed to the branch before this is merged.

- [ ] MarcinScIncubator
- [ ] ... more modules
- [ ] Await end of GCI for final feedback and additional changes (should be merged into the `newBlockFamilies` branch

Additional comments in this PR for feedback/improvements would also be appreciated :-)

PR #3343, opened 2018-04-27
New block families - rebased

Cervator:
Refresh of #3149 that has been rebased on latest develop as of this PR submission. See it for most the details. Summary of outstanding issues:

- [ ] Rotation still seems to need something to let unique rotation blocks place correctly facing a given direction. For instance all stair blocks place facing the same way, rather than the direction facing the player at placement time. @pollend mentioned not having thought of the ideal approach for this yet
- [x] Code quality: https://github.com/MovingBlocks/Terasology/pull/3221/files#diff-ea4234b8e155dbfd65deb02f7b1b7605R229 should perhaps be a more narrowly defined Exception? Additional comment in the PR itself https://github.com/MovingBlocks/Terasology/pull/3221
- [x] Code quality: Excessive `return` at https://github.com/MovingBlocks/Terasology/pull/3221/files#diff-ea4234b8e155dbfd65deb02f7b1b7605R231
- [x] Probably could use some more Checkstyle scanning etc
- [ ] The GrowingFlora trees (viewable by running the game with NeoTTA + its world) aren't using their pipe-like shape correctly and instead look like weird trunk blocks. Unless that's just me

### Related/updated module PRs

These seem to make the modules work, or at least compile. Some module sets can't be run fully yet without more fixes.

- [ ] https://github.com/Terasology/Fences/pull/8
- [ ] https://github.com/Terasology/MarcinScIncubator/pull/3
- [ ] https://github.com/Terasology/StructuralResources/pull/3 - no unique family so less work
- [ ] https://github.com/Terasology/MultiBlock/pull/6
- [ ] https://github.com/Terasology/IRLCorp/pull/16
- [ ] https://github.com/Terasology/ManualLabor/pull/23
- [ ] https://github.com/Terasology/JoshariasSurvival/pull/23

### PRs exist but need more work

Early versions of the needed changes were submitted here, but later went out of date as the new system evolved further

- [ ] https://github.com/Terasology/Signalling/pull/17
- [ ] https://github.com/Terasology/Machines/pull/25 - also clashes with the changes in #3221 and is a bit more work than I can put in right now (complex family with all faces and an archetype, logic defines it all over several passes, hard to readily tweak)

### Modules that still need a PR to fix related code

These need basic review akin to one of the GCI tasks that resulted in a bunch of the above PRs. Can review some of those PRs to see the likely changes to the modules in this list

- [ ] GrowingFlora - this one is fairly central, would be a first good target so we can test other modules that have this one in their dependency tree. New PR: https://github.com/Terasology/GrowingFlora/pull/12 - NeoTTA works with this but all tree branches lack their "pipe" shape?
- [ ] TutorialAssetSystem (and probably its wiki as well)
- [ ] ItemPipes
- [ ] LightAndShadow
- [ ] Minesweeper - seemingly just needs a standard family update
- [ ] ModularComputers - single method no longer matches a signature
- [ ] Rails
- [ ] Sample - special since it has several old families from GCI including [PRs](https://github.com/Terasology/Sample/pulls) to update
- [ ] StructuralTemplates

### Unknown

Can't compile due to broken dependencies. Could be fine after dependencies fixed (yep, seemingly build now after the GF compile fix)

- [x] Lost - probably OK, fails from GrowingFlora not building
- [x] NeoTTA - probably OK, fails from GrowingFlora not building
- [x] WoodAndStone - probably OK, fails from GrowingFlora not building

Also: potentially any module that uses block families the old school way in prefabs rather than define their own unique block families thus causing compilation errors. I haven't looked for those, nor have I done any specific play testing of affected block families beyond the basics (torch, stairs, fences...)

### Further testing

Use this rebased branch for testing modules. The older branch predates the big v2 engine merge and will show dependency issues on a large amount of modules due to version bumps made later.

Might be of particular interest to @asiekierka who may refactor further in the area as part of GSOC 2018


GooeyHub:

NicholasBatesNZ:
I took the liberty to convert [GrowingFlora](https://github.com/Terasology/GrowingFlora/pull/12) the new system.

Cervator:
Thanks @NicholasBatesNZ :-) that lets GF and things depending on it build and run, sweet!

However, either the base system has another bug or the family setup isn't quite right. When I run the full NeoTTA gameplay template with its world the trees have their branches without their pipe-like shapes. Unsure if that's a general issue or specific to that family.

@pollend is mostly unavailable for the weekend, but maybe he can check it out after :-)

NicholasBatesNZ:
Yeah @Cervator that's because someone needs to go through each .block (or write a script) and change the `rotation` property to `family` in all the modules that use these shapes. That's something I haven't done unfortunately. 

_Sent from my Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

engine/src/main/java/org/terasology/world/block/entity/neighbourUpdate/NeighbourBlockFamilyUpdateSystem.java
```diff
@@ -69,6 +71,22 @@ public void largeBlockUpdateFinished(LargeBlockUpdateFinished event, EntityRef e
         }
     }
 
+    /**
+     * notifies the adjacent block families when a block is placed next to them
+     * @param event
+     * @param entity
+     */
+    @ReceiveEvent
+    public void OnBlockPlaced(OnBlockItemPlaced event, EntityRef entity) {
```
asiekierka: Nitpicking, but shouldn't this name be lower-case to match conventions?

engine/src/main/java/org/terasology/world/block/family/BlockSections.java
```diff
@@ -0,0 +1,39 @@
...
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Describes the sections that a block can use such as top, left, right, bottom
+ * but other sections can be used.
+ * <p>
+ *     Each section will override the existing values on the base block when that family section is used. look in
```
skaldarnar: > Each section will override the existing values on the base block when that family section is used

What do you want to express by this? Does it mean that there is only one _active_ isntance of `BlockSections`  per family (or per block?)

engine/src/main/java/org/terasology/world/block/family/FreeFormSupported.java
```diff
@@ -20,8 +20,12 @@
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * Annotation used to mark if a {@link BlockFamily} supports blocks that have freeform shapes.
```
skaldarnar: Where is described what a _ freeform_ is? Should there be a link to that (`@see` could be used for such cross references)

engine/src/main/java/org/terasology/world/block/family/FreeformFamily.java
```diff
@@ -16,27 +16,113 @@
...
  * A freeform family is a pseudo block family that can be combined with any block shape to produce an actual block
  * family.
- *
  */
-public class FreeformFamily {
-    public BlockUri uri;
-    public List<String> categories;
+@RegisterBlockFamily("freeform")
+@FreeFormSupported(true)
+public class FreeformFamily extends AbstractBlockFamily {
```
skaldarnar: Hah, this might be the best spot to better describe what a "freeform family" is - maybe we have some good examples?

engine/src/main/java/org/terasology/world/block/family/MultiConnectFamily.java
```diff
@@ -0,0 +1,234 @@
...
+import org.terasology.world.block.BlockBuilderHelper;
+import org.terasology.world.block.BlockUri;
+import org.terasology.world.block.loader.BlockFamilyDefinition;
+import org.terasology.world.block.shapes.BlockShape;
+
+import java.util.Locale;
+import java.util.Set;
+
+/**
+ * Multi-Connect family describes a block family that will connect to other neighboring blocks.
```
skaldarnar: I can understand what this is about - I like it! Can we have such descriptions (with examples) for the other families as well, please?

Cervator:
[Block families](https://trello.com/c/fkknZWM8/41-block-families)

engine/src/main/java/org/terasology/world/block/family/FreeFormSupported.java
pollend: I'm not quite sure what freeform is.

jellysnake:
Also leaving a note here that #2907 will not be compatible with this and so should probably be closed/updated when this is merged

Cervator:
This has been merged now - prepare for broken things :-)

Going to be updating/merging modules for a while longer. I fixed some minor quality issues via https://github.com/MovingBlocks/Terasology/commit/d0c852d53788e92fc82a56fe0ef4b5347b85e148 and will create a new follow-up issue with the remaining requests from here

Cervator:
Done! At least with what will be handled on this one. Yay!

See #3374 for follow-ups, I've tried to catch all the outstanding issues from here not yet fixed

PR #3347, opened 2018-04-30
Add AdvancedFilter section to SelectModulesScreen v2

### LINKED ISSUE #2935: Allow filtering out library modules from module list

mkienenb:
See 

Add AdvancedFilter section to SelectModulesScreen #3072

Fixes #2935

Reworked to dynamically handle Boolean StandardModuleExtensions.
Dropped extension-specific code.
Dropped "Activate Filters" ui.
Added "Server-side-only" ui (exists as a Boolean StandardModuleExtensions so may as well allow it)

In an ideal PR, the UI checkbox list would be dynamically created, but that was more than I wanted to do to finish off this PR, so there's a hard-coded association table instead.

It's unclear yet if this works for non-local (non-downloaded) mods.  The only time they show up is if all checkboxes are unchecked, but that might be due to there being no StandardModuleExtensions loaded?  Did not investigate.


engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -551,4 +597,32 @@ private void refreshSelection() {
         setSelectedVersions(resolver.resolve(selectedModules));
         updateValidToSelect();
     }
+
+    // TODO: should dynamically generate checkbox list from boolean StandardModuleExtensions rather than hardcoding associations here
```
mkienenb: Should dynamically generate the contents of the checkbox list rather than hoping every Boolean StandardModuleExtension is listed with a matching checkbox id.   I might consider doing that in a future PR.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -359,46 +362,89 @@ public String get() {
...
 
-    private void prepareModuleList(boolean checked) {
-        if (selectModulesConfig.isChecked()) {
-            Iterator<ModuleSelectionInfo> iter = sortedModules.iterator();
-            while (iter.hasNext()) {
-                if (!iter.next().isPresent()) {
-                    iter.remove();
-                }
+    private void advancedModuleFilter() {
+        Iterator<ModuleSelectionInfo> iter = sortedModules.iterator();
```
mkienenb: Not thrilled about using an iterator and iter.remove, but that's what the code was doing before.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -359,46 +362,89 @@ public String get() {
...
-                            selectModulesConfig.setIsChecked(value);
-                            filterText(moduleSearch.getText());
-                            prepareModuleList(value);
-                        }
-                    }
-            );
+     private void filterModules() {
+        sortedModules.clear();
+        sortedModules.addAll(allSortedModules);
+        if(selectModulesConfig.isAnyStandardModuleExtensionSelected()) {
```
mkienenb: This will always happen if there's a Boolean StandardModuleExtension that isn't in the checkbox association table.   Another reason to get rid of the table.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -359,46 +362,89 @@ public String get() {
...
+            advancedModuleFilter();
```
mkienenb: It'd be better to unconditionally run the advanced filter, but the way advancedModuleFilter() is currently implemented, it's hard to do that and still OR the results together.

engine/src/main/java/org/terasology/engine/module/StandardModuleExtension.java
```diff
@@ -47,17 +59,34 @@ public String getKey() {
...
-        Boolean serverSideOnly = module.getMetadata().getExtension(SERVER_SIDE_ONLY.getKey(), Boolean.class);
-        return serverSideOnly != null && serverSideOnly;
+        return getBooleanExtension(module, SERVER_SIDE_ONLY);
     }
 
     public static boolean isGameplayModule(Module module) {
-        Boolean isGameplay = module.getMetadata().getExtension(IS_GAMEPLAY.getKey(), Boolean.class);
-        return isGameplay != null && isGameplay;
+        return getBooleanExtension(module, IS_GAMEPLAY);
     }
```
mkienenb: I dropped the other is*Module() methods which @TheFlash98 added since they aren't used anywhere yet.   Could either add them back in when they are actually used, or just leave them out.

engine/src/main/java/org/terasology/engine/module/StandardModuleExtension.java
```diff
@@ -47,17 +59,34 @@ public String getKey() {
...
+		Set<StandardModuleExtension> booleanPropertySet = EnumSet.noneOf(StandardModuleExtension.class);
+		for (StandardModuleExtension standardModuleExtension : values()) {
+			if (standardModuleExtension.getValueType().equals(Boolean.class)) {
+				booleanPropertySet.add(standardModuleExtension);
+			}
+		}
+		return booleanPropertySet;
+	}
+
+	public boolean isProvidedBy(Module module) {
```
mkienenb: I had a terrible time coming up with a name for this method.   Maybe someone else can think of a better one.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
Cervator: Could be a nice Good First Issue with a nice little write-up after this is merged 👍

engine/src/main/java/org/terasology/engine/module/StandardModuleExtension.java
Cervator: Seems a fair enough name to me. Flows well in its one usage

engine/src/main/java/org/terasology/engine/module/StandardModuleExtension.java
```diff
@@ -47,17 +59,34 @@ public String getKey() {
...
-        return isGameplay != null && isGameplay;
+        return getBooleanExtension(module, IS_GAMEPLAY);
     }
 
     public static SimpleUri getDefaultWorldGenerator(Module module) {
         String ext = module.getMetadata().getExtension(DEFAULT_WORLD_GENERATOR.getKey(), String.class);
         return ext != null ? new SimpleUri(ext) : null;
     }
+
+    private static boolean getBooleanExtension(Module module, StandardModuleExtension ext) {
```
Cervator: IntelliJ + Checkstyle gave me angry red warnings on this file due to super duper evil tab characters instead of spaces. Gasp!

engine/src/main/java/org/terasology/config/SelectModulesConfig.java
```diff
@@ -15,14 +15,46 @@
...
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+
+import org.terasology.engine.module.StandardModuleExtension;
+
 public class SelectModulesConfig {
-    private boolean isChecked;
+	public boolean localOnly;
```
Cervator: Public, yet with getter/setter?

GooeyHub:

GooeyHub:

mkienenb:
Logic changes: Added an "Uncategorized" option that will show all modules which do not have a true boolean extension.   Also disables all of the dynamic categories when enabled.   Possible future work would be to move the dynamic categories into an intended column layout under this option.

From what I can tell in the debugger, the reason all modules disappear when is-local is selected is because the offline modules aren't categorized, and not because the information isn't available on the client.  We will need to classify a few modules to have a testable environment to be certain.

Cleanup changes:

Cleaned up whitespace and formatting using eclipse Terasology config files.
Removed some dead empty comments.
Fixed code that checked for null using an NPE try-catch instead of an if statement.

GooeyHub:

Cervator:
Nice changes - ship it!

I made a tiny label tweak as "Uncategorized modules" was just a slight bit too wide to fit comfortably (and none of the others included "modules" anyway so)

Either of you interested in writing a little issue on how to make the checkbox list dynamic? Could involve a bit of detective work but make a Good First Issue

Issue #2406, opened 2016-07-14
Snow stuffs


Cervator:
Replacement for ancient issue #487 fleshed out a bit more since we have more options now.

- [ ] Add one or more sizes of snow blocks made of slabs (1/8th, 1/4th..) that can be placed. Maybe a full snow block too that doesn't have a dirt side showing (pure snow)
- [ ] Ability to / example of using them during world gen to place on top of snow-covered blocks (add to the [TutorialWorldGeneration](https://github.com/Terasology/TutorialWorldGeneration/) repo/guide?)
- [ ] Maybe experiment with making the smaller slabs penetrable. If a player can jump/climb 2 blocks a tiny added slab of fluffy snow probably shouldn't impede movement
- [ ] To hell with the previous item! Make it risky to climb in snow! Add a chance the player will fall if climbing blocks covered in snow, causing a tiny bit of damage if so
- [ ] Find a way to create snowshoes that make pure snow blocks of any height no longer penetrable (so a player can walk on top of them - this might be tricky)
- [ ] Enhance the [WeatherManager](https://github.com/Terasology/WeatherManager/) module to add snowy weather that deposits snow slabs (I think it has rain, just no graphic for it?)
- [ ] Add logic so that if for whatever reason a snow slab grows to a full block switch it to an actual non-penetrable full snow block. And/or maybe turn heavy enough snow build-up into ice blocks.
- [ ] Maybe allow pure snow slabs/blocks to be "scooped" for modules that have buckets (could then be heated to get water)
- [ ] Wax philosophically about the potential of other slab-based materials that could work similarly (pseudo-liquids like tar?)

Some of these items are probably appropriate for the general backlog (stuff we *could* do, but no real priority) but we do have snow blocks and it is a pretty basic concept so we should further the cause of the white walkers at least a little before archiving the rest. Winter is coming.

omomthings:
Hello, thinking about implementing this block, anything i should know before getting stuck with it haha 😄

skaldarnar:
A good starting point would be the [tutorial on world generation](https://github.com/Terasology/TutorialWorldGeneration). If you haven't already, do have a look at our [wiki](https://github.com/MovingBlocks/Terasology/wiki). In particular the sections [block attributes](https://github.com/Terasology/TutorialAssetSystem/wiki/Block-Attributes), potentially also on [block shapes](https://github.com/MovingBlocks/Terasology/wiki/Block-Shapes-In-Blender).

[...]

PR #3349, opened 2018-05-01
Added snow slabs and snow shoes

ashley-hu:
### Contains

Added snow shoes and snow slabs (1/8, 1/4, 1/2) for #2406 

### How to test

For snow slabs
1. Open the command console
2. Type in 'give snowslabeight'
3. You will receive the 1/8 snow slab
4. Do the same for (1/4, and 1/2) slabs, by typing 'give snowslabhalf' and 'give snowslabquarter'
5. You will receive the respective snow slabs in your inventory

For snow shoes
1. Open the command console
2. Type in 'give snowshoe'
3. The snow shoe will appear in inventory

- [ ] Add snowshoe functionality
- [ ] Add stacking for snow slabs


OvermindDL1:
Shouldn't this be a mod?

_/me is entirely for removing all 'content' from the engine_

Cervator:
Heya @ashley-hu good to see another round of stuff from your group :-)

I can confirm the new blocks work - nice job!

terasology-180502181914-1152x720

However, this time through we're going to have to do a little bit more work, which hopefully will work alright with your course. Often in open source you'll end up with some old issue that has drifted a bit away from how things work in the present, and this is one of those cases. It came up a bit in discussion earlier when I explained about how the "add risk of falling when climbing in snow" part of the issue is obsolete as climbing isn't currently a thing :-)

Stuff we need to look at here:

* You see how you're affecting 14 files, but have 53 listed commits, including a bunch involving other work you've done that's not relevant to this PR? Merging this as-is would add a lot of messy history to our project, which is bad. There are two ways to fix this
  * `git rebase` - this is a technical, fast, and clean option, but it is also tricky and I do not recommend it unless one or more of you is very comfortable with Git :-) In either case I encourage you to read up on it, try some of the GitHub or other tutorials on Git, and maybe play with it *in a local duplicate copy of your workspace* - rebase is one way you can really break your workspace if you aren't sure what you're doing, but Git makes it super easy to work on a copy
  * Simply start a clean branch and copy paste the few related files in, commit once, push to GitHub, and make a new PR (closing this one). This is my recommendation, due to another reason as well. Read on! :-)
* You don't need a unique texture for each block. By default a `SomeName.block` file will look for a `SomeName.png` texture, but as you've done elsewhere you can use block attributes to just point to the same snow texture for every block. So you can delete the duplicate files (or if copy pasting into a new branch just skip those files)
* Use "eighth" in each relevant place to indicate 1/8, not "eight"
* I'm not sure what the "Lightsnow" block is for, it isn't quite configured right. What's the goal with it? 
* Same story with the `slabStack.prefab` which is empty, so adding it doesn't actually do anything :-) You don't need to have one at all, you can just take out that part of the definition. If that's part of a future effort to make stacking work that's cool but should wait until then, generally it is frowned upon to add incomplete work.
* For the snowshoes there are a few issues
  * It might be tricky visually showing the right kind. It looks like you added *boots* not [Snowshoes](https://en.wikipedia.org/wiki/Snowshoe) but it might just be hard to get the icon made right. Feel free to try in a 32x32 pixel size instead if the added detail help make them more visually distinct. If existing artwork is being reused we need to know how it is licensed and ideally add a link somewhere to credit the author.
  * @OvermindDL1 is right, we don't really want content in the engine repo and that's another problem with the old issue - back when it was created this was a fine place to put content but now it is not :-) I should have mentioned that more in the rebooted issue. The slabs could live here in Core at least for now (easy to move later), they'd be useful, and I'm surprised we didn't have any yet. A better home 
 for the snowshoes would be https://github.com/Terasology/Equipment which could also expose you to module development (one day the slabs might move to [StructuralResources](https://github.com/Terasology/StructuralResources)). Might be best to guide via chat but if you run `groovyw module recurse Equipment` then `gradlew idea` you should get that source added locally
  * You added the snowshoes into an item atlas, but for one-off items like that it is way better supplying it as an independent image file plus the associated `.prefab` which would also later be a way to help define the custom functionality. Again the Equipment module should show you a lot of examples on how to do that :-)
  * Adding the actual logic to the snowshoes is indeed quite a bit more complicated. Another part of open source is to better direct things over time. Since the rebooted snow issue is poorly written even now we can try to address a few things here then re-submit new issues elsewhere where they fit better and close the original as partially completed. So try to submit a separate PR with the snowshoe graphic to the Equipment module and after that's merged we can work on writing up a better feature request there on adding the logic later. 

We could also write up an issue somewhere on making snow "stack" better (probably like ingots, swapping to larger slabs) or even make weather effects make snow blocks grow over time (even if you just trigger the functionality with a console command for now)

Let me know if all that makes sense, any questions you have, and which approaches you're interested in following. As usual chat is probably easier when we can find each other online at the same time :-)

dorasun:
Sorry we haven't been responding, we have exams and graduation in the next two weeks, but I will definitely take another look at the texture (I made it, thought it was inspired by the Minecraft diamond boots, so not entirely sure how licensing would work in this case) and making Snowshoes equipment instead of an item. We were actually planning to make the snowshoes look like snowshoes from the get go and I made a bunch of textures:
screen shot 2018-05-05 at 10 11 43 pm
screen shot 2018-05-05 at 10 11 52 pm
but we thought they looked like tennis racquets so we opted for boots instead, but with 32x32 pixels I think i'll have more room to make it look better so I'll work on that when I can. Our class is over so we may or may not be working on this project as actively anymore but we want to at least finish what we started so we'll get to it as soon as we can!

Cervator:
Thanks for the update, best of luck with your exams! You're more than welcome to keep contributing but no obligation nor rush, we'll be here a long time :-)

And yeah true snowshoes are tricky!

jellysnake:
Hey @dorasun @ashley-hu,
Just wanted to check in on if this has had any updates or if you're still planning on working on this in future?

PR #3350, opened 2018-05-01
Improve render graph error detection

mkienenb:
It's currently possible to call connectNode() on a node that has not been added to the render graph without throwing an error.   What happens is implementation-specific according to the general graph documentation, but we should be explicit in how we handle it.

This PR throws an exception but another approach would be to add the node explicitly.


GooeyHub:

eviltak:
Closing due to likely irrelevance. Pinging @dave2s to comment on relevance after the rendering GSoC 2019 project and to coordinate with @mkienenb as needed.

dave2s:
The problem is not relevant anymore. _Connect()_ will be implicitly called only for present nodes after the graph has been created as of #3732. This will happen on the basis of _DependencyConnections_, which are used for explicitly connecting nodes now instead.

PR #3351, opened 2018-05-03
Render graph overlay

mkienenb:
Prototype for a screen to display render graph nodes.


GooeyHub:

mkienenb:
Added stuff which normally gets dumped to the logs to the nodes.   A more clever solution would do more than a multii-lilne text dump, but it's a start.

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/ingame/metrics/RenderGraphOverlay.java
```diff
@@ -0,0 +1,160 @@
...
+    @Override
+    public void initialise() {
+        WidgetUtil.trySubscribe(this, "close", widget -> getManager().closeScreen(RenderGraphOverlay.this));
+        renderGraphTreeView = find("renderGraphTreeView", UITreeView.class);
+        renderGraphTreeView.subscribeNodeClick((event, treeNode) -> {
+            if (event.getMouseButton() == MouseInput.MOUSE_LEFT) {
+                RenderNodeInfo renderNodeInfo = (RenderNodeInfo)treeNode.getValue();
+                Node renderNode = renderNodeInfo.getNode();
+                String uriString = renderNode.getUri().toString();
+                logger.info(uriString);
```
vampcat: Is there a particular reason you're first fetching the node and then calling `toUri` on it, instead of `String uriString = renderNodeInfo.toString(); ` (which also takes care of the `node == null` case?)

mkienenb: I wasn't sure exactly what I wanted to happen on mouse-click yet.   I was kind of thinking a pop-up panel with more information about the node, but as a placeholder to detect the click and make sure it was working right, I decided to dump something to the console.    

The fact that renderNodeInfo.toString() currently shows the url or ROOT isn't behavior I planned to keep.   toString() traditionally dumps the contents of an object in a format that the developer can use for debugging, but isn't recommended for actual program logic, so I'll very likely change it later to also dump input and output information as well as the uri.   I really hate the root node = null situation.   I'd like to do away with it entirely.

But the answer for this particular section of code is that it's all going to be deleted "soon".

engine/src/main/resources/assets/ui/ingame/pauseMenu.ui
```diff
@@ -100,6 +100,21 @@
                         "offset": 8
                     }
                 }
+            },
+            {
+                "type": "UIButton",
+                "text": "RenderGraph",
```
mkienenb: Needs localization.

eviltak:
@dave2s: Can you coordinate with @mkienenb and see how this can be adapted (if needed) to the new render graph?

dave2s:
@eviltak @mkienenb  This PR is certainly of interest and should stand as one of gsoc19's followups. I've tested it before, but I didn't try to adapt it. I remember I was having some errors, trying to display the UI. Regarding the rendergraph itself, the representation is still the same so this might not need much adapting.

pollend:
looks like I blew up the PR. I saved the changes in this branch: https://github.com/MovingBlocks/Terasology/pull/4515

Issue #3354, opened 2018-05-07
Add label + filter reset button for new module categories


Cervator:
With #3347 merged we now have a filter listing for module categories. But I realized we don't have a way to reset the filters to default and the area could use a nice label. Should be fairly easy to do. Quick mockup:

![2018-05-07 18_23_37-terasology _ alpha](https://user-images.githubusercontent.com/1063833/39728295-4ead2736-5224-11e8-8d5f-1a0e46787adf.png)

So in other words move the checkboxes down a bit, add the indicated label above them, and add a new button that'll reset the filters to default below the checkboxes. Piece of cake! :-)

See https://github.com/Terasology/TutorialNui/wiki for a general intro to our UI system

jellysnake:
Closing as fixed in #3354

PR #3357, opened 2018-05-11
Adding Reset Filters button and module categories title (fix for #3354)

dacharya64:
### Contains
Fix for #3354, adds a label for the new module categories and a reset filters button that clears out all checked filters. 

### How to test

Load up game, open up module select menu, and try selecting / resetting advanced filters.

### Outstanding 
* Translations needed for the new button and title

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
@@ -309,7 +309,7 @@ public class SelectModulesScreen extends CoreScreenLayer {
                     public Boolean get() {
                         ModuleSelectionInfo info = moduleList.getSelection();
                         return info != null && info.isPresent() && !isSelectedGameplayModule(info)
-                               && (info.isSelected() || info.isValidToSelect());
+                                && (info.isSelected() || info.isValidToSelect());
                     }
                 });
                 toggleActivate.bindText(new ReadOnlyBinding<String>() {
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -407,13 +407,56 @@ public String get() {
                 }
                 filterModules();
             });
+
+            UIButton resetAdvancedFilters = find("resetFilters", UIButton.class);
+            filterModules();
+            if (resetAdvancedFilters != null) {
+
```
mkienenb: It seems like we have a lot of duplicated code now for keeping checkboxes in sync with the data.
If I remember correctly, it's possible to bind a NUI control to a data source.   I recommend that we do that rather than explicitly setting both the data source and then the checkbox.  Once weve bound the checkbox to the appropriate data source, all we should have to do is set the value on the data source.

I also think that the default values on the data source should be set by a method in the data source.   Right now you have to parse code to determine what the default values are.   (Figure out what the state of the data is at the time of the if statement that toggles the value....)

Instead, we should call selectModulesConfig.resetToDefaultValues().   Then the checkboxes should pick up the changes through bindings automatically.

engine/src/main/resources/assets/i18n/menu_en.lang
```diff
@@ -244,6 +244,8 @@
     "select-game-title": "Select Game",
     "select-modules": "Modules...",
     "select-modules-title": "Select Modules...",
+    "select-modules-filter-reset": "Reset Filters",
```
mkienenb: Maybe "Reset to Defaults" or "Reset to Default Values" might be more obvious?

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
mkienenb: Yes, I tested this and it is much cleaner this way.

For instance, the ui code for local-only can be reduced to this.  I also fixed the getter/setter to work with boolean not Boolean:
`
            UICheckbox localOnlyCheckbox = find("localOnlyCheckbox", UICheckbox.class);
            localOnlyCheckbox.bindChecked(BindHelper.bindBeanProperty("localOnlySelected", selectModulesConfig, boolean.class));
            localOnlyCheckbox.subscribe(e -> {
                filterModules();
            });
`

mkienenb: Creating a binding for the other dynamic checkboxes will require a custom binding subclass (modification of BeanBinding).    

The button action for uncategorized with a boolean bean binding needs more thought.   Probably the logic in the button action would be moved inside the selectModulesConfig.setUncategorizedSelected() method.

I can apply this PR as-is (once the extra filterModule(); method call is removed) if you are not interested in changing the code to use bindings.  I can go ahead and do that in a separate PR.

dacharya64:
I removed the unnecessary filterModues() call. I don't think I'll be able to work on this any more for now, though, since I'm working on my Google Summer of Code project.

mkienenb:
Sounds good.  I will merge this when I get a chance.

PR #3359, opened 2018-05-12
synchronize moudles with advanced filters on window open.

mkienenb:
### Contains

Fixes problem with modules screen showing full list of modules after window has been opened.
Fixes problem with modules list showing full list of downloaded modules immediately after window is opened.

### How to test

Before:
Opening the modules panel shows the local modules, then all modules once the remote modules download.  Neither view considers advanced filter settings.

After:
Open modules screen.  Note that listed modules matches advanced filter settings before and after the remote modules download.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
@@ -106,6 +106,7 @@ public class SelectModulesScreen extends CoreScreenLayer {
         }
 
         refreshSelection();
+        filterModules();
     }
 
     @Override
@@ -570,6 +571,7 @@ public class SelectModulesScreen extends CoreScreenLayer {
                 info.setOnlineVersion(remote);
             }
         }
+        filterModules();
     }
 
     @Override


engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -570,6 +571,7 @@ private void updateModuleInformation() {
                 info.setOnlineVersion(remote);
             }
         }
+        filterModules();
```
mkienenb: This one fixes the problem with the list updating after the remote modules have downloaded.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/selectModulesScreen/SelectModulesScreen.java
```diff
@@ -106,6 +106,7 @@ public void onOpened() {
         }
 
         refreshSelection();
+        filterModules();
```
mkienenb: This one fixes the problem with the list updating before the remote modules have downloaded.

GooeyHub:

TheFlash98:
Can confirm that this works. :)

PR #3368, opened 2018-05-22
IntelliJ Tweaks and Fixes

jellysnake:
* Adds a fix for `@In` not correctly marking a variable as assigned if the variable was private
* Changes the year in the copyright to always be the current year
* Adds a file template for creating a System.


jellysnake:
Yup. If I ever get around to working on the intelliJ plugin we have then that's the better place to introduce them. You can also include code templates. I've got one for a new event handler and for loggers

PR #3369, opened 2018-05-22
Add method to set entity at location

jellysnake:
This PR adds a method to `BlockEntityRegistry` that sets the entity at a given location. I've added it to this class because it already keeps track of entities linked with positions in the world.
This is something I wish to use for the multiblock style towers in my Tower Defense GSOC project but I can also see applications for it in modules like Simple Farming.

I also wish to note that this doesn't break the API as it only adds a method

### DIFF
--- a/engine/src/main/java/org/terasology/world/BlockEntityRegistry.java
+++ b/engine/src/main/java/org/terasology/world/BlockEntityRegistry.java
@@ -28,6 +28,15 @@ import org.terasology.world.block.Block;
 public interface BlockEntityRegistry {
 
     /**
+     * Sets a new entity at the given position.
+     * This new entity is not temporary and will overwrite any existing entity at this point.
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Merged, thanks! Although I just realized you had later added the WIP label so maybe I shouldn't have merged yet?

Fun fact: it *is* actually an API violation, but one that's very easy to turn into a minor enhancement: `BlockEntityRegistry` is API so adding a non-default method could lead to an implementer in a module failing as it no longer meets the contract defined by the interface.

In this case I checked a mega-workspace real quick and as expected there are no other implementers, so it is something we can forgive pretty easily at the moment :-)

The outright fix that would have avoided an API violation would be to provide a default implementation in the interface, that would have worked too. But we're in API free fire still so not a problem in either case!

PR #3337, opened 2018-04-23
Display held items of remote players in multiplayer sessions.

dave2s:
### Tackled issues: #3280

_Based on @Josharias 's firstPersonClientSystem introduced mid 2016, practically clones his changes._

[image summary](https://pasteboard.co/Hi09xqN.jpg)

### **Added java classes:**
**ThirdPersonRemoteClientSystem.java** - Essential class which glues all together. Handles displaying held items for all remote players in multiplayer session. Locally, registered for clients.
**RemotePersonHeldItemTransformComponent.java** - Component represents transformations of various held item types (used in hand,itemBase and blockItemBase prefabs)
**RemotePersonHeldItemMountPointComponent.java** - Component represents mounting point of characters for held items to attach to.
**ItemIsRemotelyHeldComponent.java** - This is used on the client side to track and clean up items that are no longer held by remote players, but are still location linked

### **Added prefabs:**
**remotePersonHeldItemMountPoint.prefab** - represents RemotePersonHeldItemMountPointComponent.java

### **Modified prefabs:**
**blockItemBase.prefab,itemBase.prefab,hand.prefab** - added remotePersonHeldItemTransform components to handle transformations of these items in game
**player.prefab** - added remotePersonHeldItemMountPoint component to attach all the held items to. Corresponds to same named .java

### **How to test and try out**
Host a server, run another instance, join the server, run around, change held items, pickup blocks, observe one another holding different types of held items (empty hand, blocks, weapons, torches)...

Run multiple instances of the game, host a multiplayer session, connect and disconnect players and watch their items. - ❗️  More than two players tends to crash often.

### **ToDo:**
- [ ] UnitTests

### Outstanding before merging ❗️ 
 
- [ ] **Client sometimes logs in with a detached camera spawned at 0 0 0.**

- [ ] **Player's who happen to log in always see hands of other players or nothing until these other players perform change of the held item while the logged in player is in the game.**

- [ ] **Lingering previous held items**

- [ ] Occassional nullpointer crashes while: - simultaneously joining server, - on disconnecting of crashed player. Needs more testing and more eyes :) I've changed these mechanics so many times I can't see clearly.

PR #3329, opened 2018-04-17
Fix to online players overlay not displayig all of the players

lgzy:
### Contains

Fixes #3318 
For some reason in function `determinePlayerAndPing` `allClients` would not contain all of the clients sometimes. I did not figure out why was that, but instead I used client entities directly from `pingMap`. It didn't contain client local to the server, but I just put it into the map with 5ms ping.

### How to test

- Start two game clients
- Have one host and the other join
- Hold down TAB and observe both names listed

PR #3379, opened 2018-06-04
Experimental merge of #3337 + #3329 with some fixes and extra logging

Cervator:
I wanted to see if PRs #3337 + #3329 could be made to work well enough to merge as part of Alpha 10 as they are quite useful. I fixed all the crashes I could find and can no longer trigger game problems. However some of the fixes are bandaids over underlying bugs.

* #3329 might need a better strategy for listing players than having the logic run yet get ignored solely on headless servers
* #3337 might also need strategy adjustments as its caching of held items seems to have issues. Even weirder *I cannot* clean up the `charactersHeldItems` map for some reason when I try to remove an entry left behind in player 1's map when a player 2 leaves.

However, the game works in both host/client + client and headless + client + client. The player online overlay works fine (still has a slight hiccup on a first frame or during joins but it functions). The held items show correctly at all times in all scenarios I've tested and don't provoke orphaned items in the world, however in some cases (player 2 leaves and rejoins in host/client + client, or a 1st player on a headless server) a dupe held item appears locally. It appears to be the held item meant to show for remote players (one kinda floats near where you'd expect it to be seen externally, while there's also the more typical static 2D version present)

In any case I may still merge this for the value it provides, even knowing the code could be better. Still, ideally we'd find the cleaner approaches and fix the remaining dupe item issue.

Making PR to get an experimental build to drop on a server for more thorough testing. May be of particular interest to @dave2s and @Josharias 

For slightly more details: in host/client + client If you exit a session with a held item it gets stuck held next session. Affects equally host and client separately. Empty hand doesn't get stuck (clears the issue if you relog). If the host restarts while player 2 has an item stuck then on rejoining player 2's item is unstuck even without empty hand.


GooeyHub:

Cervator:
I enabled archiving of the game zip in the PR job then reset the event server to run using that build + HumanoidCharacters to also check that out.

http://jenkins.terasology.org/job/TerasologyPRs/2240/artifact/build/distributions/Terasology.zip

Zip link will last only about a week or so before it gets pruned, but then the event server will poof before that.

Observations:

* I *do* see orphaned items now, visible only to a client that stays online while another disconnects. I don't know if maybe my tweaking helped cause that.
* Having HumanoidCharacters enabled but not in the zip may be causing some quirks. The module will auto-download and work if you connect to the server, but you don't seem to get the customization dialog. Going back to the main menu then trying to reconnect or something like that got me a quirky warning about a module download failure. Restarting the client fixed it
* While in-game and still having the customization dialog up it seems like the colors take effect on a *different* visible player as well. Not sure what's up with that

Further testing and review appreciated.

Cervator:
The mystery thickens.

Have found two issues:

* Even in a single player game one of the local player checks fails: an `EntityRef` being compared in `linkHeldItemLocationForRemotePlayer` to the various entities connected to `LocalPlayer` doesn't match any as actually the entity's `CharacterComponent`'s `controller` (the client) is what matches the local player client. With that check in an additional route into adding things that shouldn't be added is avoided (but I'm not sure that caused any trouble)
* In a headless multiplayer game the local player _doesn't show the actual local player_ - it ends up referring to a `NullEntityRef`. Naturally checks against that fail and even the local player passes, thus ending up with a drawn held item in 3D *in addition* to the simple 2D representation

Exactly how to reliably detect who you are in that case is next, but 1 am right now. To be continued!

Cervator:
It woooooooooorks! Well, it already worked. But now it _almost fully_ works. Yeah. 

No more duplicated held item in some cases for the local player, due to more thorough local player detection. No more orphaned items on logout and no cache kept between ticks to go dirty. May be slightly less performant and everything can probably be improved further, but the fundamentals are solid now and there's a little less code as well :-)

A few outstanding things to consider:

- [ ] Check builder sample gameplay (doesn't have held items at all, but this code is in engine not Core so it would be active even in BSG)
- [ ] Known issue still there: empty held item until first action in some cases
- [ ] Known issue: not handling empty hand right. Used to simply replace with a floating hand, but that looks kinda awkward on the humanoid characters since they, well, already have hands. Plus the texture doesn't really work right in 3D and that let me not worry about a second cache of hands. ✋ Anyway right now left it out but that ends up keeping the prior item until the next valid item. Need to fix but it is nearly 1 am again - *probably* not a difficult fix.
- [ ] Haven't tested the commands, although I did update them to the new approach

Feedback and testing hugely appreciated. I tested both single player, server/client + client multiplayer, and headless + 2 clients. Pinging @dave2s and @Josharias in particular

GooeyHub:

Cervator:
Two more fixes are in: empty hand is now handled correctly and I've partially fixed initial lack of held items.

Fixed the second part first try, think I'm getting the hang of this! Think this is the most complicated code I've written in years :p

Details on partial fix: I hooked up a check in a `postBegin` that catches all other active players as you log in, triggering the creation of their held item entities. However, that is a one-time thing for the logging in client. It doesn't cover when it is somebody else logging in for the first time. They still have to change their held item before it shows up for already-connected players.

Fixing that should be doable too, but at some point I should probably call this good enough and merge it so I can get back to that whole "Oh yeah need to release Alpha 10" thing

* Find a suitable event that is sent to clients when a new client connects. Difficulty: something like `ConnectedEvent` is only sent to the connecting client. If there isn't an event yet then it should be easy enough to catch something an authority system that then would send a unique new event to all other clients to let them know "Hey somebody new joined!"
* Add logic to the main system's `update()` where if an existing held item for a given remote player doesn't exist then go see if they have one equipped and if so trigger `linkHeldItemLocationForRemotePlayer` ... which actually might be easy enough and the way to go even if it a slight bit more stuff happening in a loop rather than one-time via event. But lunch first.

GooeyHub:

Cervator:
Final fix is in! Commit message has the details.

Much could still be improved, but as cool as it feels to spend some time with serious code I really should be doing organizing instead. I'll write up some follow-up issues sometime with more ideas :-)

GooeyHub:

PR #3383, opened 2018-06-05
New create game flow backend | Part 2

TheFlash98:
### Contains
The backend of the [new create game flow](https://goo.gl/hizreb). The player can now select different world generators and add a number of worlds to the game. Each world has a unique seed and the seed you see on the `AdvanceGameSetupScreen` is for the Universe. Each world is previewable individually on the `WorldPreGenerationScreen`. You have the option to configure and change the properties of your world in both `UniverseSetupScreen` and `WorldPreGenerationScreen`. There is a configure button which leads you to the `WorldSetupScreen`  where properties can be edited and the changes are successfully previewed in the `WorldPreGenerationScreen`.

### Outstanding before merging
- [ ] Making a world with Height Map causes a crash.
- [x] Game title (?) 
 
A video of what's done: https://www.youtube.com/watch?v=qhKGQjEONmA

PR #3385, opened 2018-06-08
New create game flow backend | Final steps

TheFlash98:
### Contains
Nothing new to review in this PR compared to #3383 I've only addressed the review by vampcat. Will be pushing the last few things needed for the create game flow here.

- [ ] HeightMap causes a crash
- [ ] Better way to update preview when zoom slider value changes 

Could use some help in the above-mentioned issues. :)

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/StartPlayingScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/StartPlayingScreen.java
@@ -41,6 +41,8 @@ import org.terasology.world.time.WorldTime;
 
 public class StartPlayingScreen extends CoreScreenLayer {
 
+    public static final ResourceUrn ASSET_URI = new ResourceUrn("engine:startPlayingScreen");
+
     @In
     private ModuleManager moduleManager;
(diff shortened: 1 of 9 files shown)


TheFlash98:
[theflash98 & multi-world](https://trello.com/c/dYU1iOul/38-theflash98-multi-world)

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/StartPlayingScreen.java
```diff
@@ -85,7 +86,7 @@ public void initialise() {
 
             SimpleUri uri = world.getWorldGeneratorInfo().getUri();
             // This is multiplied by the number of seconds in a day (86400000) to determine the exact  millisecond at which the game will start.
-            float timeOffset = 0.50f;
+            final float timeOffset = 0.50f;
```
Cervator: Unsure if @vampcat meant to make this a *constant* rather than just final. If so it would be all caps and so on as a class level variable :-)

GooeyHub:

PR #3386, opened 2018-06-08
Vector pool

Adrijaned:
### Contains

Creates a Vector(3i) memory management model to lower a bit usage of memory and ease the work for garbage collector.

### How to test

Launch game, ideally create new world, let it run for a little while, then use console command `reportVectorUsage`. You should see logs from VectorPool dumped to stdout, the first number being amount of vectors drawn from vectorPool, and the second amount of vectors returned to it. Amount of vectors drawn is higher than amount of vectors freed, that is because not everything returns its vectors (yet). With Perlin worldGen, I was able to get to about 7_000_000L vectors freed in about 10 seconds, which makes about 100 MBytes of memory saved, that gc has not to free himself.

Not everything (actually, only a bit of things) use VectorPools now, and amount of vectors drawn is way higher than the amount of vectors freed. This however serves as a nice proof of concept.

### DIFF
--- a/engine/src/main/java/org/terasology/math/VectorPools.java
+++ b/engine/src/main/java/org/terasology/math/VectorPools.java
@@ -19,7 +19,7 @@ import org.terasology.math.geom.Vector3i;
 
 public class VectorPools {
 
-    private static final int POOL_SIZE = 100;
+    private static final int POOL_SIZE = 100_000; //Might be a bit overkill, but some might methods return array of all Vectors in chunk
 
     private static int currentVector3isAvailable;
     private static final Vector3i[] VECTOR_3I_POOL = new Vector3i[POOL_SIZE];
--- a/engine/src/main/java/org/terasology/math/VectorPools.java
+++ b/engine/src/main/java/org/terasology/math/VectorPools.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2018 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.math;
+
+import org.terasology.math.geom.Vector3i;
+
+public class VectorPools {
+
+    private static final int POOL_SIZE = 100;
+
+    private static int currentVector3isAvailable;
+    private static final Vector3i[] VECTOR_3I_POOL = new Vector3i[POOL_SIZE];
+
+    private VectorPools() { }
+
+    public static synchronized Vector3i getVector3i() {
+        if (currentVector3isAvailable > 0) {
+            VECTOR_3I_POOL[currentVector3isAvailable - 1].set(0, 0, 0);
+            return VECTOR_3I_POOL[--currentVector3isAvailable];
+        }
+        return new Vector3i();
+    }
+
+    public static synchronized Vector3i getVector3i(Vector3i other) {
+        if (currentVector3isAvailable > 0) {
+            VECTOR_3I_POOL[currentVector3isAvailable - 1].set(other);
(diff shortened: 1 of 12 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

@Adrijaned: bring on the PRs! /me shoves @vampcat in front

GooeyHub:

jellysnake:
Ooh nice! An optimisation PR.
I'll definitely be checking this one out for testing.

MarcinSc:
I'd suggest at least testing, if pooling of objects actually improves performance. Current JVMs do a lot of optimizations behind the scenes, so it might turn out, that introducing a pool, actually lowers performance.

vampcat:
@MarcinSc : I've looked into TS memory usage, and garbage collection *is* a problem in TS. There are enough vectors being created every frame that the memory usage spikes up from 100 mb to 500 mb in 3-4 seconds - then falls back to 100 due to gc kicking in. Having that change to a constant 150 mb doesn't sound so bad to me.

engine/src/main/java/org/terasology/math/VectorPools.java
```diff
@@ -0,0 +1,58 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.math;
+
+import org.terasology.math.geom.Vector3i;
+
+public class VectorPools {
```
vampcat: The class should be named `VectorPool`, without the s.

(it's SolShip, never SolShips - unless it is a collection of SolShips. In this case, this is a collection of vectors, not a collection of vectorPools, so..)

engine/src/main/java/org/terasology/math/VectorPools.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private static final int POOL_SIZE = 100;
+
+    private static int currentVector3isAvailable;
+    private static final Vector3i[] VECTOR_3I_POOL = new Vector3i[POOL_SIZE];
```
vampcat: I think this should be `Vector3iPool` and not `VECTOR_3I_POOL`.

(Note that this variable is a constant reference, and not a constant itself. A small - but important - distinction)

engine/src/main/java/org/terasology/math/VectorPools.java
vampcat: Also, let's make this class more generic , that can be mixed with Templates (so that we can have `VectorPool<Vector3i>` or `VectorPool<3f>`) or inherited (to get `VectorPool3i extends VectorPool<Vector3i>` or `VectorPool3f extends VectorPool<Vector3f>`)? Makes the class way more extensible and flexible.

engine/src/main/java/org/terasology/math/VectorPools.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private VectorPools() { }
+
+    public static synchronized Vector3i getVector3i() {
```
vampcat: In line with my previous comment, this will become `getVector()`

engine/src/main/java/org/terasology/math/VectorPools.java
```diff
@@ -0,0 +1,58 @@
...
+            VECTOR_3I_POOL[currentVector3isAvailable - 1].set(x, y, z);
+            return VECTOR_3I_POOL[--currentVector3isAvailable];
+        }
+        return new Vector3i(x, y, z);
+    }
+
+    public static synchronized void free(Vector3i v) {
+        if (currentVector3isAvailable < POOL_SIZE) {
+            VECTOR_3I_POOL[currentVector3isAvailable++] = v;
+        }
```
vampcat: How about handling the underflow? :D

(more specifically, why DO we have a POOL_SIZE?)

vampcat:
/me doesn't know why he's being thrown under the bus.

MarcinSc:
@vampcat But does it actually improve performance? Let's imagine that object initialization and GC takes 5% of CPU over some extended amount of time, while pooling, thread synchronization required, etc takes 6% of CPU. Yes, you will be churning through less memory with a pool of objects, but you might be using more CPU/time.

pollend:
could we move this to TeraMath?

Adrijaned:
Moved to TeraMath (https://github.com/MovingBlocks/TeraMath/pull/52)

PR #3387, opened 2018-06-09
cleanup work in org.terasology.math

asiekierka:
### Contains

Cleanup work in org.terasology.math. Primarily marks fields which are effectively final as "final", but also performs more advanced cleanup/optimizations in Rotation.

### How to test

If the game turns on, it's probably fine.

- [ ] Should Transform.scale not also be final? Would changing it to final break any existing module?

### Outstanding after merging

- [x] ~~IntegerRange should be replaced with Guava's Ranges as they are much more flexible, but this is best fit for a separate PR~~ Nevermind, it would be better to improve IntegerRange.

### DIFF
--- a/engine/src/main/java/org/terasology/math/AABB.java
+++ b/engine/src/main/java/org/terasology/math/AABB.java
@@ -452,11 +452,11 @@ public final class AABB {
     }
 
     public boolean intersectRectangle(Vector3f from, Vector3f direction) {
-        Vector3f dirfrac = new Vector3f();
-
-        dirfrac.x = 1.0f / direction.x;
-        dirfrac.y = 1.0f / direction.y;
-        dirfrac.z = 1.0f / direction.z;
+        Vector3f dirfrac = new Vector3f(
+                1.0f / direction.x,
+                1.0f / direction.y,
+                1.0f / direction.z
+        );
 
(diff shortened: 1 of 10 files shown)


GooeyHub:

engine/src/main/java/org/terasology/math/AABB.java
```diff
@@ -452,11 +452,11 @@ public int hashCode() {
     }
 
     public boolean intersectRectangle(Vector3f from, Vector3f direction) {
-        Vector3f dirfrac = new Vector3f();
-
-        dirfrac.x = 1.0f / direction.x;
-        dirfrac.y = 1.0f / direction.y;
-        dirfrac.z = 1.0f / direction.z;
+        Vector3f dirfrac = new Vector3f(
```
vampcat: Let's rename dirfrac to dirFrac, shall we?

(i realize you did not introduce this variable in the first place, but it never hurts to sprinkle some code-quality-dust over areas that you visit)

GooeyHub:

asiekierka:
Note: The "outstanding" question is probably best answered by someone more experienced with the project.

Cervator:
Found time to test this and the game did indeed still turn on. Merged :-)

PR #3389, opened 2018-06-10
fixed rotation issue with freeform family and added more detailed logging information for BlockFamily

pollend:
I just reworded the block name to avoid confusion and also added a bit more logging information that helped diagnoses the issue.

old:
Registered BlockFamily[core:brick:engine:stair]

new:
Registered BlockFamily[freeform,core:brick:engine:stair]


GooeyHub:

Cervator:
Thanks! This seems to have done the trick :-)

PR #3391, opened 2018-06-11
Fix NPE after releasing v2 Core

ar0ne:
### Contains

Fix NPE for DetailsGameScreen after releasing V2. 
After release we don't have an old version of some modules in our classpath. In this case all your saved game with the old core version can't be showed correctly on this screen.

See Core, BuilderSampleGameplay, CoreSampleGameplay.

### How to test

1. Open DetailsGameScreen for any save with old core version.
2. See log for appropriate message.
3. Check that they aren't in list of modules.


GooeyHub:

Cervator:
Thanks for the quick fix! :-)

As noted on Slack I think we could improve this to still add a module button, but color the text red if the version wasn't found. If an older version exists maybe color it yellow then put a disclaimer in the info box that the details shown are for an older version?

PR #3388, opened 2018-06-09
New create game flow, final touches

TheFlash98:
### Contains
Disabled the time progression slider and world shape drop down. Changed the UI of the `WorldSetupScreen` and removed the Message Popup which appeared by clicking "Re-Roll".

PR #3393, opened 2018-06-14
Fix UI Slider using listener

nihal111:
@TheFlash98 I hate temp

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldPreGenerationScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldPreGenerationScreen.java
@@ -53,9 +53,10 @@ import java.util.stream.Collectors;
  * {@link UniverseSetupScreen}. Each world is still configurable and its seed
  * can be changed by the re-roll button. Note that each world has a unique seed.
  */
-public class WorldPreGenerationScreen extends CoreScreenLayer {
+public class WorldPreGenerationScreen extends CoreScreenLayer implements UISlider.UISliderOnChangeListener {
 
     public static final ResourceUrn ASSET_URI = new ResourceUrn("engine:worldPreGenerationScreen");
+
 
     @In
     private ModuleManager moduleManager;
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
```diff
@@ -265,4 +270,12 @@ public void setPrecision(int precision) {
     private int pixelOffsetFor(float val, int width) {
         return TeraMath.floorToInt(width * (val - getMinimum()) / getRange());
     }
+
+    public interface UISliderOnChangeListener {
```
Cervator: An inner interface *and* it is followed by a regular class method? That's what you get for rushing! :-)

This even triggers a Checkstyle warning about the order in which stuff is written. Just for clarity could you entirely extract the interface instead? Sure it'll be tiny but files are cheap. Then it can get some javadoc too

engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
```diff
@@ -56,6 +58,9 @@ public boolean onMouseClick(NUIMouseClickEvent event) {
         @Override
         public void onMouseRelease(NUIMouseReleaseEvent event) {
             active = false;
+            if (uiSliderOnChangeListener != null) {
+                uiSliderOnChangeListener.onSliderMouseDown(getValue());
```
Cervator: This naming seems confusing. `onMouseRelease` seems like the right phase, but it is calling an _onDown_ method? Shouldn't the listener method be named similarly? `onMouseRelease` as well or `onChangeTriggered` perhaps.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldPreGenerationScreen.java
```diff
@@ -114,18 +115,7 @@ public void setEnvironment(Context subContext) throws UnresolvedWorldGeneratorEx
...
-                return zoomValue;
-            }
-
-            @Override
-            public void set(Float value) {
-                zoomValue = value;
-                updatePreview();
-            }
-        });
+        zoomSlider.setUiSliderOnChangeListener(this);
```
Cervator: There's a conflict here although it seems minor. Didn't notice if related but `zoomSlider` here hides a class instance variable. Not necessarily something you caused but could you clarify that if possible?

Cervator: And actually I think I end up getting a related NPE while testing although it could be from a poor conflict fix

nihal111:
@cervator Yes indeed, agreed.
This PR was never meant to be merged though, that's why no PR description either. I wanted to give @theflash98 a rough direction and multiple files on pastebin would've been ugly, so I opened up this PR real quick.

nihal111:
Closing this since #3388 is merged anyway. @theflash98 could you please follow up another PR to fix this?

Issue #3310, opened 2018-04-08
Using F5 (show BT editor) in-game during multiplayer causes crash via NPE


Cervator:
### What you were trying to do

Open the behavior tree editor in-game via F5

### What actually happened

Client crashed with an NPE

### How to reproduce

* Client connected to a multiplayer server with Behaviors active (assumed, might take less)
* Press F5

### Log details and game version

Omega 900 during today's play test

```
14:03:27.201 [main] ERROR o.terasology.engine.TerasologyEngine - Uncaught exception, attempting clean game shutdown
java.lang.NullPointerException: null
	at org.terasology.rendering.nui.TextLineBuilder.addText(TextLineBuilder.java:56)
	at org.terasology.rendering.nui.TextLineBuilder.getLines(TextLineBuilder.java:51)
	at org.terasology.rendering.nui.TextLineBuilder.getLines(TextLineBuilder.java:46)
	at org.terasology.rendering.nui.itemRendering.StringTextRenderer.getPreferredSize(StringTextRenderer.java:69)
	at org.terasology.rendering.nui.widgets.UIList.getPreferredContentSize(UIList.java:117)
	at org.terasology.rendering.nui.internal.CanvasImpl.calculateRestrictedSize(CanvasImpl.java:347)
	at org.terasology.rendering.nui.internal.CanvasImpl.calculatePreferredSize(CanvasImpl.java:333)
	at org.terasology.rendering.nui.layouts.ScrollableArea.getPreferredContentSize(ScrollableArea.java:193)
	at org.terasology.rendering.nui.internal.CanvasImpl.calculateRestrictedSize(CanvasImpl.java:347)
	at org.terasology.rendering.nui.internal.CanvasImpl.calculatePreferredSize(CanvasImpl.java:333)
	at org.terasology.rendering.nui.layouts.miglayout.MigComponent.calculatePreferredSize(MigComponent.java:41)
	at org.terasology.rendering.nui.layouts.miglayout.MigLayout.layoutContainer(MigLayout.java:223)
	at org.terasology.rendering.nui.layouts.miglayout.MigLayout.getPreferredContentSize(MigLayout.java:182)
	at org.terasology.rendering.nui.layouts.miglayout.MigLayout.getMaxContentSize(MigLayout.java:188)
	at org.terasology.rendering.nui.internal.CanvasImpl.calculateMaximumSize(CanvasImpl.java:362)
[...]

PR #3394, opened 2018-06-14
Removed Behavior editor button from Main Screen

ar0ne:
### Contains

Removed Behavior Editor button, because it doesn't supposed to work without Behaviors module. 

See details in #3310 .

### How to test

Open MainMenu and check that we don't have this button anymore.


GooeyHub:

PR #3395, opened 2018-06-14
Rewrite modules tab

ar0ne:
### Contains

Small changes for GameDetailsScreen. Includes highlighting for invalid modules (version from manifest.json isn't in user's classpath). 

### How to test

After release of V2 we increased Core version from 2.0.0-SNAPSHOT to 2.0.1-SNAPSHOT.
Basically, you need old save with previous version. In this case when you'll open the screen, you should see highlighted "Core Content" module on Modules tab and warning in description.


GooeyHub:

Cervator:
Tests out 👍 

I think we could still distinguish between modules missing (red, no info available) vs wrong version (yellow, list what we have) but that can be done later. Maybe a backlog card?

At present a non-existing module (just add a gibberish one into the save manifest) results in a logged error but no indication in the game details that there's a problem (the module gets listed in the overview, but not on the module tab)

PR #3396, opened 2018-06-16
Make render logging a little quieter

jellysnake:
Currently the `RenderTaskListGenerator` class outputs the entire `INTERMEDIATE RENDERER LIST` which is really long. I've moved to to debug rather than info which makes the log easier to search through and use as this list can get very long

### DIFF
--- a/engine-tests/src/test/java/org/terasology/rendering/dag/RenderTaskListGeneratorTest.java
+++ b/engine-tests/src/test/java/org/terasology/rendering/dag/RenderTaskListGeneratorTest.java
@@ -17,11 +17,11 @@ package org.terasology.rendering.dag;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
-import java.util.Set;
 import org.junit.Test;
 import org.terasology.engine.SimpleUri;
 
 import java.util.List;
+import java.util.Set;
 
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Looks like 3 test failures, see http://jenkins.terasology.org/job/TerasologyPRs/2295/testReport :-)

Pinging @vampcat for a quick sanity check

jellysnake:
I've taken a look at this and it seems to a small issue related to a line of logging not displaying. As everything in the log output is thus one line higher it's not properly passing the test.

Cervator:
To fix the test should we find some way to enable debug logging during execution of that test then?

Also pinging @vampcat again. Poke poke :-)

Cervator:
> Also, for the test : Might I recommend we simply comment out the test for now, and add a TODO: Enable back again kind of message?

We can toss in the `@Ignore` annotation for the test, then it'll show up as an ignored test which is always something to avoid. More TODO-like encouragement.

jellysnake:
The reason the tests were not working is that as part of the changes I made it such that the marker nodes are only added on debug.

Thus to fix the tests I removed the lines where it checked for the marker nodes. However I do not know enough about the rendering system to know if this is the best approach.
The other approach to fixing the tests I can see is to revert the check for adding the nodes from `isDebugEnabled()` back to `isInfoEnabled()` which will result in the marker nodes always being added and the tests happy as they can find them.
Alternativly is to run the tests with debug logging enabled but I'm not sure how to do that.

GooeyHub:

Cervator:
Tests out now - merging :-)

PR #3397, opened 2018-06-16
Improve the biomes view

ar0ne:
### Contains

Small improving of rendering biomes on GameDetailsScreen (biomes tab).
See also: https://github.com/Terasology/PolyWorld/pull/16

### How to test

1. Create/Load game with active PolyWorld module.
2. Open GameDetailsScreen and tab biomes.
3. Check that all biomes have the same style (no biomes in UpperCase).


GooeyHub:

Cervator:
Tested this out along with the PolyWorld PR but I'm unsure about the desired end-state: for PolyWorld's Island World now "Ocean" gets listed _twice_ instead of "Ocean" + "OCEAN" and the list isn't sorted so they appear far apart. Should there be some sort of more explicit naming to distinguish the two oceans in the list or should the biomes alpha-sort so it'd be easier to tell there are two biomes with the same short name?

Also: on a note akin to the module detail bit that broke after a version bump I get an NPE when looking at an _existing_ save game with the PolyWorld biomes under their original names. NPE blocks the detail view from working and trying to load the older save causes a game crash. I wonder if there'd be a more elegant way to not break quite so bad in a case like this. May be out of scope (and we're still Alpha so save games are hardly sacred) but this is likely to be a repeat issue in the future as biomes get changed around

ar0ne:
Good catch!
img
Maybe we could sort and just highlight biomes not from Core? I think for most cases it'd be enough.

Cervator:
Yeah that sounds like a good way to do it :-)

ar0ne:
Heya @Cervator ,
Check it please when you have time! 
I added popup message which notify users when something goes wrong during load.

GooeyHub:

ar0ne:
JFYI, I decided to add to this branch more features regarding modules tab on the same screen GameDetails. I will try to ping you again when work is done.

GooeyHub:

GooeyHub:

GooeyHub:

ar0ne:
Heya @Cervator ,
I've updated the view, as you may see I added popup with list of errors. 1 - Old saved game; 2 - manually add imaginary module to manifest.json; 3 - because of bad module.
Can you check it out ?
img

jellysnake:
Quick suggestion. If there is a clash with two short names lets display the full name of the biome.
That is, rather than `Beach` & `Beach`, we should show `Core: Beach` and `OtherModule:Beach`

I've not had a chance to test this out yet but if I get a chance I shall

Cervator:
Nice work on the error handling, works really well :-)

Tested it out in a variety of ways. Merged! Thanks.

I added a card to the backlog to fully qualify biomes if there are duplicates, should be an easy addition.

Also: try to avoid merge commits in a topic branch. Either use git rebase if you need to merge (such as conflicts or related work) or even just leave the topic branch a few unrelated commits out of date

PR #3403, opened 2018-06-19
Gson TypeAdapters and Factories for TypeHandlers

eviltak:
# Motivation
Terasology uses a custom serialization system loosely based on the Gson 1.x architecture. To specify custom serialization for specific types, Terasology requires the use of implementations of `TypeHandler<T>` -- a Terasology-specific interface. As such, serialization handlers for Terasology-introduced types like `Vector3i` and `Color` implement `TypeHandler` and are only compatible with the Terasology serialization system.

To help use existing `TypeHandlers` for serializing Terasology types to JSON via Gson, the `JsonTypeHandlerAdapter` class, a legacy Gson 1.x serialization adapter, was created. `JsonTypeHandlerAdapter` can be used to register Gson type adapters for each type manually; however, this approach is suboptimal for reasons stated below.

Since manually registering a `TypeHandler`/`JsonTypeHandlerAdapter` for every possible generic variant of popularly serialized generic types like `Set`, `Map` and `List` is very cumbersome (if not downright impossible), Terasology uses a `TypeSerializationLibrary` to create/load `TypeHandler`s for types on the fly. This allows `TypeHandler`s for generic types like `Map` to be easily created on demand, rather than having to hardcode handlers for each generic variant.

Gson 2.x included the `TypeAdapterFactory` interface which, much like `TypeSerializationLibrary` creates `TypeAdapter`s for types on demand. However, it is currently not possible to levy the convenience of `TypeAdapterFactory`/`TypeSerializationLibrary` in Gson serialization with existing `TypeHandlers` because `JsonTypeHandlerAdapter`, a legacy Gson 1.x type adapter, is not supported by Gson 2.x `TypeAdapterFactory`.

To work around this, this PR introduces a new `GsonTypeHandlerAdapter` -- a Gson 2.x serialization adapter very similar in functionality to the old `JsonTypeHandlerAdapter` (renamed in this PR to `LegacyGsonTypeHandlerAdapter`) -- which allows the use of an existing `TypeHandler` as a Gson 2.x `TypeAdapter`. This PR also introduces two `TypeAdapterFactory` implementations -- `GsonTypeHandlerAdapterFactory` and `GsonTypeSerializationLibraryAdapterFactory` -- which create `GsonTypeHandlerAdapter` instances for types from their handlers on the fly.

Since `GsonTypeHandlerAdapter` instances are meant to be used only via a `TypeAdapterFactory` implementation, the `JsonTypeHandlerAdapter` is retained and renamed to `LegacyGsonTypeHandlerAdapter` in cases where individual `TypeHandler`s/`TypeAdapter`s are required.

# Impact
With the Terasology serialization system, there is no way to arbitrarily serialize objects of any compound type purely containing serializable data (like `class GameMap { Map<String, Rect2i> regions; }`) -- you must create a `TypeHandler` for the class (`GameMap` in our example) which manually (de)serializes all fields.

With the introduction of `JsonTypeHandlerAdapter`, you could levy the automagic of Gson by registering a `JsonTypeHandlerAdapter` in a `Gson` object for types having a `TypeHandler`. However, this would still be very restrictive as individual type handlers/adapters would not be able to handle all generic variants for a generic type like `Map` or `Set` (the very reason for which `TypeSerializationLibrary` was created).

With the introduction of the new `TypeAdapter` and `TypeAdapterFactory` implementations in this PR, it is now possible to levy the convenience of on-demand `TypeHandler` creation of `TypeSerializationLibrary` by using a `GsonTypeSerializationLibraryAdapterFactory`, which internally uses a `TypeSerializationLibrary` to create `TypeHandler`s for types on the fly (which are then wrapped by a `GsonTypeHandlerAdapter`). This also means that if a `TypeHandler` for the type of a given object is not found, the object will be serialized using the default Gson rules, allowing for more flexibility.

This makes serializing compound types using Terasology's handlers very easy:

```java
class GameMap { 
    Map<String, Rect2i> regions; 
    EntityRef mainEntity;
}

static Gson createGson(TypeSerializationLibrary typeSerializationLibrary) {
    return new GsonBuilder()
        .registerTypeAdapterFactory(
            new GsonTypeSerializationLibraryAdapterFactory(typeSerializationLibrary)
        )
        .create();
}

String serializeGameMap(GameMap gameMap) {
    Gson gson = createGson(entityManager.getTypeSerializerLibrary());    // Includes EntityRef handler
    return gson.toJson(gameMap);
}

GameMap deserializeGameMap(String json) {
    Gson gson = createGson(entityManager.getTypeSerializerLibrary());    // Includes EntityRef handler
    return gson.fromJson(json, GameMap.class);
}
```

# Changes
- Added `GsonTypeHandlerAdapter`
- Added `GsonTypeHandlerAdapterFactory`
- Added `GsonTypeSerializationLibraryAdapterFactory`
- Added `GsonMapExclusionStrategy`
- Added `GsonFactory`
- Added `TypeHandlerEntry`
- Added `ReflectionUtil.readField` helper
- Renamed `JsonTypeHandlerAdapter` to `LegacyGsonTypeHandlerAdapter`

# Caveats
- When `GsonTypeSerializationLibraryAdapterFactory` encounters a type which does not have a `TypeHandler` in `TypeSerializationLibrary`, the `TypeSerializationLibrary.getHandlerFor` logs errors. Logging errors, however, is unnecessary in this case because when a `TypeHandler` is not found, Gson switches to its default serialization handlers and is (more often than not) able to serialize the type correctly. This is specifically noticeable when serializing an object of a compound type with no existing `TypeHandler` (like `GameMap` in the above example). This can be worked around by making `getHandlerFor` log errors only on request.
- `TypeSerializationLibrary.getHandlerFor` returns null when a `Map` type with non-`String` keys is passed to it, raising an error in the stock Terasology serializer and ensuring that only `Map`s with `String` keys are serialized. In Gson, however, when a non-`String` keyed `Map` is serialized, the keys are serialized using `toString()`, effectively breaking deserialization. This can be prevented by either **a)** making Gson exclude non-String key Maps manually (using an `ExclusionStrategy` or similar); or **b)** [Enabling complex `Map` key serialization](https://google.github.io/gson/apidocs/com/google/gson/GsonBuilder.html#enableComplexMapKeySerialization--) in Gson, which serializes the `Map` to a list of key-value pairs (for non-primitive/complex key types), thus allowing deserialization.
- Deserialization of maps with primitive non-String key types does not work -- check [this comment](https://github.com/MovingBlocks/Terasology/pull/3403#issuecomment-401583540) on this PR for more details.

# Considerations
- [x] Add a static helper method (like in the example above) that creates a `Gson` object for a `TypeSerializationLibrary` while keeping in mind the following considerations.
- [x] Either **a)** make Gson exclude non-String key Maps manually (using an `ExclusionStrategy` or similar); ~or **b)** [enable complex `Map` key serialization](https://google.github.io/gson/apidocs/com/google/gson/GsonBuilder.html#enableComplexMapKeySerialization--) in Gson, which serializes the `Map` to a list of key-value pairs (for non-primitive/complex key types), thus allowing deserialization. Refer to the link for more details.
(Caveat 2)~
- [x] ~Extend primitive type handlers in Terasology to account for serialized values enclosed in a string/quotes _if_ option **b)** is chosen above
(Caveat 3)~

Comments and suggestions (especially on the **Caveats** and the **Considerations**) are welcome.

### DIFF
--- a/engine/src/main/java/org/terasology/persistence/typeHandling/gson/GsonFactory.java
+++ b/engine/src/main/java/org/terasology/persistence/typeHandling/gson/GsonFactory.java
@@ -20,7 +20,7 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.TypeAdapterFactory;
 import org.terasology.persistence.typeHandling.TypeSerializationLibrary;
 
-public class GsonUtility {
+public class GsonFactory {
     public static GsonBuilder createDefaultGsonBuilder() {
         return new GsonBuilder()
                 .setExclusionStrategies(new GsonMapExclusionStrategy());
--- a/engine/src/main/java/org/terasology/persistence/typeHandling/gson/GsonFactory.java
+++ b/engine/src/main/java/org/terasology/persistence/typeHandling/gson/GsonFactory.java
@@ -20,12 +20,26 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.TypeAdapterFactory;
 import org.terasology.persistence.typeHandling.TypeSerializationLibrary;
 
+/**
+ * Class containing static factory methods for generating {@link Gson} objects that follow Terasology
+ * serialization rules and support Terasology TypeHandlers.
+ */
 public class GsonFactory {
+    /**
+     * Create a {@link GsonBuilder} with options set to comply with Terasology JSON serialization rules.
+     */
     public static GsonBuilder createDefaultGsonBuilder() {
         return new GsonBuilder()
                 .setExclusionStrategies(new GsonMapExclusionStrategy());
     }
 
+    /**
+     * Create a {@link Gson} object which uses type handlers loaded from the given
+     * {@link TypeSerializationLibrary} and complies with Terasology JSON serialization rules.
+     *
+     * @param typeSerializationLibrary The {@link TypeSerializationLibrary} to load type handler
+     *                                 definitions from
+     */
     public static Gson createGsonWithTypeSerializationLibrary(TypeSerializationLibrary typeSerializationLibrary) {
         TypeAdapterFactory typeAdapterFactory =
                 new GsonTypeSerializationLibraryAdapterFactory(typeSerializationLibrary);
@@ -35,11 +49,17 @@ public class GsonFactory {
                 .create();
     }
 
-    public static Gson createGsonWithTypeHandlers(TypeHandlerEntry<?>... typeHandlerPairs) {
+    /**
+     * Create a {@link Gson} object which uses the given type handlers and complies with Terasology
+     * JSON serialization rules.
+     *
+     * @param typeHandlerEntries The type handlers to use during serialization.
+     */
+    public static Gson createGsonWithTypeHandlers(TypeHandlerEntry<?>... typeHandlerEntries) {
         GsonTypeHandlerAdapterFactory typeAdapterFactory = new GsonTypeHandlerAdapterFactory();
 
-        for (TypeHandlerEntry typeHandlerPair : typeHandlerPairs) {
-            typeAdapterFactory.addTypeHandler(typeHandlerPair.type, typeHandlerPair.typeHandler);
+        for (TypeHandlerEntry typeHandlerEntry : typeHandlerEntries) {
+            typeAdapterFactory.addTypeHandler(typeHandlerEntry.type, typeHandlerEntry.typeHandler);
         }
 
         return createDefaultGsonBuilder()
--- a/engine/src/main/java/org/terasology/persistence/typeHandling/gson/GsonFactory.java
+++ b/engine/src/main/java/org/terasology/persistence/typeHandling/gson/GsonFactory.java
@@ -55,11 +55,12 @@ public class GsonFactory {
      *
      * @param typeHandlerEntries The type handlers to use during serialization.
      */
+    @SuppressWarnings("unchecked")
     public static Gson createGsonWithTypeHandlers(TypeHandlerEntry<?>... typeHandlerEntries) {
         GsonTypeHandlerAdapterFactory typeAdapterFactory = new GsonTypeHandlerAdapterFactory();
 
         for (TypeHandlerEntry typeHandlerEntry : typeHandlerEntries) {
-            typeAdapterFactory.addTypeHandler(typeHandlerEntry.type, typeHandlerEntry.typeHandler);
+            typeAdapterFactory.addTypeHandler(typeHandlerEntry);
         }
 
         return createDefaultGsonBuilder()
--- a/engine-tests/src/test/java/org/terasology/persistence/typeHandling/gson/GsonTypeHandlerAdapterTest.java
+++ b/engine-tests/src/test/java/org/terasology/persistence/typeHandling/gson/GsonTypeHandlerAdapterTest.java
@@ -16,7 +16,6 @@
 package org.terasology.persistence.typeHandling.gson;
 
 import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
 import org.junit.Assert;
@@ -31,15 +30,9 @@ public class GsonTypeHandlerAdapterTest {
     private static final String OBJECT_JSON_HEX = "{\"color\":DEADBEEF,\"i\":-123}";
     private static final TestClass OBJECT = new TestClass(new Color(0xDEADBEEF), -123);
 
-    private final ColorTypeHandler typeHandler = new ColorTypeHandler();
-
-    private final GsonTypeHandlerAdapterFactory typeHandlerAdapterFactory = new GsonTypeHandlerAdapterFactory() {{
-        addTypeHandler(Color.class, typeHandler);
-    }};
-
-    private final Gson gson = new GsonBuilder()
-            .registerTypeAdapterFactory(typeHandlerAdapterFactory)
-            .create();
+    private final Gson gson = GsonUtility.createGsonWithTypeHandlers(
+            TypeHandlerEntry.of(Color.class, new ColorTypeHandler())
+    );
 
     /**
      * {@link GsonTypeHandlerAdapter#read(JsonReader)} is tested by deserializing an object from JSON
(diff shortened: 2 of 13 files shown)


GooeyHub:

mkienenb:
> This can be worked around by making getHandlerFor log errors only on request.

`TypeSerializationLibrary` logged at "_error_" severity when the process failed.   I'd recommend logging  at "_warning_" severity when `TypeAdapterFactory` uses a default serialization handler.

eviltak:
@mkienenb I greatly appreciate your recommendation, but I think it will be better to not log anything in this case. `TypeSerializationLibrary.getHandlerFor` does log errors when it cannot find/create a `TypeHandler` for the given type, which is what I assume you mean by "when the process failed". This is understandable when using the Terasology serialization system since type serialization is impossible in the absence of a `TypeHandler`. However, when the method is used in this context, the absence of a `TypeHandler` for a type does not signify a failure -- Gson can still serialize the type using its own default object serialization. As such, I do not see the need to log warnings when `getHandlerFor` fails to return a `TypeHandler`; there is nothing to warn the user about. If no `TypeHandler` is found and Gson fails to serialize the object on its own, Gson will throw an exception.

Furthermore, if we continue logging warnings/errors in this case, the logs will be spammed with these (unnecessary, according to me) warnings/errors. Errors will be logged for every type that Gson encounters which does not have a `TypeHandler`, and this will be a quite frequent phenomenon. For example, the work in progress record and replay module serializes thousands of events during a record, and each of these events is of a type that does not have a corresponding `TypeHandler` and has to go through the Gson object serialization process. For each event object that is serialized, `getHandlerFor` will log an error for the absence of a `TypeHandler`. This will lead to the log being spammed with thousands of missing `TypeHandler` messages which are irrelevant since the serialization will continue anyway.

mkienenb:
@eviltak It wasn't clear to me whether having the Gson 2.x  default serialization triggered was intended as a band-aid temporary workaround until the mod author got around to fixing it himself, or if Gson 2.x default serialization was intended to be the right way for serializing those objects.    The quoted line (and the fact it was in the caveats section) made me think that the mod author was supposed to fix these.   Your response seems to indicate that is not the case.

eviltak:
@mkienenb so that's where the misunderstanding stemmed from :smile: I listed this point under caveats because I hadn't gotten around to fixing it yet and wanted to discuss it with others. Gson 2.x default serialization is meant to be the right way for serializing these objects, as you correctly noted.

eviltak:
Some testing has revealed that deserialization of a `Map` with a primitive (but non-String) key type like `Integer` fails with the `GsonTypeSerializationLibraryAdapterFactory`. As per valid JSON, Gson serializes primitive keys (like `31`) to a string by including quotes around the serialized primitive value (like `"31"`). When these keys are deserialized, the `GsonTypeSerializationLibraryAdapterFactory` ensures that the Terasology type handler is used for the primitive type and passes the string-serialized value (`"31"`) to the type handler. However, primitive type handlers in Terasology expect to simply get the numeric value (`31`), take one look at the quotes (`"31"`), consider it a string and return a null value. The way I see it, this has two workarounds:

- Make Gson exclude non-String key Maps manually (using an ExclusionStrategy or similar) -- Caveat 2 option a); or
- Extend the primitive type handlers in Terasology to deserialize serialized primitive values enclosed in a string/quotes.

Cervator:
Good find. I don't know if this is much of an answer as much as an obvious preference but the less custom code and the more in _compliance_ we are with the frameworks we use (including upgraded versions) the better, IMHO. And the harder it is to screw up code the better as well :-)

If authors have to remember to add in `enableComplexMapKeySerialization` somewhere or take special steps to deal with values inside maps (or end up with unusually ugly JSON like the second example in the link) then that's bad. If a helper method can hide all that and make it easy (and pretty) then great. Especially if it can be enforced somehow.

We did just do a major release, but are still in alpha, so if we should aim to overhaul this area further (rewriting all type handlers) for improvements that'll be worth it then by all means aim for it. Should `LegacyGsonTypeHandlerAdapter` or anything related maybe be marked as `@deprecated`  to make it clearer what may need effort to be fully up to date?

Trying to poke people occasionally to get more review on this.

eviltak:
I do plan on wrapping all required settings like calling `enableComplexMapSerialization` or adding an `ExclusionStrategy` for non-String keyed maps in a helper method that creates a "Terasology recommended" Gson object for serialization (consideration 1).

Regarding the ugliness of `enableComplexMapSerialization` -- Gson will, by default, serialize maps with complex keys to a list of key value pairs (the ugly second example). However, to retain a map-like structure, we could serialize the keys to JSON, and store them as a string JSON key. This will retain the `key: value` structure of JSON maps and should, in theory, _not_ require us to extend type handlers to deserialize serialized primitive values enclosed in a string/quotes.

Marking `LegacyGsonTypeHandlerAdapter` as deprecated may be a good idea. `GsonTypeHandlerAdapterFactory` fulfills more or less the same functionality as multiple registered`LegcayGsonTypeHandlerAdapter`s (mapping serialization of specified types to their TypeHandlers), and we do not know if Gson will continue to support legacy Gson 1.x type adapters in later releases.

engine/src/main/java/org/terasology/persistence/typeHandling/gson/GsonFactory.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Create a {@link Gson} object which uses the given type handlers and complies with Terasology
+     * JSON serialization rules.
+     *
+     * @param typeHandlerEntries The type handlers to use during serialization.
+     */
+    @SuppressWarnings("unchecked")
+    public static Gson createGsonWithTypeHandlers(TypeHandlerEntry<?>... typeHandlerEntries) {
```
eviltak: `TypeHandlerEntry` is used to ensure that the `TypeHandler` actually is a type handler for the intended type, because this check is not performed manually further down the line. On creation of a `TypeHandlerEntry`, the compiler ensures that the `TypeHandler` passed is generic over the same type as is specified by the `class`.

GooeyHub:

Cervator:
Thanks for all the extra care, attention, and patience on this one :-)

Merged it locally in a mega-workspace and tried a variety of ways to provoke errors but couldn't find any. All appears to be well!

PR #3404, opened 2018-06-21
Fix remotely held items @commands broken by #3379, brought by #3337

dave2s:
### Commands brought with ThirdPersonRemoteClientSystem(#3337) were broken since #3379. This PR fixes them. These commands provide ability to set position and rotation of items held by remote players for local client trough ingame console. For debugging purposes

#3379
### Contains
Edited ingame commands for remote held item translation rotation and mountpoint scale (the 3rd one is probaly useless though, might as well get rid of it). These commands, introduced by #3337 stopped working after #3379 changed the code to suit the refactored logic. 

Changes: shortened, made it work. Redundant checks for _relatesToLocalPlayer(remotePlayer)_ removed as only real remote player entities should have _RemotePersonHeldItemMountPointComponent_ already (this check is performed when adding those). Unnecessary call to _ensureClientSideEntityOnHeldItemMountPoint()_ deleted.

### How to test

Start multiplayer game with 2 players minimum. Aim your screen at the other player/s. Fire up your console ("~") and use one of the commands:
*setRemotePlayersHeldItemMountPointTranslations \<Float x\> \<Float y\> \<Float z\>* for moving remote held items about. (in meters, 1 meter is one block i suppose?)
*setRemotePlayersHeldItemMountPointRotations \<Float x\> \<Float y\> \<Float z\>* for rotating remote held items about. (in degrees)

observe visual changes of items held by other players :)

Probably nothing.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/ThirdPersonRemoteClientSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/ThirdPersonRemoteClientSystem.java
@@ -121,39 +121,30 @@ public class ThirdPersonRemoteClientSystem extends BaseComponentSystem implement
 
     @Command(shortDescription = "Sets the held item mount point translation for remote characters")
     public void setRemotePlayersHeldItemMountPointTranslations(@CommandParam("x") float x, @CommandParam("y") float y, @CommandParam("z") float z) {
-        for (EntityRef remotePlayer : entityManager.getEntitiesWith(CharacterComponent.class, PlayerCharacterComponent.class)) {
-            if (relatesToLocalPlayer(remotePlayer)) {
-                RemotePersonHeldItemMountPointComponent newComponent = remotePlayer.getComponent(RemotePersonHeldItemMountPointComponent.class);
-                if (newComponent != null) {
-                    newComponent.translate = new Vector3f(x, y, z);
-                    ensureClientSideEntityOnHeldItemMountPoint(OnActivatedComponent.newInstance(), remotePlayer, newComponent);
-                }
+        for (EntityRef remotePlayer : entityManager.getEntitiesWith(RemotePersonHeldItemMountPointComponent.class)) {
+            RemotePersonHeldItemMountPointComponent remoteMountPointComponent = remotePlayer.getComponent(RemotePersonHeldItemMountPointComponent.class);
+            if (remoteMountPointComponent != null) {
+                remoteMountPointComponent.translate.set(x, y, z);
             }
         }
     }
 
     @Command(shortDescription = "Sets the held item mount point rotation for remote characters")
     public void setRemotePlayersHeldItemMountPointRotations(@CommandParam("x") float x, @CommandParam("y") float y, @CommandParam("z") float z) {
-        for (EntityRef remotePlayer : entityManager.getEntitiesWith(CharacterComponent.class, PlayerCharacterComponent.class)) {
-            if (relatesToLocalPlayer(remotePlayer)) {
-                RemotePersonHeldItemMountPointComponent newComponent = remotePlayer.getComponent(RemotePersonHeldItemMountPointComponent.class);
-                if (newComponent != null) {
-                    newComponent.rotateDegrees = new Vector3f(x, y, z);
-                    ensureClientSideEntityOnHeldItemMountPoint(OnActivatedComponent.newInstance(), remotePlayer, newComponent);
-                }
+        for (EntityRef remotePlayer : entityManager.getEntitiesWith(RemotePersonHeldItemMountPointComponent.class)) {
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/logic/players/ThirdPersonRemoteClientSystem.java
```diff
@@ -121,39 +121,30 @@ public void ensureHeldItemIsMountedOnLoad(OnChangedComponent event, EntityRef cl
...
-            if (relatesToLocalPlayer(remotePlayer)) {
-                RemotePersonHeldItemMountPointComponent newComponent = remotePlayer.getComponent(RemotePersonHeldItemMountPointComponent.class);
-                if (newComponent != null) {
-                    newComponent.rotateDegrees = new Vector3f(x, y, z);
-                    ensureClientSideEntityOnHeldItemMountPoint(OnActivatedComponent.newInstance(), remotePlayer, newComponent);
-                }
+        for (EntityRef remotePlayer : entityManager.getEntitiesWith(RemotePersonHeldItemMountPointComponent.class)) {
+            RemotePersonHeldItemMountPointComponent remoteMountPointComponent = remotePlayer.getComponent(RemotePersonHeldItemMountPointComponent.class);
+            if (remoteMountPointComponent != null) {
+                remoteMountPointComponent.rotateDegrees.set(x, y, z);
```
dave2s: modifies the existing vector rather than ovewriting it with a new one

dave2s:
Would have added this to the previous PR but its' branch is deleted

GooeyHub:

engine/src/main/java/org/terasology/logic/players/ThirdPersonRemoteClientSystem.java
```diff
@@ -121,39 +121,30 @@ public void ensureHeldItemIsMountedOnLoad(OnChangedComponent event, EntityRef cl
...
-        for (EntityRef remotePlayer : entityManager.getEntitiesWith(CharacterComponent.class, PlayerCharacterComponent.class)) {
-            if (relatesToLocalPlayer(remotePlayer)) {
-                RemotePersonHeldItemMountPointComponent newComponent = remotePlayer.getComponent(RemotePersonHeldItemMountPointComponent.class);
-                if (newComponent != null) {
-                    newComponent.translate = new Vector3f(x, y, z);
-                    ensureClientSideEntityOnHeldItemMountPoint(OnActivatedComponent.newInstance(), remotePlayer, newComponent);
-                }
+        for (EntityRef remotePlayer : entityManager.getEntitiesWith(RemotePersonHeldItemMountPointComponent.class)) {
+            RemotePersonHeldItemMountPointComponent remoteMountPointComponent = remotePlayer.getComponent(RemotePersonHeldItemMountPointComponent.class);
+            if (remoteMountPointComponent != null) {
```
jellysnake: Very minor thing is that you *technically* don't need this check.
The `entityManager.getEntitiesWith` ensures that they will have the component and thus you don't need to check if it is null.

engine/src/main/java/org/terasology/logic/players/ThirdPersonRemoteClientSystem.java
```diff
@@ -121,39 +121,30 @@ public void ensureHeldItemIsMountedOnLoad(OnChangedComponent event, EntityRef cl
...
-            if (relatesToLocalPlayer(remotePlayer)) {
-                RemotePersonHeldItemMountPointComponent newComponent = remotePlayer.getComponent(RemotePersonHeldItemMountPointComponent.class);
-                if (newComponent != null) {
-                    newComponent.scale = scale;
-                    ensureClientSideEntityOnHeldItemMountPoint(OnActivatedComponent.newInstance(), remotePlayer, newComponent);
-                }
+        for (EntityRef remotePlayer : entityManager.getEntitiesWith(RemotePersonHeldItemMountPointComponent.class)) {
+            RemotePersonHeldItemMountPointComponent remoteMountPointComponent = remotePlayer.getComponent(RemotePersonHeldItemMountPointComponent.class);
+            if (remoteMountPointComponent != null) {
+                remoteMountPointComponent.scale = scale;
```
jellysnake: Have you checked that you don't need to call `remotePlayer.saveComponent` on the remote component here?

engine/src/main/java/org/terasology/logic/players/ThirdPersonRemoteClientSystem.java
dave2s: Right I missed this, thanks 👍

engine/src/main/java/org/terasology/logic/players/ThirdPersonRemoteClientSystem.java
dave2s: It is now called in every update. It should probably be in here too but..would be called twice in a split second. What do you think?

GooeyHub:

Cervator:
Tested - works great! Nice work :-)

PR #3405, opened 2018-06-21
New button widget "Open In Browser"

ar0ne:
### Contains

Add new button widget with predefined action opening default web browser.

### How to test

Add the button widget to some existing screen and set up a link.
Check that default web browser open the link.

e.g. `tryFind("openInBrowser", UIButtonWebBrowser.class).ifPresent( b -> b.setLink("https://github.com"));`

- [ ] Check on Windows.
- [ ] Check on Mac
- [ ] Check on *nix

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UIButtonWebBrowser.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UIButtonWebBrowser.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2018 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.rendering.nui.widgets;
+
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.awt.Desktop;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * Button with predefined action - open the link in default web browser
+ */
+public class UIButtonWebBrowser extends UIButton {
+
+    private static final Logger logger = LoggerFactory.getLogger(UIButtonWebBrowser.class);
+
+    private String link = "";
+
+    public UIButtonWebBrowser() {
+        this.subscribe(openInDefaultBrowser);
+    }
+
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Nice work :-)

This widget is somewhat special and opens up a potentially malicious way modules can trick a user into opening a specific URL in a browser. I wonder if there'd be an easy way for us to include a disclaimer popup before the browser open request is honored?

Potentially with the option to ignore further confirmations for a given module or even all uses of the widget, until the config file is edited anyway. There should be similar logic in place for the first time a user opens the host game or join game option to pick a player name.

ar0ne:
Hm, that makes sense. So, I added confirmation popup and I'd like to propose next code snippet.
```
UIButtonWebBrowser buttonWebBrowser = find("yourButtonId", UIButtonWebBrowser.class);
        if (buttonWebBrowser != null) {
            buttonWebBrowser.setLink("https://github.com");
            buttonWebBrowser.setConfirmationHandler(() -> {
                TwoButtonPopup confirmationPopup = getManager().pushScreen(TwoButtonPopup.ASSET_URI, TwoButtonPopup.class);
                confirmationPopup.setMessage(translationSystem.translate("${engine:menu#button-web-browser-confirmation-title}"), translationSystem.translate("${engine:menu#button-web-browser-confirmation-message}") + "\n" + buttonWebBrowser.getLink());
                confirmationPopup.setLeftButton(translationSystem.translate("${engine:menu#dialog-yes}"), buttonWebBrowser::doConfirmation);
                confirmationPopup.setRightButton(translationSystem.translate("${engine:menu#dialog-no}"), () -> {
                });
            });
        }
```
Should get something like this:

GooeyHub:

mkienenb:
I think the problem is that the module -- caller of find("yourButtonId", UIButtonWebBrowser.class); -- cannot be the one to provide confirmation since the module is untrusted.

It's not likely to be an issue when the engine in the caller, but we have to plan for modules as well as engine use.

What Cervator is suggesting is that the confirmation take place inside of the widget code.

Maybe the widget accesses an Environment variable Map<Url, Boolean> isUrlConfirmedMap to determine if this particular URL has already been confirmed.  If not, it puts up a display panel.....

It's going to be a bit tricky no matter how it's done.

jellysnake:
I can second mkienenb and say that I'd like to see all the code regarding confirmation within the widget. I don't think that perhaps an environment variable is needed. Simply encapsulate the `confirmed` boolean to wholly inside the widget.

One other thing is that I'd like to see a popup if the opening failed. You've got a bunch of log outputs but nothing that's displayed to the user.

Cervator:
Tested on Win10 - works great! :-)

Yep the confirmation needs to be done within the widget, a module should only configure the button and supply the link.

A checkbox could be added to the button that goes "[ ] Accept all future links from [root url]?" - save whitelisted URLs to a section in `config.cfg` somewhere probably. But that could be a future enhancement. Main thing needed to merge this is just forcing the confirmation :-)

ar0ne:
Hi All, I move the logic within the widget, please take a look.
Also, attached the gist: https://gist.github.com/ar0ne/4138988b167e0ba661071d2118424335

GooeyHub:

PR #3406, opened 2018-06-22
New widget "Image Slideshow"

ar0ne:
### Contains

This is new ui widget for slideshows images.

### How to test

Currently, we don't have a bunch of associated images. But you could add for example "default" image.

e.g. GameDetailsScreen(line: 383).
```
    protected void setPreviewImage(TextureRegion texture) {
        UIImageSlideshow slider = find("preview", UIImageSlideshow.class);
        if (slider != null) {
            slider.addImage(texture);
            slider.addImage(Assets.getTexture(DEFAULT_PREVIEW_IMAGE_URI).get());
        }
    }
```

Not sure but now it works automatically. Do we need to add buttons "next" and "prev" ? Or better to add new card and put it to backlog ?

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
@@ -15,8 +15,6 @@
  */
 package org.terasology.rendering.nui.widgets;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.terasology.math.geom.Vector2i;
 import org.terasology.rendering.assets.texture.TextureRegion;
 import org.terasology.rendering.nui.Canvas;
@@ -32,31 +30,34 @@ import java.util.Date;
 import java.util.List;
 
 /**
- * A widget to display image slideshows
+ * A widget to display image slideshows.
  */
 public class UIImageSlideshow extends CoreWidget {
 
-    private Logger logger = LoggerFactory.getLogger(UIImageSlideshow.class);
-
     private Binding<TextureRegion> currentImage = new DefaultBinding<>();
-
-    private long timestamp = new Date().getTime();
+    private long timestamp = getCurrentTimestamp();
     private boolean active = true;
+    private List<TextureRegion> images = new ArrayList<>();
 
-    @LayoutConfig
+    /**
+     * An index number of current image.
+     */
     private int index = 0;
 
+    /**
+     * A speed of changing images in seconds.
+     */
     @LayoutConfig
-    private int speed = 5; // seconds
+    private int speed = 5;
 
+    /**
+     * Whether the slideshow infinite.
+     */
     @LayoutConfig
     private boolean infinite = true;
 
     @LayoutConfig
     private boolean ignoreAspectRatio;
-
-
-    private List<TextureRegion> images = new ArrayList<>();
 
     @Override
     public void onDraw(Canvas canvas) {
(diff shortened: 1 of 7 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -0,0 +1,142 @@
...
+                canvas.drawTexture(currentImage.get(), Color.WHITE);
+            }
+        }
+    }
+
+    @Override
+    public void update(float delta) {
+        if (isActive() && timestamp + speed * 1000 < getCurrentTimestamp()) {
+            timestamp = getCurrentTimestamp();
+            nextImage();
```
jellysnake: You can use the `delta` here instead of making calls to Date.
Since `delta` is the time since last call you can do something like this:

```java
public void update(float delta) {
    imageDisplayTime += delta;
    if (imageDisplayTime >= speed) {
        imageDisplayTime = 0f;
        /* Call your image switching here */
    }
}
```

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -0,0 +1,142 @@
...
+    @LayoutConfig
+    private boolean infinite = true;
+
+    @LayoutConfig
+    private boolean ignoreAspectRatio;
+
+    @Override
+    public void onDraw(Canvas canvas) {
+        if (currentImage.get() != null) {
+            canvas.drawTexture(currentImage.get(), Color.WHITE);
```
jellysnake: Should this line be here? Won't it result in the image being drawn twice, once in the `if` below and once here?

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -0,0 +1,142 @@
...
+     */
+    @LayoutConfig
+    private boolean infinite = true;
+
+    @LayoutConfig
+    private boolean ignoreAspectRatio;
+
+    @Override
+    public void onDraw(Canvas canvas) {
+        if (currentImage.get() != null) {
```
jellysnake: I'm making a suggestion that instead of manually drawing the image here, you make this widget use an `UIImage`.

You can simply call it's draw here and use a binding to automatically update the image in it when the image index here is altered.

You can also delegate to its `getPreferredContentSize`.

jellysnake: Alternatively you can consider extending the `UIImage` which will also allow you to do similar and utilise it's default methods.

GooeyHub:

Cervator:
Tests out, nice work :-) I concur with Jelly's comments, happy to merge this after those are addressed.

On Slack we discussed the possibility of a timed cycle variant vs manual back/forward. That could be done in a separate PR or along with this one, up to you.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
ar0ne: Nope. thanks!

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
ar0ne: Great idea, thanks!

GooeyHub:

GooeyHub:

Issue #3407, opened 2018-06-24
Better Java version handling especially for Java 10


Cervator:
Report in from Twitter, see https://twitter.com/Samuel3744/status/1010657742833537024

In short: A user has only Java 10 installed. Prompt is appearing that indicates no suitable *Java 8* installation is available. 

- [ ] Will the game actually run with Java 10? I haven't tried, am unsure if anybody did. It wasn't even that long ago we looked at Java 9 a bit
- [ ] Speaking of Java 9: Did we actually make sure everything was well there now that it is stable and released? 
- [ ] Is our JRE detection going to cause trouble if it is insisting on Java 8 somewhere even if 9 (or 10) work? I know in Gradle we just require a Java 8 *compatible* JDK so later ones are fine too.

Somewhat unsure what is triggering the Java 8 prompt. I'm wondering if it might actually be the TerasologyLauncher, not the game itself. In that case we might need a fix over there as well.

In either case Java 10 is coming up occasionally so it would be good if we could do some research to prepare (and also confirm Java 9 as fine)

mkienenb:
Don't bother?

http://www.oracle.com/technetwork/java/javase/eol-135779.html

Java 9 was EOL as of March (2018)
Java 10 will be EOL as of September (2018)

Probably better to skip straight to Java 11.

Cervator:
Slight clarification then: I'm looking for *compatibility* with Java 9 and 10. Not at all saying we should start requiring either or begin using their fancy features. We're still aiming to build things with Java 8 and so on :-)

mkienenb:
I completely get what you are saying, but targetting Java 11 is a better approach.   Java 9 has already been replaced by anyone who is keeping up-to-date.   java 10 will be gone by September.

Cervator:
Eh? I mean the idea is to support Java 8, 9, and 10, all at the same time. 11 too when it comes out. We have users right now needing that. Including me on my Mac for that sake, I can't dev without PR #3409 :-)

Oooor I could try to downgrade Java. On a Mac. On my second week ever using one. Eehhh. And I'm a dev. Imagine the amount of regular users that are *not* maintaining their Java and have no idea which they're on - potentially for years to come. It is like the routine issues with old integrated graphic chips, just with Java.

Cervator:
[...]

PR #3409, opened 2018-06-27
Upgrade to Gradle 4.7 for Java 10 compatibility (version parsing issue)

Cervator:
Relates to #3407 - at least on MacOS you can run the game seemingly OK with Java 10, but without this PR you couldn't *build* the game on Java 10 due to a trivial version parsing change in how Gradle detects the Java version.

Still would be good with more research to see if there are any other obvious issues with 9 or 10, this just allows you to use Gradle again.

I recall having noted several potential issues in upgrading Gradle beyond 3.3 or so. So it might not be ideal to merge this yet - especially as I believe one relates to how Gradle outputs when running in CI (the logging changes and can end up deducing the wrong mode to run in)

Anyway, anybody else able to test out this simple change on Win + Linux? Try out a few Gradle commands (like `gradlew idea`, check, jar, clean, etc) and try to re-open the game in IntelliJ after regenerating project files.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Good news everyone! Apparently the logging issue got fixed in a later Gradle 4 than the one I tried last time. No trace of it and if there are no others we could probably go ahead and upgrade. Lots of other good improvements.

Gradle 5 on the other hand will *definitely* cause some stuff to break, but hopefully not too much

PR #3416, opened 2018-07-02
Record & Replay - Fix issue of holding left click

iaronaraujo:
### Contains

Fixes the issue in which holding left click during a recording was causing replays to not reproduce block destruction correctly.

### How to test

Record a game in which you destroy blocks by holding left click and then watch its replay to check that everything was destroyed correctly.

- [ ] This PR fixes the "left click" problem, but the same issue exists when holding "right click".


GooeyHub:

GooeyHub:

Cervator:
Tested it out. Pretty subtle, but I could provoke a difference before vs after :-)

Merged, thanks!

PR #3419, opened 2018-07-03
Extract Inventory from Core to separate module

jellysnake:
This PR moves all the code that directly relates to providing an Inventory, from Core into it's own module.

This new module can be found under https://github.com/jellysnake/Inventory, however will be moved into the official lineup when/if this PR is merged.

The only potential issues that will come from this is if a module references an asset that is longer located in core. That will most likely require someone with a mega-workspace to check and locate any such issues.

# Testing
Confirm that other modules that include code related to the inventory still compile correctly.
Confirm that playing the game normally still works.


GooeyHub:

jellysnake:
Note that the build error is due to the module, Inventory, not being located in the official lineup. Hence the build fails when attempting to locate that module

Cervator:
Compiles in a mega-workspace! Not bad.

> The only potential issues that will come from this is if a module references an asset that is longer located in core. That will most likely require someone with a mega-workspace to check and locate any such issues.

Oh right. There are **6544** such occurrences. I'm not kidding, that's accurate :-)

Admittedly a fair amount of that bloat looks like StructureTemplates where every single block from Core ends up counting as one hit. For instance MedievalCities is responsible for more than half the hits, at a total of 3738 over just 14 ST `.prefab` files. So that'd clean up pretty quick.

400ish might be temp files, in a clean new workspace without saves that would go away. GooeysQuests has another 518 again in STs. Lost has 521

Daunting, but might not be as bad as the first glance hints at. Want to go ahead and move the module under the Terasology org so we can build it formally? That might make it a little easier to rename and test things.

And just to clarify this is essentially a clean move of everything inventory without changes? Any plans to refactor anything substantial like with your prior inventory efforts?

General note for anybody testing: Requires a reset of `config.cfg` since otherwise you'll end up with a bunch of duplicate key bindings (toolbar no worky etc)

After that reset I can play the game although oddly I'm left-clicking right through a chest's inventory screen, causing an attack to whatever is behind the UI. Comes with a NPE:

```
18:47:46.589 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
	at org.terasology.logic.inventory.InventoryUtils.putItemIntoSlot(InventoryUtils.java:392)
	at org.terasology.logic.inventory.InventoryUtils.moveItem(InventoryUtils.java:331)
```

jellysnake:
Wow, that's.. uh. a lot of usages. I'm surprised that there are that many, when I've only moved across 8 assets, of which 5 are UI screens and 3 are UI skins. I may have to grab my own mega-workspace and check for those assets specifically.


Yes this is a clean move. There are no changes, but this move will make it easier to refactor. I may end up tweaking the inventory a bit in later stages but for now my efforts are focused in other places. (GSoC)

I'll move the module across to the offical org. Should I also setup a jenkins build or leave that til later?

Cervator:
Feel free to set up Jenkins (clone an existing module job, make sure it is using the two template steps) - also try to get the Jenkins hook configured here on GitHub (Settings -> Integrations & services -> service "Jenkins (GitHub plugin)" -> Hook URL `http://jenkins.terasology.org/github-webhook/` make sure to include the `/`) 👍 

Probably I screwed up the usage search. I somehow got into thinking we were talking about *any* asset in Core, like blocks used in StructureTemplates. So I searched for stuff like `Core:` which included lots of blocks. Saw the chest first thing in the file list and my mind went off on a tangent. I should re-search *just* for the set of assets after digging them out of the list. Sorry about that :-)

On the plus side we now know how much it would take to move the entirety of Core out somewhere 😁

GooeyHub:

jellysnake:
All conflicts are caused by both this PR and Replay&Record adding new things to a constructor. It is easily fixed by manually merging the two small changes

jellysnake:
Conflicts can be ignored, as files are being removed and changes have been replicated in Terasology/Inventory

skaldarnar:
List of files/assets removed from Core: 

```
core:dropItemRegionDefault
core:inventoryDefault
core:itemTooltip
core:dropItemRegion
core:inventoryHud
core:transferItemCursor
core:containerScreen
core:inventoryScreen
```

skaldarnar:
> After that reset I can play the game although oddly I'm left-clicking right through a chest's inventory screen, causing an attack to whatever is behind the UI. Comes with a NPE:

@jellysnake could you please look into this. I'm experiencing the same oddity. I tracked it down to `InventoryCell.getTransferEntity()` return the Null entity (see https://github.com/MovingBlocks/Terasology/blob/develop/modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryCell.java#L202-L204)

GooeyHub:

PR #3420, opened 2018-07-03
removed redudant vector handler in Util

pollend:
This function is already implemented in Vector3f. This implementation is redundant in the codebase.


GooeyHub:

jellysnake:
It appears that a rouge file, `target/.history`, has snuck in. Should probably remove that.

Although this only seems to be used in one place inside of the engine, can you confirm it's not used anywhere else?
I'd say it's probably better to mark the method as Deprecated and then to remove it in the next breaking release. As it stands it's a relatively minor issue, probably would be better if we went though and removed all redundant methods/cleaned up the whole Vector utils file.

pollend:
I'll do that.

PR #3422, opened 2018-07-04
New constructor in `OnBlockItemPlaced` to have a third argument for event instigator

l0ftyWhizZ:
Added a new constructor in `OnBlockItemPlaced` to have a third argument for event instigator, and corresponding changes made in `BlockItemSystem`.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/items/OnBlockItemPlaced.java
+++ b/engine/src/main/java/org/terasology/world/block/items/OnBlockItemPlaced.java
@@ -20,12 +20,25 @@ import org.terasology.entitySystem.event.Event;
 import org.terasology.math.geom.Vector3i;
 
 /**
+ * This event gets called whenever a block item is placed in the world
  */
 public class OnBlockItemPlaced implements Event {
+    /**
+     * The position where the block is placed
+     */
     private Vector3i position;
-    private EntityRef placedBlock;
-    private EntityRef instigator;
 
+    /**
+     *  The entity corresponding to the placed block
+     */
+    private EntityRef placedBlock;
+
+    /**
+     * The entity which placed the block
+     */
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

GooeyHub:

engine/src/main/java/org/terasology/world/block/items/OnBlockItemPlaced.java
```diff
@@ -24,17 +24,27 @@
 public class OnBlockItemPlaced implements Event {
     private Vector3i position;
     private EntityRef placedBlock;
+    private EntityRef instigator;
 
     public OnBlockItemPlaced(Vector3i pos, EntityRef placedBlock) {
```
jellysnake: Can you mark this as @Deprecated.

engine/src/main/java/org/terasology/world/block/items/OnBlockItemPlaced.java
```diff
@@ -24,17 +24,27 @@
 public class OnBlockItemPlaced implements Event {
     private Vector3i position;
     private EntityRef placedBlock;
+    private EntityRef instigator;
```
jellysnake: Ideally this would be set to `EntityRef.NULL` rather than `null`
That can either be done here or by calling the three arg constructor from the two arg constrctor

GooeyHub:

GooeyHub:

Cervator:
Tested out along with latest GooKeeper, thanks! :-)

One tip for the future: When deprecating a constructor like this simply make it call the new one with `null` for the new field, rather than have it keep its old logic. That way if the new constructor does something different (now or in the future) the old constructor has a chance to benefit from it too.

No big deal in this case though so merging :-)

PR #3424, opened 2018-07-05
Up-Down-Enter keys support for UIList widget

ar0ne:
### Contains

IMHO, it might be useful. At present, you can't switch list items by up and down keys, you can use mouse only. 

### How to test

1. Open any screen with UIList widget (e.g. SelectGameScreen). 
2. Select any item from list (focus should be on UIList), and check that **Up**, **Down** and **Enter** keys work properly.

Nothing.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UIList.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UIList.java
@@ -16,6 +16,7 @@
 package org.terasology.rendering.nui.widgets;
 
 import com.google.common.collect.Lists;
+import org.terasology.input.Keyboard;
 import org.terasology.input.MouseInput;
 import org.terasology.math.Border;
 import org.terasology.math.geom.Rect2i;
@@ -25,8 +26,10 @@ import org.terasology.rendering.nui.Canvas;
 import org.terasology.rendering.nui.CoreWidget;
 import org.terasology.rendering.nui.databinding.Binding;
 import org.terasology.rendering.nui.databinding.DefaultBinding;
+import org.terasology.rendering.nui.events.NUIKeyEvent;
 import org.terasology.rendering.nui.events.NUIMouseClickEvent;
 import org.terasology.rendering.nui.events.NUIMouseDoubleClickEvent;
+import org.terasology.rendering.nui.events.NUIMouseOverEvent;
 import org.terasology.rendering.nui.itemRendering.ItemRenderer;
 import org.terasology.rendering.nui.itemRendering.ToStringTextRenderer;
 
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Weird, the first time I tried this out I noticed some bizarre behavior on the advanced screen in the module list. I could activate a module fine, but then it wouldn't highlight and I couldn't disable it again. I could also sometimes double-mouse-click grayed out modules (online only) and have their button caption change (Activate/Deactivate).

Thought maybe it was a funny interaction with #3429 which I had in my workspace at the same time, but I rolled back this PR and it still happened. But then I cleaned my workspace really well and it went away - including *after* merging this PR again. So maybe that's a non-issue. Wanted to mention it just in case it might have been or get spotted by anybody else.

For a couple simple improvements: Could the spacebar also work like enter? That seems fairly typical. And you can currently use the arrow keys to get outside the visible area since it doesn't trigger scrolling. Not entirely sure how easy that would be to improve - interact with the scrolling code and try to advance the scroll up/down when a selection gets out of the visible range?

Another maybe interesting interaction: using the mousewheel to scroll while above the scrollable area takes the focus away so the arrow keys no longer work until the next time you select something in the list. Unsure if that'd be an easy fix.

This also really makes me miss having tab ordering, which we still need to implement (there's an ancient issue floating around on that).

GooeyHub:

GooeyHub:

ar0ne:
Hey @Cervator  ,
I've updated PR and provided more features. Can you check it again? 
Honestly, I can't reproduce your issue. Tell me please if you still have it. 
Also, at present, I have no idea how can we fix the scrolling code and always show visible items, but I've added card to my backlog, maybe I'll get inspired in future.

Cervator:
Looking good now, cannot reproduce :-)

One minor fix needed now though - one tweak went too far

> Another maybe interesting interaction: using the mousewheel to scroll while above the scrollable area takes the focus away so the arrow keys no longer work until the next time you select something in the list. Unsure if that'd be an easy fix.

Now the behavior is that mousewheel changes the _selected item_ rather than scroll the list at all. That's more problematic since now you can't scroll without hovering immediately on top of the scrollbar. That part could just be reverted, really. The minor advantage of being able to still change item selection via arrows despite using the mousewheel isn't really worth the loss of scroll being the primary action for the mousewheel :-)

Ideal behavior: mouse wheel scrolls the viewport of the _list_ but not the selected item within the list. Doing so doesn't cause arrow scrolling to need a re-selection, using arrow keys still changes the selected item.

If that isn't easy to achieve then just don't worry about it and leave mousewheel for scrolling :-)

One potential additional tweak: Also support up/down _movement_ keys to scroll? I dunno if that's worth it. Especially since the user can rebind those away from `W` / `S`

GooeyHub:

Cervator:
Retested and merged! Thank you :-)

Issue #3384, opened 2018-06-08
Highlight possibility active buttons of UITabBox widget


ar0ne:
### What you were trying to do
I've used UITabBox widget on my new screen GameDetailsScreen (#3382). And realized that we need to redo this widget, because we can't highlight active buttons. 

Actually, I've spent few days on that, and, unfortunately, looks like I don't have nice solution yet...
I'd like just to track this issue further, add to my GSoC board and fix it when I will have good solution/idea.

### How to reproduce
1. Fetch branch from PR(#3382) if it's not in develop yet.
2. Open Details for saved game.
3. Play with tabs and see that they're not highlighted.

Cervator:
Closing as completed by #3429 - thanks @ar0ne ! :-)

PR #3429, opened 2018-07-06
Highlight active tab buttons

ar0ne:
### Contains

Currently, we could highlight button only when we are pressing mouse down. It's fine, but we also need to have a way how to highlight some "active" buttons, e.g. active tab in UITabBox widget.

Original issue #3384 .

### How to test

1. Open GameDetailsScreen. 
2. Check that tabs and tab buttons work correctly. As well as only one active button was highlighted.

Nothing.


GooeyHub:

PR #3290, opened 2018-03-16
CollisionShape API formalization

eviltak:
#3199 introduced a wrapper over collision shapes to remove engine and module dependencies on the physics engine. However, the introduced wrappers are merely functional and have inconsistent and incomplete public APIs. This PR formalizes the `CollisionShape` API and establish a consistent set of methods in the relevant interfaces.

# Changes
## Additions
- [x] `SphereShape.getRadius`
- [x] `BoxShape.getExtents`
- [x] `ConvexHullShape.getVertices`
- [ ] Child shape getters and removal methods to `CompoundShape`
- [ ] `CompoundShapeChild` interface 
- [ ] `CapsuleShape` interface
- [ ] `CylinderShape` interface

## Modifications
- [ ] Make `BulletCompoundShapeChild` public and an implementor of `CompoundShapeChild`
- [ ] Rename `CollisionShapeFactory` methods to use `create` prefix instead of `getNew`

## Removals
- [x] Remove `BoxShape.getHalfExtentsWithoutMargin` in favour of `BoxShape.getExtents`

PR #3432, opened 2018-07-08
Terra bullet (gdx native)

pollend:
This implementation uses a version of libgdx bullet extension to work directly with Termath along with a voxel implementation. there are some out standing features that need to be address and fixed before  this would be considered workable but this implementation works against the current development branch of Terasology.

https://github.com/MovingBlocks/Terasology/pull/3121
https://github.com/MovingBlocks/Terasology/pull/3042

https://github.com/pollend/TeraBullet-2

# How to Test

~~You will have to include the modified libgdx library into lib and install the maven libraries from there (mvn install). At the moment the way to load the native library is to just copy the native-build into the native folder in the root of the project. running the ant script in TeraMath should do everything for you in terms of building the library.~~

The bullet builds are in artifactory so there is no need to copy anything into lib.


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/logic/inventory/ItemPickupAuthoritySystem.java
```diff
@@ -99,10 +99,8 @@ public void updateExtentsOnBlockItemBoxShape(OnAddedComponent event, EntityRef i
             return;
         }
 
-        if (blockFamily.getArchetypeBlock().getCollisionShape() instanceof BoxShape) {
-            BoxShape collisionShape = (BoxShape) blockFamily.getArchetypeBlock().getCollisionShape();
-            Vector3f extents = collisionShape.getHalfExtentsWithoutMargin();
-            extents.scale(2.0f);
+        if (blockFamily.getArchetypeBlock().getCollisionShape() instanceof btBoxShape) {
+            Vector3f extents = ((btBoxShape) blockFamily.getArchetypeBlock().getCollisionShape()).getHalfExtentsWithoutMargin();
```
eviltak: After #3199, this will not work. Please revert this change.

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletBoxShape.java
```diff
@@ -15,36 +15,28 @@
...
-        return new BulletBoxShape(halfExtentsWithMargin);
+        Vector3f halfExtentsWithMargin = boxShape.getHalfExtentsWithMargin();
+        return new BulletBoxShape(rot.rotate(halfExtentsWithMargin));
     }
 
     @Override
     public Vector3f getHalfExtentsWithoutMargin() {
-        javax.vecmath.Vector3f out = new javax.vecmath.Vector3f();
-        return VecMath.from(boxShape.getHalfExtentsWithoutMargin(out));
+        return boxShape.getHalfExtentsWithoutMargin().mul(2);
```
eviltak: I recommend that you leave out the `.mul(2)` for now since the method has not been renamed to `getExtents` (though you have already merged the commit that renames the method to `getExtents` from #3290 -- not sure what has happened here)

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletCapsuleShape.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.physics.bullet.shapes;
+
+import com.badlogic.gdx.physics.bullet.collision.btCapsuleShape;
+import org.terasology.math.geom.Quat4f;
+import org.terasology.physics.shapes.CollisionShape;
+
+public class BulletCapsuleShape extends BulletCollisionShape {
```
eviltak: Ideally this should also implement an engine `CapsuleShape` interface, but it looks like I may have forgotten to add the interface to the CollisionShape API. I will add it in #3290, after which it can be implemented here.

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletCylinderShape.java
```diff
@@ -0,0 +1,39 @@
...
+package org.terasology.physics.bullet.shapes;
+
+import com.badlogic.gdx.physics.bullet.collision.btConvexHullShape;
+import com.badlogic.gdx.physics.bullet.collision.btCylinderShape;
+import org.terasology.math.geom.Quat4f;
+import org.terasology.math.geom.Vector3f;
+import org.terasology.physics.shapes.CollisionShape;
+
+
+public class BulletCylinderShape extends  BulletCollisionShape{
```
eviltak: Like `CapsuleShape`, looks like the `CylinderShape` interface is missing too. I shall add it in #3290.

engine/src/main/java/org/terasology/physics/shapes/CollisionShapeFactory.java
```diff
@@ -62,4 +64,13 @@ default BoxShape getNewUnitCube() {
...
     SphereShape getNewSphere(float radius);
+
+
+    /**
+     * Returns the colliding body associated with the entity
+     *
+     * @param entityRef
+     * @return
+     */
+    BulletCollisionShape getShapeFor(EntityRef entityRef);
```
eviltak: Since a `Factory` is responsible for creating objects, I would prefer moving this method somewhere else. Besides, this method does not need to be overridable -- similar to your implementation in `BulletCollisionShapeFactory`, the final implementation will look for components on the entity and return shapes created from a `CollisionShapeFactory` instance. Also, in the interest of remaining physics engine agnostic, the return type should be made `CollisionShape`. Sample final implementation:

```java
CollisionShape getShapeFor(EntityRef entity, CollisionShapeFactory collisionShapeFactory) {
        BoxShapeComponent box = entityRef.getComponent(BoxShapeComponent.class);
        if (box != null) {
            Vector3f halfExtents = new Vector3f(box.extents);
            return collisionShapeFactory.getNewBox(halfExtents.mul(.5f));
        }
        SphereShapeComponent sphere = entityRef.getComponent(SphereShapeComponent.class);
        if (sphere != null) {
            return collisionShapeFactory.getNewSphere(sphere.radius);
        }
        CapsuleShapeComponent capsule = entityRef.getComponent(CapsuleShapeComponent.class);
        if (capsule != null) {
            // This method does not yet exist in CollisionShapeFactory
            return collisionShapeFactory.getNewCapsule(capsule.radius,capsule.height);
        }
        CylinderShapeComponent cylinder = entityRef.getComponent(CylinderShapeComponent.class);
        if (cylinder != null) {
            // This method does not yet exist in CollisionShapeFactory
            return collisionShapeFactory.getNewCylinder(new Vector3f(cylinder.radius, 0.5f * cylinder.height, cylinder.radius));
        }
        HullShapeComponent hull = entityRef.getComponent(HullShapeComponent.class);
        if (hull != null) {
            FloatBuffer buffer = BufferUtils.createFloatBuffer( hull.sourceMesh.getVertices().size());
            TFloatIterator iterator = hull.sourceMesh.getVertices().iterator();
            int numPoints = 0;
            while (iterator.hasNext()) {
                numPoints++;
                buffer.put(iterator.next());
                buffer.put(iterator.next());
                buffer.put(iterator.next());
            }
            return collisionShapeFactory.getNewConvexHull(buffer,numPoints,3 * Float.BYTES);
        }
        CharacterMovementComponent characterMovementComponent = entityRef.getComponent(CharacterMovementComponent.class);
        if (characterMovementComponent != null) {
            return collisionShapeFactory.getNewCapsule(characterMovementComponent.radius,characterMovementComponent.height);
        }
        throw new IllegalArgumentException("Creating physics object that requires a ShapeComponent or CharacterMovementComponent, but has neither. Entity: " + entityRef);
    }
```

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletCollisionShapeFactory.java
```diff
@@ -44,4 +58,44 @@ public CompoundShape getNewCompoundShape() {
     public SphereShape getNewSphere(float radius) {
         return new BulletSphereShape(radius);
     }
+
+    @Override
+    public BulletCollisionShape getShapeFor(EntityRef entityRef) {
```
eviltak: Like I explain below, I think this method should be moved elsewhere.

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletCollisionShapeFactory.java
```diff
@@ -44,4 +58,44 @@ public CompoundShape getNewCompoundShape() {
...
+                buffer.put(iterator.next());
+            }
+            return new BulletConvexHullShape(buffer,numPoints,3 * Float.BYTES);
+        }
+        CharacterMovementComponent characterMovementComponent = entityRef.getComponent(CharacterMovementComponent.class);
+        if (characterMovementComponent != null) {
+            return new BulletCapsuleShape(characterMovementComponent.radius,characterMovementComponent.height);
+        }
+        logger.error("Creating physics object that requires a ShapeComponent or CharacterMovementComponent, but has neither. Entity: {}", entityRef);
+        throw new IllegalArgumentException("Creating physics object that requires a ShapeComponent or CharacterMovementComponent, but has neither. Entity: " + entityRef);
```
eviltak: If we are already throwing an exception, a logger error may be redundant.

engine/src/main/java/org/terasology/engine/subsystem/lwjgl/BaseLwjglSubsystem.java
```diff
@@ -38,6 +39,7 @@ public void preInitialise(Context context) {
         if (!initialised) {
             initLogger();
             LWJGLHelper.initNativeLibs();
+            Bullet.init(true);
```
eviltak: The introduction of this bullet-specific line of code in the engine directly breaks the independence of the engine from the physics engine. If calling this method is absolutely necessary, we may have to look at an abstraction for physics engine initialization (like we did with the CollisionShape API). This could probably be done in `PhysicsEngineManager`, as that is what currently sets up bullet as the game physics engine.

engine/src/main/java/org/terasology/physics/engine/RigidBody.java
```diff
@@ -120,6 +121,9 @@
      */
     void setOrientation(Quat4f orientation);
 
+    Matrix4f getWorldTransform();
+    Matrix4f setWorldTransform(Matrix4f trans);
```
eviltak: Will it be better to use the `Transform` class here (or modify `Transform` to be created from/use `Matrix4f`) instead?

GooeyHub:

GooeyHub:

pollend:
the steps in the original PR are obsolete. the artifacts are now in artifactory so the basic steps just involve running the project.

GooeyHub:

GooeyHub:

PR #3434, opened 2018-07-08
Fix the issue with ALL UISliders

TheFlash98:
Fixes the issue due to the new listeners on UISliders. :)


GooeyHub:

PR #3438, opened 2018-07-16
Replay Test: Test support for Record & Replay

iaronaraujo:
### Contains

Testing support to Record & Replay through the creation of a new type of test: Replay Tests. It is now possible to write asserts to check game states while a replay is playing. There is also an example test and a template to help in the creation of Replay Tests. For more information, check https://github.com/MovingBlocks/Terasology/wiki/Replay-Tests

### How to test

Run one of the tests in ExampleReplayTest or follow the tutorial in the new wiki page about Replay Tests to create your own replay test.

- [ ] Since replay tests are "headed" they cannot be run by Jenkins, so it is important that the annotation "@Ignore" is present in every replay test.
- [ ] It is known that selecting more than one replay test to run causes issues in the tests that come after the first one, which causes the second test to not run.


GooeyHub:

iaronaraujo:
We talked over in the slack and decided that Replay Tests should be moved to a new module: ReplayTestModule.

Therefore, the things that were moved to the new module was deleted here. There are still some small refactors and an addition of an important method in EventSystemReplayImpl!

Cervator:
Would you mind if we close this and restart the PR with just the few final changes? There's also a (minor) conflict now that I merged the refactor PR :-)

Rebase would be an alternative (and you can do that into another branch to squash the reverted commits) but since the nature of the PR changed it seems more proper to just do a new PR and write it up fresh.

The wiki page might also be a bit out of date now vs the module approach?

iaronaraujo:
Closed this PR and opened this one: https://github.com/MovingBlocks/Terasology/pull/3445

PR #3443, opened 2018-07-20
Switched "getDelta()" to "getDeltaMs()" in the ServerCharacterPredictionSystem.

iaronaraujo:
### Contains

Changes "getDelta()" to "getDeltaMs()" in the ServerCharacterPredictionSystem since it is more appropriate and this could be causing weird behavior.

It is not known if this will really help anything, but after discussing about this in the #architecture channel in the Slack, we arrived at the conclusion that the `getDelta()` method was not appropriate for what it was being used.


GooeyHub:

PR #3448, opened 2018-07-24
Tweak advanced game setup screen + small improvements

ar0ne:
### Contains

Small refactoring and I've added possibility to open ModuleDetails screen from AdvancedGameSetup screen.

### How to test

1. Open AdvancedGameSetup screen
2. Try to open ModuleDetails screen

Nothing.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
@@ -52,11 +52,12 @@ import org.terasology.rendering.nui.databinding.Binding;
 import org.terasology.rendering.nui.databinding.ReadOnlyBinding;
 import org.terasology.rendering.nui.itemRendering.AbstractItemRenderer;
 import org.terasology.rendering.nui.layers.mainMenu.ConfirmPopup;
-import org.terasology.rendering.nui.layers.mainMenu.MessagePopup;
 import org.terasology.rendering.nui.layers.mainMenu.GameManifestProvider;
+import org.terasology.rendering.nui.layers.mainMenu.MessagePopup;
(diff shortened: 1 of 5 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
```diff
@@ -466,6 +468,27 @@ public String get() {
                         });
                     }
                 }
+
+                final UIButton moduleDetails = find("moduleDetails", UIButton.class);
```
Naman-sopho: You should replace `moduleDetails` and other such strings with constants declared at the top of the class.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/moduleDetailsScreen/ModuleDetailsScreen.java
```diff
@@ -141,9 +143,7 @@ public void initialise() {
 
         bindDependencyDescription();
 
-        WidgetUtil.trySubscribe(this, "close", button -> {
-            triggerBackAnimation();
-        });
+        WidgetUtil.trySubscribe(this, "close", button -> triggerBackAnimation());
```
Naman-sopho: Seems like `close` should also be a constant.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
ar0ne: Great idea, I plan to do refactoring of new screens and overall UI next week or two. 
I don't want to do it now, because there are few PRs which will might have conflicts. Better to finish with logic, and only then polish this class.

Cervator:
Thanks for the PR @ar0ne and thanks for review .+ testing @Naman-sopho !

PR #3450, opened 2018-07-24
Add categories to module details screen

ar0ne:
### Contains

Just add field **"Categories"** to module details screen.
pic

### How to test

In the attached screenshot you may find _changed_ categories. 
I've added "isWorld" just for testing and you can do the same.
Also I've fixed "serverSideOnly" for PolyWorld ([PR](https://github.com/Terasology/PolyWorld/pull/18)).

Nothing.


jellysnake:
Tested this, and was unable to break it.
Will merge

PR #3452, opened 2018-07-25
[Defect] Not always all modules are available for filtering

ar0ne:
### Contains

I've spotted out that sometimes I can't find all existing modules on AdvancedGameSetup screen. After investigation, I understood that in this case I can see only "remote" and "filtered" modules from previous session. 

### How to test

1. Open AdvancedGameSetup screen.
2. Filter "libraries"
3. Exit game. 
4. Open AdvancedGameSetup screen.
5. Uncheck libraries.
6. Check that available only "libraries", even when you selected "local".

Nothing.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
@@ -640,16 +640,18 @@ public class AdvancedGameSetupScreen extends CoreScreenLayer {
                 if (info == null) {
                     info = ModuleSelectionInfo.remote(remote);
                     modulesLookup.put(remote.getId(), info);
-                    int pos = Collections.binarySearch(sortedModules, info, moduleInfoComparator);
-                    if (pos < 0) { // not yet in the (sorted) list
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
```diff
@@ -640,16 +640,18 @@ private void updateModuleInformation() {
...
-                        sortedModules.add(-pos - 1, info); // use "insertion point" to keep the list sorted
-                        allSortedModules.clear();
-                        allSortedModules.addAll(sortedModules);
-                    }
                 }
                 info.setOnlineVersion(remote);
             }
         }
+        sortedModules.clear();
+        allSortedModules.clear();
```
Naman-sopho: You might want to add a comment explaining the difference between `sortedModules` and `allSortedModules` variables or change their names to something better.

ar0ne: Hm, these names were used by somebody long time ago and from my perspective they're quite descriptive. 
sortedModules - represent modules for moduleList (UIList widget).
allSortedModules - it's some kind of back-up, contains all sorted modules for cases when we for example reset filter.
I don't think that I can improve it. Maybe you have some ideas?

PR #3454, opened 2018-07-26
Improvements for UIImageSlideshow widget

ar0ne:
### Contains

I redid slideshow widget, now images can be switched manually or by time period, or even both methods.

### How to test

_GameDetailsScreen.java_
```
    public void setPreviews(final List<UIImage> textures) {
        if (textures != null && !textures.isEmpty()) {
            previewSlideshow.clean();
            textures.forEach(previewSlideshow::addImage);

            // ADD RANDOM IMAGES
            previewSlideshow.addImage(Assets.getTexture("engine:defaultPreview").get());
            previewSlideshow.addImage(Assets.getTexture("engine:default").get());
            previewSlideshow.addImage(Assets.getTexture("engine:char").get());
            previewSlideshow.addImage(Assets.getTexture("engine:openbook").get());
        }
    }
```
PIC

Nothing.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/gameDetailsScreen/GameDetailsScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/gameDetailsScreen/GameDetailsScreen.java
@@ -35,7 +35,6 @@ import org.terasology.module.ResolutionResult;
 import org.terasology.naming.Name;
 import org.terasology.naming.NameVersion;
 import org.terasology.registry.In;
-import org.terasology.rendering.assets.texture.TextureRegion;
 import org.terasology.rendering.nui.Canvas;
 import org.terasology.rendering.nui.CoreScreenLayer;
 import org.terasology.rendering.nui.WidgetUtil;
@@ -49,11 +48,12 @@ import org.terasology.rendering.nui.layers.mainMenu.moduleDetailsScreen.ModuleDe
 import org.terasology.rendering.nui.layers.mainMenu.savedGames.GameInfo;
 import org.terasology.rendering.nui.layouts.ScrollableArea;
 import org.terasology.rendering.nui.widgets.UIButton;
+import org.terasology.rendering.nui.widgets.UIImage;
 import org.terasology.rendering.nui.widgets.UILabel;
 import org.terasology.rendering.nui.widgets.UIList;
 import org.terasology.rendering.nui.widgets.UITabBox;
 import org.terasology.rendering.nui.widgets.UIText;
-import org.terasology.rendering.nui.widgets.slideshow.UITimedImageSlideshow;
+import org.terasology.rendering.nui.widgets.UIImageSlideshow;
 import org.terasology.utilities.time.DateTimeHelper;
 import org.terasology.world.biomes.Biome;
 import org.terasology.world.biomes.BiomeManager;
@@ -103,6 +103,7 @@ public class GameDetailsScreen extends CoreScreenLayer {
     private Map<String, List<String>> blockFamilyIds;
     private List<String> errors;
     private UIButton openModuleDetails;
+    private UIImageSlideshow previewSlideshow;
 
     @Override
     public void initialise() {
@@ -118,14 +119,17 @@ public class GameDetailsScreen extends CoreScreenLayer {
         worldDescription = find("worldDescription", ScrollableArea.class);
         descriptionContainer = find("descriptionContainer", ScrollableArea.class);
         openModuleDetails = find("openModuleDetails", UIButton.class);
+        previewSlideshow = find("preview", UIImageSlideshow.class);
 
         if (descriptionContainer != null && worldDescription != null && descriptionTitle != null &&
-                gameModules != null && gameWorlds != null && biomes != null && blocks != null && openModuleDetails != null) {
+                gameModules != null && gameWorlds != null && biomes != null && blocks != null && openModuleDetails != null ||
+                previewSlideshow != null) {
 
             setUpGameModules();
             setUpGameWorlds();
             setUpBlocks();
             setUpBiomes();
+            setUpPreviewSlideshow();
 
             openModuleDetails.subscribe(button -> openModuleDetailsScreen());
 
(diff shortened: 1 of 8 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/gameDetailsScreen/GameDetailsScreen.java
```diff
@@ -118,14 +119,17 @@ public void initialise() {
         worldDescription = find("worldDescription", ScrollableArea.class);
         descriptionContainer = find("descriptionContainer", ScrollableArea.class);
         openModuleDetails = find("openModuleDetails", UIButton.class);
+        previewSlideshow = find("preview", UIImageSlideshow.class);
 
         if (descriptionContainer != null && worldDescription != null && descriptionTitle != null &&
-                gameModules != null && gameWorlds != null && biomes != null && blocks != null && openModuleDetails != null) {
+                gameModules != null && gameWorlds != null && biomes != null && blocks != null && openModuleDetails != null ||
```
jellysnake: Should this not be `&& previewSlideshow != null`, not `||`

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -29,15 +29,16 @@
...
 /**
- * This widget displays images in sequence in an image slideshow. Switching images automatically after a time period.
+ * This widget displays images in sequence in an image slideshow.
+ * Switching images automatically after a time period or manually.
  */
-public class UITimedImageSlideshow extends CoreWidget {
+public class UIImageSlideshow extends CoreWidget {
 
     private Binding<UIWidget> currentImage = new DefaultBinding<>();
     private List<UIImage> images = new ArrayList<>();
```
jellysnake: I would like to see this have `@LayoutConfig` as well

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -29,15 +29,16 @@
...
 
 /**
- * This widget displays images in sequence in an image slideshow. Switching images automatically after a time period.
+ * This widget displays images in sequence in an image slideshow.
+ * Switching images automatically after a time period or manually.
  */
-public class UITimedImageSlideshow extends CoreWidget {
+public class UIImageSlideshow extends CoreWidget {
 
     private Binding<UIWidget> currentImage = new DefaultBinding<>();
```
jellysnake: ### Option B
Change this type to `UIImage` (no binding)

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
jellysnake: ### Option A
Change this to a `List<TextureRegion>`

### Option B
Change this to a `List<UIWidget>` and rename appropriately

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -51,16 +52,30 @@
...
 
+    /**
+     * Whether the slideshow automatically switch images.
+     */
+    @LayoutConfig
+    private boolean auto = true;
+
     @Override
     public void onDraw(Canvas canvas) {
         if (currentImage.get() != null) {
```
jellysnake: Can be replaced with `currentImage.isPresent()`

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -51,16 +52,30 @@
...
             currentImage.get().onDraw(canvas);
         }
     }
 
+    @Override
+    public Vector2i getPreferredContentSize(Canvas canvas, Vector2i sizeHint) {
+        if (currentImage.get() != null) {
```
jellysnake: Can be replaced with `currentImage.isPresent()`

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -128,22 +116,73 @@ public void clean() {
...
+                if (infinite) {
+                    index = 0;
+                } else {
+                    stop();
+                }
+            } else {
+                index++;
+            }
+            if (prevIndex != index) {
+                currentImage.set(images.get(index));
```
jellysnake: ### Option B
Change this from setting `currentImage` to the new image, just set the _texture_ to the new texture.

If you wanted to get really fancy you could also use a binding, and bind the texture to `images.get(index)` via `ReadOnlyBinding` this would make the texture automatically update without you having to set it.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -128,22 +116,73 @@ public void clean() {
...
+                if (infinite) {
+                    index = size - 1;
+                } else {
+                    stop();
+                }
+            } else {
+                index--;
+            }
+            if (prevIndex != index) {
+                currentImage.set(images.get(index));
```
jellysnake: See comment in `nextImage()`

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
```diff
@@ -128,22 +116,73 @@ public void clean() {
...
+
+    public float getSpeed() {
+        return speed;
+    }
+
+    public void setSpeed(float speed) {
+        this.speed = speed;
+    }
+
+    public List<UIImage> getImages() {
```
jellysnake: ### Option A
This would then return a list of `UIWidget`

### Option B
This would then return a list of `TextureRegion`

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/gameDetailsScreen/GameDetailsScreen.java
ar0ne: thanks, fixed

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
ar0ne: Honestly, both these options break Single responsibility principle. I'm pretty sure that we aren't going to have another types for slideshow in future. 
So, I mean we don't need to have so generic type UIWidget for **ImageSlideshow**. It just doesn't make any sense. The same for TextureRegion. I'd try to avoid using textures in this widget, just because it shouldn't care about so specific type of content, which slideshow uses.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
ar0ne: isPresent is method from Optional. What do you mean? 
Do you propose to make *currentImage* optional?

engine/src/main/java/org/terasology/rendering/nui/widgets/UIImageSlideshow.java
ar0ne: To be honest, I've understood that this kind of bindings decreases performance of whole UI. 
And in all my recent PRs I'm trying do not use it too often, only where it really might help.

GooeyHub:

GooeyHub:

Cervator:
Tested out and merged :-)

@jellysnake did your comments did addressed? I see a bunch of collapsed review comments and further commits. If not maybe you and/or @ar0ne could turn it into a nice Good First Issue request or so? Seems like it might be a good candidate

Cervator:
Another quick for reference bit: I caught a couple bugs where this PR hadn't been applied to the record and replay screens, causing them to fail dealing with the preview/image change. Fixed in https://github.com/MovingBlocks/Terasology/commit/fb5cdefd013b038d264fb56f3e56e2577a493912

Also took out a newline in a log statement, those tend to be evil as they break the flow of the log :-)

PR #3459, opened 2018-07-31
Save screenshots regularly

ar0ne:
### Contains

We wanted more game preview images, we got it!

### How to test

Just need to create a new game or load saved one. Then play and check that game takes screenshots regularly, and you can see them in detailed screens.

* [x]  **Replays** take screenshots and saves them to `saves` dir.


GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/persistence/internal/ReadWriteStorageManager.java
```diff
@@ -439,6 +443,8 @@ private void startAutoSaving() {
         saveTransaction = createSaveTransaction();
         saveThreadManager.offer(saveTransaction);
 
+        saveGamePreviewImage();
```
Cervator: IMHO this is too frequent to do automatic screenshots. Autosave is IIRC every 30 seconds? That's 120 an hour, or about 8 hours of gameplay to get to a thousand screenshots, which at an average size of 100kb would be 100 MB of screenshots :-)

I'd stick to generating an automatic screenshot on every manual save (exiting the world). Optionally we could add something under player settings to automatically take screenshots every x but I'm not convinced there'd really be a demand for that.

engine/src/main/java/org/terasology/persistence/internal/StoragePathProvider.java
```diff
@@ -26,6 +26,7 @@
 public class StoragePathProvider {
     private static final String PLAYERS_PATH = "players";
     private static final String WORLDS_PATH = "worlds";
+    private static final String PREVIEWS_PATH = "previews";
```
Cervator: Rather than add a unique path for previews I suggest we just consider regular screenshots and previews the same thing and store both in a "screenshots" in the game save dir. I'm not sure if we should duplicate screenshots to a global "screenshots" in the data dir root or even have one at all (the three tile images are just for debugging and could probably be disabled or get an option to enable them somewhere)

engine/src/main/java/org/terasology/persistence/internal/GamePreviewImageProvider.java
```diff
@@ -0,0 +1,147 @@
...
+            final String fileName = path.getFileName().toString().split("\\.")[0];
+            try {
+                counter = Math.max(counter, Integer.parseInt(fileName) + 1);
+            } catch (NumberFormatException e) {
+                logger.warn("Could not parse {} as integer (not an error)", fileName, e);
+            }
+        }
+        return counter + JPG_FILE_TYPE;
+    }
+}
```
Cervator: Trivial: newline at the end of every text file

Cervator:
Noticed another quirk: If you run a replay at the end on exiting a preview is saved - but there may not even be an associated save game (much like exiting from a multiplayer session). So I ended up with a `saves/Game94/previews/1.jpg` - I guess the recording originally came from a "Game94" ? I had renamed my "saves" dir for unrelated testing, so ended up with just that odd orphaned dir.

May be worth also checking what happens in multiplayer

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/persistence/internal/GamePreviewImageProvider.java
ar0ne: Fixed

engine/src/main/java/org/terasology/persistence/internal/StoragePathProvider.java
ar0ne: From my perspective, we should not mix user's screenshots and random preview images. Generally, because they are for different purposes. And it might be real pain in the neck to manage copies or links in global screenshot folder for all existing and old saved games.

GooeyHub:

GooeyHub:

ar0ne:
Rebase this branch to #3460.

Issue #3444, opened 2018-07-21
Command line toggle for the module sandbox


Cervator:
At present to disable the module sandbox for testing you need to edit `ModuleManagerImpl.java` like so: https://gist.github.com/Cervator/85da99b3400d880f6eae4ac3ca992676 (diff it against a normal version locally to see the 4 changed lines)

To ease testing it would nice to have a command line flag (`-permissiveSecurity` ?) and potentially an IntelliJ run config that enables it. See `Terasology.java` for the current usage of command line flags and `ide.gradle` for the run configs (clone an existing run config block then add the flag)

`ModuleManagerImpl.java` would need to be edited to support either case and default to security being on like normal. If the flag is found then instead switch to permissive mode like in the linked Gist.

To complete this we should also add details to https://github.com/MovingBlocks/Terasology/wiki/Modding-API as well as the usage text in `Terasology.java`

_Edit:_ As an example look at how `NO_SAVE_GAMES` in `Terasology.java` is used. It writes a config to `System` which can then be later read in the location needing the value - `InitialiseWorld`:

`boolean writeSaveGamesEnabled = context.get(Config.class).getSystem().isWriteSaveGamesEnabled();`

starlord1311:
can i take up this issue? @Cervator

Cervator:
Sure @starlord1311 - you don't need to ask, although posting a comment that you're going to try it out can be helpful if others are interested as well. Just coordinate via the issue :-)

And welcome!

starlord1311:
@Cervator can you tell me what i am supposed to do?

Cervator:
@starlord1311 so in these kinds of issues marked with Good First Issue we should already have put a decent amount of the details in the initial description, but it is up to you to dig deeper and figure out exactly what to implement :-)

In this case I pointed to a couple key places in the code that already do similar things, and where to put the new stuff. So you'll want to look at those places and try to understand what's going on.

Look at an existing command line flag in `Terasology.java` and look around for how it is used. In IntelliJ you can CTRL+left click classes or right click things then select "Find Usages" to jump between a single feature spread across different files.

Copy-paste the Gist's contents into your local copy of `ModuleManagerImpl.java` to see the 4 lines that change, and understand why. Then consider what conditional logic you can add to make the effective change vary without having to change the code. You'll want to hook a new `if` up to the new command line flag.

[...]

PR #3461, opened 2018-08-03
Gradle support for JNLua and its natives

Cervator:
For @asiekierka's GSOC project. Should allow Kallisti to build as a module which then would allow KComputers to depend on it instead of having to embed it as source. Also should package up JNLua and its natives along with the game zip.

To test: merge this and try `gradlew idea` (which should trigger `extractNatives`)

https://github.com/MovingBlocks/JNLua/pull/2 included the part to deal with natives in Jenkins. Two jobs are now involved:

* http://jenkins.terasology.org/job/JNLua/ - builds the jar and Win/Linux natives. Runs on a Debian builder node I prepared manually. We should be able to automate that or just make a snapshot to restore as needed
* http://jenkins.terasology.org/job/JNLua_Mac/ - builds Mac natives then zips all of them and publishes to Artifactory. Runs on my Mac - the job will trigger after the first one finishes, but queue waiting for the Mac to become available. That's nice and visible in Jenkins so I should notice (or be poked if needed)

After this is merged we can build module-Kallisti then make KComputers depend on it.

Only remaining todo will be #3444 until we sort out the sandbox. Not sure if we also need some whitelisting but atm that won't be enough so gotta turn security permissive for now

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,9 +78,12 @@ dependencies {
 
     // Config for our code analytics lives in a centralized repo: https://github.com/MovingBlocks/TeraConfig
     codeMetrics group: 'org.terasology.config', name: 'codemetrics', version: '1.2.0', ext: 'zip'
+
+    // Natives for JNLua (Kallisti, KComputers)
+    natives group: 'org.terasology.jnlua', name: 'jnlua_natives', version: '0.1.0-SNAPSHOT', ext: 'zip'
 }
 
-task extractWindowsNatives(type:Sync) {
+task extractWindowsNatives(type:Copy) {
     description = "Extracts the Windows natives from the downloaded zip"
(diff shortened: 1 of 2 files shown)


GooeyHub:

pollend:
could this be extended down the line to include other natives. I was thinking about the bullet natives and the openVR natives.

asiekierka:
Looks good, but I think making all natives rely on a similar system (post-GSoC, at least) is a very good idea.

Cervator:
Yeah I definitely want to cleanup natives some more. OpenVR also needs a proper build job that publishes to Artifactory then use from there instead of embed in the repo. Same for future stuff like Bullet. Just a question of available effort really.

PR #3462, opened 2018-08-03
Improvements to the Record and Replay classes' JavaDoc

iaronaraujo:
### Contains

JavaDoc improvements for the Record and Replay classes

### DIFF
--- a/engine/src/main/java/org/terasology/recording/CharacterStateEventPositionMap.java
+++ b/engine/src/main/java/org/terasology/recording/CharacterStateEventPositionMap.java
@@ -22,8 +22,9 @@ import java.util.HashMap;
 import java.util.Map;
 
 /**
- * Saves the "position" and "velocity" variable for every CharacterStateEvent so that the player location can be
- * reproduced correctly during a replay.
+ * When a CharacterMoveInputEvent is caught by ServerCharacterPredictionSystem, a new CharacterStateEvent is created and
+ * processed without passing through the EventSystem. Since this event is important to reproduce the player's location,
+ * this class is used to record its "position" and "velocity" attribute and update it during a Replay.
  */
 public class CharacterStateEventPositionMap {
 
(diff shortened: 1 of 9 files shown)


GooeyHub:

engine/src/main/java/org/terasology/recording/CharacterStateEventPositionMap.java
```diff
@@ -22,8 +22,9 @@
 import java.util.Map;
 
 /**
- * Saves the "position" and "velocity" variable for every CharacterStateEvent so that the player location can be
- * reproduced correctly during a replay.
+ * When a CharacterMoveInputEvent is caught by ServerCharacterPredictionSystem, a new CharacterStateEvent is created and
```
Cervator: General note for javadoc: Try to come up with a once sentence summary  that fits on one line and ends with a `.` - that goes into the summary section for the class in the javadoc site. If you end up with a multi-sentence summary only the first sentence makes it which can make the docs read kinda funny.

Then get into as much detail as you like after the first sentence/paragraph 👍

engine/src/main/java/org/terasology/recording/DirectionAndOriginPosRecorder.java
```diff
@@ -21,8 +21,10 @@
 import java.util.Deque;
 
 /**
- * Saves the variables 'direction' and 'originData' to be used on replays since holding the mouse buttons can cause weird
- * behavior during replays.
+ * During Record and Replay development, it was noticed that sometimes holding the right or left mouse button caused
```
Cervator: Same story plus you don't generally want to add historical information in a summary. You could have that in a separate paragraph later that goes into justification/reasoning. So a first sentence/paragraph here might be:

"Responsible for managing some variables used in a workaround during recording"

Then next sentence/paragraph you can go into the details, and perhaps what the future may hold as far as a fix that could avoid the workaround

engine/src/main/java/org/terasology/recording/EventCatcher.java
```diff
@@ -21,7 +21,8 @@
 import java.util.List;
 
 /**
- * This class is responsible for catching the events during a Record and send the desired ones to the RecordedEventStore.
+ * Gets the events sent in the EventSystemImpl and adds their copies to the RecordedEventStore if they are of a type
+ * selected to be recorded.
```
Cervator: How would you find out whether something is selected to be recorded? Would be a great spot for a javadoc link to an associated class :-)

PR #3469, opened 2018-08-08
Reduce memory allocations

theobisproject:
### Contains

Reduction of memory allocations while playing the game. As result this will lead to fewer garbage collections.  
Found this while recording a view minutes of gameply with Java Mission Control.

### How to test

Run the game with the debug information displayed and compare the memory information.  
Alternatively check the allocations with some profiler.

Maybe replace all occurences of Sides.values() in the codebase with a static variable?

### DIFF
--- a/engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
+++ b/engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
@@ -201,8 +201,8 @@ public class StandardBatchPropagator implements BatchPropagator {
         Block block = world.getBlockAt(pos);
         for (Side side : BLOCK_SIDES) {
             byte spreadValue = rules.propagateValue(value, side, block);
-            Vector3i adjPos = side.getAdjacentPos(pos);
             if (rules.canSpreadOutOf(block, side)) {
+                Vector3i adjPos = side.getAdjacentPos(pos);
                 byte adjValue = world.getValueAt(adjPos);
                 if (adjValue < spreadValue && adjValue != PropagatorWorldView.UNAVAILABLE) {
                     Block adjBlock = world.getBlockAt(adjPos);
--- a/engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
+++ b/engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
@@ -28,7 +28,6 @@ import org.terasology.world.chunks.LitChunk;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Stream;
 
 /**
  * Batch propagator that works on a set of changed blocks
--- a/engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
+++ b/engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
@@ -36,7 +36,6 @@ import java.util.Set;
 public class StandardBatchPropagator implements BatchPropagator {
 
     private static final byte NO_VALUE = 0;
-    private static final Side[] BLOCK_SIDES = Side.values();
 
     private PropagationRules rules;
     private PropagatorWorldView world;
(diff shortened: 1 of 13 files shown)


GooeyHub:

Cervator:
Nice find! Did you happen to capture any logging, details, or screenies? How did you set up with Java Mission Control and even find it in the first place?

@GooeyHub: add to whitelist

@asiekierka / @pollend / - could either of you weigh in on this? The general area may change architecturally in a major way . Maybe @Adrijaned or @vampcat would have some thoughts as well (vector pool musings from time to time - speaking of, https://github.com/MovingBlocks/TeraMath/pull/53 might have stalled)

GooeyHub:

engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
```diff
@@ -28,6 +28,7 @@
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
```
Adrijaned: This import seems useless here

vampcat:
This looks good to me as well - although I do wonder why we can't make a static function `Side.getValues()` that essentially returns the same thing, but doesn't require the user class to create a static SIDES variable?

theobisproject:
@Cervator To enable the Java Mission Control capturing simply add the following JVM parameters when starting the game.
`-XX:+UnlockCommercialFeatures -XX:+FlightRecorder -XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints -XX:StartFlightRecording=name=MyRecording,settings=profile,filename=terasology.jfr,dumponexit=true `
Then finding the issue is not that hard in this case. As it can be seen in the first screenshot the `Side[]` has the most allocated memory. When selecting it a stacktrace is shown at the button. Then I had simply followed the stacktrace and looked for `Side[]` allocations.

**Current develop branch memory allocations**
current_develop_branch_memory_allocations

**Memory allocation with my fix**
reducedmemory_branch_memory_allocations

**How I measured**

1. Start Terasology from scratch
2. Load already generated world
3. Smash some sand blocks
4. Close the game

@vampcat This could be another method to fix this issue. For me this fix was the most obvious fix and everyone can immediately see what the difference between the old and the new version is. The static function is hiding this a bit.
There is even one more possible solution to this: a static EnumSet in the Side enumeration.

For a bit of background on this problem there is a recent article with possible solutions and some links to the JDK mailing list about the topic. https://marxsoftware.blogspot.com/2018/07/memory-hogging-enumvalues.html

@Adrijaned I would delay the addition of JavaDoc until we decided on the final implementation.

engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
theobisproject: Yes, you are correct. Deleted it.

GooeyHub:

Cervator:
Thanks for the great details @theobisproject ! Yeah we have a lot of low hanging fruit when it comes to profiling and fixing easy issues improving performance.

Poking @vampcat for a quick follow-up on the implementation detail 👍

vampcat:
@theobisproject That's some sweet improvement we're getting! I knew that vector3i creation was a problem, I had no clue enum was an even bigger problem!

I'm perfectly happy with both (actually, all 3) ways to implement this :)

theobisproject:
@vampcat I will then go for the `EnumSet` in the `Side` class solution. While this is not the most efficient solution regarding memory (this would be the array) it provides all benefits the `Set` interfaces provides (someone could use streams if he wants).

GooeyHub:

Cervator:
Tested out and merged - thanks!

@theobisproject would you by chance be up for putting your instructions and anything else related to profiling you can come up with into a wiki page?

theobisproject:
@Cervator Yes, I can write an entry in the wiki. Is there anything special I should know. Never done it before.
For the instructions wouldn't it be better to add them as new run configration in IntelliJ to the already existing ones?

Issue #3266, opened 2018-02-23
Item remains grabbed after closing inventory, even in pause menu


Maximetinu:
### What you were trying to do
I've found this bug while I was messing around with the inventory. If I try to grab an item and then I close the inventory, the item still remains on a limb. If I open pause menu it's still showing up along with the cursor. Then, if I close and open again the game or auto save happens with the item grabbed and I exit the game then, the transfering item is lost.

### What actually happened
Far as I can tell by reading the code, the transfering item is coded as a single-slot inventory, as can be seen in the prefab file and the CharacterInventorySystem class (line 83). Despite being stored in the CharacterComponent as movingItem, it seems that the movingItem is not being persisted on save.

### How to reproduce
* Step 1 - Open Inventory and click an item
* Step 2 - With the item grabbed, press ESC twice
* Step 3 - As you can see, the item is attached to your cursor even in pause menu. Then click on "Exit to the main menu".
* Step 4 - Load the game again. Now, the item is lost, as the game was saved when you leave the world and the moving item wasn't saved.

### Log details and game version
Current latest stable: 74 alpha 2018-02-05T04:45:49Z

### Computer details
OS: Win 10 x64 CPU: i5-4690 RAM: 8GB Graphic Card: GeForce GTX 970 but irrelevant as it happens in other computers too

### Video
[Vimeo video showing the bug](https://vimeo.com/257146394)

Maximetinu:
I'm going to try solving this issue.

To begin with, I think it should put the grabbed item in the first free inventory slot when the inventory is closed, and in case of it's full, drop the item. It seems that the pause menu cannot be open if the inventory is already opened, but if possible, it should be controlled too. Also, if auto save happens while an item is being grabbed, to avoid losing the item the same behaviour should happen before saving, thus trying to store the item in the inventory first. You can tell me what do you think about this rules.

On the surface, I think I will have to check the inventory logic inside the core module, starting with the CharacterInventorySystem which is where the transferSlot is created, and the CharacterComponent and the stuff around it. Correct me if I'm wrong or I missed something. Of course, I still have to go deeper in code.

jellysnake:
Thanks for spotting this. Your proposed solution seems like a reasonable one, however I'd generalise it to happening when there is a change in the NUI Screen being shown. So if a new screen is opened ontop of the inventory, if the inventory screen is closed etc etc. I think there is a method that is called when this occurs.
As for the saving, I think that it would be best to move the item to the inventory -.
[...]

PR #3470, opened 2018-08-10
Change InventoryScreen.class so movingItem not lost on exit

burtojae:
### Contains

Fixes issue [#3266](https://github.com/MovingBlocks/Terasology/issues/3266) by taking code from the InventoryCell and DropItem region classes and modifying it to work in the onClose() method of the InventoryScreen class.

Instead of the old behavior, the item in the transfer slot should now return to the player's inventory if there is space, or drop on the ground if the inventory is full.

### How to test

Create a world with the Core Gameplay template and open your inventory. Left click on an item so you have it grabbed and then press either escape or i to close your inventory. The item should have returned to your inventory with the same logic that shift+left-click uses to move items from the hotbar to the main inventory and back.

Now open your inventory again and fill every empty slot with a torch. Place the remaining torches in the slot where the pickaxe was so you have a pickaxe on your cursor and an inventory full of torches. Close your inventory. The pickaxe should drop on the ground in front of you with the same logic that drops an item when you left click with it off of an inventory cell.

- [ ] Currently does not have the same behavior applied to auto-save as suggested by @Maximetinu and @jellysnake.

### DIFF
--- a/modules/Core/src/main/java/org/terasology/logic/inventory/InventoryUIClientSystem.java
+++ b/modules/Core/src/main/java/org/terasology/logic/inventory/InventoryUIClientSystem.java
@@ -42,13 +42,21 @@ import java.util.List;
 
 @RegisterSystem(RegisterMode.CLIENT)
 public class InventoryUIClientSystem extends BaseComponentSystem {
-
-    int itemMovedSlot = -1;
+    
+    EntityRef movingItemItem = EntityRef.NULL;
+    
+    int movingItemCount = 0;
 
     private static final Logger logger = LoggerFactory.getLogger(InventoryCell.class);
 
     @In
     private NUIManager nuiManager;
+    
+    @In
+    private InventoryManager invManager;
+    
+    @In
+    private LocalPlayer localPlayer;
 
     @Override
     public void initialise() {
@@ -101,7 +109,7 @@ public class InventoryUIClientSystem extends BaseComponentSystem {
     }
 
     private EntityRef getTransferEntity() {
-        return CoreRegistry.get(LocalPlayer.class).getCharacterEntity().getComponent(CharacterComponent.class).movingItem;
+        return localPlayer.getCharacterEntity().getComponent(CharacterComponent.class).movingItem;
     }
 
     @Override
@@ -114,12 +122,14 @@ public class InventoryUIClientSystem extends BaseComponentSystem {
           similar to what was needed here to take them out of the transfer
           slot and sort them into the inventory.
         */
-        EntityRef playerEntity = CoreRegistry.get(LocalPlayer.class).getCharacterEntity();
-        EntityRef movingItem = playerEntity.getComponent(CharacterComponent.class).movingItem;
+        EntityRef playerEntity = localPlayer.getCharacterEntity();
+        EntityRef movingItemSlot = playerEntity.getComponent(CharacterComponent.class).movingItem;
 
-        EntityRef movingItemItem = CoreRegistry.get(InventoryManager.class).getItemInSlot(movingItem, 0);
+        movingItemItem = invManager.getItemInSlot(movingItemSlot, 0);
+        
+        movingItemCount = invManager.getStackSize(movingItemItem);
 
-        EntityRef fromEntity = movingItem;
+        EntityRef fromEntity = movingItemSlot;
         int fromSlot = 0;
 
         InventoryComponent playerInventory = playerEntity.getComponent(InventoryComponent.class);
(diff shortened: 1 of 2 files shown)


GooeyHub:

jellysnake:
I'll do a proper eyeball of the code later. Just triggering the build bot for now

@GooeyHub ok to test

Cervator:
Welcome @burtojae and thanks for your first PR! Good fix and nice description 👍 

I've had a talk with @GooeyHub who should now listen to @jellysnake (had to add as a bot admin in configure for http://jenkins.terasology.org/job/TerasologyPRs/)

@GooeyHub: ok to test

GooeyHub:

Cervator:
@burtojae that's correct - modules wouldn't be able to hook directly into the auto-save process. *However* that usually just ends up indicating that we haven't yet added a useful hook into the engine that modules can hang stuff on :-)

So rather than start your logic related in a module trying to interact with `ReadWriteStorageManager` you could add a new event in the engine auto save system that gets triggered every auto-save that advertises "Hey, an auto save is happening, is anybody interested in doing anything?". Module code can then introduce an event handler that catches that and acts on it before the auto-save continues.

You may still need to move logic a little around in case you need something more to happen in the engine during auto-save, or if you need more information to come along with the event being triggered by the auto-save, or sending some back that the auto-save can react to if a subscriber wanted something to happen.

That might end up going down a deeper rabbit hole so I'll just leave you with that for now - let me know if it helps and if you would like more related info or examples. Ideally hop on Discord or ask for a Slack invite sometime and we can work on it live sometime when the right people are online 👍

jellysnake:
Actually, there is already a hook in place for this.  
Systems have four methods called `preSave`, `preAutoSave`, `postSave` and `postAutoSave`.

Those methods are called just before and just after a save is made. I'd suggest using these methods in the `InventoryUIClientSystem` to call a method on the screen which handles the actual moving of the object around.

You can access the screen from the system by using `NuiManager#isOpen()` and `NuiManager#getScreen()` to check for and get the screen. This will then give you an object that you can use to call the method on thee screen.

GooeyHub:

GooeyHub:

jellysnake:
Minor thing I've noticed from skimming your code.

Using `CoreRegistry` to get classes is actually deprecated in favour of `@In`

The syntax for `@In` is:

```java
@In
private ClassToInject fieldName;
```

For example the `NuiManager` in this class is being injected via `@In`.

Anything you can get through `CoreRegistry` you can also inject. The only caveat is that it only works in some types of classes. However Systems are one of those so you can use it here.

burtojae:
Is there a good way of comparing two stacks of items to see if they both contain the same item? The canStackTogether() method in the InventoryManager doesn't work for that, and comparing items directly with == only seems to work for unstackable items like tools. I'm trying to figure out the best way to put the right item(s) back into the transfer slot after a save. 

I got rid of the CoreRegistry uses and replaced them with injected classes instead by the way.

GooeyHub:

modules/Core/src/main/java/org/terasology/logic/inventory/InventoryUIClientSystem.java
```diff
@@ -24,17 +26,37 @@
...
 import org.terasology.rendering.nui.NUIManager;
+import org.terasology.rendering.nui.layers.ingame.inventory.InventoryCell;
+
+import java.util.ArrayList;
+import java.util.List;
 
 @RegisterSystem(RegisterMode.CLIENT)
 public class InventoryUIClientSystem extends BaseComponentSystem {
+    
+    EntityRef movingItemItem = EntityRef.NULL;
```
Cervator: You probably want some sort of access modifier here, likely `private`

modules/Core/src/main/java/org/terasology/logic/inventory/InventoryUIClientSystem.java
```diff
@@ -24,17 +26,37 @@
...
+    
+    int movingItemCount = 0;
```
Cervator: Here too. Access modifiers help clarity :-) You can also javadoc these fields just for a quick one-liner about what they're for (ideally using the single line `/** Some description */` format)

modules/Core/src/main/java/org/terasology/logic/inventory/InventoryUIClientSystem.java
```diff
@@ -24,17 +26,37 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(InventoryCell.class);
```
Cervator: Logger usually goes at the very top of the fields. If you are using an IDE look for the Checkstyle plugin it should highlight these

modules/Core/src/main/java/org/terasology/logic/inventory/InventoryUIClientSystem.java
```diff
@@ -69,4 +91,121 @@ public void onToggleInventory(InventoryButton event, EntityRef entity, ClientCom
             // do not consume the event, so that the inventory will still open
         }
     }
+
+
+    /*
+      The numbersBetween() and getTransferEntity() methods were
```
Cervator: Even if this is a private method still use the `/**` javadoc approach instead of a multi-line plain comment. Javadoc is nice! And usually in an IDE if you `/**<enter>` on a method, field, or class the IDE will auto-generate a Javadoc block for you including return values and parameters. 

Don't be shy about making lines longer, our max line length is like 175 but for comments you could try to stay below 120 or so.

modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryScreen.java
```diff
@@ -43,4 +61,103 @@ public EntityRef get() {
...
     }
+
+    /*
+      The numbersBetween() and getTransferEntity() methods were
+      originally in the InventoryCell class. They were copied over
+      to here because they are private functions that the
+      moveItemSmartly() method needs to function. The first section
+      of code in onClosed() is based on the moveItemSmartly() method.
+    */
+    private List<Integer> numbersBetween(int start, int exclusiveEnd) {
```
Cervator: Incidentally now that we seem to have this method in three different places it might be time to add it to https://github.com/MovingBlocks/TeraMath instead as a static method then just use it from there. Don't worry about that for this PR (gets a little more complicated to involve a lib update) but could be an idea for what to try next :-)

modules/Core/src/main/java/org/terasology/rendering/nui/layers/ingame/inventory/InventoryScreen.java
```diff
@@ -43,4 +61,103 @@ public EntityRef get() {
...
+        List<Integer> numbers = new ArrayList<>();
+        for (int number = start; number < exclusiveEnd; number++) {
+            numbers.add(number);
+        }
+        return numbers;
+    }
+
+    private EntityRef getTransferEntity() {
+        return CoreRegistry.get(LocalPlayer.class).getCharacterEntity().getComponent(CharacterComponent.class).movingItem;
```
Cervator: Three more sneaky usages of CoreRegistry hiding in this class. Plus a couple imports that could be removed.

There may be some cases in which `@In` would initialize a local service too early meaning it then won't work, but give it a try and see if it works

Cervator:
Whoops, forgot to actually leave a comment with the review.

Thanks for the catch @jellysnake - I forgot we had the autosaves there too :-)

Appreciate you continuing with this @burtojae ! I made a few minor comments with tiny requested changes. This could probably done with less code, but since you're copying in / adapting existing code that might be better in a separate PR.

I don't know off-hand of a stack check like that, but Jelly might

jellysnake:
I know that one of our summer of code projects, light & shadow, had a similar desire to compare entities in the inventory.  

Here I'd recommend a similar approach to what L&S used which is to make a new component and use it as a flag component. You can check if an entity has a given component using `.hasComponent(TheComponentClass.class);`

So you would add the flag component just before adding it to the inventory and remove it just after taking it out.

A completely alternative approach to the problem, if dealing with the inventory proves to be too fiddly, is to make a new component with a single EntityRef field. Then to save the held item add a new instance of this component to the local player. 

To undo after save, simply remove the component from the player.

An aside to consider as well that I realised whilst typing the comment is that you should also consider the game being loaded from save, and handle getting the saved item in that case

jellysnake:
@burtojae just following up on this PR. Any updates/new work?

skaldarnar:
This should be redirected against [Terasology/Inventory](https://github.com/Terasology/Inventory) as the inventory code was extracted to its own module by #3419.

@burtojae @jellysnake can you take care of this?

jellysnake:
Closing this as moved to a new PR in Terasology/Inventory#2 as the related code has been extracted to there

PR #3473, opened 2018-08-31
Fix liquid rendering

### LINKED ISSUE #2012: trimmedLoweredCube shape and water rendering (maybe larger hard coded rendering issue)
We found and "fixed" some shapes with issues recently related to face configuration being off so adjacent blocks could cause the wrong block face to occlude (#1999). However, trimmedLoweredCube has turned out to be special as it is being used for water blocks.

I hadn't thought about that before but it makes sense - water is slightly lower inside a block space than solid blocks to make water appear slightly lower. Yet you don't see this open space in water blocks below the surface. Previously the top face in the shape was set to TOP rather than CENTER.

That made water blocks occlude as if two full faces were next to each other, yet a different block made with the trimmedLoweredCube shape would show the gap. Under actual water you just get the general underwater effect with the blocks otherwise occluded until I guess some magic happens at land.

Having the shape instead set with its top face to CENTER correctly let normal blocks show that face rather than let you see through the world, but when the rendering for water came along and tried to get the TOP side's appearance it would crash with an NPE. That shape now only had CENTER, no TOP. Didn't notice when originally testing the shape fix as I was doing so in Cities where water is a lot more rare.

I'm not really sure what the fix is here, it is certainly outside my reach, I'm surprised I was even able to track it this far (and I'm sure I'm off on the details somehow anyway). There are comments in the area suggesting the rendering needs cleanup ( \* waves to @emanuele3d \* )

I switched the CENTER face in the shape back to TOP temporarily so it doesn't crash the game and added a guard to that one part of the rendering to check if the side actually has valid rendering data. But we probably should re-fix the shape and straighten out the rendering one day.
##

4Denthusiast:
### Contains

Fixes various problems relating to the meshes for liquid surfaces. I think this resolves #2012, but it's unclear exactly what problems that issue refers to. This breaks the API by removing Block.[get/set]LoweredLiquidMesh and replacing it with 2 new meshes for the bottom 0.9 units and top 0.1 units of the block. It would be possible to keep the original names for one of these, but that seemed inappropriate as neither of them mean the same as the original.

Generally, rather than making each liquid block responsible for the bottom 0.9 of itself and the top 0.1 of the block below, this PR makes each liquid block responsible for both portions of itself, leading to clearer and more correct logic.

It would be desirable to render surfaces at the boundary between different liquids, too, but there are separate difficulties with that, so I haven't attempted it here.

### How to test

An air-filled hole in the bottom of a body of water is one of the clearest examples of something that used to render incorrectly. Another case is one water block above another, with a solid block to the side of one of them (when viewed from the inside, there's some z-fighting in the old version). There may be others, too, but I can't remember what exactly.

If you want to be thorough in testing, this PR also changes how solid blocks are rendered, although the results should be the same.


GooeyHub:

PR #3477, opened 2018-09-06
Add rendering enums as API

4Denthusiast:
Since #3476 renders most of #3309 obsolete, this is the remaining parts, revealing the bits of API necessary for a custom implementation of BlockMeshGenerator, such as FlowingLiquids uses for making meshes for liquid blocks.

Although this is not currently necessary due to the module access restrictions being disabled, it still seems desirable to have this marked as API as it's actually used in a module (and harmless).

The things revealed do relate to some rather ad-hoc internals of rendering which may be changed at some point, but if that's a problem, it indicates that changes need to be made elsewhere so that access to these isn't necessary.

### Contains

Reveals ChunkVertexFlag and ChunkMesh.RenderType as API


4Denthusiast:
I've been looking more at how separation of chunk rendering into phases is done, and I think that it might make sense to change the ChunkMesh.RenderTypes enum, so it might not be a good idea to add it to the API yet.

PR #3479, opened 2018-09-07
Make water-waves actually vertical

### LINKED ISSUE #2450: Animated water looks absurdly strange from different angles

Trying to turn on animated water and look nice.
### What actually happened

When I attempted to turn on the animated water (though successfully turned on), the behaviors of the animated water seemed strange. When looked from different viewports (e.g. from the north and the south), the level of the water varies differently, and constantly changes according to viewport. Probably because the water is rendered separately from the blocks and entities, it seems to have a de-sync between water rendering and block rendering, where they should stick to each other at borders.

In other words, at the borders where water meet blocks, from some angles of view we may see through the water (a gap between the water surface and the edge of the land).
### How to reproduce
- Turn on animated water
- View from different angles

Pre-Alpha #63 / Alpha 3.
### Computer details
- Intel Core 2 Duo T7500
- NVidia Quadro FX570M
- 1GB Java Heap Space

**A screenshot would be added later.**
##

4Denthusiast:
### Contains

Change the offset for animated water from being vertical in view-space to being vertical in world-space. Resolves #2450.

It still doesn't look good, but it does look better.

### How to test

Turn on animated water in video settings. Look at it from various angles.

### DIFF
--- a/engine/src/main/resources/assets/shaders/chunk_vert.glsl
+++ b/engine/src/main/resources/assets/shaders/chunk_vert.glsl
@@ -159,7 +159,7 @@ void main()
         vec4 normalAndOffset = calcWaterNormalAndOffset(vertexWorldPos.xz);
 
         waterNormalViewSpace = gl_NormalMatrix * normalAndOffset.xyz;
-        vertexViewPos.y += normalAndOffset.w + waterOffsetY;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Checks out - thanks! 👍 

Does look a bit funnier yet with FlowingLiquids enabled now, but that's a whole other can of worms :-)

PR #3487, opened 2018-09-21
Cleaner Math: static Vector3i operators

LinusVanElswijk:
### Contains
- Adds static Vector3i operators.
- Replaces calls to member operator, where it makes code more concise and reduces clutter.
- Does _not_ increase the number of Vector3i constructor calls.

### How to test
- Run the game
- Run the engine tests 
- Can add Unit tests, if requested.

### DIFF
--- a/engine/src/main/java/org/terasology/math/ChunkMath.java
+++ b/engine/src/main/java/org/terasology/math/ChunkMath.java
@@ -167,15 +167,10 @@ public final class ChunkMath {
     }
 
     public static Region3i getChunkRegionAroundWorldPos(Vector3i pos, int extent) {
-        Vector3i minPos = new Vector3i(-extent, -extent, -extent);
-        minPos.add(pos);
-        Vector3i maxPos = new Vector3i(extent, extent, extent);
-        maxPos.add(pos);
+        Vector3i minPos = Vector3i.one().mul(-extent).add(pos);
+        Vector3i maxPos = Vector3i.one().mul(+extent).add(pos);
 
-        Vector3i minChunk = calcChunkPos(minPos);
-        Vector3i maxChunk = calcChunkPos(maxPos);
-
-        return Region3i.createFromMinMax(minChunk, maxChunk);
+        return Region3i.createFromMinMax(calcChunkPos(minPos), calcChunkPos(maxPos));
     }
 
     // TODO: This doesn't belong in this class, move it.
--- a/engine/src/main/java/org/terasology/math/Side.java
+++ b/engine/src/main/java/org/terasology/math/Side.java
@@ -23,6 +23,8 @@ import org.terasology.math.geom.Vector3i;
 import java.util.EnumMap;
 import java.util.EnumSet;
 
+import static org.terasology.math.Vector3iOperators.add;
+
 /**
  * The six sides of a block and a slew of related utility.
  * <br><br>
@@ -295,9 +297,7 @@ public enum Side {
     }
 
     public Vector3i getAdjacentPos(Vector3i position) {
-        Vector3i result = new Vector3i(position);
-        result.add(vector3iDir);
-        return result;
+        return add(position, vector3iDir);
     }
 
     public Side getRelativeSide(Direction direction) {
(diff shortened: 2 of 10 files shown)


GooeyHub:

LinusVanElswijk:
If accepted, I will try to add Vector3f static operators as well

pollend:
you don't need these static operators. It makes the code slightly shorter but you end up hiding the implementation.

GooeyHub:

Cervator:
Heya @LinusVanElswijk nice to see you again :-)

Seeing the `new Vector` part get moved to a central spot like this made me wonder about some recent discussion about using pools of vector objects to achieve performance improvements. Maybe that would be one potential neat extra

Maybe throw some javadoc on the new methods in the utility class? Clear enough that it is vector math, but still, good to have just a few words 👍 

@GooeyHub: add to whitelist

GooeyHub:

pollend:
@LinusVanElswijk maybe the alternative would be just to add it to TeraMath. Vector3i.add(a1,a2)

LinusVanElswijk:
I agree that it would be better if it was part of TeraMath.

If we compare
```
Vector3i z = new Vector3i(x).add(y);
Vector3i z = add(x,y);
```
Then the second version is a lot more clear IMO. And the visual clutter of the copy constructor adds up if the vector math is any more complex than this.

Cervator:
> Don't know if we can convince eachother :)

Yep that's a common challenge :-)

To me as somewhat of a lay person for this topic I do like the really short version a little better, but the total amount of code we're talking about is pretty small and risks drifting into [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding) - however, @pollend is really deep in the area to where it might matter more to be able to see at a glance that it is indeed a `new` instance of the object being created.

We talked a bit on Slack yesterday (incidentally @LinusVanElswijk you still have an account there, come say hi sometime! Let me know if you need a replacement invite for any reason) and I got into wondering about alternatives that would satisfy both requirements and/or achieve more than just wrap a tiny bit of code.

* My earlier mention of vector pools relate to https://github.com/MovingBlocks/TeraMath/pull/53 and I figure it would be easier to adopt pooling if the `new` part of the code is hidden within a wrapper. It makes it more intuitive to non-experts while experts probably can rely on the javadoc and deeper inspection for added details
* Moving to TeraMath came up, the `Vector3i` defined there actually already has a bunch of similar utility methods, but they're not static, and several are already overloaded. Seems like we should do whatever we do over there - but we might still need a separate class for full static wrappers
  * Adding them directly to `Vector3i` sounds pleasing to me but in a naive fashion: you might bump into overloading issues with the static and without static you still get the longer `something = Vector3i.add(x, y)
* Speaking of TeraMath it seems like every other project has their own math lib these days, and we've talked from time to time about adopting/merging others. I dunno how if at all any of them deal with this kind of question.
* Alternatively I started looking at the methods on the engine side and they could probably even be minimized further to where _more_ code ends up in TeraMath. But this might be too situational to give perfect solutions to all the things.

## Before (whole method, only 2 lines actually changed)

```java
  void someMethod() {
    Vector3i something = .getChunkZipPosition(chunkPos)
  }
...
    public Vector3i getChunkZipPosition(Vector3i chunkPos) {
        Vector3i result = new Vector3i(chunkPos);
        result.div(CHUNK_ZIP_DIM);	
        if (chunkPos.x < 0) {
            result.x -= 1;
        }
         if (chunkPos.y < 0) {
             result.y -= 1;
         }
         if (chunkPos.z < 0) {
             result.z -= 1;
         }
         return result;
```

## After (whole thing moved to TeraMath)

```java
  someMethod() {
    Vector3i something = TeraMath.divAbs(chunkPos, CHUNK_ZIP_DIM)
  }
```

At which point I guess you could also make `divAbs` statically available for the shorter yet `divAbs(chunkPos, CHUNK_ZIP_DIM)` and we're back to talking about the pros and cons of static wrappers again ;-)

To move this sideways a bit @LinusVanElswijk if you're interested in vector math in general (you could just have spotted some static utility you thought might look nicer) would you be interested in helping out with some of the open PRs over with TeraMath, and rebuilding this PR there? Mostly it is just @pollend wishing there were more people able to review that sort of stuff. @msteiger and @immortius have done some solid work there in the past but are usually low on time nowadays. Am curious what they might think about this situation.

LinusVanElswijk:
I'll consider the alternatives. I think it is safe to close this PR for now.

PR #3456, opened 2018-07-26
Type handling (serialization) improvements

eviltak:
A follow up of #3453, this PR simplifies the interface used to serialize objects and makes it extremely easy to serialize arbitrary objects using either Protobuf or Gson. `TypeHandlingIntegrationTest` shows how easy it is to serialize an arbitrary generic object with supertype references in fields (the test linked below includes both serialization and deserialization via Protobuf):

https://github.com/MovingBlocks/Terasology/blob/c9fd8934282be2d0bf57968a6e1c9ecc8cffc5b0/engine-tests/src/test/java/org/terasology/persistence/typeHandling/TypeHandlingIntegrationTest.java#L87-L111

For context, the class being serialized in the above test is this (where `Animal` is an abstract class with two subclasses):

```java
    private static class SomeClass<T> {
        private T data;
        private List<T> list = Lists.newArrayList();
        private Set<Animal> animals = Sets.newHashSet();
    }
```

# Changes
- Remove `DeserializationContext`
- Add factories for type handlers (`Rect2f`, `Rect2i`, `Region3i`) previously relying on `DeserializationContext.deserializeAs` which look up the type handlers they need
- Remove `SerializationContext.create(T, Class<? extends T>)`
- Rename `SerializationContext` to `PersistedDataSerializer`
- Rename `PersistedDataSerializer.createXYZ` methods to `serializeXYZ`
- Add `ArrayTypeHandler` (+ Factory)

# TODO
- [x] Add null handling in every TypeHandler -- null/default value handling takes place in `ObjectFieldMapTypeHandler` but still look for a better way to easily handle null values without having to `if(value == null) { return serializer.serializeNull(); }` in every single type handler
- [ ] Further simplify external Gson and Protobuf serialization workflow 
- [x] ~Load types from name via a `ClassLoader`(s) in `RuntimeDelegatingTypeHandler`to stay within the sandbox~ To be handled in a later PR
- [ ] Rebase onto `develop` after #3453 is merged

## Broken modules
- [ ] Dialogs (used by L&S)
- [ ] DynamicCities
- [ ] LightAndShadow - GSOC sensitive
- [ ] MasterOfOreon - GSOC sensitive
- [ ] Tasks (used by L&S)
- [ ] TutorialDynamicCities

PR #3453, opened 2018-07-26
Additional TypeHandlers

eviltak:
This PR adds a number of additional TypeHandlers and TypeHandlerFactories to the engine for added functionality.

`TypeHandler`/`TypeHandlerFactory` | Description
--- | ---
`ObjectFieldMapTypeHandler` (and Factory) | Serializes objects as a fieldName -> fieldValue map. Used as the last resort. Replaces `MappedContainerTypeHandler` and makes `@MappedContainer` redundant.
`AssetTypeHandlerFactory` | Generates an `AssetTypeHandler` for each `T extends Asset` so that they do not need to be manually registered.
`RuntimeDelegatingTypeHandler` | Delegates serialization of a value to its runtime type if needed. Used in cases where a subclass instance is referred to as a supertype. Meant for internal use only, and is used in `Collection`, `Map`, and object as a field map serialization.

Minor changes:
- Add `ReflectionUtil.resolveType`
- Resolve types in `ReflectionUtil.getTypeParameterForSuper`
- Use `ReflectionUtil.getTypeParameterForSuper` in `Collection` and `StringMap` type handlers.

PR #3449, opened 2018-07-24
Collection serialization improvements

eviltak:
# Goal
This PR mainly aims to improve the way collections are serialized. Currently, each `TypeHandler` contains code for serialization of both individual type instances and collections of type instances. This leads to a lot of hidden code duplication, with different `(de)serializeCollection` implementations doing almost the same thing. This PR introduces `CollectionTypeHandler`, a type handler that (in conjunction with `CollectionTypeHandlerFactory`) works for any `Collection<E>`. A unified `CollectionTypeHandler` has the following implications:
- No need for separate `TypeHandler`s for `Set`, `Queue`, `List`, etc.
- `TypeHandler.serializeCollection` and `deserializeCollection` are eliminated, simplifying existing `TypeHandler`s
- Many collection-related methods in `SerializationContext` and `DeserializationContext` are made redundant, greatly simplifying their implementations
- Collections are serialized in one format without variation. Should users wish to serialize a specific collection in a different (and possibly more efficient) way, they can add a custom `TypeHandler`/`TypeHandlerFactory`

# Changes
- Added `InstanceCreator` -> Functional interface for specifying how types with no-args constructors are instantiated
- Added `ConstructorLibrary` -> A library of `ObjectConstructor`/`InstanceCreator` instances for creating objects of various types.
- Added `CollectionTypeHandler` -> The `TypeHandler` that performs the brunt of collection serialization
- Removed `TypeHandler.serializeCollection` and `deserializeCollection`
- Removed `SimpleTypeHandler`, which was made redundant by the above removal
- Removed `SerializationContext.create(Collection<T>, Class<T>)`
- Removed `DeserializationContext.deserializeCollection`

# Prerequisites
This PR includes commits from #3446 and requires that it be merged.

PR #3446, opened 2018-07-23
TypeHandlerFactory

eviltak:
# Motivation
Terasology uses the `TypeHandler` interface to describe custom serialization formats for various types. Unfortunately, type handlers for generic types like `List` and `Map` rely on the type handlers for their type parameters and hence must be made dynamically.`TypeSerializationLibrary.getHandlerFor` generates such type handlers on demand. While the `TypeSerializationLibrary` makes it easy to add type handlers for custom types (thus allowing modules to use their own types in serializable types like components/events), it is impossible to dynamically generate type handlers without modifying engine code (`getHandlerFor`). The introduction of `TypeHandlerFactory` aims to solve this problem.

PR #3489, opened 2018-09-22
Temporarily revert part of the serialization overhaul

Cervator:
This is a fun one. Have pictures!

@eviltak has been doing some great work trying to overhaul our serialization, type handling, etc. However, that is a tricky field and when one major round got merged it had some subtle crashes for specific modules that I missed. A second major round made it into a second PR which fixed some/all of the issues from the first one but introduced *other* subtle new issues. Both sets also required a few module compile fixes. Then RL started eating @eviltak's time so further fixes became hard to find time for.

In the interest of stabilizing `develop` so we can do a proper release in advance of project birthday celebrations next weekend + the regular scheduled play test the weekend after that I'm trying to cleanly revert the first major round of changes then immediately re-introduce them in a parking branch where round two has then been merged on top. Git makes this easy even if it ends up looking funny:

image

To summarize that differently through the magic of bullet points:

* PR #3449 contained one major round of changes, and itself ended up containing #3446 - got merged
* PR #3456 contained a second major round of changes, and ended up containing #3453 - stalled
* This PR is a revert of #3449 meaning round one will disappear from develop
* Branch https://github.com/MovingBlocks/Terasology/tree/newSerialization then has that revert cleanly reversed and round two merged in on top - this would eventually be PRed when ready

Comparing merge commits with PRs seem to indicate the numbers look right:

2018-09-22 17_12_20-collection serialization improvements by eviltak pull request 3449 movingbl

That's PR #3449. Those numbers should match this PR in file count and they do. So the right stuff from #3449 is getting reverted

2018-09-22 17_11_49-comparing develop newserialization movingblocks_terasology

That's a would-be PR created from the `newSerialization` branch to `develop` and it should match what's in PR #3456 without any indication that the revert mattered at all, since it was reversed in that branch. And it does! Other than the 3 merge/revert commits in the picture up top, anyway. The file count matches.

Less certain what'll happen after this PR is actually merged. Will a PR from `newSerialization` *still* cleanly just list the contents of both rounds of changes, or will conflicts / other weirdness arise? I dunno! That's the beauty of Git, I guess. Fingers crossed 🤞 

Beyond this PR there is a small set of module PRs that went with round one, those will also be undone and likely get their own `newSerialization` branches. Roughly the same modules are affected by round two, but in both cases there were unrelated modules that would fail at runtime but compile fine. That'll be prepared soon.

Aiming to self-merge this when ready.


GooeyHub:

PR #3495, opened 2018-09-26
Remove liquid data

4Denthusiast:
Remove liquid data from chunks, and the classes for its encoding.

The existing systems for liquid data clearly relate to a feature that was never implemented, and have done so for years. It introduces additional complexity and a small runtime cost (the memory used) for no current benefit.

The Fluid module makes use of the removed LiquidData class, but I think that that was inappropriate in the first place, as it doesn't allow new fluids to be added by modules, requires other systems to keep fluid data aligned with block data, and is tied up with the mostly unimplemented liquids-on-top-of-blocks thing.

There seems to be some sort of error relating to loading existing worlds, but I don't think world backwards-compatibility is much of an issue at this point, so I ignored it.

There seems to be the issue again of me using a slightly different version of the protobuf compiler, leading to loads of non-functional changes in the affected file.

### Contains

Removes [get/set]Liquid from ChunkImpl, WorldProvider, ect. Removes LiquidData and its associated tests.

- [ ] Fix the Fluid module.
- [x] Check for other modules this breaks.
- [x] Decide whether this change is actually desirable.


GooeyHub:

Cervator:
Quite a fair PR, desirable, and long overdue as indeed the original usage of the system disappeared _years_ ago - but as a fun historical fact it *did* exist and work - mostly :-) I can still remember massive floods from unlucky seed values resulting in liquid source blocks being placed high up.

Affected modules:

- [ ] Fluids module as indicated above - needs fix
- [ ] FlexiblePathfinding - simple stub implementations to remove
- [ ] TutorialWorldGeneration - one class did set the liquids but as we know nothing does anything with it
- [ ] ShatteredPlanes - same
- [ ] SubstanceMatters - has an apparent usage of _typing_ in `FluidSubstanceCommonSystem` - new way to just declare substance types? We've in the past talked about a simple typing/tagging system ("This block is called Water, it is tagged with: liquid, fluid, wet, water...")
- [ ] Machines - uses typing, checks depth
- [ ] CombatSystem - just a simple usage for world gen
- [ ] DynamicCities - world gen usage ... and a typing for resources? Placing cities near water?
- [ ] AnotherWorld - world gen usage and maybe an attempt as something layering with water?
- [ ] WoodAndStone - uses that thing from AnotherWorld
- [ ] NeoTTA - same since it was cloned from WoodAndStone
- [ ] Lost - also uses the layering thing

Most of them seem trivial / could be gently removed. Main question may be if we want to retain some sort of liquid typing in the engine? Or possibly move that into a module. leaving liquids undifferentiated otherwise

4Denthusiast:
Having some list of liquid types could be useful, but I don't think that that deserves to be in the engine. In most cases, the set of blocks which are marked as liquid should be sufficient (and avoids possible problems of synchronising that with an additional list). In the case that additional fluid types are used in something like Fluid, and can't be placed in the world, that's something that has to be supported by that module anyway, and I think Fluid already has some sort of liquid registry.

jellysnake:
@e-aakash
Thanks for finishing up the PRs.
I'll do a xheck

GooeyHub:

PR #3496, opened 2018-09-30
Add memory for confirmed trusted URLs to let the game open in a browser

### LINKED ISSUE #3414: Add memory for confirmed "safe" URLs to let the game open in a browser
With #3405 we have a UI widget `UIButtonWebBrowser` that when clicked can open a URL in a browser. It requires confirmation from the user before doing so to avoid malicious modules sending a user off to some dubious site unknowingly.

At present that confirmation will be required every time. The purpose of this issue is to add a list of remembered safe sites so the user can OK a given URL and not be prompted again. Probably that should be added by including a checkbox (unchecked by default) on the confirmation dialog. If it is checked then on confirming the request add that URL to some new section of `config.cfg`

Any confirmation dialog then should go check that section first before popping up.

For similar functionality see `EnterUsernamePopup` which is used by `JoinGameScreen` when a user previously hasn't configured a player name. This should work similarly, although the config value will be a list rather than a single String.

On deleting `config.cfg` the list will be wiped and the user should be re-prompted again.

llvieira:
Closes #3414.

image

### Contains
1. Add a new checkbox on the confirmation dialog.

### How to test
1. Add the button widget to some existing screen and set up a link. For example, you can change the `updateOpenInBrowserButton()` in `ModuleDetailsScreen.java` to the folowing:
```java
private void updateOpenInBrowserButton() {
        openInBrowser.setEnabled(false);
        openInBrowser.setLink("https://github.com/llvieira")
            .setNuiManager(getManager())
            .setTranslationSystem(translationSystem)
}
``` 
2. Go the the screen where you enabled the open-in-browser functionality.
2. Mark the checkbox to save the link. All the trusted URLs will be stored in `config.cfg`.
```shell
"webBrowserConfig": {
    "trustedUrls": [
      "https://github.com/llvieira"
    ]
  }
```
3. Close the game, open it again, go to the same screen.
4. Check that the default web browser open the link without asking for permission.
5. Try the same without making the checkbox. The popup will always appear.

### TODO - WIP part
1. Just save the root domain instead of the entirely URL. **[GCI task]**

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/ConfirmUrlPopup.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/ConfirmUrlPopup.java
@@ -16,15 +16,16 @@
 package org.terasology.rendering.nui.layers.mainMenu;
 
 import org.terasology.assets.ResourceUrn;
+import org.terasology.config.WebBrowserConfig;
 import org.terasology.rendering.nui.CoreScreenLayer;
 import org.terasology.rendering.nui.WidgetUtil;
 import org.terasology.rendering.nui.widgets.UIButton;
 import org.terasology.rendering.nui.widgets.UICheckbox;
 import org.terasology.rendering.nui.widgets.UILabel;
 
-public class ConfirmURLPopup extends CoreScreenLayer {
+public class ConfirmUrlPopup extends CoreScreenLayer {
 
-    public static final ResourceUrn ASSET_URI = new ResourceUrn("engine:confirmURLPopup!instance");
+    public static final ResourceUrn ASSET_URI = new ResourceUrn("engine:confirmUrlPopup!instance");
 
     private Runnable leftActon;
     private Runnable rightAction;
@@ -33,13 +34,6 @@ public class ConfirmURLPopup extends CoreScreenLayer {
     public void initialise() {
         WidgetUtil.trySubscribe(this, "leftButton", button -> buttonCallback(leftActon));
         WidgetUtil.trySubscribe(this, "rightButton", button -> buttonCallback(rightAction));
-
-        UICheckbox saveURL = find("saveURL", UICheckbox.class);
-        saveURL.setChecked(false);
-        saveURL.subscribe(e -> {
-            // save url in config.cfg
-            System.out.println("Save URL checkbox");
-        });
     }
 
     private void buttonCallback(Runnable action) {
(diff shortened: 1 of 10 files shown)


engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/moduleDetailsScreen/ModuleDetailsScreen.java
```diff
@@ -199,18 +199,23 @@ public String getString(Module value) {
...
-            return;
-        }
-        final String moduleOrigin = getOriginModuleUrl(moduleManager.getRegistry().getLatestModuleVersion(module.getMetadata().getId()));
-        if (StringUtils.isNotBlank(moduleOrigin)) {
-            openInBrowser.setEnabled(true);
-            openInBrowser.setLink(moduleOrigin)
-                    .setNuiManager(getManager())
-                    .setTranslationSystem(translationSystem);
-        }
+        openInBrowser.setEnabled(true);
```
llvieira: It is just for testing... :)

GooeyHub:

GooeyHub:

Cervator:
Just for the issue's reference: the WIP part here is since the issue asks for the _base_ of the address (the domain part) to be remembered, but the PR as-is stores specific links, which may be better/safer anyway.

There's also a bit of code styling/testing code to clean up then we can merge and consider doing the root domain option separately or even as a GCI task :-)

llvieira:
I made the changes, @Cervator. Could you review it again please? :smiley:

GooeyHub:

llvieira:
I made the changes, @Cervator. Could you please review it again?

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/ConfirmUrlPopup.java
```diff
@@ -0,0 +1,73 @@
...
+        WidgetUtil.trySubscribe(this, "rightButton", button -> buttonCallback(rightAction));
+    }
+
+    private void buttonCallback(Runnable action) {
+        getManager().popScreen();
+        action.run();
+    }
+
+    public void setLeftButton(String text, Runnable action) {
+        find("leftButton", UIButton.class).setText(text);
```
Cervator: Could use some null checks here and several other places, in case something goes wrong with finding the UI element

llvieira: Sure, I did it now.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Tests out - works beautifully! Well done and thanks :-)

PR #3498, opened 2018-10-04
hotfix that fixes the rotation issue with structured templates

pollend:
this is a quick hot-fix that fixes the block building in structured templates. A freeformfamily is now considered a side-defined family. this can be verified by building a structured template shape with stairs and verifying that the stairs are in the right direction.


GooeyHub:

Cervator:
Tested out in GooeysQuests and MedievalCities - looking good! Thank you so much :-)

PR #3511, opened 2018-10-21
Allow modules limited API access to files

### LINKED ISSUE #3366: Allow modules limited API access to files
## Overview

_edit: @asiekierka is working on this as of August 2018_

This is a more narrow approach than something like #1703, specifically focused on giving modules limited access to files, still without allowing modules access to  `java.io.File` etc. Feedback appreciated on the concept, consider this an initial proposal. This would be in high demand for some projects like @asiekierka's GSOC item, and maybe a good way to satisfy _some_ of the needs of a more full-fledged permission system for opening up just about anything.

There are two approaches involved:

1. Create a subdir in the game data dir that allows modules to read, write, and delete to actual files, via `@API` listed engine methods doing the actual file manipulation
1. Allow modules `@API` access to an engine method that opens a file chooser for the user to manually select a given file for reading (and/or writing?)

### "Safe" file space

A minimal implementation here would define a new subdir in the game data dir (see `PathManager`) such as `<user home>/Terasology/sandbox` and make engine methods that wrap common file operations so they're only allowed to happen within that directory. Example:

Module class `X` calls engine method `public byte[] readSandboxFile(String filename)` which reads the `filename` at the root of the `sandbox` dir then provides the contents in a byte array to the calling module to do with as it pleases.

Now I doubt providing a byte array is necessarily the best way to provide the contents, but then I'm not paid to be on top of the best practices surrounding I/O these days, in fact I'm not paid at all! :-)

I don't know if it makes sense to provide a stream of some sort instead, a bunch of related things were put on the whitelist in the past by https://github.com/MovingBlocks/Terasology/pull/2494/files - there are also other fun things like virtual file systems that might come into play.

A more improved implementation might include some ideas like:

* Use `sandbox/global` for the space that any module has access to. Then allow exclusive access by modules to paths like `sandbox/modules/<module name>` allowing modules to have private spaces other modules cannot read. I'm not sure if there are glaring ways to get around this since modules can access each other just fine at present.
* Allow modules themselves to expose access to their private store for other modules. Maybe some simple little whitelist file like `sandbox/modules/MyModule/whitelist.txt` with a list of approved module names? Not sure this even has a good use case.

### File chooser request

Much like the above, but rather than exclusively target a safe sandboxed file space request a user manually select any file on the file system via traditional file chooser dialog. Again solely provide the data for a module to do something with, relying on the engine itself to do the I/O proper.

Maybe make it so the engine pops a disclaimer before showing the file chooser dialog, warning the user that modules can do anything they want with the related data.

Am unsure if this should be read-only, or if there also should be an option to write (or even overwrite?) a target file. That would be a very nice way to save a StructureTemplate exported prefab directly to a target file of the user's choice.

Key point here is that only the engine is able to engage the user with the file chooser dialog, requiring manual interaction with the user. No direct file access allowed at all.

## Concerns

This seems like a pretty limited risk, but I am sure there still are some. A module could excessively write to files in the file sandbox, causing I/O slowdowns or eating up a bunch of space. Maybe there should be a limit to how much I/O activity a module can request and/or a limit set on how much space a module can use. Probably easy enough to handle in a module's private dir, maybe the global space can have a higher but still finite cap.

Also: Unit tests please! Maybe a good thing to add to the [Malicious](https://github.com/Terasology/Malicious) module.

Pinging @engiValk for some nefarious speculation.

llvieira:
### Contains

I'm following the "safe" file space approach to solve this issue (closes #3366). I've changed the `PathManager` to create a new sub-directory ( `<user home>/Terasology/sandbox`) in the game data directory. I've also created the `SandboxFileManager` which wraps common file operations so they're only allowed to happen within that sub-directory. The `SandboxFileManager` is marked with the `@API` annotation so that it could be accessed by any module.

#### Writing files
1. Using any module like `BuilderSampleGameplay`, you could implement `postSave()` method in the system class, like this:
```java
@Override
public void postBegin() {
    SandboxFileManager sandboxFileManager = new SandboxFileManager();
    byte[] someBytes = new byte[2];
    someBytes[0] = 'h';
    someBytes[1] = 'i';

    Consumer<OutputStream> consumer = outputStream -> {
        try {
            outputStream.write(someBytes);
        } catch (IOException e) {
            logger.error("error", e);
        }
    };

    sandboxFileManager.writeFile("SomeFile", consumer);
}
``` 
2. Then after running the game with _Builder Sample Gameplay_, you can check if `SomeFile` was created into `sandbox` directory.

#### Reading files
1. Implement `postSave()` method, like the following:
```java
@Override
public void postBegin() {
    SandboxFileManager sandboxFileManager = new SandboxFileManager();

    Consumer<InputStream> consumer = inputStream -> {
        try {
           int value = inputStream.read();
           
           while (value != -1) {
               doSomething(value);
               value = inputStream.read();
            }
        } catch (IOException e) {
            logger.error("error", e);
        }
    };

    sandboxFileManager.readFile("SomeFile", consumer);
}
``` 

_Edit_: The **access denied** issue that I had described when I opened this pull request was solved by wrapping the outright IO stuff with `AccessController.doPrivileged(...)`.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ModuleInputStream.java
+++ b/engine/src/main/java/org/terasology/engine/module/ModuleInputStream.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2018 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.engine.module;
+
+import org.terasology.module.sandbox.API;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This class is a wrapper for {@link InputStream} which should be used by modules.
+ * <p>
+ * By using this class, the caller is not allowed to close the stream at all.
+ * For security reasons, only {@link SandboxFileManager} have the permissions to close it,
+ * since the caller may completely forget to do so.
+ */
+@API
+public class ModuleInputStream {
+
+    private InputStream inputStream;
+
+    public ModuleInputStream(InputStream inputStream) {
+        this.inputStream = inputStream;
+    }
+
+    public int read() throws IOException {
+        return inputStream.read();
+    }
+
+    public int read(byte[] b) throws IOException {
+        return inputStream.read(b);
+    }
+
+    public int read(byte[] b, int off, int len) throws IOException {
+        return inputStream.read(b, off, len);
+    }
+
+    public long skip(long n) throws IOException {
+        return inputStream.skip(n);
+    }
+
+    public int available() throws IOException {
+        return inputStream.available();
+    }
+
+    public void mark(int readLimit) {
+        inputStream.mark(readLimit);
+    }
+
+    public void reset() throws IOException {
+        inputStream.reset();
+    }
+
+    public boolean markSupported() {
+        return inputStream.markSupported();
+    }
+}
--- a/engine/src/main/java/org/terasology/engine/module/ModuleInputStream.java
+++ b/engine/src/main/java/org/terasology/engine/module/ModuleInputStream.java
@@ -42,8 +42,43 @@ public class ModuleInputStream extends InputStream {
     }
 
     @Override
+    public int read(byte[] b) throws IOException {
+        return inputStream.read(b);
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+        return inputStream.read(b, off, len);
+    }
+
+    @Override
+    public long skip(long n) throws IOException {
+        return inputStream.skip(n);
+    }
+
+    @Override
+    public int available() throws IOException {
+        return inputStream.available();
+    }
+
+    @Override
     public void close() throws IOException {
         throw new IOException("You must not close the stream. " +
                 "The SandboxFileManager will do it automatically.");
+    }
+
+    @Override
+    public synchronized void mark(int readLimit) {
+        inputStream.mark(readLimit);
+    }
+
+    @Override
+    public synchronized void reset() throws IOException {
+        inputStream.reset();
+    }
+
+    @Override
+    public boolean markSupported() {
+        return inputStream.markSupported();
     }
 }
--- a/engine/src/main/java/org/terasology/engine/module/SandboxFileManager.java
+++ b/engine/src/main/java/org/terasology/engine/module/SandboxFileManager.java
@@ -20,12 +20,14 @@ import org.slf4j.LoggerFactory;
 import org.terasology.engine.paths.PathManager;
 import org.terasology.module.sandbox.API;
 
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.stream.Stream;
+import java.util.function.Consumer;
 
 /**
  * This class wrap common file operations so they're only allowed to happen
@@ -45,17 +47,23 @@ public class SandboxFileManager {
     }
 
     /**
-     * Read the file that matches the passed filename.
+     * Reads the file that matches the passed filename.
      *
      * @param filename
-     * @return String stream.
+     * @param consumer Consumer to read the file.
      */
-    public Stream<String> readFile(String filename) {
-        Path sandboxPath = pathManager.getSandboxPath(filename);
+    public void readFile(String filename, Consumer<ModuleInputStream> consumer) {
+        String sandboxPath = pathManager.getSandboxPath(filename).toString();
 
-        return AccessController.doPrivileged((PrivilegedAction<Stream<String>>) () -> {
+        AccessController.doPrivileged((PrivilegedAction<Void>) () -> {
             try {
-                return Files.lines(sandboxPath);
+                InputStream inputStream = new FileInputStream(sandboxPath);
+                ModuleInputStream moduleInputStream = new ModuleInputStream(inputStream);
+
+                // consumer to read the file, if it exists
+                consumer.accept(moduleInputStream);
+
+                inputStream.close();
             } catch (IOException e) {
                 logger.error("Could not read the file: " + filename, e);
             }
--- a/engine/src/main/java/org/terasology/engine/module/SandboxFileManager.java
+++ b/engine/src/main/java/org/terasology/engine/module/SandboxFileManager.java
@@ -32,7 +32,7 @@ import java.util.function.Consumer;
 /**
  * This class wrap common file operations so they're only allowed to happen
  * within Terasology/sandbox directory.
- *
+ * <p>
  * It gives modules the ability to read and write to the file system in a safe way.
  */
 @API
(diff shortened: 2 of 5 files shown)


GooeyHub:

engine/src/main/java/org/terasology/engine/module/SandboxManager.java
```diff
@@ -0,0 +1,66 @@
...
+import org.slf4j.LoggerFactory;
+import org.terasology.engine.paths.PathManager;
+import org.terasology.module.sandbox.API;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+@API
+public class SandboxManager {
```
Cervator: Maybe `SandboxFileManager` just for added clarity?

engine/src/main/java/org/terasology/engine/paths/PathManager.java
```diff
@@ -266,6 +268,14 @@ public Path getConfigsPath() {
         return configsPath;
     }
 
+    /**
+     *
+     * @return Path in which the modules are allowed to save files.
+     */
+    public Path getSandboxesPath() {
```
Cervator: IMHO `getSandboxPath` not sandboxES

engine/src/main/java/org/terasology/engine/module/SandboxManager.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(SandboxManager.class);
+
+    private PathManager pathManager;
```
Cervator: Maybe you've tried this and/or this class might not be in a place where `@In` would work, but that'd be a nice easy way to initialize this var. Maybe something to try :-)

engine/src/main/java/org/terasology/engine/module/SandboxManager.java
Cervator: Also: Javadoc is great, both at class level and method level. No need to wait to finish the whole thing before adding some nice docs :-)

engine/src/main/java/org/terasology/engine/module/SandboxManager.java
llvieira: @Cervator I tried this but I got a `NullPointerException`. Is it added to the `CoreRegistry`? At least I didn't find any occurrences of something like `@In private PathManager pathManager`.

engine/src/main/java/org/terasology/engine/paths/PathManager.java
llvieira: Fixed!

GooeyHub:

engine/src/main/java/org/terasology/engine/module/SandboxFileManager.java
```diff
@@ -0,0 +1,104 @@
...
+        this.pathManager = PathManager.getInstance();
+    }
+
+    /**
+     * Read the file that matches the passed filename.
+     *
+     * @param filename
+     * @return An array of bytes.
+     */
+    public byte[] readFile(String filename) {
```
asiekierka: Can we not offer a Stream of some kind? Reading to a byte[] can go awry on memory usage given very large files.

llvieira: Yes, we can. I managed to return a `Stream<String>` instead of `byte[]`. It's now really efficiently for large files.

GooeyHub:

asiekierka:
@llvieira try-with-resource with a callback, perhaps? That way, once the callback returns, you can close it.

llvieira:
@asiekierka what would have inside the callback? If it just return an `InputStream` and it's closed soon after, then I think we would have the same problem: we cannot read from a closed `InputStream`. I've tried to figure out solutions like that and I didn't found anything useful yet. Perhaps, the idea is just not yet clear in my mind. :-)

asiekierka:
@llvieira No, you perform all your reads *in* the callback. If you want to read the whole file, you can just do that; if you're more specific, you can do that as well. If you fear other code using close() in place of the sandbox API method, just wrap the InputStream so that it's not possible.

It's just a suggestion, though. Do you have a better idea?

llvieira:
@asiekierka as far as I know if we wrap the `InputStream` to avoid the caller not closing the stream, we'll need to close it inside the wrapper `read` methods, as follows:
```java
public int read() throws IOException {
    int value = 0;
    try {
        value = inputStream.read();
    } finally {
        if (value == -1) {
            inputStream.close();
        }
    }

    return value;
}
```

However, if somehow the caller wants to reposition the stream after the read operation has been completed, it would be impossible using this wrapper. So, given these limitations I'm not sure if it would be a good idea. Perhaps you are talking about a different approach for the wrapper that I don't gather yet. 

Regarding the callback, I'm still in doubt. Would `readFile(String filename)` have to receive the `InputStream` interface and within the method to call some read function from the interface? moreover, would the caller pass in some particular `InputStream` implementation  to `readFile` method, like `FileInputStream`? My concept of Java callbacks is: you create an interface which has one or more methods to be called when some event happens. Then, any class which needs to be notified when events occur implements this interface. So the idea of `readFile` having a callback is still not clear to me. Could you please give an example of how the callback could be used in this scenario? 

Sorry if I'm asking too much, but I have to have this design clear in my mind to start changing the code. :-)

asiekierka:
@llvieira 

What i was thinking of is:

1. Have the callback function NOT run close(). At all.
2. When the callback function finishes operation and returns to .readFile(), .readFile() simply calls .close() on the InputStream.
3. The InputStream received by the callback functions throws an Exception when the callback function tries to .close() it, as that responsibility falls on the .readFile() method in the API after the callback function finishes.

Is this more clear?

llvieira:
@asiekierka yep, I got it now. Could you please check my last commit to see if I made the changes accordingly? Really appreciate your help. :smiley:

GooeyHub:

engine/src/main/java/org/terasology/engine/module/ModuleInputStream.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * This class is a wrapper for {@link InputStream} which should be used by modules.
+ * <p>
+ * By using this class, the caller is not allowed to close the stream at all.
+ * For security reasons, only {@link SandboxFileManager} have the permissions to close it,
+ * since the caller may completely forget to do so.
+ */
+@API
+public class ModuleInputStream {
```
asiekierka: If I were you, I would make this extend InputStream, put @Override on the proxied methods, and add an exception throw on close(). This would allow usage of the modified InputStream in other libraries which already expect an InputStream-type object.

asiekierka:
Yes, this is way better! I do have one bit of input, as shown in the above comment.

Writing to files could be realized in the same fashion, though with some provisions for not exceeding a given amount of space, say, it might be a bit more complicated and is definitely out of this PR's scope.

jellysnake:
This PR seems to be tantalisingly close to mergable.
@asiekierka what's your thoughts on llvieira's comment?

engine/src/main/java/org/terasology/engine/module/ModuleInputStream.java
llvieira: @asiekierka sorry, I had misunderstood you, now I see your point. I'm gonna update it. :smiley:

GooeyHub:

GooeyHub:

Cervator:
This pull request **introduces 1 alert** when merging 52cd46ccfac69cc2e6525c73e625aeeee7b45e8f into b02e53d71ea3f5ef2000405f631481fb74f6ede2 - [view on LGTM.com](https://lgtm.com/projects/g/MovingBlocks/Terasology/rev/pr-97be0615c2f01ca95653dcd5cf826bd636ce0d72)

**new alerts:**

* 1 for Inefficient output stream

---

asiekierka:
This is exactly what I had envisioned - as such, I'm okay with it.

GooeyHub:

llvieira:
@asiekierka thanks for all your help, after reading our comments again, it became much clearer how I would have to implement it.

jellysnake:
And so, with the check off from Asie I'm going to merge this long awaited PR :)

Issue #3510, opened 2018-10-21
Fix up LGTM errors


jellysnake:
So we've recently just enabled [LGTM](https://lgtm.com/projects/g/MovingBlocks/Terasology) to this project and there is a number of alerts coming popping up.  
We should probably try and fix most of these, barring false positives of course.

I've tried to group them into different types and detail my thoughts on them.

## Errors
| Error Type | LGTM rule | Details | Fixed? |
| --- | --- | --- | --- |
| [Command line is built with string concatenation.](https://lgtm.com/projects/g/MovingBlocks/Terasology/alerts/?mode=tree&severity=error&ruleFocus=7870097) | [rule](https://lgtm.com/rules/7870097/) | Seems to be an easy fix, and a worthwhile security flaw to patch. Just need to follow details in the rules | ☑️ #3512 |
| [Array index out of bounds](https://lgtm.com/projects/g/MovingBlocks/Terasology/alerts/?mode=tree&severity=error&ruleFocus=2049320662) | [rule](https://lgtm.com/rules/2049320662/) | Easily fixed by adding a length check to the line `if (c == '&') {` just before the errors | ☑️ #3514 |
| [Container contents are never accessed](https://lgtm.com/projects/g/MovingBlocks/Terasology/alerts/?mode=tree&severity=error&ruleFocus=890083) | [rule](https://lgtm.com/rules/890083/) | IntelliJ seems to agree with this. I'd recommend removing and with thorough tests to ensure there are no side effects or that it isn't accessed in an odd manner elsewhere | ☑️ #3512 |
| [Container contents are never initialized](https://lgtm.com/projects/g/MovingBlocks/Terasology/alerts/?mode=tree&severity=error&ruleFocus=5980077) | [rule](https://lgtm.com/rules/5980077/) | Again, like above I think remove and test. | ☑️ #3512 |
| [Inconsistent equals and hashCode](https://lgtm.com/projects/g/MovingBlocks/Terasology/alerts/?mode=tree&severity=error&ruleFocus=6770060) | [rule](https://lgtm.com/rules/6770060/) | I think this falls under the label of 'over zealous checking'. As far as I can see this class is never used in such a way that hashcode is needed. Don't think we should fix this, as it's just not worth it | ☑️ #3518 |
[...]

PR #3514, opened 2018-10-21
Fixes the LGTM array out of bounds error

mwvdev:
### Contains

Fixes the LGTM array out of bounds error in #3510. When starting to work on this issue, I noticed that the method didn't work as intended so this has been fixed as part of this issue.

### How to test

Run the unit tests added as part of this PR. Verify that LGTM no longer flags this method as having errors (I was unable to test this locally since LGTM doesn't seem to support analyzing forks).

### Comments

I've tried keeping the fix as close to the original code as possible. In order to maintain the behavior of throwing an exception if an unsupported entity is encountered, I chose to add the length guard at each individual entity, since adding the check near `if (c == '&') {` would change this behavior. It would also need to check entities of length 3 and 4.

### DIFF
--- a/engine-tests/src/test/java/org/terasology/rendering/nui/widgets/browser/data/basic/HTMLLikeParserParameterizedTest.java
+++ b/engine-tests/src/test/java/org/terasology/rendering/nui/widgets/browser/data/basic/HTMLLikeParserParameterizedTest.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2018 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


GooeyHub:

Cervator:
This pull request **fixes 10 alerts** when merging 6f37d9d95d8489f0e9e1820af5eb1a22486bb8f1 into 9118708ad3f34f0bc05572fe54ad47538c3164bb - [view on LGTM.com](https://lgtm.com/projects/g/MovingBlocks/Terasology/rev/pr-9412d3ef616930737bdbe428ba078ae0a475f7b9)

**fixed alerts:**

* 10 for Array index out of bounds

---

jellysnake:
LGTM fixes and new tests? Even better. I'll merge this

PR #3518, opened 2018-10-22
Fixes the LGTM error "Inconsistent equals and hashCode"

mwvdev:
### Contains

Fixes the LGTM hashCode error in #3510.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFbo.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/stateChanges/BindFbo.java
@@ -81,11 +81,14 @@ public final class BindFbo implements StateChange {
             return this;
         }
 
-        // TODO: Add .hashCode()
-
         @Override
         public boolean equals(Object obj) {
             return (obj instanceof UnbindFbo);
+        }
+
+        @Override
+        public int hashCode() {
+            return UnbindFbo.class.hashCode();
         }
 
         @Override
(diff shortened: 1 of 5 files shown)


GooeyHub:

Cervator:
This pull request **fixes 6 alerts** when merging 7d8ea300dfe4c919d62217782d936513e34b70e8 into 9118708ad3f34f0bc05572fe54ad47538c3164bb - [view on LGTM.com](https://lgtm.com/projects/g/MovingBlocks/Terasology/rev/pr-df250ecc46a67a8afd721f3c0ba1e0431848d832)

**fixed alerts:**

* 6 for Inconsistent equals and hashCode

---

Cervator:
Thanks for the fixes :-)

I'm not really very knowledgeable about `hashCode` generation - usual example code seems to include some random numbers and the hash codes for any fields added in. Any particular reason or tactic behind the choice you made @mwvdev ?

Poking @vampcat or anybody else who might have some additional feedback

@GooeyHub: add to whitelist

GooeyHub:

jellysnake:
So from what I gather this hashcode implementation is calling the hashcode on the class object, rather than the instance.
This is consistent with the current equals. At the moment equals will be true if the two objects are both of type `UnbindFbo`. Hence the logical extension is that all instances of `UnbindFbo` should hash to the same value.

This is done by using the hash function for the _class_ object rather than an instance.

mwvdev:
@jellysnake Exactly. I noticed that the objects were stateless and the only "identifying" part of the object was its class. For stateless objects I think this is a good way to do it.

PR #3522, opened 2018-10-24
#3508 avoid zero and negative available size in ColumnLayout

### LINKED ISSUE #3508: Input settings screen does not handle some window sizes gracefully

Look at the input settings menu to gaze at its sheer awesomeness

### What actually happened

It looked ... less awesome than expected, as it seems like the margins are messed up somehow when the overall window size is of a certain size or less

2018-10-20 22_51_41-terasology _ alpha

When I resized the window and made it larger suddenly the buttons popped into shape properly

2018-10-20 22_50_56-terasology _ alpha

### How to reproduce

* Run the game at default settings (no prior `config.cfg` present locally)
* Go to Settings - Input - behold the monstrous proportions of things
* Cry in despair
* Try to increase the overall game window a bit
* Bask at the awesomely structured working version of the page

### Extra

See the https://github.com/Terasology/TutorialNui for an introduction to our UI framework. This issue has happened occasionally in the past, maybe when new elements have been added that have somehow thrown off the available dimensions / margin settings on that page? It could possible relate to the presence of specific modules that introduce input setting items (the list is built dynamically, part of what causes the slowdown in trying to load the page first time). 

Overly wide descriptions of each input setting? Yet it may happen primarily on shrinking the window _vertically_ ? We could probably eliminate the "Terasology" logo element on that page to help make more vertical space available. But still, the buttons shouldn't go screwy just because the window gets resized a bit. There must be a better way to gracefully degrade

JustPRV:
### Contains

Fixes #3508.
It turned out when availableSize.y (org/terasology/rendering/nui/layouts/ColumnLayout.java:232) become negative or zero it causes incorrect row height calculation as Canvas size become (0, 0) and preferable Label height becomes significantly higher as it expects Label text to be split to as many lines as possible for 0 widths. 
Fix looks a bit hacky and I believe there is a better solution, but I hope this PR will at least help in a further investigation.

### How to test

See test case from #3508.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layouts/ColumnLayout.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layouts/ColumnLayout.java
@@ -224,10 +224,10 @@ public class ColumnLayout extends CoreLayout<LayoutHint> {
         Vector2i availableSize = new Vector2i(areaHint);
         int numRows = TeraMath.ceilToInt((float) widgetList.size() / columns);
         if (numRows > 0) {
-            availableSize.y -= verticalSpacing * (numRows - 1);
+            availableSize.y = Math.max(1, availableSize.y - verticalSpacing * (numRows - 1));
         }
         if (columns > 0) {
-            availableSize.x -= horizontalSpacing * (columns - 1);
+            availableSize.x = Math.max(1, availableSize.x - horizontalSpacing * (columns - 1));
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Thanks for the fix! Helps bunches. Added you to the credits 👍

PR #3524, opened 2018-10-24
Allow to change resolutions used in full-screen mode

### LINKED ISSUE #3505: Add resolution picker
It has been 7 years and we have somehow managed so far without adding an actual option for picking the resolution to run the game at! Weird and recent realization. Destination Sol does have a resolution picker so it might have some related code for inspiration - although it uses LibGDX so maybe not.

For Terasology usually the game just runs in windowed mode, but it is a plenty valid use case to switch the resolution used in full screen.

To complete: Add a new option under the video settings to pick from available resolutions in a drop-down. Apply the updated resolution via some sort of OpenGL/LWJGL magic (surely there'll be some handy howto examples on Google)

Edvinas01:
Adds resolution picker to video settings menu and resolves #3505. 

It can be found here:

The resolution picker works in the following way:
- By default the best matching resolution will be shown in the picker.
- After picking any resolution in windowed full-screen or windowed modes, it will only be persisted in `config.cfg` without actually changing the resolution.
- After picking any resolution in full-screen mode, game resolution will change to the selected one and will be persisted in `config.cfg` file.

When changing resolutions on Linux Mint 18 with dual-monitor setup, sometimes the game screen appears to be cut-off in the middle. This seems to happen based on which screen I start the game and occurs randomly.


jellysnake:
Tested this out and it does indeed work.
However it's also very easy to accidentally select the wrong entry like I did and you end up with a wholly unusable screen.

I think we should add some form of timer, where after the timer the resolution changes revert, unless confirmed. This way if you accidentally select the wrong one you aren't stuck

jellysnake:
@GooeyHub: Ok to test

GooeyHub:

Edvinas01:
@jellysnake good point, I'll play around with this idea.

Edvinas01:
Added a popup which will appear once resolution is changed in full-screen mode. If the timer runs out, resolution will be restored to the previous one:
</img>

GooeyHub:

engine/src/main/java/org/terasology/engine/subsystem/lwjgl/LwjglResolution.java
```diff
@@ -0,0 +1,52 @@
...
+        this.displayMode = displayMode;
+    }
+
+    public DisplayMode getDisplayMode() {
+        return displayMode;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
Cervator: Oh noes! Minor code style violation: always use brackets. Easy to fix tho :-)

PR #3527, opened 2018-10-25
Fixed the LGTM error "Type mismatch on container modification"

mwvdev:
### Contains

Fixes the "Type mismatch on container modification" from #3510.

### How to test

The input code was fairly tightly coupled so I had trouble unit testing this approach. As an alternative, I have [attached a snippet in this Gist](https://gist.github.com/mwvdev/a2c868c2679c345f220c1155213de289) to verify that this fix works. The snippet also shows that the LGTM error is unlikely to be a false positive.

### DIFF
--- a/engine/src/main/java/org/terasology/input/lwjgl/JInputControllerDevice.java
+++ b/engine/src/main/java/org/terasology/input/lwjgl/JInputControllerDevice.java
@@ -208,7 +208,7 @@ public class JInputControllerDevice implements ControllerDevice {
      * @param controller the controller to remove
      */
     private void removeController(Controller controller) {
-        controllers.remove(controller.getName());
+        controllers.remove(controller);
         logger.info("Removed controller: " + controller.getName());
     }
 


GooeyHub:

Cervator:
This pull request **fixes 1 alert** when merging c63076a29b6a80cf70a0ae7d70ba64ff7b2bb18b into 7cf495ee093ff7d3829498288660138d266761d8 - [view on LGTM.com](https://lgtm.com/projects/g/MovingBlocks/Terasology/rev/pr-95659df8a72b125063feb808a8fda2e19cf07424)

**fixed alerts:**

* 1 for Type mismatch on container modification

---

Cervator:
Thanks for the nice Gist! Any reason to not add that formally to our engine tests?

mwvdev:
Thanks! I considered it but chose to supply it as a gist instead because if the implementation in JInputControllerDevice changes then the attached test cases will likely keep passing, i.e. if the container was switched with a Controller aware container that knew to call getName, then the attached test case would no longer be relevant. Then again, it might be better than leaving it out. What do you think? Should we add it?

PR #3531, opened 2018-10-26
Added wildcard support to groovyw <item> get

BenjaminAmos:
# Description
This PR adds wildcard support to the `groovyw <item> get <object>` e.g. `groovyw <item> get *`.

# Testing
Testing this PR could take some considerable time and effort to check all of the use-cases are correct, so these changes should only apply if wildcards (`*` or `?`) are detected in the input argument (see [Line 61 of util.groovy](https://github.com/BenjaminAmos/Terasology/blob/715d2450d077613a3fc3c6e5ee4ca26f3a9e9b7c/config/groovy/util.groovy#L61)). As such, I will list the series of tests that I have tried so far:
- Command: `groovyw module get *`
    - Expected result: All the available modules under the @Terasology organisation should be obtained.
- Command: `groovyw module get ?????Blocks`
    - Expected result: The [FunnyBlocks](https://github.com/Terasology/FunnyBlocks) module should be obtained.
- Command: `groovyw module get *StructureTemplates*`
    - Expected result: Both the [StructureTemplates](https://github.com/Terasology/StructureTemplates) and [AdditionalStructureTemplates](https://github.com/Terasology/AdditionalStructureTemplates) modules should be obtained.
- TODO: Add some more test examples - See [my other pull request](https://github.com/MovingBlocks/DestinationSol/pull/342) for another (although only functional with DestinationSol) set of examples.

# Outstanding Work
- [X] Such a large change needs significant testing. This change needs to be tested extensively to be proven useable, although I think that it should work most of the time. I still need to finish doing this, as I have not fully tested it yet.
    - Further tests to perform:
        - [X] Test that the regex characters (``[``, ``]``, ``(``, ``)``, ``*``, ``.``, ``\``) are not injected into the code
        - [X] Test if multiple wildcard selectors can be used to obtain completely different modules, e.g. using the command `groovyw module get *de* Fo*`, which should obtain all the modules that either have the letter pair "de" in them or that start with "Fo".

# Notes
This should not break any existing functionality, as the new wildcard handling code is only used when the `*` or `?` symbols are found.

The wildcard strings will have to be escaped using single quotes (e.g. ``./groovyw module get '*'`` rather than ``./groovyw module get *``) on POSIX platforms (Linux and MacOS), as otherwise the terminal appears to automatically expand them (replacing the arguments with the names all the files in the current directory matching that descriptor) before passing them to the program. I have modified the Windows script to automatically escape the arguments, although I am not sure if it is possible with POSIX platforms to do so.

### DIFF
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -355,15 +355,15 @@ class common {
      * @param regex the regex that the retrieved items should match
      */
     String[] retrieveAvailableItemsWithRegexMatch(String regex) {
-        ArrayList<String> selectedModules = new ArrayList<String>()
-        String[] moduleList = retrieveAvailableItems()
-        for (String module : moduleList) {
-            if (module.matches(regex)) {
-                selectedModules.add(module)
+        ArrayList<String> selectedItems = new ArrayList<String>()
+        String[] itemList = retrieveAvailableItems()
+        for (String item : itemList) {
+            if (item.matches(regex)) {
+                selectedItems.add(item)
             }
         }
 
-        return ((String[]) selectedModules.toArray());
+        return ((String[]) selectedItems.toArray());
     }
 
     /**
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -31,9 +31,11 @@ class common {
     /** The default name of a git remote we might want to work on or keep handy */
     String defaultRemote = "origin"
 
+    /** Should we cache the list of remote items */
     boolean itemListCached = false
 
-    String[] cachedItemList;
+    /** For keeping a list of remote items that can be retrieved */
+    String[] cachedItemList
 
     /**
      * Initialize defaults to match the target item type
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -31,6 +31,10 @@ class common {
     /** The default name of a git remote we might want to work on or keep handy */
     String defaultRemote = "origin"
 
+    boolean itemListCached = false
+
+    String[] cachedItemList;
+
     /**
      * Initialize defaults to match the target item type
      * @param type the type to be initialized
@@ -297,6 +301,9 @@ class common {
      * @return a String[] containing the names of items available for download.
      */
     String[] retrieveAvailableItems() {
+        if (itemListCached) {
+            return cachedItemList
+        }
 
         // TODO: We need better ways to display the result especially when it contains a lot of items
         // However, in some cases heavy filtering could still mean that very few items will actually display ...
@@ -323,7 +330,49 @@ class common {
             }
         }
 
-        return itemTypeScript.filterItemsFromApi(mappedPossibleItems)
+        String[] items = itemTypeScript.filterItemsFromApi(mappedPossibleItems)
+
+        return items;
+    }
+
+    /**
+     * Retrieves all the available items for the target type in the form of a list that match the specified regex.
+     *
+     * Forming the regex:
+     * "\Q" starts an explicit quote (which includes all reserved characters as well)
+     * "\E" ends an explicit quote
+     * "\w*" is equivalent to "*" - it selects anything that has the same characters
+     * (in the range of a-z, A-Z or 1-9) as before and after the asterisk
+     * "." is equivalent to "?" - it selects anything that has the rest of the pattern but any
+     * character in the "?" symbol's position
+     * So, "\Q<INPUT_PART1>\E\w*\Q\<INPUT_PART1>E", selects anything that starts with INPUT_PART1
+     * and ends with INPUT_PART2 - This regex expression is equivalent to the input argument
+     * "INPUT_PART1*INPUT_PART2"
+     *
+     * @return a String[] containing the names of items available for download.
+     * @param regex the regex that the retrieved items should match
+     */
+    String[] retrieveAvailableItemsWithRegexMatch(String regex) {
+        ArrayList<String> selectedModules = new ArrayList<String>()
+        String[] moduleList = retrieveAvailableItems()
+        for (String module : moduleList) {
+            if (module.matches(regex)) {
+                selectedModules.add(module)
+            }
+        }
+
+        return ((String[]) selectedModules.toArray());
+    }
+
+    /**
+     * Retrieves all the available items for the target type in the form of a list that match the specified regex.
+     *
+     * @param wildcardPattern the wildcard pattern that the retrieved items should match
+     * @return a String[] containing the names of items available for download.
+     */
+    String[] retrieveAvalibleItemsWithWildcardMatch(String wildcardPattern) {
+        String regex = ("\\Q" + wildcardPattern.replace("*", "\\E\\w*\\Q").replace("?", "\\E.\\Q") + "\\E")
+        return retrieveAvailableItemsWithRegexMatch(regex);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


GooeyHub:

config/groovy/util.groovy
```diff
@@ -54,7 +54,35 @@ switch(cleanerArgs[0]) {
         } else {
             // Note: processCustomRemote also drops one of the array elements from cleanerArgs
             cleanerArgs = common.processCustomRemote(cleanerArgs)
-            common.retrieve cleanerArgs, recurse
+            ArrayList<String> selectedModules = new ArrayList<String>()
+            String[] moduleList = common.retrieveAvailableItems()
```
Cervator: I think this should be relocated to inside the wildcard block, otherwise it looks like _every_ item `get` operation would hit the GitHub API to retrieve every possible item, even if it already knows exactly which items it cares about

config/groovy/util.groovy
```diff
@@ -54,7 +54,35 @@ switch(cleanerArgs[0]) {
...
+
+            for (String arg : cleanerArgs) {
+                if (!arg.contains('*') && !arg.contains('?')) {
+                    selectedModules.add(arg)
+                } else {
+                    for (String module : moduleList) {
+                        /**
+                         * Forming the regex:
```
Cervator: Could this please be moved to a method in `common.groovy` ? Then you can attach the javadoc properly and we can keep `util.groovy` more slim. Somewhat of a code spell to start bloating logic inside a switch, cleaner to just deal solely with the logic about where to delegate.

config/groovy/util.groovy
```diff
@@ -214,6 +242,9 @@ def printUsage() {
     println "'-condensed-list-format' to group items by starting letter for the 'list' sub-command (default with many items)"
     println ""
     println "Example: 'groovyw module get Sample -remote jellysnake' - would retrieve Sample from jellysnake's Sample repo on GitHub."
+    println "Example: 'groovyw module get *' - would retrieve all the modules in the Terasology organisation on GitHub."
+    println "Example: 'groovyw module get *de*' - would retrieve all the modules in the Terasology organisation on GitHub" +
```
Cervator: Maybe throw in an example of the `?` syntax here? Possibly instead of / merged into this `*de*` example to save space

Also, if the Windows/Unix case remains a concern that should probably be spelled out here. If a Linux user wants to retrieve all modules would they have to enter `./groovyw module get '*'` ?

groovyw.bat
```diff
@@ -58,7 +58,24 @@ set _SKIP=2
 :win9xME_args_slurp
 if "x%~1" == "x" goto execute
 
-set CMD_LINE_ARGS=%*
+@REM set CMD_LINE_ARGS=%*
```
Cervator: Could you add a comment above this block to help spell out what it does? In particular since this is probably the first major custom bit of scripting added to the `groovyw` scripts themselves - I've otherwise tried to keep them as close to their `gradlew` parents as possible :-)

config/groovy/util.groovy
BenjaminAmos: I've merged the examples into each other and have added a message about the Unix issue.

config/groovy/common.groovy
```diff
@@ -323,7 +332,49 @@ class common {
...
+     * character in the "?" symbol's position
+     * So, "\Q<INPUT_PART1>\E\w*\Q\<INPUT_PART1>E", selects anything that starts with INPUT_PART1
+     * and ends with INPUT_PART2 - This regex expression is equivalent to the input argument
+     * "INPUT_PART1*INPUT_PART2"
+     *
+     * @return a String[] containing the names of items available for download.
+     * @param regex the regex that the retrieved items should match
+     */
+    String[] retrieveAvailableItemsWithRegexMatch(String regex) {
+        ArrayList<String> selectedModules = new ArrayList<String>()
```
Cervator: Trivial: shouldn't this (and the next few lines) be _items_ not _modules_ ?

config/groovy/util.groovy
```diff
@@ -54,7 +54,19 @@ switch(cleanerArgs[0]) {
         } else {
             // Note: processCustomRemote also drops one of the array elements from cleanerArgs
             cleanerArgs = common.processCustomRemote(cleanerArgs)
-            common.retrieve cleanerArgs, recurse
+            ArrayList<String> selectedModules = new ArrayList<String>()
```
Cervator: More `modules` -> `items` (and a couple more further down)

config/groovy/util.groovy
```diff
@@ -214,6 +226,13 @@ def printUsage() {
     println "'-condensed-list-format' to group items by starting letter for the 'list' sub-command (default with many items)"
     println ""
     println "Example: 'groovyw module get Sample -remote jellysnake' - would retrieve Sample from jellysnake's Sample repo on GitHub."
+    println "Example: 'groovyw module get *' - would retrieve all the modules in the Terasology organisation on GitHub."
+    println "Example: 'groovyw module get Sa??l*' - would retrieve all the modules in the Terasology organisation on GitHub" +
+            " that start with \"Sa\", have any two characters after that, then an \"l\" and then end with anything else." +
+            " This should retrieve the Sample repository from the Terasology organisation on GitHub."
+    println ""
+    println "*NOTE*: On UNIX platforms (MacOS and Linux), the arguments must be escaped with single quotes e.g. groovyw module get '*'."
```
Cervator: Maybe this tiny tweak?

```suggestion
    println "*NOTE*: On UNIX platforms (MacOS and Linux), the wildcard arguments must be escaped with single quotes e.g. groovyw module get '*'."
```

BenjaminAmos:
The 'modules' should now be correctly referred to as items. I have also applied the modified usage message.

Cervator:
Alright, finally getting back to this, as I merged in  and then thought I remembered there being another Groovy related PR i could get real quick. Might as well grab both, right? Can't possibly take that long to take this for another spin, right? 

_Github: Wrong!_

So since this PR was made the way GitHub paginates in their API requests apparently changed. When you reach the end rather than no additional request ... you get page 1 again. Fun! That took some puzzling out. After I realized it and looked for a fix I had a weird amount of trouble just getting a plain URL from what we get back for the next page request, and in the end gave in to a hack for the sake of expediency: if `page=1` is found in the URL we're done ... 

Only, while that finally got me results some of the testing gave funny results. A bit more poking around and I realized that, of yeah, the initial page request says it wants `page=100` per request, so ... yeah. 

Made me think of the olden days when I had time to actually code regularly. _Gods I was ~~strong~~ skilled then_ - actually nah not really but at least I might catch the simple stuff :p

After that I finally got expected results, added my hacky addition via https://github.com/MovingBlocks/Terasology/commit/dbc32f0da97e46c447ae2293a00e5babaf92c85b to a temp branch, tested more both on Mac and Windows, noting the need to have quotes around wildcards on my Mac. Also tweaked at the way caching works so you don't need an API hit if you know which exact modules you want (in that case just construct the URLs, don't need to ask GitHub for all the things that exist to then validate whichever you care about)

Finally validated `groovyw module update-all` and `groovyw module get *` on my Windows machine. So even though my added code is a little hacky I have a decent feeling for this now and really want to get back to merging all the things already.

@BenjaminAmos - sorry for taking so long to merge this and thank you again so much for the effort you put into it and other things. I'm sad we haven't seen you around since GCI, your efforts are highly appreciated. We just started GSOC and you could easily help mentor any of the projects, earning more Google swag, optional stipends, travel opportunities etc. DestSol may also participate in our inaugural "Terasology Summer of Code" where you'd fit right in. Just get back in touch sometime - I hope to hear from you again!

Issue #3466, opened 2018-08-06
Improve Record and Replay Serialization


iaronaraujo:
### The Problem

Right now, to add a new event type to Record and Replay (RR), It is necessary to add code to the `EventCopier` and `RecordedEventSerializer` that copies every attribute of the event and transforms it into JSON. This makes the addition of new event types time consuming and makes those two classes even bigger, which may cause its reading to be confusing and tiresome.

### Suggestion

Once the PR 3456 is merged, there should be enough resources to greatly improve how the events are serialized and copied, making the whole serialization more general, without having to write code for each and every type of event supported by RR.

iaronaraujo:
Another thing I just thought about. Once this is done, it is possible that RR will need a way to make sure the Entity IDs do not change from the recording to the replay. This may include chunks IDs.

PR #3533, opened 2018-10-28
Refactor R&R serialization

llvieira:
### Contains

With `CollectionTypeHandler`, the way `RecordedEvents` are serialized are now greatly improved, making the whole serialization more general, without having to write code for each and every type of event supported by R&R. 

By the way, before in `RecordedEventSerializer` we had **523** lines and now only **112** including javadocs. :smile: 

### How to test

- Create a record.
- Replay the record.

- [x] Refactor R&R serialization, see #3466. 
- [ ] Fix the problem with `RuntimeDelegatingTypeHandler`. It currently looks up the class via `Class.forName`, but this must be changed because `Class.forName` uses the engine class loader to load the class, which a) does not have access to module classes and b) _does_ have access to all engine classes, which is a security hole (modules may indirectly access internal engine classes by using this).
- [ ] Update this branch with `develop` to merge it afterwards.


GooeyHub:

Cervator:
This pull request **fixes 2 alerts** when merging 77eb98de8e940854a5d631d50098e21c2bf784a4 into 7a2c6cb46f7b5c6bb66ff7f7b768be4bd77c3ab0 - [view on LGTM.com](https://lgtm.com/projects/g/MovingBlocks/Terasology/rev/pr-db7b3ab31cabdcebe6e2f1bf76f95095aee4f870)

**fixed alerts:**

* 1 for Dereferenced variable may be null
* 1 for Potential input resource leak

---

engine/src/main/java/org/terasology/persistence/typeHandling/coreTypes/RuntimeDelegatingTypeHandler.java
```diff
@@ -56,7 +56,12 @@ public RuntimeDelegatingTypeHandler(TypeHandler<T> delegateHandler, TypeInfo<T>
     public PersistedData serializeNonNull(T value, PersistedDataSerializer serializer) {
         // If primitive, don't go looking for the runtime type, serialize as is
         if (typeInfo.getRawType().isPrimitive()) {
-            return delegateHandler.serialize(value, serializer);
+//            LOGGER.info("DelegateHandler: " + delegateHandler);
+            if (delegateHandler != null) {
+                return delegateHandler.serialize(value, serializer);
+            }
```
eviltak: This change should not be needed since primitive types will always have a TypeHandler (thus ensuring `delegateHandler` will never be null). The problem is propagating from somewhere else. We _could_ make this change here, but this is just deferring the change elsewhere.

engine/src/main/java/org/terasology/persistence/typeHandling/coreTypes/RuntimeDelegatingTypeHandler.java
```diff
@@ -56,7 +56,12 @@ public RuntimeDelegatingTypeHandler(TypeHandler<T> delegateHandler, TypeInfo<T>
...
+
+            return null;
```
eviltak: Always use `serializer.serializeNull()`. Returning null `PersistedData` references will have unforeseen repercussions. With the new `TypeHandler` refactor (and all other new code I introduce in the Terasology codebase), I aim use `Optional` instead of nulls to make absent values more obvious.

engine/src/main/java/org/terasology/persistence/typeHandling/coreTypes/RuntimeDelegatingTypeHandler.java
```diff
@@ -56,7 +56,12 @@ public RuntimeDelegatingTypeHandler(TypeHandler<T> delegateHandler, TypeInfo<T>
     public PersistedData serializeNonNull(T value, PersistedDataSerializer serializer) {
         // If primitive, don't go looking for the runtime type, serialize as is
         if (typeInfo.getRawType().isPrimitive()) {
-            return delegateHandler.serialize(value, serializer);
+//            LOGGER.info("DelegateHandler: " + delegateHandler);
```
eviltak:
```suggestion
```

engine/src/main/java/org/terasology/persistence/typeHandling/coreTypes/RuntimeDelegatingTypeHandler.java
```diff
@@ -82,6 +87,8 @@ public PersistedData serializeNonNull(T value, PersistedDataSerializer serialize
             }
         }
 
+//        LOGGER.info("DelegateHandler: " + delegateHandler);
```
eviltak:
```suggestion
```

engine/src/main/java/org/terasology/recording/RecordAndReplaySerializer.java
```diff
@@ -119,7 +117,7 @@ private void serializeFileAmount(Gson gson, String recordingPath) {
             writer.close();
             logger.info("File Amount Serialization completed!");
         } catch (Exception e) {
-            logger.error("Error while serializing file amount:", e);
+            logger.error("Error while serializing file amount", e);
```
eviltak: Unless absolutely needed, in the future please refrain from making such small irrelevant changes: the removal of the semi-colon in the error log makes no difference whatsoever.

engine/src/main/java/org/terasology/persistence/typeHandling/coreTypes/RuntimeDelegatingTypeHandler.java
llvieira: Entirely agree with you. We need to figure out the source of the problem. I've logged out all the events to find the problem, then I noticed that after the `KEY_SCAPE` event some weird events just popped out. These events are represented by a character `'\u001B' 27` (which I think is a escape character), so it belongs to primitive types. However, when I logged out its respective `delegateHandler` it had a `null` value. In other words, no `TypeHandler` for this weird char. Therefore, adding the if statement is the workaround that I found.

At this picture we can see some of the event logs, including the weird one:
image

Maybe another PR to fix that problem would be great, because I still have no idea where it came from, but yeah it has nothing to do with serialization.

engine/src/main/java/org/terasology/persistence/typeHandling/coreTypes/RuntimeDelegatingTypeHandler.java
llvieira: Oh, great! I can use that as part of the workaround before fixing the problem we commented above.

llvieira:
@eviltak thanks for your review, really appreciated that. And yes, the other changes like fixing `RuntimeDelegatingTypeHandler` I will make in separate pull requests. For now, I just want to get R&R running with this amazing serialization improvement. I hope we get all those things merged still this year. By the way, thanks for the serialization overhaul, without it these changes would never be possible.

engine/src/main/java/org/terasology/persistence/typeHandling/coreTypes/RuntimeDelegatingTypeHandler.java
llvieira: I have merge this branch with `develop`, there is a chance that this problem is fixed there. We are 95 commits behind `develop`, that's bad. :/

llvieira: I need to merge this branch with `develop`, there is a chance that this problem is fixed there. We are 95 commits behind `develop`, that's bad. :/

llvieira: Yep, figured out! I'm going to implement the `CharTypeHandler` and remove this workaround. Thanks for your help! :-)

eviltak:
@llvieira I'm going to have to ask you to create a new PR branched off of the newly updated `newSerialization` branch tip (I just pushed some commits) with just the changes to `RecordedEventSerializer`. `newSerialization` now has all the fixes you included in your PR, so you shouldn't need to touch any other files. Also, please do not merge in/rebase over `develop`: this will be done when `newSerialization` itself is merged into `develop`. Thank you for all your work, it should come to fruition very soon!

PR #3537, opened 2018-10-31
Add option to trust all URLs to the same website

nschnitzer:
### Contains

Adds option to trust all URLs with a certain hostname.

### How to test

Try and open URLs with the same hostname after selecting the option.

### Outstanding before merging

If anything. You can use neat checkboxes! Feel free to delete if not needed

- [ ] Need to consider use case x
- [ ] Still have to adjust the wiki doc
- [ ] Will need translation work

PR #3539, opened 2018-11-01
Add option to trust all URLs to the same website

llvieira:
This PR contains the work made by @nschnitzer for his GCI task. The original PR #3537 get automatically closed by git after a wrong push in the attempt to fix some indentation issues.

### Contains

Adds option to trust all URLs with a certain hostname.

### How to test

Try and open URLs with the same hostname after selecting the option.

Credit @nschnitzer.


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UIButtonWebBrowser.java
```diff
@@ -147,16 +154,33 @@ private boolean hasConfirmation() {
...
     public UIButtonWebBrowser setUrl(String url) {
+        boolean trustedHostName = false;
         boolean trustedUrl = false;
+
         if (webBrowserConfig != null) {
-            trustedUrl = webBrowserConfig.isUrlTrusted(url);
+            try {
+                String hostname = new URL(url).getHost();
+                trustedHostName = webBrowserConfig.isHostNameTrusted(hostname);
+            } catch (MalformedURLException e) {}
```
jellysnake: I think we shouldn't just catch and silently ignore? Would be nice if we could somehow alert that there is a malformed URL.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/moduleDetailsScreen/ModuleDetailsScreen.java
```diff
@@ -192,9 +193,7 @@ public String getString(Module value) {
     }
 
     private List<DependencyInfo> getSortedDependencies(final Module module) {
-        return module.getMetadata()
-                .getDependencies().stream()
-                .sorted(Comparator.comparing(DependencyInfo::getId))
+        return module.getMetadata().getDependencies().stream().sorted(Comparator.comparing(DependencyInfo::getId))
```
jellysnake: It seems like this file is just formatting changes. Some of them are good but also some of them i'd argue actually make it less clear. In particular the ones involving the chained calls like this.

We should probably just discard all the changes to this file as it seems superfluous. Also seems like formatting issues are the biggest problem here haha

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/ConfirmUrlPopup.java
```diff
@@ -73,21 +74,38 @@ public void setMessage(String title, String message) {
...
+                    webBrowserConfig.addTrustedHostName(hostName);
                 } else {
-                    webBrowserConfig.removeTrustedUrl(url);
+                    webBrowserConfig.removeTrustedHostName(hostName);
                 }
             });
+
+            UICheckbox saveUrl = find("saveUrl", UICheckbox.class);
+
+            if (saveUrl != null) {
```
jellysnake: I highly doubt this should be nested in the `saveHostName` if. They should be separate imo

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/moduleDetailsScreen/ModuleDetailsScreen.java
llvieira: Entirely agree. All changes discarded.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/ConfirmUrlPopup.java
llvieira: Yes, they should be separate. Fixed!

GooeyHub:

PR #3543, opened 2018-11-12
Fixing PMD Issues

claytonj222:
### Contains

### How to test


GooeyHub:

engiValk:
Hey there,

Thanks for taking the time to do these changes and make a PR. Unfortunately due to the size of this PR and the lack of documentation of these changes in your commit message I'm going to close it. In the future it would be beneficial to aim for smaller Pull Requests and ensuring your commit message is clear and precise as to exactly what your changes will be doing in detail.

Also the issue of file conflicts would need to be addressed. Feel free to make a new Pull Request with the above feedback taken into consideration.

-engiValk

jellysnake:
To add onto the above, the reason why there are so many files is because you have included thing that should not have been committed. For instance, the IntelliJ project files.

Another reason is that there are a number of changes that are not to do with the logging changes you've made. This is likely from autoformatting files with some seeing turned on.

It seem like you didn't set up the game by [following our developer guide](https://metaterasology.github.io/docs/developerIntro/fullWorkspace.html). Try setting it up with that and then we can try and review the PR properly for you.

Issue #3547, opened 2018-11-15
Error on attempting to delete saves


jellysnake:
When trying to delete a save the game will error. This is because the world preview image is located in the save directory and the reader that opens the file is not closed. Hence when the game tries to delete the game folder it can't because the folder is in use.

This can then also bleed over into trying to save a game later, if that game has the same name as the attempted-deleted save.

This can likely be fixed by closing the IO stream after loading the preview image.

# Reproducing
Just try and delete any save.
Once it fails, create a new world (with no name) and then try and save.

pardovot:
No solution yet, but a few things meanwhile:
The problem is coming from 2 sources combined, the first one is ```takeGamePreview``` method in ```ScreenGrabber``` class, which saves the preview image.
The line: ```this.savedGamePath = GamePreviewImageProvider.getNextGamePreviewImagePath(saveDirPath);```

Which really leads to ```getNextGamePreviewImagePath``` in ```GamePreviewImageProvider``` Class.

And also ```loadPreviewImages``` method in ```SelectionScreen``` class by calling to ```getAllPreviewImages``` method in ```GamePreviewImageProvider```
```final List<BufferedImage> bufferedImages = GamePreviewImageProvider.getAllPreviewImages(gameInfo.getSavePath());```

Tried closing the IO stream in ```getAllPreviewImages```, but it did not solve the problem.

The problem seems to be a combination of the 2, as by only calling each one of them seperately the problem still persists, and only by not calling both of them the problem is fixed, but then the image preview is not created.

I'd say the problem is something regarding the actual way java manages the files, might be wrong.

pardovot:
The problem is the following method:
```    private static List<Path> getAllPathsToFilesInFolder(final Path dirPath) {
        // create folder if not exists yet
        createDirectory(dirPath);

        try {
            return Files.list(dirPath).filter(Files::isRegularFile).collect(Collectors.toList());
        } catch (IOException e) {
            logger.warn("Can't correctly read files from {}", dirPath);
            return Collections.emptyList();
        }

    }```
In the ```GamePreviewImageProvider``` class, more specifically - ```return Files.list(dirPath).filter(Files::isRegularFile).collect(Collectors.toList());``` This line causes the problem.

Tried somehow closing the stream, didn't work so far.

Cervator:
Closing as fixed by #3549 - thanks @pardovot !

PR #3549, opened 2018-11-18
Fixed deleted saves error.

pardovot:
Fix for deleted saves error #3547 

The return line used a stream which remained open and caused the folder to be inaccessible, giving an error in the delete process.
I've put the stream in a try with resources construct which closed the stream right after using it.

### How to test

Make a new save, then try to delete it.


GooeyHub:

GooeyHub:

Cervator:
Tested out - works! Thank you so much :-)

PR #3551, opened 2018-11-20
Make the order of overlapping widgets change

RatMoleRat:
Allows the order of display (what renders in front of what) of overlapping windows to change when the semicolon key is pressed. 
This can be demonstrate by taking two overlapping UI screens - I used the core inventory screen and the crafting screen from the BasicCrafting module to test, although others should work - and press the semicolon key to see the front one move to the back and the back one to the front.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/SortOrder.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/SortOrder.java
@@ -43,7 +43,6 @@ public class SortOrder extends BaseComponentSystem {
     private static int current;
     private static ArrayList<Integer[]> layersFilled; //arg1 of the Integer[] is the layer depth, arg2 is the number of things on that layer
     private static int index;
-    private static int uiIndex;
     private static boolean inSortOrder;
     private static ArrayList<CoreScreenLayer> enabledWidgets;
     private static boolean initialized = false;
@@ -104,7 +103,6 @@ public class SortOrder extends BaseComponentSystem {
         }
         current = 0;
         index = 0;
-        uiIndex = -1;
         layersFilled = new ArrayList<>();
         enabledWidgets = new ArrayList<>();
         used = new ArrayList<>();
@@ -230,10 +228,6 @@ public class SortOrder extends BaseComponentSystem {
     }
     public static ArrayList<CoreScreenLayer> getEnabledWidgets() {
         return enabledWidgets;
-    }
-    public static int makeIndex() {
-        uiIndex++;
-        return uiIndex;
     }
 
     public static boolean isInitialized() {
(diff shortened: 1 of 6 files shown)


GooeyHub:

engiValk:
@GooeyHub: ok to test

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/SortOrder.java
```diff
@@ -0,0 +1,246 @@
...
+     * Initializes sort order
+     * @param event
+     * @param player
+     */
+    @ReceiveEvent
+    public void onPlayerSpawnedEvent(OnPlayerSpawnedEvent event, EntityRef player) {
+        initialized = true;
+        Map<Integer, BindableButton> keys = bindsManager.getKeyBinds();
+        if (keys.containsKey(Keyboard.Key.SEMICOLON.getId())) {
+            keys.get(Keyboard.Key.SEMICOLON.getId()).subscribe(new BindButtonSubscriber() {
```
Cervator: Semi colon might be good for testing, but not very intuitive - maybe we could use something more like alt-tabbing. Some other games and apps use ctrl-tab or shift-tab, think that would be possible? Maybe even both - shift being an order reverser.

engine/src/main/java/org/terasology/rendering/nui/SortOrder.java
```diff
@@ -0,0 +1,246 @@
...
+import org.terasology.registry.In;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Deque;
+import java.util.Map;
+
+
+@RegisterSystem(RegisterMode.ALWAYS)
+public class SortOrder extends BaseComponentSystem {
```
Cervator: If this is a `System` class ideally end the name in `System` :-)

engine/src/main/java/org/terasology/rendering/nui/internal/NUIManagerInternal.java
```diff
@@ -104,7 +105,14 @@ public NUIManagerInternal(CanvasRenderer renderer, Context context) {
         ModuleAwareAssetTypeManager maaTypeManager = context.get(ModuleAwareAssetTypeManager.class);
         maaTypeManager.getAssetType(UIElement.class).ifPresent(type -> type.disposeAll());
     }
-
+    @Override
```
Cervator: One blank line between method blocks please

RatMoleRat:
I made the various changes, and I now have it set up to change the order when the alt key is pressed, because that seems more intuitive than the semicolon, and because tab is linked to another UI widget popping up. I could still change it to tab, but figured that alt is better because it doesn't seem to already be used by anything.

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/SortOrder.java
Cervator: I'd be curious if we could suppress the online players view if any modifier key is also used, but that could be something to worry about later :-)

GooeyHub:

PR #3554, opened 2018-11-23
Player can't create a game with special name from the player

iHDeveloper:
### Contains
I found that when I create a game in the singleplayer. And put for example `Terasology World` in the game name textbox. Then the game should put that name But sadly the game generates a name for the game instead of using my given name. So, I made a small change to fix this issue.

### How to test
- Open the game
- Go to singleplayer
- Click on `Create` ( If you don't have any game it will show you the create screen )
- Fill any name from your mind
- Click on `Play`
- Click `Esc`
- Go back to the game menu. You will find the game with the given name that you wanted it to be.

- [ ] Does this will cause any issues in the future?


iHDeveloper:
Also, When the player clicks `Create` and he or she leaves the game name textbox empty. Then, the game will generate a name for the game instead of showing a dialog that the text box is empty.

Cervator:
Tests out - works well! Thanks :-) Also added you to the credits

Issue #3567, opened 2018-12-02
Extract biome-per-block to a module


Cervator:
With #3476 merged we can now add per-block data fields in a module. Previously we implemented biome-per-block in #1242 which @Adrijaned then included in a [nice wiki page on biomes](https://github.com/Terasology/TutorialWorldGeneration/wiki/Biomes)

However, biome-per-block isn't necessarily needed in every case, leaving to that field being unused per-block. Things like the new `BiomeChangeEvent` in #3560 wouldn't need to check movement constantly to see if a player is changing biome, only when the new module is active. Speaking of, we have an `OnEnterBiomeEvent` already, came in via #2954, is that a duplicate?

One tricky bit is that we have both biome-per-block and custom biome systems, including `CoreBiome` etc in the Core module. See for instance #2303 for some code changes made to _facet_ based biomes, which are less accurate (2D I believe) but perform better. 

Could the other biome stuff also be extracted? Possibly, but that may be far trickier since it is in use within the engine/Core. It would be like extracting Health or Inventory. It likely *should* be done at some point, but this issue is just to investigate and if possible deal with the per-block biomes, which are rarely used (see https://github.com/Adrijaned/DeadIslands for one usage)

Working this item would involve moving just the per-block biome stuff to a new module, which would remove the associated classes from this repo, breaking API, but I believe we could forgive that. It would also need to involve updating the tutorial on worldgen and sample code since the new module would be a dependency.

For a bit more historical background consider reading https://forum.terasology.org/threads/modding-per-block-storage-extensions.806 or other forum threads on the topic.

Adrijaned:
Oh, didn't know about the existence of the other event... but mine will be easier to extract into the module :grin:. I'll try to tackle this after I finish some other work, if no one has by then

Cervator:
Great :-) Let me know if you want to have a GCI task created for it when you're ready.

PR #3570, opened 2018-12-03
Removing biomes, to be moved to module

Adrijaned:
Works now; just protobuf needs to be recompiled


Adrijaned:
Fixes #3567

GooeyHub:

jellysnake:
On the topic of Core needing to depend on this, A similar situation existing with my (languishing) Inventory PR #3419.
A solution proposed and partially hashed out was to switch to having a set of core modules, that can constitute similar to Core does at the moment. Biomes and Inventory would be good fits for this at present.

Alternatively we can instead look at making the core dependency on Biomes/Inventory/Others optional but that may require more work than it's worth given how they'd likely be enabled 9 times out of 10.

There was some write ups on this topic but we'll have to wait for Loremaster Cervator to share them (seeing as I've no idea where they were)

GooeyHub:

Adrijaned:
This should be nearly finished. Only the rendering bug remains, and the fact that 75e4b91 somehow broke movement...

GooeyHub:

Adrijaned:
Movement is no longer broken; just the rendering bug remains and this should be up for review!

GooeyHub:

Adrijaned:
Not sure if I didn't actually break something. Someone (on windows, presumably) still has to recompile the proto files (I couldn't get `protoc` to work)

jellysnake:
I can't easily find the module in question so if you've made a repo for it then link it here. And if you haven't you might as well go ahead and make one.
That's the reason the build keeps failing, similar to the issue with the inventory pull.


Other than that, how did biomes end up breaking rendering? Where do they intersect that much?

Adrijaned:
The module is in this PR? (it was added to the engine repo for now)

Adrijaned:
Regarding the biomes and rendering - I believe there was a special case for grassBlocks somewhere in rendering code that was offseting the color a tiny bit based on humidity and the other two fields in the od biome - I might have messed up that a bit while removing the biomes, Only the rendering of grass is not working as expected, and only the _sides_ of grass are broken now, the green rendering as white.

Adrijaned:
Thanks to an invaluable input from @vampcat, rendering now works correctly

jellysnake:
Is this ready to be tested out and then ideally merged now? Or is there still some outstanding known tweaks/bugs to be made.

Adrijaned:
This should be ready to review&merge, nothing outstanding I know about

GooeyHub:

GooeyHub:

GooeyHub:

PR #3571, opened 2018-12-04
feat: Add AFK state for players in multiplayer

iHDeveloper:
### Contains
As a gamer sometimes I have to go away from the keyboard for some reason and I want to tell my friends who I'm playing with that I'm away so they won't get mad at me or something. So, I made an `afk` command to do that goal.

### How to test
- Pull from [My Fork](https://github.com/iHDeveloper/Terasology) from branch `afk-command`
- Run the game from the source code
- Make sure you in multiplayer mode like (Hosting game or Playing online)
- Show the game console
- Type `afk`
- You can now go away without telling your friends that I'm AFK in the chat :smile: 

- [x] Client can't send request to the **server**
- [x] Rebuild the way that client and server communicate in the `AFKSystem`
- [x] Auto disable the AFK when the player moves
- [x] Test it


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/logic/afk/AFKSystem.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @In
+    private NetworkSystem networkSystem;
+
+    private AFKComponent component;
+
+    @Override
+    public void initialise() {
+        component = new AFKComponent();
+        localPlayer.getClientEntity().addComponent(component);
```
jellysnake: So this does work but I'd recommend just directly adding the component to the players prefab. That way it can be hooked into via out `delta` and `overrides` systems

engine/src/main/java/org/terasology/logic/afk/AFKSystem.java
```diff
@@ -0,0 +1,89 @@
...
+import org.terasology.logic.console.commandSystem.annotations.Command;
+import org.terasology.logic.permission.PermissionManager;
+import org.terasology.logic.players.LocalPlayer;
+import org.terasology.network.NetworkMode;
+import org.terasology.network.NetworkSystem;
+import org.terasology.physics.events.MovedEvent;
+import org.terasology.registry.In;
+import org.terasology.registry.Share;
+
+@Share(AFK.class)
```
jellysnake: So just so you are aware, this is only needed if you want other systems to be able to inject yours via `@In`. Given that you actually don't expose any methods I'd say that this isn't really needed.

kaen:
I tested this out locally. First very minor note is that the message in the console uses a curly brace where a square bracket is expected: `[AFK} You are AFK!`

More importantly, it seems that this only works as intended for the host going afk. Here are some scenarios I tested by hosting with (no saves) launch configuration, then launching 2nd and 3rd client from the configuration dropdown and connecting to localhost server.

### Scenario 1 (works as intended)
- host types afk
- host shows afk on their own player list
- host shows afk on 2nd client
- host shows afk on 3rd client
- host returns from afk
- host shows not afk on both client lists

### Scenario 2 (not working as intended)
- 2nd client types afk
- 2nd client shows afk on their own list
- 2nd client *does not* show afk on 3rd client's list
- 2nd client *does not* show afk on host's list

### Scenario 3 (not working as intended)
- host, 2nd, and 3rd client type afk
- host list only shows host afk
- 2nd client shows host and 2nd client afk
- 3rd client show host and 3rd client afk
- clients are not shown afk on other client's lists

Cervator (dismissed):
Can confirm as well - AFK shows for the player running the command, but that's it :-)

Also, I get the impression the player is meant to see the "Welcome back, you are no longer AFK!" message in the little chat message bar at the bottom of the screen? It pops up when a player starts moving, but appears blank to me, while the message gets logged in the regular console OK. Unsure if that's bug here or if any input to the console will trigger a blank chat notification as well.

A few times it seemed like I went un-afk without really doing anything. Maybe it is overly sensitive? Perhaps instead of using the move event use the one where the player has sent an actual message? Then they're certainly no longer afk. But if a cat poked their mouse - no more afk! :-)

Can also validate that the afk command doesn't work in single player, where it would be kinda silly. But on the other hand having extra logic for that may not be needed.

I suspect you need one more layer to make this work. See how `AFKSystem` (which should really be `AfkSystem`, likewise the component should be `AfkComponent`) is set to `ALWAYS` ? That means it'll do the same thing both on clients and the server. Right now it works solely locally.

You may well want to have a _Client_ only system that registers and handles the command, but sends a note to the server "Hey I'm going AFK" (you send an event that's being listened to in an _Authority_ system) and then the server can send a `BroadcastEvent` that the client systems catch with the name of the now-AFK player in it. Then each client can update their local state to know that player X is AFK and show that in the overlay. When the player comes back you repeat the process to let everybody know that player is back now

So in other words:

* `AfkClientSystem` registers the command and when triggered sends an event (that's replicated / sent to the server somehow, I forget how) "Hey I'm going AFK, my player name is X"
* `AfkAuthoritySystem` listens to the event indicating that player X is going AFK, re-issues a `BroadcastEvent` to everybody with that information
* `AfkClientSystem` has a listener for the broadcast event, when received you now know to mark that player AFK in the overlay

You should be able to look at the players online overlay itself to see how it transfers state around to the different clients connected with their ping so it all shows up correctly everywhere :-)

engine/src/main/java/org/terasology/logic/afk/AFKSystem.java
iHDeveloper: **Done** I deleted it because I don't actually need it.

engine/src/main/java/org/terasology/logic/afk/AFKSystem.java
```diff
@@ -0,0 +1,89 @@
...
+    @In
+    private NetworkSystem networkSystem;
+
+    private AFKComponent component;
+
+    @Override
+    public void initialise() {
+        component = new AFKComponent();
+        localPlayer.getClientEntity().addComponent(component);
+        logger.info("Successfully! Initialised the AFK system");
```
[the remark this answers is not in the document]
iHDeveloper: **Done** I removed it because I think we don't need it.

engine/src/main/java/org/terasology/logic/afk/AFKSystem.java
iHDeveloper: **Done** I made the component to be added in `client.prefab` in `player`.

GooeyHub:

iHDeveloper:
I think we are ready to merge it. I tested it with my friend [Falcoph](https://github.com/falcoph). Thanks, @Cervator for your notes It wasn't gonna happen now without them. Also, Thanks, @kaen for the test scenarios and now this pull can pass all of those scenarios 😄

skaldarnar:
I will test this out when I get a bit free time ;-) 
One thing I'm wondering about is whether you need the boolean field on `AFKComponent`, or whether the presence of the component is enough to indicate that a player is actually afk, and the absence denotes the opposite (not afk).

iHDeveloper:
Thanks for your help @skaldarnar. I implemented new features in the list below:
- Shows `Idle` in the Discord RPC when the player is afk
- Auto `afk` with no active for like 5 minutes
- Auto `unafk` with any move or even touch your keyboard

You can test them and give us the result of your test. That's will be helpful from you 😄 
**Note:** You can also try the scenarios that @kaen did in the last test result from him. 😉

GooeyHub:

jellysnake:
I probably should have been clearer with my explanation.
This is mostly a rehash of my discord message, but here it won't scroll away.

What I meant was that the `DelayManager` listens for changes made to `PeriodicActionComponent`
So rather than re-implementing the delay logic from `DelayManager` I was intending for you to look at the logic only in `addPeriodicAction` and then copy parts of it.
You'll still use `DelayManager` behind the scenes, just changing how you access it.

To get more technical and specific, you should delete all the methods you copied except for `addPeriodicAction`, the `PeriodicActionTriggeredEvent` handler and `getPeriodicComponent`.
Then call the add method in `postBegin` and that should work.

If you can't get it to work then you could either just do the auto afk because it's not strictly needed for the task, or try and move it to the Authority method you have.

GooeyHub:

GooeyHub:

GooeyHub:

skaldarnar (approved):
Tests out fine with two clients in a local setup after updating this PR according to NUI extraction. 

I'm not sure whether this should live directly in the engine, or whether it should be extracted to a module, but that can be done in a follow-up.

So, better late than never - LGTM!

engine/src/main/java/org/terasology/logic/afk/AfkAuthoritySystem.java
```diff
@@ -0,0 +1,89 @@
...
+    @In
+    private DelayManager delayManager;
+
+    @In
+    private LocalPlayer localPlayer;
+
+    @In
+    private NetworkSystem networkSystem;
+
+    private int count;
```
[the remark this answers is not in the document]
skaldarnar: I've created https://github.com/MovingBlocks/Terasology/issues/4242 to keep track of this. @iHDeveloper might this be your next contribution? 😉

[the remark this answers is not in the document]
iHDeveloper: @skaldarnar Sure. I'm currently busy with exams. But, I will work on it during the weekend 😁

PR #3573, opened 2018-12-10
Apply special keyboards option in Input settings

Koward:
### Contains

This PR adds new buttons on the Input settings screen to quickly change your key mappings if you have a special keyboard like AZERTY, DVORAK or NEO.
The mappings were already available through console commands but that would not save them in the settings and that was difficult to reach for a casual player.

There is a localization of the title and the buttons for both English and French.

Screenshot:
screenshot from 2018-12-10 10-42-31

### How to test

1. On the main screen, open Settings then Input. You should see the new section "Special keyboards".
2. Try it by clicking on Apply for some special keyboard like AZERTY.
3. The changes in mapping should be instantly reflected in the binding buttons below (in Movement, Interaction, etc.) if they were supposed to change for the special keyboard you chose.
4. You might want to check in-game and ensure the new mappings are effectively applied.
* Note: the steps are also be available from the in-game Settings menu


GooeyHub:

jellysnake:
@gooeyhub ok to test

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings/InputSettingsScreen.java
```diff
@@ -48,23 +45,18 @@
...
-import org.terasology.rendering.nui.widgets.UILabel;
-import org.terasology.rendering.nui.widgets.UISlider;
-import org.terasology.rendering.nui.widgets.UISpace;
-
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
+import org.terasology.rendering.nui.widgets.*;
```
jellysnake: See other comments about stars

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings/InputSettingsScreen.java
```diff
@@ -27,11 +27,8 @@
 import org.terasology.engine.module.ModuleManager;
 import org.terasology.engine.subsystem.config.BindsManager;
 import org.terasology.i18n.TranslationSystem;
-import org.terasology.input.BindButtonEvent;
-import org.terasology.input.Input;
-import org.terasology.input.InputCategory;
-import org.terasology.input.InputSystem;
-import org.terasology.input.RegisterBindButton;
+import org.terasology.input.*;
```
jellysnake: Terasology style strongly avoids star imports. Please change this back to importing items individually. [If you are using IntelliJ then this guide will help](https://stackoverflow.com/questions/3587071/disable-intellij-starred-package-imports)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings/InputSettingsScreen.java
```diff
@@ -94,6 +86,35 @@ public void initialise() {
...
+            BindCommands.DVORAK.forEach(this::setPrimaryBind);
+            bindsManager.registerBinds();
+        });
+        UIButton neo = new UIButton();
+        neo.setText(translationSystem.translate("${engine:menu#input-settings-apply}"));
+        neo.subscribe(event -> {
+            BindCommands.NEO.forEach(this::setPrimaryBind);
+            bindsManager.registerBinds();
+        });
+        mainLayout.addWidget(new UILabel("specialKeyboards", "subheading", translationSystem.translate("${engine:menu#special-keyboards}")));
```
jellysnake: The preferred way to edit UI files is by manually altering the `.ui` files for the screen you are after. [In your case that would be this one](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/ui/menu/inputSettingsScreen.ui)
If you need help with how to use our UI system then [check out our tutorial](https://github.com/Terasology/TutorialNui) or [hop onto our discord](http://discord.gg/Terasology)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings/InputSettingsScreen.java
```diff
@@ -94,6 +86,35 @@ public void initialise() {
         mainLayout.setVerticalSpacing(8);
         mainLayout.setFamily("option-grid");
 
+        UIButton azerty = new UIButton();
```
jellysnake: This would probably work better using a `UIDropdown` instead as that's designed to choose between multiple options, which is what's being done here.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings/InputSettingsScreen.java
Koward: Thank you, I must have missed this guideline.
This has now been corrected.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings/InputSettingsScreen.java
Koward: I agree a `UIDropdown` would feel more natural. I think it should be designed with a different behavior than the button and the console command.

At the moment, "Apply" only changes a set of mappings. Other mappings are untouched.
That means if I blindly replaced the buttons by a dropdown, moving from one state to another and then back would not give the same result.
For example, moving from AZERTY to NEO then back to AZERTY would leave some NEO mappings that were applied. In my opinion that is not what a user would expect from a dropdown.

For that reason I think I will change the behavior to restore defaults all keys before applying the special keyboard bindings when selecting a new keyboard state.

GooeyHub:

GooeyHub:

GooeyHub:

Issue #3577, opened 2018-12-18
Follow-ups to #3561 - UI improvements / fixes


Cervator:
After merging and testing #3561 I noticed some issues, although I'm not 100% sure they're all coming in via that PR. Would like to get them fixed and some more found as I expect quirks - pretty substantial new feature :-)

- [ ] Sliders didn't work for me on the audio menu
- [ ] Pressing `ESC` seemed to break further attempts to tab around on a page (due to losing focus somehow?)
- [ ] The server status on "Join Server" doesn't seem to refresh? That usually happened when a server entry in the list became selected. Does seem like it gets queried, just no info shows up. Joining seems to still work
- [ ] Alignment in the settings menus seem off, but that may just be from my workspace. Would be good to confirm. Seems like a margin issue.

Found some more? Add them here! Fixed some? PR! Got additional details/research about some of the items? Post!

RatMoleRat:
Sliders are not currently able to be tabbed to at all, which is why they don't work. Adding support for them (and input boxes, so that they could be tabbed to and then typed in) would be nice.

Another issue (though sort of another personal preference) is that it is hard to tell when the UIList is tabbed to and able to be selected from. Adding some kind of visual indication that the list is what's currently selected through tabbing might make it look better.

The PR I just made should make tabbing after `Esc` work.

Cervator:
Thanks! Updated the list :-)

mayant15:
Larger menus (like video settings) should scroll as we tab. The widget in focus should always be visible.

PR #3581, opened 2018-12-21
Fixed tabbing between widgets after Escape is pressed

RatMoleRat:
### Contains

Fixes widget tabbing so that it can be used after `Esc` has been pressed (the second problem mentioned in #3577)

### How to test

Press `Esc` on the main menu or other screens, then `Tab` in order to tab through the widgets.


GooeyHub:

PR #3588, opened 2018-12-25
Fixes the display mode dropdown

### LINKED ISSUE #3587: Game behaves poorly on interacting with the windowed/fullscreen option

Change from windowed to fullscreen mode. Expected that interacting with (clicking/activating) the display mode drop-down would show all the available options on one interaction then let me choose with a second interaction

### What actually happened

Game window disappeared (as if about to change display mode) but game kept running and the whole OS started mildly freaking out (seems like it outright crashed an OS process for @Adrijaned on Linux). It feels like the game gets stuck in a state where it is trying to swap windowed state constantly

Theory is that it relates to how activating the UI widget works, likely after tabbing was introduced. Actually tabbing to the widget and varying the display mode with down arrow works. Only clicking the widget with the mouse makes the game freak out.

It seems like too much happens at once. Interacting with the drop-down should first reveal the options then not activate a given item until a second interaction (second mouse click or hitting enter after selecting an option with the arrow keys)

May similarly affect the drop-down widget in general? See also #3581 and #3561 for potential clues

### How to reproduce

* Run game
* Go to video settings
* Click on the Display Mode drop-down (using tab plus arrow keys still work fine)

If other drop-downs are also affected it may be easier to test with those. After interacting poorly with the display mode drop-down i have to force-kill the game - it may destabilize your OS, beware!

No applicable logging, latest develop branch as of this write-up

### Computer details

Win10 with a 4k monitor in case it matters

RatMoleRat:
### Contains

Fixes #3587.

### How to test

Go to the display mode dropdown (inside the video settings menu) and click it with the mouse. The dropdown will now function as expected.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
@@ -110,7 +110,7 @@ public class UIDropdown<T> extends ActivatableWidget {
             for (int i = 0; i < optionListeners.size(); ++i) {
                 if (optionListeners.get(i).isMouseOver()) {
                     canvas.setMode(HOVER_MODE);
-                } else if (i==highlighted) {
+                } else if (i==highlighted && TabbingManager.focusedWidget != null && TabbingManager.focusedWidget.equals(this)) {
                     canvas.setMode(HOVER_MODE);
                     setSelection(getOptions().get(highlighted));
                 } else {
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Thanks for the fix! Tested and merged :-)

I have a tiny follow-up in mind but it is far less critical and maybe easy to approach whenever so I'm going to create a quick Good First Issue for it.

PR #3592, opened 2018-12-27
Fixes dropdowns when tabbing and allows tabbing to text entry boxes

RatMoleRat:
### Contains

I found that #3588 didn't fully resolve (or maybe caused?) some effects of the problem outlined in #3587, just in widget tabbing rather than using the mouse. As in, tabbing to and then opening the display mode dropdown caused the game to break. This fixes that and allows tabbing to text entry widgets, as well.

### How to test

- To test the dropdown fix: 
    - Tab to the display mode dropdown (inside the video settings menu), then press enter to expand it. Once it is open, press the arrow keys to move through the choices, and notice that the selection works.
- To test the text entry boxes: 
    - A good place to test is the filter on the module selection screen. Confirm that you can tab to it and that typing in it filters the modules, as it should.

- [ ] After changing the display mode, it takes two presses of `Enter` to activate the next widget by tabbing. Once another widget has been activated or deactivated once, future attempts to activate widgets seem to require just one press of `Enter`, as expected.


GooeyHub:

Cervator:
Thanks @RatMoleRat :-)

I merged this then submitted #3593 to cover the follow-up.

PR #3600, opened 2019-01-23
Fixed image scaling

### LINKED ISSUE #3598: UIImage scaling improvements
#3595 needed to be reverted due to widespread effects on UIImage scaling. However, it demonstrated that the current options for UIImage scaling, fitting and stretching, may not be sufficient for all use cases, and that other scaling modes (like filling) should be available.

Additionally, Cervator proposed that the decision between stretching and filling should be able to be made based on setting a limit to the amount an image is able to be stretched, after which the filling method is used instead. This would require a new scale mode, taking an additional parameter.

To do:
- [ ] Allow all scaling modes to be used with UIImage
- [ ] New scale mode: Context-sensitive stretch / fill

Vishwaak:
### Contains

This PR fixes #3598 

### How to test

Opening the game will help you understand the change in image scaling.


GooeyHub:

PR #3612, opened 2019-02-07
Make world generation random again

### LINKED ISSUE #3610: Worldgen seed not unique
I was using noise to generate my world based on the seed provided to my world generator. I assumed this was the seed shown in the world-setup screen.

### What actually happened
Instead of the seed shown in the world-setup screen, the method `BaseFacetedWorldGenerator .setWorldSeed()` got called with parameter `LOTR World-00`. This is supposed to be unique according to docs at `WorldPreGenerationScreen.createSeed()`, but it appears that the integer which gets appended is only incremented when remaining in the menu-area. When starting a world or restarting the game, it gets reset to 0. This makes the world seed be dependant on only the world-generator name and as this stays constant, the generated noise will be the same.
### How to reproduce
* Start the game and create a world using the Perlin generator from the Core module (I used my own generator but this seems to be the case for all generators). Don't touch the seed, maybe set some breakpoints in the setSeed() method of the FacetProviders or BaseFacetedWorldGenerator to check the integer or string you get passed. Let the world generate. 

* Generate a second world and use the same approach. 

This world (if the issue is an actual issue) should be the same as the first. Note that the seed which we saw in the menu was different.

### EDIT: clearer reproduction steps

1) Run the game (either from scratch or with additional modules)
2) Select a world generator (Perlin is a good choice because you immediately see the preview, allowing you to verify more easily)
3) Run and repeat.

TimVerhaegen:
### Contains

Fixes #3610, making the worlds generate randomly again.

### How to test

1) Use the reproduction steps from the issue to verify the bug exists for your workspace.
2) Use this branch and repeat the steps given in the issue. The generation should now be random. 

Create a world with the same world-generator twice and compare the result. (It should be a different world).

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/GameManifestProvider.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/GameManifestProvider.java
@@ -78,7 +78,7 @@ public class GameManifestProvider {
         String seed;
         if (universeWrapper.getTargetWorld() != null) {
             uri = universeWrapper.getTargetWorld().getWorldGenerator().getUri();
-            seed = universeWrapper.getTargetWorld().getWorldName().toString() + 0;
+            seed = universeWrapper.getTargetWorld().getWorldGenerator().getWorldSeed();
             gameManifest.setSeed(seed);
         } else {
             uri = config.getWorldGeneration().getDefaultGenerator();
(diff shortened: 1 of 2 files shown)


GooeyHub:

jellysnake:
Provided that this change tests out should be able to go ahead and merge it.

Cervator:
Tested out - can confirm it seems to do what it says it does!

I'm still wondering if we're handling the seed right in every case, but without multi-world even fully functional it probably isn't worth losing sleep over. For instance if I'm reading it right a quick start will simply set the seed to the universe seed, while an advanced start will involve both the universe seed _and_ the world name + increment. 

Anyway, it was indeed weird to see the same preview for two separate runs through the advanced flow. Thanks for catching and fixing that :-)

PR #3613, opened 2019-02-10
Visan

costinvisan:
The problem is that when in MenuControlSystem time.setPaused(true); was called the SwipeAnimation it will doesn't work, cause of course any animation work over time and how is that possible if time was paused. The solution was actually pretty simple just to let the menu system pass in an estimated delta time even if game is paused.


GooeyHub:

Issue #3553, opened 2018-11-23
Can't go back after enabling Animated Menus option in Video Settings


Graviton48:
### What you were trying to do
Just saw the `Animated Menus` option in Video settings while in-game and checked/ticked it to try it out.

### What actually happened
After Animated Menus option was enabled , the back button didn't seem to respond...It was not taking me to the Settings Screen (Audio settings, video settings , etc.)

### How to reproduce

* Step 1 Go to Video Settings in-game
* Step 2 Enable `Animated Menus` option
* Step 3 Press `Back` button

### Computer details
CPU- Intel Core-i3 3220
GPU- Nvidia Geforce GTX 1050 2GB DDR5
RAM- 8 GB DDR3
Windows- 10

Cervator:
Closing as fixed by #3615 ! Thanks all :-)

PR #3614, opened 2019-02-10
Issue #3553 solved. Can't go back after enabling Animated Menus

costinvisan:
#3553 
### The problem
In MenuControlSystem if networkMode is off the time is paused, so any animation cannot happen over time.
### Solution 
 Even if the time is paused we let the menu system pass in an estimated delta time(in SwipeAnimationSystem in case animDelta is 0 it is set it to 0.035f).

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/animation/SwipeMenuAnimationSystem.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/animation/SwipeMenuAnimationSystem.java
@@ -177,9 +177,9 @@ public class SwipeMenuAnimationSystem implements MenuAnimationSystem {
     public void update(float delta) {
         float animDelta = delta;
 
-        if (animDelta > 0.1f) {
+        if (animDelta > 0.1f || animDelta == 0.0f) {
             // avoid skipping over fast animations on heavy load
-            animDelta = 0.1f;
+            animDelta = 0.035f;
         }
 
         flyIn.update(animDelta);
--- a/engine/src/main/java/org/terasology/rendering/nui/animation/SwipeMenuAnimationSystem.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/animation/SwipeMenuAnimationSystem.java
@@ -56,6 +56,8 @@ public class SwipeMenuAnimationSystem implements MenuAnimationSystem {
 
     private float scale;
 
+    private static float PAUSED_ANIMATION_FRAME_TICK_TIME = 0.035f;
+
     /**
      * Creates default animations
      * @param duration the duration of the animation in seconds
@@ -181,14 +183,15 @@ public class SwipeMenuAnimationSystem implements MenuAnimationSystem {
             // avoid skipping over fast animations on heavy load
             animDelta = 0.1f;
         }
-        if(animDelta == 0.0f) {
-            // in case time is paused
-            animDelta = 0.035f;
+        if (animDelta < 0.0001f) {
+            // when we are inGameState and get to pauseMenu the time is paused
+            // so we need to fake the time is still running at a set rate
+            animDelta = PAUSED_ANIMATION_FRAME_TICK_TIME;
         }
 
         flyIn.update(animDelta);
         flyOut.update(animDelta);
-    }   
+    }
 
     @Override
     public Rect2i animateRegion(Rect2i rc) {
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

engine/src/main/java/org/terasology/rendering/nui/animation/SwipeMenuAnimationSystem.java
```diff
@@ -181,10 +181,14 @@ public void update(float delta) {
             // avoid skipping over fast animations on heavy load
             animDelta = 0.1f;
         }
+        if(animDelta == 0.0f) {
```
Cervator: Trivial code style issue: space between `if` and `(` - also may help with a newline between the two blocks.

Quirky float risk: comparisons can get weird with them, maybe do `animDelta < 0.0001f` instead?

engine/src/main/java/org/terasology/rendering/nui/animation/SwipeMenuAnimationSystem.java
```diff
@@ -181,10 +181,14 @@ public void update(float delta) {
...
+            // in case time is paused
```
Cervator: You could expand on this comment a bit with our discoveries about how the game being paused stops time and results in zeroed out delta values, so to make sure the animations still run we "fake" that time is still running at a set rate

engine/src/main/java/org/terasology/rendering/nui/animation/SwipeMenuAnimationSystem.java
```diff
@@ -181,10 +181,14 @@ public void update(float delta) {
...
+            animDelta = 0.035f;
+        }
 
         flyIn.update(animDelta);
         flyOut.update(animDelta);
-    }
+    }
```
Cervator: Trivial code style: extra spaces that serve no reason but to make the diff ever so slightly bigger

engine/src/main/java/org/terasology/rendering/nui/animation/SwipeMenuAnimationSystem.java
```diff
@@ -181,10 +181,14 @@ public void update(float delta) {
             // avoid skipping over fast animations on heavy load
             animDelta = 0.1f;
         }
+        if(animDelta == 0.0f) {
+            // in case time is paused
+            animDelta = 0.035f;
```
Cervator: It helps to avoid "magic numbers" and instead make a constant at the class level with a descriptive name, like `private static float PAUSED_ANIMATION_FRAME_TICK_TIME` or something

GooeyHub:

GooeyHub:

AndyTechGuy:
I tested it out and it works well for me 👍. Only two things to say:
1. The current animation constant makes the animation extremely fast on my computer, to the point that I cannot see the animation occur. Is this just for me or should the constant be lowered?
2. This is a great temporary solution, but a good task for the future would probably be replacing this with something that doesn't require a time estimation. For example: separating world updates (stops when paused) and UI updates (doesn't stop when paused), so the delta for animations doesn't go to zero when the world is paused.

engine/src/main/java/org/terasology/rendering/nui/animation/SwipeMenuAnimationSystem.java
```diff
@@ -56,6 +56,8 @@ public float getVertScale() {
 
     private float scale;
 
+    private static float PAUSED_ANIMATION_FRAME_TICK_TIME = 0.035f;
```
Cervator: It took inspecting more closely in source while testing but this line remains slightly off:

1. It should be at the top of the class declaration (a check for things being in a standard order)
1. It should be declared as `private static final float ...` to be a true constant, without which it triggers a naming convention warning

I'm just committing that during the merge since I moved it to the top of the class and added `final` already trying to make the IDE happy. Had me scratching my head for a moment why it was unhappy about the name, but it wasn't the name as much as the missing `final` :-)

Not a big deal at all, works fine just project best practices and all that!

I also added the following TODO and will write up a separate issue with details:

`// TODO: Convert away from a faked time interval to handling UI updates to non-game objects in a non-gametime loop?`

Cervator:
Tested out and am merging, thanks @costinvisan ! And @AndyTechGuy too for the review and ideas.

Added a todo and will write up a new issue for the enhancement to handle non-game UI components independent from game time :-)

Issue #3625, opened 2019-03-03
Resized game window resets to default after a game restart


Cervator:
### What you were trying to do

Stuff! Unrelated to the bug. Had resized my game window and expected the game to re-open with the window in the same position at the same size as before.

### What actually happened

Settings related to the size (and maybe position?) reset to default so I had to re-resize it.

Should be a fairly easy bug fix. Search around in the code base for "resize" or look at the related UI screens like `VideoSettingsScreen`

### How to reproduce

* Open Game
* Resize window
* Close and re-open game: no more resized window

PR #3627, opened 2019-03-03
Save window position and size on shutdown

mayant15:
Window size and position are saved to config.cfg
Fixes #3625 

### How to test

- Start the game
- Resize or move the window
- Restart the game, size and position should be saved

PR #3628, opened 2019-03-03
Window size saved and debugEnabled set false before shutdown

costinvisan:
Issue #3625 
**Contains**
Fixes for resizing the game window to default after the game is restarted. Also I thought it would be nice that the debug info should be disabled right before the game is shutted down.
**How to test**
- Open the game, resize the window(should be in windowed display mode). Shut the game down, open again and we should get same position and size.
- For debug info just get in any world enable debug system, close the game, reopen and get in any world. We can see that debug is disabled.


GooeyHub:

Cervator:
Heya @costinvisan - be careful with making assumptions, especially when you're still learning about the project :-)

It may _seem_ like a nicety to auto-disable debug text, but it might actually be part of some developers' workflow to run the game multiple times in succession where leaving debug text up is convenient.

For this PR that's a trivial bit of code to revert during merge, but if it was any larger it might hold up a PR that's otherwise ready. In this case #3627 has the same fix (and a second part of the fix) without the extra, so I've gone ahead and just merged that while I'll close this. Please don't feel bad and don't worry you still get "GSOC points" for having done this :-)

PR #3630, opened 2019-03-06
Extend tabbing to UISlider

mayant15:
Fixes the first problem mentioned in #3577 
Sliders can now be activated with Tab and used with the arrow keys.
This also fixes the activation persistence problem of sliders, i.e, holding down a slider and scrolling prevented the slider from deactivating.

### How to test
- Go to a menu with sliders (eg. Audio Settings)
- Navigate to a slider using Tab or Shift+Tab
- Increase its value using RIGHT/UP key, or decrease using LEFT/DOWN key

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
@@ -22,7 +22,13 @@ import org.terasology.input.events.MouseWheelEvent;
 import org.terasology.math.geom.Rect2i;
 import org.terasology.math.TeraMath;
 import org.terasology.math.geom.Vector2i;
-import org.terasology.rendering.nui.*;
+import org.terasology.rendering.nui.ActivatableWidget;
+import org.terasology.rendering.nui.BaseInteractionListener;
+import org.terasology.rendering.nui.Canvas;
+import org.terasology.rendering.nui.InteractionListener;
+import org.terasology.rendering.nui.LayoutConfig;
+import org.terasology.rendering.nui.SubRegion;
+import org.terasology.rendering.nui.TabbingManager;
 import org.terasology.rendering.nui.databinding.Binding;
 import org.terasology.rendering.nui.databinding.DefaultBinding;
 import org.terasology.rendering.nui.events.NUIKeyEvent;
@@ -171,7 +177,7 @@ public class UISlider extends ActivatableWidget {
             return DISABLED_MODE;
         }
 
-        if (active || (TabbingManager.focusedWidget != null && TabbingManager.focusedWidget.equals(this))) {
+        if (active || this.equals(TabbingManager.focusedWidget)) {
             return ACTIVE_MODE;
         } else if (tickerListener.isMouseOver()) {
             return HOVER_MODE;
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
```diff
@@ -16,26 +16,24 @@
...
 import org.terasology.math.geom.Rect2i;
 import org.terasology.math.TeraMath;
 import org.terasology.math.geom.Vector2i;
-import org.terasology.rendering.nui.BaseInteractionListener;
-import org.terasology.rendering.nui.Canvas;
-import org.terasology.rendering.nui.CoreWidget;
-import org.terasology.rendering.nui.InteractionListener;
-import org.terasology.rendering.nui.LayoutConfig;
-import org.terasology.rendering.nui.SubRegion;
+import org.terasology.rendering.nui.*;
```
pollend: we don't use .* operator when importing library. you can configure intellj to expand everything. 

https://odoepner.wordpress.com/2015/08/12/configure-intellij-to-use-default-eclipse-java-import-layout/

engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
```diff
@@ -173,14 +171,39 @@ public String getMode() {
...
+        if (active || (TabbingManager.focusedWidget != null && TabbingManager.focusedWidget.equals(this))) {
             return ACTIVE_MODE;
         } else if (tickerListener.isMouseOver()) {
             return HOVER_MODE;
         }
         return DEFAULT_MODE;
     }
 
+    private void changeValue(float delta) {
+        float newValue = TeraMath.clamp(getValue() + delta, 0, getRange() + getMinimum());
```
jellysnake: Wouldn't `getMinimum()` be more appropriate here than `0`.

engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
```diff
@@ -173,14 +171,39 @@ public String getMode() {
             return DISABLED_MODE;
         }
 
-        if (active) {
+        if (active || (TabbingManager.focusedWidget != null && TabbingManager.focusedWidget.equals(this))) {
```
jellysnake: If you do `this.equals(TabbingManager.focusedWidget)` then you don't need the bulky null check here, as `equals` returns false for null

engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
```diff
@@ -173,14 +171,39 @@ public String getMode() {
...
     }
 
+    private void changeValue(float delta) {
+        float newValue = TeraMath.clamp(getValue() + delta, 0, getRange() + getMinimum());
+        setValue(newValue);
+    }
+
+    @Override
+    public boolean onKeyEvent(NUIKeyEvent event) {
+        if (event.isDown() && TabbingManager.focusedWidget.equals(this)) {
```
jellysnake: See my above point about null checks. If you've got one above it's probably good to have one here.

engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
```diff
@@ -173,14 +171,39 @@ public String getMode() {
...
+            int keyId = event.getKey().getId();
+            if (keyId == Keyboard.KeyId.RIGHT || keyId == Keyboard.KeyId.UP) {
```
jellysnake: +1 for making it work on both up and right.

engine/src/main/java/org/terasology/rendering/nui/widgets/UISlider.java
```diff
@@ -173,14 +171,39 @@ public String getMode() {
...
+                this.changeValue(-1 * getIncrement());
+                return true;
+            }
+        }
+        return super.onKeyEvent(event);
+    }
+
+    @Override
+    public void onMouseWheelEvent(MouseWheelEvent event) {
+        event.consume();
```
jellysnake: Just curious of the logic behind consuming this event. I don't know the code well enough to work it out through a review

mayant15: As far as I understand, consume in supposed to stop the event from propagating after this widget receives it, right? I put it there because I wanted to end that event as soon as it started and not interfere with mouse button release.

Cervator:
Tests out - merging! Thanks :-)

PR #3632, opened 2019-03-11
changes for custom world name

shaunakdixit:
### Contains

Now world can have custom name

### How to test

create a new advance game 

- click on configure button after adding world.

- change name and click on done.

- click on continue and world pregeneration screen will appear

- click on configure button and here also you can change name 

- after changing name go back to universe setup screen and you can see changes has been reflected back

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
@@ -103,8 +103,8 @@ public class UniverseSetupScreen extends CoreScreenLayer {
     private Context context;
     private int worldNumber;
     private String selectedWorld = "";
-    public int indexOfWorld;
-    public WorldSetupWrapper customWorld;
+    public int indexOfSelectedWorld;
+    public WorldSetupWrapper copyOfSelectedWorld;
 
     @Override
     public void initialise() {
@@ -176,7 +176,7 @@ public class UniverseSetupScreen extends CoreScreenLayer {
             @Override
             public void set(String value) {
                 selectedWorld = value;
-                indexOfWorld = findIndex(worlds, selectedWorld);
+                indexOfSelectedWorld = findIndex(worlds, selectedWorld);
             }
         });
 
@@ -201,7 +201,6 @@ public class UniverseSetupScreen extends CoreScreenLayer {
         WidgetUtil.trySubscribe(this, "addGenerator", button -> {
             if (worldGenerator.getSelection().getUri().toString().equals("Core:heightMap")) {
                 getManager().pushScreen(MessagePopup.ASSET_URI, MessagePopup.class).setMessage("HeightMap not supported", "HeightMap is not supported for advanced setup right now, a game template will be introduced soon.");
-
             } else {
                 addNewWorld(worldGenerator.getSelection());
                 worldsDropdown.setOptions(worldNames());
@@ -249,7 +248,7 @@ public class UniverseSetupScreen extends CoreScreenLayer {
             worldsDropdown.setOptions(worldNames());
         }
         selectedWorld = "";
-        indexOfWorld =findIndex(worlds, selectedWorld);
+        indexOfSelectedWorld = findIndex(worlds, selectedWorld);
     }
 
     private Set<Name> getAllEnabledModuleNames() {
@@ -280,19 +279,19 @@ public class UniverseSetupScreen extends CoreScreenLayer {
     private void addNewWorld(WorldGeneratorInfo worldGeneratorInfo) {
         selectedWorld = worldGeneratorInfo.getDisplayName() + '-' + worldNumber;
         worlds.add(new WorldSetupWrapper(new Name(worldGeneratorInfo.getDisplayName() + '-' + worldNumber), worldGeneratorInfo));
-        indexOfWorld = findIndex(worlds, selectedWorld);
+        indexOfSelectedWorld = findIndex(worlds, selectedWorld);
         worldNumber++;
     }
 
     /**
-     * This method refresh the dropdown menu
-     * it is called from worldSetupScreen and also
-     * set selected world in dropdown Bar
+     * This method refreshes the worlds drop-down menu
+     * when world name is changed and updates variable selectedWorld.
+     * @param worldsDropdown
      */
-    public void refresh(UIDropdownScrollable worldsDropdown) {
+    public void refreshWorldDropdown(UIDropdownScrollable worldsDropdown) {
         worldsDropdown.setOptions(worldNames());
-        customWorld = worlds.get(indexOfWorld);
-        selectedWorld = customWorld.getWorldName().toString();
+        copyOfSelectedWorld = worlds.get(indexOfSelectedWorld);
+        selectedWorld = copyOfSelectedWorld.getWorldName().toString();
     }
 
     /**
@@ -331,11 +330,18 @@ public class UniverseSetupScreen extends CoreScreenLayer {
         }
     }
 
-    public int findIndex(List<WorldSetupWrapper> worldsList,String worldName) {
-        for(int i = 0 ;i<worldsList.size();i++) {
-            WorldSetupWrapper random = worldsList.get(i);
-            Name customName= random.getWorldName();
-            if(customName.toString().equals(worldName)) {
+    /**
+     * This method finds index of selected world from array list.
+     * and return -1 if unable to find it.
+     * @param worldsList
+     * @param worldName
+     * @return
+     */
+    public int findIndex(List<WorldSetupWrapper> worldsList, String worldName) {
+        for (int i = 0; i < worldsList.size(); i++) {
+            WorldSetupWrapper currentWorldFromList = worldsList.get(i);
+            Name customName = currentWorldFromList.getWorldName();
+            if (customName.toString().equals(worldName)) {
                 return i;
             }
         }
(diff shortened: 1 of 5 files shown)


GooeyHub:

shaunakdixit:
> Test it and everything works fine. The code nicely done.

thanks :-)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
```diff
@@ -103,6 +103,8 @@
     private Context context;
     private int worldNumber;
     private String selectedWorld = "";
+    public int indexOfWorld;
+    public WorldSetupWrapper customWorld;
```
Cervator: Maybe a better name here - is this meant to reflect a custom _name_ for a given world? All worlds are sort of custom themselves.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
```diff
@@ -103,6 +103,8 @@
     private Context context;
     private int worldNumber;
     private String selectedWorld = "";
+    public int indexOfWorld;
```
Cervator: This could perhaps also be named slightly better. `indexOfSelectedWorld` ?

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
```diff
@@ -246,6 +249,7 @@ public void onOpened() {
             worldsDropdown.setOptions(worldNames());
         }
         selectedWorld = "";
+        indexOfWorld =findIndex(worlds, selectedWorld);
```
Cervator: Trivial: Missing whitespace. Be sure to check the affected files with https://github.com/MovingBlocks/Terasology/wiki/Checkstyle - would have caught this one :-)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
```diff
@@ -276,9 +280,21 @@ private void recursivelyAddModuleDependencies(Set<Name> modules, Name moduleName
     private void addNewWorld(WorldGeneratorInfo worldGeneratorInfo) {
         selectedWorld = worldGeneratorInfo.getDisplayName() + '-' + worldNumber;
         worlds.add(new WorldSetupWrapper(new Name(worldGeneratorInfo.getDisplayName() + '-' + worldNumber), worldGeneratorInfo));
+        indexOfWorld = findIndex(worlds, selectedWorld);
         worldNumber++;
     }
 
+    /**
+     * This method refresh the dropdown menu
```
Cervator: For properly formatted Javadoc in an IDE just enter `/**<enter>` and it'll expand into a Javadoc block that includes any `return` or `param` tags etc. In this case you'll want to describe what `worldsDropdown` involves as well as indicate method-level javadoc (which is what you've got already). 

For that it tends to be better to describe the method in abstract - what is it _intended_ to be used for, not necessarily explicit places it is called from (which may change over time).

For instance: "Refreshes the worlds drop-down to catch any world name updates or a changed selection"

It may also be better to be more explicit than `refresh` - ideally get some reference to _what_ is being refreshed in the method name :-)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
```diff
@@ -315,6 +331,17 @@ public void setEnvironment(UniverseWrapper wrapper) {
         }
     }
 
+    public int findIndex(List<WorldSetupWrapper> worldsList,String worldName) {
```
Cervator: Needs javadoc (you can describe in the return tag what `-1` means for instance) and whitespace improvements.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
```diff
@@ -315,6 +331,17 @@ public void setEnvironment(UniverseWrapper wrapper) {
...
+        for(int i = 0 ;i<worldsList.size();i++) {
+            WorldSetupWrapper random = worldsList.get(i);
```
Cervator: Better variable name than `random` please - how is it random anyway? It is the specific world in each location in the list processed by the `for`

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
```diff
@@ -315,6 +331,17 @@ public void setEnvironment(UniverseWrapper wrapper) {
...
+            Name customName= random.getWorldName();
+            if(customName.toString().equals(worldName)) {
```
Cervator: `equals` or `ignoreCaseEquals` here? What happens if somebody tries to have two worlds that differ in just capitalization? What _should_ happen?

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldPreGenerationScreen.java
```diff
@@ -207,6 +211,13 @@ public void onOpened() {
         }
     }
 
+    /**
+     * Set selectedWorld when configure from WorldPreGenerationScreen
```
Cervator: Javadoc is great, but use the expanded version that covers all args/returns etc (missing javadoc here on `setNewName` - which is a poor name, you could simply flip it and call it `newNameToSet` for a better feel)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldSetupScreen.java
```diff
@@ -67,22 +70,42 @@
...
 
     @Override
     public void initialise() {
         setAnimationSystem(MenuAnimationSystems.createDefaultSwipeAnimation());
 
         WidgetUtil.trySubscribe(this, "close", button -> {
+            final UniverseSetupScreen universeSetupScreen = getManager().createScreen(UniverseSetupScreen.ASSET_URI, UniverseSetupScreen.class);
+            final WorldPreGenerationScreen worldPreGenerationScreen =getManager().createScreen(WorldPreGenerationScreen.ASSET_URI, WorldPreGenerationScreen.class );
+            UIText customWorldName = find("customisedWorldName", UIText.class);
+            if(customWorldName.getText().isEmpty()==false) {
```
Cervator: Minor whitespace style issues again

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldSetupScreen.java
```diff
@@ -67,22 +70,42 @@
...
+            universeSetupScreen.refresh(worldsDropdown);
+            worldPreGenerationScreen.setName(newWorldName);
             triggerBackAnimation();
         });
     }
 
+    /**
+     * This method sets the world name in title as well as in UITextBox
+     */
+    private void setName(UIText customWorldName) {
```
Cervator: Poor method name: if you are `set`ting the variable `customWorldName` the method should be called `setCustomWorldName` by convention.

Private methods don't _necessarily_ require javadoc, sometimes it can be useful if they're complex. Setters/getters usually don't justify it.

engine/src/main/java/org/terasology/rendering/world/WorldSetupWrapper.java
```diff
@@ -49,6 +49,8 @@ public Name getWorldName() {
         return this.worldName;
     }
 
+    public void setWorldName(Name newName) { this.worldName = newName;}
```
Cervator: Whitespace please (methods should be multi-line even if short)

Cervator:
Oh, did find one big problem: the world name field really does need better validation. Try to submit a world with an empty name. The field should probably trim any excess whitespace then test for least 1 character

Ideally it would also test for alphanumeric characters only? You can put total gibberish into the name field :-) That part I'd consider optional with an issue describing the need being an OK alternative

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
shaunakdixit: **copyOfSelectedWorld**  is it ok ?

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
shaunakdixit: It is not related to user.
this check the world name created by game like "perlin-0" or "flat-1" and return their indexes

shaunakdixit:
there was an issue i rectified that if name is empty

shaunakdixit:
it will change back to last updated or original name

GooeyHub:

Cervator:
Reviewed again, tested, and merged, thanks :-)

There were still a few trivial issues that probably weren't worth another review cycle so I tweaked them real quick - see https://github.com/MovingBlocks/Terasology/commit/12207a8a2265f44876cc8a94ef095fef404a0c8c for what I mean

Validation for world name could still be better (ideally you want to popup a warning when world name is empty, not just return as if it was changed, but then the original value remains. Will submit separate issue for that 👍

Issue #2535, opened 2016-10-12
Height change extras


nihal111:
With #2511 that enables change in Height and eyeHeight. There are a few bugs and extras that need to be dealt with:
- [x] Increase/Decrease jump speed according to size. Giants should be able to jump higher and dwarfs lower. 
- [x] Define a function or maybe just a proportionality constant (w.r.t height) for this.
- [ ] Increase/Decrease damage from fall. (30m giant should not die of a 15m fall). Define function/proportionality constant same as above.
- [x] Giants should take longer strides, increase runFactor?
- [x] Unable to pick up blocks off ground when height > 10. Bug!
- [x] Physics bug: Eye height translates gazeMountPoint to a value more than expected.
- [x] Trickier: Enable giants to be able to automatically climb up (no jump) blocks of smaller elevation.
- [x] Height too small, makes you fall through blocks.
- [ ] Potions should use a AffectHeightComponent instead of affecting the PlayerConfig settings. These should override the Default (config) values and should be serialized to the world save.
##

Cervator:
Throwing a bunch of assorted labels on this since it has multiple sub-items, and assigning @nihal111 to indicate work may be going on. It may be possible for additional contributors to work some of the items. Just be sure to coordinate if anybody else is interested as well :-)

TheFlash98:
Hey, I'd like to contribute to this issue @nihal111 😄

nihal111:
@TheFlash98 the crouch feature is buggy (not multiplayer ready) at this point. That needs to be changed first.
https://github.com/MovingBlocks/Terasology/pull/2961 has some discussion.
Maybe you could start with that. But that would require you to learn how character prediction and other things work in multiplayer mode.

TheFlash98:
Any help on where I could learn that?

EaguaireSama:
This seems to be an old issue ,can I work on this?

Cervator:
Hi @EaguaireSama - sure! Although just earlier today a user named Prince on Discord showed interest in it as well, unless that's just you under another name? In any case there are probably enough items here for several users, and some of them may take some digging first. Some items are fairly straight forward, others less so, may be ideal to discuss on chat and sort of (re)learn the current state of the relevant codebase together as parts are getting kind of dusty. I would also say some of our desires to refactor movement would tie into this.

Just be sure to dig around in the linked issues / PRs, maybe review and comment on #3644 specifically so we could see about merging that and continuing the items here on top :)

[...]

PR #3635, opened 2019-03-11
Adding additional features to Height change

sugam45:
Adds additional features to playerHeight as mentioned in #2535 .
With the increase of height player now gets enhanced jump height and speed as well as get immune to fall damage to a certain extent with the increase in height.

## **How to Test**
-  Start the game,
+ Write the command `playerHeight <value>` in the console.


GooeyHub:

Issue #3636, opened 2019-03-13
Better custom world name validation


Cervator:
Follow-up to #3632 which lets you enter a custom world name. Ideally we should attach more validation to such a custom name. Right now if you submit an empty value the name simply won't change, but there's no indication for that unless the user looks. Ideal setup would instead validate the value of the field looking for a few potential issues:

* Blank world name: Tell user to enter a name
* World name matches an existing world name (even with differing capitalization): Tell user name already used
* World name with gibberish in it: Maybe there are some special characters we should avoid? Yet we might want to support different language scripts for world name .... unsure about this one
* Does the world name fully apply to the world-specific _seed_ as well? I named two worlds the same and their world seeds varied slightly

Cervator:
Closing as complete (enough) with #3637 !

PR #3637, opened 2019-03-13
World name similarity checks

kinshukk:
### Contains

Added checks for similar world names. If a similarly named world is found, a popup is shown.

### How to test

1. Create game -> Advanced Setup -> Universe Setup
2. Add a couple of worlds
3. Rename first to, say, "blocksEverywhere" (by selecting world and going into Configure menu)
4. Try to rename second world to the same name.

- [ ] Haven't checked for special characters.
- [ ] Didn't quite get that different seed thing mentioned in #3636

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldSetupScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldSetupScreen.java
@@ -91,23 +91,12 @@ public class WorldSetupScreen extends CoreScreenLayer {
             } else if (customWorldName.getText().equalsIgnoreCase(world.getWorldName().toString())) {
                 //same name as before: go back to universe setup
                 goBack = true;
+            } else if (universeSetupScreen.worldNameMatchesAnother(customWorldName.getText())) {
+                //if same name is already used, inform user with a  popup
+                getManager().pushScreen(MessagePopup.ASSET_URI, MessagePopup.class).setMessage("Name Already Used!", "Please use a different name for this world");
             } else {
-                boolean worldNameMatchesAnother = false;
-
-                //check for case insensitive equality of world names
-                for (String worldName : universeSetupScreen.worldNames()) {
-                    if (worldName.equalsIgnoreCase(customWorldName.getText())) {
-                        worldNameMatchesAnother = true;
-                        break;
-                    }
-                }
-
-                if (worldNameMatchesAnother) {
-                    getManager().pushScreen(MessagePopup.ASSET_URI, MessagePopup.class).setMessage("Name Already Used!", "Please use a different name for this world");
-                } else {
-                    //no match found: go back to universe setup
-                    goBack = true;
-                }
+                //no match found: go back to universe setup
+                goBack = true;
             }
 
             if (goBack) {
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldSetupScreen.java
```diff
@@ -81,18 +81,48 @@ public void initialise() {
...
+
+                //check for case insensitive equality of world names
+                for (String worldName : universeSetupScreen.worldNames()) {
+                    if (worldName.equalsIgnoreCase(customWorldName.getText())) {
+                        worldNameMatchesAnother = true;
+                        break;
+                    }
+                }
+
+                if (worldNameMatchesAnother == true) {
```
RatMoleRat: You don't need the `==true` in this if statement - `if (worldNameMatchesAnother) {` does the same thing.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldSetupScreen.java
```diff
@@ -81,18 +81,48 @@ public void initialise() {
...
+                    getManager().pushScreen(MessagePopup.ASSET_URI, MessagePopup.class).setMessage("Name Already Used!", "Please use a different name for this world");
+                } else {
+                    //no match found: go back to universe setup
+                    goBack = true;
+                }
+            }
+
+            if (goBack == true) {
```
RatMoleRat: Same as above.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/WorldSetupScreen.java
kinshukk: Done :)

kinshukk: Checks passing now, artifactory was temporarily down.

Cervator:
@GooeyHub: add to whitelist

GooeyHub:

Cervator:
Tests out fine but I found one sneaky case not covered, and honestly I only just barely thought of it myself: If you add a world, rename it _to its increment + 1_ (as in, rename Perlin-1 to Perlin-2) then add a new world the new world ... is named the same as the prior one you just renamed :-)

@kinshukk are you up for finding where that new name is made up and adding a check there specifically to see if an existing world has been renamed to match the increment, and in that case keep incrementing it until its name is unique?

kinshukk:
Oops yeah I totally missed that one, will look into it.

GooeyHub:

GooeyHub:

shaunakdixit:
@kinshukk I would like to add one more thing. Don't consider increment of +1 it could be any number suppose if I rename world to **Flat-0** to **Flat-7** and then I add a new world **Flat-8** then You have to print name from **0 (Since 0 is free now we can use it again) to  6 then directly jump to 9**.
I believe best possible solution is to create a list of world number and check that number exist in list if it  is present in it then skip check next number.
https://github.com/MovingBlocks/Terasology/blob/12207a8a2265f44876cc8a94ef095fef404a0c8c/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java#L277-L285

kinshukk:
That might be ideal, but I think the point of having a world number is just so that the worlds have _unique_ names by default, not for having them in any particular order. Adding more checks for now available world numbers would add unnecessary complexity IMO.

shaunakdixit:
Thats totally up to cervator :-)

Cervator:
Yep for this issue I'm just hunting for world name uniqueness. Good catch tho @shaunakdixit !

Tested out, all looks well, merged! Welcome to the credits section @kinshukk and thanks :D

PR #3640, opened 2019-03-16
Create a Collision System

costinvisan:
### Contains
Collisions events handled by a CollisionSystem. Should work as berfore cause it was a pretty much the same thing as in HealthAuthoritySystem, just gotta send a DoDamageEvent.

### How to test
- get in singleplayer mode collide on vertical or horizontal fast enough to take some damage.

- [ ] Maybe we should do also a CollisionComponent

### DIFF
--- a/modules/Core/src/main/java/org/terasology/logic/actions/CollisionSystem.java
+++ b/modules/Core/src/main/java/org/terasology/logic/actions/CollisionSystem.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2019 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.logic.actions;
+
+import org.terasology.audio.StaticSound;
+import org.terasology.audio.events.PlaySoundEvent;
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.event.ReceiveEvent;
+import org.terasology.entitySystem.systems.BaseComponentSystem;
+import org.terasology.entitySystem.systems.RegisterMode;
+import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.logic.characters.CharacterSoundComponent;
+import org.terasology.logic.characters.CharacterSoundSystem;
+import org.terasology.logic.characters.events.HorizontalCollisionEvent;
+import org.terasology.logic.characters.events.VerticalCollisionEvent;
+import org.terasology.logic.health.DoDamageEvent;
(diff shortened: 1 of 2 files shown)


GooeyHub:

modules/Core/src/main/java/org/terasology/logic/actions/CollisionSystem.java
```diff
@@ -0,0 +1,112 @@
...
+import org.terasology.logic.health.EngineDamageTypes;
+import org.terasology.logic.health.HealthComponent;
+import org.terasology.math.geom.Vector3f;
+import org.terasology.registry.In;
+import org.terasology.utilities.random.FastRandom;
+import org.terasology.utilities.random.Random;
+
+
+@RegisterSystem(RegisterMode.AUTHORITY)
+public class CollisionSystem extends BaseComponentSystem {
```
Cervator: Some simple javadoc at the class level would be a nice addition :+1:

modules/Core/src/main/java/org/terasology/logic/actions/CollisionSystem.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @In
+    private org.terasology.engine.Time time;
```
Cervator: You can probably turn this into an unqualified reference and just have an import statement instead, that's more typical

modules/Core/src/main/java/org/terasology/logic/actions/CollisionSystem.java
```diff
@@ -0,0 +1,112 @@
...
+        if (speed > health.fallingDamageSpeedThreshold) {
+            int damage = (int) ((speed - health.fallingDamageSpeedThreshold) * health.excessSpeedDamageMultiplier);
+            if (damage > 0) {
+                entity.send(new DoDamageEvent(damage, EngineDamageTypes.PHYSICAL.get(), EntityRef.NULL, EntityRef.NULL));
+            }
+        }
+    }
+
+    /**
+     * Same as method onLand but now it's about horizontal collision
```
Cervator: Like noted on chat it would be nice to have each method documented independently (without a need to read anything else) then possibly just refer to the others as extras (can use `@link` and `@see` in Javadoc if you like)

modules/Core/src/main/java/org/terasology/logic/actions/CollisionSystem.java
```diff
@@ -0,0 +1,112 @@
...
+    private org.terasology.engine.Time time;
+
+    private Random random = new FastRandom();
+
+    /**
+     * React to any vertical collision of an entity that has a
+     * health component and establishes if the character should
+     * take damage and what's the amount
+     * @param event vertical collision
+     * @param entity
```
Cervator: Remember to add a little something document-y to each `@param` tag even if it seems obvious :-) Can help thin out the main description if the param tags are used well

modules/Core/src/main/java/org/terasology/logic/actions/CollisionSystem.java
```diff
@@ -0,0 +1,112 @@
...
+            if (damage > 0) {
+                entity.send(new DoDamageEvent(damage, EngineDamageTypes.PHYSICAL.get(), EntityRef.NULL, EntityRef.NULL));
+            }
+        }
+    }
+
+    /**
+     * For a horizontal collision we check if we can play
+     * the character sounds for landing
+     * @param event vertical collision
```
Cervator: Same with these param tags. For instance `@param event the event that triggered the vertical collision`

Issue #2753, opened 2017-01-11
Time out gracefully and return to menu when failing to spawn the player


Cervator:
At the moment it is possible under some conditions to fail to spawn the player, usually in the case of a bug affecting world gen causing all the chunk generator threads to die (so no world to spawn in!)

The symptom for that at present is simply stalling on the Create/Load world screen. We need to come up with a way to time out and return gracefully to the menu instead, presenting an error message to the user that no suitable spawn position could be found and that there may be something wrong with the world.

In #2750 related code was touched and an `Optional` prepared that could be checked as part of this.

To reproduce introduce a bug in a world generator that'll crash chunk generation threads on purpose.

Shouldn't be too terribly difficult.

zeokav:
@Cervator Won't timing out, in general, not fit for extremely slow machines (worst case (?)). 
Also, I finally got done with my internship so I'd like to start again with some more bite-size because I lost touch. 😛 
How do I mess with the world generator?

Cervator:
Probably there is a way to keep track of time and state to where you can tell if there is still world being generated (Chunk generation is still succeeding on occasion), and suppress / extend the timeout in that case :-)

To mess with world generation find a `Rasterizer` class and introduce an error in `generateChunk` - that way when the error is encountered it'll crash that ChunkGenerator thread while the game itself continues (so it'll stall trying to find a spawn point)

zeokav:
What I'm trying to do right now - 
Making generateChunk throw an exception in case there's a timeout while generating a chunk for which I use a Timer with maybe a 1000ms delay for each chunk (Couldn't figure out how else to check for non responsiveness). Also wanted to know if there's a wrapper for handling exceptions or if I can just throw them like I would in normal java. 
The error is handled by popping up a new screen telling the user what the problem was. 
I also wanted to know how to rollback to the menu undoing all that was done till the point of error.

Cervator:
That might not work since it is generateChunk causing the chunk generator threads to die, and there are a bunch of them. The timeout needs to be hooked up to the game loading progress itself (like whatever kicks off the progress bar after the player starts creating a world) and if chunks haven't become available (or no spawn position can be found) *then* time out and return to the main menu.

Since you would hook that up at the same place that triggers the menu state change probably there is an easier place there to return back to where you came from. I wouldn't want to try interacting with that from a chunk generator thread :-)

zeokav:
[...]

PR #3642, opened 2019-03-17
Graceful spawn fail when chunk loading times out

Loonride:
### Contains

I decided to take a go at this issue for my first PR: https://github.com/MovingBlocks/Terasology/issues/2753

I added a method in GameState to listen for new chunks, specifically so that StateLoading can keep track of when new chunks load and time out back to the main menu if a chunk does not load for a long period (Once chunks are actually supposed to start loading).

### How to test

Player should spawn without issues when everything goes as expected. If an error is introduced in the chunk generation process, the player will fail to spawn, and the game will time out back to the main menu.

To introduce an error in the chunk generation process:
Add `int num = 4 / (4 - 4);` at the start of the `generateChunk` method in BuilderWorldRasterizer: https://github.com/MovingBlocks/Terasology/blob/dbfe54ea07bcecabe90a1fae5bdd67e1e0380b38/modules/BuilderSampleGameplay/src/main/java/org/terasology/BuilderSampleGameplay/world/BuilderWorldRasterizer.java

Nothing must be done before merging, but it should be considered if 20000 is a good timeout time. The `onChunkLoaded` method I added in StateLoading could be used in the future to add more movement to the progress bar, if we can find a way to predict how many chunks will be loaded during initial load.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/modes/StateLoading.java
+++ b/engine/src/main/java/org/terasology/engine/modes/StateLoading.java
@@ -58,6 +58,7 @@ import org.terasology.engine.modes.loadProcesses.RegisterSystems;
 import org.terasology.engine.modes.loadProcesses.SetupLocalPlayer;
 import org.terasology.engine.modes.loadProcesses.SetupRemotePlayer;
 import org.terasology.engine.modes.loadProcesses.StartServer;
+import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.game.Game;
 import org.terasology.game.GameManifest;
 import org.terasology.network.JoinStatus;
@@ -67,6 +68,7 @@ import org.terasology.rendering.nui.NUIManager;
 import org.terasology.rendering.nui.internal.CanvasRenderer;
 import org.terasology.rendering.nui.internal.NUIManagerInternal;
 import org.terasology.rendering.nui.layers.mainMenu.loadingScreen.LoadingScreen;
+import org.terasology.world.chunks.event.OnChunkLoaded;
 
 import java.util.Queue;
 
@@ -87,6 +89,10 @@ public class StateLoading implements GameState {
 
     private int progress;
     private int maxProgress;
+
+    private static final long chunkGenerationTimeout = 20000;
+    private boolean chunkGenerationStarted;
+    private long timeLastChunkGenerated;
 
     /**
      * Constructor for server or single player games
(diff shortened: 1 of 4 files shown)


GooeyHub:

engine/src/main/java/org/terasology/engine/modes/StateLoading.java
```diff
@@ -88,6 +90,10 @@
     private int progress;
     private int maxProgress;
 
+    private static final long chunkGenerationTimeout = 20000;
```
Cervator: All caps on constants please, and toss it up to the top of the class :-) 

Default seems OK. Maybe there'd be a way to make it part of `config.cfg` or even expose it in the settings in case any players need a longer timeout?

engine/src/main/java/org/terasology/engine/modes/loadProcesses/LoadingChunkEventSystem.java
```diff
@@ -0,0 +1,42 @@
...
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.event.ReceiveEvent;
+import org.terasology.entitySystem.systems.BaseComponentSystem;
+import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.registry.In;
+import org.terasology.world.WorldComponent;
+import org.terasology.world.chunks.event.OnChunkLoaded;
+
+@RegisterSystem
+public class LoadingChunkEventSystem extends BaseComponentSystem {
```
Cervator: Would be helpful to add a bit of Javadoc here please :-)

engine/src/main/java/org/terasology/engine/modes/loadProcesses/LoadingChunkEventSystem.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @In
+    private Context context;
+
+    @ReceiveEvent(components = {WorldComponent.class})
+    public void onNewChunk(OnChunkLoaded chunkAvailable, EntityRef worldEntity) {
```
Cervator: And possibly here (javadoc)

Cervator:
@GooeyHub: add to whitelist

jellysnake:
@GooeyHub ok to test

Triggering bot and labelling PR

GooeyHub:

GooeyHub:

Loonride:
Opened new PR with cleaned up git: https://github.com/MovingBlocks/Terasology/pull/3645

PR #3644, opened 2019-03-18
Changes in other parameters with height change

sugam45:
### Height change

Adds additional features to playerHeight as mentioned in #2535 .
With the increase of height player now gets enhanced jump height and speed as well as get immune to fall damage to a certain extent with the increase in height.
### How to test

  +  Start the game,

   + Write the command playerHeight <value> in the console.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
+++ b/engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
@@ -239,9 +239,19 @@ public class MovementDebugCommands extends BaseComponentSystem {
         try {
             ClientComponent clientComp = client.getComponent(ClientComponent.class);
             CharacterMovementComponent move = clientComp.character.getComponent(CharacterMovementComponent.class);
+
+            Optional<Prefab> prefab = Assets.get(new ResourceUrn("engine:player"), Prefab.class);
+            CharacterMovementComponent moveDefault = prefab.get().getComponent(CharacterMovementComponent.class);
+
+            float jumpSpeedConst = 0.5f;
+            float runFactorConst = 0.1f;
+            float stepHeightConst = 0.05f;
             if (move != null) {
                 float prevHeight = move.height;
                 move.height = amount;
+                move.jumpSpeed = moveDefault.jumpSpeed + (move.height-moveDefault.height)*jumpSpeedConst;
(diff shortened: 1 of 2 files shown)


GooeyHub:

msteiger:
test this please

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -239,9 +239,19 @@ public String playerHeight(@Sender EntityRef client, @CommandParam("height") flo
...
+            Optional<Prefab> prefab = Assets.get(new ResourceUrn("engine:player"), Prefab.class);
+            CharacterMovementComponent moveDefault = prefab.get().getComponent(CharacterMovementComponent.class);
+
+            float jumpSpeedConst = 0.5f;
+            float runFactorConst = 0.1f;
+            float stepHeightConst = 0.05f;
             if (move != null) {
                 float prevHeight = move.height;
                 move.height = amount;
+                move.jumpSpeed = moveDefault.jumpSpeed + (move.height-moveDefault.height)*jumpSpeedConst;
```
msteiger: If the custom height is significantly lower than the default, the modification factor can be a "large" negative number. Are you sure that it cannot be larger than the default jump speed? The player would be unable to jump otherwise.

modules/Core/src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -231,9 +235,14 @@ public void onHeal(DoHealEvent event, EntityRef entity) {
...
         HealthComponent health = entity.getComponent(HealthComponent.class);
+        CharacterMovementComponent move = entity.getComponent(CharacterMovementComponent.class);
+
+        Optional<Prefab> prefab = Assets.get(new ResourceUrn("engine:player"), Prefab.class);
+        CharacterMovementComponent moveDefault = prefab.get().getComponent(CharacterMovementComponent.class);
+
         float speed = Math.abs(event.getVelocity().y);
 
-        if (speed > health.fallingDamageSpeedThreshold) {
+        if (speed > health.fallingDamageSpeedThreshold*Math.max(move.height/moveDefault.height,1.0f)) {
```
msteiger: I think that this should be moved to a separate line. Also, the damage needs to be adjusted to this variable threshold, not the original value..

GooeyHub:

skaldarnar:
Superseded by #3907

PR #3645, opened 2019-03-18
Graceful spawn fail on chunk load timeout: git cleaned up

Loonride:
### Contains

@Cervator Cleaned up git a bit as requested. Follow-up of previous PR which I will close: https://github.com/MovingBlocks/Terasology/pull/3642, solving https://github.com/MovingBlocks/Terasology/issues/2753.

I have added Javadoc to the new LoadingChunkEventSystem, and have moved the timeout time to system config. I told the user to check the log for more info if the failure occurs.

### How to test

As stated before:
Player should spawn without issues when everything goes as expected. If an error is introduced in the chunk generation process, the player will fail to spawn, and the game will time out back to the main menu.

To introduce an error in the chunk generation process:
Add `int num = 4 / (4 - 4);` at the start of the `generateChunk` method in BuilderWorldRasterizer: https://github.com/MovingBlocks/Terasology/blob/dbfe54ea07bcecabe90a1fae5bdd67e1e0380b38/modules/BuilderSampleGameplay/src/main/java/org/terasology/BuilderSampleGameplay/world/BuilderWorldRasterizer.java

2 Things I can still potentially add, if you would like:

- As the time since last chunk load increases (eg. every 5 seconds) the message changes with likelihood that something went wrong, indicating to the user that the game has not gone unresponsive. Also maybe a more detailed error message once user is sent back to home screen.
- Setting to change how long the timeout is (I am not sure which category in the settings this would fit under)

### DIFF
--- a/engine/src/main/java/org/terasology/config/SystemConfig.java
+++ b/engine/src/main/java/org/terasology/config/SystemConfig.java
@@ -30,6 +30,7 @@ public class SystemConfig {
     private boolean debugEnabled;
     private boolean monitoringEnabled;
     private boolean writeSaveGamesEnabled;
+    private long chunkGenerationFailTimeoutInMs;
     private String locale;
 
     public long getDayNightLengthInMs() {
@@ -90,6 +91,14 @@ public class SystemConfig {
 
     public void setWriteSaveGamesEnabled(boolean writeSaveGamesEnabled) {
         this.writeSaveGamesEnabled = writeSaveGamesEnabled;
+    }
+
+    public long getChunkGenerationFailTimeoutInMs() {
+        return chunkGenerationFailTimeoutInMs;
+    }
+
+    public void setChunkGenerationFailTimeoutInMs(long chunkGenerationFailTimeoutInMs) {
+        this.chunkGenerationFailTimeoutInMs = chunkGenerationFailTimeoutInMs;
     }
 
     public Locale getLocale() {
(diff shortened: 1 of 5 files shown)


GooeyHub:

Cervator:
Tested again and merging! Thanks so much :-) Added you to the game credits as well

Appreciate the config setting - that's probably enough really, it is more of a debug-level config item anyway.

Progress bar could be improved further, there's also an issue or two about letting some things in the menu run in a background thread so the UI doesn't lock up. Feel free to pursue if interested :-)

The only real issue I noticed while testing this was that I could provoke the game into not exiting gracefully after a timeout, at least in some cases. I suspect that's more pre-existing behavior as we have a bug report or two written up for that as well. Unsure if this aggravates it or not - some less than happy path flows back from some game state just seems able to forget a thread or two somewhere. Especially if something bad happens during multiplayer setup. Another thing to feel free to investigate further, may be a bit trickier though.

Loonride:
@Cervator Glad I could help! I will definitely have a look into those other issues as well. As for the situations where graceful timeout fails, is it happening before the AwaitCharacterSpawn step of the load? The timeout logic kicks in at that step, but maybe something is still going wrong on occasion before that point.

Cervator:
Nah, graceful timeout isn't failing at all. The entire game fails to exit cleanly under certain circumstances. That might be if unusual paths through the game state system are taken, something in the multiplayer system fails to dispose, etc :-)

Issue #3323, opened 2018-04-15
Rendering issue with layers drawn in the wrong order?


Cervator:
### What you were trying to do

Trying to test GooeysQuests and manually placing dungeons.

`give dungeonEntranceViaMountain` gives you a tool you can use to click on suitable terrain to spawn that initial StructureTemplate (ST) and additional ones behind it to form a dungeon. A wireframe preview of the ST should display, drawn over the terrain so you can actually see it (individual blocks deep into the terrain are outlined)

### What actually happened

The wireframe doesn't display correctly. It looks like it only shows where other parts of the UI would normally be drawn in the foreground anyway. See screenie below

This seems to be relatively recent, I think I saw it last Saturday during the play test, but am not quite sure.

### How to reproduce

* Start the game with the GooeysQuests module active (in the Perlin world)
* `give dungeonEntranceViaMountain` in the game console
* Hold that item and face a mountain, notice how weirdly the wireframe shows / doesn't show

![2018-04-15 00_58_46-terasology - c__dev_terasology_git_modulemegaworkspace_terasology - engine-te](https://user-images.githubusercontent.com/1063833/38775138-ca5f924e-4048-11e8-9d1c-0f96d459e364.png)

PR #3650, opened 2019-03-30
Overlay rendering order fix

Loonride:
### Contains

Fixes https://github.com/MovingBlocks/Terasology/issues/3323

In order to make sure that the outline shown in the issue renders on top, all the other world rendering Nodes must be processed first. This took a bit longer than it should have, but it taught me a lot about rendering and the DAG :)

### How to test

As described in the issue above:
- Start the game with the GooeysQuests module active
- `give dungeonEntranceViaMountain` in the game console
- See the wireframe render above everything else

I can update the renderer map sketchboard as well: https://sketchboard.me/dAEsEaNXcHvD#/.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
@@ -323,6 +323,10 @@ public final class WorldRendererImpl implements WorldRenderer {
         Node overlaysNode = new OverlaysNode("overlaysNode", context);
         renderGraph.addNode(overlaysNode);
         renderGraph.connect(finalHazeNode, overlaysNode);
+
+        renderGraph.connect(opaqueObjectsNode, overlaysNode);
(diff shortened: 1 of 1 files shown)


GooeyHub:

dave2s:
This seems like an easy fix, although it would be nice to verify other overlays after this change? Not sure if this can break something else, @vampcat? Integration to #3732 will be quick.

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -323,6 +323,10 @@ private void addWorldRenderingNodes(RenderGraph renderGraph) {
         Node overlaysNode = new OverlaysNode("overlaysNode", context);
         renderGraph.addNode(overlaysNode);
         renderGraph.connect(finalHazeNode, overlaysNode);
```
dave2s: Technically, we might not need this line anymore?

PR #3653, opened 2019-04-03
Fix Groovy Wrapper Logger Warning

arpitkamboj:
### Description

Groovy Wrapper currently gives a logger warning whenever it tries to run a git command internally. This PR fixes the warning by importing the sl4fj logger class into the common groovy script.

### How to test

Try running `groovyw module get <name>` with and without the fix. The warning should be gone after the fix. 

_Note_ - Make sure to delete the module before re-using the command because the script will skip the git commands and the fix will have no effect in that situation.
The warning is only shown when the groovy script uses the git module.

### Result

_**Before fixing**_

**_After fixing_**

### DIFF
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -1,5 +1,9 @@
 import groovy.json.JsonSlurper
 
+@Grab(group='org.slf4j', module='slf4j-api', version='1.6.1')
+@Grab(group='org.slf4j', module='slf4j-nop', version='1.6.1')
+import org.slf4j.*
+
 @GrabResolver(name = 'jcenter', root = 'http://jcenter.bintray.com/')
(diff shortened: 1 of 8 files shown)


GooeyHub:

Cervator:
Thanks for the fix! This will be nice to get out of the way :-)

Could you please remove the `import org.slf4j.*` line (tests out fine without) and rebase? Your commits are pretty messy for a 3 (should be 2) line change. Try avoiding merges and PRs in topic branches/forks - should almost always be able to do clean pulls to avoid merge commits.

@GooeyHub: add to whitelist

GooeyHub:

GooeyHub:

skaldarnar:
Merged in 6d1e5ed15d479005e35ad41432f60e4aeb80a9b9

PR #3658, opened 2019-04-06
Feature: Read shaders from source folder on recompile

Loonride:
### Contains

When the `recompileShaders` command was run in the game in the past, it would use the shader files in the build folder, not the source folder. This made it harder to develop and test shaders. This PR introduces a somewhat hacky method to force loading of shader files from the source folder upon recompile. It will not do this on initial compilation, since most users and players will not be using this feature, or possibly will not have the local `src` folder. If the `recompileShaders` command is run without the proper `src` folder, or with a file missing, loading from the source folder for that particular file will simply not happen.

### How to test

- Start the game
- Change something in the shader source folder
- run recompileShaders command and see the shader change

- I did not implement this for metadata, because it is such a small feature, unlikely to be useful when recompiling
- I considered making the loaded file info from the source folder overwrite the build folder files, but decided against it in case someone accidentally starts making changes in the build folder and realizes their mistake (functionally there is no difference between it being overwritten or not overwritten)
- The source file loading happens in the `ShaderData` class. The class was not designed for loading the data, but I think it is the best intermediary between the `GLSLShader` and the `GLSLShaderFormat`, which is usually responsible for the loading but I don't believe is reachable in any way to be used for this use case.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/assets/shader/GLSLShaderFormat.java
+++ b/engine/src/main/java/org/terasology/rendering/assets/shader/GLSLShaderFormat.java
@@ -31,9 +31,7 @@ import org.terasology.assets.format.AssetFileFormat;
 import org.terasology.assets.module.annotations.RegisterAssetFileFormat;
 import org.terasology.naming.Name;
 
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.Reader;
+import java.io.*;
 import java.lang.reflect.Type;
 import java.nio.file.PathMatcher;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
```diff
@@ -124,6 +124,19 @@ int linkShaderProgram(int featureHash) {
         return shaderProgram;
     }
 
+    @Override
+    public void nonInitialRecompile() {
+        // try to load the shaders from the source folder, not the build folder,
+        // on recompiles caused by the recompileShaders command
+        try {
+            shaderProgramBase.reloadFromSource();
+        }
```
dave2s: Checkstyle would like to see 
`} catch (IOException e) {`

engine/src/main/java/org/terasology/rendering/assets/shader/GLSLShaderFormat.java
```diff
@@ -31,9 +31,7 @@
 import org.terasology.assets.module.annotations.RegisterAssetFileFormat;
 import org.terasology.naming.Name;
 
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.Reader;
+import java.io.*;
```
dave2s: Your IDE probably did this. You should probably revert this change since importing whole io here is really not effective

GooeyHub:

GooeyHub:

GooeyHub:

Loonride:
I realized that changing the render options also calls `recompileAllShaders`, and we do not want that to be grabbing them from source like we want to when the `recompileShaders` command is used in the game. I renamed the new shader method to `developmentRecompile` to clearly indicate that this recompile should not be used unless the user explicitly wants to develop shaders. For that reason I added a `recompileAllShaders(boolean development)` method to differentiate the recompiling modes.

eviltak:
Thanks for your work @Loonride! @dave2s can you take a look at the latest changes after your last review, and merge this PR if it is good?

dave2s:
To recap: This PR enables to recompile shaders from source folder on runtime without rewriting already compiled shaders in the build folder. To do this it uses  2 aditional AssetDataFile variables in ShaderData object which it saves to and reads from instead of rewriting the build folder and reloading from there. Previous version of the command recompiled from build folder, which has runtime copy of said shaders. For me, I would maybe rename the `boolean development` to at least `boolean fromSource`. Other than that, seeing as this is for debugging only and the legacy code should behave the same, I'd just merge it as it seems to work as intended.

dave2s:
Continued by #3765 now

PR #3659, opened 2019-04-06
Fix: Reflect sky in water

Loonride:
### Contains

Sky was not being reflected in the water:
Terasology-190403002354-1366x706

This is because the sky sphere was not being culled from the correct side. The sky is now culled from the correct side in the relevant reflection node.

### How to test

- Run the game with global reflective water
- See some sky reflections in the water

None


GooeyHub:

engine/src/main/java/org/terasology/rendering/dag/nodes/BackdropReflectionNode.java
```diff
@@ -26,24 +26,14 @@
...
 import org.terasology.rendering.opengl.FBO;
 import org.terasology.rendering.opengl.FBOConfig;
 import org.terasology.rendering.opengl.fbms.DisplayResolutionDependentFBOs;
 import org.terasology.rendering.world.WorldRenderer;
 
-import static org.lwjgl.opengl.GL11.glCallList;
-import static org.lwjgl.opengl.GL11.glEndList;
-import static org.lwjgl.opengl.GL11.glGenLists;
-import static org.lwjgl.opengl.GL11.glNewList;
+import static org.lwjgl.opengl.GL11.*;
```
pollend: we don't use the start import '*' operator. 

https://kb.novaordis.com/index.php/IntelliJ_Disable_Star_Imports

Loonride: Thanks for pointing that out. I know I read on some Terasology style guide that we should let IntelliJ do the default '*' at 5, so I guess whatever I found was outdated.

GooeyHub:

GooeyHub:

msteiger:
I think that imports have been updated. The actual code changes seem to be straightforward, but I haven't verified the fix.

jellysnake:
Tested this out. Compiles fine, doesn't crash. I just, can't really notice a difference.
I suspect this may be that I'm actually not sure what I'm supposed to be looking for.

Here's an example, I took 2 screenshots (not through the in game feature) and compared them side by side.

| One | Two |
| ---   |   --- |
|[Without](https://i.imgur.com/Y0xT5uq.jpg) | [With](https://i.imgur.com/qinH7lp.png) |

If you look at the two, it's hard to tell which is which, but the one on the left, _'one'_ is without this PR and the one on the right, 'two' is with this PR.

Can you provide me an example screenshot or how I should be testing this to make sure it works?

pollend:
I remove the water effect so this could be more clear. 

with `addDesiredStateChange(new SetFacesToCull(GL_BACK));`
image

with `addDesiredStateChange(new SetFacesToCull(GL_FRONT));`
image

code left as is.
image

not sure why you want to cull the sky from the water reflection?

pollend:
one thing that might be a bug but it's pretty well hidden with the refraction shading effect. you can see the backsides of the culled geometry from the land.

image

PR #3662, opened 2019-04-21
Update Modules.md

casals:
### Contains

Small documentation fixes on Modules list:

- Corrected link for GooeyDefence (originally listed as GooeyDefense). 
- Included FlexibleMovement in the modules list.

### How to test

- Go to the Modules list
- Check if the new GooeyDefence link is working properly
- Check if the new FlexibleMovement link is working properly

- [x] Check if it makes sense to include FlexibleMovement as a listed module.
- [x] Check if instead of fixing the link, renaming the GooeyDefence repository to make it grammatically correct makes more sense.
- [x] Consider removing the following entries from the Modules list: (i) Core and CoreSampleGameplay (since they actually don't exist as repositories and can't be installed using Gradle); and TutorialGraphicTweaks (since it's striked out in the original list but the link is still working).

### DIFF
--- a/docs/Modules.md
+++ b/docs/Modules.md
@@ -63,13 +63,14 @@ Here's a list of modules bundled with the game by default (this line-up will cha
 * [Equipment](https://github.com/Terasology/Equipment) - an equipment system with a character screen (`c`) showing stats and slots for various equipment types
 * [EventualSkills](https://github.com/Terasology/EventualSkills) - a time-based skill system, a bit akin to EVE Online's system
 * [Fences](https://github.com/Terasology/Fences) - fences!
+* [FlexibleMovement](https://github.com/Terasology/FlexibleMovement) - a movement implementation for FlexiblePathfinding (see below)
 * [FlexiblePathfinding](https://github.com/Terasology/FlexiblePathfinding) - a more flexible variant pathfinding framework
 * [FlowingLiquids](https://github.com/Terasology/FlowingLiquids) - a more advanced flowing liquid system than SimpleLiquids - treats liquids as finite, makes them propagate, and current will even drag you around!
 * [Fluid](https://github.com/Terasology/Fluid) - adds support for fluid in non-world situations (such as for storage in workstations)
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Thanks!

On FlexibleMovement the general condition for inclusion here is that the module is in the "stable lineup" for modules and gets shipped with the game. However, there are exceptions like modules that go unstable and get removed, ideally temporarily. And I know we want to put FM to work. Have you looked at it to check out how broken it is? It does go well with FlexiblePathfinding so it would be nice to have both available. With a disclaimer on the line here (may not be stable / included yet) I'm happy to have it described

On GooeyDefence - yeah spelling sometimes varies, especially for things like "initialize" vs "initialise" - generally whichever spelling comes first tends to win ... 😁 

Core / CoreSampleGameplay - should probably stay since they're available in the game and thus selectable and descriptions are handy, although we should make it clear they're currently embedded in the engine repo (as is BuilderSampleGameplay for that sake). Over time they might get extracted out, like the Inventory module just now (was pulled out of Core but now lives in its own repo)

TutorialGraphicTweaks - yeah that one is so old at this point any updated rendering tutorial work is likely to start from scratch. Feel free to delete it from the list and I'll delete / move the repo as well :-)

casals:
On FlexibleMovement: I was just checking the code and it will require some work. All *Node classes depend on org.terasology.logic.behavior.tree.Node/Status/Task (which I imagine have long been deprecated). I'll include the unstable disclaimer and raise an issue related to it (I'll work on it in the next days).

On GooeyDefence / Core / CoreSampleGameplay: OK, thanks for clarifying it 😄 

On TutorialGraphicTweaks: I'll remove it from the list.

I'll amend my original commit with these changes.

Cervator:
Thanks! Tested out the two links and merged :-)

I moved the graphics repo to the GooeyTests org. In theory its link should still work but redirect to the new resting place, but this way it'll stay out of search results in the main org and such. Should probably do the same for other ancient broken modules.

Also added you to the credits, even little things count 👍

PR #3666, opened 2019-04-27
Fixed OpenGL line width warning

### LINKED ISSUE #3665: OpenGL Line Width Warning
When hovering over anything that creates a hitbox outline, the following error gets printed out every tick:

`[main] WARN  OpenGL - [api] [deprecated] API_ID_LINE_WIDTH deprecated behavior warning has been generated. Wide lines have been deprecated. glLineWidth set to 2.000000. glLineWidth with width greater than 1.0 will generate GL_INVALID_VALUE error in future versions`

This is due to several `glLineWidth` calls where the parameter is 2 (unsurprisingly). This means the fix is very simply just changing these calls to use 1 instead.

PAndaContron:
This PR fixes #3665. It is a very simple fix, since there are only 3 places where a method call leads to a call to `glLineWidth`. All of these calls have been modified to have a line width of only 1. This should prevent any warnings from OpenGL.

Also, it seems like Eclipse automatically replaced all of the tabs with spaces (which shouldn't be an issue since most of the project uses spaces instead anyway), and reordered the imports. I don't think these should be major issues though.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/LocalPlayerSystem.java
@@ -15,6 +15,8 @@
  */
 package org.terasology.logic.players;
 
+import java.util.List;
+
 import org.terasology.assets.ResourceUrn;
 import org.terasology.config.Config;
 import org.terasology.engine.SimpleUri;
@@ -78,8 +80,6 @@ import org.terasology.world.WorldProvider;
 import org.terasology.world.block.Block;
 import org.terasology.world.block.BlockComponent;
 import org.terasology.world.block.regions.BlockRegionComponent;
-
-import java.util.List;
 
 // TODO: This needs a really good cleanup
 // TODO: Move more input stuff to a specific input system?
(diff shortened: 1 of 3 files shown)


GooeyHub:

msteiger:
test this please

msteiger:
If the default line width is one and it's never changed to a different value, then those calls should not be necessary. So I assume you could just remove them ..

GooeyHub:

Cervator:
Hey @PAndaContron and thanks for the PR :-)

The whitespace & import changes actually _are_ problematic as they'll keep toggling back and forth between the preferred Eclipse & IntelliJ conventions. There are some style related files under `config/eclipse` you should be able to grab that'll make Eclipse follow the conventions from IntelliJ. Could you try to apply those and shrink the diff to avoid the style changes please?

PAndaContron:
Would the whitespace need to be changed back too? I thought Terasology used spaces instead of tabs, that's pretty much the only reason why I have that option enabled in Eclipse.

Cervator:
I don't actually see any tab characters at all in the diff (not sure if my spot check just missed them?), just that Eclipse has a different indentation _guide_ for `switch` blocks and such, so some things are at different indentation levels :-) And yes please try to revert/reformat those bits if able.

PAndaContron:
Got it, so should I just add a new commit to fix the formatting or something else?

PAndaContron:
That should fix it!

Cervator:
Thanks :-) Looks good to me from a style etc perspective, tagging a few potential 3d wizards who could perhaps also weigh in on the functionality. Haven't personally tested yet either.

Cervator:
Actually can't request review from @dave2s yet - sent an org invite so we can do that more properly :-)

PAndaContron:
@msteiger I think it would make more sense to leave them in just for clarity. I doubt it has a significant performance impact, and it helps to clarify that this code requires the line width to be 1. Also, in the future the line width may be changed by other methods, even though it is not right now, which would then change all of these line widths as well (this is, I assume, why the call happens every frame and not just once).

dave2s:
Another possibility might be to query for supported line widths and act accordingly. I'll try it

vampcat:
I totally agree with msteiger here. Since the call is deprecated, it makes sense to remove them rather than modify them. Add in the fact that using the calls does nothing either way, and you have a perfect reason to remove them.

dave2s:
> I totally agree with msteiger here. Since the function is deprecated, it makes sense to remove the calls rather than modify them. Add in the fact that using the calls do nothing either way, and you have a perfect reason to remove them.

What do you mean the calls do nothing either way? They do something if your vendor supports it, it's been probably always in this state - no guaranteed support. 
image

PAndaContron:
Can confirm that the fix by @dave2s works, and probably better fix than this one.

msteiger:
Is there a PR for the approach by @dave2s ? I would do the check for supported width only once since it never changes during runtime and store it somewhere, maybe even in a singleton-style fashion, so everyone can use it. Otherwise you would allocate the FloatBuffers every frame in every renderer. Also, I find it a bit odd to render it with different thickness depending on something that is not obvious to the player.

To sum it up, I would just go with a line width of 1 everywhere and avoid the complexity, but it's not my call, of course.

dave2s:
There isn't a PR. I just wanted to test if this could work and give another potential option. As you are saying, one would have to do the efficiency changes. I can make a PR if you want, but seeing as everyone is kind of univoiced on the total removal, I don't see a point there anymore.
Also, I have to agree with this:
>  Also, I find it a bit odd to render it with different thickness depending on something that is not obvious to the player.

PAndaContron:
Possibly line thickness could be added to the video settings, with the warning that some systems may not support thick lines. The default would be set by checking the max supported line width, then the user could change it from there. It's worth noting that, on my system at least, even though thick lines aren't supported, they're still displayed, so if the user is willing to put up with OpenGL warnings they can still use thicker lines.

jellysnake:
It seems to me there are two options:
1. Just remove the calls, and use the default width of `1`
2. Query to find supported widths, and use `1` or `2` depending on if the latter is supported.

Option one is the simplest, but also the least flexible. Option 2 has the advantage of allowing other systems to use thicker lines than this one, but is more complex.

I personally vote going with updating this PR to use option one. It's simpler and I'm unsure of if the added benefits of checking for line widths would ever be used. We can add a link/ref to dave2s' code on the issue so it's there for reference, should it come up in future

@PAndaContron @dave2s @vampcat does anyone have any objections?

Issue #3668, opened 2019-04-29
Gameplay Template scrollbar exits to main menu when clicked, instead of being able to drag and scroll


Qwertygiy:
### What you were trying to do
Clicking the knob beside the Gameplay Template options in the main menu to drag it and scroll it up and down between choices.

### What actually happened
The list of options closed as soon as the knob was clicked, rather than allowing itself to be dragged.

### How to reproduce

* Step 1 Click singleplayer game.
* Step 2 Open Gameplay Template list with the [v] button.
* Step 3 Click on the round scroll knob on the right side of the list.

![Screenshot_20190428_205937](https://user-images.githubusercontent.com/2421117/56872579-8a068600-69f8-11e9-9e06-7ef040eaa223.png)

### Log details and game version
TerasologyOmega, 2.2.0, default installation.

### Computer details
Model: Toshiba Satellite C55D-A5206
OS: 64-bit Ubuntu
OS Version: 4.15.0.47-generic (as Kubuntu 18.04)
CPU: AMD A4-5000 APU
GPU: Radeon integrated graphics
Driver: X.org AMD KABINI (DRM 2.50.0, LLVM 7.0.0)
JVM: OpenJDK 1.8.0
OpenGL: 4.4 (Compatibility Profile) Mesa 18.2.2

Qwertygiy:
Further information:

- The NUI Editor dropdown has the same issue.
![image](https://user-images.githubusercontent.com/2421117/56873633-ae1a9500-6a01-11e9-9e00-9107f352cf2c.png)

- The Video Settings scrollbar does *not* have the same issue.
![image](https://user-images.githubusercontent.com/2421117/56873641-c5598280-6a01-11e9-91ee-a97e484de894.png)

My initial guess with this information is that the scrollbar knob is not properly capturing the mouse click before the options in the dropdown list, as if the knob is either "behind" the options; a lower priority than the options; or not "interrupting" the click's progress to prevent it from reaching the options.

profgrammer:
The universe setup screen in the advanced setup also has the same problem. Clicking a location on the scroll bar makes the menu disappear, but on reopening it we can see that the knob is at the position we clicked on before. 
![worldgen](https://user-images.githubusercontent.com/25369498/57569422-6ee14080-7412-11e9-9ea4-fba17d4e3f74.png)

AlexandrSunLight:
Same in the languge select menu. 
[...]

PR #3676, opened 2019-05-15
Minor changes in dropdown menus

AlexandrSunLight:
When I checked #3668 I saw that keys didn't work in dropdown menus at all. After that I found out, that TabbingManager.focusedWidget mention in UIDropdown and UIDropdownScrollable classes creates NullPointerException.

### Contains
PR fixes: 
-highlight of menu options with mouse or keyboard when widget is open
-change options with keyboard when widget is closed
-"enter" key event added (it chooses highlighted option)

### How to test
Open video settings or language settings and check out a dropdown widgets. "Up", "Down", "Left", "Right" and "Enter" keys should work. Options should be highlighted when cursor is over it AND when cursor is somewhere else.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
@@ -26,7 +26,6 @@ import org.terasology.rendering.nui.BaseInteractionListener;
 import org.terasology.rendering.nui.Canvas;
 import org.terasology.rendering.nui.InteractionListener;
 import org.terasology.rendering.nui.SubRegion;
-import org.terasology.rendering.nui.TabbingManager;
 import org.terasology.rendering.nui.databinding.Binding;
 import org.terasology.rendering.nui.databinding.DefaultBinding;
 import org.terasology.rendering.nui.events.NUIKeyEvent;
@@ -110,7 +109,8 @@ public class UIDropdown<T> extends ActivatableWidget {
             for (int i = 0; i < optionListeners.size(); ++i) {
                 if (optionListeners.get(i).isMouseOver()) {
                     canvas.setMode(HOVER_MODE);
-                } else if (i==highlighted && TabbingManager.focusedWidget != null && TabbingManager.focusedWidget.equals(this)) {
+                    highlighted = i;
+                } else if (i==highlighted) {
                     canvas.setMode(HOVER_MODE);
                 } else {
                     canvas.setMode(DEFAULT_MODE);
@@ -137,7 +137,7 @@ public class UIDropdown<T> extends ActivatableWidget {
     public String getMode() {
         if (!isEnabled()) {
             return DISABLED_MODE;
-        } else if (opened || (TabbingManager.focusedWidget != null && TabbingManager.focusedWidget.equals(this))) {
+        } else if (opened) {
             return ACTIVE_MODE;
         }
         return DEFAULT_MODE;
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
```diff
@@ -110,7 +109,8 @@ public void onDraw(Canvas canvas) {
             for (int i = 0; i < optionListeners.size(); ++i) {
                 if (optionListeners.get(i).isMouseOver()) {
                     canvas.setMode(HOVER_MODE);
-                } else if (i==highlighted && TabbingManager.focusedWidget != null && TabbingManager.focusedWidget.equals(this)) {
+                    highlighted = i;
+                } else if (i==highlighted) {
```
Cervator: Minor whitespace issue here, could you grab a Checkstyle IDE plugin to validate that your changes follow our code guidelines please? :-)

Trivial but good to keep clean!

engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
```diff
@@ -264,31 +264,31 @@ public void changeHighlighted(boolean increase) {
...
-                if (opened) {
-                    setOpenedReverse(false);
-                }
-                return true;
-            } else if (keyId == Keyboard.KeyId.RIGHT) {
-                if (!opened) {
-                    setOpenedReverse(false);
-                }
-                return true;
+            switch (keyId){
```
Cervator: Another trivial missing space, but I'll leave any additional ones unremarked on as Checkstyle should catch them all 👍

engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdownScrollable.java
```diff
@@ -312,44 +312,6 @@ public void setOpenedReverse(boolean selectionSet) {
         }
     }
 
-    public void changeHighlighted(boolean increase) {
```
Cervator: Something in this removal might have removed the intended yellow highlight box around a UI widget that has been selected by tabbing to it?

Cervator:
@GooeyHub: add to whitelist

(the LGTM check is broken due to a known issue, no big deal)

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
```diff
@@ -137,7 +137,7 @@ public Vector2i getPreferredContentSize(Canvas canvas, Vector2i areaHint) {
     public String getMode() {
         if (!isEnabled()) {
             return DISABLED_MODE;
-        } else if (opened || (TabbingManager.focusedWidget != null && TabbingManager.focusedWidget.equals(this))) {
```
RatMoleRat: The removed bit is important! It needs to know if the dropdown is the focusedWidget (in other words, if it is the current selection by tabbing) in order to highlight the widget with yellow - that's what returning `ACTIVE_MODE` does. You can probably find another workaround to avoid using `TabbingManager`, but you need some way to tell if it's been tabbed to 🙂

RatMoleRat: This line is important - `ACTIVE_MODE` is what highlights the widget when it is tabbed to. Without checking to see if it's the focused widget (that is, if it's the currently tabbed-to widget), it won't be highlighted when selected by tabbing (unless it's opened). You might be able to find a work-around to avoid using `TabbingManager`, but you still need something to determine if it should be highlighted.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdownScrollable.java
RatMoleRat: IIRC `highlighted` is actually for which item in the dropdown is currently selected (via tabbing or otherwise) because that gets a box around it as well. I think the culprit is in the parent class - `UIDropdown` - where it determines if the widget is in active mode or not.

RatMoleRat:
It looks good and I don't have any suggestions about code changes except for the highlighting, which I added a comment about how to fix. You could *maybe* leave the highlighting code as-is (if it doesn't throw the exception) or just try to find a work around.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
AlexandrSunLight: Got it! Btw, I've spotted, that TabbingManager resets every time when mouse clicking or mousewheel turns (at processMouseClick method of nui\internal\CanvasImpl class). Not an issue, just the matter of opinion. Do you think, that it is more comfortable to start over every time widget gets inactive or more convinient will be highlight the next one after tabbing?

GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/widgets/UIDropdown.java
RatMoleRat: Without looking over that code section again (so I might be mistaken), I would think that it is meant to keep it sort of consistent - if you click somewhere else, tabbing shouldn't continue from that random spot that it was at before the click. I guess the way to deal with it at the moment is using the method that you mentioned, so that the player knows that tabbing will start from the beginning of the widgets. If that plan were to change, it should end up starting tabbing from the freshly clicked widget.

jellysnake:
Thanks @AlexandrSunLight for the PR (Especially as it wasn't linked to an issue)!

PR #3678, opened 2019-05-21
Arrays for ui text

### LINKED ISSUE #3629: Allow UI text to be specified in an array
Presently, multiline text has to be written in ui files on a single line, potentially resulting in absurdly long lines. The common work around to this is to specify text as an array of strings to be joined by newlines. This should be implemented for UIText, and potentially other widgets.

profgrammer:
### Contains

Fixes #3629 

### How to test

`UIText` elements now have a new attribute: `linesOfText` which is an array of strings. Programatically, use the function `setLinesOfText()` to set the text array. Also, setting the text array will render the `text` attribute to be an empty string, and vice versa, since it is logical to use only one of the two at any given point. 

- [ ] Some cases where `text.set()` is being used instead of `setText()`. Don't know if I should change those instances to `setText()`.


GooeyHub:

eviltak:
@GooeyHub: ok to test

From first impressions, in the future I'd look at renaming `textArr` to `textArray`. Abbreviations like `arr` simply make code harder to read without too much of a productivity advantage, especially with autocomplete :)

I'd also use a more informative name like `lines` or even `linesOfText` (or something similar) which better conveys what the contents of the field are actually used for.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
```diff
@@ -667,7 +674,12 @@ public void bindText(Binding<String> binding) {
      * @return The text contained by the text box
      */
     public String getText() {
-        return text.get();
+        String arrayText = "";
+        if(textArr != null){
+            for(int i = 0;i < textArr.length;i++) arrayText += textArr[i] + "\n";
```
eviltak: I'd replace this with a String join call -- it's more robust and is easier to read.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
```diff
@@ -667,7 +674,12 @@ public void bindText(Binding<String> binding) {
...
+        }
+        if(text.get() != "") return text.get() + "\n" + arrayText;
+        return text.get() + arrayText;
```
eviltak: I don't know if this behavior (of concatenating `text` and `arrayText` in that order) should be standard or not (I don't think it should because IMO _either_ `textArr` or `text` should be used at a given time), but if it is, it should be documented somewhere.

eviltak: Also, this will cause an NPE given that setting the text array sets `text` to null.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
```diff
@@ -680,6 +693,7 @@ public void setText(String val) {
         boolean callEvent = !prevText.equals(val);
 
         text.set(val != null ? val : "");
+        linesOfText = null;
```
jellysnake: like above, i'd prefer a blank entry over null

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
```diff
@@ -202,11 +206,13 @@ public void onDraw(Canvas canvas) {
...
     }
 
+
     protected void drawAll(Canvas canvas, int multilineWidth) {
+
         if (text.get() == null) {
             text.set("");
         }
-        if (text.get().equals("")) {
+        if (text.get().equals("") && linesOfText == null) {
```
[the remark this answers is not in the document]
jellysnake: This is still outstanding

jellysnake:
I'm going to merge them in spite of the ongoing discussion on the associated issue about alternative ways to implement this.
When, and indeed if, an alternative method is attempted that self same PR can revert the changes in this PR

eviltak:
Can we revert this (and maybe #3689) given that #3698 is merged?

PR #3679, opened 2019-05-25
Add try-catch to updateItem

jellysnake:
Adds a try catch to the update item call.
This allows the exception to be caught, and the item to be skipped.

# Testing
Break a module's git in some way so that `groovyw module update-all` doesn't work
Then verify that the module is skipped and the rest of the command works fine.


GooeyHub:

GooeyHub:

Cervator:
Coincidentally just had to update a mega-workspace on a rarely used system. Grabbed this and then #3531 while I was at it so I could take advantage plus test while doing it (other PR turned into a big of a saga, but it worked out!). So I can conform this as good, merged! Thanks :-)

PR #3682, opened 2019-05-29
LightShafts calculated from the Moon during night time and with half exposure.

dave2s:
### Contains
Quick tweak for sun shafts being present during night, which looks really bad.
This PR presents a quick solution remapping light shafts' source position to the moon with lowered exposure during night. An overlap instead of a switch would be nicer, or perhaps to just shorten the interval for moonshafts, start with sun a little bit more under the horizon.

### How to test
- Turn on light shafts in the video settings
- Open console '~', write setWorldTime [time], where time is any float in *days*. 
                    0,1,2 -> midnights;  0.5,1.5,2.5 - noons;   
- Verify that with the beginning of night the shafts start to appear from the moon and stop appearing from the sun under the horizon. And vice-versa

[ ] - Nothing yet

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/LightShaftsNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/LightShaftsNode.java
@@ -107,7 +107,8 @@ public class LightShaftsNode extends ConditionDependentNode {
         lightShaftsMaterial = getMaterial(LIGHT_SHAFTS_MATERIAL_URN);
 
         int textureSlot = 0;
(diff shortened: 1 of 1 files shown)


GooeyHub:

jellysnake:
Can confirm this tests out. It's a bit weird that the shafts instantly snap from being off of the moon to the sun. But that's only at sunrise/sunset and isn't worth blocking this PR for. It's much better than it coming from the ground aha.
I'm not a rendering expert but the code change looks simple enough that I'm happy to merge this in now.

PR #3687, opened 2019-06-13
Fixes invalid values for 'port' in AddServerPopup

skaldarnar:
This fix applies `.trim()` to the text to ignore trailing whitespaces.

The 'port' value in the `AddServerPopup` can end up invalid due to
whitespaces, such as newline characters (`\n`). This will disable the _OK_ button and the user cannot add a new server.


skaldarnar:
I'm wondering whether there is a more complex bug in the background that the text in the text widget can end up having newline characters which cannot be deleted...
@Cervator did you encounter anything like that before?

eviltak:
Thanks @skaldarnar! Since @jdrueckert has kindly confirmed that the changes fix the issue, I'll go ahead and merge.

PR #3689, opened 2019-06-14
Fix getText() function in UIText

### LINKED ISSUE #3688: Terrible things happen when starting a _second_ game without a game restart

Test unrelated things! Which sometimes involve jumping into more than one world.

### What actually happened

Inconsistently would get warped worlds with messed up data, from broken chunks to entirely messed up block data (hills of torches, pure water world, etc)

### How to reproduce

This is the minimal use case I could consistently reproduce real quick

* Step 0: (Unsure if needed: start from a clean game, no saves or config)
* Step 1: Start the game from source, create a new game, pick Core, quick start
* Step 2: Verify all looks well, exit back to the main menu
* Step 3: Create a new game, pick Core, quick start - bad things happen

Similar issues have happened in the past (malformed block ids through some scenario), but I thought they were all long gone. Game state handling between multiple games in one game launch has remained a fragile area, however. Chances are something isn't being reset properly when leaving the first game world.

darshan3:
### Contains

Fixes #3688

### How to test

Start from a clean game with no save games. Try reproducing #3688, in both the ways stated over there. The new game should load properly now.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
@@ -680,12 +680,13 @@ public class UIText extends WidgetWithOrder {
         String arrayText = "";
         if(linesOfText.length > 0) {
             arrayText = String.join("\n", linesOfText);
+            if(text.get() != "") {
+                return text.get() + "\n" + arrayText;
+            }
+            return text.get() + arrayText;
+        } else {
+            return text.get();
         }
-
-        if(text.get() != "") {
-            return text.get() + "\n" + arrayText;
-        }
-        return text.get() + arrayText;
     }
 
     /**
--- a/engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
@@ -17,6 +17,7 @@ package org.terasology.rendering.nui.widgets;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
+import com.google.common.collect.ObjectArrays;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.input.Keyboard;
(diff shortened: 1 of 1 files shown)


darshan3:
The getText() function was incorrectly adding a '\n' character, which leads to parsing error `GameProvider.getNextGameName()`. Hence, when it incorrectly returns the game name as 'Game 1' even if saved games are present with the same name. This leads to further errors in loading and saving the game in the same directory.

darshan3:
I would also request @jellysnake or @eviltak  to have a look once since I don't know how multiline text works and haven't tested it out for the changes in this pr.

engine/src/main/java/org/terasology/rendering/nui/widgets/UIText.java
```diff
@@ -677,15 +677,15 @@ public void bindText(Binding<String> binding) {
...
-        if(linesOfText.length > 0) {
-            arrayText = String.join("\n", linesOfText);
-        }
-
-        if(text.get() != "") {
-            return text.get() + "\n" + arrayText;
+        if (linesOfText.length > 0) {
+            String arrayText = String.join("\n", linesOfText);
+            if (text.get().equals("")) {
+                return text.get() + "\n" + arrayText;
```
skaldarnar: In this case `text.get()` is the empty string `""` - why are we prepending a newline here? Should it rather be:

```java
 if (text.get().equals("")) {
    return arrayText;
} else {
    return text.get() + "\n" + arrayText;
}
```

darshan3: My mistake, it should be `!text.get().equals("")`. I will fix it.

darshan3: Should I just return `arrayText` instead of `text.get() + arrayText`? There was a discussion in original pr #3678 regarding this but resolved eventually.

Cervator:
Tested out a fair bit - merging! :-)

Since this fixes a blocker I figure merge sooner, ask more questions later if needed. Existing saves may still cause problems even after the fix (since I guess a bad name might persist that way) but in a clean workspace I didn't see any issues colliding with new saves.

eviltak:
Reverted with #3678.

PR #3702, opened 2019-07-14
Log warning when module specified in override.cfg not found

### LINKED ISSUE #3673: Log a warning if module defined in override.cfg doesn't exist
The `override.cfg` file holds settings for the headless server. These are applied over the ones in `config.cfg`. Modules to be activated by the headless server are also defined here.

Right now, if a module that doesn't exist is defined in the override file, the server just ignores it and moves on. It would be better if an error or at least a warning is thrown in these cases. 

The `StateHeadlessSetup` class might be a good place to start. It sets up all configuration options which you would control using the in-game menu UIs.

askneller:
### Contains

Fixes #3673
Logs warning in `StateHeadlessServer` when module is not found in module registry.

### How to test

Add a non-existing module (e.g. TestMissingModule) to the `override.cfg` file and then start the headless server (make sure you have no saved games otherwise it will load the config from the first one). In the logs you should see the message "ModuleRegistry has no latest version for module <module>".

Note: I noticed that the modules from `override.cfg` are loaded first, and then `config.cfg`. This tripped me up while testing. First you'll have to delete any entry in `config.cfg` for this
```
"defaultModSelection": {
    "modules": []
```
otherwise what you specify in `override.cfg` will be completely ignored.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/subsystem/headless/mode/StateHeadlessSetup.java
+++ b/engine/src/main/java/org/terasology/engine/subsystem/headless/mode/StateHeadlessSetup.java
@@ -15,6 +15,8 @@
  */
 package org.terasology.engine.subsystem.headless.mode;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.terasology.config.Config;
 import org.terasology.config.WorldGenerationConfig;
 import org.terasology.context.Context;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Hi @askneller and welcome! Thanks for the well-written PR :-)

Triggering bot for auto-testing

@GooeyHub: add to whitelist

GooeyHub:

jellysnake:
Works and LGTM so i'm merging in.

PR #3705, opened 2019-07-21
Move controller settings to new page, make ControllerInfo generic

### LINKED ISSUE #3648: Add new "Controller Settings" page
At present the game has a single "Input Settings" sub-menu for things like hot keys. The entries on that specific page are dynamically generated based on an annotation scan for things like `@RegisterBindButton` - each such button and its bindings are put into the Input Settings and the user is able to remap as desired. For an example look at the `FrobButton` class, which by default is the `e` key for interacting with a target object, but also has a secondary default for a `CONTROLLER_BUTTON` with an id of `FOUR`

At present if an actual game controller (for example a USB PS4 or Xbox controller) is connected to the system then the game is launched additional game controller specific options are added to the very bottom of the Input Settings page, such as the ability to configure a deadzone for a given axis. However, the options are fairly limited at present and it is hard to troubleshoot as only those limited options will show up when a working controller is present. See #2125 for some outstanding related issues.

Over time more and more creative controller types have appeared, and in addition to whatever LWJGL 2 allows us in input control (which is becoming quite dated and problematic) there is `OpenVRInput` which attempts to find VR controllers, such as the hand wavy things from the Vive (or, I assume, the Oculus). And, amusingly & apparently, accelerometers on laptops, which can cause weird drift if you happen to move your laptop in funny ways with the game running ...

There are other weird things that may register as a game controller. I've for instance seen my Win8 tablet apparently register its touch screen as a game controller? And who knows what else. There's no obvious way to interface with (or even block) these controllers - unless you're connecting a standard game controller that we can recognize and pop in basic settings for you get nothing.

Lets change that!

This issue requests a new "Controller Settings" sub-menu to go next to the regular input settings, with the current basic settings like deadzone config moved into it. As such this issue may not be easy to work _without_ an actual controller, but it may be workable to a limited degree.

Rather than solely list controllers we know what to do with the page should list _every_ detected controller that could somehow interact with the game and at the very least give the user some sort of option to disable them / ignore input from them inside the game. If no controllers at all are detected then naturally just display a single message on that page going "No controllers found" or so.

If we do believe a specific controller can be used (even creatively, like apparently balancing your laptop to move!) then try to list reasonable configurations for said controller, in its own section. To begin with this may still be nothing but the original deadzone config, maybe an inverse axis or two, etc. Basic stuff.

It would be a nice extra to also list _which_ "buttons" are associated with a given controller on that page. They should still primarily be assigned on the input settings page, but another nice (but likely far more complicated) extra would be to then pick a given controller button and pick a function from a drop-down, which would then replace either the regular or secondary binding for that key. 

Example: A joystick is connected, and supports `CONTROLLER_BUTTON` `JOY1` through `JOY8` - each such entry becomes a button on the page. The user clicks the `JOY5` button and gets a drop-down of all functions that can be mapped to that button. User selects "Frob" and the `FrobButton` secondary binding becomes `JOY5` (I figure it may make more sense for controllers to occupy secondary bindings)

@darshan3 has expressed interest in working some on this issue, but it probably has room for more (particularly users with actual game controllers, especially unusual ones)

We have lots of "controller" related issues, see [this search](https://github.com/MovingBlocks/Terasology/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+controller) - maybe of particular interest is #2933 since we purposefully broke some motion controller stuff that had impacted something else negatively. Maybe some of the stuff linked from there will be good to review for some inspiration (and any fixes would be very appreciated)

askneller:
### Contains

Fixes #3648. I moved the controller settings to a separate UI page. The page lists each connected controller and its settings. To do this I fleshed out ```ControllerInfo``` with lists of buttons, axes, and hats, so that the info is more generic and conforms more to the information that ```JInput``` provides. This also required some changes in ```InputSystem``` and ```JInputControllerDevice``` for handling the new axes (there are no changes related to handling of buttons, I wanted to make minimal possible changes).

### How to test

Connect a controller. Go to Settings -> Controllers and you will see all connected controllers.
Some notes about testing (possibly for future issues):
- The system seems to assume only one connected controller. It seems that ```AbstractBindableAxis``` has no reference to a particular controller, so multiple controllers would probably be able to control the one binding (I didn't test this). Same with ```BindableButton```. Because of this I've listed all button bindings at the bottom of the settings page.
- Axis inversion appears to be buggy. I can see from logging that ```InputSystem.processControllerAxisInput``` is inverting ```axis.setTargetValue```, but the resultant movement in game still goes in the non-inverted direction. I was also able to re-create on the ```develop``` branch (n.b. I only looked at this very briefly).


GooeyHub:

askneller:
@Cervator Would you be able to look at this when possible please?

askneller:
Closing this in favour of #3839.

PR #3709, opened 2019-07-23
convereted camera to JOML

pollend:
This PR ports the camera over to work with JOML vectors instead of termath. I think I break some of the shading effects with this port because god rays don't work quite correctly. some of the matrices are not functioning how I'm expecting them to or methods with similar parameters and breaking. I would like to add more test to verify the camera more but this is a good first step. I moved the display width and height into lwjldisplay so it should allow more better testing.

- `LwjglDisplayDevice.java` - Moved display width and height in LWjgl display for testing.
- `JomlUtil.java` - Utility to convert from termath < -- > joml
- `Material.java` - additional joml methods we're added to allow for either joml or termath 

A lot of small changes with nodes that translates between termath and joml. 

### problems

- mul method in termath takes two vectors multiplies them and sets the original vector. mul in joml takes the current vector multiplies it by the first and uses an optional 3rd to set the vector.

- for quaternion angle will offset it by (2pi  - angle) if it's greater then 0 and this seems to break camera rotation 

- god rays don't seem to function correctly

- for floatbuffers terasology uses matrices in row - major and not column major so getTransposed it used instead.


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

PR #3711, opened 2019-07-25
Implement movable selection feature

Naman-sopho:
### Contains

Implementation for a new feature which enables the player to select a suitable position for a selected region after specifying its size.

There is no change in the block selection logic so any modules depending on old logic should work fine. In order to use the new feature, the `MovableSelectionStartEvent` must be sent which binds the camera position with the region and the player can select the final position of the region by clicking the left mouse button.

### How to test

Refer to the PR [MasterOfOreon#50](https://github.com/Terasology/MasterOfOreon/pull/50) for a working example and steps to test.


GooeyHub:

PR #3712, opened 2019-07-27
Plug in a few missing BiomesAPI spots

Cervator:
Short and simple, just ran into it by coincidence trying to update all the modules, that script didn't like BiomesAPI being present without being its own Git repo.

Making PR to doc, to see if somebody wants some easy merge practice, and in case somebody else manages to spot anything else I missed.


GooeyHub:

PR #3713, opened 2019-07-30
Fix APIScanner NPE when running MTE tests

eviltak:
there are _two_ packages named `org.terasology.input.binds.inventory`: one in the engine, one in the Inventory module. The engine one is `@API`, but the Inventory one is not. When the APIScanner looks for API annotated types, Reflections _sometimes_ thinks it's returning the engine package but actually returns the Inventory package. APIScanner is unable to find the annotation on the Inventory package,  causing an NPE.

This is fixed by removing the API annotation from the package and moving it to the only class in the package, `UseItemButton`.

Resolves Terasology/ModuleTestingEnvironment#9.


GooeyHub:

Cervator:
Tested this locally but weirdly can't seem to push my merge from the local workspace. Merging via UI button instead.

Thanks for the detective work @eviltak this is an important fix!

PR #3716, opened 2019-08-03
Deprecated Perlin noise over Simplex

MaxBorsch:
Converted Core Perlin world generator to instead use Simplex noise. 

Deprecated Perlin noise in favor of Simplex noise, which is a newer algorithm from Ken Perlin with fewer directional artifacts, lower computational overhead for higher dimensions.

### DIFF
--- a/engine/src/main/java/org/terasology/utilities/procedural/PerlinNoise.java
+++ b/engine/src/main/java/org/terasology/utilities/procedural/PerlinNoise.java
@@ -20,8 +20,10 @@ import org.terasology.utilities.random.FastRandom;
 
 /**
  * Improved Perlin noise based on the reference implementation by Ken Perlin.
+ * @deprecated Prefer using {@link SimplexNoise}, it is comparable to Perlin noise (fewer directional artifacts, lower computational overhead for higher dimensions).
  *
  */
+@Deprecated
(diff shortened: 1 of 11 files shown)


GooeyHub:

engine/src/main/java/org/terasology/utilities/procedural/PerlinNoise.java
```diff
@@ -20,8 +20,10 @@
 
 /**
  * Improved Perlin noise based on the reference implementation by Ken Perlin.
+ * @deprecated Prefer using SimplexNoise, it is comparable to Perlin noise (fewer directional artifacts, lower computational overhead for higher dimensions).
```
skaldarnar: I think you can put a link to the class. Not sure about the syntax how to achieve this in JavaDoc, though...

MaxBorsch: Added link

GooeyHub:

skaldarnar:
There are most likely more broken modules as they are using the (now gone) `Perlin` classes ... 

See https://github.com/search?q=org%3AMovingBlocks+org%3ATerasology+perlin&type=Code

Cervator:
Yep, just noticed TutorialWorldGeneration as one of them, probably with more hits in its wiki. Fixes there in particular would be appreciated 👍

PR #3718, opened 2019-08-10
Fix NPEs related to movable selections

Naman-sopho:
### Contains

Fix for two NPEs thrown by the `LocalPlayerBlockSelectionByItemSystem`:
- When the cam target had no `LocationComponent`, for eg: when the player is looking up at the sky
- When `onLeftMouseButtonDown` event was not triggered after a selection had been made, in which case the `blockSelectionComponentEntity` would be null


GooeyHub:

Cervator:
Tests out, thanks 👍

PR #3721, opened 2019-08-12
Use TypeInfo to store value type in Setting

eviltak:
The value type is stored as a `TypeInfo` which stores information like generic parameters. The `FlexibleConfig` API remains unchanged for now -- the `FlexibleConfig.newSetting` method still takes a `Class` parameter.


GooeyHub:

PR #3724, opened 2019-08-14
Add ShortBuffer and FloatBuffer to API whitelist

SoniEx2:
These are useful for audio buffers. While it's still necessary to
copy to a `ByteBuffer` when passing it on to Terasology, this
helps a lot during synthesis and effect processing.

Additionally, these are analogous to `ByteBuffer` and `IntBuffer`. So
it seems reasonable to have these too. (And in a future PR,
maybe `DoubleBuffer`, `LongBuffer` and `CharBuffer`?)

### Contains

### How to test

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
+++ b/engine/src/main/java/org/terasology/engine/module/ExternalApiWhitelist.java
@@ -89,7 +89,9 @@ public final class ExternalApiWhitelist {
             .add(Logger.class)
             .add(java.awt.datatransfer.UnsupportedFlavorException.class)
             .add(java.nio.ByteBuffer.class)
+            .add(java.nio.ShortBuffer.class)
             .add(java.nio.IntBuffer.class)
+            .add(java.nio.FloatBuffer.class)
             .add(java.nio.file.attribute.FileTime.class) // java.util.zip dependency
             // This class only operates on Class<?> or Object instances,
             // effectively adding a way to access arrays without knowing their type


GooeyHub:

Cervator:
@GooeyHub: ok to test

Thanks @SoniEx2

GooeyHub:

eviltak:
@SoniEx2 thanks for taking the initiative and making a PR! If you'd like, you can add the other buffers to the whitelist in this PR itself; I personally don't see why they shouldn't be whitelisted :)

SoniEx2:
should I add Buffer as well?

SoniEx2:
also, what about `Buffer*Exception`s, etc?

Cervator:
I would say to just add what you are immediately using, or we could keep suggesting classes till we're out of Java :-)

Speaking of - it would be nice to actually see some of the code this is being used for. Open source and all that.

I'm going to merge this as-is and suggest for you to feel free to submit a second PR for the other plain Buffer classes if you're going to use them, ideally with some actual details for what you're using them for (audio stuff - I get that much. Like links to a module repo eventually). That'll also give you another chance to heed the PR template 👍 

Thanks for the review and confirmation @eviltak!

SoniEx2:
so basically, I'm making a non-Terasology-specific voxel game DAW/noteblock library. (don't worry, its license is explicitly incompatible with Minecraft.) just plain Java and you write your own bindings on top.

I'm not sure if you can `.flip()` FloatBuffers and other buffers without the `Buffer` class being in the sandbox. I haven't tried it yet. so maybe I'll need to add it as well.

the point of the library, however, is to be extensible. it'll provide the basics and then you attach whatever you want to it. FM synts, ADSR envelopes, whatever.

PR #3726, opened 2019-08-16
TypeRegistry Optimization

eviltak:
Includes a number of improvements and optimizations for the newly-introduced `TypeRegistry` system.

- The `TypeRegistry` now only scans types that are either whitelisted or declared in a module (engine included) to exclude types that will rarely (if ever) need looking up via the registry.
- Type scanning and loading is now parallelized to improve performance on average.
- The `TypeRegistry` has been separated from the `ModuleManager`, and is manually reloaded every time a new `ModuleEnvironment` is loaded.


GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/engine/module/ModuleManagerImpl.java
```diff
@@ -128,7 +125,7 @@ public ModuleManagerImpl(Config config, TypeRegistryImpl typeRegistry) {
     }
 
     public ModuleManagerImpl(Config config, List<Class<?>> classesOnClasspathsToAddToEngine, TypeRegistryImpl typeRegistry) {
```
pollend: Any parcitular reason why you still need the typeregistry when you've removed it from the class?

engine/src/main/java/org/terasology/reflection/internal/TypeRegistryImpl.java
```diff
@@ -42,12 +51,36 @@
...
+        if (i == -1) {
+            return false;
+        }
+
+        String packageName = typeName.substring(0, i);
+
+        return ExternalApiWhitelist.PACKAGES.contains(packageName) || WHITELISTED_CLASSES.contains(typeName);
+    }
+
     public void reload(ModuleEnvironment environment) {
```
pollend: maybe move this into the interface so you can mock it for testing.

GooeyHub:

engine/src/main/java/org/terasology/engine/module/ModuleManagerImpl.java
eviltak: I must've missed this constructor overload. Thanks for catching it!

engine/src/main/java/org/terasology/reflection/internal/TypeRegistryImpl.java
eviltak: That's a good idea. I'm thinking I might just get rid of the interface and make it a single class, since it isn't meant to be accessed outside the engine.

GooeyHub:

engine/src/main/java/org/terasology/reflection/TypeRegistry.java
```diff
@@ -15,14 +15,127 @@
...
+        }
+
+        // Here allClassLoaders contains child class loaders followed by their parent. The list is
+        // reversed so that classes are loaded using the originally declaring/loading class loader,
+        // not a child class loader (like a ModuleClassLoader, for example)
+        Collections.reverse(allClassLoaders);
+
+        classLoaders = allClassLoaders.toArray(new ClassLoader[0]);
+
+        reflections = new Reflections(
```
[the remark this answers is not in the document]
eviltak: I'm guessing this is an Android-only problem due to how Reflections scans for loaded types. I'll note this down somewhere and see if I can create/modify a gradle task to do this!

engine/src/main/java/org/terasology/reflection/TypeRegistry.java
```diff
@@ -15,14 +15,127 @@
...
+            }
+
+            reflections.merge(module.getReflectionsFragment());
+        }
+    }
 
-public interface TypeRegistry {
-    <T> Set<Class<? extends T>> getSubtypesOf(Class<T> type);
+    public <T> Set<Class<? extends T>> getSubtypesOf(Class<T> type) {
+        Iterable<String> subTypes = reflections.getStore().getAll(SubTypesScanner.class.getSimpleName(), type.getName());
```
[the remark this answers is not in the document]
eviltak: I did this to parallelize the loading of classes as well. When many subtypes are found -- about 6000 `Object` subtypes just in the engine, core modules and whitelisted packages -- the loading of these many classes becomes somewhat of a bottleneck. The newly added `ReflectionUtil.loadClasses` method loads classes from the given class loaders in parallel to drastically reduce the time taken. 

As a benchmark (on a warmed up JVM in my slow temp work machine), the `TypeRegistryTest.testWhitelistedTypes` test (which calls `getSubtypesOf(Object.class)` once), runs in ~8 seconds with `reflections.getSubtypesOf`. With the parallelized class loading method, it runs in 4 seconds. The time taken to run the other two tests also drops by about 1 second each.

GooeyHub:

PR #3733, opened 2019-08-25
Add 'get-all' command to groovy script

skaldarnar:
### Contains

This is to avoid issues with `groovyw get *` on different platforms.

- Windows: `get *` would expand to files in current directory
- *nix: the wildcard needs to be escaped in single quotes `get '*'`

With this PR, a unified command for getting all modules is available.

### How to test

Try the new groovy command `groovyw module get-all` to retrieve all modules into a mega workspace.

Nothing. 

While looking at the code I was wondering whether we can improve the whole groovy scripting, though.

- use proper logging instead of `println` statements
- use [`CliBuilder`](http://docs.groovy-lang.org/2.4.7/html/gapi/groovy/util/CliBuilder.html) for better command line interface (less manual code)

### DIFF
--- a/config/groovy/util.groovy
+++ b/config/groovy/util.groovy
@@ -73,7 +73,14 @@ switch(cleanerArgs[0]) {
             common.retrieve(((String[])selectedItems.toArray()), recurse)
         }
         break
-
+    case "get-all":
+        println "Preparing to get all modules"
+        ArrayList<String> selectedItems = new ArrayList<String>()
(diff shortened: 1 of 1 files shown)


Cervator:
Actually, found another issue: It stopped early (?) - "Underworld" was the last module I got. Missed everything past that point. Guessing that is a paging issue and may be preexisting

GooeyHub:

PR #3735, opened 2019-09-02
Added kanji font pngs and updated Japanese menu text.

nakasya:
### Contains

Kanji fonts for Japanese (NotoSansCJKjp-Regular.orf)
Added and updated many Japanese translations for menu_jp.lang.

### How to test

Change the language to 日本語(Japanese) from the settings menu Settings>Player>Language.


GooeyHub:

Cervator:
Hi there @nakasya and nice work! While I can't read Japanese I can see the character set finally work! While in the past we just couldn't see alternative scripts in the UI at all. Are you familiar with internationalization, fonts, and so on? We badly need some more effort there, we've got several other languages translated that are missing the character support :-)

Previously I was under the impression we'd have to add chunks of Noto separately and use code to switch between them, or something similarly more complex than simply adding more stuff to the same font file. I didn't remember why, but I did recall that #2677 talks about the topic a lot, and links to other resources. In short: awesome to see it work here, but it sounds like it might not be sustainable to add every language that way? What do you think? Would be thrilled to have some more feedback, ideas, and if this approach does work then some more additional sets if you're up for it! See that issue for more details :-)

@GooeyHub: ok to test

GooeyHub:

nakasya:
Hi Cervator,
Unfortunately, I am not familiar with internationalization other than Japanese.　Chinese usually uses about 1000 more kanji characters than Japanese. Japanese and Chinese are quite different.

I think Terasology's font system is not  sustainable. For example, I guess that Minecraft uses system fonts in Japanese. Because it is different from other language fonts.
I will read the #2677 talks and explore the possibilities.

Thanks.

Cervator:
Much appreciated :-)

Yeah we picked Noto because it is essentially available for _everything_ yet we didn't get super far in then actually figuring out how to best handle all the different character sets.

nakasya:
Hello,
I reviewed the font system on Terasology for internationalization. This requires a font change system as you have written.
There are more than 8000 Chinese characters and so on, and the image will be huge(more than I remember). So I decided to select fonts only for those used in menu text (e.g. menu_ja.lang).

This method is already working in Japanese (I made cli-tool which output text file that can use with BMFont). Regular, medium and large can be displayed with a small amount of memory.
Since only the fonts used are used, the Pngs are small. Although, all original alphabets and numbers are included, even if they are not used in menu_*.lang.
I think that the language of the menu can be internationalized by collecting these into fonts and subfolders for each country.

e.g.
./engine/src/main/resources/assets/fonts/NotoSans-Regular.fnt
./engine/src/main/resources/assets/fonts/NotoSans-Regular-Medium.fnt
./engine/src/main/resources/assets/fonts/NotoSans-Regular-Large.fnt
		V
./engine/src/main/resources/assets/fonts/ja/NotoSans-Regular.fnt
./engine/src/main/resources/assets/fonts/ja/NotoSans-Regular-Medium.fnt
./engine/src/main/resources/assets/fonts/ja/NotoSans-Regular-Large.fnt
./engine/src/main/resources/assets/fonts/ch/NotoSans-Regular.fnt
./engine/src/main/resources/assets/fonts/ch/NotoSans-Regular-Medium.fnt
./engine/src/main/resources/assets/fonts/ch/NotoSans-Regular-Large.fnt
./engine/src/main/resources/assets/fonts/ko/NotoSans-Regular.fnt
so on.

PRO:
All language fonts can be displayed and the memory is small.
CON:
Free text such as chat and dialogue cannot be used because there is no font.

BTW, I have a question.
I have a question about UI internationalization.
The UI skin system seems to be unique system when I read the wiki page, but is it possible to change the font?
For example, I want to change the font folder in the selected language, but I can't find the code-location.

Of course, if you have other good ideas, please let me know.
*I'm not a native English speaker, so please ask questions if you don't understand from text.

Cervator:
Hey again @nakasya - sorry for taking a little time to get back to this :-)

Been thinking about it a bit and just talked a little about it on chat, will ping you so you have a marker in case you'd like to read it. A summary:

* We have a "conlang" - a constructed language, or at least a made up font. This naturally isn't within Noto and should be its own thing. It'd be used for some in-game displays but not all - unless we cheat and only ever use it indirectly (just create outright images anytime we need text in that font/language)
* Yep "free text" would be a problem, and worse: even if we have _complete_ fonts we still can't globally switch as there may be a need to use multiple fonts, such as showing the conlang, or if multiple players are using different languages with unique alphabets in multiplayer (how to show English if the other player is running in Japanese)

As noted on chat an "uber font" (just cram everything into a single font) would solve some of those issues but not all (custom fonts in certain modules for flavor) plus it could result in gigantic files. Not actually sure if heavy fonts are a typical cause for performance issues but it surely wouldn't _help_

Generating partial fonts ... well, it would work in limited circumstances, but you'd still need switching for flavor fonts, in-game chat, etc. So I'm not sure if that'd be worth it? Another question is how much "free text" would even normally be needed. One simple example beyond chat that I can think of is in-game books and editing them yourself

So: ideally it seems like we'd support font switching, of full sets, both globally (setting a default) and per widget or even per-line (chat). In that case even if some of the character sets are big we could probably use them unmodified from Google? A normal widget would just default to the globally set font, but you could override/customize and indicate a specific one. Which come to think of it, you can already indicate in `.ui` files or something, hmm 🤔 

But yeah, we need the switching code first of all, at least connected to the user's default language. I haven't gone digging yet, but surely somebody has bumped into it before. I think @Halamix2 was into that area long ago, and @ZPendi more recently with DestSol. @eviltak is a good resource for UI, as is @rzats if around again yet :-)

qwc:
@jellysnake and @eviltak said MERGE IT.

PR #3751, opened 2019-10-03
Better punctuation

aRealCyborg:
### My description 

I feel like the punctuation of the read me would be better with some added . and a few , that I added. 
I am not a master at grammar or anything but I feel like that this would be "more correct"
If you think the changes are not good it is ok, I wanted to try to be helpful.
For line 22: "We encourage contributions from anybody and try to keep a warm and friendly community and maintain a [code of conduct](https://github.com/MovingBlocks/Terasology/blob/develop/docs/CODE_OF_CONDUCT.md)" I did not add a . because it would make it look weird.

### Contains

Adds better punctuation. 

### How to test
Please read it to decide if you agree.


GooeyHub:

Cervator:
Merged and added one more comma while in there. Plus included you in the credits. Thanks :-)

PR #3755, opened 2019-10-10
Fixes Game Details screen by restoring Worlds & Biomes page

Qwertygiy:
Partially reverts ff190cf3e59f8e1a59b6d59ddec2a63dd71071ee which removed page2 from the UI, breaking the screen entirely. 

The Biomes section of the UI will remain non-functional until a new implementation is made using BiomesAPI, but it is now merely an empty label, rather than erroring the entire screen.

To test, create any singleplayer world, return to main menu, and then click the "Details" button on that world. It should show the modules and blocks used in that world without creating an error popup.


GooeyHub:

PR #3757, opened 2019-10-10
Adjustments to Block Updates

Qwertygiy:
Previously, the OnChangedBlock event was not being created when blocks were placed by the player.

Additionally, the event sent to the block *region* was separate from the event sent to the *block.*

This PR corrects those two problems with minimal changes to the codebase. It should solve Terasology/SimpleLiquids#4

Wider-reaching changes, such as fully replacing usage of the BlockChange class with the OnChangedBlock event, should remain within the scope of a separate PR.


GooeyHub:

PR #3743, opened 2019-09-18
Bump Gradle to latest 4.x

Cervator:
Seems to work OK in my mega-workspace. Should resolve an issue where you might fail to compile the game with a Java too new for our prior version of Gradle 4.x. Gradle 5.x on the other hand is bound to bring in a lot more change.

Also tested it with a sample `groovyw get` module check. Anybody else want to see if they can come up with something Gradle-related that'll break with this?

Will see if Jenkins builds this OK as well. Probably can merge pretty soon. Pinging @pollend for review, maybe anybody else with a mega-workspace and interest, like @jellysnake @eviltak @Qwertygiy  or somebody :-)

PR #3764, opened 2019-10-17
Upgrade Artifactory plugin to fix compatibility issue with latest Gradle version bump

Cervator:
The engine Jenkins build started failing after #3743, the minor Gradle upgrade. At least that's my theory based on logging in Jenkins. With this version bump to the associated Artifactory plugin for Gradle it publishes normally again, at least locally. PR test won't show it (no publish step) so this PR is for doc purposes and will be merged immediately to test the fix in Jenkins.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ buildscript {
 
     dependencies {
         // Artifactory plugin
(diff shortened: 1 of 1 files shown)


GooeyHub:

PR #3765, opened 2019-10-17
Feature: Read shaders from source folder on recompile by Loonride

dave2s:
Continuation of Loonride's #3658.

@eviltak mentioned there was 

> a couple of questions about how the `development` flag is being handled in the PR.

This PR suggests following changes :
Rename boolean 'development' to 'fromSource' for ShaderManager(headless/-/impl)'s recompileAllShaders(bool fromSource) and rename Shader.developmentRecompile() to [GLSL/-]Shader.recompileFromSource()

_I think it's preferable to name methods based on what they do rather than what they are used for. Especially since the latter is not self-explanatory (hence the variable rename, to be aligned with the method name and more self-explanatory)._ 

Legacy #3658 PR's description:
> Contains
> When the recompileShaders command was run in the game in the past, it would use the shader files in the build folder, not the source folder. This made it harder to develop and test shaders. This PR introduces a somewhat hacky method to force loading of shader files from the source folder upon recompile. It will not do this on initial compilation, since most users and players will not be using this feature, or possibly will not have the local src folder. If the recompileShaders command is run without the proper src folder, or with a file missing, loading from the source folder for that particular file will simply not happen.
> 
> How to test
> Start the game
> Change something in the shader source folder
> run recompileShaders command and see the shader change
> Outstanding before merging
> I did not implement this for metadata, because it is such a small feature, unlikely to be useful when recompiling
> I considered making the loaded file info from the source folder overwrite the build folder files, but decided against it in case someone accidentally starts making changes in the build folder and realizes their mistake (functionally there is no difference between it being overwritten or not overwritten)
> The source file loading happens in the ShaderData class. The class was not designed for loading the data, but I think it is the best intermediary between the GLSLShader and the GLSLShaderFormat, which is usually responsible for the loading but I don't believe is reachable in any way to be used for this use case.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
+++ b/engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
@@ -125,6 +125,19 @@ public class GLSLShader extends Shader {
     }
 
     @Override
+    public void nonInitialRecompile() {
+        // try to load the shaders from the source folder, not the build folder,
+        // on recompiles caused by the recompileShaders command
+        try {
+            shaderProgramBase.reloadFromSource();
+        }
+        catch (IOException e) {
+            logger.warn("Shader source files not found for {}", getUrn());
+        }
+        recompile();
+    }
+
+    @Override
     public void recompile() {
         registerAllShaderPermutations();
         // TODO: reload materials
--- a/engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
+++ b/engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
@@ -130,8 +130,7 @@ public class GLSLShader extends Shader {
         // on recompiles caused by the recompileShaders command
         try {
             shaderProgramBase.reloadFromSource();
-        }
-        catch (IOException e) {
+        } catch (IOException e) {
             logger.warn("Shader source files not found for {}", getUrn());
         }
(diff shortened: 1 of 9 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
```diff
@@ -124,6 +124,18 @@ int linkShaderProgram(int featureHash) {
         return shaderProgram;
     }
 
+    @Override
+    public void developmentRecompile() {
```
eviltak: This method was renamed in `Shader`, we might need to rename it here to prevent an error (the Jenkins build fails because there is no `developmentRecompile` method in `Shader` to `@Override`)

dave2s: Sorry, my bad. I noticed the mail as I was already off. I'll fix it right away.

GooeyHub:

dave2s:
So I reviewed the original PR and you can check the comment I left there. #3658 
This PR is about the change in naming of the original PR's featured methods/arguments. I hereby ask anyone and @eviltak, who mentioned the doubts, to give their opinions on this change. 

I also wonder, whether the recompileFromSource() should be added to the HeadlessShader.java

PR #3770, opened 2019-10-20
Pass `context` directly to `loadConfig` method in AutoConfigManager

skaldarnar:
This addresses a warning reported by the [Checker Framework](https://checkerframework.org/).

I think having the context as a private member is not necessary here and
we can instead pass it directly to `loadConfig` where it is needed.


GooeyHub:

Issue #1975, opened 2015-10-12
Documentation overhaul


Cervator:
For the upcoming v1.0.0 release of the engine we should really overhaul our documentation to present a more comprehensive and consistent introduction to developing with the engine (and maybe playing the game in general). Need the help from multiple people here and new contributors can easily help by both pointing out weaknesses and figuring out then documenting simple things :-)

Big list of todos - not expecting we'll finish them all. Keep in mind a lot of these could be approached from two different contexts: how to code a new thing for feature x vs how to use feature x as a player
- [x] Game readme. Needs to be streamlined a lot, it is too bloated. A ton of it probably should be moved into the wiki (see #1974) and some other stuff into the few "magic" files like `CONTRIBUTING` (see #463). Mainly the readme should link to all kinds of stuff and show a brief into with a screenshot or two.
- [x] Hot keys. Currently in readme and on at least one outdated wiki page. Any good way to automatically scrape this from code? Or maybe just keep to the very basics (readme) - see #1470. Simple example on how to create a new hot key (or how to _avoid_ new keys in favor of better systems using existing functionality in-game) would be good as well. Maybe the Sample module should implement a custom hot key as a demo, showing a simple custom UI window? _Scraping for this implemented via #2051 - need to put to use_
- [x] Console commands. With the awesome in-game help system we have this in great shape. We probably just need to describe how to _develop_ for commands and delete any attempted lists of commands outside the game other than the very basics (again see #1470)
- [x] Put more hints into the launch scripts and the -help arguments about the launcher - see #1174
- [ ] Minimum requirements page - #1123
- [ ] Better organize + publish more intro videos to our gameplay templates like JoshariasSurvival or TTA (#932)
- [ ] The Sample module needs more stuff in it - #739
- [ ] Better introduction to using GitHub with the new tools available since we originally started, like the GitHub native client. We need a minimal hassle setup for contributors that don't necessarily need to code, such as artists just committing assets. #588
- [x] Slimmer version of the Dev Setup page in general, it too has bloated. Also a suggestion to have a basic overview video for those who follow visual instructions better (#217)
- [ ] Better doc + visibility for working on models / shapes in Blender. We have a nice forum thread at http://forum.terasology.org/threads/art-resources-guides-tutorial-videos-contributing-info.786 but again could use a nice short and sweet wiki page with details and links. #587 and maybe #508
[...]

PR #3776, opened 2019-10-30
Fixed some JavaDoc Errors

rajiv-thorat:
Hello,

Please consider this pull request.

Thanks

### Contains
(1) BindsConfig : Invalid symbols in link (.W)
(2) EntityInfoComponent : Missing import to link (StorageManager)
(3) ViewDistanceChangedEvent : Invalid link (ViewDistanceChangeRequest -> ViewDistanceChangedEvent)

(4) Errors in BindsManager could not be fixed : updateDefaultBinds is private.
(5) Error in LayeredZoneRegionFunction could not be fixed : regionFunction in Zone is private.

Can be linked to #1975


GooeyHub:

engine/src/main/java/org/terasology/config/BindsConfig.java
```diff
@@ -54,7 +54,7 @@
  * One example for a binding-input combination is the {@link ForwardsButton} 
  * which is defined in the engine with the id <code>forwards</code>.
  * The default input binding for the forward movement is the W-Key.
- * Therefore the binds for <code>engine:forwards</code> would contain an {@link Input} object with type {@link InputType#KEY} and id {@link KeyId.W}.
+ * Therefore the binds for <code>engine:forwards</code> would contain an {@link Input} object with type {@link InputType#KEY} and id {@link KeyId}.
```
Qwertygiy: "W" is a valid member of KeyId, and is correctly mentioned; it should be changed to KeyId#W, which is the correct JavaDoc syntax, rather than be removed.

rajiv-thorat: Ahh, yes. Thanks for that. Now trying to figure out how to add commits to an existing PR.

engine/src/main/java/org/terasology/rendering/world/viewDistance/ViewDistanceChangedEvent.java
```diff
@@ -22,7 +22,7 @@
 /**
  * Notifies the server that the clients view distance changed.
  *
- * If you want to change the view distance send {@link ViewDistanceChangeRequest} to the local player's client entity.
+ * If you want to change the view distance send {@link ViewDistanceChangedEvent} to the local player's client entity.
```
Adrijaned: I'm pretty sure that this claim is not true. The particular commit responsible for removal of ViewDistanceChangeRequest can be seen at https://github.com/MovingBlocks/Terasology/commit/66ec2c96a1ba7cff79e71eba7cf24b3bb1ded2cb#diff-a4fba91c13c6b1e0e34060570701a54a, and from the looks of it, it doesn't look like the behaviour was reimplemented.

rajiv-thorat: If I understand your comment correctly, you are saying that the claim that ViewDistanceChangeRequest is required to be sent to change the view distance is not true because it was deleted as seen in the attached diff, and never re-implemented. (and my change is valid :P)

Or do you mean that other than the fact ViewDistanceChangeRequest was deleted and never re-implemented, the claim that ViewDistanceChangedEvent is needed to change the view distance is false. (although I can see ViewDistanceChangedEvent being used in ServerViewDistance, I am not familiar with the functionality)

Thanks for the feedback though.

Adrijaned: The latter. ViewDistanceChangedEvent is sent only from client to server to inform the server of client changing the view distance, but as far as I can tell, there is no way to comunicate something like that the other way.

rajiv-thorat: Hmm, makes sense. Please check the change and let me know if that reflects the functionality.

rajiv-thorat:
Please excuse my ignorance about the functionality. I am confused by the last comment. My current understanding is this:

Before any of my changes, the JavaDoc suggests that to change the view distance, ViewDistanceChangeRequest must be sent to local player's client entity. This may or may not have been true before the [changes](https://github.com/MovingBlocks/Terasology/commit/66ec2c96a1ba7cff79e71eba7cf24b3bb1ded2cb#diff-a4fba91c13c6b1e0e34060570701a54a) you pointed out. This class deletion clearly makes the statement in the JavaDoc invalid.
My initial assumption was that someone refactored the class without updating the JavaDoc. You pointed out that the data flows from the client to the server, and not the other way around. So this meant that my change in edfce6a4056f13349221070676b77c57b9f2d2dd is not valid, notwithstanding my assumption.
In 77a24e3a84b2c65f7e1994742aba805fb55d5d0d, I reworded the statement to say that the ViewDistanceChangedEvent is sent by the client to the server, to reflect my understanding of the code.

> in its current form, it basically just rephrases the one above it.

Since the first changes suggests that the data flows from server to the client (which is wrong), and the last commit suggests that the data flows from the client to the server (which seems to me is in accordance with the code), your comment does not seem to make sense.

Please let me know if I have completely missed the point (:P) . If you still think that the line should be remove, I will send the changes asap. Also, if this was your intent from the beginning, sorry, I seem to have missed that.

Adrijaned:
Firstly, in general, avoid using `{@link}` to reference what you are currently writing documentation for - generally the word `this` is used for that.
Secondly, to make that line you changed correct, you would have to change the beginning of it from `To change` to `When changing`. Only then would it actually reflect what is meant to be happening in the code.
Thirdly, when this change would have been done to the latest version as of now, the result wouldn't be claiming anything new except what was already said above (`sent to the server to notify...` vs `client sends to server when...`)

PR #3783, opened 2019-11-05
Properly implement OnChangedBlock event for setBlocks() calls

Qwertygiy:
Previously, #3757 attempted to fix the lack of the OnChangedBlock event being called when a block was placed by the player.

For the specific use case being tested, where the listener did not care about a LocationComponent, it was sufficient. But upon further review, it was still not being properly issued. Instead of the event being sent to the *specific* world-block entity that was being updated, it was being sent to the block *type* entity.

This time, the event should be sent to the proper location, and is being sent from the proper world provider.

### DIFF
--- a/engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
+++ b/engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
@@ -125,6 +125,21 @@ public class EntityAwareWorldProvider extends AbstractWorldProviderDecorator imp
     }
 
     @Override
+    public Map<Vector3i, Block> setBlocks(Map<Vector3i, Block> blocks) {
+        if (GameThread.isCurrentThread()) {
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
```diff
@@ -124,6 +124,21 @@ public Block setBlock(Vector3i pos, Block type) {
         return null;
     }
 
+    @Override
+    public Map<Vector3i, Block> setBlocks(Map<Vector3i, Block> blocks) {
+        if (GameThread.isCurrentThread()) {
```
Cervator: Might it be worth adding a quick comment to a line or two here explaining the issue, or even overriding / appending some Javadoc in this case?

PR #3785, opened 2019-11-10
Gltf support

immortius:
### Contains

Adds initial support for gltf mesh, skeletal mesh and animations.

### How to test

spawnPrefab "engine:cubesm"

Let me know!


GooeyHub:

engine/src/main/java/org/terasology/engine/subsystem/headless/HeadlessGraphics.java
```diff
@@ -31,9 +31,7 @@
 import org.terasology.engine.subsystem.headless.renderer.HeadlessRenderingSubsystemFactory;
 import org.terasology.engine.subsystem.headless.renderer.ShaderManagerHeadless;
 import org.terasology.rendering.ShaderManager;
-import org.terasology.rendering.assets.animation.MeshAnimation;
-import org.terasology.rendering.assets.animation.MeshAnimationData;
-import org.terasology.rendering.assets.animation.MeshAnimationImpl;
+import org.terasology.rendering.assets.animation.*;
```
skaldarnar: No start imports 😉

engine/src/main/java/org/terasology/engine/subsystem/lwjgl/LwjglGraphics.java
```diff
@@ -34,9 +34,7 @@
 import org.terasology.engine.subsystem.RenderingSubsystemFactory;
 import org.terasology.rendering.ShaderManager;
 import org.terasology.rendering.ShaderManagerLwjgl;
-import org.terasology.rendering.assets.animation.MeshAnimation;
-import org.terasology.rendering.assets.animation.MeshAnimationData;
-import org.terasology.rendering.assets.animation.MeshAnimationImpl;
+import org.terasology.rendering.assets.animation.*;
```
skaldarnar: No star import 😉

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -527,7 +527,7 @@ public String spawnPrefab(@Sender EntityRef sender, @CommandParam("prefabId") St
 
         Optional<Prefab> prefab = Assets.getPrefab(prefabName);
         if (prefab.isPresent() && prefab.get().getComponent(LocationComponent.class) != null) {
```
skaldarnar: Do you have an idea how to use `Optional` in a more idiomatic way? I think the logic can be expressed as follows, but I'm not sure how to end up with the status strings as before...

```java
Assets.getPrefab(prefabName)
  .filter(prefab -> prefab.getComponent(LocationComponent.class) != null)
  .ifPresent(prefab -> entityManager.create(prefab.get(), spawnPos);
```

engine/src/main/java/org/terasology/logic/location/LocationComponent.java
```diff
@@ -72,12 +73,18 @@ public void setLocalPosition(Vector3f newPos) {
...
+     * @return The direction this entity is facing, relative to any parent
+     */
     public Vector3f getLocalDirection() {
         Vector3f result = Direction.FORWARD.getVector3f();
         getLocalRotation().rotate(result, result);
         return result;
     }
 
+    /**
+     * @return The location of this component, relative to any parent
```
skaldarnar: Is it location or rotation? And what is the difference between direction and rotation?

engine/src/main/java/org/terasology/logic/location/LocationComponent.java
```diff
@@ -202,4 +244,6 @@ public int hashCode() {
     public boolean shouldReplicate(FieldMetadata<?, ?> field, boolean initial, boolean toOwner) {
         return initial || replicateChanges;
     }
+
```
skaldarnar: Important new lines 😉

engine/src/main/java/org/terasology/math/MatrixUtils.java
```diff
@@ -17,9 +17,7 @@
 package org.terasology.math;
 
 import org.lwjgl.BufferUtils;
-import org.terasology.math.geom.Matrix3f;
-import org.terasology.math.geom.Matrix4f;
-import org.terasology.math.geom.Vector3f;
+import org.terasology.math.geom.*;
```
skaldarnar: No star imports 😉

engine/src/main/java/org/terasology/rendering/gltf/GLTFAnimationFormat.java
```diff
@@ -0,0 +1,228 @@
...
+import org.terasology.assets.format.AssetDataFile;
+import org.terasology.assets.management.AssetManager;
+import org.terasology.assets.module.annotations.RegisterAssetFileFormat;
+import org.terasology.math.AABB;
+import org.terasology.math.geom.Quat4f;
+import org.terasology.math.geom.Vector3f;
+import org.terasology.rendering.assets.animation.MeshAnimationBundleData;
+import org.terasology.rendering.assets.animation.MeshAnimationData;
+import org.terasology.rendering.assets.animation.MeshAnimationFrame;
+import org.terasology.rendering.gltf.model.*;
```
skaldarnar: No star imports 😉

engine/src/main/java/org/terasology/rendering/gltf/GLTFMeshFormat.java
```diff
@@ -0,0 +1,90 @@
...
+package org.terasology.rendering.gltf;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.terasology.assets.ResourceUrn;
+import org.terasology.assets.format.AssetDataFile;
+import org.terasology.assets.management.AssetManager;
+import org.terasology.assets.module.annotations.RegisterAssetFileFormat;
+import org.terasology.rendering.assets.mesh.MeshData;
+import org.terasology.rendering.gltf.model.*;
```
skaldarnar: No star imports 😉

engine/src/main/java/org/terasology/engine/subsystem/headless/HeadlessGraphics.java
immortius: No prob. I assume the IntelliJ default settings just got me there.

engine/src/main/java/org/terasology/logic/location/LocationComponent.java
immortius: Prexisting code, but direction would be a vector pointing the way the entity is facing, while rotation is a quaternion I imagine.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
immortius: Hmm, I think I would do:
```    return Assets.getPrefab(prefabName).map(prefab -> {
            LocationComponent loc = prefab.getComponent(LocationComponent.class);
            if (loc != null) {
                entityManager.create(prefab, spawnPos);
                return "Done";
            } else {
                return "Prefab cannot be spawned (no location component)";
            }
        }).orElse("Unknown prefab");````

Comment (skaldarnar) [reply engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java]: That would work... I'm always hesitant to have side effects in `map` and try to use `ifPresent` (`forEach` in Scala) for that purpose. 
I think your proposal is better than using the combination of `isPresent` and `get()` (multiple times). 👍

**Comment by Cervator** [comment]
Added some more potential reviewers :-)

@Quarternius just updated the Metal Renegades robots to a single mesh approach that should work, would be good to get somebody to try those out. See https://github.com/MetaTerasology/MetalRenegades

It would also be wonderful if we could have some help adding a page on the new `glTF` approach in https://github.com/Terasology/TutorialAssetSystem/wiki with a sample model added there as well.

If we can merge this soon'ish we could probably make a GCI task to take one of Quarternius' many models that seem fitting for the game to test them out with the new format while commenting on any difficulties doing so

**Comment by GooeyHub** [comment]

File: engine/src/main/java/org/terasology/math/MatrixUtils.java [hunk 9]
```diff
@@ -245,4 +247,10 @@ public static Matrix3f calcNormalMatrix(Matrix4f mv) {
         result.transpose();
         return result;
     }
+
+    public static Quat4f extractRotation(BaseMatrix4f m) {
```
brylie: It might be good here to add a docstring explaining the formulas. They are somewhat dense, and not obvious to the uninitiated. Is there some upstream documentation for these formulas?

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
```diff
@@ -74,62 +73,57 @@ public Bone getRootBone() {
...
-            positions.add(bone.getObjectPosition());
-            rotations.add(bone.getObjectRotation());
+            transforms[bone.getIndex()] = bone.getObjectTransform();
         }
-        return getVertexNormals(positions, rotations);
+        return getVertexNormals(Arrays.asList(transforms));
     }
 
-    public List<Vector3f> getVertexPositions(List<Vector3f> bonePositions, List<Quat4f> boneRotations) {
+    public List<Vector3f> getVertexPositions(List<Matrix4f> boneTransforms) {
```
brylie: Is there a process for auto-generating documentation? Would documenting these methods potentially be useful for new developers to become familiar with the skeletal mesh API?

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
```diff
@@ -74,62 +73,57 @@ public Bone getRootBone() {
...
             }
-            results.add(vertexPos);
+            skinMat.transformPoint(pos);
+            results.add(pos);
         }
         return results;
     }
 
-    public List<Vector3f> getVertexNormals(List<Vector3f> bonePositions, List<Quat4f> boneRotations) {
+    public List<Vector3f> getVertexNormals(List<Matrix4f> boneTransforms) {
```
brylie: Add docstring?

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
```diff
@@ -74,62 +73,57 @@ public Bone getRootBone() {
...
+                skinMat.add(jointMat);
             }
-            results.add(vertexNorm);
+            skinMat.transformVector(norm);
+            results.add(norm);
         }
         return results;
     }
 
     public int getVertexCount() {
```
brylie: Add docstring?

brylie:
This looks pretty dang nice!

In general, what is the guideline for documenting public methods and classes? Does Terasology use or plan to use a documentation generator, to help developers better understand the APIs?

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
immortius: If the documentation is autogenerated then it is likely not providing any information of value (since it would all be derived from the method name/parameter names/etc.

I'll chuck some javadoc on these methods. I'm not sure they would be used much - primarily they exist for the rendering system and I'm not sure there is much reason to otherwise interact with them, maybe custom collision checking against exact mesh position I guess.

GooeyHub:

Cervator:
Tested out the new model, works!

Tested out old models: not so much, which is fair if we're getting a superior format anyway. But we need to convert the old ones to `glTF` real quick first then as part of merging this PR :-)

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
brylie: OK, I just noticed they were public methods, so thought they might be used by a developer somehow.

Cervator:
Update: There is now a rebased version of this PR in #3808 - easier to use for testing!

GooeyHub:

casals:
@immortius about the floating bug discussion mentioned at artwork: there was a bug affecting the `heightOffset` parameter being masked by another bug - but this PR fixes the second one, so now we can get rid of the first as well :) Basically what's happening now is: this parameter is being processed/rendered off-scale, and with the modifications you made there's no practical effect right now (try setting if to -100 or 100 and you won't see any difference in-game). Here's the fix:

at engine/src/main/java/org/terasology/rendering/logic/SkeletonRenderer.java :

- remove `worldPositionCameraSpace.y += skeletalMesh.heightOffset;` from line 272 (cut);
- place it at line 253, right before the AABB transformation (`aabb = aabb.transform(worldRot, worldPos, worldScale);`)

This, combined with Behaviors PR 23, should correct both the floating bug and the weird rotation observed on deer.

PAndaContron:
Not sure if this is a known issue, but I found that the rotation field in gltf files appears to be completely ignored. This can be replicated by going to the "Object Properties" tab in Blender and changing the rotation values, then exporting it. None of these rotations have any effect, despite being exported with the file.

PAndaContron:
Looking through the code, it appears that rotation is used when loading bones, but not when loading meshes. This might also cause some issues with animations, although I have not tested that.

PAndaContron:
I made a PR #3812 to add support for transformations, although it needs more testing.

Cervator:
New Oreons are ready! Many thanks to @Quaternius :-)

https://github.com/MetaTerasology/Oreons

I think that was the bulk of the remainder, if we can just get them all tested and PRed to their module we might finally be able to merge this while retaining most the models (and the gameplay they're used in). 

The Light & Shadow critters are still in an odd place. We have _most_ the red team, but in an old storage location at https://github.com/MovingBlocks/TeraMisc/tree/master/lightandshadow/redteam - need to move to a new MetaTerasology repo. Additionally I don't see the knight in there, yet I know we have the game asset to spawn it. There's also the Dominog. I think there was even an adventurer? But maybe that was just the 2D concept art by @SuperSnark. Those might be findable via forum or elsewhere, and then likewise just need a re-export to `glTF` along with testing. Just a few more volunteer hours and we should be ready :-)

Pinging @brylie @PAndaContron @casals @Kanonbell @Naman-sopho extra in case of interest in helping with the last few (from comments here and on model PRs, along with Naman for MOO itself). Right this moment MOO may be stuck in dependency figuring-outing as we're reworking some fundamental modules, but that shouldn't be too much hassle to sort out.

Now behold! The Oreons v2.0:

https://cdn.discordapp.com/attachments/541688465226596362/710248508279488532/4.gif
https://cdn.discordapp.com/attachments/541688465226596362/710248502646538352/1.gif
https://cdn.discordapp.com/attachments/541688465226596362/710248517989302322/5.gif
https://cdn.discordapp.com/attachments/541688465226596362/710248503791583312/2.gif
https://cdn.discordapp.com/attachments/541688465226596362/710248507692417044/3.gif

brylie:
I am willing to help with the model PRs. What specifically should I do?

Cervator:
Hey @brylie :-)

That would be super awesome. Essentially it is one or two things:

1. If there is already a contributed `gltf` PR for a given module make sure it actually tests out and the creature reacts as intended in-game (animations trigger and so on). This may be a little tricky in some cases if it isn't obvious how to actually trigger some animations. We used to have (still might live in the new Debugging module?) an animation tool / UI where you could select and test a model in-game, but i'm not sure if that exists anymore, or would even work for `gltf`
2. If there is no PR and we just have loose `gltf` files somewhere then put the files together with adjusted prefabs so they can be tested like in the first point, and make a module PR.

jdrueckert:
[GLTF Support](https://trello.com/c/ONiUzizJ/5-gltf-support)

Cervator:
This has been merged via #4150! Just with rebase etc so this PR doesn't show as merged, but it is! 🎉 

Thank you @immortius and everybody else involved :-) 

Now time for continuing the cleanup with #4156 😁

PR #3786, opened 2019-11-12
Create LocationChangedEvent for moving entities

Qwertygiy:
### Contains
This PR adds two classes, LocationChangedEvent and LocationChangedSystem, as well as adds two variables to LocationComponent -- lastPosition and lastRotation.

This allows other systems to listen for a LocationChangedEvent to handle moving entities such as players, items, or creatures, without having to constantly check their own cache of previous locations.

### How to test
Create a class that implements BaseComponentSystem and subscribes for a LocationChangedEvent. It should trigger whenever a valid entity (such as the player or their camera) moves position or rotates.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/location/LocationChangedEvent.java
+++ b/engine/src/main/java/org/terasology/logic/location/LocationChangedEvent.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2019 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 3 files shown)


GooeyHub:

engine/src/main/java/org/terasology/logic/location/LocationChangedEvent.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public LocationChangedEvent(LocationComponent newLocation, Vector3f oPosition, Quat4f oRotation, Quat4f nRotation)
+    {
+        this(newLocation, oPosition, oRotation, newLocation.position, nRotation);
+    }
+
+    public LocationChangedEvent(LocationComponent nComponent, Vector3f oPosition, Quat4f oRotation, Vector3f nPosition, Quat4f nRotation)
+    {
+        oldPosition = ImmutableVector3f.createOrUse(oPosition);
```
mayant15: This would always create a new `ImmutableVector3f` because `Vector3f` and it's immutable counterpart are not related. So every time a new `LocationChangedEvent` is created, these 4 new objects will be created again.

image

Qwertygiy: The point is that it is *not* for blocks -- it is meant to be block *item* entities, players, animals, etc. This allows the existence of sensors that check for the presence of such mobile entities without requiring a collision check, or for each sensor to constantly scan all entities for their location.

I will check again and see if immutable is unnecessary, but my recollection is that it is required so that event recipients cannot modify the very mutable "last position" vector and mess up any other event recipients.

PR #3803, opened 2019-12-14
Highlight issues modul details page

godfather2:
### Contains

Colors the bad modules in the module dependencies section in red.
I have been fiddling with the code for a couple of days and it seems likely that you can't color the modules in the 'Modules' list and the text in the 'installed' label without making some serious changes to the code.

### How to test

Mess with versions on any module so that it's installed version is out of bounds.

- [ ] Colors the modules in the 'Modules' list
- [x] Colors the modules in the 'Module Dependencies' list
- [ ] Colors the text in the 'installed' label

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/moduleDetailsScreen/ModuleDetailsScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/moduleDetailsScreen/ModuleDetailsScreen.java
@@ -276,6 +276,21 @@ public class ModuleDetailsScreen extends CoreScreenLayer {
         });
     }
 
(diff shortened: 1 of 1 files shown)


GooeyHub:

DarkWeird:
1. You violate Variable Naming Conversion, using variable name which starts with uppercase.
2. You can


use getMajor() 



instead

toString ().charAt()



• it will bug with major version greater then 10.

godfather2:
Fixed the naming issue. getMajor() doesn't work for 'installed version'.I used regex instead.

e-aakash:
Thanks for the work.
But the issue has been resolve in https://github.com/MovingBlocks/Terasology/pull/3840, hence closing this PR

PR #3809, opened 2019-12-22
JNBullet

pollend:
This implementation uses a JNI wrapper around native bullet library with tweaks to the code to handle block world . there are some out standing features that need to be address and fixed before  this would be considered workable but this implementation works against the current development branch of Terasology. At the moment blocks will jitter a bit with the matrix that comes back from bullet. I'm not entirely sure what is wrong but normalizing the Quaterninion seems to address some of the jiggling problems. bullet seems stable but something is wrong with how Terasology is handling the matrix from bullet. Friction seems to address the other problem of blocks jiggling and never settling. At the moment I would like some more people to test this against the old branch and work out if there are any significant benefits against the jbullet.

## Issues

- physics entities seem to glitch and jump. the problem seems to be with how quaternions are assigned with motion state. 
  - engine/src/main/java/org/terasology/physics/bullet/EntityMotionState.java

## Previous Branches 

- https://github.com/MovingBlocks/Terasology/pull/3121
- https://github.com/MovingBlocks/Terasology/pull/3042

## Relevant Links

- https://forum.terasology.org/threads/bullet-and-box2d.2272/
- https://github.com/MovingBlocks/NativeBullet 
- https://github.com/MovingBlocks/bullet3/tree/307d59279542c23055dcba5d7e62f5372e7cf2fa


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

pollend:
here is the other part to the performance problem. bullet is fast but it's probably heavily being throttled between calls through JNI.  getvoxel takes just as much time as a step for bullet so I'm guessing the timing is heavily affecting the performance. multi-threading does not seem to improve the performance substantially so it's a memory throttling problem with JNI. I guess some other method will be needed to sync the state between bullet and java. 

image

image

GooeyHub:

PR #3812, opened 2019-12-24
Added support for mesh transformations

PAndaContron:
### Contains

Adds support for the `matrix`, `rotation`, `translation`, and `scale` properties for GLTF meshes.

### How to test

Using Blender, select the mesh, then go to `Object Properties` and open the `Transform` drop down. Adjustments to location, rotation, and scale should now affect the mesh in the game.

- [ ] More testing! I haven't tested much of this change although I have double-checked my math a few times. So far I've only tested rotations, which appear to produce the right result, however I have not tested translations and scaling, or any combination of the 3.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/gltf/GLTFMeshFormat.java
+++ b/engine/src/main/java/org/terasology/rendering/gltf/GLTFMeshFormat.java
@@ -98,8 +98,8 @@ public class GLTFMeshFormat extends GLTFCommonFormat<MeshData> {
 
     private void applyTransformations(GLTF gltf, TFloatList vertices, TFloatList normals) {
         GLTFNode node = null;
-        for(GLTFNode curr : gltf.getNodes()) {
-            if(curr.getMesh() == 0) {
+        for (GLTFNode curr : gltf.getNodes()) {
+            if (curr.getMesh() == 0) {
                 node = curr;
                 break;
             }
@@ -107,7 +107,7 @@ public class GLTFMeshFormat extends GLTFCommonFormat<MeshData> {
 
         Matrix4f transform = new Matrix4f(Matrix4f.IDENTITY);
 
-        if(node != null) {
+        if (node != null) {
             if (node.getMatrix() == null) {
                 Vector3f position = new Vector3f();
                 Quat4f rotation = new Quat4f(Quat4f.IDENTITY);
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/gltf/GLTFMeshFormat.java
```diff
@@ -86,9 +91,71 @@ public MeshData load(ResourceUrn urn, List<AssetDataFile> inputs) throws IOExcep
...
 
             readBuffer(loadedBuffers.get(indicesBuffer.getBuffer()), indicesAccessor, indicesBuffer, mesh.getIndices());
+            applyTransformations(gltf, mesh.getVertices(), mesh.getNormals());
             return mesh;
         }
     }
 
+    private void applyTransformations(GLTF gltf, TFloatList vertices, TFloatList normals) {
+        GLTFNode node = null;
+        for(GLTFNode curr : gltf.getNodes()) {
```
Cervator: Minor whitespace style issues here and elsewhere, make sure you have Checkstyle hooked up and run it in the IDE against modified files 👍

PAndaContron: I fixed the issues in the file I modified, should I fix the rest of the GLTF checkstyle issues as well or leave those for now?

GooeyHub:

engine/src/main/java/org/terasology/rendering/gltf/GLTFMeshFormat.java
Cervator: Thanks! Generally if you touch a file feel free to fix other uncontroversial style fixes within it, so long as that wouldn't overwhelm the diff (make it hard to spot your true changes). Same with related files, but the further they get from the core changes and the more changes overall the more likely it'll be a better idea to do it in a different PR. Bit of a judgement call case by case.

In this case you're adding to an unmerged PR so it may be wise to avoid more change in case the other PR gets modified again. But any _new_ code you introduce at least wouldn't add any more diffs to style right :-)

tl;dr: I'd leave the rest as-is in this case :)

PAndaContron:
I added support for transformations inherited from parent nodes. The mesh's node transformation is applied first, then each parent node going up the tree. As usual, I have not yet fully tested this, although I can confirm that it does not have any effect on any mesh nodes without parents.

GooeyHub:

Cervator:
This has been merged via #4150! Just with rebase etc so this PR doesn't show as merged, but it is! 🎉

PR #3813, opened 2019-12-25
Fix for Z-fighting on back faces of transparent blocks

Qwertygiy:
This prevents the interior faces of partially transparent blocks from being merged into the exterior faces of the blocks behind them.

It would be wise to only adjust interior faces, and not exterior faces, but as of yet I have been unable to determine the proper way to accomplish this. This means that currently, it will not fix z-fighting between two blocks that are both partially transparent.

To test this, create or use a block marked "translucent" with the full cube shape, and a texture that is partially transparent so that you can see inside the block. When placed adjacent to a solid block, the interior face of the transparent block should be fully visible and overlaying the face of the block behind it.

### DIFF
--- a/engine/src/main/resources/assets/shaders/chunk_vert.glsl
+++ b/engine/src/main/resources/assets/shaders/chunk_vert.glsl
@@ -178,4 +178,11 @@ void main()
 
     vertexProjPos = gl_ProjectionMatrix * vertexViewPos;
     gl_Position = vertexProjPos;
+
+#if defined (FEATURE_ALPHA_REJECT)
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Tested out with the ice block and could validate the before-after fix on this, so - merged! :-)

At first I wasn't super sure if it at all related to the weirdness of how some blocks look when held, but no change there. Not actually sure if we have an issue on it or if it just floating around as tribal knowledge (maybe even just stuck in my head)

Terasology-191228202155-1152x720

Issue #3566, opened 2018-12-02
Attempt to handle unexpected (old or weird) OSes that may not support our directory detection


Cervator:
### What you were trying to do

A user attempted to run the game on a Windows XP system. Yes, WinXP still exists. No, it isn't likely to run the game well even if it didn't instantly crash :-)

It made me think though that there might be other current (or future) OSes we might encounter and maybe we can harden the process for a second chance at being able to run the game.

### What actually happened

`Exception in thread "main" java.lang.UnsatisfiedLinkError: Error looking up function 'SHGetKnownFolderPath': The specified procedure could not be found.` - apparently XP can't handle the Windows specific call trying to get a default directory to store game data.

### How to reproduce

* Run the game on Windows XP. If you can find something with XP on it

### Possible improvement

I've come up with two brief ideas that might be able to get around this issue, which is namely using native/OS specific code trying to get a game data dir to call default.

* If path detection fails fall back on showing the user a popup asking for a valid path to use (similar to what the game launcher does) - so wrap exception detection around the problematic call then attempt a fallback
* If the user supplies -homedir then skip that whole part of the code. Update the docs to say if you hit an issue like that try to supply -homedir yourself

SanchitUke:
Hi!
I would like to work on this issue. I have not done any exception handling before. Can you direct me where to start? Thank you

sladyn98:
@Cervator Looks like a good beginner issue to work on. Can I start working on this

Adrijaned:
Hi @sladyn98, feel free to start working on this, as as far as I know noone else is yet.

sladyn98:
https://github.com/MovingBlocks/Terasology/blob/4e95d495f21d5a687a48d1edb0f389fb66d7e1cc/facades/PC/src/main/java/org/terasology/engine/Terasology.java#L365

My plan was to add a FileChooser to the above point and instead of catching an IO Exception would catch a general Exception, thereby catching the OS error.This would give the user a chance to select the file path and thus proceed normally.
```JFileChooser jfc = new JFileChooser(FileSystemView.getFileSystemView().getHomeDirectory());```

DarkWeird:
@sladyn98 its will be problem in headless server

sladyn98:
@DarkWeird So in that case we could use the fallback path that has been defined for the operating system.

DarkWeird:
@sladyn98 
I don't think what use homedir - pretty idea.
Looks like game files directly in homedir.
[...]

PR #3823, opened 2020-01-13
Handle unexpected (old or weird) OSes that may not support our directory detection.

sladyn98:
### Contains

This PR deals with errors thrown during directory detection, specifically giving a chance to run the game again if the default directory detection fails.
In case of headless servers we have defaulted to choosing the home path as the current directory.
Fixes https://github.com/MovingBlocks/Terasology/issues/3566

### How to test

Run the game on Windows XP. If you can find something with XP on it or a wierd OS that throws our directory detection out of the window.

- [ x ] Still have to adjust the wiki doc 
The doc has to be updated if in case we hit any such OSes in the future.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/paths/PathManager.java
+++ b/engine/src/main/java/org/terasology/engine/paths/PathManager.java
@@ -32,6 +32,9 @@ import java.nio.file.Paths;
 import java.util.List;
 
 import javax.swing.JFileChooser;
+import org.terasology.context.Context;
+import org.terasology.engine.subsystem.DisplayDevice;
+
 
 /**
  * Manager class that keeps track of the game's various paths and save directories.
@@ -52,6 +55,8 @@ public final class PathManager {
     private static final String REGEX = "[^A-Za-z0-9-_ ]";
 
     private static PathManager instance;
+
+    private static Context context;
     private Path installPath;
     private Path homePath;
     private Path savesPath;
@@ -151,7 +156,7 @@ public final class PathManager {
      * @param rootPath Path to use as the home path.
      * @throws IOException Thrown when required directories cannot be accessed.
      */
-    public void useOverrideHomePath(Path rootPath) throws Exception {
+    public void useOverrideHomePath(Path rootPath) throws IOException {
         this.homePath = rootPath;
         updateDirs();
     }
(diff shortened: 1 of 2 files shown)


GooeyHub:

Qwertygiy:
@GooeyHub add to whitelist.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -363,6 +364,12 @@ private static void handleLaunchArguments(String[] args) {
 
         } catch (IOException e) {
             reportException(e);
+            try {
+                PathManager.getInstance().chooseHomePathManually();
+            } catch (IOException ex) {
+                reportException(e);
```
SoniEx2: ex/e?

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -363,6 +364,12 @@ private static void handleLaunchArguments(String[] args) {
...
+                System.exit(0);
+            }
             System.exit(0);
```
SoniEx2: shouldn't this be taken out?

facades/PC/src/main/java/org/terasology/engine/Terasology.java
sladyn98: Yeah Correcting :+1:

facades/PC/src/main/java/org/terasology/engine/Terasology.java
sladyn98: Yes it should :+1:

engine/src/main/java/org/terasology/engine/paths/PathManager.java
```diff
@@ -188,6 +196,24 @@ public void useDefaultHomePath() throws IOException {
         updateDirs();
     }
 
+    /**
+     * Gives user the option to manually choose home path.
+     * @throws IOException Thrown when required directories cannot be accessed.
+     */
+    public void chooseHomePathManually() throws IOException {
```
skaldarnar: Should/can this be part of `useDefaultHomePath`?

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -363,7 +364,12 @@ private static void handleLaunchArguments(String[] args) {
 
         } catch (IOException e) {
             reportException(e);
-            System.exit(0);
+            try {
+                PathManager.getInstance().chooseHomePathManually();
```
skaldarnar: If this can recover any exception that was thrown before, there is no need to `reportException` before trying the backup. At least, it should not be reported as an error (maybe as a warning, or probably just on INFO level).

sladyn98: Yeah , you have a valid point I just thought that it would be good to report an exception without stopping execution. A Warning sounds like a much better approach  :)

engine/src/main/java/org/terasology/engine/paths/PathManager.java
sladyn98: It could be but then it would have to catch the exception inside the `useDefaultHomePath`, moreover this provides a much modular approach and a generic solution to all further weird OSeS

PR #3825, opened 2020-01-16
Add a RunsAfter annotation to allow ordering of rasterizers

askneller:
### Contains

Adds a `RunsAfter` annotation that indicates one rasterizer must be run after another.

### How to test

Add the annotation to a rasterizer. E.g.
```
@RegisterPlugin
@RunsAfter(CaveRasterizer.class)
public class SomeRasterizer implements WorldRasterizerPlugin {
```
During world generation you can place a breakpoint at `WorldBuilder` line 108. When it hits the breakpoint inspect the `orderedRasterizers` list and confirm that `SomeRasterizer` is after `CaveRasterizer`.


GooeyHub:

Qwertygiy:
Looks good to me. There is one minor thing; I feel like the name should imply that the annotated rasterizer *requires* the other rasterizer to be run before it can be run, much as the Requires annotation exists for facets. 

This is a good solution for situations similar to that, where the rasterizers could not logically be run without the presence of their predecessor, like the CaveFloor generation not doing anyone a lick of good without a Cave existing! But the name RunsAfter could suggest that there's no dependency involved and that it's merely for ordering purposes, which it shouldn't be.

askneller:
Good point. How about I change it to something like `RequiresRasterizer` so that it doesn't conflict with the existing `Requires`?

Qwertygiy:
Sounds perfect to me!

GooeyHub:

Issue #3760, opened 2019-10-16
Feature: Possibility to search for LAN games


qwc:
It should be possible to search for local area network games automatically. 
Some extra button in the server list. 

(Might refine that issue later)

Cervator:
Since this is a fairly well understood topic I think I'd just go search for and ~steal~ draw inspiration from any sort of similar implementation we can find :-) There might even be some useful snippets in Java somewhere already!

This might be easy enough to do if such a search finds a good example and we can just plug in another tab or button on the join game screen

sladyn98:
I would like to work on this issue
https://www.baeldung.com/java-broadcast-multicast This looks like a decent implementation for broadcasting.

pollend:
go ahead, we always like new PR's :P.

sladyn98:
I thought of the following approach for this.
We could add a `broadCastServer.java` class which would broadcast the game on the local area network, along with the port and the modules available. This list of servers would then show up in the join game server. This could be supported by a button that says `Find local area games` and a button to activate/deactivate the broadcast.
cc @DarkWeird

DarkWeird:
@sladyn98 yeah. Sounds nice.
What about ipv4/v6, headless server and broadcast ip (nitpick, ICANN determinate some ip addresses for broadcast)?

sladyn98:
IPV4 supports broadcasting, In IPv6, multicast support has been made mandatory, and there is no explicit concept of broadcasting,in multicasting unfortunately we have to make the recipients get added to the group to be able to receive the messages so we might have to stick with just broadcasting.
For headless servers the code for stays the same with the `ip.getHostAddress()` becoming the host IP

DarkWeird:
@sladyn98 wait your PR :3

ahv15:
Could I try working on this issue, based off how the previous assignee had attempted it. TBH I am not too sure how this works exactly but this seemed like one of the few beginner issues that nobody is working on for the time being and any help would be appreciated.

Cervator:
Heya @ahv15 - yes, please! In this case there's an existing PR you can look at for inspiration, and all the comments to learn more about it, what it missing, and so on. Plus it would be great to get that work salvaged and the feature completed :-)

ahv15:
Ok sure, I will take a shot at it.

skaldarnar:
@ahv15 also, make sure to hop on Discord if you have questions or are stuck somewhere 😉

ahv15:
[...]

PR #3828, opened 2020-01-25
Ability to Broadcast games on the Local Area Network

sladyn98:
Ability to Broadcast games on the Local Area Network

### Contains

`JoinGameScreen.java` - Added the broadcast button to be enabled only when the server is not broadcasting, along with a custom message of all the modules, ports and IpAddresse's

`Broadcast.java` - Added the ability to broadcast the game on all local listening servers.
The message is broadcasted every 5 minutes.
"Fixes https://github.com/MovingBlocks/Terasology/issues/3760"

### How to test

The broadcast button is added to the User Interface and the respective broadcasted message can be read on the Join Game Server on a machine present on the LAN.

- :heavy_check_mark:  UI button for broadcasting
- :heavy_check_mark: Make sure Broadcasting is enabled for headless server.
-  [ ]  Need to add documentation for this.

### DIFF
--- a/engine/src/main/java/org/terasology/network/BroadCastServer.java
+++ b/engine/src/main/java/org/terasology/network/BroadCastServer.java
@@ -27,8 +27,8 @@ public class BroadCastServer {
 
     private static boolean turnOnBroadcast;
     private static final Logger logger = LoggerFactory.getLogger(BroadCastServer.class);
-    private static final String DISCOVERY_REQUEST = "DISCOVERY_REQUEST";
-    private static final String DISCOVERY_RESPONSE = "DISCOVERY_RESPONSE";
+    private  final String discoveryRequest = "DISCOVERY_REQUEST";
+    private  final String discoveryResponse = "DISCOVERY_RESPONSE";
     private  DatagramSocket receiveSocket;
     private  DatagramSocket sendSocket;
 
@@ -70,16 +70,16 @@ public class BroadCastServer {
     }
 
     private void listenToBroadCast() throws Exception {
-        byte[] buffer = DISCOVERY_REQUEST.getBytes(StandardCharsets.UTF_8);
+        byte[] buffer = discoveryRequest.getBytes(StandardCharsets.UTF_8);
         DatagramPacket packet = new DatagramPacket(buffer, buffer.length);
         getReceiveSocket().receive(packet);
         logger.info("Discovery package received! -> " + packet.getAddress() + ":" + packet.getPort());
 
         //Validate the data sent.
         String data = new String(packet.getData()).trim();
-        if (data.equals(DISCOVERY_REQUEST)) { // validate command
+        if (data.equals(discoveryRequest)) { // validate command
             // Send response
-            byte[] response = new byte[DISCOVERY_RESPONSE.length()];
+            byte[] response = new byte[discoveryResponse.length()];
             DatagramPacket responsePacket = new DatagramPacket(response, response.length, packet.getAddress(), packet.getPort());
             getReceiveSocket().send(responsePacket);
             logger.info("Response sent to: " + packet.getAddress() + ":" + packet.getPort());
@@ -115,18 +115,18 @@ public class BroadCastServer {
 
     public void sendBroadCast() throws Exception {
         // Discovery request command
-        byte[] data = DISCOVERY_REQUEST.getBytes(StandardCharsets.UTF_8);
+        byte[] data = discoveryRequest.getBytes(StandardCharsets.UTF_8);
         DatagramPacket packet = new DatagramPacket(data, data.length, InetAddress.getByName("255.255.255.255"), 8002);
         getSendSocket().send(packet);
         logger.info("Discovery package sent!" + packet.getAddress() + ":" + packet.getPort());
 
         // Discovery response command
-        byte[] response = new byte[DISCOVERY_RESPONSE.length()];
+        byte[] response = new byte[discoveryResponse.length()];
         DatagramPacket responsePacket = new DatagramPacket(response, response.length);
         getSendSocket().receive(responsePacket);
         logger.info("Discovery response received!" + responsePacket.getAddress() + ":" + responsePacket.getPort());
         String responseData = new String(responsePacket.getData(), StandardCharsets.UTF_8);
-        if (responseData.equals(DISCOVERY_RESPONSE)) { // Check valid command
+        if (responseData.equals(discoveryResponse)) { // Check valid command
             logger.info("Found server!" + responsePacket.getAddress() + ":" + responsePacket.getPort());
         }
     }
--- a/engine/src/main/java/org/terasology/network/BroadCastServer.java
+++ b/engine/src/main/java/org/terasology/network/BroadCastServer.java
@@ -38,10 +38,10 @@ public class BroadCastServer {
     private static DatagramSocket socket;
     private static boolean turnOnBroadcast;
     private static final Logger logger = LoggerFactory.getLogger(JoinGameScreen.class);
-    private NetworkSystem networkSystem;
+    private static ScheduledExecutorService service;
 
 
-    public boolean isTurnOnBroadcast() {
+    public boolean isBroadCastTurnedOn() {
         return turnOnBroadcast;
     }
 
@@ -50,6 +50,7 @@ public class BroadCastServer {
     }
 
     public void startBroadcast(String message) {
+        service = Executors.newSingleThreadScheduledExecutor();
         Runnable runnable = new Runnable() {
             public void run() {
                 try {
@@ -58,13 +59,17 @@ public class BroadCastServer {
                         broadcast(message, inadr);
                     }
                 } catch (Exception e) {
+                    logger.info("Broadcasting has been interrupted " + e.getMessage());
                     e.printStackTrace();
                 }
             }
         };
-        ScheduledExecutorService service = Executors
-            .newSingleThreadScheduledExecutor();
         service.scheduleAtFixedRate(runnable, 0, 300, TimeUnit.SECONDS);
+    }
+
+    public void stopBroadCast() {
+        logger.info("Shutting down BroadCasting");
+        service.shutdown();
     }
 
     /**
--- a/engine/src/main/java/org/terasology/network/BroadCastServer.java
+++ b/engine/src/main/java/org/terasology/network/BroadCastServer.java
@@ -42,19 +42,20 @@ public class BroadCastServer {
         BroadCastServer.turnOnBroadcast = true;
         try {
             sendBroadCast();
-            service = Executors.newSingleThreadScheduledExecutor();
-            Runnable runnable = new Runnable() {
+            Executors.newSingleThreadExecutor().execute(new Runnable() {
+                @Override
                 public void run() {
-                    logger.info("Listening to broadcast");
-                    try {
-                        while(turnOnBroadcast) {
-                            listenToBroadCast();
+                    while (true) {
+                        try {
+                            while (turnOnBroadcast) {
+                                listenToBroadCast();
+                            }
+                        } catch (Exception e) {
+                            logger.error("Broadcasting has been interrupted" + e.getMessage());
                         }
-                    } catch (Exception e) {
-                        logger.error("Broadcasting has been interrupted" + e.getMessage());
                     }
                 }
-            };
+            });
         } catch (Exception e) {
         logger.info("Broadcasting has been interrupted " + e.getMessage());
         }
--- a/engine/src/main/java/org/terasology/network/BroadCastServer.java
+++ b/engine/src/main/java/org/terasology/network/BroadCastServer.java
@@ -33,6 +33,8 @@ public class BroadCastServer {
     private static boolean turnOnBroadcast;
     private static final Logger logger = LoggerFactory.getLogger(JoinGameScreen.class);
     private static ScheduledExecutorService service;
+    private static String discoveryRequest = "__DISCOVERY_REQUEST__";
+    private static String discoveryResponse = "__DISCOVERY_RESPONSE__";
 
     public boolean isBroadCastTurnedOn() {
         return turnOnBroadcast;
@@ -69,16 +71,16 @@ public class BroadCastServer {
     }
 
     private void listenToBroadCast() throws Exception {
-        byte[] buffer = "__DISCOVERY_REQUEST__".getBytes();
+        byte[] buffer = discoveryRequest.getBytes();
         DatagramPacket packet = new DatagramPacket(buffer, buffer.length);
         getReceiveSocket().receive(packet);
         logger.info("Discovery package received! -> " + packet.getAddress() + ":" + packet.getPort());
 
         //Validate the data sent.
         String data = new String(packet.getData()).trim();
-        if (data.equals("__DISCOVERY_REQUEST__")) { // validate command
+        if (data.equals(discoveryRequest)) { // validate command
             // Send response
-            byte[] response = new byte["__DISCOVERY_RESPONSE".length()];
+            byte[] response = new byte[discoveryResponse.length()];
             DatagramPacket responsePacket = new DatagramPacket(response, response.length, packet.getAddress(), packet.getPort());
             getReceiveSocket().send(responsePacket);
             logger.info("Response sent to: " + packet.getAddress() + ":" + packet.getPort());
@@ -114,18 +116,18 @@ public class BroadCastServer {
 
     public void sendBroadCast() throws Exception {
         // Discovery request command
-        byte[] data = "__DISCOVERY_REQUEST__".getBytes();
+        byte[] data = discoveryRequest.getBytes();
         DatagramPacket packet = new DatagramPacket(data, data.length, InetAddress.getByName("255.255.255.255"), 8002);
         getSendSocket().send(packet);
         logger.info("Discovery package sent!" + packet.getAddress() + ":" + packet.getPort());
 
         // Discovery response command
-        byte[] response = new byte["__DISCOVERY_RESPONSE__".length()];
+        byte[] response = new byte[discoveryResponse.length()];
         DatagramPacket responsePacket = new DatagramPacket(response, response.length);
         getSendSocket().receive(responsePacket);
         logger.info("Discovery response received!" + responsePacket.getAddress() + ":" + responsePacket.getPort());
         String responseData = new String(responsePacket.getData()).trim();
-        if (responseData.equals("__DISCOVERY_RESPONSE__")) { // Check valid command
+        if (responseData.equals(discoveryResponse)) { // Check valid command
             logger.info("Found server!" + responsePacket.getAddress() + ":" + responsePacket.getPort());
         }
     }
--- a/engine/src/main/java/org/terasology/network/BroadCastServer.java
+++ b/engine/src/main/java/org/terasology/network/BroadCastServer.java
@@ -30,12 +30,17 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.terasology.config.Config;
+import org.terasology.registry.In;
 import org.terasology.rendering.nui.layers.mainMenu.JoinGameScreen;
 
 
 public class BroadCastServer {
 
     private static DatagramSocket socket;
+
+    @In
+    private Config config;
     private static boolean turnOnBroadcast;
     private static final Logger logger = LoggerFactory.getLogger(JoinGameScreen.class);
     private static ScheduledExecutorService service;
@@ -49,7 +54,8 @@ public class BroadCastServer {
         BroadCastServer.turnOnBroadcast = turnOnBroadcast;
     }
 
-    public void startBroadcast(String message) {
+    public void startBroadcast() {
+        String message = buildBroadCastMessage();
         service = Executors.newSingleThreadScheduledExecutor();
         Runnable runnable = new Runnable() {
             public void run() {
@@ -109,6 +115,13 @@ public class BroadCastServer {
                 .forEach(broadcastList::add);
         }
         return broadcastList;
+    }
+
+
+    private String buildBroadCastMessage() {
+        logger.info("Building Broadcast Message");
+        int serverPort = config.getNetwork().getServerPort();
+        return "Hello I am running on server" + config.getNetwork().getMasterServer() + "and Port:" + serverPort;
     }
 
 }
--- a/engine/src/main/java/org/terasology/network/BroadCastServer.java
+++ b/engine/src/main/java/org/terasology/network/BroadCastServer.java
@@ -20,19 +20,18 @@ import java.net.DatagramSocket;
 import java.net.InetAddress;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class BroadCastServer {
 
-    private static DatagramSocket receiveSocket;
-    private static DatagramSocket sendSocket;
     private static boolean turnOnBroadcast;
     private static final Logger logger = LoggerFactory.getLogger(BroadCastServer.class);
-    private static ScheduledExecutorService service;
     private static final String DISCOVERY_REQUEST = "DISCOVERY_REQUEST";
     private static final String DISCOVERY_RESPONSE = "DISCOVERY_RESPONSE";
+    private  DatagramSocket receiveSocket;
+    private  DatagramSocket sendSocket;
+
 
     public boolean isBroadCastTurnedOn() {
         return turnOnBroadcast;
@@ -45,26 +44,29 @@ public class BroadCastServer {
             Executors.newSingleThreadExecutor().execute(new Runnable() {
                 @Override
                 public void run() {
-                    while (true) {
+                    while (!Thread.currentThread().isInterrupted()) {
                         try {
                             while (turnOnBroadcast) {
                                 listenToBroadCast();
                             }
                         } catch (Exception e) {
                             logger.error("Broadcasting has been interrupted" + e.getMessage());
+                            stopBroadCast();
                         }
                     }
                 }
             });
         } catch (Exception e) {
         logger.info("Broadcasting has been interrupted " + e.getMessage());
+        stopBroadCast();
         }
     }
 
     public void stopBroadCast() {
         BroadCastServer.turnOnBroadcast = false;
         logger.info("Shutting down BroadCasting");
-        service.shutdown();
+        this.sendSocket.close();
+        this.receiveSocket.close();
     }
 
     private void listenToBroadCast() throws Exception {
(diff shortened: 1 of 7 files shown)


GooeyHub:

sladyn98:
CC @DarkWeird

DarkWeird:
It's wip now?
I don't sure what using all upped non loopback interface is good idea. I haven't idea how make this part better :3
Please, add todo list to Outstanding before merging.

sladyn98:
@DarkWeird I added the Button for the broadcast but when I run the game it does not show up , am I missing something ?

DarkWeird:
I am not familiar with NUI.
Use it:
https://github.com/Terasology/TutorialNui

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
```diff
@@ -352,6 +357,21 @@ public Boolean get() {
                 refresh();
             });
         }
+
+        UIButton broadCastServerButton = find("broadcast", UIButton.class);
+        if (broadCastServerButton != null) {
+            broadCastServerButton.bindEnabled(new ReadOnlyBinding<Boolean>() {
+                @Override
+                public Boolean get() {
+                    return broadCastServer.isTurnOnBroadcast();
```
Torpedo99:
```suggestion
                    return !broadCastServer.isTurnOnBroadcast();
```
I think you probably would want to enable the button when not broadcasting and disable it when it is.

sladyn98: Yeah thanks for the suggestion :+1:

DarkWeird:
Don't forget cleanup resources (shutdown executor on exit)
Maybe better integrate server to network subsystem, or separate.
Be careful, we have some works with nui extraction in separate lib.
I wanna see your code, working on headless server too :3

sladyn98:
> Don't forget cleanup resources (shutdown executor on exit)

Yes adding it :)

> Maybe better integrate server to network subsystem, or separate.

With respect to integration with integration with network subsytem, this is already separate module isn't it ?
> Be careful, we have some works with nui extraction in separate lib.

Yeah  it does not break anything for now, Do I need to take care of something specific ?
> I wanna see your code, working on headless server too :3

Working on the headless server part,

DarkWeird:
I think better implement as part of network system, or near.

Your broadcast placed at nui's package now :3

DarkWeird:
1. Fix imports, plz (broadcastserver have import on nui)
2. Why broadcastserver in headless setup use DI stuffs(its work?), in nui not
3. What with multiple network interfaces?
If you bind server on all interfaces - it is 0.0.0.0 - unreachable target port. You need broadcast every interface ip.
If specific(one)- you not need broadcast to others network.

SoniEx2:
maybe use UPnP or something?

sladyn98:
@DarkWeird 

> 2. Why broadcastserver in headless setup use DI stuffs(its work?), in nui not

I am not sure what DI stuff, could you explain what do you mean ? 

>If you bind server on all interfaces - it is 0.0.0.0 - unreachable target port. You need broadcast every interface ip.

```
   Vector<String> container=new Vector<String>();
        container.add(message);
        String m=classname.getMYIP();
               //here u have your own IP address..now suppose your IP is 192.168.1.1.basically m sending hello packet to every other machine.if that machine is active ,it will reply n we get to know that machine is active......
               for(int i=1;i<255;i++)
           {
                        try
                               {
                     
                    String n="192.168.1."+i;//192.168.1.1,192.168.1.2 n so on.........to 192.168.1.255
                     
                        new Thread(new client(n,container)).start();//create client threads n send hello packet         to every other machine......
                        System.out.println(i);
                     
                }catch(Exception e){e.printStackTrace();}
             
        }
```

We could do this , it is a much brute force approach and only checks for 255 machines.
What we are doing here is basically

- Determine your own IP address
- Determine your own netmask
- Scan all the addresses (except the lowest, which is your network address and the highest, which is your broadcast address)

sladyn98:
> maybe use UPnP or something?

I guess this is very susceptible to attacks, it is a bit of a dangerous approach ?

SoniEx2:
properly configured UPnP isn't any more risky than UDP broadcast.

improperly configured UPnP doesn't suddenly become more dangerous because apps are using it. improperly configured UPnP is inherently dangerous, but doesn't affect the safety of terasology.

whether UPnP is properly or improperly configured depends on the router, not on terasology.

DarkWeird:
@sladyn98 
1. Dependency injection. In our case - @In annotation.
2. I say about several network cards (wifi/lan, several lan)
You can start server on all network interface(use ip 0.0.0.0, as ip for bind server) or specific(192.168.1.1 (lan for example) or 10.0.1.1(wifi for example))
If you start server on wifi, your code will broadcast invalid ip for lan(for wifi it bradcast too)
If you start on any network card(interface), your code will broadcast invalid ip for one of network interface(lan or wifi)

Maybe, some problem will be with docker, vpn, etc. With all what generate virtual network interfaces, if it installed/used on system

DarkWeird:
@SoniEx2 i think implemention of UPnP can be as separate PR. 
IIRC, UPnP configure router's routing from WAN to LAN, isn't it?

Hmm, broadcast to WAN ?

SoniEx2:
Please read up on UPnP https://en.wikipedia.org/wiki/Universal_Plug_and_Play

It is a more suitable solution to this problem.

sladyn98:
> Please read up on UPnP https://en.wikipedia.org/wiki/Universal_Plug_and_Play
> 
> It is a more suitable solution to this problem.

Thank you so much for the feedback, really appreciate it. Will Read up on this.

sladyn98:
@DarkWeird Looking at the dependency injection, rather I left it to the class to be responsible for the creation for the object. It would work fine, if i hard code the object creation, If you suggest that is a better approach, I am definitely on board :)

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -78,50 +77,46 @@ public void stopBroadCast() {
...
+        // Send from localhost:1901
+        InetSocketAddress srcAddress = new InetSocketAddress(localhost, ssdpSearchPort);
+        // Send to 239.255.255.250:1900
+        InetSocketAddress dstAddress = new InetSocketAddress(InetAddress.getByName(ssdpIP), ssdpPort);
+        // ----------------------------------------- //
+        //       Construct the request packet.       //
+        // ----------------------------------------- //
+        StringBuffer discoveryMessage = new StringBuffer();
+        discoveryMessage.append("M-SEARCH * HTTP/1.1\r\n");
+        discoveryMessage.append("HOST: " + ssdpIP + ":" + ssdpPort + "\r\n");
```
sladyn98: This looks to be sufficient in order to broadcast both the port and address.

sladyn98:
CC @DarkWeird @SoniEx2

DarkWeird:
Upnp Looks good. 
@SoniEx2
Is it working without router?
It can work in vpn? (Hamachi e.g)

Localhost as source address looks odd. Its working?

DarkWeird:
Don't forget change System.out to logger usage

sladyn98:
> Is it working without router?

Yes it is working with router. I checked the original program with wireshark and it got the packets.

sladyn98:
> It can work in vpn? (Hamachi e.g)

IIUC VPN has the purpose of connecting networks so if we  want to reach another system via a VPN you will have to establish a network connection , a Java API for all of this protocols will be (nearly) impossible, since VPN is handled by OS drivers and not on the application level (where java has its place) in most cases.
Although we could detect if a vpn is up and then try and stop the broadcast since outer connection would not be possible if that is the use case we are targeting.
```
try {
    for( NetworkInterface intf : Collections.list(NetworkInterface.getNetworkInterfaces())) {
        // Pass over dormant interfaces
        if(!intf.isUp() || intf.getInterfaceAddresses().size() == 0)
            continue;
            if ("tun0".equals(intf.getName())){
                // The VPN is up
                break;
            } }}
```

DarkWeird:
@sladyn98 using tun0 - bad idea.
Different vpn clients use different names, and can setup different names.

You don't must think about special handler for vpn(on application level they looks like common network).
I want know, your code can properly work with vpn? (Some guys want play together on self private server, but haven't providers, which open port/ip for connection from external net)

SoniEx2:
please use an UPnP library. it'll also save you figuring out the IP. it'll handle all that for you.

sladyn98:
> please use an UPnP library. it'll also save you figuring out the IP. it'll handle all that for you.

Appreciate the feedback, do you mean cling?

SoniEx2:
I'd suggest using https://github.com/cybergarage/cybergarage-upnp because that's still maintained.

I2P uses it, so it's probably gonna be maintained for a long time.

sladyn98:
@SoniEx2 Yeah had a look :) I guess this implementation would suffice for now, what do you think ?

sladyn98:
> @sladyn98 using tun0 - bad idea.
> Different vpn clients use different names, and can setup different names.
> 
> You don't must think about special handler for vpn(on application level they looks like common network).
> I want know, your code can properly work with vpn? (Some guys want play together on self private server, but haven't providers, which open port/ip for connection from external net)

Yeah this would work over private networks smoothly :+1:

DarkWeird:
1. Remove unused channel field
2. Move strings(request, response) to static
3. Add explicit encoding for byte-string transformation(toBytes,new String)

sladyn98:
> 3. Add explicit encoding for byte-string transformation(toBytes,new String)

@DarkWeird Could you let me know where exactly do you prefer this encoding. The `getByte` Method does do the same isn't it

DarkWeird:
This methods using system default encoding (windows-125* on windows, utf-8 on linux(or any other encoding, if user has very interesting ideas :) )).
One string with different encodings can   represent as different bytes.

Further, it has not zero chance to use different, non equals encoding.

Oh, java encoding called as charset.

My(and many libraries authors) recommendation: explicitly use UTF-8 ( new String(bytes, charset), getBytes(charset))

sladyn98:
@DarkWeird I added the encoding. Thanks

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
```diff
@@ -67,12 +69,12 @@
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
-/**
- */
 public class JoinGameScreen extends CoreScreenLayer {
     public static final ResourceUrn ASSET_URI = new ResourceUrn("engine:joinGameScreen");
 
     private static final Logger logger = LoggerFactory.getLogger(JoinGameScreen.class);
+    private Channel channel;
```
DarkWeird: unused fields.

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,135 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.terasology.rendering.nui.layers.mainMenu.JoinGameScreen;
+
+
+public class BroadCastServer {
+
+    private static DatagramSocket receiveSocket;
+    private static DatagramSocket sendSocket;
+    private static boolean turnOnBroadcast;
```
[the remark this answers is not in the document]
sladyn98: Thanks got it. I was getting used to the DI framework hence I made its field static, now I understand

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,135 @@
...
+        try {
+            if (sendSocket == null) {
+                sendSocket = new DatagramSocket(8001, InetAddress.getLocalHost());
+                sendSocket.setBroadcast(true);
+            }
+        } catch (Exception e) {
+            logger.error("Broadcast Exception Encountered" + e.getMessage());
+        }
+        return sendSocket;
+    }
```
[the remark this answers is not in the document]
sladyn98: We won't be needing a new socket right? Because we would be using the same one to send the request. Do you think there is a better way to do this?

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public void stopBroadCast() {
+        logger.info("Shutting down BroadCasting");
+        service.shutdown();
+    }
+
+    private void listenToBroadCast() throws Exception {
+        byte[] buffer = discoveryRequest.getBytes(StandardCharsets.UTF_8);
+        DatagramPacket packet = new DatagramPacket(buffer, buffer.length);
+        getReceiveSocket().receive(packet);
```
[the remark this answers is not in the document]
sladyn98: Is there any other way to keep listening?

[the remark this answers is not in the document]
DarkWeird: For blocking  input-output - while-loop with condition onstop(you can use your boolean turnOn)

engine/src/main/java/org/terasology/network/BroadCastServer.java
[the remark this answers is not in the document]
DarkWeird: I'm would do separate socket intialization and getters, because it can throw exception.
If it will throw exception, then broadcast feature cannot will work properly at least.
If you will use one class instance - you use one socket.
Additional, You can split client and server functionality

engine/src/main/java/org/terasology/network/BroadCastServer.java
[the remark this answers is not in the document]
sladyn98: Sounds cool:+1

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,135 @@
...
+                        listenToBroadCast();
+                    } catch (Exception e) {
+                        logger.error("Broadcasting has been interrupted" + e.getMessage());
+                    }
+                }
+            };
+            service.scheduleAtFixedRate(runnable, 0, 300, TimeUnit.SECONDS);
+        } catch (Exception e) {
+        logger.info("Broadcasting has been interrupted " + e.getMessage());
+        e.printStackTrace();
```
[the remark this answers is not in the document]
DarkWeird: I suggest found info about logging level and exceptions handling for youself. Something like a best practices or similar.

My english so bad and knowledge not well-formed for explanations. :(

engine/src/main/java/org/terasology/network/BroadCastServer.java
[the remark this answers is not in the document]
sladyn98: In my opinion this code should catch broadcast errors if the sockets are null or the ports are unavailable so that the broadcast can be terminated immediately. I do not think we need more modularity  in terms of client and server functionality because this offers us the best error handling.

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,135 @@
...
+        return turnOnBroadcast;
+    }
+
+    public void setTurnOnBroadcast(boolean turnOnBroadcast) {
+        BroadCastServer.turnOnBroadcast = turnOnBroadcast;
+    }
+
+    public void startBroadcast() {
+        try {
+            sendBroadCast();
```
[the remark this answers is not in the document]
sladyn98: Yes that would ensure that the responses are caught in that thread. We could start a separate thread to hear broadcasts if thats the best way forward. What do you think?

sladyn98:
@DarkWierd got most of the changes committed :)

DarkWeird:
@sladyn98 
Your Runnable runnable with listenToBroadCast not using now. 

I don't think what PR done.
You print response in log. How is regular player can use your feature? Regular player don't use console.

sladyn98:
@DarkWeird 
> You print response in log. How is regular player can use your feature? Regular player don't use console.

Yeah this was something I was waiting for feedback on. Do we want to build a separate area for displaying the servers or just show up on the JoinGameScreen.?

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,131 @@
...
+                        try {
+                            while (turnOnBroadcast) {
+                                listenToBroadCast();
+                            }
+                        } catch (Exception e) {
+                            logger.error("Broadcasting has been interrupted" + e.getMessage());
+                        }
+                    }
+                }
+            });
```
DarkWeird: 1. Every invoke ```startBroadcast()``` will produce new unclosable Thread within ```newSingleThreadExecutor```
    With '2'(see next) you can eat all your CPU(or problem with shared access to listening socket), if you will click many times on button.
2. Using ```while (true)``` -  bad idea. It's Busy waiting (I googled it now. it said - anti-pattern :) ) 
3. You don't close sockets, listening socket will be listen while it not recieve packet or program exited, even  your turn broadcast off. 


I see two ways to fix it all:
1. working with something like ```UDPThread```(```extends Thread```) with closing socket on interrupt [how there](https://stackoverflow.com/a/4671080)
    create and interrupt it on start and stop broadcast
2.  not create new executor every time 
     instead ```while (true)``` use one of ```java.util.concurrent.*``` classes or ```Object#wait``` and ```Object#notify``` primitives
    setting timeouts for listening (avoid problem in '3')

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,131 @@
...
+        } catch (Exception e) {
+            logger.error("Broadcast Exception Encountered" + e.getMessage());
+        }
+        return receiveSocket;
+    }
+
+    private DatagramSocket getSendSocket() {
+        try {
+            if (sendSocket == null) {
+                sendSocket = new DatagramSocket(8001, InetAddress.getLocalHost());
```
DarkWeird: Seems, it is not working in LAN. Try test this PR on several computers.

DarkWeird:
> @DarkWeird
> 
> > You print response in log. How is regular player can use your feature? Regular player don't use console.
> 
> Yeah this was something I was waiting for feedback on. Do we want to build a separate area for displaying the servers or just show up on the JoinGameScreen.?

Better separate area with something like 'LAN' title. Or something like separate Tab (if it is possible).
When I saw your PR I thought about Counter-Strike 1.6 'LAN Games' :)

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,131 @@
...
+
+public class BroadCastServer {
+
+    private static DatagramSocket receiveSocket;
+    private static DatagramSocket sendSocket;
+    private static boolean turnOnBroadcast;
+    private static final Logger logger = LoggerFactory.getLogger(BroadCastServer.class);
+    private static ScheduledExecutorService service;
+    private static final String DISCOVERY_REQUEST = "DISCOVERY_REQUEST";
+    private static final String DISCOVERY_RESPONSE = "DISCOVERY_RESPONSE";
```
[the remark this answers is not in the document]
sladyn98: `turnOnBroadcast` needs to remain static right :/

engine/src/main/java/org/terasology/network/BroadCastServer.java
sladyn98: > Better separate area with something like 'LAN' title. Or something like separate Tab (if it is possible).
> When I saw your PR I thought about Counter-Strike 1.6's 'LAN Games' :)

Yeah will work on this :)

sladyn98: I tried this on wire shark and it does capture the packets but I have no access to more computers, could someone else try it

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -45,26 +44,29 @@ public void startBroadcast() {
             Executors.newSingleThreadExecutor().execute(new Runnable() {
                 @Override
                 public void run() {
-                    while (true) {
+                    while (!Thread.currentThread().isInterrupted()) {
```
sladyn98: @DarkWeird Added the thread interruption condition, so that we don't wait continously

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -45,26 +44,29 @@ public void startBroadcast() {
...
                         } catch (Exception e) {
                             logger.error("Broadcasting has been interrupted" + e.getMessage());
+                            stopBroadCast();
                         }
                     }
                 }
             });
         } catch (Exception e) {
         logger.info("Broadcasting has been interrupted " + e.getMessage());
+        stopBroadCast();
```
sladyn98: @DarkWeird Stopping broadcast on interruption would close the sockets thereby preventing unnecessary open sockets

engine/src/main/java/org/terasology/network/BroadCastServer.java
DarkWeird: If you using linux or msys, you can try to use 'nc' tool.
You can try to use docker or virtual machine, don't forget setup network.

DarkWeird: If you using linux or msys, you can try to use 'nc' tool.
You can try to use docker or virtual machine, don't forget setup network.

engine/src/main/java/org/terasology/network/BroadCastServer.java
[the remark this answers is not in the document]
DarkWeird: Why? With DI you will have only one instance.

engine/src/main/java/org/terasology/network/BroadCastServer.java
sladyn98: Yeah tried it on two computers it does show up :) Can anyone confirm

engine/src/main/java/org/terasology/network/BroadCastServer.java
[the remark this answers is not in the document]
sladyn98: Yeah

sladyn98:
Should this PR involve the development of a new UI component, since the IP addresses for the broadcast would anyway show up in the Join game section. What do you guys think ?
cc @skaldarnar @DarkWeird

engine/src/main/java/org/terasology/network/BroadCastServer.java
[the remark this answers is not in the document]
sladyn98: Resolved :+1:

Cervator:
IMHO I would try to just finish out this PR at its current level of functionality so we can merge it and start fresh with a second effort involving the UI. This has been an awesome PR but the longer one keeps going the more unwieldy the PR gets and the higher the risk of code going stale / conflicting with changes made in the meantime. Small relatively quick PRs ftw - although some topics to warrant some deeper review and some extra effort cycles :-)

sladyn98:
Cool @DarkWeird  This looks done to me then ?

sladyn98:
Gentle Ping @DarkWeird @Crevator Can we get this over the line :)

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,133 @@
...
+public class BroadCastServer {
+
+    private static boolean turnOnBroadcast;
+    private static final Logger logger = LoggerFactory.getLogger(BroadCastServer.class);
+    private  final String discoveryRequest = "DISCOVERY_REQUEST";
+    private  final String discoveryResponse = "DISCOVERY_RESPONSE";
+    private  DatagramSocket receiveSocket;
+    private  DatagramSocket sendSocket;
+
+
```
skaldarnar, line 35: I think I would order these members differently. First the _static final_ fields (the string constants can also be final I would think), second the static members, and finally the class members.

The string constants should then be written in all-caps.

```suggestion
    private static final Logger logger = LoggerFactory.getLogger(BroadCastServer.class);
    private static final String DISCOVERY_REQUEST= "DISCOVERY_REQUEST";
    private static final String DISCOVERY_RESPONSE = "DISCOVERY_RESPONSE";    
    
    private static boolean turnOnBroadcast;

    private DatagramSocket receiveSocket;
    private DatagramSocket sendSocket;
```

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,133 @@
...
+    public boolean isBroadCastTurnedOn() {
+        return turnOnBroadcast;
```
skaldarnar, line 37: maybe use _active_ as word here? `BroadCastServer.isActive()` may be a bit more intuitive.

engine/src/main/java/org/terasology/network/BroadCastServer.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public void startBroadcast() {
+        BroadCastServer.turnOnBroadcast = true;
+        try {
+            sendBroadCast();
+            Executors.newSingleThreadExecutor().execute(new Runnable() {
```
skaldarnar, line 44: Can use a lambda expression for the runnable here.

jdrueckert:
image
:thinking:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
```diff
@@ -92,6 +85,9 @@
     @In
     private StorageServiceWorker storageServiceWorker;
 
+    @In
+    private BroadCastServer broadCastServer;
```
skaldarnar, line 89: `BroadCastServer` is just a regular class without any annotations or such. Thus, I don't think it can just be injected like this.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
```diff
@@ -352,6 +348,22 @@ public Boolean get() {
                 refresh();
             });
         }
+
+        UIButton broadCastServerButton = find("broadcast", UIButton.class);
+        if (broadCastServerButton != null) {
+            broadCastServerButton.bindText(new ReadOnlyBinding<String>() {
+                @Override
+                public String get() {
+                    if (!broadCastServer.isBroadCastTurnedOn()) {
```
skaldarnar, line 357: In case `broadCastServer` cannot be injected this will be `null`, leading to a NullPointerException. That probably breaks the UI screen then.

skaldarnar:
Hey @sladyn98, are you still around? 

Please do continue work on this and ask questions if you're still interested in this. Otherwise, I'm going to close this PR due to inactivity.

SoniEx2:
I could fork this tbh

skaldarnar:
@SoniEx2 be our guest 🙃 broadcasting games in the local network would be a nice feature 🤓

sladyn98:
@SoniEx2 @skaldarnar I am so sorry guys, I do not think I can complete this feature, I am working with Jenkins for GSoC 2020 and I  feel really bad for leaving this code incomplete. TBH it took a lot more than I initially expected, :) but I enjoyed working on it. Thanks for the reviews

skaldarnar:
Thanks for the heads up @sladyn98 - closing this PR (for now).

PR #3833, opened 2020-02-16
Expose JOML as a package to modules

sin3point14:
### Contains

Exposes JOML as a package to Modules

### How to test

impot JOML, JomlUtil in any module and replace any vector/matrix with JOML counterpart and convert it to TeraMath conterpart with JomlUtil.from()

PR #3814, opened 2019-12-28
Bump for another Checkstyle config tweak in the TeraConfig project

Cervator:
Found another minor error, at least reported by the very latest IntelliJ's updated Checkstyle plugin. Unsure if it would encounter errors on older IntelliJs or via command line, would be good to get confirmation from somebody else before merging.

See https://github.com/MovingBlocks/TeraConfig/pull/8

PR #3806, opened 2019-12-15
Gradle upgrading

DarkWeird:
### Contains

Upgrade to Gradle 6.0
Upgrade FindBugs (Abandoned) -> SpotBugs
Simplification module dependencies src-bin determination (gradle way)
Replace artifactory to gradle's maven-publish plugin (support all maven repos)
Replacement compile to implementation (faster compile, smaller modules classpath)

### How to test
1. Basic:
   * Run ```gradlew build```
   * Must work (allTests passed, exists tests from MTE and Modules)
2. Client:
   * Run ```gradlew game```
   * Game must started and work normal (all needed modules loaded, assets available)
3. Server:
    * Run ```gradlew server```
    * Server must started and work normal (all needed modules loaded)
4. TeraEd:
    * Run ```gradlew editor```
    * Editor must started and work normal
5. Distribution:
    * Run ```gradlew distPC```
    * Grab distribution and launch it
    * Game must started and work normal (all needed modules loaded, assets available)
6. Publish:
    * Run ```gradlew publishToMavenLocal```
    * Check maven local repository for valid artifacts (.jar, sources.jar, javadoc.jar)
 7. Publish to Artifactory:
    * set environment variables ```ORG_GRADLE_PROJECT_mavenUser``` and ```ORG_GRADLE_PROJECT_mavenPass```
    * Run ``` gradlew publishAllPublicationsToMovingBlocksRepository``` (taskname has been generated, i called repository as ```MovingBlocks```)
    * Check artifactory repository for valid artifacts (.jar, sources.jar, javadoc.jar)
### Outstanding before merging (Move to Future PRs)

- [x] Try replace artifactory with generic maven plugin (Need testing with artifactory)
- [ ] ~~Generate more hardened build.gradle for modules (Terasology have modules that have invalid version reference in module.txt)~~
     * ~~fixed versions from module.txt - non "+" version~~
     * ~~regenerating on module.txt changes only.~~
- [ ] https://github.com/Terasology/ItemPipes/pull/12
- [ ] https://github.com/MovingBlocks/FacadeServer/pull/30
- [ ] https://github.com/Terasology/Health/pull/33

PR #3835, opened 2020-02-23
Gradle upgrade superset including Jenkins work and more!

Cervator:
I've built this primarily on #3806 with great thanks to @DarkWeird !

It finishes some of the more quirky parts of the overall logistics, does other minor stuff, enhances support for testing things outside our "live" environment in Artifactory, Jenkins, and on GitHub, updates to use Jenkins pipelines properly and the Jenkins Templating Engine (for modules). Also pulled in a couple other PRs and tested other stuff yet.

I believe this is merge-ready now and have new build jobs in the CBC Jenkins queued up to then start building everything fresh and neat in the new Jenkins, including a working Omega zip pulling in modules from new GitHub org mega-jobs automatically created by scanning for repos on GitHub - no more manually jobs in Jenkins and PRs will get tested for modules as well as the engine!

There are still improvements to make before we can use the new system for releases, primarily the approach remains hard coded to use the engine develop branch and snapshot versions for modules (much like before). Some tweaking needs to happen like consider `develop` a producer of snapshots and `master` a producer of released artifacts.. Not hard, just more work to do.

Hard to directly test some of this since Jenkins, but see #3806 for some of the central details and try to find any sort of Gradle task that appears broken. It is quite possible there could still be edge cases related to more rarely used stuff or rare combos involving user environments.

* Merges #3806 (which has links to more associated PRs)
* Merges #3833 (tested with https://github.com/Terasology/GooeyDefence/pull/36)
* Merges #3814 (which in turn includes https://github.com/MovingBlocks/TeraConfig/pull/8)
* Involved updates to https://github.com/MovingBlocks/ModuleJteConfig (module mega-jobs)
* Involves https://github.com/MovingBlocks/FacadeServer/pull/30 although that may need more testing
* Bumps version for engine and its embedded modules to `3.0.0` - this will require a bunch of module version bumps that'll be easiest to do in a big refactor done while merging this in a mega-workspace.
* Maintains the bundled version data used in the main menu and as part of the launcher (I think?) but can't yet test the launcher with new URLs as the CBC Jenkins still isn't exposed publicly (and has more complicated URLs in any case)
* Opens up properly attaching module build steps for maintaining the module browsing site (hi @majordwarf!) since those details can now be added to the JTE template Jenkinsfile.


GooeyHub:

build.gradle
```diff
@@ -185,16 +183,104 @@ clean.doLast {
...
+        resolutionStrategy.dependencySubstitution {
+            substitute module("org.terasology.engine:engine") because "we have sources!" with project(":engine")
+            substitute module("org.terasology.engine:engine-tests") because "we have sources!" with project(":engine-tests")
+        }
+    }
+}
+
+// Magic for replace remote dependency on local project (source)
+// For exists modules
+project(":modules").subprojects.forEach { proj ->
```
Cervator: This block appears to be what can cause a stall with `gradlew idea` when `org.gradle.parallel=true` is set in `gradle.properties` - FYI @DarkWeird 

If I comment this block out then the stall doesn't happen (but I'd assume module dependencies won't work). Probably relates to multiple threads working different modules for the `ideaModule` task yet this block iterates through all modules and for each goes and looks for every other module and substitutes stuff? That seems a classic case for `ConcurrentModificationException` being thrown when working with a `Map` just to come up with a possibly similar example :-)

We could still merge and advise to keep parallel processing off, with the intent to later get rid of `gradlew idea` entirely - *but* there could be a possibility of similar stalls in other Gradle logic that needs to iterate over modules?

Running Gradle in parallel mode is nice for efficiency but probably not really high priority yet. Could be more of an optimization thing to review later :-)

PR #3836, opened 2020-02-24
Update to BlockShape format: Replace BlockPart with String

Qwertygiy:
This PR updates the block shape system to rely on String-based maps as opposed to Enum-based maps. There are some minor performance benefits; particularly, values can be mapped to the "default" key instead of being copied to every side. But the primary benefit is that block shapes are no longer limited to the seven faces of top, bottom, left, right, front, back, and center; this allows for significantly more advanced block design and texture mapping.

All existing shapes should remain functional; this update should not change the behavior of any existing shape files or the appearance of any blocks.

A number of modules will require minor updates to gain compatibility with this PR.

### DIFF
--- a/engine-tests/src/test/java/org/terasology/world/block/shape/BlockShapeTest.java
+++ b/engine-tests/src/test/java/org/terasology/world/block/shape/BlockShapeTest.java
@@ -80,16 +80,23 @@ public class BlockShapeTest extends TerasologyTestingEnvironment {
 
         ObjectArrayList<javax.vecmath.Vector3f> points = ((com.bulletphysics.collision.shapes.ConvexHullShape) bulletConvexHullShape.underlyingShape).getPoints();
         for (int x = 0; x < points.size(); x++) {
-            fuzzVectorTest(test[x], VecMath.from(points.get(x)));
-
+            fuzzVectorTest(test[x], points);
         }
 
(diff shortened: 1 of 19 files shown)


GooeyHub:

GooeyHub:

GooeyHub:

skaldarnar:
With the recent discussion on case sensitivity and https://github.com/MovingBlocks/gestalt/pull/77 open I'm wondering whether a string-based approach is really advisable. At least we have to make sure the it is case insensitive, and I would vote for using `Name` to achieve that.

skaldarnar:
@Qwertygiy any news on this? Should we try to get this merged with a safer alternative than plain strings, or can this be closed?

4Denthusiast:
I'm in favour of closing this PR. I don't understand the reason why more sides would be needed. The separation of the mesh into parts seems to be mostly for the purpose of disabling parts of the mesh that are covered. Unless you're planning on making this system much more complicated than is necessary, this doesn't need any more than the BlockParts corresponding to sides, for things that will get occluded if the adjacent blocks have full faces, and the Center BlockPart for everything else.

skaldarnar:
Closed this because of the concerns about string-based maps. We can/should come back to this at some later point.

PR #3839, opened 2020-02-27
Controller settings

### LINKED ISSUE #3648: Add new "Controller Settings" page

EdvinArdo:
#### Contains
Fixes #3648. Builds on #3705, which moved controller settings to its own page.

It is now possible to rebind controller buttons. This is done via
sending a ControllerButtonEvent that is then handled by NuiManagerInternal
and UIInputBind.
    
The controller settings page now contains an axis sensitivity assigned to every axis.
The name labels were previously the raw names from the controller, but are now converted to a display name.

### How to test

Open Settings -> Input and rebind some buttons to controller buttons. Try using the changed buttons in-game.

Start the game with a controller plugged in. Open Settings -> Controllers and adjust the Axis Sensitivity sliders. Look and move around in-game.

Has been tested with an XBOX 360 controller on Linux.


EdvinArdo:
@Cervator could you take a look at this? Thanks!

Cervator:
Hi @EdvinArdo - nice work! Pings / notifications from GitHub can be very easy to miss due to the amount of repos and such we maintain, better to poke us on Discord or so :-)

It might still take just a bit of time to get to reviewing and testing this, but we very much appreciate the work!

skaldarnar:
Heyo @EdvinArdo, are you still around to address the requested changes? Finishing this up would be highly appreciated!

EdvinArdo:
Heyo! Sorry, I've lost interest and probably won't be finishing up :/

Cervator:
I put an extra note into #3648 to review this if somebody in the future gets back to that topic. Thank you again @EdvinArdo for the work here, even if we weren't able to get it merged :-)

PR #3841, opened 2020-03-03
feat: improve 'playerHeight' command

### LINKED ISSUE #2535: Height change extras

godfather2:
### Contains

Attempt to fix #2535 .
The character's jump speed,movement speed,interaction range and step height are a function of the player's height.I acquired the functions through play testing and graph plotters.
Earlier, a change in the player's height was made through a change in the his position through teleportation.This is probably what caused the player to sink through blocks at dwarf heights.This pull request, on the other hand just changes the eye height of the player.
- [x] Increase/Decrease jump speed according to size. Giants should be able to jump higher and dwarfs lower.
- [x] Giants should take longer strides, increase runFactor?
 - [x] Unable to pick up blocks off ground when height > 10. Bug!
 - [x] Physics bug: Eye height translates gazeMountPoint to a value more than expected.
 - [x] Trickier: Enable giants to be able to automatically climb up (no jump) blocks of smaller elevation.
- [x] Height too small, makes you fall through blocks. 
"Increase/Decrease damage from fall. (30m giant should not die of a 15m fall). Define function/proportionality constant same as above."->this sub issue is dealt by a pull request in the health module

### How to test

Just change the player's height and his capabilities should increase or decrease accordingly.

- [ ] Potions should use a AffectHeightComponent instead of affecting the PlayerConfig settings. These should override the Default (config) values and should be serialized to the world save.


GooeyHub:

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -18,12 +18,8 @@
 import org.terasology.entitySystem.entity.EntityManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.terasology.logic.characters.CharacterImpulseEvent;
+import org.terasology.logic.characters.*;
```
jellysnake: See: https://github.com/MovingBlocks/Terasology/wiki/Common-Style-Points#starred-imports

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +231,44 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
         return "";
     }
 
+    private float function1(float parameter){
```
jellysnake: See: https://github.com/MovingBlocks/Terasology/wiki/Common-Style-Points#descriptive-variable-names

skaldarnar: Indeed, please use a _descriptive name_ here. What is this function doing? What to the numbers mean?

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +231,44 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+        double p = (double)parameter;
+        return (float) pow(p,0.6) * 0.78f ;
```
skaldarnar: What's the reason to cast between float and doubles here? Wouldn't the following work?

```java
private double calculateSomething(double p) {
    return pow(p, 0.6) * 0.78;
}
```

I guess this could even be inlined.

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +231,44 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+    }
+
+    private float function2(float parameter){
```
skaldarnar: Same here, please use a better name for the method.

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +231,44 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+
     @Command(shortDescription = "Sets the height of the player", runOnServer = true,
             requiredPermission = PermissionManager.CHEAT_PERMISSION)
     public String playerHeight(@Sender EntityRef client, @CommandParam("height") float amount) {
         try {
             ClientComponent clientComp = client.getComponent(ClientComponent.class);
             CharacterMovementComponent move = clientComp.character.getComponent(CharacterMovementComponent.class);
-            if (move != null) {
+            CharacterComponent charComp=clientComp.character.getComponent(CharacterComponent.class);
+            EntityRef player = client.getComponent(ClientComponent.class).character;
```
skaldarnar: This can be replaced by a reference to `clientComp` which is retrieved in the first line of this try-catch block.

```suggestion
            EntityRef player = clientComp.character;
```

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +231,44 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+            GazeMountPointComponent gazeMountPointComponent = player.getComponent(GazeMountPointComponent.class);
+
+            float ratio = amount / 1.6f;
+            float constant= function1(amount);
```
skaldarnar: What constant is calculated based off the height (amount)? Is it the eye level or something like that?

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +231,44 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+
+            if (move != null && gazeMountPointComponent != null) {
                 float prevHeight = move.height;
+
                 move.height = amount;
+                gazeMountPointComponent.translate.y = amount;
+                move.jumpSpeed = 12 * constant;
+                move.stepHeight = 0.35f * ratio;
+                move.distanceBetweenFootsteps = ratio;
+                move.runFactor = function2(amount);
```
skaldarnar: If `function2` computes the run factor (whatever that is) you could call it `getRunFactor` instead.

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
godfather2: The values returned from the functions are assigned to float variables.Hence the cast

godfather2:
@skaldarnar "Potions should use a AffectHeightComponent instead of affecting the PlayerConfig settings. These should override the Default (config) values and should be serialized to the world save."
This is actually a follow up.I only mentioned it because it was mentioned as one of the extras in the issue

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +236,48 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+        return (float) pow(d,0.6) * 3.9f ;
+    }
+
+    private float getRunFactor(float amount){
+       return 0.15f + (0.64f * amount) - (0.006f * amount * amount);
+    }
+
     @Command(shortDescription = "Sets the height of the player", runOnServer = true,
             requiredPermission = PermissionManager.CHEAT_PERMISSION)
     public String playerHeight(@Sender EntityRef client, @CommandParam("height") float amount) {
```
skaldarnar, line 255: Height has a limitation of `@Range(min = 0, max = 5)` in `CharacterMovementComponent`, that should probably also be checked here...

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +236,48 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
         return "";
     }
 
+    private float getJumpSpeed(float amount){
+        double d = (double)amount;
+        return (float) pow(d,0.6) * 9.36f ;
```
skaldarnar, line 241: The default `jumpSpeed` seems to be 10 (and that's the maximum at the same time o.O ). thus, I would assume that setting the player height to the default should result in the default jump speed, but substituting `amount` by 1.6 yiels a value ~12.4 (≠10).

This is how the jump speed is adjusted according to this function, for `amount` in the interval from 0 to 5:
image
see [Wolfram Alpha](https://www.wolframalpha.com/input/?i=pow%28x%2C+0.6%29+*+9.36%2C+x+in+%5B0%2C5%5D)


see [CharacterMovementComponent.java#L53-L54](https://github.com/godfather2/Terasology/blob/heightExtras/engine/src/main/java/org/terasology/logic/characters/CharacterMovementComponent.java#L53-L54)

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +236,48 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+    }
+    
+    private float getInteractionRange(float amount){
+        double d = (double)amount;
+        return (float) pow(d,0.6) * 3.9f ;
+    }
+
+    private float getRunFactor(float amount){
+       return 0.15f + (0.64f * amount) - (0.006f * amount * amount);
```
skaldarnar, line 250: The run factor seems to be a value in the range from 0 to 10, with 1.5 being the default. 

Again, calculating the new run factor with the default height of 1.6 yields a value of 1.15864 (≠1.5).

This is how the run factor scales with the given function for `amount` between 0 and 5. The curve looks almost linear, to be hones:
image
see [Wolfram Alpha](https://www.wolframalpha.com/input/?i=0.15+%2B+%280.64+*+x%29+-+%280.006+*+x%5E2%29+%2C+x+in+%5B0%2C5%5D)

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +236,48 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
     }
 
+    private float getJumpSpeed(float amount){
+        double d = (double)amount;
+        return (float) pow(d,0.6) * 9.36f ;
+    }
+    
+    private float getInteractionRange(float amount){
+        double d = (double)amount;
+        return (float) pow(d,0.6) * 3.9f ;
```
skaldarnar, line 246: The default interaction range seems to be 5 (see [CharacterComponent.java#L37](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/logic/characters/CharacterComponent.java#L37)). With this formula the interaction range for height of 1.6 is 5.17055 (≠ 5).

This is what the plot looks like:
image
see [Wolfram Alpha](https://www.wolframalpha.com/input/?i=pow%28x%2C0.6%29+*+3.9%2C+x+in+%5B0%2C5%5D)

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +236,52 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
         return "";
     }
 
+    private float getJumpSpeed(float amount){
+        double d = (double)amount;
+        return Math.round(pow(d,0.55) * 8.45f);
+    }
```
skaldarnar, line 242:
```suggestion
    private float getJumpSpeed(float height) {
        return Math.round(Math.pow(height/DEFAULT_HEIGHT, 0.55f) * DEFAULT_JUMP_SPEED);
    }
```

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +236,52 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+    
+    private float getInteractionRange(float amount){
+        double d = (double)amount;
+        return Math.round(pow(d,0.6) * 3.9f);
+    }
```
skaldarnar, line 247:
```suggestion
    private float getInteractionRange(float height){
        return Math.round(pow(height/DEFAULT_HEIGHT, 0.6) * DEFAULT_INTERACTION_RANGE);
    }
```

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +236,52 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+
+    private float getRunFactor(float amount){
+       return Math.round(5.5f + (6.4f * amount) - (0.06f * amount * amount)) / 10f;
+    }
```
skaldarnar, line 251: Maybe use a similar formular to the ones you use above? Is there any reason this should be quadratic?
```suggestion
    private float getRunFactor(float height){
       return Math.round(Math.pow(height/DEFAULT_HEIGHT, 0.5) * DEFAULT_RUN_FACTOR);;
    }
```

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,22 +236,52 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
     public String playerHeight(@Sender EntityRef client, @CommandParam("height") float amount) {
         try {
             ClientComponent clientComp = client.getComponent(ClientComponent.class);
             CharacterMovementComponent move = clientComp.character.getComponent(CharacterMovementComponent.class);
-            if (move != null) {
+            CharacterComponent charComp = clientComp.character.getComponent(CharacterComponent.class);
+            EntityRef player = clientComp.character;
+            GazeMountPointComponent gazeMountPointComponent = player.getComponent(GazeMountPointComponent.class);
+
+            float defaultHeight = 1.6f;
```
skaldarnar, line 263: These should not be re-defined here. When the default values in CharacterMovementComponent are changed these will be outdated, and will likely cause some weird behavior and bugs.

Instead, `CharacterMovementComponent` should expose 

```java
    public  static final float DEFAULT_HEIGHT = 1.6f;
    public  static final float DEFAULT_JUMP_SPEED = 10.f;
```

godfather2: i just looked at the `restoreSpeed` command. This is how it accesses default values `CharacterMovementComponent moveDefault = prefab.get().getComponent(CharacterMovementComponent.class);` for instance ` move.jumpSpeed = moveDefault.jumpSpeed;` Should i do this?

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
godfather2: well....its just that i painstakingly constructing the other 2 functions manually,before i realized that it is much easier to arrive at them through regression,and that is how i acquired the third function.I must admit @skaldarnar you are really thorough

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,28 +238,59 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
         return "";
     }
 
+    private float getJumpSpeed(float ratio) {
+        return (float) Math.pow(ratio, 0.74f) * 6f + 4f;
+    }
```
skaldarnar, line 243: The default jump speed for the player is set in the prefab to 12, thus calculating with 10 here sets the default values off (sorry for not catching this earlier). I like the fact the we have a constant value here to ensure that the player can always jump some minimum height. 
When deriving this from the default jump speed I would also increase the influence of the constant and reduce the influence of the first part:
```suggestion
    private float getJumpSpeed(float ratio, float defaultValue) {
        return (float) Math.pow(ratio, 0.74f) * 0.4f * defaultValue + 0.6f * defaultValue;
    }
```

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,28 +238,59 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
-                return "Height of player set to " + amount + " (was " + prevHeight + ")";
+        if (amount >= 1 && amount <= 25) {
+            try {
+                ClientComponent clientComp = client.getComponent(ClientComponent.class);
+                CharacterMovementComponent move = clientComp.character.getComponent(CharacterMovementComponent.class);
+                CharacterComponent charComp = clientComp.character.getComponent(CharacterComponent.class);
+                EntityRef player = clientComp.character;
+                GazeMountPointComponent gazeMountPointComponent = player.getComponent(GazeMountPointComponent.class);
+                Optional<Prefab> prefab = Assets.get(new ResourceUrn("engine:player"), Prefab.class);
+                CharacterMovementComponent moveDefault = prefab.get().getComponent(CharacterMovementComponent.class);
```
skaldarnar, line 264: As we'll need both the default movement component and the default character component later I'll fetch both here:
```suggestion

                Prefab playerDefaults = Assets.getPrefab("engine:player").get();
                CharacterMovementComponent moveDefault = playerDefaults.getComponent(CharacterMovementComponent.class);
                CharacterComponent characterDefault = playerDefaults.getComponent(CharacterComponent.class)
```

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,28 +238,59 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+
+                if (move != null && gazeMountPointComponent != null) {
```
skaldarnar, line 266: Ideally we'll need a lot more null checks here (for all the components). I'll refactor that part after this PR is merged, though.

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,28 +238,59 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+                    float defaultEyeHeight = config.getPlayer().getEyeHeight();
+                    float foreHeadSize = moveDefault.height - defaultEyeHeight;
+                    float ratio = amount / moveDefault.height;
+
+                    move.height = amount;
+                    move.jumpSpeed = getJumpSpeed(ratio);
+                    move.stepHeight = moveDefault.stepHeight * ratio;
+                    move.distanceBetweenFootsteps = moveDefault.distanceBetweenFootsteps * ratio;
+                    move.runFactor = getRunFactor(ratio);
+                    charComp.interactionRange = getInteractionRange(ratio);
```
skaldarnar, line 277: Inline with the method changes above we'll need to pass the default values over now:
```suggestion
                    move.jumpSpeed = getJumpSpeed(ratio, moveDefault.jumpSpeed);
                    move.stepHeight = ratio * moveDefault.stepHeight;
                    move.distanceBetweenFootsteps = ratio * moveDefault.distanceBetweenFootsteps;
                    move.runFactor = getRunFactor(ratio, moveDefault.runFactor);
                    charComp.interactionRange = getInteractionRange(ratio, characterDefault.interactionRange);
```

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,28 +238,59 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+                    gazeMountPointComponent.translate.y = amount - foreHeadSize;
+                    Location.removeChild(player, gazeMountPointComponent.gazeEntity);
+                    Location.attachChild(player, gazeMountPointComponent.gazeEntity, gazeMountPointComponent.translate, new Quat4f(Quat4f.IDENTITY));
+
+                    player.saveComponent(gazeMountPointComponent);
```
skaldarnar, line 282: As this is the same code as for the `playerEyeHeight` command I would refactor this later on to use a common 
```java
updateGazeMountPointHeight(player, amount - foreHeadSize);
```

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,28 +238,59 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+                    clientComp.character.saveComponent(move);
+
+                    return "Height of player set to " + amount + " (was " + prevHeight + ")";
+                }
+                return "";
+            } catch (NullPointerException e) {
```
skaldarnar, line 288: If we did proper null checks this try-catch block should not be necessary (in general, catching a NPE is always a code smell as you could have done a null-check before).

engine/src/main/java/org/terasology/logic/debug/MovementDebugCommands.java
```diff
@@ -233,28 +238,61 @@ public String stepHeight(@Sender EntityRef client, @CommandParam("height") float
...
+            try {
+                ClientComponent clientComp = client.getComponent(ClientComponent.class);
+                CharacterMovementComponent move = clientComp.character.getComponent(CharacterMovementComponent.class);
+                CharacterComponent charComp = clientComp.character.getComponent(CharacterComponent.class);
+                EntityRef player = clientComp.character;
+                GazeMountPointComponent gazeMountPointComponent = player.getComponent(GazeMountPointComponent.class);
+
+                Prefab playerDefaults = Assets.getPrefab("engine:player").get();
+                CharacterMovementComponent moveDefault = playerDefaults.getComponent(CharacterMovementComponent.class);
+                CharacterComponent characterDefault = playerDefaults.getComponent(CharacterComponent.class)
```
skaldarnar, line 266:
```suggestion
                CharacterComponent characterDefault = playerDefaults.getComponent(CharacterComponent.class);
```

skaldarnar:
I think that this command should, under the hood, use a `ScaleCharacterEvent` such that other systems can also react on it (e.g., the Health system also needs to adjust values on height change). 

Merging this now as clear improvement and good progress to get the logic right.

PR #3802, opened 2019-12-13
Restore pre-BiomeAPI foliage shading

Qwertygiy:
During the extraction of biomes from the engine, block color sources and offsets were removed due to their integration with biomes. This resulted in rather dead-looking grass and leaves, as well as consequences with other color-shifted blocks not being color-shifted.

This PR re-implements the system as close to the original as possible without making reference to biomes, and instead referring to the world's humidity and temperature facets directly.

image

PR #3844, opened 2020-03-06
Revert "Restore pre-BiomeAPI foliage shading (#3802)"

Qwertygiy:
This reverts commit 8ad3d164

The method in which I had implemented this breaks the server/client boundary by asking for server-specific information about the world in the client. It works fine in single-player, but of course it causes crashes when attempting to connect to servers. This feature still needs to be reimplemented, but it will be a much larger job that requires more infrastructure for obtaining the required information from the server.


GooeyHub:

Issue #3725, opened 2019-08-16
External API Whitelist Review


eviltak:
The external API whitelist (defined in the `ExternalApiWhitelist` class) has a number of inclusions and exclusions that require more discussion. This issue aims to consolidate the few isolated issues and conversations that have been made on the topic.

## Pressing concerns (usually involving reflection)
- Removal of `MethodAccess` (#3060)
- Removal of the `jdk.internal.reflect` package
- Removal of the `sun.reflect` package

## Not-so-pressing concerns
- #2531: Java AWT whitelisting -- rather than expose the entire package, expose specific subpackages
- #2791: Removing `guava` from the whitelist -- no evidence of a security hole yet

PR #3847, opened 2020-03-08
Minor Gradle dependency tweak

Cervator:
I'm still trying to understand the `apiElements` and other fancy more modern Gradle stuff introduced with #3835, but I believe this tweak might be going in the right direction.

GSON not being in the `apiElements` list is why I believe the build of the Kallisti module (which admittedly won't function at runtime without security off) failed in Jenkins - it used to build OK then fail expectedly at runtime unless you ran with security disabled.

Likewise I had added JOML to the legacy `compile` configuration as it was added during the Gradle 6 upgrade finished in #3835, but now believe I understand it better, with it needing to be both `implementation` and `apiElements` - or maybe something else so we can declare it fully in just one statement. It isn't actually used at all in modulespace yet but it will be and has already been whitelisted.

My belief is that being in `apiElements` doesn't whitelist anything on its own, just declares an intent, then we secondarily whitelist using our normal methods.

I'm also wondering if GSON _should_ be on the whitelist? Pros/cons? Could be one less obstacle to making Kallisti work within the sandbox. Note that it builds in two different ways, this PR only relates to how it can build as a Terasology module (with our standard module build file), there's no impact to its [standalone build](http://jenkins.terasology.org/view/All/job/KallistiStandalone/) which should still just work

PR just for doc & discussion, merging right away to continue testing in Jenkins. But happy to have my understanding updated and for us to improve this segment of Gradle better yet!

Maybe of interest to @DarkWeird @skaldarnar @immortius @asiekierka @engiValk (security angle). Maybe slightly related to #3725

### DIFF
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -69,11 +69,13 @@ configurations.all {
 dependencies {
     // White list packages {@see ExternalApiWhiteList class}
     apiElements group: 'com.google.guava', name: 'guava', version: '23.0'
+    apiElements group: 'com.google.code.gson', name: 'gson', version: '2.6.2'
     apiElements group: 'net.sf.trove4j', name: 'trove4j', version: '3.0.3'
     apiElements group: 'org.lwjgl.lwjgl', name: 'lwjgl', version: LwjglVersion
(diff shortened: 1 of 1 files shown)


GooeyHub:

PR #3848, opened 2020-03-08
getWorld optimization

otradovec:
Two occurrences of for loop, which is executed max once >> changed to if

### DIFF
--- a/engine/src/main/java/org/terasology/world/sun/DefaultCelestialSystem.java
+++ b/engine/src/main/java/org/terasology/world/sun/DefaultCelestialSystem.java
@@ -27,6 +27,7 @@ import org.terasology.world.WorldProvider;
 import org.terasology.world.time.WorldTime;
 
 import java.math.RoundingMode;
+import java.util.Iterator;
 
 import static org.terasology.world.time.WorldTime.DAY_LENGTH;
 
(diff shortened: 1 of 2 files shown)


GooeyHub:

Cervator:
Hey again @otradovec - I'm not sure this code change is an improvement. It is a bit more awkward code for the same effect, even if the `for` loop itself is kinda goofy/wasteful when the _expectation_ is for there to be just one world entity. Probably we ultimately need to improve the backing code to where we either have the option to return a singleton entity or better handle the _potential_ for multiple (as we'll support multiple worlds sooner or later)

In other words it makes already awkward code that needs to be refactored fundamentally into more _explicitly_ awkward (if technically more accurate) code, if that makes sense. Chances are we do this in many places over the codebase, and really should just have a better way in general to handle singleton entities :-)

Curious if @immortius would have a comment for how we should architecturally approach singleton entities.

otradovec:
@Cervator
I have not though about having more worlds, then it makes more sence. Something like nether? I have never played this game, because I wasn't able to run it, but the code is interesting.

Cervator:
Yep exactly 👍 You did find some code that _does_ need improvement, just not quite sure how to handle it yet :-)

PR #3851, opened 2020-03-11
Added Refresh on changing from listed to custom in join game

meetcshah19:
Contains

The server info field "ping" remains unchanged when switching from Listed servers to Custom servers, ideally it should get refreshed according to the server whose information is being displayed at the moment.This PR fixes it.

How to test

1. Go to the join game screen 
2. Select a server in both listed and custom section  
3. Now changing the tab from listed to custom and vice-versa updates the ping field according to the server being displayed (it was not so earlier).


GooeyHub:

PR #3852, opened 2020-03-11
3850 Fix

### LINKED ISSUE #3850: Server connection test updates actively selected server not the origin server

Look at servers accurately on the Join Game screen showing multiple servers

### What actually happened

Clicked a bad server (offline) first by accident, then selected a working server. When the failed ping test came back from the bad server it showed the result (being unable to connect to the server) on the good server, despite the other stats showing online and all is well.

Some threaded check holding that call probably needs to pick its update target better. Maybe at present it just updates the field regardless of the selected server, when perhaps it should be a call-back with the specific server queried? And if that isn't visible /selected then don't show it. 

### How to reproduce

* Make sure there's both a working and not-working server in the join game list
* Select the broken server (should start a server query that'll time out)
* Quickly select a working server and wait a second or two, the connection to the good server will show that it "failed"

meetcshah19:
Contains

Fixes #3850 

How to test

Same steps used to Reproduce the issue

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
@@ -473,7 +473,7 @@ public class JoinGameScreen extends CoreScreenLayer {
     private void refreshPing() {
         String address = visibleList.getSelection().getAddress();
         int port = visibleList.getSelection().getPort();
-        String name= visibleList.getSelection().getName();
+        String name = visibleList.getSelection().getName();
         UILabel ping = find("ping", UILabel.class);
         ping.setText("Requested");
 
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
@@ -473,6 +473,7 @@ public class JoinGameScreen extends CoreScreenLayer {
     private void refreshPing() {
         String address = visibleList.getSelection().getAddress();
         int port = visibleList.getSelection().getPort();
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
```diff
@@ -473,6 +473,7 @@ private String getModulesText(Future<ServerInfoMessage> info) {
     private void refreshPing() {
         String address = visibleList.getSelection().getAddress();
         int port = visibleList.getSelection().getPort();
+        String name= visibleList.getSelection().getName();
```
jdrueckert:
```suggestion
        String name = visibleList.getSelection().getName();
```

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
```diff
@@ -486,6 +487,8 @@ private void refreshPing() {
                 }
             } catch (IOException e) {
                 String text = translationSystem.translate("${engine:menu#connection-failed}");
+                // Check if selection name is same as earlier when response is recieved before updating ping field
```
jdrueckert:
```suggestion
                // Check if selection name is same as earlier when response is received before updating ping field
```

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/JoinGameScreen.java
```diff
@@ -486,6 +487,8 @@ private void refreshPing() {
...
+                if(name.equals(visibleList.getSelection().getName()))
```
jdrueckert:
```suggestion
                if (name.equals(visibleList.getSelection().getName()))
```

PR #3854, opened 2020-03-13
Fix for game crashing when a player re-spawns in multiplayer

meetcshah19:
Contains

The game crashes when any player on the server tries to respawn when multiple players are present in the server.

How to test

1. Start a local server and two clients using run configurations .
2. Join the server from both the clients.
3. Die and respawn any player.(This causes the game to crash in the version before fix)

What have I done 

When a 'RespawnResquestEvent' is fired two listeners 'setSpawnLocationOnRespawnRequest' and 'onRespawnRequest' are triggered . Although the priority of 'setSpawnLocationOnRespawnRequest' was higher there is a chance that it finishes execution after 'onRespawnRequest' (they might run concurrently ) and in such an event the other player might try to render the respawned player before his spawn location is set which led to an error.

The code i have written runs RespawnResquestEvent after setSpawnLocationOnRespawnRequest thereby avoiding the error.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
@@ -210,7 +210,7 @@ public class PlayerSystem extends BaseComponentSystem implements UpdateSubscribe
         removeRelevanceEntity(entity);
     }
 
-    @ReceiveEvent(priority = EventPriority.PRIORITY_CRITICAL, components = {ClientComponent.class})
+//    @ReceiveEvent(priority = EventPriority.PRIORITY_CRITICAL, components = {ClientComponent.class})
     public void setSpawnLocationOnRespawnRequest(RespawnRequestEvent event, EntityRef entity) {
         ClientComponent clientComponent = entity.getComponent(ClientComponent.class);
         EntityRef character = clientComponent.character;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Good research, and I think this fix would indeed work! However, it can be improved :-)

For a quick bit of background: our event system is what helps facilitate communication between entity systems and extension with new usages of existing events/systems. Event handlers are set at different priority systems to leave space for future expansion - for instance a module may intercept the `RespawnRequestEvent` with a priority _between_ the current two levels ("critical" and "trivial") then selectively change just the location the player will respawn at. 

The AdventureAssets seemingly does this with an event handler at the "high" level. With this change in place that handler would now run first, before the logic in the old critical handler here has executed. That may not matter (I haven't analyzed the code that closely) but that's why a change with this approach may lead to bugs.

Likewise another approach can capture an event and consume it before it gets to lower priority handlers.

Two potential quick options (may need more thought):

1. If the two event handlers being independent doesn't actually matter (nobody ever needed the option, or the game changed to where it is no longer relevant) then probably just fold the two methods into one (especially to avoid confusion as the handler losing the annotation still _looks_ like a handler). In this case you'd want to check how that'd affect the handler in AdventureAssets.
1. Instead of a method call from the second handler to the first instead _send_ an event to offer a chance for other handlers to take action, including the old first handler. This is a pattern used elsewhere, although it tends to use _different_ events since sending the same event as the one you're in the middle of handling could lead to weirdness. Like the first option you'd want to examine current users to see how they'd be affected.

That's my real quick bit of feedback, but it is mostly from the top of my head so I can't guarantee its accuracy, but hopefully it'll shed some insight on the situation - together with checking out AdventureAssets. I can also recommend looking at Inferno as it also redirects the player to a different spawn point when they would have otherwise died, although I think that one may intercept death instead of a respawn :-)

meetcshah19:
New PR is up

Issue #3415, opened 2018-07-01
Better isolate involvement of OpenVR regarding FirstPersonHeldItemMountPointComponent


Cervator:
This in a way is a follow-up to #2611 which has far more todo items. But an issue that has come up lately is that if the game is run in an unusual context (environment-based unit tests in Jenkins, the server facade, etc) initialization of natives can be problematic and in some cases nonsensical such as supporting VR while running a headless testing client.

Specifically `FirstPersonHeldItemMountPointComponent` has references to `VRControllerState_t` which comes from inside the `engine/libs/jopenvr.jar` (moving that to a jar hosted by Artifactory is another todo) - however it is not necessarily present on the classpath in some cases. It works fine when running the game normally via PC facade.

Goal for this issue is to better isolate that somehow so the game can initialize everything in a headless context.

Our normal approach is to pick headless style subsystems instead of headed ones, see for instance `Terasology.populateSubsystems` - in this case `OpenVRInput` is only initialized if not headless. However, the `FirstPersonHeldItemMountPointComponent` usage is escaping that as it is linked to some of the general controller setup as well.

It might be possible to reverse how some of the callback stuff works so it is the OpenVR side that initiates its listener stuff without having to pass in OpenVR specific classes. Then unless that stuff is actively loaded and enabled (this happens regardless of whether VR is enabled in-game, which is another issue) the call never happens and `VRControllerState_t` is not needed on the classpath

Maybe of interest to @dave2s or perhaps @indianajohn if still around :-)

When fixed this should ease @Inei1's troubles when trying to bundle the server facade with the Omega zip while also allowing us to have fully functional environment-based unit tests for the first time in ages.

Resulting error (from starting server facade or running unit tests like the ones in the ModuleTestingEnvironment in Jenkins):

```
ERROR o.t.e.metadata.ComponentLibrary - Class not found, FirstPersonHeldItemMountPointComponent
java.lang.NoClassDefFoundError: jopenvr/VRControllerState_t
        at java.lang.Class.getDeclaredMethods0(Native Method)
        at java.lang.Class.privateGetDeclaredMethods(Unknown Source)
        at java.lang.Class.privateGetMethodRecursive(Unknown Source)
        at java.lang.Class.getMethod0(Unknown Source)
        at java.lang.Class.getMethod(Unknown Source)
        at org.terasology.utilities.ReflectionUtil.findMethod(ReflectionUtil.java:109)
[...]

PR #3856, opened 2020-03-16
More Gradle follow-up, more focused on code analytics

Cervator:
Although it may also hold some promise in fixing some quirks due to how with Gradle 6 we got a new default target build path (`build/classes/java/main` instead of plain `build/classes`) - turned out there were more places where that mattered than the few I found and tried to apply fixes for. 

This PR specifically fixes the ModuleTestingEnvironment, which for some reason failed to find _any_ world generators as the test for which module provided one would run into one path with the `java/main` fragment and one without. 

I tried to puzzle that out for a while but didn't really get anywhere. This solution may not be perfect (somewhat going backwards) but it reduces uncertainties while we improve stuff further and find the time to make more proper fixes. I expect a next big round might be when we switch 100% to Gradle vs using any IDE-specific stuff.

This restores how Gradle and IntelliJ builds to the same paths (both regular source and test source, to a different spot each) so that in theory stuff should work the same via Gradle as in IntelliJ (... and good luck with Eclipse or Netbeans). Ultimately we should delegate all building, run configs, etc to Gradle, at which point we might be able to fix some of these quirks. Others may take some classpath tweaking in Gestalt - maybe delay that till the reintegration of the latest release?

Goes with:

* https://github.com/MovingBlocks/ModuleJteConfig/pull/2 - updated module build file (at least demonstrates some stuff, needs more work)
* https://github.com/Terasology/Health/pull/37 - found another couple spots where tests depended on packages no longer normally available in the engine, fairly easy fixes
* I am *not* convinced this affects the mystery world generator resolution issue we're having these days, although I can't tell since it isn't happening for me right now. Pinging @sin3point14 to maybe try this out and see if there's _any_ difference?

Outstanding:

* For some reason the engine tests in `TypeRegistryTest` intermittently stall for me, both in Gradle and in IntelliJ. Sometimes I've been able to run the first one to completion even if it stalls for a minute or two, other times they just hang. Anybody else able to try it out on other OSes might be informative (I'm on Win10)
* Two of the MTE tests still fail, but seemingly from getting stuck in an infinite loop in `LocalChunkProvider.checkForUnload()` - huh? Not sure that'd relate to these changes. `ClientConnectionTest` and `ExampleTest` - the others work fine now.
* Module tests in Jenkins still fail, as is tradition, thanks to #3415 - confirmed via Health, although its MTE tests do work locally now, honest!
* Got a local fix for ItemPipes as well
* Need to enable the analytics visualization in the engine, just started in a module since way faster build time.

Pinging @DarkWeird for possible interest

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,4 @@ blacklist.json
 # Profiling related
 terasology.jfr
 local.properties
+*.hprof
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,12 +5,25 @@ node ("default-java") {
         sh 'chmod +x gradlew'
     }
     stage('Build') {
-        sh './gradlew clean build distForLauncher'
+        // Oddly it seems the first execution of gradlew correctly runs in plain console mode, for later steps we have to force it?
+        sh './gradlew clean distForLauncher'
         archiveArtifacts 'gradlew, gradle/wrapper/*, modules/Core/build.gradle, config/**, facades/PC/build/distributions/Terasology.zip, build/resources/main/org/terasology/version/versionInfo.properties, natives/**'
     }
     stage('Publish') {
         withCredentials([usernamePassword(credentialsId: 'artifactory-gooey', usernameVariable: 'artifactoryUser', passwordVariable: 'artifactoryPass')]) {
-            sh './gradlew publish -PmavenUser=${artifactoryUser} -PmavenPass=${artifactoryPass}'
+            sh './gradlew --console=plain publish -PmavenUser=${artifactoryUser} -PmavenPass=${artifactoryPass}'
         }
+    }
+    stage('Analytics') {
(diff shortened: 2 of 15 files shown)


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Tested by several others and tweaked some more. Passed in Jenkins via our test org jobs.

#3859 submitted to not forget some of the potentially problematic automated tests.

More analytics visualizations enabled and showing in the new Jenkins now. Tried to hook up the Sonar cloud scanner again as well (got an earlier experimental branch), but that may need a bit more work and API additions to create module projects and such automatically, plus it ends up doubling the build duration as I guess the Sonar site builds the project from scratch (at least LGTM runs in parallel)

PR #3858, opened 2020-03-16
Modified search to show if an asset is broken

### LINKED ISSUE #3741: Harden 'search' command against malformed assets

Search for stuff in the in-game console, like `search stone`

### What actually happened

Got a one-line error in the game console. Turned out an asset was misconfigured (was pointing to a template asset in the wrong module). After fixing that another similar case hit where the JSON in another asset was malformed, yet the error didn't give any clue _which _ asset.

Turns out I guess that to be able to search the engine goes and looks at every possible asset which means if it encounters even a single broken asset it'll error out and the search fails.

#1402 is a bigger goal to better validate assets (warnings at compile time somehow? Code scanning step?). But maybe we can harden `search` to better log what is breaking it?

#2874 also seems mildly similar

### How to reproduce

* Break an asset (like remove a comma or mess up a module name in a URI)
* Run the game and make sure the asset is within the active module set
* Run a `search` command for anything - get error!

### Example log 

`An error occurred while executing command 'search': Unable to resolve based block definition 'CoreBlocks:rock'`

(in that case the `rock` template actually lived in the CoreAssets module instead)

vedant-shroff:
### Contains

Fixed #3741 to harden the search for blocks. The console now shows which asset is broken when a search command is issued in the in game console . 

### How to test

Break an asset (eg : ladders) and write the command 'search stone' or search for any asset in the console. The search will return 0 results and return a message indicating that ladders is broken . Then fix it and run the same command and it should work.

If it works perfectly , can also be applied to prefabs and commands in the search

### DIFF
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -267,11 +267,9 @@ public class CoreCommands extends BaseComponentSystem {
             }
             // If a prefab is broken , it will throw an exception
             catch(Exception e){
-                List<String> EmptyList=new ArrayList<String>();
                 console.addMessage("Note : Search may not return results if invalid assets are present");
                 console.addMessage("Error parsing : "+curUrn.toString());
                 console.addMessage(e.toString());
-                // return(EmptyList);
                 continue;
             }
         }
--- a/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
@@ -49,6 +49,7 @@ import org.terasology.logic.console.suggesters.CommandNameSuggester;
 import org.terasology.logic.console.suggesters.ScreenSuggester;
 import org.terasology.logic.console.suggesters.SkinSuggester;
 import org.terasology.logic.inventory.events.DropItemEvent;
+import org.terasology.logic.location.Location;
 import org.terasology.logic.location.LocationComponent;
 import org.terasology.logic.permission.PermissionManager;
 import org.terasology.math.Direction;
@@ -83,15 +84,10 @@ import org.terasology.world.block.loader.BlockFamilyDefinition;
 
 import java.io.IOException;
 import java.net.UnknownHostException;
+import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
-import java.util.List;
-import java.util.Set;
-import java.util.LinkedList;
-import java.util.Locale;
-import java.util.Optional;
-import java.util.Arrays;
 
 /**
  * Adds a series of useful commands to the game. Likely these could be moved to more fitting places over time.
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -83,15 +84,10 @@
 
 import java.io.IOException;
 import java.net.UnknownHostException;
+import java.util.*;
```
jellysnake: One of our style rules is to avoid star imports. IntelliJ defaults to using them however.

You'll need to go into settings and change the number of imports required before a star is used up to 99. It's pain that we can't automatically do that at the moment but oh well

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -253,10 +249,29 @@ public String search(@CommandParam("searched") String searched) {
      * @return List of blocks that match searched string
      */
     private List<String> findBlockMatches(String searchLowercase) {
-        return assetManager.getAvailableAssets(BlockFamilyDefinition.class)
-                .stream().<Optional<BlockFamilyDefinition>>map(urn -> assetManager.getAsset(urn, BlockFamilyDefinition.class))
-                .filter(def -> def.isPresent() && def.get().isLoadable() && matchesSearch(searchLowercase, def.get()))
-                .map(r -> new BlockUri(r.get().getUrn()).toString()).collect(Collectors.toList());
+        ResourceUrn curUrn;
+        List<String> l=new ArrayList<String>();
```
jellysnake: We try and have all variables have useful names, even if they're not used much. Here I'd suggest something like `matchingBlocks` or even just `outputList`

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -253,10 +249,29 @@ public String search(@CommandParam("searched") String searched) {
...
+        for (ResourceUrn urn : assetManager.getAvailableAssets(BlockFamilyDefinition.class)) {
+            // Current urn for logging purposes to find the broken urn
+            curUrn = urn;
+            try{
+                Optional<BlockFamilyDefinition> def = assetManager.getAsset(urn, BlockFamilyDefinition.class);
+                if (def.isPresent() && def.get().isLoadable() && matchesSearch(searchLowercase, def.get())) {
+                    String s = new BlockUri(def.get().getUrn()).toString();
```
jellysnake: See my above comment about naming. (Although this variable could easily be joined with the following line

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -253,10 +249,29 @@ public String search(@CommandParam("searched") String searched) {
...
+                    l.add(s);
+                }
+            }
+            // If a prefab is broken , it will throw an exception
+            catch(Exception e){
+                List<String> EmptyList=new ArrayList<String>();
+                console.addMessage("Note : Search may not return results if invalid assets are present");
+                console.addMessage("Error parsing : "+curUrn.toString());
+                console.addMessage(e.toString());
+                return(EmptyList);
```
jellysnake: Would it be possible to instead of returning here, to just keep looping through the `for` block? 
That way the broken assets will be skipped but the rest of the command will still work and the printing will still work.

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
```diff
@@ -523,17 +538,16 @@ public String spawnPrefab(@Sender EntityRef sender, @CommandParam("prefabId") St
         } else {
             dir.set(Direction.FORWARD.getVector3f());
         }
-        Quat4f rotation = Quat4f.shortestArcQuat(Direction.FORWARD.getVector3f(), dir);
```
jellysnake: The cleanup of the method is appreciated, but this line seems like it has just been removed. Is it actually not needed, because it looks like it's used in spawning the prefab.

vedant-shroff: Fixed , Turns out I was using a slightly older branch which had an outdated version of this method , have replaced it with the correct implementation along with other changes

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
vedant-shroff: Good idea, Sorry for not implementing this initially , it was on my mind but I forgot about it once I was done with the functionality . This is the new output after this change.
Screenshot from 2020-03-20 19-26-20

engine/src/main/java/org/terasology/logic/console/commands/CoreCommands.java
vedant-shroff: Fixed! Will keep this in mind from now on .

pollend:
good job I was able to validate this:

image

PR #3855, opened 2020-03-15
Multipayer Respawn Crash Bug Fix

meetcshah19:
Contains

The game crashes when any player on the server tries to respawn when multiple players are present in the server.

How to test

1. Start a local server and two clients using run configurations .
2. Join the server from both the clients.
3. Die and respawn any player.(This causes the game to crash in the version before fix)

Other Modules Affected

The fix will cause AdventureAssets module's spawn location to not work (need to handle the spawn override using the new listener , A PR for that https://github.com/Terasology/AdventureAssets/pull/25 

What have I done 

The bug was due to a player being rendered before his spawn location was set (because making the player available  for rendering and setting his spawn location was being done asynchronously).
I have fixed the issue by making it synchronous and also made default spawnLocation modification  possible by making a BeforeRespawnEvent to handle it.

PR #3860, opened 2020-03-19
Multipayer Respawn Crash Bug Fix

meetcshah19:
Contains

The game crashes when any player on the server tries to respawn when multiple players are present in the server.

How to test

- Start a local server and two clients using run configurations .
- Join the server from both the clients.
- Die and respawn any player.(This causes the game to crash in the version before fix)

### DIFF
--- a/engine-tests/src/test/java/org/terasology/documentation/ApiScraper.java
+++ b/engine-tests/src/test/java/org/terasology/documentation/ApiScraper.java
@@ -61,7 +61,7 @@ public final class ApiScraper {
                 location = apiClass.getResource('/' + apiClass.getName().replace('.', '/') + ".class");
             }
 
-            if (location == null) {
+            if (location == null ) {
                 System.out.println("Failed to get a class/package location, skipping " + apiClass);
                 continue;
             }
--- a/engine-tests/src/test/java/org/terasology/documentation/ApiScraper.java
+++ b/engine-tests/src/test/java/org/terasology/documentation/ApiScraper.java
@@ -61,7 +61,7 @@ public final class ApiScraper {
                 location = apiClass.getResource('/' + apiClass.getName().replace('.', '/') + ".class");
             }
 
-            if (location == null ) {
+            if (location == null) {
                 System.out.println("Failed to get a class/package location, skipping " + apiClass);
                 continue;
             }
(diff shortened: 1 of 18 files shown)


GooeyHub:

Cervator:
So I'm not really sure about this - not on its own, at least. This just covers up the bug that #3855 attempted to fix, right? And that fix seemed reasonable, just not sure why it didn't fully work.

Hardening against a bug like this is a somewhat controversial topic. If you add loads of code to test against conditions that _should_ never happen, and simply skip logic if found, you might end up hiding real bugs now or in the future (and the hardening itself could be buggy, such as a typo in one of dozens of similar checks). At the least `WARN` level logging should note that the unexpected value occurred.

The null check is really a logic check that could as well (and probably should be) a test for whether the entity *has* the component in the first place, rather than "cheating" by just seeing "Hey did we get the thing?". So that's kinda valid. Then also checking that it doesn't have an uninitialized value ... it never _should_ happen, and therefore the code may be excessive, and some schools of thought suggest that in that case you _should_ crash with a runtime error to bring attention to the bug. 

I've leaned toward hardening in specific cases where we know there's a raised risk of bad data (such as a typical case where modules may be expected to return data that could be bad), but not blanket hardening just in case. I'm not really sure what's ideal here, it can be a hard judgement call.

In any case this just hides a bug we know to exist. So we still need to come up with a better way of handling the respawn. What we still outstanding / wrong with the other fix?

meetcshah19:
@Cervator The intention behind this PR was to just make multiplayer playable until we get to the root cause of the NaN values,my bad to not mention that earlier. The earlier fix was based on the logic that listeners for the RespawnRequest event were being run on different threads and that the enitity was being made available for rendering before its location was set and which is not the case (I have tried my best to follow the event chain properly and seems fine to me). I don't know how it fixed the problem in one scenario according according to your testing. I did log the entity getting the NaN values in it's location and it seems to be a player. Although I wasn't able to determine where these values are set (which probably is the root cause) and could surely use some help from someone who is more familiar with the code. Also the issue in not only in multiplayer.The game seems to crash in all server + client scenarios. MeshRenderer and ChunkRelevanceRegion are the classes which get the exception. Also weirdly after the fix #3855 the game crashes sometimes and doesn't the other times (I have had times when it didn't crash with 1 server + 2 clients). Just mentioning the last info to help anyone else trying to debug.

Cervator:
Hey @meetcshah19 yeah I pretty much came to the same conclusion, and just merged this and submitted a reminder issue to revisit in the future. At least this way it shouldn't be so much of a headache in the meanwhile :-)

Thanks again for helping hunt this down!

PR #3857, opened 2020-03-16
WIP: Feature/joml conversion 1

pollend:
This is a continuation of #3811 and #3837 this PR just ports the rest of the code base over. I added a rounding method to JOMLUtility and we might want to look into adding a PR to JOML adding some kind of rounding constructor between Vector3f <--> Vector3i. I found that trying to port over things by peace meal is more problematic for the long term. There are some odd movement problems that still need to be addressed in the code but it somewhat works for the most part.

PR #3863, opened 2020-03-23
WIP: started working on converion joml for location

pollend:
This is a continuation from  #3857 with a minimum amount of change on movement code. There still seems to be some problems with camera movement that is hard kind of hard to pin out.


GooeyHub:

PR #3865, opened 2020-03-25
feat(JOML): Migrate MeshRenderer/SkeletonRenderer

pollend:
I went in a bit of a different direction and I'll probably work outwards in. so this is just a conversion of the SkeletonRenderer since it does not have a lot of dependencies and building up those utility methods will help with porting the rest of everything. 

This can be tested with WildAnimals and spawning a deer in. any block entities should use the MeshRenderer. Player hand and anything held is also handled by the MeshRenderer. If everything works correctly then the player hand should work default with held items, explosions, the walk cycle on deers in Wildanimals should walk correctly.

```
./groovyw module recurse WildAnimals
./spawnprefab deer
```
I noticed a rotation problem but this seems to also be a problem with the base develop branch.

### DIFF
--- a/engine/src/main/java/org/terasology/math/MatrixUtils.java
+++ b/engine/src/main/java/org/terasology/math/MatrixUtils.java
@@ -48,10 +48,18 @@ public final class MatrixUtils {
         return buffer;
     }
 
+    /**
+     *
+     * Copies the given matrix into a newly allocated FloatBuffer
+     * the other of the elements in row-major.
+     * transformation operations happen on the inverse in the library
+     *
+     * @param m the matrix to copy
+     * @return a new Float buffer containing the matrix in row-major form
+     */
     public static FloatBuffer matrixToFloatBuffer(Matrix4fc m) {
         return m.getTransposed(BufferUtils.createFloatBuffer(16));
     }
-
 
     /**
      * Copies the given matrix into a newly allocated FloatBuffer.
@@ -75,7 +83,7 @@ public final class MatrixUtils {
      * Copies the given matrix into an existing FloatBuffer.
      * The order of the elements is column major (as used by OpenGL).
      *
-     * @param m  the matrix to copy
+     * @param m the matrix to copy
      * @param fb the float buffer to copy the matrix into
      * @return The provided float buffer.
      */
@@ -134,7 +142,7 @@ public final class MatrixUtils {
      *
      * @param m  the matrix to copy
      * @param fb the float buffer to copy the matrix into
-     * @return The provided float buffer.
+     * @return the provided float buffer with matrix elements in column major order.
      */
     public static FloatBuffer matrixToFloatBuffer(Matrix4fc m, FloatBuffer fb) {
         return m.getTransposed(fb);
@@ -255,10 +263,6 @@ public final class MatrixUtils {
         Matrix4f result = new Matrix4f();
         result.mul(m, vm);
         return result;
-    }
-
-    public static org.joml.Matrix4f calcModelViewMatrix(Matrix4fc m, Matrix4fc vm) {
-        return new org.joml.Matrix4f(vm).mul(m);
     }
 
     public static Matrix3f calcNormalMatrix(Matrix4f mv) {
--- a/engine/src/main/java/org/terasology/math/MatrixUtils.java
+++ b/engine/src/main/java/org/terasology/math/MatrixUtils.java
@@ -18,6 +18,7 @@ package org.terasology.math;
 
 import org.joml.Matrix3fc;
 import org.joml.Matrix4fc;
+import org.joml.Vector3ic;
 import org.lwjgl.BufferUtils;
 import org.terasology.math.geom.Matrix3f;
 import org.terasology.math.geom.Matrix4f;
@@ -40,8 +41,10 @@ public final class MatrixUtils {
      *
      * @param m the matrix to copy
      * @return A new FloatBuffer containing the matrix in column-major form.
-     * @deprecated used JOML method that uses Matrix4fc
+     * @deprecated This is scheduled for removal in an upcoming version
+     *             method will be replaced with JOML implementation {@link #matrixToFloatBuffer(Matrix4fc)}.
      */
+    @Deprecated
     public static FloatBuffer matrixToFloatBuffer(Matrix4f m) {
         FloatBuffer buffer = BufferUtils.createFloatBuffer(16);
         matrixToFloatBuffer(m, buffer);
@@ -67,7 +70,8 @@ public final class MatrixUtils {
      *
      * @param m the matrix to copy
      * @return A new FloatBuffer containing the matrix in column-major form.
-     * @deprecated used JOML method that uses Matrix3fc
+     * @deprecated This is scheduled for removal in an upcoming version
+     *             method will be replaced with JOML implementation {@link #matrixToFloatBuffer(Matrix3fc)}.
      */
     public static FloatBuffer matrixToFloatBuffer(Matrix3f m) {
         FloatBuffer buffer = BufferUtils.createFloatBuffer(9);
@@ -75,6 +79,13 @@ public final class MatrixUtils {
         return buffer;
     }
 
+    /**
+     * Copies the given matrix into a newly allocated FloatBuffer.
+     * The order of the elements is column major (as used by OpenGL).
+     *
+     * @param m the matrix to copy
+     * @return A new FloatBuffer containing the matrix in column-major form.
+     */
     public static FloatBuffer matrixToFloatBuffer(Matrix3fc m) {
         return m.getTransposed(BufferUtils.createFloatBuffer(9));
     }
@@ -86,7 +97,10 @@ public final class MatrixUtils {
      * @param m the matrix to copy
      * @param fb the float buffer to copy the matrix into
      * @return The provided float buffer.
+     * @deprecated This is scheduled for removal in an upcoming version
+     *             method will be replaced with JOML implementation {@link #matrixToFloatBuffer(Matrix3fc, FloatBuffer)}.
      */
+    @Deprecated
     public static FloatBuffer matrixToFloatBuffer(Matrix3f m, FloatBuffer fb) {
         fb.put(m.m00);
         fb.put(m.m10);
@@ -102,6 +116,14 @@ public final class MatrixUtils {
         return fb;
     }
 
+    /**
+     * Copies the given matrix into an existing buffer.
+     * The order of the elements is column major (as used by OpenGL).
+     *
+     * @param m  the matrix to copy
+     * @param fb the float buffer to copy the matrix into
+     * @return The provided float buffer.
+     */
     public static FloatBuffer matrixToFloatBuffer(Matrix3fc m, FloatBuffer fb) {
         return m.getTransposed(fb);
     }
@@ -113,7 +135,10 @@ public final class MatrixUtils {
      * @param m  the matrix to copy
      * @param fb the float buffer to copy the matrix into
      * @return The provided float buffer.
+     * @deprecated This is scheduled for removal in an upcoming version
+     *             method will be replaced with JOML implementation {@link #matrixToFloatBuffer(Matrix4fc, FloatBuffer)}.
      */
+    @Deprecated
     public static FloatBuffer matrixToFloatBuffer(Matrix4f m, FloatBuffer fb) {
         fb.put(m.m00);
         fb.put(m.m10);
@@ -282,6 +307,11 @@ public final class MatrixUtils {
         return result;
     }
 
+    /**
+     * Calculated normal view matrix from a modelViewMatrix
+     * @param mv ModelViewMatrix
+     * @return 3x3 normal matrix
+     */
     public static org.joml.Matrix3f calcNormalMatrix(Matrix4fc mv) {
         return mv.get3x3(new org.joml.Matrix3f()).invert().transpose();
     }
--- a/engine/src/main/java/org/terasology/math/MatrixUtils.java
+++ b/engine/src/main/java/org/terasology/math/MatrixUtils.java
@@ -94,6 +94,10 @@ public final class MatrixUtils {
         return fb;
     }
 
+    public static FloatBuffer matrixToFloatBuffer(Matrix3fc m, FloatBuffer fb) {
+        return m.getTransposed(fb);
+    }
+
     /**
      * Copies the given matrix into an existing FloatBuffer.
      * The order of the elements is column major (as used by OpenGL).
@@ -122,6 +126,18 @@ public final class MatrixUtils {
 
         fb.flip();
         return fb;
+    }
+
+    /**
+     * Copies the given matrix into an existing FloatBuffer.
+     * The order of the elements is column major (as used by OpenGL).
+     *
+     * @param m  the matrix to copy
+     * @param fb the float buffer to copy the matrix into
+     * @return The provided float buffer.
+     */
+    public static FloatBuffer matrixToFloatBuffer(Matrix4fc m, FloatBuffer fb) {
+        return m.getTransposed(fb);
     }
 
     public static org.joml.Matrix4f createViewMatrix(float eyeX, float eyeY, float eyeZ, float centerX, float centerY, float centerZ, float upX, float upY, float upZ) {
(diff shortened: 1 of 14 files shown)


GooeyHub:

GooeyHub:

pollend:
here is the skeletal renderer working with this branch. https://youtu.be/JcaLoCOVeEk

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
```diff
@@ -20,11 +20,11 @@
 import com.google.common.collect.Maps;
 import gnu.trove.list.TIntList;
 import gnu.trove.list.array.TIntArrayList;
+import org.joml.Quaternionf;
+import org.joml.Vector2f;
+import org.joml.Vector3f;
```
PS-Soundwave, line 25: Some uses of Vector3f are not equivalent in this file.

engine/src/main/java/org/terasology/rendering/md5/MD5SkeletonLoader.java
```diff
@@ -47,16 +48,16 @@
...
             "\\s+" + MD5Patterns.INTEGER_PATTERN);
-    private Pattern triPattern = Pattern.compile("tri\\s+" + MD5Patterns.INTEGER_PATTERN + 
-            "\\s+" + MD5Patterns.INTEGER_PATTERN + "\\s+" + MD5Patterns.INTEGER_PATTERN + 
+    private Pattern triPattern = Pattern.compile("tri\\s+" + MD5Patterns.INTEGER_PATTERN +
+            "\\s+" + MD5Patterns.INTEGER_PATTERN + "\\s+" + MD5Patterns.INTEGER_PATTERN +
             "\\s+" + MD5Patterns.INTEGER_PATTERN);
-    private Pattern weightPattern = Pattern.compile("weight\\s+" + MD5Patterns.INTEGER_PATTERN + 
-            "\\s+" + MD5Patterns.INTEGER_PATTERN + "\\s+" + MD5Patterns.FLOAT_PATTERN + 
+    private Pattern weightPattern = Pattern.compile("weight\\s+" + MD5Patterns.INTEGER_PATTERN +
+            "\\s+" + MD5Patterns.INTEGER_PATTERN + "\\s+" + MD5Patterns.FLOAT_PATTERN +
```
PS-Soundwave, line 60: Remove these; inflates repo size.

pollend: these are used to parse the lines in the MD5 file? do you want to move this into the source code or make them static?

PS-Soundwave: I meant all the added spaces

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
pollend: what cases does this not work?

skaldarnar:
>  noticed a rotation problem but this seems to also be a problem with the base develop branch.

Are you talking about the rotation of items held in hand? I think this is ancient... about time to fix it? ^^

engine/src/main/java/org/terasology/math/MatrixUtils.java
```diff
@@ -94,6 +94,10 @@ public static FloatBuffer matrixToFloatBuffer(Matrix3f m, FloatBuffer fb) {
         return fb;
     }
 
+    public static FloatBuffer matrixToFloatBuffer(Matrix3fc m, FloatBuffer fb) {
```
skaldarnar, line 97: What exactly does this do (aka, please add Javadoc)? 

My guess is taht the matrix `m` will stay unchanged, while the buffer `fb` will be filled with the matrix values?

engine/src/main/java/org/terasology/math/MatrixUtils.java
```diff
@@ -124,6 +128,18 @@ public static FloatBuffer matrixToFloatBuffer(Matrix4f m, FloatBuffer fb) {
         return fb;
     }
 
+    /**
+     * Copies the given matrix into an existing FloatBuffer.
+     * The order of the elements is column major (as used by OpenGL).
+     *
+     * @param m  the matrix to copy
+     * @param fb the float buffer to copy the matrix into
+     * @return The provided float buffer.
```
skaldarnar, line 137:
```suggestion
     * @return the provided float buffer with matrix elements in column major order.
```

engine/src/main/java/org/terasology/math/MatrixUtils.java
```diff
@@ -124,6 +128,18 @@ public static FloatBuffer matrixToFloatBuffer(Matrix4f m, FloatBuffer fb) {
         return fb;
     }
 
+    /**
+     * Copies the given matrix into an existing FloatBuffer.
```
skaldarnar, line 132: 👍

engine/src/main/java/org/terasology/math/MatrixUtils.java
```diff
@@ -241,6 +257,10 @@ public static Matrix4f calcModelViewMatrix(Matrix4f m, Matrix4f vm) {
         return result;
     }
 
+    public static org.joml.Matrix4f calcModelViewMatrix(Matrix4fc m, Matrix4fc vm) {
```
skaldarnar, line 260: Again, doc would be highly appreciated! Which parameters is modified, which is not? I think these are common pitfalls, and conventions that are not very obvious if you're not working with these classes all the time.

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/Bone.java
```diff
@@ -62,27 +62,30 @@ public Vector3f getLocalPosition() {
         Vector3f pos = new Vector3f(objectSpacePos);
         if (parent != null) {
             pos.sub(parent.getObjectPosition());
-            Quat4f inverseParentRot = new Quat4f();
-            inverseParentRot.inverse(parent.getObjectRotation());
-            inverseParentRot.rotate(pos, pos);
+            Quaternionf inverseParentRot = new Quaternionf();
+            parent.getObjectRotation().invert(inverseParentRot);
+//            inverseParentRot.inverse(parent.getObjectRotation());
```
skaldarnar, line 67: What is this comment about? Can it be removed?
```suggestion
```

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/Bone.java
```diff
@@ -62,27 +62,30 @@ public Vector3f getLocalPosition() {
...
-        Quat4f rot = new Quat4f(rotation);
+    public Quaternionf getLocalRotation() {
+        Quaternionf rot = new Quaternionf(rotation);
         if (parent != null) {
-            Quat4f inverseParentRot = new Quat4f();
-            inverseParentRot.inverse(parent.getObjectRotation());
-            rot.mul(inverseParentRot, rot);
+            Quaternionf inverseParentRot = new Quaternionf();
+            parent.getObjectRotation().invert(inverseParentRot);
+//            inverseParentRot.inverse(parent.getObjectRotation());
```
skaldarnar, line 86: same here 
```suggestion
```

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/Bone.java
```diff
@@ -62,27 +62,30 @@ public Vector3f getLocalPosition() {
...
+            rot.premul(inverseParentRot);
+//            inverseParentRot.mul(rot,rot);
```
skaldarnar, line 88: .. and here
```suggestion
```

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
skaldarnar: I only see the change from `scale` to `mul`with scalar value, which looks correct to me.

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshDataBuilder.java
```diff
@@ -75,7 +76,7 @@ public SkeletalMeshDataBuilder addMesh(Bone bone, MeshBuilder builder) {
     public SkeletalMeshDataBuilder addBox(Bone bone, Vector3f offset, Vector3f size, float u, float v) {
         MeshBuilder meshBuilder = new MeshBuilder();
         meshBuilder.setTextureMapper(textureMapper);
-        meshBuilder.addBox(offset, size, u, v);
+        meshBuilder.addBox(JomlUtil.from(offset), JomlUtil.from(size), u, v);
```
skaldarnar, line 79: I'm not really happy about the utility method name... `JomlUtil.from(...)` converts a JOML vector to a TeraMath vector? The naming suggests the other way around to me 🤔 

(nothing major, though)

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
```diff
@@ -191,23 +190,22 @@ private void renderEntitiesByMaterial(SetMultimap<Material, EntityRef> meshByMat
...
                         continue;
                     }
                     if (meshComp.mesh.isDisposed()) {
                         logger.error("Attempted to render disposed mesh");
                         continue;
                     }
 
-                    location.getWorldRotation(worldRot);
-                    location.getWorldPosition(worldPos);
+                    worldRot.set(JomlUtil.from(location.getWorldRotation()));
```
skaldarnar, line 201: I really like this change... I'm always confused by "getter" methods that actually mutate the method argument o.O

engine/src/main/java/org/terasology/rendering/logic/SkeletonRenderer.java
```diff
@@ -18,6 +18,10 @@
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import jdk.nashorn.internal.scripts.JO;
```
skaldarnar, line 21: Accidently import? I suppose using some JDK internal scripts would be a bad idea...
```suggestion
```

engine/src/main/java/org/terasology/rendering/md5/MD5ParserCommon.java
```diff
@@ -29,15 +29,16 @@
...
     private MD5ParserCommon() {
     }
 
     static {
         CORRECTION_MATRIX = new Matrix3f(-1, 0, 0, 0, 0, 1, 0, 1, 0);
-        CORRECTION_QUATERNION = new Quat4f(0, 0, 0, 1);
-        CORRECTION_QUATERNION.set(CORRECTION_MATRIX);
+        CORRECTION_QUATERNION = new Quaternionf(0, 0, 0, 1);
+        CORRECTION_QUATERNION.setFromNormalized(CORRECTION_MATRIX);
+//        CORRECTION_QUATERNION.set(CORRECTION_MATRIX);
```
skaldarnar, line 41:
```suggestion
```

engine/src/main/java/org/terasology/rendering/md5/MD5SkeletonLoader.java
skaldarnar: whut? I don't get the problem here... 🙄

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
skaldarnar: Clarification after discussion on Discord:

> e.g. Vector2f.sub(Vector2f, Vector2f) is present in both APIs but does something different in each
> one is dest.sub(minu, subtra) and the other is minu.sub(subtra, dest)

Would be good to compile a list of these incompatibilities and make sure to check them in the ported files.

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshDataBuilder.java
pollend: from one to the other is what I was thinking but you can think the same with regards to, one to the other. you can argue it either way and both are easy to interpret and make sense of.

engine/src/main/java/org/terasology/math/MatrixUtils.java
pollend: I don't use the method anyways. it just wraps a multiplication operation where a simple comment would probably be easier to interpret then a method use-case.

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
PS-Soundwave: semantics of ll. 163-164 change

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshDataBuilder.java
PS-Soundwave: also seems somewhat insignificant since use should be avoided in new code and the class is slated to be removed

engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
kBlaszczyk: The "dest" parameter is common practice in joml though. It's purpose is not to have an unnecessary memory allocation.

engine/src/main/java/org/terasology/rendering/assets/skeletalmesh/SkeletalMeshData.java
```diff
@@ -172,13 +172,14 @@ private void calculateNormals() {
```
kBlaszczyk, line 172: There is a `sub()` error in line 163-164. (above)
It should be:
```
vertices.get(indices.get(i * 3 + 1)).sub(baseVert, v1);
vertices.get(indices.get(i * 3 + 2)).sub(baseVert, v2);
```

engine/src/main/java/org/terasology/rendering/md5/MD5AnimationLoader.java
```diff
@@ -224,7 +225,7 @@ private void readBounds(BufferedReader reader, MD5 md5) throws IOException {
             max.x = Math.max(a.x, b.x);
             max.y = Math.max(a.y, b.y);
             max.z = Math.max(a.z, b.z);
-            md5.bounds[i] = AABB.createMinMax(min, max);
+            md5.bounds[i] = AABB.createMinMax(JomlUtil.from(min), JomlUtil.from(max));
```
kBlaszczyk, line 228: The lines above can be simplified using joml methods:
```
Vector3f min = new Vector3f(a).min(b);
Vector3f max = new Vector3f(a).max(b);
```

pollend: its a list of AABB bounds. unless we want to start using the AABB  bound that JOML already implements. the way it is right now is probably fine.

kBlaszczyk: I'm talking about this stuff:
```
Vector3f min = new Vector3f();
min.x = Math.min(a.x, b.x);
min.y = Math.min(a.y, b.y);
min.z = Math.min(a.z, b.z);
Vector3f max = new Vector3f();
max.x = Math.max(a.x, b.x);
max.y = Math.max(a.y, b.y);
max.z = Math.max(a.z, b.z);
```

pollend: I've been debating if it is really necessary. I think all the md5 stuff is going to go away at some point.

engine/src/main/java/org/terasology/math/MatrixUtils.java
pollend: only one method is calling this and its just a multiplication so I don't really see a value in this method call. I think I'll probably remove it at a later point.

pollend:
@Cervator I would squash it and just use that short description.

PS-Soundwave:
Pending confirmation that my most recent changes don't break bone rendering, I think we're done.

If we could get your blessing, @skaldarnar ~

PR #3867, opened 2020-03-25
ported skypshere implementation to JOML

pollend:
this port just focuses on porting over skysphere over to JOML.


GooeyHub:

PR #3869, opened 2020-03-27
added serilization compatibility for joml along with a basic test for…

pollend:
I added joml and moved termath to a legacy namespace along with a basic test to show that joml and termath vectors will serialize both for a given component. There are also some stub object I setup for gson that will start to play a role when I start porting in stuff like blocks and block families. I think it would be wise to introduce some more TypeSerializerTest to show that object will serialize back and forth correctly. At some point I think will remove the legacy termath stuff from TypeSerializer.

### DIFF
--- a/engine-tests/src/main/java/org/terasology/testUtil/TeraAssert.java
+++ b/engine-tests/src/main/java/org/terasology/testUtil/TeraAssert.java
@@ -16,7 +16,11 @@
 package org.terasology.testUtil;
 
 import com.google.common.collect.Lists;
+import org.joml.Vector2fc;
+import org.joml.Vector3fc;
+import org.joml.Vector4fc;
 import org.terasology.math.geom.Quat4f;
+import org.terasology.math.geom.Vector2f;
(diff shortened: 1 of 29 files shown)


GooeyHub:

GooeyHub:

Cervator:
Adding reminder here in case we aim to merge this soon (hopefully): we should also add the upgrade to JOML 1.9.23 to potentially fix a weirdly abrupt (if still subtle) shadow/lighting issue

pollend:
does 1.9.23 resolve this shadow issue you mention?

Cervator:
Haven't tested yet, but that was Soundwave's suspicion yep.

Cervator:
Tested a fair bit locally again, along with some bug hunting that may or may have not involved JOML. Found that #3881 likely is JOML, #3882 is likely not, but this PR doesn't seem to affect anything, I can play multiplayer, save/load a world, run bunches of unit tests etc 👍 

Included the bump to JOML 1.9.23

PR #3871, opened 2020-03-30
More Gradle: now with makeshift release management! Ready for v3.0.0

Cervator:
This ate my entire weekend, although just the spare time between GSOC proposal review, baby maintenance, and so on. Up way too late so will be brief :tm: - had imagined I might merge this straight to develop then PR to master for the v3.0.0 release, but too late and now too big.

Realized I needed to be able to vary how artifacts get published, either as releases or snapshots to Artifactory. This used to be done by having two separate jobs in the legacy Jenkins, one tied to `master` that made releases, one tied to `develop` that made snapshots. The actual version numbers were modified *by hand* as part of doing a release (take off `-SNAPSHOT`, push to `master`, bump version and reattach `-SNAPSHOT` ...)

New Jenkins has a multi-branch pipeline job. Shiny. But it was publishing _everything_ as snapshots, even pull requests getting built, meaning at any point in time a _module_ build could run against some engine artifact that came out of an unmerged PR. Not great.

Okay not too bad. Quick conditional to vary publishing by branch just for now. Oh, and really then I might as well just leave the version numbers without `-SNAPSHOT` and attach that dynamically if we're in `develop` ! Awesome bonus. Wait, but then what about modules. They use the same shared `publish.gradle` to keep things standard. That means they'll publish releases when building from `master` branches 🤔 

Alright, too much work to make `develop` branches for everything, default them, have devs do both branches as a standard... besides we might be able to do better like with Git tags (ohai @skaldarnar) instead of branches, and hey now checks are in place in code to where we could probably just look at tags instead of branches. But this needs to work right now.

Awkward release management to the rescue! We'll _default_ modules to not have release management enabled, meaning even while building `master` branches unless `"isReleaseManaged" : true` is added to `module.txt` the new stuff won't trigger and we'll still just make snapshots. And the embedded modules with the engine will get that set, but nobody else yet. It isn't super pretty, but ... it works.

For any reviewers: The actual change in `build.gradle` for modules is minimal, I just had to move some things around. This has been tested with the Nanoware job line in Jenkins and repo line in Artifactory (which also took some extra effort to get right)

Bonus: The code will now gently (or angrily, if red text is angry) warn about modules that still use `-SNAPSHOT` in their own version in `module.txt`, so we can eventually clean that out (or just move the version out of code into tags). Thanks @sin3point14 for that!

### Outstanding

- [ ] Goes with https://github.com/MovingBlocks/ModuleJteConfig/pull/3
- [ ] Javadoc fails on modules that have none, leading to failed build (huh, did this not trigger earlier?)
- [ ] Analytics run _after_ publishing - that's a sinful allowance since there are no quality gates in recording the analytics yet so no way to fail the build anyway and this way it gets to the interesting parts sooner
- [ ] Think about next steps sooner or later, this isn't perfect but gets us closer ...
- [ ] Modules in theory can now run release builds, if enabled. However these will still use the "build harness" from the `develop` build of the engine. More work to do and scenarios to satisfy - but we'll get there. Shout out to @PS-Soundwave

### DIFF
--- a/config/gradle/publish.gradle
+++ b/config/gradle/publish.gradle
@@ -17,15 +17,29 @@ publishing {
                 maven {
                     name = 'TerasologyOrg'
 
-                    def repoViaEnv = System.getenv()["PUBLISH_REPO"]
+                    String repoViaEnv = System.getenv()["PUBLISH_REPO"]
+                    String gitBranch = System.getenv()["BRANCH_NAME"]
                     if (rootProject.hasProperty("publishRepo")) {
+                        // This first option is good for local testing, you can set an explicit target repo in gradle.properties
                         url = "http://artifactory.terasology.org/artifactory/$publishRepo"
                         println "Changing PUBLISH repoKey set via property to $publishRepo"
                     } else if (repoViaEnv != null && repoViaEnv != "") {
+                        // Support varying publish target based on branch name
+                        if (gitBranch != null && gitBranch != "" && gitBranch.equals("master")) {
+                            println "We're working on a 'master' branch so replacing 'snapshot' with 'release' in the Artifactory repo"
+                            repoViaEnv = repoViaEnv.replace("snapshot", "release")
+                        }
+                        // Second one is aimed at Jenkins, where env vars can be set based on jobs/folders.
                         url = "http://artifactory.terasology.org/artifactory/$repoViaEnv"
                         println "Changing PUBLISH repoKey set via env var to $repoViaEnv"
                     } else {
-                        url = 'http://artifactory.terasology.org/artifactory/terasology-snapshot-local'
+                        // Finally this handles defaults while still supporting varying the publish target based on branch name
+                        if (gitBranch != null && gitBranch != "" && gitBranch.equals("master")) {
+                            println "We're working on a 'master' branch so using'release' in the Artifactory repo"
+                            url = 'http://artifactory.terasology.org/artifactory/terasology-release-local'
+                        } else {
+                            url = 'http://artifactory.terasology.org/artifactory/terasology-snapshot-local'
+                        }
                         println "PUBLISH repoKey is terasology-snapshot-local (default value)"
                     }
 
--- a/config/gradle/publish.gradle
+++ b/config/gradle/publish.gradle
@@ -17,30 +17,45 @@ publishing {
                 maven {
                     name = 'TerasologyOrg'
 
-                    String repoViaEnv = System.getenv()["PUBLISH_REPO"]
-                    String gitBranch = System.getenv()["BRANCH_NAME"]
                     if (rootProject.hasProperty("publishRepo")) {
-                        // This first option is good for local testing, you can set an explicit target repo in gradle.properties
+                        // This first option is good for local testing, you can set a full explicit target repo in gradle.properties
                         url = "http://artifactory.terasology.org/artifactory/$publishRepo"
-                        println "Changing PUBLISH repoKey set via property to $publishRepo"
-                    } else if (repoViaEnv != null && repoViaEnv != "") {
-                        // Support varying publish target based on branch name
-                        if (gitBranch != null && gitBranch != "" && gitBranch.equals("master")) {
-                            println "We're working on a 'master' branch so replacing 'snapshot' with 'release' in the Artifactory repo"
-                            repoViaEnv = repoViaEnv.replace("snapshot", "release")
-                        }
-                        // Second one is aimed at Jenkins, where env vars can be set based on jobs/folders.
-                        url = "http://artifactory.terasology.org/artifactory/$repoViaEnv"
-                        println "Changing PUBLISH repoKey set via env var to $repoViaEnv"
+                        println "Changing PUBLISH repoKey set via Gradle property to $publishRepo"
                     } else {
-                        // Finally this handles defaults while still supporting varying the publish target based on branch name
-                        if (gitBranch != null && gitBranch != "" && gitBranch.equals("master")) {
-                            println "We're working on a 'master' branch so using'release' in the Artifactory repo"
-                            url = 'http://artifactory.terasology.org/artifactory/terasology-release-local'
-                        } else {
-                            url = 'http://artifactory.terasology.org/artifactory/terasology-snapshot-local'
+                        // In this case we are going to fletch the publish repo together from a few things
+                        // First if we have an override from the environment to use a different target publish org
+                        String deducedPublishRepo = System.getenv()["PUBLISH_ORG"]
+                        if (deducedPublishRepo == null || deducedPublishRepo == "") {
+                            // If not then default
+                            deducedPublishRepo = "terasology"
                         }
-                        println "PUBLISH repoKey is terasology-snapshot-local (default value)"
+                        String releaseRepoFragment = "-release-local"
+                        String snapshotRepoFragment = "-snapshot-local"
+
+                        // Secondly we're going to find out if we're doing a release or a snapshot - this gets a little more complicated
+                        String gitBranch = System.getenv()["BRANCH_NAME"]
+                        if (gitBranch != null && gitBranch != "" && gitBranch.equals("master")) {
+                            // Okay we're in an environment where a branch name is set to 'master' so we might be doing a release
+                            // This is the funny part. Modules aren't ready globally to accept master branch == release, so check a prop that defaults to bypass
+                            if (project.hasProperty("bypassModuleReleaseManagement")) {
+                                if (bypassModuleReleaseManagement == "true") {
+                                    println "Release management not enabled for " + project.name + ", using snapshot repo despite building 'master' branch"
+                                    deducedPublishRepo += snapshotRepoFragment
+                                } else {
+                                    println "Release management *is* enabled for " + project.name + ", using release repo since building 'master' branch"
+                                    deducedPublishRepo += releaseRepoFragment
+                                }
+                            } else {
+                                println "We're working on a 'master' branch with defaults so using the release publish repo"
+                                deducedPublishRepo += releaseRepoFragment
+                            }
+                        } else {
+                            // No master branch so we for sure are just doing snapshots
+                            println "We're not working with a branch name set that's 'master' so assuming we're publishing snapshots"
+                            deducedPublishRepo += "-snapshot-local"
+                        }
+                        println "The final deduced publish repo is $deducedPublishRepo"
+                        url = "http://artifactory.terasology.org/artifactory/$deducedPublishRepo"
                     }
 
                     if (rootProject.hasProperty("mavenUser") && rootProject.hasProperty("mavenPass")) {
--- a/config/gradle/publish.gradle
+++ b/config/gradle/publish.gradle
@@ -22,38 +22,21 @@ publishing {
                         url = "http://artifactory.terasology.org/artifactory/$publishRepo"
                         println "Changing PUBLISH repoKey set via Gradle property to $publishRepo"
                     } else {
-                        // In this case we are going to fletch the publish repo together from a few things
-                        // First if we have an override from the environment to use a different target publish org
+                        // Support override from the environment to use a different target publish org
                         String deducedPublishRepo = System.getenv()["PUBLISH_ORG"]
                         if (deducedPublishRepo == null || deducedPublishRepo == "") {
                             // If not then default
                             deducedPublishRepo = "terasology"
                         }
-                        String releaseRepoFragment = "-release-local"
-                        String snapshotRepoFragment = "-snapshot-local"
 
-                        // Secondly we're going to find out if we're doing a release or a snapshot - this gets a little more complicated
+                        // Check the active Git branch and some module logic to see whether we're doing a release or snapshot
                         String gitBranch = System.getenv()["BRANCH_NAME"]
-                        if (gitBranch != null && gitBranch != "" && gitBranch.equals("master")) {
-                            // Okay we're in an environment where a branch name is set to 'master' so we might be doing a release
-                            // This is the funny part. Modules aren't ready globally to accept master branch == release, so check a prop that defaults to bypass
-                            if (project.hasProperty("bypassModuleReleaseManagement")) {
-                                if (bypassModuleReleaseManagement == "true") {
-                                    println "Release management not enabled for " + project.name + ", using snapshot repo despite building 'master' branch"
-                                    deducedPublishRepo += snapshotRepoFragment
-                                } else {
-                                    println "Release management *is* enabled for " + project.name + ", using release repo since building 'master' branch"
-                                    deducedPublishRepo += releaseRepoFragment
-                                }
-                            } else {
-                                println "We're working on a 'master' branch with defaults so using the release publish repo"
-                                deducedPublishRepo += releaseRepoFragment
-                            }
+                        if (isMaster(gitBranch) && !shouldBypassModuleRelease()) {
+                            deducedPublishRepo += "-release-local"
                         } else {
-                            // No master branch so we for sure are just doing snapshots
-                            println "We're not working with a branch name set that's 'master' so assuming we're publishing snapshots"
                             deducedPublishRepo += "-snapshot-local"
                         }
+
                         println "The final deduced publish repo is $deducedPublishRepo"
                         url = "http://artifactory.terasology.org/artifactory/$deducedPublishRepo"
                     }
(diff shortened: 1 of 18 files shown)


GooeyHub:

config/gradle/publish.gradle
```diff
@@ -17,16 +17,45 @@ publishing {
                 maven {
                     name = 'TerasologyOrg'
 
-                    def repoViaEnv = System.getenv()["PUBLISH_REPO"]
                     if (rootProject.hasProperty("publishRepo")) {
```
skaldarnar: I'm wondering whether we should start to scope the properties in the gradle builds, similar to what most plugins to. This would become `terasology.publishRepo` then...

config/gradle/publish.gradle
```diff
@@ -17,16 +17,45 @@ publishing {
...
+                                }
+                            } else {
+                                println "We're working on a 'master' branch with defaults so using the release publish repo"
+                                deducedPublishRepo += releaseRepoFragment
+                            }
+                        } else {
+                            // No master branch so we for sure are just doing snapshots
+                            println "We're not working with a branch name set that's 'master' so assuming we're publishing snapshots"
+                            deducedPublishRepo += "-snapshot-local"
+                        }
```
skaldarnar: The whole logic in here boils down to just this, doesn't it?

```groovy
if (isMaster(maybeGitBranch) && !shouldBypassModuleRelease()) {
	deducedPublishRepo += "-release-local"
} else {
	deducedPublishRepo += "-snapshot-local"
}
```

I think with the two additional functions this code is really readable, and the nesting is reduced.

```groovy
def isMaster(gitBranch) {
  return gitBranch != null && gitBranch != "" && gitBranch.equals("master");
}

def shouldBypassModuleRelease() {
  return project.hasProperty("bypassModuleReleaseManagement") && 
	bypassModuleReleaseManagement == "true";
}
```

config/gradle/publish.gradle
```diff
@@ -17,16 +17,45 @@ publishing {
...
+                                    deducedPublishRepo += releaseRepoFragment
+                                }
+                            } else {
+                                println "We're working on a 'master' branch with defaults so using the release publish repo"
+                                deducedPublishRepo += releaseRepoFragment
+                            }
+                        } else {
+                            // No master branch so we for sure are just doing snapshots
+                            println "We're not working with a branch name set that's 'master' so assuming we're publishing snapshots"
+                            deducedPublishRepo += "-snapshot-local"
```
skaldarnar:
```suggestion
                            deducedPublishRepo += snapshotRepoFragment
```

config/gradle/publish.gradle
skaldarnar: This would loose a couple of console lines, but overall the structure in the build file is more comprehensible imho.

modules/BiomesAPI/build.gradle
```diff
@@ -35,17 +25,25 @@ buildscript {
...
     env = System.getenv()
+
+    // This is a fun one ... when versions switched to dynamic -SNAPSHOT or not based on branch existing modules using `master` would suddenly try publishing releases
+    // This won't work without additionally doing constant version bumps (perhaps via Git tags) - but too much work to switch around all modules at once
+    // Complicating things more the use of publish.gradle to centralize logic means modules and engine bits are treated the same, yet we need to vary modules
+    // Temporary workaround: default modules to bypass release management: master branch builds will still make snapshot builds for the snapshot repo
+    // If a module actually wants release management simply include `"isReleaseManaged" : true` in module.txt - this is needed for the engine repo embedded modules
+    // One option would be to slowly convert modulespace to default to a `develop` + `master` setup living in harmony with associated automation/github tweaks
+    // Alternatively one more round of refactoring could switch to Git tags, a single `master` branch, and possible other things to help match snaps/PR builds somehow?
```
skaldarnar: Heavyly in favor of git tags, as most modules don't have complicated branching going on anyways. But you're right, should not be done in this PR, and needs some proper thoughtwork to figure out details and edge cases...

modules/BiomesAPI/build.gradle
```diff
@@ -35,17 +25,25 @@ buildscript {
...
 
 ext {
     // Read environment variables, including variables passed by jenkins continuous integration server
     env = System.getenv()
+
+    // This is a fun one ... when versions switched to dynamic -SNAPSHOT or not based on branch existing modules using `master` would suddenly try publishing releases
+    // This won't work without additionally doing constant version bumps (perhaps via Git tags) - but too much work to switch around all modules at once
+    // Complicating things more the use of publish.gradle to centralize logic means modules and engine bits are treated the same, yet we need to vary modules
+    // Temporary workaround: default modules to bypass release management: master branch builds will still make snapshot builds for the snapshot repo
+    // If a module actually wants release management simply include `"isReleaseManaged" : true` in module.txt - this is needed for the engine repo embedded modules
```
skaldarnar: I don't like the fact that the generic `module.txt` now holds information specific to our build system. As it is not officially documented anywhere, I guess its not much of a deal. Nevertheless, I would like to get rid of it asap.

modules/BiomesAPI/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
 // Gradle uses the magic version variable when creating the jar name (unless explicitly set somewhere else I guess)
 version = moduleConfig.version
 
+import org.gradle.internal.logging.text.StyledTextOutputFactory
+import static org.gradle.internal.logging.text.StyledTextOutput.Style
+
+// The only case in which we make module non-snapshots is if release management is enabled and BRANCH_NAME is "master"
+if (moduleConfig.isReleaseManaged && env.BRANCH_NAME == "master") {
```
skaldarnar: Why no null check for the branch name here? Shouldn't this be a similar check to `isMaster(...)` as proposed above?

modules/BiomesAPI/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+    // This is mildly awkward since we need to bypass by default, yet if release management is on (true) then we set the bypass to false ..
+    ext.bypassModuleReleaseManagement = false
+    if (version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (version.endsWith("-SNAPSHOT")) {
```

modules/BiomesAPI/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+        version -= "-SNAPSHOT"
+        def out = services.get(StyledTextOutputFactory).create("an-ouput")
+        out.withStyle(Style.FailureHeader).println("WARNING: Module " + project.name + " is explicitly versioned as a snapshot in module.txt, please remove '-SNAPSHOT'")
+    }
+} else {
+    if (!version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (!version.endsWith("-SNAPSHOT")) {
```

modules/BuilderSampleGameplay/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
 // Gradle uses the magic version variable when creating the jar name (unless explicitly set somewhere else I guess)
 version = moduleConfig.version
 
+import org.gradle.internal.logging.text.StyledTextOutputFactory
+import static org.gradle.internal.logging.text.StyledTextOutput.Style
+
+// The only case in which we make module non-snapshots is if release management is enabled and BRANCH_NAME is "master"
+if (moduleConfig.isReleaseManaged && env.BRANCH_NAME == "master") {
```
skaldarnar: see above (`isMaster(...)` check).

modules/BuilderSampleGameplay/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+    // This is mildly awkward since we need to bypass by default, yet if release management is on (true) then we set the bypass to false ..
+    ext.bypassModuleReleaseManagement = false
+    if (version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (version.endsWith("-SNAPSHOT")) {
```

modules/BuilderSampleGameplay/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+        version -= "-SNAPSHOT"
+        def out = services.get(StyledTextOutputFactory).create("an-ouput")
+        out.withStyle(Style.FailureHeader).println("WARNING: Module " + project.name + " is explicitly versioned as a snapshot in module.txt, please remove '-SNAPSHOT'")
+    }
+} else {
+    if (!version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (!version.endsWith("-SNAPSHOT")) {
```

modules/Core/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
 // Gradle uses the magic version variable when creating the jar name (unless explicitly set somewhere else I guess)
 version = moduleConfig.version
 
+import org.gradle.internal.logging.text.StyledTextOutputFactory
+import static org.gradle.internal.logging.text.StyledTextOutput.Style
+
+// The only case in which we make module non-snapshots is if release management is enabled and BRANCH_NAME is "master"
+if (moduleConfig.isReleaseManaged && env.BRANCH_NAME == "master") {
```
skaldarnar: see above

modules/Core/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+    // This is mildly awkward since we need to bypass by default, yet if release management is on (true) then we set the bypass to false ..
+    ext.bypassModuleReleaseManagement = false
+    if (version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (version.endsWith("-SNAPSHOT")) {
```

modules/Core/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+        version -= "-SNAPSHOT"
+        def out = services.get(StyledTextOutputFactory).create("an-ouput")
+        out.withStyle(Style.FailureHeader).println("WARNING: Module " + project.name + " is explicitly versioned as a snapshot in module.txt, please remove '-SNAPSHOT'")
+    }
+} else {
+    if (!version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (!version.endsWith("-SNAPSHOT")) {
```

modules/CoreSampleGameplay/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
 // Gradle uses the magic version variable when creating the jar name (unless explicitly set somewhere else I guess)
 version = moduleConfig.version
 
+import org.gradle.internal.logging.text.StyledTextOutputFactory
+import static org.gradle.internal.logging.text.StyledTextOutput.Style
+
+// The only case in which we make module non-snapshots is if release management is enabled and BRANCH_NAME is "master"
+if (moduleConfig.isReleaseManaged && env.BRANCH_NAME == "master") {
```
skaldarnar: see above

modules/CoreSampleGameplay/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+    // This is mildly awkward since we need to bypass by default, yet if release management is on (true) then we set the bypass to false ..
+    ext.bypassModuleReleaseManagement = false
+    if (version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (version.endsWith("-SNAPSHOT")) {
```

modules/CoreSampleGameplay/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+        version -= "-SNAPSHOT"
+        def out = services.get(StyledTextOutputFactory).create("an-ouput")
+        out.withStyle(Style.FailureHeader).println("WARNING: Module " + project.name + " is explicitly versioned as a snapshot in module.txt, please remove '-SNAPSHOT'")
+    }
+} else {
+    if (!version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (!version.endsWith("-SNAPSHOT")) {
```

templates/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
 // Gradle uses the magic version variable when creating the jar name (unless explicitly set somewhere else I guess)
 version = moduleConfig.version
 
+import org.gradle.internal.logging.text.StyledTextOutputFactory
+import static org.gradle.internal.logging.text.StyledTextOutput.Style
+
+// The only case in which we make module non-snapshots is if release management is enabled and BRANCH_NAME is "master"
+if (moduleConfig.isReleaseManaged && env.BRANCH_NAME == "master") {
```
skaldarnar: see above

templates/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+    // This is mildly awkward since we need to bypass by default, yet if release management is on (true) then we set the bypass to false ..
+    ext.bypassModuleReleaseManagement = false
+    if (version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (version.endsWith("-SNAPSHOT")) {
```

templates/build.gradle
```diff
@@ -69,11 +67,44 @@ for (dependency in moduleConfig.dependencies) {
...
+        version -= "-SNAPSHOT"
+        def out = services.get(StyledTextOutputFactory).create("an-ouput")
+        out.withStyle(Style.FailureHeader).println("WARNING: Module " + project.name + " is explicitly versioned as a snapshot in module.txt, please remove '-SNAPSHOT'")
+    }
+} else {
+    if (!version.contains("-SNAPSHOT")) {
```
skaldarnar:
```suggestion
    if (!version.endsWith("-SNAPSHOT")) {
```

config/gradle/publish.gradle
Cervator: Had not thought about that before. Even in a plain build file, rather than a plugin? I've been curious if some of our logic might go better _in_ a plugin, where it would make all kinds of sense, since then you'd probably use neato little blocks to configure it.

modules/BiomesAPI/build.gradle
Cervator: Yep, it is a workaround for sure. Lets me get through this big release without an even bigger overhaul. It can be an undocced feature likely never to be used in module land as we can hopefully finish improving things soon.

modules/BiomesAPI/build.gradle
Cervator: Mixture of "way too deep into the rabbit warren, just want to finish!" and from this spot actually being null safe since `==` rather than `equals` :-)

Why is the _other_ place null checking and using `String.equals` ? Rabbit holes and 2 am coding ;-)

config/gradle/publish.gradle
Cervator: Thank you! I had even _more_ logic madness in there while trying to find a solution that would cover all the scenarios, this was actually an already boiled down version, but closing in on 2 am was not letting me see clearly in shrinking it further. This is way cleaner! Don't need the logging now that it tests out fine :-)

modules/BiomesAPI/build.gradle
Cervator: Applied this, thanks! Actually cleaned it up a bit more yet now that I can brain again.

GooeyHub:

Cervator:
Thanks a lot for the fresh eyes on this! Helped a lot, cleaner now, tests out still, hopefully looks better now. Gotta fix the Javadoc thing soon then will merge and do the release 👍

facades/subprojects.gradle
```diff
@@ -1,9 +1,13 @@
 // This magically allows subdirs in this subproject to themselves become sub-subprojects in a proper tree structure
 new File(rootDir, 'facades').eachDir { possibleSubprojectDir ->
-    def subprojectName = 'facades:' + possibleSubprojectDir.name
-    println "Processing facade $subprojectName, including it as a sub-project"
-    include subprojectName
-    def subprojectPath = ':' + subprojectName
-    def subproject = project(subprojectPath)
-    subproject.projectDir = possibleSubprojectDir
+    if (!possibleSubprojectDir.name.startsWith(".")) {
```
skaldarnar, line 3: What about a general approach like this (in the future): https://medium.com/@ungesehn/use-gitignore-for-gradle-task-excludes-e5d011e99f71

`.settings` should be gitignores regardless of where in the file tree...

GooeyHub:

Cervator:
Fixed the Javadoc issue, validated with the Nanoware jobs etc, and merged! Initial engine develop job ran fine :shipit:

PR #3872, opened 2020-03-30
Add World generation listeners

askneller:
### Contains

This PR adds functionality that allows systems in modules to "listen in" on aspects of world generation by requesting to be notified when a certain class of facet (e.g. ```SurfaceHeightFacet```) has been generated for a region. 

It does so by creating a ```WorldGeneratorListenerLibrary``` similar in implementation to ```WorldGeneratorPluginLibrary```. Once a facet has been provided to a ```Region``` (via ```RegionImpl.getFacet()```) a map of ```FacetProviderListener```s are notified of the facets to which they have registered to listen.

My motivation in this was that I wanted to find parts of terrain that had certain surface features (in my case mountain-like peaks) that I would use at a later point during the game to do things like generate ominous looking towers. I didn't do this via regular world generation providers and rasterizers because I wanted to check areas that spanned more than one region (currently a group of 4). I figured it might also be a convenient way of storing other locations for later usage, like spawning buried treasure or something in a cave once a certain event has occurred etc.

So this is what I came up with. Any suggestions for improvement would be greatly appreciated. Or if this is not a direction you want to take the code fair enough, thanks for considering it.

### How to test

You can add a class ```DummyListener``` to ```CoreSampleGameplay``` using the following:
```
package org.terasology.coresamplegameplay;

import org.slf4j.Logger;
import org.terasology.world.generation.Facet;
import org.terasology.world.generation.FacetProviderListener;
import org.terasology.world.generation.GeneratingRegion;
import org.terasology.world.generation.WorldFacet;
import org.terasology.world.generation.facets.SurfaceHeightFacet;
import org.terasology.world.generator.plugin.RegisterFacetListener;

@RegisterFacetListener({@Facet(SurfaceHeightFacet.class)})
public class DummyListener implements FacetProviderListener {

    private static final Logger logger = LoggerFactory.getLogger(DummyListener.class);

    private static final Object lock = new Object();
    private static final int max = 25; // Just to illustrate notify
    private static Integer handled = 0; // Just to illustrate notify

    @Override
    public <F extends WorldFacet> void notify(GeneratingRegion region, F facet) {
        synchronized (lock) {
            if (handled < max) {
                logger.info("Have been notified of facet {} on region {}",
                        facet.getClass().getSimpleName(), region.getRegion());
                handled++;

                // From here the class can do some quick checks about the facet to see if the region is of
                // interest, and if so offload the processing to another system.
            }
        }
    }
}
```

When you load a new game you should see a registering message in the logs like this:
```
19:00:13.045 [main] INFO  o.t.w.g.p.DefaultWorldGeneratorListenerLibrary - Registering FacetListener 'DummyListener'
19:00:13.045 [main] INFO  o.t.w.g.p.DefaultWorldGeneratorListenerLibrary - DummyListener listens to class org.terasology.world.generation.facets.SurfaceHeightFacet
```
The listener will then print out 25 log messages as it is notified:
```
19:00:14.809 [main] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-16, 0, -16), Size: (33, 3, 33))
19:00:14.904 [Chunk-Generator-2] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-64, 0, 32), Size: (32, 64, 32))
19:00:14.905 [Chunk-Generator-5] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-32, -128, 0), Size: (32, 64, 32))
19:00:14.906 [Chunk-Generator-4] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-64, -128, 0), Size: (32, 64, 32))
19:00:14.906 [Chunk-Generator-3] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-64, -64, 0), Size: (32, 64, 32))
19:00:14.908 [Chunk-Generator-6] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-64, 64, 0), Size: (32, 64, 32))
19:00:14.909 [Chunk-Generator-1] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-64, 0, -32), Size: (32, 64, 32))
19:00:14.914 [Chunk-Generator-7] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-64, -64, 32), Size: (32, 64, 32))
19:00:14.917 [Chunk-Generator-0] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-64, 0, 0), Size: (32, 64, 32))
19:00:15.226 [Chunk-Generator-5] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, 0, 0), Size: (32, 64, 32))
19:00:15.236 [Chunk-Generator-0] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, -64, 0), Size: (32, 64, 32))
19:00:15.241 [Chunk-Generator-2] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, 0, -32), Size: (32, 64, 32))
19:00:15.241 [Chunk-Generator-3] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-32, 0, 0), Size: (32, 64, 32))
19:00:15.248 [Chunk-Generator-1] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, 0, 32), Size: (32, 64, 32))
19:00:15.249 [Chunk-Generator-4] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, 64, 0), Size: (32, 64, 32))
19:00:15.262 [Chunk-Generator-7] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (32, 0, 0), Size: (32, 64, 32))
19:00:15.265 [Chunk-Generator-6] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-32, 0, -32), Size: (32, 64, 32))
19:00:15.289 [Chunk-Generator-0] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, -64, 32), Size: (32, 64, 32))
19:00:15.289 [Chunk-Generator-3] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, 0, -64), Size: (32, 64, 32))
19:00:15.292 [Chunk-Generator-5] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-32, -64, 0), Size: (32, 64, 32))
19:00:15.305 [Chunk-Generator-2] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-32, 0, 32), Size: (32, 64, 32))
19:00:15.359 [Chunk-Generator-5] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (-32, 64, 0), Size: (32, 64, 32))
19:00:15.364 [Chunk-Generator-3] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, 0, 64), Size: (32, 64, 32))
19:00:15.366 [Chunk-Generator-0] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, 64, -32), Size: (32, 64, 32))
19:00:15.367 [Chunk-Generator-7] INFO  o.t.coresamplegameplay.DummyListener - Have been notified of facet SurfaceHeightFacet on region (Min: (0, -128, 0), Size: (32, 64, 32))
```


GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/terasology/world/generation/FacetProviderListener.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.world.generation;
+
+import org.terasology.world.generation.facets.SurfaceHeightFacet;
```
4Denthusiast, line 18: SurfaceHeightFacet is now deprecated. This should be updated to SurfacesFacet.

engine/src/main/java/org/terasology/world/generation/FacetProviderListener.java
```diff
@@ -0,0 +1,64 @@
...
+    default void initialize() {
+        // don't do anything
+    }
+
+    /**
+     * Has this listener been initialised. Because listeners can listen to multiple facets it's possible that
+     * initialise may be called more than once.
+     * @return true if this listener has already been initialised, false otherwise.
+     */
+    default boolean isInitialised() {
```
4Denthusiast, line 48: The fact that this method is part of the interface makes it look like initialize wouldn't be called if isInitialised returns true. This isn't actually the case, so what is this method for? If it's only for internal use (i.e. for the listener to keep track of whether it has been initialised yet), it shouldn't be part of the interface. Also, you've spelt "inisialised" two different ways.

engine/src/main/java/org/terasology/world/generation/FacetProviderListener.java
```diff
@@ -0,0 +1,64 @@
...
+     * <br/>
+     * <strong>This call is made from the world generation thread, not the main thread</strong>. As such it should not
+     * do any long running processing and should be thread safe. Ideally it should check if the region-facet combination
+     * is of interest, and if so notify another system to do any actual processing <em>on a separate thread</em>.
+     *
+     * @param region A region that is undergoing generation.
+     * @param facet The facet that was provided to the region.
+     * @param <F>
+     */
+    <F extends WorldFacet> void notify(GeneratingRegion region, F facet);
```
4Denthusiast, line 63: What's the advantage of this polymorphic type over simply having the signature `void notify(GeneratingRegion region, WorldFacet facet)`? It would make sense if the type parameter were an argument of FacetProviderListener. I assume you didn't do that so that one listener could listen for multiple facets, but without doing it that way, I don't see how this makes sense.

engine/src/main/java/org/terasology/world/generator/plugin/WorldGeneratorListenerLibrary.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.world.generator.plugin;
+
+import java.util.List;
+import java.util.Map;
+
+@FunctionalInterface
+public interface WorldGeneratorListenerLibrary {
```
4Denthusiast, line 22: What sort of non-default WorldGeneratorListenerLibrary do you imagine there might be? Why have this as a separate interface?

engine/src/main/java/org/terasology/world/generator/plugin/WorldGeneratorListener.java
```diff
@@ -0,0 +1,28 @@
...
+import org.terasology.world.generation.FacetProviderListener;
+
+/**
+ * A category of {@link Class}es that would like to know about certain aspects of world generation, but do not seek
+ * to <em>influence</em> world generation.
+ * <br/>
+ * Current subclasses are:
+ * {@link FacetProviderListener}
+ */
+public interface WorldGeneratorListener {
```
4Denthusiast, line 27: This relates somewhat to my overall comment that you might as well run the listeners at the end once all the data is available, but why would you need other sorts of WorldGenerationListener, and even if this did have multiple sub-interfaces, what would be the point of this interface, which doesn't have any methods? If there's nothing that you can do with both a FacetProviderListener and (hypothetically) a RasterizerListener (not that I think that actually makes much sense), where would you need to deal with something that's either one or the other, but you don't care which?

modules/Core/src/main/java/org/terasology/core/world/generator/worldGenerators/PerlinFacetedWorldGenerator.java
```diff
@@ -82,6 +86,7 @@ protected WorldBuilder createWorld() {
                         //.addRasterizer(new GroundRasterizer(blockManager))
                 .addRasterizer(new SolidRasterizer())
                 .addPlugins()
+                .addListeners()
```
4Denthusiast, line 89: The WorldBuilder doesn't need to keep track of the order of the listeners relative to anything else, and you can't add individual listeners in the WorldBuilder as non-plugins, because the addListener method doesn't have the right signature for that, so addListeners might as well be integrated into the WorldBuilder constructor rather than being called later.

4Denthusiast (changes requested):
I think that it would be more useful if the listeners were simply called after all of the facet providers and all of the rasterizers are finished. That way, they have access to all of the data, which both simplifies the logic of registering and calling listeners, and makes it easier for a listener to make use of multiple facets at once, for example, a listener relating to doing something with caves might need both the CavesFacet and the ElevationFacet, to do its thing only below a certain depth. Also, sometimes having the finished chunk available might be useful too.

4Denthusiast:
I expect that most listeners would basically put the data needed in some concurrency-compatible datastructure to be picked up by a system on the main thread, so if you could implement that as part of this system instead, so that the listeners get called in the main thread, that would be useful to avoid duplication. If you do make that change but don't move the listeners to the end of world generation, there would be a small risk of concurrency errors in the case that either a listener or a Provider or Rasterizer without the appropriate annotation updates a facet (although that would generally be a bad idea already).

modules/Core/src/main/java/org/terasology/core/world/generator/worldGenerators/PerlinFacetedWorldGenerator.java
sin3point14: +1 to this, it'd also save the time of going around and updating every worldgen module out there and this seems like a important enough feature to be added by default

sin3point14 (changes requested):
I'd like to know how many times will the `notify` function be called if a facet with extended borders which spans across multiple chunks is registered by a listener? Looking at the logs, I have a feeling that it'd notify for every chunk(as a facet is created for every chunk the extended facet lies in) which I think kind of defeats the purpose of detecting the specific features, though i could be wrong here.

askneller:
Thanks for the review @4Denthusiast , @sin3point14, i'll try and get to this soon.

jdrueckert:
Hey there @askneller any news on integrating the review comments?
Looks like it needs some conflict resolution, too :upside_down_face:

askneller:
Hi @jdrueckert, I'll be able to look at this later this week, Maybe Wednesday or Thursday

engine/src/main/java/org/terasology/world/generation/FacetProviderListener.java
askneller: Fixed this

askneller:
So from what I can gather following the code, here is my understanding of the chunk generation process:

1. The ```RelevanceSystem``` keeps track of which chunks are required to be loaded, based on their proximity to players. 
2. During the system's update it may call ```chunkProvider.createOrLoadChunk``` when required. The provider (using as an example ```LocalChunkProvider```) will submit a task to its ```ChunkProcessingPipeline``` 
3. That will load the chunk if possible. If not possible the task calls ```WorldGenerator.createChunk```, which in turn calls ```World.rasterizeChunk```. 
4. rasterizeChunk will create a ```Region``` for the chunk, loop through all of the rasterizers and invoke ```rasterizer.generateChunk```, passing in the ```Chunk``` and ```Region```. 
5. When a rasterizer calls ```Region.getFacet```, _this_ is when ```RegionImpl``` will loop though the facet provider chains, processing the providers as necessary.

If I'm not mistaken the call to ```WorldGenerator.createChunk``` (at 3) is the beginning of the off-main-thread work, is that correct? Can anyone with better knowledge of world generation suggest what might be the best way to get the facet information back onto the main thread? From what I can tell all generated facet information is stored in the Region, and isn't preserved anywhere outside ```WorldImpl.rasterizeChunk```.

engine/src/main/java/org/terasology/world/generation/FacetProviderListener.java
askneller: Don't know I was thinking there. Removed from the interface.

engine/src/main/java/org/terasology/world/generation/FacetProviderListener.java
askneller: Good point, I've just made it ```WorldFacet```.

engine/src/main/java/org/terasology/world/generator/plugin/WorldGeneratorListenerLibrary.java
askneller: I guess I was cargo cult following the plugin library style. I've removed the interface and just made it a single class.

modules/Core/src/main/java/org/terasology/core/world/generator/worldGenerators/PerlinFacetedWorldGenerator.java
askneller: I've integrated this into the constructor.

engine/src/main/java/org/terasology/world/generator/plugin/WorldGeneratorListener.java
askneller: As above I guess I was cargo cult-ing again. I'll rename ```FacetProviderListener``` to ```WorldGeneratorListener``` and make it the one interface.

DarkWeird:
> So from what I can gather following the code, here is my understanding of the chunk generation process:
> 
>     1. The `RelevanceSystem` keeps track of which chunks are required to be loaded, based on their proximity to players.
> 
>     2. During the system's update it may call `chunkProvider.createOrLoadChunk` when required. The provider (using as an example `LocalChunkProvider`) will submit a task to its `ChunkProcessingPipeline`
> 
>     3. That will load the chunk if possible. If not possible the task calls `WorldGenerator.createChunk`, which in turn calls `World.rasterizeChunk`.
> 
>     4. rasterizeChunk will create a `Region` for the chunk, loop through all of the rasterizers and invoke `rasterizer.generateChunk`, passing in the `Chunk` and `Region`.
> 
>     5. When a rasterizer calls `Region.getFacet`, _this_ is when `RegionImpl` will loop though the facet provider chains, processing the providers as necessary.
> 
> 
> If I'm not mistaken the call to `WorldGenerator.createChunk` (at 3) is the beginning of the off-main-thread work, is that correct? Can anyone with better knowledge of world generation suggest what might be the best way to get the facet information back onto the main thread? From what I can tell all generated facet information is stored in the Region, and isn't preserved anywhere outside `WorldImpl.rasterizeChunk`.

you right in steps 1-3. about 4,5 - idk.

One way - enrich `Chunk` with those data.
Second way - returns from `WorldGenerator.createChunk` not chunk, but data class with chunk and additional data.(like metadata or facet-ralated data_

jdrueckert:
Hey there @askneller 👋 happy new year!
Are you currently still blocked here or did @DarkWeird's proposed solutions help?
I see that there was some uncertainty about world gen stuff, so I'm pinging @4Denthusiast for help.

The main question seemed to be the following:
> Can anyone with better knowledge of world generation suggest what might be the best way to get the facet information back onto the main thread?

askneller:
Hello @jdrueckert , thank you, and happy new year to you too!

I'm not really blocked, I've just been on a break. As for what DarkWeird suggested I think maybe the second method would be better, as the first seems as though it might blow out the chunk size when it's saved. Does that sound ok?

4Denthusiast:
I agree that it would be best to not permanently save all of the facet data automatically. I would also recommend that you merge this data into the existing `OnChunkGenerated` event, which has a similar purpose.

4Denthusiast:
Sorry to have taken a circuitous route to this point, but I basically don't think this feature is actually worth the added complexity that would be required. I looked into the chunk generation system to work out how I would get the facet data from the `Region` in `WorldRegion.rasterizeChunk` to the `OnChunkGenerated` event, and it seems like it would be quite messy (and there's also a risk of having to send a lot more data over the network if that event gets replicated to the client in multiplayer, but it would still seem like a bad idea to have this not be an event, and also I don't actually know exactly how events interact with multiplayer).

The other reason I don't think this feature is worth the complexity is that it seems to me like most things that could be done by having the facet data available to other systems could be done without it, either by putting them into the world generator, or by using the information that's available afterwards in blocks, entities and extraData. For your example of generating towers, getting information from outside the current chunk's region is what borders are for, and the `ElevationFacet` (probably the most relevant for finding mountain peaks) usually has a border big enough that it's twice the size of the chunk itself in each direction already, and this could be extended easily and not too computationally expensively (because it's only a 2D facet, so it's pretty cheap to compute). Alternatively, you could add the towers with a delay, but just use the block data directly rather than facet data. Either of these approaches would probably work better than saving the facet data for later (e.g. neither of them would suffer from the problem where the blocks might be changed before the tower was generated, leaving it misaligned with the terrain).

If you have a more concrete proposal for a feature that you actually intend to implement soon based on this PR, then I might be convinced to approve this (subject to the other changes already mentioned). Similarly, if you do manage to work out a neat way to get the facet data into the `OnChunkGenerated` event without making the code much more complex, messy or inefficient, then it might be worth merging, but I don't think that that will actually be possible.

I'm leaving this PR open for now in case you do want to try to fix it anyway, or other people disagree, but if I have convinced you to give up on it, then please close the PR so it's not cluttering up the list, and we can maybe make an issue or something instead to note thoughts on this feature.

askneller:
Fair enough. Having looked more into borders you are right that this isn't really necessary. Thanks for the comments all. I'll close this now.

PR #3873, opened 2020-03-30
Fixed UI bug at Advanced Game Setup window, Issue #3818

### LINKED ISSUE #3818: Double clicking an un-downloaded module messes up UI.
### What I did
You can activate a module by either clicking the Activate button or by double clicking the module name. Generally modules cannot be activated if they haven't been downloaded, since the activate button is grayed out, **but** you can still double click them (**if** you haven't activated/deactivated a module from the list yet). This broke quite some features of the UI
### What happened
The activate button changed it's text from "Deactivated" to "Activated" despite the module not being present locally.
unknown

This also caused the module highlight feature to break. Activated modules no longer showed a green color.
unknown (1)
Also, the activated modules could no longer be (apparently) deactivated.
### How to reproduce

* Open Terasology and try making a new world
* Double click on a module which has not been downloaded yet
* Notice how the activate button changes text but remains grayed out
* Try activating a downloaded module and notice how it does not change it's color.
* Try deactivating any downloaded and activated module, it would not work.

stefaniamak:
### Result

Double-clicking on a module that has not been downloaded, doesn't break the Advanced Game Setup selection anymore.

### Original Issue 

* UI falsely indicated that (not-downloaded) module as active after double click
* Activate-Deactivate button of an activated module remained "Deactivate" no matter how many times clicked
* Activated modules no longer were in green color

Closes #3818 

### Changes

* Moved the `refreshSelection();` from `onOpened()` to `updateModuleInformation()`

## How to test

* Go to the Advanced settings option while creating a new world in Terasology
* Double click on a module that has not been downloaded yet

Note: The issue occurred only during **the first double click**. Meaning, if you double click an already downloaded module, and afterward double click one that is not, the issue does not occur even without the changes.

## Outstanding before merging

I was unsure if the `refreshSelection();` at `onOpened()` influences other sections irrelevant to the issue I was working on, thus I left it as a comment instead of deleting it completely. The issue does not occur either way. Should be deleted if not retrieved.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
@@ -124,7 +124,7 @@ public class AdvancedGameSetupScreen extends CoreScreenLayer {
             info.setExplicitSelection(config.getDefaultModSelection().hasModule(info.getMetadata().getId()));
         }
 
-        refreshSelection();
+        //refreshSelection();
         filterModules();
     }
 
@@ -694,6 +694,7 @@ public class AdvancedGameSetupScreen extends CoreScreenLayer {
         allSortedModules.addAll(sortedModules);
 
         filterModules();
+        refreshSelection();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


GooeyHub:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
```diff
@@ -124,7 +124,7 @@ public void onOpened() {
             info.setExplicitSelection(config.getDefaultModSelection().hasModule(info.getMetadata().getId()));
         }
 
-        refreshSelection();
+        //refreshSelection();
```
skaldarnar: This can indeed be removed.
```suggestion
```

PR #3880, opened 2020-04-07
Added Gitpod files

casals:
### Contains

Files required to run Terasology using Gitpod

### How to test

- Create an account on Gitpod (easier with Github credentials)
- Create a new workspace using this PR link: https://github.com/MovingBlocks/Terasology/pull/<PR number>
- After the workspace is created and the init scripts are executed you should see Terasology's initial screen shown in the VNC machine panel

Related discussion: https://community.gitpod.io/t/lwjgl-openal/1105

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,12 @@
+image: 
+  file: gitpod.dockerfile
+
+tasks:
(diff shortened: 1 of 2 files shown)


GooeyHub:

PR #3885, opened 2020-04-08
bumped version of joml and resolved direction of lightshafts

pollend:
Looks like the transform for light shafts was in the wrong direction so they were facing away from the light source. applying a transpose to the viewmatrix corrects the direction even. 

before: 
image

after:
image

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/dag/nodes/LightShaftsNode.java
+++ b/engine/src/main/java/org/terasology/rendering/dag/nodes/LightShaftsNode.java
@@ -15,6 +15,7 @@
  */
 package org.terasology.rendering.dag.nodes;
 
+import org.joml.Matrix4f;
 import org.joml.Vector3f;
 import org.joml.Vector4f;
 import org.terasology.assets.ResourceUrn;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/rendering/dag/nodes/LightShaftsNode.java
```diff
@@ -146,7 +147,7 @@ public void process() {
 
         sunPositionWorldSpace4.set(sunDirection.x * 10000.0f, sunDirection.y * 10000.0f, sunDirection.z * 10000.0f, 1.0f);
         sunPositionScreenSpace.set(sunPositionWorldSpace4);
-        activeCamera.getViewProjectionMatrix().transform(sunPositionScreenSpace);
+        new Matrix4f(activeCamera.getViewProjectionMatrix()).transpose().transform(sunPositionScreenSpace);
```
httpdigest, line 150: [Matrix4f.transformTranspose(Vector4f)](https://github.com/JOML-CI/JOML/commit/094f9fd49afec20eb3d46b03e16a411c9be2fe4e) will be available in JOML 1.9.25.

pollend: that's nice to know about. I'll keep that in mind when we do another version bump.

Issue #3670, opened 2019-05-01
Time out of sync for players in multiplayer


mayant15:
### What you were trying to do
Hosted a game locally and connected to it with another client.

### What actually happened
The player that connected later did not have the same time as the host.

### How to reproduce
* Start a multiplayer game locally
* Start up another client
* Wait 5-10 minutes before connecting

![day-night](https://user-images.githubusercontent.com/32035777/57034699-e0c8c580-6c6d-11e9-8e0c-0c2afe11f2cb.jpg)

jellysnake:
This should be easily fixable by sending the world time (and dilation) when the player is connected and using that rather than starting from zero

Cervator:
On a related note the time changing shortcuts (like the arrow keys in debug mode) also exclusively change the time on the client who does it. Time should be synchronized from the server to the clients and if some modification is done to time that request should be sent to the server to enact and broadcast.

jdrueckert:
Fixed by #4292

PR #3890, opened 2020-04-15
Time out of sync for players in multiplayer

godfather2:
### Contains

Attempts to fix https://github.com/MovingBlocks/Terasology/issues/3670
The World time changed using debug mode by any client is now reflected in all clients.
This sketch board explains how i fixed it,by comparing it against the Ping System: https://sketchboard.me/fB4oPzAlhMkt#/.

### How to test

Start up a multiplayer game locally and change the World time using the debug mode,and you will see the time change being replicated in all clients.

The client that connects later lags the other clients by around 0.03 days. This is because the World time is updated on every frame, and the first frame is run some time after the new client's WorldProvider receives the host's World time. I tried fixing it by sending an event to reset time, right before the first frame of the new client(using the postBegin() method). But for some reason, the event handler is unable to catch this event. I am guessing that i am not sending the event using the right entity.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
@@ -31,12 +31,12 @@ import org.terasology.input.events.KeyEvent;
 import org.terasology.input.events.MouseAxisEvent;
 import org.terasology.logic.characters.CharacterComponent;
 import org.terasology.logic.debug.DebugProperties;
+import org.terasology.logic.players.event.WorldtimeResetEvent;
 import org.terasology.network.ClientComponent;
 import org.terasology.registry.In;
 import org.terasology.rendering.nui.NUIManager;
 import org.terasology.rendering.nui.layers.ingame.metrics.DebugOverlay;
 import org.terasology.world.WorldProvider;
-
 
 @RegisterSystem(RegisterMode.CLIENT)
 public class DebugControlSystem extends BaseComponentSystem {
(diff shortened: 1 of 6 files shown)


engine/src/main/java/org/terasology/logic/players/DebugControlSystem.java
```diff
@@ -164,4 +160,9 @@ public void onMouseX(MouseAxisEvent event, EntityRef entity) {
         }
     }
 
+    private void timeTravel(EntityRef entity, KeyEvent event, float timeInDays) {
```
Cervator, line 163: Protip: even if this is a private method it is a good place to add Javadoc to make it clear what the method does :-)

engine/src/main/java/org/terasology/logic/players/event/FindClientSystem.java
```diff
@@ -0,0 +1,44 @@
...
+import org.terasology.entitySystem.event.ReceiveEvent;
+import org.terasology.entitySystem.systems.BaseComponentSystem;
+import org.terasology.entitySystem.systems.RegisterMode;
+import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.logic.time.WorldtimeResyncEvent;
+import org.terasology.network.ClientComponent;
+import org.terasology.registry.In;
+
+@RegisterSystem(RegisterMode.AUTHORITY)
+public class FindClientSystem extends BaseComponentSystem {
```
Cervator, line 29: I think this could be named better? It does more than just find clients, it sends them time resync events. Seems more like it should be a `SynchronizeClientTimeSystem` or something?

pollend:
> Code looks fine, just a couple comments +1
> 
> However, I can't get it to work :-(
> 
> My test setup:
> 
>     * One client that hosts a multiplayer game with CoreSampleGameplay
> 
>     * Second client that joins
> 
>     * Observed that the second client _did_ seem to join with a time more in sync with the host than starting at the default value
> 
>     * Attempted arrow keys to change time on either game window, neither worked at all - didn't change time on the local client or the other :-(
> 
>     * Attempted use of `setWorldTime 0.9` on either client to see if they would sync with the other client - also didn't change the time for the _other_ client, but did change the time correctly on the local client

I can't get this to work either. do we sync up time if it is changed through a debug command?

Cervator:
> I can't get this to work either. do we sync up time if it is changed through a debug command?

Not at present, as far as I know, which is one of the desires - time should stay in sync in whichever way it may be changed :-)

PR #3891, opened 2020-04-16
 chore: Add JavaDoc and fix some prefabs

skaldarnar:
doc: Add notes about event target and sender entity to ChatMessageEvent

fix: Remove components from prefabs that are no longer in engine
	- MinionMove is in Behaviors
	- Breather is in Breathing

### DIFF
--- a/engine/src/main/java/org/terasology/logic/chat/ChatMessageEvent.java
+++ b/engine/src/main/java/org/terasology/logic/chat/ChatMessageEvent.java
@@ -26,7 +26,7 @@ import org.terasology.network.OwnerEvent;
 /**
  * Indicate that a message should be printed to the chat console.
  * <p>
- * Needs to be send against a client entity, e.g., the targeted entity should have a {@link org.terasology.network.ClientComponent}.
+ * Needs to be sent against a client entity, e.g., the targeted entity should have a {@link org.terasology.network.ClientComponent}.
  */
 @OwnerEvent
 public class ChatMessageEvent implements MessageEvent {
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/terasology/logic/chat/ChatMessageEvent.java
```diff
@@ -24,8 +24,9 @@
 import org.terasology.network.OwnerEvent;
 
 /**
- * A chat message
- *
+ * Indicate that a message should be printed to the chat console.
+ * <p>
+ * Needs to be send against a client entity, e.g., the targeted entity should have a {@link org.terasology.network.ClientComponent}.
```
jdrueckert, line 29:
```suggestion
 * Needs to be sent against a client entity, e.g., the targeted entity should have a {@link org.terasology.network.ClientComponent}.
```

engine/src/main/java/org/terasology/logic/chat/ChatMessageEvent.java
```diff
@@ -35,6 +36,17 @@
     protected ChatMessageEvent() {
     }
 
+    /**
+     * A chat message is associated with an entity that sent it.
```
jdrueckert, line 40:
```suggestion
     * A chat message is associated with the entity that sent it.
```

engine/src/main/java/org/terasology/logic/chat/ChatMessageEvent.java
```diff
@@ -35,6 +36,17 @@
...
+     * <p>
+     * This entity may often be a player, but it can be any entity with a {@link org.terasology.logic.common.DisplayNameComponent}.
+     * If the entity also has a {@link org.terasology.network.ColorComponent} the sender's name will be colorized.
```
jdrueckert, line 43:
```suggestion
     * If the entity also has a {@link org.terasology.network.ColorComponent}, the sender's name will be colorized.
```

jdrueckert: also not sure whether it's "colorized" or "coloured"

skaldarnar: 🤷‍♂️ 🎨

Issue #2611, opened 2016-11-18
VR support follow-ups and potential extraction of the VR Provider eventually


Cervator:
There is no rush on this but I wanted to record it all as future reminders.

@IndianaJohn did a great amount of work in #2479 adding support for OpenVR, and embedded it into Terasology for the time being. It could live as an external library with some pros and cons.

He's got a repo over at https://github.com/indianajohn/jopenvrwrapper

* `OpenVRProvider.java` uses `com.sun` packages which is discouraged, but this might be required as part of using JNA - which is indeed in such a package
* Per PR notes the provider is paired to our engine dependency on JNA, with potential impact if that changes in the future
* The embedding of the provider inside the Terasology engine was also due to a shared dependency on LWJGL (which is another topic of future interest - do we go to LWJGL 3 or LibGDX?)
  * To get around having the provider depend on a potentially different version of LWJGL you could simply use a Gradle `provided` dependency edit in our engine to avoid a version conflict. So long as it is a minor difference, of course, probably wouldn't quite work if one depends on LWJGL 2 and the other LWJGL 3 ...
* The VR support uses JOML for math instead of adding more stuff to TeraMath and we may indeed be able to slowly migrate to JOML overall. @Immortius recently got past one major obstacle in getting some stuff added to JOML. @msteiger also likes math and I like pinging people :-)
* We might want to check on how OSVR works when @tdgunes manages to get a hold of the headset we got from OSVR for testing. Or maybe @emanuele3d will be able to get his hands on a headset a little longer soon :D
* We might also want to extract out the `libopenvr` natives, akin to how LWJGL does it, and extract them at Gradle prep time the same way (rather than keep them in source). This is a hassle, but oh well.
* There is some generated source code in the provider as well as some packages without a domain. If we really want to go nuts we may want to do some cleanup to get a tidier setup (but that's mostly effort for aesthetics)
* We can probably retire/archive the [TeraOVR repo](https://github.com/MovingBlocks/TeraOVR) sometime
* Maybe we should switch the "Oculus Rift Support" option in the Video Settings back on as "VR Support" or do we not want to expose it yet? I guess it may also require a game restart? It could pop an info box explaining that if needed

tdgunes:
[...]

PR #3897, opened 2020-04-19
Comment out a bit more VR stuff to avoid a native resolution issue in module standalone workspaces

Cervator:
The way we build modules standalone has tended to have a fair share of quirks since the Gradle from this engine workspace gets reused in clever/evil/dumb ways. But long story short one such quirk goes away with this PR, even if it is only a workaround for #3415. Really #2611 and making VR work again should be the proper fix, which would improve handling the OpenVR natives files and so on.

In the meantime this gets us one step closer to better _advanced_ tests in module jobs in Jenkins or anywhere else other than a full engine workspace + modules. Normal unit tests are fine, but those requiring a running game environment tend to be more tricky.

See https://github.com/Terasology/ModuleTestingEnvironment/issues/12 for the next problem I've encountered. 

Testing this is problematic, really can only test in a local workspace that the game acts normally. To test it more properly need to try a standalone module workspace which that MTE issue can explain, but even so that relies on a binary version of the engine. As of this writing the most recent one from Nanoware/Terasology will correctly show what this PR changes, while the most recent one from MovingBlocks/Terasology will show the old issue.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
+++ b/engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
@@ -15,21 +15,18 @@
  */
 package org.terasology.logic.players;
 
-import jopenvr.VRControllerState_t;
 import org.joml.Matrix4f;
 import org.terasology.entitySystem.Component;
 import org.terasology.entitySystem.Owns;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.math.TeraMath;
 import org.terasology.math.geom.Vector3f;
-import org.terasology.rendering.openvrprovider.ControllerListener;
 import org.terasology.math.geom.Quat4f;
-import org.terasology.rendering.openvrprovider.OpenVRProvider;
 
 /**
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
```diff
@@ -78,6 +63,7 @@ public void buttonStateChanged(VRControllerState_t stateBefore, VRControllerStat
      * @param pose - the controller pose - a homogenous transformation matrix.
      * @param handIndex - the hand index - 0 for left and 1 for right.
      */
+    /* commented out due to a natives issue and VR not working at the moment anyway
```
skaldarnar, line 66: `//TODO:` has nice discoverability in many editors, IDEs, and code analysis toosl 😉 

```suggestion
    //TODO: commented out due to a natives issue and VR not working at the moment anyway
    /*
```

Cervator: Yep not sure why I didn't do that. Obvious improvement is obvious :-) Thanks!

PR #3899, opened 2020-04-20
feat(JOML): Migrate methods of WorldProvider

pollend:
I will start going through some of the more major systems in Terasology and provide JOML alternatives that we can start to use in place of the teramath implementations. we should be able to start removing the termath methods when these endpoints get updated in module space. I think I will start to do this for some of the more major systems that terasology uses to make the transition easier. Components are another issue that will be harder to address.

These methods are now backed by their joml equivalent so any actions on a given block should be the same from before. I expect breaking blocks and placement of blocks should still work the same regardless.

### DIFF
--- a/engine-tests/src/test/java/org/terasology/world/EntityAwareWorldProviderTest.java
+++ b/engine-tests/src/test/java/org/terasology/world/EntityAwareWorldProviderTest.java
@@ -270,6 +270,16 @@ public class EntityAwareWorldProviderTest extends TerasologyTestingEnvironment {
     }
 
     @Test
+    public void testEntityBecomesTemporaryWhenChangedFromAKeepActiveBlockJoml() {
+        worldProvider.setBlock(new org.joml.Vector3i(), keepActiveBlock);
+        EntityRef blockEntity = worldProvider.getBlockEntityAt(new Vector3i(0, 0, 0));
+        worldProvider.setBlock(new org.joml.Vector3i(), airBlock);
+        worldProvider.update(1.0f);
+        assertFalse(blockEntity.isActive());
+    }
+
+
+    @Test
     public void testEntityBecomesTemporaryIfForceBlockActiveComponentRemoved() {
         EntityRef blockEntity = worldProvider.getBlockEntityAt(new Vector3i(0, 0, 0));
         blockEntity.addComponent(new ForceBlockActiveComponent());
--- a/engine/src/main/java/org/terasology/world/WorldProvider.java
+++ b/engine/src/main/java/org/terasology/world/WorldProvider.java
@@ -40,7 +40,7 @@ public interface WorldProvider extends WorldProviderCore {
     boolean isBlockRelevant(Vector3i pos);
 
     /**
-     *
+     * An active block is in a chunk that is available and fully generated.
      * @param pos The position
      * @return Whether the given block is active
      */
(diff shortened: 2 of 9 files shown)


engine-tests/src/test/java/org/terasology/world/EntityAwareWorldProviderTest.java
```diff
@@ -269,6 +269,16 @@ public void testEntityBecomesTemporaryWhenChangedFromAKeepActiveBlock() {
         assertFalse(blockEntity.isActive());
     }
 
+    @Test
+    public void testEntityBecomesTemporaryWhenChangedFromAKeepActiveBlockJoml() {
```
Cervator, line 273: I love this test name 👍

engine/src/main/java/org/terasology/world/WorldProvider.java
```diff
@@ -29,19 +31,49 @@
...
      */
+    @Deprecated
     boolean isBlockRelevant(Vector3i pos);
 
+    /**
+     *
+     * @param pos The position
+     * @return Whether the given block is active
+     */
+    boolean isBlockRelevant(Vector3ic pos);
```
Cervator, line 47: Should we consider (in general) adding default interface implementations when adding new methods like this? That could avoid having to update _some_ things in modulespace. In this case I'm not sure it makes sense as I only see one non-engine usage over in FlexiblePathfinding.

engine/src/main/java/org/terasology/world/WorldProvider.java
```diff
@@ -29,19 +31,49 @@
...
+     * @param pos The position
      * @return Whether the given block is active
+     * @deprecated This is scheduled for removal in an upcoming version
+     *             method will be replaced with JOML implementation {@link #isBlockRelevant(Vector3ic)}.
      */
+    @Deprecated
     boolean isBlockRelevant(Vector3i pos);
 
+    /**
+     *
```
Cervator, line 43: Some sort of descriptor for the method itself here please :-) Looks like the old method has one, can just copy it forward.

engine/src/main/java/org/terasology/world/internal/WorldProviderWrapper.java
```diff
@@ -44,26 +46,51 @@ public boolean isBlockRelevant(Vector3i pos) {
...
+    }
+
     @Override
     public boolean isBlockRelevant(Vector3f pos) {
         return isBlockRelevant(new Vector3i(pos, RoundingMode.HALF_UP));
     }
 
+    @Override
+    public boolean isBlockRelevant(Vector3fc pos) {
+        return isBlockRelevant(new org.joml.Vector3i(pos, org.joml.RoundingMode.HALF_UP));
```
Cervator, line 61: Explicit package like this makes me sad - can we avoid it via imports or is something in the way of that?

pollend: nope, Vector3i conflicts with Vector3i from termath.

engine/src/main/java/org/terasology/world/WorldProvider.java
pollend: It can't hurt at thh moment but a lot of this is temporary and I all the old methods are going to go away anyways.

engine/src/main/java/org/terasology/world/WorldProvider.java
skaldarnar:
```suggestion
     * An active block is in a chunk that is available and fully generated.
```

PR #3900, opened 2020-04-20
Feature/joml conversion full

kBlaszczyk:
This is a continuation of #3857.

I fixed lots of joml misusages and improved the usage of the library along the way.
I've found some problems in the Camera class which I pointed out with a FixMe. These should be addressed in another PR.

To run this PR, the Health and Inventory module need to be set to the following code base:
https://github.com/kBlaszczyk/Health/tree/refactoring/joml-conversion
https://github.com/kBlaszczyk/Inventory/tree/refactoring/joml-conversion

Once these are checked out, the tests and the game should run.


engine-tests/src/test/java/org/terasology/math/MatrixUtilsTest.java
```diff
@@ -0,0 +1,157 @@
+package org.terasology.math;
```
kBlaszczyk, line 1: I heavily refactored the MatrixUtils. These tests verify that MOST of it is behaving just like the old implementation.
MOST because I'm pretty sure to have detected an issue in the old implementation. I'll point to that in the appropriate change.

engine-tests/src/test/java/org/terasology/math/Vector3iTest.java
```diff
@@ -189,6 +184,8 @@ public void testHash() {
     }
 
     @Test
```
kBlaszczyk, line 186: Disabled Test. I pointed the reason out. Please comment.

engine/src/main/java/org/terasology/logic/nameTags/NameTagClientSystem.java
```diff
@@ -85,7 +85,7 @@ private void createOrUpdateNameTagFor(EntityRef entity, NameTagComponent nameTag
             nameTag = nameTagBuilder.build();
             nameTagEntityToFloatingTextMap.put(entity, nameTag);
 
-            Location.attachChild(entity, nameTag, offset, new Quat4f(1, 0, 0, 0));
+            Location.attachChild(entity, nameTag, offset, new Quaternionf(1, 0, 0, 0));
```
kBlaszczyk, line 88: Is this a useful quaternion or a mistake? I'm tempted to say it's supposed to be the identity quaternion.

engine/src/main/java/org/terasology/logic/players/FirstPersonHeldItemMountPointComponent.java
```diff
@@ -85,13 +84,8 @@ public void poseChanged(Matrix4f pose, int handIndex) {
...
         trackingDataReceived = true;
-        Matrix4f adjustedPose = pose.mul(toolAdjustmentMatrix);
-        translate = new Vector3f(adjustedPose.m30(), adjustedPose.m31(), adjustedPose.m32());
-        org.joml.Vector4f jomlQuaternion = org.terasology.rendering.openvrprovider.OpenVRUtil.convertToQuaternion(adjustedPose);
-        if (rotationQuaternion == null) {
-            rotationQuaternion = new Quat4f(jomlQuaternion.x, jomlQuaternion.y, jomlQuaternion.z, jomlQuaternion.w);
-        } else {
-            rotationQuaternion.set(jomlQuaternion.x, jomlQuaternion.y, jomlQuaternion.z, jomlQuaternion.w);
-        }
+        Matrix4f adjustedPose = pose.mul(toolAdjustmentMatrix, new Matrix4f());
```
kBlaszczyk, line 87: I made the null check obsolete by making the members final. The client shouldn't be able to reset or set them to null anyway. The rest is simplification.

engine/src/main/java/org/terasology/math/MatrixUtils.java
```diff
@@ -61,14 +38,8 @@ public static FloatBuffer matrixToFloatBuffer(Matrix4fc m) {
      * @return A new FloatBuffer containing the matrix in column-major form.
      * @deprecated used JOML method that uses Matrix3fc
```
kBlaszczyk, line 39: I'm not sure about deprecating these methods, since they still add value by taking care of the buffer creation.
If anything I'd deprecate `matrixToFloatBuffer(Matrix3fc m, FloatBuffer fb)`, but they aren't currently marked for deprecation. ("they" as in both, the mat3 and mat4 variations)

engine/src/main/java/org/terasology/math/MatrixUtils.java
```diff
@@ -102,159 +74,43 @@ public static FloatBuffer matrixToFloatBuffer(Matrix3f m, FloatBuffer fb) {
...
+            float eyeX, float eyeY, float eyeZ,
+            float centerX, float centerY, float centerZ,
+            float upX, float upY, float upZ
+    ) {
+        return new Matrix4f().setLookAt(
+                eyeX, eyeY, eyeZ,
+                centerX, centerY, centerZ,
+                upX, upY, upZ
+        );
     }
```
kBlaszczyk, line 91: NOTE:
The original implementation of this method did:
Matrix4f().setLookAt(...).setTranslation(-eyeX, -eyeY, -eyeZ)
I removed setting the translation component because I trust joml more. Opinions?

engine/src/main/java/org/terasology/persistence/typeHandling/TypeHandlerLibrary.java
```diff
@@ -198,17 +168,11 @@ private static void populateWithDefaultHandlers(TypeHandlerLibrary serialization
 
         serializationLibrary.addTypeHandlerFactory(new TextureRegionAssetTypeHandlerFactory());
 
-        serializationLibrary.addTypeHandler(Vector4f.class, new LegacyVector4fTypeHandler());
-        serializationLibrary.addTypeHandler(Vector3f.class, new LegacyVector3fTypeHandler());
-        serializationLibrary.addTypeHandler(Vector2f.class, new LegacyVector2fTypeHandler());
-        serializationLibrary.addTypeHandler(Vector3i.class, new LegacyVector3iTypeHandler());
-        serializationLibrary.addTypeHandler(Vector2i.class, new LegacyVector2iTypeHandler());
```
kBlaszczyk, line 205: I'm PRETTY sure these can be safely removed. It is possible however, that they are still used to serialize objects containing TeraMath related members.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
```diff
@@ -36,9 +36,11 @@
 public abstract class Camera {
 
     /* CAMERA PARAMETERS */
-    protected final Vector3f position = new Vector3f(0, 0, 0);
+    protected final Vector3f position = new Vector3f();
     protected final Vector3f up = new Vector3f(0, 1, 0);
-    protected Vector3f viewingDirection = new Vector3f(1, 0, 0);
+    // FixMe this is clearly not the viewing direction, it's the rotation axis
+    // The usage of this value outside of this class should be thoroughly analyzed when fixing this.
```
kBlaszczyk, line 42: This has to be addressed. There need to be getters for both the actual viewing direction and the axis angle. Some clients are using the viewing axis, (most likely) thinking it is the viewing direction.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
```diff
@@ -205,17 +207,14 @@ public Vector3f getViewingDirection() {
      * Get the orientation of the camera.
      * @return the orientation direction, a quaternion.
      */
-    public Quat4f getOrientation() {
-        return new Quat4f(JomlUtil.from(viewingDirection), viewingAngle);
```
kBlaszczyk, line 209: This is the proof, that `viewingDirection` is the viewing axis.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
```diff
@@ -205,17 +207,14 @@ public Vector3f getViewingDirection() {
...
-     Try to set the viewing direction.
-     * @param direction
-     */
-    public void setOrientation(Quat4f direction) {
-        viewingDirection = new Vector3f(direction.x,direction.y,direction.z);
-        viewingAngle = direction.getAngle();
+    public void setOrientation(Quaternionf quaternion) {
+        AxisAngle4f axisAngle = new AxisAngle4f(quaternion);
+        viewingDirection.set(axisAngle.x, axisAngle.y, axisAngle.z);
+        viewingAngle = axisAngle.angle;
```
kBlaszczyk, line 217: Note that the old implementation wrongfully interpreted the provided quaternion as an AxisAngle, which is mathematically wrong. I fixed this, but client code needs to be reviewed urgently.

engine/src/main/java/org/terasology/math/MatrixUtils.java
pollend: I would probably try to get rid of MatrixUtils. It doesn't really add much especially if we are just wrapping a single function.

engine/src/main/java/org/terasology/rendering/collada/ColladaLoader.java
```diff
@@ -466,14 +456,14 @@ private MD5Joint createMD5Joint(Element jointNodeElement) throws ColladaParseExc
...
+        Quaternionf[] rotationArray = new Quaternionf[inverseBindMatrixArray.length / 16];
         for (int i = 0; i < inverseBindMatrixArray.length / 16; ++i) {
             int offset = i * 16;
-            Matrix4f matrix4f = new Matrix4f(Arrays.copyOfRange(inverseBindMatrixArray, offset, offset + 16));
-            Quat4f rotation = new Quat4f();
-            rotation.set(matrix4f);
-            rotationArray[i] = rotation;
+            Matrix4f matrix = new Matrix4f(FloatBuffer.wrap(
+                    Arrays.copyOfRange(inverseBindMatrixArray, offset, offset + 16)
+            )).transpose();
```
kBlaszczyk, line 465: I assume the old implementation worked. If so, the data in the array is stored row major and therefore the same data loaded into a joml matrix has to be transposed. What are we using the collada loader for? How can this be verified?

engine/src/main/java/org/terasology/math/MatrixUtils.java
kBlaszczyk: `calcViewProjectionMatrix()` might still be useful to make these calculations less error prone. But I technically agree.

engine/src/main/java/org/terasology/rendering/cameras/Camera.java
kBlaszczyk: Issue got resolved here: #3905

PR #3902, opened 2020-04-21
partial port of kinematicCharacterMover

pollend:
Here is a partial port of the KinematicCharacterMover. Player character should still move the same way without the changes in this file. I'll probably go through and tweaks some of the uses of Quat4f and replace them with their JOML alternatives. The kinematic character mover is an internal class to the engine so there is close to no impact on module space unless I start updated events that are emitted from the character mover.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterMovementSystemUtility.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterMovementSystemUtility.java
@@ -15,13 +15,12 @@
  */
 package org.terasology.logic.characters;
 
+import org.joml.Quaternionf;
+import org.joml.Vector3f;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.logic.location.LocationComponent;
+import org.terasology.math.JomlUtil;
 import org.terasology.math.TeraMath;
-import org.terasology.math.geom.BaseQuat4f;
-import org.terasology.math.geom.BaseVector3f;
-import org.terasology.math.geom.Quat4f;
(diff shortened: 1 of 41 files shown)


engine/src/main/java/org/terasology/logic/characters/CharacterMovementSystemUtility.java
```diff
@@ -63,27 +62,27 @@ public void setToState(EntityRef entity, CharacterStateEvent state) {
         movementComp.footstepDelta = state.getFootstepDelta();
         entity.saveComponent(movementComp);
 
-        setPhysicsLocation(entity, state.getPosition());
+        setPhysicsLocation(entity, JomlUtil.from(state.getPosition()));
 
         // set the pitch to the character's gaze entity
-        Quat4f rotation = new Quat4f(0f, TeraMath.DEG_TO_RAD * state.getPitch(), 0f);
+        Quaternionf rotation = new Quaternionf().rotationYXZ(0f, TeraMath.DEG_TO_RAD * state.getPitch(), 0f);
```
kBlaszczyk, line 68: Could be simplified to:
```suggestion
        Quaternionf rotation = new Quaternionf().rotationX(TeraMath.DEG_TO_RAD * state.getPitch());
```

pollend:
I tested this with a couple modules, Default Standard, Light and Shadow. the movement seems unaffected so I think this should be ok to merge.

Cervator:
Did some more testing with multiplayer and a physics / movement focus. Looks good 👍

PR #3903, opened 2020-04-22
resolved entityref issue

pollend:
This resolve the issue from: https://github.com/Terasology/JoshariasSurvival/issues/31

I'm not quite sure the exact cause but EnvironmentSwitchHandler adds a new typehandler to the context. I guess there is some hidden thread issue were not aware of but this seems to resolve the problem for the moment. we need a more permanent fix to resolve this at some point in the future. 

## Testing
- create a new default world
- view inventory
- exit world and reload
- inventory should still be there!!


Cervator:
Closing manually, the first commit got pushed so this is resolved!

PR #3905, opened 2020-04-22
chore: Improve camera math usage

kBlaszczyk:
This PR fixes math misusage in the Camera class.

The `viewingDirection` member was set to the xyz-values of a given quaternion in `setOrientation()`.
This value was then used in `getOrientation()` as a viewing axis to define a quaternion from an axis and an angle.
The xyz-values of a quaternion do not represent a viewing axis, nor do they represent a viewing direction.

The value was also used throughout the application as actual viewing direction.
I fixed the calculation of the viewing direction and kept the viewing axis as well.

To verify, check if the camera behaves as expected.


engine/src/main/java/org/terasology/rendering/cameras/Camera.java
```diff
@@ -203,19 +211,18 @@ public Vector3f getViewingDirection() {
 
     /**
      * Get the orientation of the camera.
-     * @return the orientation direction, a quaternion.
+     * @return the orientation
      */
     public Quat4f getOrientation() {
-        return new Quat4f(JomlUtil.from(viewingDirection), viewingAngle);
+        return new Quat4f(from(viewingAxis), viewingAngle);
```
pollend, line 217: would it be more clear to use JomlUtil.from instead of importing the static method.

kBlaszczyk: alright

pollend:
I verified the behavior and this seems to work correctly. if there isn't anything outstanding I think we should be safe to merge this.

PR #3907, opened 2020-04-24
feat: Add events to scale character

skaldarnar:
This PR refactors the adaptions for height changes to be tied to a new events: `ScaleCharacterEvent`.

Two events for slightly different sementics regarding the way an entity should be scaled:

- `ScaleByRequest(factor)` an event to request that an entity should be scaled by the given factor. This is a semantic operation, and as such is consumable (to cancel the request).
- `ScaleToRequest(value)` an event to request that an entity should be scaled to a given value. Again, this is a semantic operation and is consumable.

If the events are not cancelled, they'll eventually reach an event handler with low priority that "accepts" the request and turns it into a notification event for systems to react on:

- `OnScaleEvent(oldValue, newValue)` is an immutable event that informs about a change. Systems should listen to this event in case they want to act upon scaling of characters.

This event is used by the debug command, and it is intended to be used by modules to trigger changes to a player's (or mob's) height. This could be a shrinking potion, or a magic effect. 

---

There are some TODO comments in the code for follow-ups. Further adjustments to the functions for adjusting the values of the character component are also required.
These event also allow to react on scaling a character for more components, e.g., adjust the size of the floating cube in multiplayer, ...


skaldarnar:
Thinking even more about the way the event(s) are supposed be used I would like to propose the following.

Two events for slightly different sementics regarding the way an entity should be scaled:

- `ScaleByEvent(factor)` to request or trigger that an entity should be scaled by the given factor. This is a semantic operation, and as such should be consumable (to cancel the request).
- `ScaleToEvent(value)` to request that an entity should be scaled to a give value. Again, this is a semantic operation and should be consumable.

If the events are not cancelled, they'll eventually reach an event handler with low priority that "accepts" the request and turns it into a _notification event_ for systems to react on:

- `ScaledEvent(oldValue, newValue)` is an immutable event that _informs_ about a change. Systems should listen to this event in case they want to act upon scaling of characters. 

For instance, the `GazeMountPointSystem` will react by recalculating the mount point translation based on the new height. The `CharacterSystem` may adjust values such as speed or jump height by using the scaling factor. 

---

The reason I'm _over-engineering_ this slightly is because I would like understand how we envision the whole ECS to work with the game. I would like to establish best-practice patterns right in the engine and very core modules for reference. 
The suggestion above takes different usages for the scaling event into account and tries to find a unified pattern that works well for all cases...

@jellysnake @Cervator @PS-Soundwave @pollend your input on this is highly appreciated!

pollend:
are there some potential edge cases if we also have a third person camera or do we somehow split these two implementation in the logic. two separate systems with two separate ways to handle the view.

engine/src/main/java/org/terasology/logic/characters/CharacterSystem.java
```diff
@@ -62,7 +64,10 @@
 import org.terasology.world.block.BlockComponent;
 import org.terasology.world.block.regions.ActAsBlockComponent;
 
+import java.util.Optional;
+
 /**
+ *
  */
```
pollend, line 71: are you planning on added some javadocs to this class?

engine/src/main/resources/assets/prefabs/player/player.prefab
```diff
@@ -74,11 +67,5 @@
         "translate": [-0.4, 0.45, 0.3],
         "scale": 0.35
   },
-  "GazeMountPoint": {
```
pollend, line 77: does the mount point update before the game starts?

engine/src/main/resources/assets/prefabs/player/player.prefab
```diff
@@ -20,13 +20,6 @@
   },
   "VisualCharacter": {
   },
-  "BoxShape": {
-    "extents": [
-      1.5,
-      1.8,
-      1.5
-    ]
-  },
```
pollend, line 29: do we still need the BoxShape and does it still serve a purpose? I remember there was some discussion about the there being two colliders on the player.

pollend:
I managed to run this PR and everything seems to be in order. we should be pretty good in regards to merging this.

engine/src/main/java/org/terasology/logic/characters/CharacterSystem.java
skaldarnar: eventually yes - this is just an auto-formatting change. The `CharacterSystem` does a lot of stuff right now, maybe it can be slimmed down to have a better idea how each part is working...

engine/src/main/resources/assets/prefabs/player/player.prefab
skaldarnar: Currently yes, it is updated upon player spawn with the value from the player settings. I want to touch the player settings and see what we can get rid of. 
I'd likely remove these settings, but keep the scale event upon player spawn to make sure the gaze mount point is correctly computed. I'd like to avoid managing default values in several places, thus, I'd like to get it out of the prefab...

engine/src/main/resources/assets/prefabs/player/player.prefab
skaldarnar: yeah, we discussed this on Discord - nobody seems to know what this is used for, and the numbers are off from the capsule shape used otherwise for the player. 
People suggested to remove it now - we can easily bring it back if we encounter some bugs (hard part will be to connect these potential bugs with the lack of a `BoxShapeComponent` on the player, though)

Cervator:
@skaldarnar - bit of a belated response here, but I wanted to add some more support for the idea of using events this way, marking it as a best practice 👍 Very much reminds me of what we did with the Health system, where healing effects can arise from multiple sources, get modified / cancelled, then eventually trigger an actual heal.

PR #3908, opened 2020-04-24
JOML migration for AudioManager

kBlaszczyk:
This PR migrates the AudioManager and related classes to JOML.

Every method I converted without keeping the original was unused throughout the engine and all major modules.
(the only module I couldn't check was FlexibleMovement, which seems to be broken for unrelated reasons atm)

To verify, run the game and check if the sound is working as usual.

### DIFF
--- a/engine/src/main/java/org/terasology/audio/AudioManager.java
+++ b/engine/src/main/java/org/terasology/audio/AudioManager.java
@@ -15,7 +15,7 @@
  */
 package org.terasology.audio;
 
-import org.joml.Quaternionf;
+import org.joml.Quaternionfc;
 import org.joml.Vector3fc;
 import org.terasology.assets.AssetFactory;
 import org.terasology.math.geom.Quat4f;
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/terasology/audio/AudioManager.java
```diff
@@ -123,8 +133,12 @@
      * @param position The new position
      * @param orientation The new orientation (in a quaternion)
      * @param velocity The new velocity
+     * @deprecated This method is scheduled for removal in an upcoming version.
+     *             Use the JOML implementation instead: {@link #updateListener(Vector3fc, Quaternionf, Vector3fc)}.
      */
+    @Deprecated
     void updateListener(Vector3f position, Quat4f orientation, Vector3f velocity);
+    void updateListener(Vector3fc position, Quaternionf orientation, Vector3fc velocity);
```
pollend, line 141: can we change it to use Quaternionfc. I don't think we modify the orientation in anyway so the interface would make more sense here.

kBlaszczyk: yes, of course.

PR #3910, opened 2020-04-25
feat: add retain components component

jdrueckert:
### Contains

This PR adds `RetainComponentsComponent` which is a component intended to list component classes that are supposed to be retained when placing a block.

Specifying a list of components to be retained when placing a block is necessary as on placing a block, `EntityAwareWorldProvider::updateBlockEntityComponents` removes all components from the entity of the block to be placed that
- are not any of `NetworkComponent`, `BlockComponent`, and `LocationComponent`
- don't have `retainUnalteredOnBlockChange` metadata
- are not listed in the block prefab
- are not in the set of components to be retained
This especially affects components that are added to an entity programmatically, e.g. the `ManualLabor:AssemblyTable` which receives amongst others an `InventoryComponent` in `Machines:MachineCommonSystem` leading to newly placed assembly tables being broken.

The set of component given to `updateBlockEntityComponents` was hard-coded to `emptySet` in `EntityAwareWorldProvider::setBlocks` until now. With this PR there's now a check for `RetainComponentsComponent` in the `blockEntity` in `EntityAwareWorldProvider::setBlocks`.
If it's present, instead of `emptySet`, the set of components to retain that is stored in the `RetainComponentsComponent` component are passed.

### How to test

Broken state:
Programmatically add a component to an entity and observe that it's removed in `EntityAwareWorldProvider::updateBlockEntityComponents`.

Fixed state:
Before adding the new component, create a `RetainComponentsComponent` with this new component and add it to the entity. Observe that it's not removed anymore. Also observe, that the `RetainComponentsComponent` is removed.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/common/RetainComponentsComponent.java
+++ b/engine/src/main/java/org/terasology/logic/common/RetainComponentsComponent.java
@@ -1,11 +1,31 @@
+/*
+ * Copyright 2020 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.terasology.logic.common;
 
 import com.google.api.client.util.Sets;
 import org.terasology.entitySystem.Component;
 
-import java.util.Collections;
 import java.util.Set;
 
+/**
+ * This component is intended to list component classes that are supposed to be retained when placing a block.
+ *
+ * If a block entity has a component that is not part of its prefab, retaining this component results in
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/terasology/logic/common/RetainComponentsComponent.java
```diff
@@ -0,0 +1,11 @@
+package org.terasology.logic.common;
```
skaldarnar, line 1: We're missing a license header here.

engine/src/main/java/org/terasology/logic/common/RetainComponentsComponent.java
```diff
@@ -0,0 +1,11 @@
...
+
+import com.google.api.client.util.Sets;
+import org.terasology.entitySystem.Component;
+
+import java.util.Collections;
+import java.util.Set;
+
+public class RetainComponentsComponent implements Component {
```
skaldarnar, line 9: Please add Javadoc to explain what this component is intended for. I think you can take parts the PR description for that.

engine/src/main/java/org/terasology/logic/common/RetainComponentsComponent.java
```diff
@@ -0,0 +1,11 @@
...
+    public Set<Class<? extends Component>> components = Sets.newHashSet();
```
skaldarnar, line 10: We should make sure that this works with de-/serialization, e.g., check whether we can specify this also via a prefab...

Issue #3889, opened 2020-04-14
Water under the bridge... sort of.


misterHippo:
As per the note on Discord, water blocks that are surrounded by other blocks don't look right.

This is obviously because the rendering isn't done when a block is fully enclosed. However, due to the 'height' of the water block, that idea doesn't float.

HAHAHAHA see what I did there? :D

*ahem*

See this pic in case you've been hiding under a block.

![image](https://user-images.githubusercontent.com/605912/79177259-bf90c300-7e45-11ea-9398-01549ccac39b.png)

skaldarnar:
Can confirm this is an issue with the `engine:loweredCube` shape with occluding blocks. The same effect happens when placing `CoreBlocks:Dirt:engine:loweredCube` the same way (`give dirt 16 loweredcube`). 
It does not happen for `engine:halfBlock` though, not sure what the difference is. Should be fairly simple to debug and fix.

Cervator:
For a bit more historical background: water is a bit of a special case as it is meant to not appear flush with the block line at the _surface_ of the body of water - that's why it uses the lowered cube shape. However it'd look silly if every layer of water had a little gap at the top, so there's some special handling for that somewhere. That might be what causes some quirks around other related cases.

cvennel:
Hi there. Is this as simple as adding a check in BlockMeshGeneratorSingleShape?

In the isSideVisibleForBlockTypes method I tried adding a conditional of if the block above is not water return true for this block. I can open a PR if that sounds fine. I was trying to follow the new contributor guides but wasn't sure where to ask questions first.

Before:
![Before](https://user-images.githubusercontent.com/35314989/80318138-f1a30d00-87d5-11ea-987d-713f8a46bbd6.png)

and After
![after](https://user-images.githubusercontent.com/35314989/80318142-f4056700-87d5-11ea-88b2-e3ea8edeeb72.png)

Cervator:
Hi @cvennel ! This is fine although our Discord is sometimes better as it is easy to miss all the GitHub notifications.

If that fixes the problem then that's probably fine for a PR already :) Maybe we need to improve the setup further (for instance for things other than water), but if this fixes the initial bug then we can submit an issue for further refactoring later. That might be a much bigger task.

cvennel:
[...]

PR #3911, opened 2020-04-26
fix: Add logic check to allow visible water under other blocks

cvennel:
### Contains

Fixes bug #3889

### How to test

Verify water surface is visible when under other block as in below photos:

Before:
Before

After
after

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/primitives/BlockMeshGeneratorSingleShape.java
+++ b/engine/src/main/java/org/terasology/rendering/primitives/BlockMeshGeneratorSingleShape.java
@@ -142,6 +142,10 @@ public class BlockMeshGeneratorSingleShape implements BlockMeshGenerator {
             return false;
         }
 
+        if (currentBlock.isWater() && (side == Side.TOP) && !blockToCheck.isWater()){
+            return true;
+        }
+
         return currentBlock.isWaving() != blockToCheck.isWaving() || blockToCheck.getMeshGenerator() == null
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/rendering/primitives/BlockMeshGeneratorSingleShape.java
```diff
@@ -142,6 +142,10 @@ private boolean isSideVisibleForBlockTypes(Block blockToCheck, Block currentBloc
             return false;
         }
 
+        if (currentBlock.isWater() && (side == Side.TOP) && !blockToCheck.isWater()){
```
skaldarnar, line 145: Please add a `//TODO: ...` comment here to indicate that this only fixes part of a broader problem:

```suggestion
        //TODO: This only fixes the "water under block" issue of the top side not being rendered.
        //      Find a more general way to handle this for "engine:loweredCube" shapes.
        if (currentBlock.isWater() && (side == Side.TOP) && !blockToCheck.isWater()){
```

cvennel: Thanks Skaldarnar, good point; I've added the TODO. I also put a note about IntelliJ warning `!blockToCheck.isLiquid()` always returns true

pollend:
@Cervator @skaldarnar if there isn't anything else we should merge this.

Cervator:
LGTM and I wouldn't mind merging, just am already up too late for the day :-)

Would be nice to fix further but that's probably beyond this PR. The TODO may also be enough for the reminder rather than making a full separate issue.

pollend:
The existing issue will still be open so it should be fine #3889

PR #3912, opened 2020-04-27
feat(JOML): Migrate PlayerFactory

pollend:
I ported over character spawning logic to JOML. I moved `SpiralIterable` to math and I also added added and deprecated `attachChild` in Location. The changes should be functionally the same from the old code. Do we want to move `SpiralIterable`to a different namespace is the place I'm using ok? 

## How to Test
- create a new world
- player should spawn without issue

### DIFF
--- a/engine/src/main/java/org/terasology/math/SpiralIterable.java
+++ b/engine/src/main/java/org/terasology/math/SpiralIterable.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright 2020 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.math;
+
+import com.google.common.base.Preconditions;
+import org.joml.Vector2i;
+import org.joml.Vector2ic;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+/**
+ * An {@link Iterable} that iterates in
+ * a square-shapes spiral around the central point (inclusive).
+ * <br><br>
+ * The iteration starts in positive x direction.
+ * <br><br>
+ * The iterating vector is reused. <b>Do not attempt to store the instance</b> e.g. in a collection.
+ * @author Martin Steiger
+ */
+public class SpiralIterable implements Iterable<Vector2ic>{
+
+    /**
+     * (MAX_SIDELEN * 2 + 1) ^2 must be < Integer.MAX_VALUE
+     */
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/terasology/math/SpiralIterable.java
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * An {@link Iterable} that iterates in
+ * a square-shapes spiral around the central point (inclusive).
+ * <br><br>
+ * The iteration starts in positive x direction.
+ * <br><br>
+ * The iterating vector is reused. <b>Do not attempt to store the instance</b> e.g. in a collection.
+ * @author Martin Steiger
+ */
+public class SpiralIterable implements Iterable<Vector2ic>{
```
skaldarnar, line 33: I think the documentation on this can be improved, but that's unrelated to this PR.

engine/src/main/java/org/terasology/math/SpiralIterable.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.NoSuchElementException;
+
+/**
+ * An {@link Iterable} that iterates in
+ * a square-shapes spiral around the central point (inclusive).
+ * <br><br>
+ * The iteration starts in positive x direction.
+ * <br><br>
+ * The iterating vector is reused. <b>Do not attempt to store the instance</b> e.g. in a collection.
+ * @author Martin Steiger
```
skaldarnar, line 31: remove the author tag, we don't use them (anymore)
```suggestion
```

engine/src/main/java/org/terasology/math/SpiralIterable.java
```diff
@@ -0,0 +1,183 @@
...
+    public Iterator<Vector2ic> iterator() {
+
+        return new Iterator<Vector2ic>() {
+            private int radius = 1;
+            private int leg;
+            private int x = -1;
+            private int y;
+            private int index;
+
+            private Vector2i pos = new Vector2i();
```
kBlaszczyk, line 92: I'm not sure about this one...
The client iterates over read only vectors. So he probably doesn't expect their value to change. Each `next()` call however, changes the value of the previously returned vector, since it's always the same instance being returned.
Therefore I think it would be cleaner to return new instance every time. (see next comment)

engine/src/main/java/org/terasology/math/SpiralIterable.java
```diff
@@ -0,0 +1,183 @@
...
+                            ++radius;
+                        }
+                        break;
+                }
+
+                index++;
+                int finalX = center.x() + x * scale;
+                int finalY = center.y() + (clockwise ? y : -y) * scale;
+                pos.set(finalX, finalY);
+                return pos;
```
kBlaszczyk, line 132: `return new Vector2i(finalX, finalY)`

engine/src/main/java/org/terasology/math/SpiralIterable.java
pollend: I think the point is to reduce GC by avoiding creating a new instance. @skaldarnar how much does this matter with regards to performance and would it be better to make a new instance?

kBlaszczyk: I just realized, that it even points this out in the javadoc: `The iterating vector is reused.`
So it was definitely done consciously. The GC argument is valid (even though we are having tons of unnecessary vector allocations at runtime anyway), so I guess I can live with that.

PR #3913, opened 2020-04-28
fix: Add type handler for `Class<? extends Component>` 

skaldarnar:
Follow up to https://github.com/MovingBlocks/Terasology/pull/3910

Potential fix for this oversight: https://github.com/Terasology/Machines/pull/29#issuecomment-620378868

I tried to write a `TypeHandler` for `Class<? extends Component>` (see https://github.com/Terasology/Machines/pull/29#issuecomment-620846792) but failed adding it to the type handler library.

Not sure what the best/easiest/recommended way to go forward is ... 🙄 

@Cervator @jdrueckert would be great if you could test this ... The set should be filled like this now:

```java
retainComponents.componentTypeNames.add(InventoryAccessComponent.class.getName());
```


skaldarnar:
Unfortunately, this does not work as the specific component classes cannot be loaded from the engine (`ClassNotFoundException`)... 😞

PR #3917, opened 2020-04-30
feat(JOML): Migrate Particle System

pollend:
A lot of the logic was pretty straightforward so moving the particle system over to JOML was not too bad. There are not a lot of dependencies from this part of the codebase so moving over the rest of the components is fairly trivial when compared to other parts of the engine. This mostly focuses on porting over particles in the engine. 

Modules:
- https://github.com/Terasology/Health/pull/40

## Test
- download health module: ` ./groovyw module recurse Health` and use branch from https://github.com/Terasology/Health/pull/40
- create core world and verify that particles behave correctly when breaking blocks.

### DIFF
--- a/engine/src/main/java/org/terasology/particles/functions/generators/TextureOffsetGeneratorFunction.java
+++ b/engine/src/main/java/org/terasology/particles/functions/generators/TextureOffsetGeneratorFunction.java
@@ -38,6 +38,6 @@ public class TextureOffsetGeneratorFunction extends GeneratorFunction<TextureOff
 
         final int randomOffsetIndex = random.nextInt(component.validOffsets.size());
         final Vector2f randomOffset = component.validOffsets.get(randomOffsetIndex);
-        particleData.textureOffset.set(randomOffset.x(), randomOffset.y());
+        particleData.textureOffset.set(randomOffset);
     }
 }
--- a/engine/src/main/java/org/terasology/particles/functions/generators/TextureOffsetGeneratorFunction.java
+++ b/engine/src/main/java/org/terasology/particles/functions/generators/TextureOffsetGeneratorFunction.java
@@ -15,7 +15,7 @@
  */
 package org.terasology.particles.functions.generators;
 
-import org.terasology.math.geom.Vector2f;
+import org.joml.Vector2f;
 import org.terasology.particles.ParticleData;
 import org.terasology.particles.ParticleDataMask;
 import org.terasology.particles.components.generators.TextureOffsetGeneratorComponent;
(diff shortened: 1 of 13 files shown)


pollend:
we should probably address this PR: https://github.com/MovingBlocks/Terasology/pull/3230

engine/src/main/java/org/terasology/particles/functions/generators/TextureOffsetGeneratorFunction.java
```diff
@@ -35,9 +35,9 @@ public void onEmission(TextureOffsetGeneratorComponent component, ParticleData p
         if (component.validOffsets.size() == 0) {
             return;
         }
-        
+
         final int randomOffsetIndex = random.nextInt(component.validOffsets.size());
         final Vector2f randomOffset = component.validOffsets.get(randomOffsetIndex);
-        particleData.textureOffset.set(randomOffset.getX(), randomOffset.getY());
+        particleData.textureOffset.set(randomOffset.x(), randomOffset.y());
```
kBlaszczyk, line 41: Could be simplified:
```suggestion
        particleData.textureOffset.set(randomOffset);
```

engine/src/main/java/org/terasology/particles/rendering/SpriteParticleRenderer.java
```diff
@@ -99,7 +99,7 @@ public void drawParticles(Material material, ParticleRenderingData<ParticleDataS
             GL13.glActiveTexture(GL13.GL_TEXTURE0);
             glBindTexture(GL11.GL_TEXTURE_2D, particleSystem.particleData.texture.getId());
 
-            material.setFloat2("texSize", particleSystem.particleData.textureSize.getX(), particleSystem.particleData.textureSize.getY());
+            material.setFloat2("texSize", particleSystem.particleData.textureSize.x(), particleSystem.particleData.textureSize.y());
```
kBlaszczyk, line 102: Could be simplified:
```suggestion
            material.setFloat2("texSize", particleSystem.particleData.textureSize);
```

Cervator:
Okay this took longer than the few minutes I had hoped an already-approved PR with some addressed review comments would take, looking for an easy PR to merge real quick ... 😁 

First had to resolve a conflict on here, in one or two modules as well, deal with the weirdness that the WeatherManager PR seems obsolete (guess the module changed since and no longer needed the compat fix). Then turns out some other modules broke as well, found and fixed those, some more broke from unrelated issues, and even though GooeyDefence does work it crashes on shutdown again for something unrelated. Fun! But as far as I can test this PR and the associated module ones test out and things are getting better - thank you! 👍

PR #3919, opened 2020-04-30
feat(JOML): Migrate BlockMeshLoader

pollend:
I ported over the JsonBlockShapeLoader to json. The next part is trying to port overt the Tesselator. 

## Test
this logic is used to load in blocks so just loading in a world should be enough to verify the functionality.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/internal/BlockBuilder.java
+++ b/engine/src/main/java/org/terasology/world/block/internal/BlockBuilder.java
@@ -17,6 +17,7 @@ package org.terasology.world.block.internal;
 
 import com.google.common.base.Strings;
 import com.google.common.collect.Maps;
+import org.terasology.math.JomlUtil;
 import org.terasology.utilities.Assets;
 import org.terasology.math.Rotation;
 import org.terasology.math.Side;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/terasology/world/block/internal/BlockBuilder.java
```diff
@@ -193,7 +194,7 @@ private BlockAppearance createAppearance(BlockShape shape, Map<BlockPart, BlockT
             BlockPart targetPart = part.rotate(rot);
             textureAtlasPositions.put(targetPart, atlasPos);
             if (shape.getMeshPart(part) != null) {
-                meshParts.put(targetPart, shape.getMeshPart(part).rotate(rot.getQuat4f()).mapTexCoords(atlasPos, worldAtlas.getRelativeTileSize(), frameCount));
+                meshParts.put(targetPart, shape.getMeshPart(part).rotate(JomlUtil.from(rot.getQuat4f())).mapTexCoords(JomlUtil.from(atlasPos), worldAtlas.getRelativeTileSize(), frameCount));
```
kBlaszczyk, line 197: Some line breaks would be nice.
This applies to multiple changes made in this PR, do we have a convention about how many characters per line are acceptable?

Cervator: That's a little wide, yeah. The project standard (defined in Checkstyle config) is 175, which is pretty wide, so apply best judgement even if approaching that. This exceeds it by a bit.

PR #3922, opened 2020-04-30
chore: Extract CoreItems from Core

skaldarnar:
This is an attempt to extract more things from the `Core` module to eventually get rid of it. The assets and source code related to items can now be found here:
> https://github.com/Terasology/CoreItems

Requires https://github.com/Terasology/CoreAssets/pull/4 for the `CoreItems` module to work correctly.


modules/CoreSampleGameplay/src/main/java/org/terasology/coresamplegameplay/PlayerStartingInventorySystem.java
```diff
@@ -41,52 +41,72 @@
     @ReceiveEvent(components = InventoryComponent.class)
     public void onPlayerSpawnedEvent(OnPlayerSpawnedEvent event, EntityRef player) {
         BlockItemFactory blockFactory = new BlockItemFactory(entityManager);
-        // Goodie chest
+
+        EntityRef bonusChest = getBonusChest(blockFactory);
```
Cervator, line 45: Should we consider replacing this whole thing with the new capability from https://github.com/Terasology/Inventory/pull/3 ? Or I guess maybe it can't do filled chests? 🤔

Cervator:
On second look, wow, WoodAndStone uses the crossbow? And actually has a recipe for it? I had no idea 🙈 💥 

Pity that's probably too broken to really get to 😞

modules/CoreSampleGameplay/module.txt
```diff
@@ -4,7 +4,8 @@
     "displayName" : "Core Gameplay",
     "description" : "Minimal gameplay template. Little content but a few starting items.",
     "dependencies" : [
-        {"id": "Core", "minVersion": "3.0.0"}
+        {"id": "Core", "minVersion": "3.0.0"},
+        {"id": "CoreItems", "minVersion": "1.0.0"}
```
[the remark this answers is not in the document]
skaldarnar: @jdrueckert started to go through (Omega) module space to look for usages and directly depend on `CoreItems`. I think in some cases this even lead to removing the dependency to `Core` 🎉

modules/CoreSampleGameplay/src/main/java/org/terasology/coresamplegameplay/PlayerStartingInventorySystem.java
skaldarnar: Right, the filled chests are currently missing... that's why I gave this a quick improvement before giving the `StartingInventoryComponent` the finishing touches 😉

skaldarnar:
Good remarks, pretty much sums up the next items on my todo list ^^ Would like to go step-by-step to not be confused by what went where....

@jdrueckert and I discussed whether we want to have a basic sword and crossbow in the core items, but figured out that we eventually want to _get rid of `CoreItems`_ - move some stuff to `CombatSystem` (or CombatAssets.... 🤔 ) others to Tools, others to Furniture, ... you get it 😉 

I'll clean up CoreItem a bit more (remove coal, I think it's not dropped from coal blocks) and figure something out for the crossbow. 

The "damage stuff" has a lots of TODO items to move it to Health, actually.... also on my list, feel free to beat me to it. 

The magic axe is not totally gone - it lives forth in `gooeysFist` and is now even stronger 💪

PR #3923, opened 2020-05-01
feat(JOML): Migrate BlockEntityRegistry

pollend:
updated BlockEntityRegistry with joml methods. the methods themselves back up their joml implementations, same as the worldprovider PR.


Cervator:
Looked easy enough, poked around in a mega-workspace and couldn't spot anything too odd. Merged! Thanks 👍

PR #3930, opened 2020-05-02
Seed ui and save issues

### LINKED ISSUE #3883: World seed value seemingly gets ignored if you use "Start Playing" from Advanced Game Setup

Start the game with the exact same seed to get the same world - got lazy and tried to start playing straight off the Advanced Game Setup (AGS) screen after having entered the seed value

### What actually happened

Got a different world than expected. Found that if I used "Continue to Universe Setup" and manually added my Perlin world then it worked fine. If you use the button on the AGS page it does seem to pick the default world for the gameplay template, but it skips the seed value somehow.

### How to reproduce

* Create a new game, enter a seed value on the AGS page or just copy the default to keep handy
* Continue through remaining steps adding a Perlin world manually and entering the world
* Observe the splendor of your chosen world
* Restart the game client (possible session bleed if you just go back to the main menu, avoid to be sure)
* Enter the same exact seed value on the AGS page, but this time then press "Start Playing"
* Observe a completely different world

I did this with GooeysQuests, but expect it'll affect any set of modules

### Troubleshooting tip

Start by examining what actions happen based on the buttons clicked (search for the button captions if needed to find the associated files). How does the flow vary if you start the game from the AGS page rather than go all the way through the create game flow?

### LINKED ISSUE #3915: Seed value does not appear on Game Details Screen
See the seed of a saved world through the Game Details Screen.
### What actually happened
The spot where the seed should be written was blank.
### How to reproduce

* Go to Singleplayer
* Select one of the saved games _(if there aren't any, create a new one)_
* Click the details button to go at the game's Game details window
* Notice that the seed value does not appear

I only tested the issue through Single Player and with Core Gameplay template saves, but I believe the issue occurs under any circumstances.

Game version: **78 - alpha**

### Related screenshots
issue_screenshot_seed_not_appearing

### LINKED ISSUE #3929: Seed value equals the default-blank value at the save files, if created through the Advanced Game Setup
Save the seed value of a new world while creating that world through the Advanced Game Setup settings.
### What actually happened
When opening the manifest file of the game save file, the seed value equals the default value, which is: "".
### How to reproduce

* Setup the conditions for the save file:
    * Create a new world
    * Go to the advanced settings
    * Start Playing
    * Exit Terasology
* See the bug at that save file:
    * Open the manifest file, which is located at **Terasology>saves>_YourGameFileName_>manifest**
    * Notice seed value being equal to "", instead of the given one. _(see screenshot below)_
seed is not saved

stefaniamak:
### Result

Seed value gets saved at the save file of the created world.

### Original Issue 

* Seed value was not getting saved at the world's save file, thus:
    * Correct seed value did not appear on Game Details Screen (it was never saved in the first place)

The problem seemed to be that the `gameManifest` never gets the seed value.

Closes #3915 
Closes #3929 
Possibly also Closes #3883 

### Changes

* Added the seed value `seed` at the `gameManifest`.

## How to test

1. Create different games through the Advanced Game Settings by either adding your seed value, **or** by keeping the generated one
2. Follow the How to reproduce issue #3915 **or** How to reproduce issue #3929

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/GameManifestProvider.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/GameManifestProvider.java
@@ -79,11 +79,12 @@ public class GameManifestProvider {
         if (universeWrapper.getTargetWorld() != null) {
             uri = universeWrapper.getTargetWorld().getWorldGenerator().getUri();
             seed = universeWrapper.getTargetWorld().getWorldGenerator().getWorldSeed();
-            gameManifest.setSeed(seed);
         } else {
             uri = config.getWorldGeneration().getDefaultGenerator();
             seed = universeWrapper.getSeed();
         }
+        gameManifest.setSeed(seed);
+
         String targetWorldName = "";
         if (universeWrapper.getTargetWorld() != null) {
             targetWorldName = universeWrapper.getTargetWorld().getWorldName().toString();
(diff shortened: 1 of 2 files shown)


stefaniamak:
Issue #3931 still remains.

PR #3935, opened 2020-05-05
feat(JOML): migrated Diamond3iIterator ->  Diamond3iIterable

pollend:
I moved Diamond3iIterable/SpiralIterable into its own namespace and marked AABB and Diamon3Iterator as deprecated. I'm not sure how to handle Region3i since it's used heavily throughout the code base.  probably a better way to handle the changes proposed in #3934.


engine/src/main/java/org/terasology/math/Diamond3iIterable.java
```diff
@@ -0,0 +1,162 @@
...
+                    } else {
+                        offset[2] = -(level[0] - TeraMath.fastAbs(offset[0]) - TeraMath.fastAbs(offset[1]));
+                    }
+                }
+                return pos;
+            }
+        };
+    }
+
+    public static final class Builder {
```
kBlaszczyk, line 132: I'd prefer the inner class on top, but I'm not sure if we have style guidelines for that.

engine/src/main/java/org/terasology/math/Diamond3iIterable.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    public static Diamond3iIterable.Builder shell(Vector3ic origin, int radius) {
+        return new Diamond3iIterable.Builder(origin, radius).start(radius - 1);
+    }
+
+    @Override
+    public Iterator<Vector3ic> iterator() {
+        Vector3i pos = new Vector3i();
+        final int[] level = {this.startDistance};
+        final int[] offset = {-this.startDistance,0,0};
```
kBlaszczyk, line 95: The value access would be clearer using a Vector3i here.

engine/src/main/java/org/terasology/math/Diamond3iIterable.java
```diff
@@ -0,0 +1,162 @@
...
+
+        return new Iterator<Vector3ic>() {
+            @Override
+            public boolean hasNext() {
+                return level[0] < endDistance;
+            }
+
+            @Override
+            public Vector3ic next() {
+                pos.set(origin.x + offset[0], origin.y + offset[1], origin.z + offset[2]);
```
kBlaszczyk, line 105: If a `Vector3i` was used no represent the offset, the statement could be simplified to:
`origin.add(offset, pos);`

engine/src/main/java/org/terasology/math/Diamond3iIterable.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     *
+     * @param origin center point of Diamond3iIterator
+     * @param startDistance inner distance outwards
+     * @param maxDistance maxDistance away
+     */
+    private Diamond3iIterable(Vector3ic origin, int startDistance, int maxDistance) {
+        Preconditions.checkArgument(startDistance < maxDistance, "startDistance must be < maxDistance");
```
kBlaszczyk, line 54: +/- 1 confusion:
```suggestion
        Preconditions.checkArgument(startDistance <= maxDistance, "startDistance must be <= maxDistance");
```

engine/src/main/java/org/terasology/math/Diamond3iIterable.java
```diff
@@ -0,0 +1,162 @@
...
+        return new Diamond3iIterable.Builder(origin, end).start(start);
+    }
+
+    /**
+     * iterates in a shell that is 1 block wide
+     * @param origin center of shell
+     * @param radius distance for 1 block wide shell
+     */
+    public static Diamond3iIterable.Builder shell(Vector3ic origin, int radius) {
+        return new Diamond3iIterable.Builder(origin, radius).start(radius - 1);
```
kBlaszczyk, line 88: +/- 1 confusion:
```suggestion
        return new Diamond3iIterable.Builder(origin, radius).start(radius);
```

engine/src/main/java/org/terasology/math/Diamond3iIterable.java
```diff
@@ -0,0 +1,162 @@
...
+    @Override
+    public Iterator<Vector3ic> iterator() {
+        Vector3i pos = new Vector3i();
+        final int[] level = {this.startDistance};
+        final int[] offset = {-this.startDistance,0,0};
+
+        return new Iterator<Vector3ic>() {
+            @Override
+            public boolean hasNext() {
+                return level[0] < endDistance;
```
kBlaszczyk, line 100: +/- 1 confusion:
```suggestion
                return level[0] <= endDistance;
```

engine/src/main/java/org/terasology/math/Diamond3iIterable.java
```diff
@@ -0,0 +1,162 @@
...
+        private int startDistance = 0;
+
+        /**
+         * Default endDistance starts at 1
+         * @param origin center region for iterator
+         * @param endDistance maximums radius
+         */
+        private Builder(Vector3ic origin, int endDistance){
+            this.origin = origin;
+            this.endDistance = endDistance + 1;
```
kBlaszczyk, line 144: +/- 1 confusion:
```suggestion
            this.endDistance = endDistance;
```

engine/src/main/java/org/terasology/math/Diamond3iIterable.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        /**
+         * Default start distance is 0
+         * @param start  the minimum radius
+         */
+        public Diamond3iIterable.Builder start(int start) {
+            this.startDistance = start + 1;
```
kBlaszczyk, line 152: +/- 1 confusion:
```suggestion
            this.startDistance = start;
```
that was the last one.

engine-tests/src/test/java/org/terasology/math/Diamond3iIteratorTest.java
```diff
@@ -35,13 +35,13 @@
...
     public void testZeroDistanceIteration() {
-        Iterator<Vector3ic> iter = Diamond3iIterable.region(new org.joml.Vector3i(),0).build().iterator();
+        Iterator<Vector3ic> iter = Diamond3iIterable.region(new org.joml.Vector3i(),1).build().iterator();
         assertEquals(Lists.newArrayList(new Vector3i()), Lists.newArrayList(iter));
     }
 
     @Test
     public void testOneDistanceIteration() {
-        Iterator<Vector3ic> iter = Diamond3iIterable.region(new Vector3i(),1).build().iterator();
+        Iterator<Vector3ic> iter = Diamond3iIterable.region(new Vector3i(),2).build().iterator();
```
kBlaszczyk, line 44: No! that's not a good idea.
The changes I suggested kept the unit tests unchanged and intact. You changed the API and now every usage of that method would need to be adjusted.
Besides, the API was fine - If I create a diamond with radius 1, I expect the coordinates to be (-1, 0, 0), (1, 0, 0), ...
This behavior needs to be kept, my changes only optimized the internals.

engine/src/main/java/org/terasology/math/Diamond3iIterable.java
```diff
@@ -102,26 +102,26 @@ public boolean hasNext() {
 
             @Override
             public Vector3ic next() {
-                pos.set(origin.x + offset[0], origin.y + offset[1], origin.z + offset[2]);
-                if (offset[2] < 0) {
-                    offset[2] *= -1;
-                } else if (offset[1] < 0) {
-                    offset[1] *= -1;
-                    offset[2] = -(level[0] - TeraMath.fastAbs(offset[0]) - TeraMath.fastAbs(offset[1]));
+                pos.set(origin.x + offset.x, origin.y + offset.y, origin.z + offset.z);
```
kBlaszczyk, line 105: As suggested before, why not use vector math?
```suggestion
                origin.add(offset, pos);
```
alternatively: `pos.set(origin).add(offset)`

PR #3936, opened 2020-05-05
feat(JOML): Migrate Region3i

pollend:
### Contains

This is a migration of `Region3i` to JOML. the process it a bit more complication and Its kind of a combination of Deprecating and replacing the implementation in the engine. If we add an AABBi type object to joml then will probably just deprecate `Region3i`. https://github.com/JOML-CI/JOML/issues/234 

- I use min and max to define Region3i vs min and size. 
- max is subtracted by 1 so I kept that the same so the class would be compatible but going forward `getMin(int component)` and `getMax(int component)`
- minX, minY, minZ, maxZ, maxY, maxZ public. 
- `createFromCenterExtents` can be accomplished by `new Region3i(0,0,0,0,0,0).inflate(3,3,3)`
- iterator and subtract reuses the same vector

### Modules

- https://github.com/Terasology/Health/pull/41

### DIFF
--- a/engine/src/main/java/org/terasology/math/Region3i.java
+++ b/engine/src/main/java/org/terasology/math/Region3i.java
@@ -20,8 +20,6 @@ import org.joml.Math;
 import org.joml.Vector3fc;
 import org.joml.Vector3i;
 import org.joml.Vector3ic;
-import org.terasology.audio.AudioEndListener;
-import org.terasology.audio.StaticSound;
 import org.terasology.math.geom.BaseVector3f;
 import org.terasology.math.geom.BaseVector3i;
 import org.terasology.math.geom.Vector3f;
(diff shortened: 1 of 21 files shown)


engine/src/main/java/org/terasology/math/Region3i.java
```diff
@@ -73,253 +223,676 @@ public static Region3i createFromMinAndSize(BaseVector3i min, BaseVector3i size)
...
         return result;
     }
 
+    /**
+     * get center of {@link Region3i} and write to <code>dest</code>
+     *
+     * @param dest will hold the result
+     * @return dest
+     */
+    public Vector3f getCenter(Vector3f dest) {
```
kBlaszczyk, line 712: You probably intended to use the JOML-type here. You are still importing TeraMath.Vector3f and are therefore using it in this method.

pollend: looks like I overlooked the import.

pollend:
I think I'll close this in favor of waiting for this: https://github.com/JOML-CI/JOML/issues/234

Issue #3742, opened 2019-09-08
Game does not launch


RastaManKing:
Launch the game

Game splash screen flashed then game crashed

Step 1 - Install game (im using Arch Linux)
Step 2 - Launch game
Step 3 - Watch it not start

[rastamanking@GreenGoblin ~]$ terasology
PathManager being initialized. Initial code location is /usr/share/terasology/libs/engine-2.2.0.jar
Switched it to expected working dir: /usr/share/terasology
installPath was null, running from IDE or headless server? Setting to: /usr/share/terasology
Found the natives dir: /usr/share/terasology/natives
21:05:45.303 [main] INFO  o.terasology.engine.TerasologyEngine - Initializing Terasology...
21:05:45.316 [main] INFO  o.terasology.engine.TerasologyEngine - [buildNumber=78, buildId=78, buildTag=jenkins-TerasologyStable-78, buildUrl=http://jenkins.terasology.org/job/TerasologyStable/78/, gitBranch=master, gitCommit=a05742f087c3b9a680284ccc1b91c8b1e38c3dc6, dateTime=2019-02-03T04:57:37Z, displayVersion=alpha, engineVersion=2.2.0]
21:05:45.316 [main] INFO  o.terasology.engine.TerasologyEngine - Home path: /home/rastamanking/.local/share/terasology
21:05:45.317 [main] INFO  o.terasology.engine.TerasologyEngine - Install path: /usr/share/terasology
21:05:45.317 [main] INFO  o.terasology.engine.TerasologyEngine - Java: 12.0.2 in /usr/lib/jvm/java-12-jdk
21:05:45.317 [main] INFO  o.terasology.engine.TerasologyEngine - Java VM: Java HotSpot(TM) 64-Bit Server VM, version: 12.0.2+10
21:05:45.317 [main] INFO  o.terasology.engine.TerasologyEngine - OS: Linux, arch: amd64, version: 5.2.11-arch1-1-ARCH
21:05:45.317 [main] INFO  o.terasology.engine.TerasologyEngine - Max. Memory: 1536 MiB
21:05:45.317 [main] INFO  o.terasology.engine.TerasologyEngine - Processors: 8
[...]

PR #3937, opened 2020-05-07
Change how LWJGL path is set. Improves compat with newer JRE

keturn:
We had our own NativeHelper code to try to manipulate private variables on the classloader. It didn't work very well across JRE versions, as seen in #3742. Fortunately, [LWJGL has its own mechanism for dealing with library paths](http://wiki.lwjgl.org/wiki/LWJGL_Hidden_Switches.html#Hidden_Switches).

In specific, this PR addresses the error that appeared in logs as:

```
ERROR o.terasology.utilities.NativeHelper - Couldn't link static libraries.
java.lang.NoSuchFieldException: usr_paths
at java.base/java.lang.Class.getDeclaredField
at org.terasology.utilities.NativeHelper.addLibraryPath
```

### How to test

* Run Terasology in a Java 8 environment.
* Run Terasology in a Java 14 environment.

If it still works under Java 8, and it gets to the main screen in Java 14, that's better than it was before!

The other things in `natives/` could potentially be affected by this. That includes what, OpenAL, controller support, and _Lua?_

### DIFF
--- a/.idea/runConfigurations/TerasologyPC__permissive_security_.xml
+++ b/.idea/runConfigurations/TerasologyPC__permissive_security_.xml
@@ -4,7 +4,7 @@
     <module name="Terasology.facades.PC.main" />
     <option name="PROGRAM_PARAMETERS" value="-homedir -noCrashReport -permissiveSecurity" />
     <shortenClasspath name="MANIFEST" />
-    <option name="VM_PARAMETERS" value="-Xms256m -Xmx1536m" />
+    <option name="VM_PARAMETERS" value="-Xms256m -Xmx1536m -Djava.library.path=$TERASOLOGY_NATIVES$" />
     <extension name="coverage">
       <pattern>
         <option name="PATTERN" value="org.terasology.engine.*" />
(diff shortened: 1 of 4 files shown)


keturn:
Sound is working, and it looks like it's using the bundled `libopenal`. I set another property that took care of `libjinput`.

Does lua work? So far I have managed to start a world with KComputers in it that hasn't crashed, but I don't know if that proves lua works or not.

keturn:
Okay. I give up, for the moment, on testing lua. Until I see an easier test plan.

Apparently OpenComputers will fall back to LuaJ if JNLua can't load, so I'm not sure if running a minimal Computer would prove anything? I'd have to specifically test whether the suspend/persistence features work.

or have a specific idea of what to look for in the logs. It might log something when it succeeds or fails in loading the lua engine, but I'm not sure.

Given that the stuff that uses JNLua _already_ needs special launch-time arguments to work (for the permissions requirements), I think testing JNLua is not-release-blocking. JNLua users could do their own path settings at launch time, alongside their other settings.

keturn:
Okay @Cervator, I've added examples of how to set the native library path.

* for gradle, in :facades:PC:game 
* for IDEA, in the permissive security run config. Requires going to Settings / Path Variables / TERASOLOGY_NATIVES to set.

having to manually set that variable value is a stopgap, but streamlining how the build handles the natives directory is for another time.

Cervator:
Okay tested around a bit more and updated KComputers with extra details including the potential workarounds: https://github.com/Terasology/KComputers/commit/7b2c93cf2324bb4b270da8ca504841d372433cb9

Still thinking we should remove from lineup right now, but at least this way it'll be more doable to play with it anyway, and hopefully later recover it fully and move forward :-)

Thanks @keturn

PR #3939, opened 2020-05-07
feat: Particle Pool Sharing

kBlaszczyk:
This PR fixes the engine's usage of particle emitters referencing the same particle pool.

It turns out that particle pools could already be shared among multiple `ParticleEmitterComponent`s.
If done, however, the engine would update and render the pools multiple times (once for each particle emitter referencing it).

# Test
Start a new game and verify that block mining causes the expected particles to be emitted.
I'll also add a PR to Terasology/WeatherManager that fixes the module's performance and relies on this one.

### DIFF
--- a/engine/src/main/java/org/terasology/particles/ParticleSystemManagerImpl.java
+++ b/engine/src/main/java/org/terasology/particles/ParticleSystemManagerImpl.java
@@ -36,6 +36,10 @@ import org.terasology.physics.Physics;
 import org.terasology.registry.In;
 import org.terasology.registry.Share;
 
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Stream;
 
 /**
@@ -107,11 +111,16 @@ public class ParticleSystemManagerImpl extends BaseComponentSystem implements Up
     @Override
     public Stream<ParticleRenderingData> getParticleEmittersByDataComponent(Class<? extends Component> particleDataComponent) {
         return particleUpdater.getParticleEmitters().stream()
-                .filter(p -> p.ownerEntity.hasComponent(particleDataComponent))
-                .map(particleEmitterComponent ->
-                        new ParticleRenderingData<>(
-                                particleEmitterComponent.ownerEntity.getComponent(particleDataComponent),
-                                particleEmitterComponent.particlePool
-                        ));
+                .filter(emitter -> emitter.ownerEntity.hasComponent(particleDataComponent))
+                .filter(distinctByKey(emitter -> emitter.particlePool))
+                .map(emitter -> new ParticleRenderingData<>(
+                        emitter.ownerEntity.getComponent(particleDataComponent),
+                        emitter.particlePool
+                ));
+    }
+
+    private static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor) {
+        Set<Object> seen = ConcurrentHashMap.newKeySet();
+        return element -> seen.add(keyExtractor.apply(element));
     }
 }
--- a/engine/src/main/java/org/terasology/particles/ParticleSystemManagerImpl.java
+++ b/engine/src/main/java/org/terasology/particles/ParticleSystemManagerImpl.java
@@ -111,8 +111,8 @@ public class ParticleSystemManagerImpl extends BaseComponentSystem implements Up
     @Override
     public Stream<ParticleRenderingData> getParticleEmittersByDataComponent(Class<? extends Component> particleDataComponent) {
         return particleUpdater.getParticleEmitters().stream()
-                .filter(emitter -> emitter.ownerEntity.hasComponent(particleDataComponent))
-                .filter(distinctByKey(emitter -> emitter.particlePool))
+                .filter(emitter -> emitter.ownerEntity.hasComponent(particleDataComponent))  // filter emitters, whose owning entity has a particleDataComponent
+                .filter(distinctByKey(emitter -> emitter.particlePool))  // filter emitters referencing a unique particle pool
                 .map(emitter -> new ParticleRenderingData<>(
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/particles/ParticleSystemManagerImpl.java
```diff
@@ -107,11 +111,16 @@ public void update(float delta) {
...
     public Stream<ParticleRenderingData> getParticleEmittersByDataComponent(Class<? extends Component> particleDataComponent) {
         return particleUpdater.getParticleEmitters().stream()
-                .filter(p -> p.ownerEntity.hasComponent(particleDataComponent))
-                .map(particleEmitterComponent ->
-                        new ParticleRenderingData<>(
-                                particleEmitterComponent.ownerEntity.getComponent(particleDataComponent),
-                                particleEmitterComponent.particlePool
-                        ));
+                .filter(emitter -> emitter.ownerEntity.hasComponent(particleDataComponent))
+                .filter(distinctByKey(emitter -> emitter.particlePool))
```
pollend, line 115: Here are some comments that should clarify the functionality of each filter. 

```suggestion
                .filter(emitter -> emitter.ownerEntity.hasComponent(particleDataComponent)) // check if owning entity has particleDataComponent
                .filter(distinctByKey(emitter -> emitter.particlePool)) // each particle emitter has a unique pool
```

engine/src/main/java/org/terasology/particles/ParticleSystemManagerImpl.java
```diff
@@ -107,11 +111,16 @@ public void update(float delta) {
...
+                .map(emitter -> new ParticleRenderingData<>(
+                        emitter.ownerEntity.getComponent(particleDataComponent),
+                        emitter.particlePool
+                ));
+    }
+
+    private static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor) {
+        Set<Object> seen = ConcurrentHashMap.newKeySet();
+        return element -> seen.add(keyExtractor.apply(element));
```
skaldarnar, line 124: Took me a few minutes to understand what is happening here... In the end, pretty elegant solution.

engine/src/main/java/org/terasology/particles/updating/ParticleUpdaterImpl.java
```diff
@@ -116,9 +118,10 @@ public void removeEmitter(final EntityRef emitter) {
     public void update(final float delta) {
         movingAvgDelta = TeraMath.lerp(movingAvgDelta, delta, 0.05f);
 
-        for (ParticleEmitterComponent registeredParticleSystem : ImmutableList.copyOf(registeredParticleSystems)) {
-            updateParticleSystem(registeredParticleSystem, delta);
-        }
+        Collection<ParticleEmitterComponent> particleEmitters = ImmutableList.copyOf(registeredParticleSystems);
+        particleEmitters.forEach(x -> updateParticleEmitters(x, delta));
+        particleEmitters.forEach(x -> updateParticleData(x, delta));
```
skaldarnar, line 123: Is it important that particle emitters and particale data are updated in separate loops, i.e., that the particle data is updated after _all_ emitters have been updated? If so, please provide a brief comment stating the reason here.

Otherwise, this could be done in a single `forEach`:
```suggestion
        particleEmitters.forEach(x -> {
            updateParticleEmitters(x, delta);
            updateParticleData(x, delta);
        });
```

engine/src/main/java/org/terasology/particles/updating/ParticleUpdaterImpl.java
```diff
@@ -299,29 +302,39 @@ private void updateEmitter(final ParticleEmitterComponent particleEmitter, final
         }
     }
 
-    private void updateParticleSystem(final ParticleEmitterComponent partSys, final float delta) {
-        if (partSys.enabled && (partSys.particleSpawnsLeft == ParticleEmitterComponent.INFINITE_PARTICLE_SPAWNS || partSys.particleSpawnsLeft > 0)) {
-            updateEmitter(partSys, 0, delta); // Emit particles
+    private void updateParticleEmitters(final ParticleEmitterComponent emitter, final float delta) {
```
skaldarnar, line 305: Would be nice to see some Javadoc here for future contributors (and people not familiar with all the details, just like me ^^)

engine/src/main/java/org/terasology/particles/updating/ParticleUpdaterImpl.java
```diff
@@ -299,29 +302,39 @@ private void updateEmitter(final ParticleEmitterComponent particleEmitter, final
...
+        if (emitter.enabled && (emitter.particleSpawnsLeft == ParticleEmitterComponent.INFINITE_PARTICLE_SPAWNS || emitter.particleSpawnsLeft > 0)) {
+            updateEmitter(emitter, 0, delta); // Emit particles
         }
 
-        updateParticles(partSys, delta); // Update particle lifetime and Affectors
+        updateEmitterLifeTime(emitter, delta);
+    }
+
+    private void updateParticleData(final ParticleEmitterComponent particleSystem, float delta) {
```
skaldarnar, line 313: Same here. In combination with the comment further up this might also be the right spot to describe invariants or assumptions.

engine/src/main/java/org/terasology/particles/updating/ParticleUpdaterImpl.java
kBlaszczyk: comment suggestion:
```
// It's important to update all emitters before the particle data inside the pools gets updated.
// This ensures that all freshly revived particles are also being updated.
```
To explain it further:
In the code you suggested a particle pool would be updated and in following loop iterations, other emitters might revive additional particles in that pool. These weren't part of the update.

Please provide feedback if the comment makes that clear enough. I'll add it then.

skaldarnar: It does, thanks for explaining! 👍

PR #3904, opened 2020-04-22
Added a new Biomes Metric mode and fixed OnBiomeChangedEvent

vedant-shroff:
### Contains

Added a new Biomes Metric mode to the debug mode showing information regarding the current biome in which the user is present. 
Also fixes the OnBiomeChanged event and block extra data storage by passing the chunk of the block as an argument to the setExtraData function. 
### How to test

Load/Create a world, Press F3 to toggle the debug mode, Press F4 to cycle between the metrics modes till the Biomes Metrics Mode is reached. This should contain the current Biome. Test by moving around with this enabled and watch the biome name change.

PR #3942, opened 2020-05-09
chore: Remove core module line-up

skaldarnar:
As discussed in the **stabilization** workstream we are going to remove all non-essential modules from the engine's `modules` folder and move/split them to respective stand-alone modules.

## New Module Line-up
image

Code and assets have moved to different repositories. Some were split, some merged, and some just moved to a new home.

- ⤴ https://github.com/Terasology/CoreAdvancedAssets
- ✳ https://github.com/Terasology/CoreAssets
- 🚚  https://github.com/Terasology/CoreSampleGameplay
- ⤴ https://github.com/Terasology/CoreWorlds
- ⤴ https://github.com/Terasology/Debugging
- ⤴ https://github.com/Terasology/Drops
- ⤴ https://github.com/Terasology/Explosives
- ⤴ https://github.com/Terasology/Furnishings
- 🚚 https://github.com/Terasology/BiomesAPI

Two modules should be deprecated:

- ❌ ~~[Terasology/CoreBlocks](https://github.com/Terasology/CoreBlocks)~~
- ❌ ~~[Terasology/CoreItems](https://github.com/Terasology/CoreItems)~~

✳ _extended_ ⤴ _extracted_ 🚚 _moved_ ❌ _deprecated_

## Related PRs

Closes #3922
Closes #3926 
Closes #3925 

Affects #3904


docs/Modules.md
```diff
@@ -57,7 +57,7 @@ Here's a list of modules bundled with the game by default (this line-up will cha
 * Core - mandatory content needed for normal game launch
 * [CoreAssets](https://github.com/Terasology/CoreAssets) - base assets extracted from Core
 * [CoreBlocks](https://github.com/Terasology/CoreBlocks) - base blocks extracted from Core
-* CoreSampleGameplay - gameplay front for Core - allows modules to depend on Core without the default starting inventory
+* [CoreSampleGameplay](https://github.com/Terasology/CoreSampleGameplay) - default gameplay with a few core modules enabled
```
skaldarnar, line 60: What is this file used for? I don't think it's up to date, is it?

engine-tests/out/test/resources/gameManifest.json
```diff
@@ -10,7 +10,7 @@
       "title": "main",
       "seed": "naqrhaKV1lUofG7PLIJcOfDlUiPs3Jmp",
       "time": 25538256,
-      "worldGenerator": "Core:heightMap"
+      "worldGenerator": "BuilderSampleGameplay:BuilderSampleWorld"
```
skaldarnar, line 13: I don't know what this `gameManifest`.json` is used for, to be honest. I just replaced this value with the builder sample world as that is the only one still living in the engine repo....

Does anybody know whether the file is still used?

engine-tests/src/test/resources/gameManifest.json
```diff
@@ -10,7 +10,7 @@
       "title": "main",
       "seed": "naqrhaKV1lUofG7PLIJcOfDlUiPs3Jmp",
       "time": 25538256,
-      "worldGenerator": "Core:heightMap"
+      "worldGenerator": "BuilderSampleGameplay:BuilderSampleWorld"
```
skaldarnar, line 13: Same here, the file seems to be same as the one in the `out` folder...

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
```diff
@@ -56,6 +56,8 @@
     /**
      * Name of the core gameplay module. The core gameplay module adds a minimal environment for some basic gameplay.
      */
+    //TODO: CSG moved out of the engine repo, so this should either be removed or reference the Builder Sample Gameplay
+    //      which still lives in the engine repo.
     public static final Name CORE_GAMEPLAY_MODULE = new Name("coresamplegameplay");
```
skaldarnar, line 61:
```suggestion
    public static final Name CORE_GAMEPLAY_MODULE = new Name("buildersamplegameplay");
```

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
```diff
@@ -56,6 +56,8 @@
     /**
      * Name of the core gameplay module. The core gameplay module adds a minimal environment for some basic gameplay.
      */
+    //TODO: CSG moved out of the engine repo, so this should either be removed or reference the Builder Sample Gameplay
```
skaldarnar, line 59: I'm not sure what these constants are used for... would it make sense to change this to `buildersamplegameplay`?

engine/src/main/java/org/terasology/input/internal/BindCommands.java
```diff
@@ -58,7 +58,6 @@ public String bindKey(@CommandParam("key") String key, @CommandParam("function")
         AZERTY.put(KeyId.Z, new SimpleUri("engine:forwards"));
         AZERTY.put(KeyId.S, new SimpleUri("engine:backwards"));
         AZERTY.put(KeyId.Q, new SimpleUri("engine:left"));
-        AZERTY.put(KeyId.A, new SimpleUri("Core:dropItem"));
```
skaldarnar, line 61: I think the whole special handling for different keyboard layouts is "broken" in the sense that it does not work with modules at all... we might just remove it at this point... 🤷

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
```diff
@@ -198,7 +198,9 @@ public void set(String value) {
         });
 
         WidgetUtil.trySubscribe(this, "addGenerator", button -> {
-            if (worldGenerator.getSelection().getUri().toString().equals("Core:heightMap")) {
+            //TODO: there should not be a reference from the engine to some module - the engine must be agnostic to what
+            //      modules may do
+            if (worldGenerator.getSelection().getUri().toString().equals("CoreWorlds:heightMap")) {
```
skaldarnar, line 203: I don't know how to deal with this, but as it is, that's a potential bug waiting around the corner... at least, there should be a generic way for modules to mark that some world gen is not suited for advanced universe setup

engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
```diff
@@ -163,8 +163,9 @@ private void commonDestroyed(DoDestroyEvent event, EntityRef entity, Block block
         if (blockDamageModifierComponent == null || !blockDamageModifierComponent.skipPerBlockEffects) {
             // dust particle effect
             if (entity.hasComponent(LocationComponent.class) && block.isDebrisOnDestroy()) {
+                //TODO: particle system stuff should be split out better - this is effectively a stealth dependency on
```
skaldarnar, line 166: This was a TODO even before this PR... @kBlaszczyk you just touched the particle system ... 🙄 Jokes aside, how can we get this out of the engine?

templates/override.cfg
```diff
@@ -6,6 +6,6 @@
     "defaultGameplayModuleName": "CoreSampleGameplay"
   },
   "worldGeneration": {
-    "defaultGenerator": "Core:facetedperlin"
+    "defaultGenerator": "CoreWorlds:facetedperlin"
```
skaldarnar, line 9: what is this `override.cfg` used for?

docs/Modules.md
Cervator: I've done best effort to keep `Modules.md` in sync manually with the omega lineup, yeah, but that's a poor approach. More automation would be better (like the module site)

engine-tests/out/test/resources/gameManifest.json
Cervator: Good question. Being that it was set to a generator that doesn't work ... probably either not being used or that particular value doesn't matter anyway :-)

We don't generally delete unit tests though, so maybe something is hiding somewhere and still at least trying to use this 🤔 Unless maybe it got moved out as part of the Record & Replay system, or for the MTE?

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
Cervator: That's fine by me and IIRC we have some logic somewhere else that attempts to detect JoshariasSurvival as the default when the full module lineup is present anyway.

engine/src/main/java/org/terasology/engine/TerasologyConstants.java
Cervator: Some of the constants are surely still in use, but maybe not all. I think this should be fine to also change to BSG

engine/src/main/java/org/terasology/input/internal/BindCommands.java
Cervator: Yeah the azerty and the one (or two?) other keyboard layout commands we have might be good examples of the sort of code we took in from _one_ contributor who found it useful at some point in the past, but it may not be worth maintaining as-is if we can gain some benefit from cleaning up and later re-adding if there's a better way to do it and more interest in using it.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
Cervator: Yeah, that whole project itself (multiworld) really needs to be resumed and finished. Along with the better phasing of things during game setup so you can get things like block registration and hot keys sorted out before you enter the world.

templates/override.cfg
Cervator: See https://github.com/MovingBlocks/Terasology/blob/develop/docs/Playing.md#configuring-a-server :-) Also works for local development when you need to test via headless server.

engine/src/main/java/org/terasology/input/internal/BindCommands.java
pollend: how would we want to handle different keyboard layouts? I also don't really like how we use annotations to register bindings against a whole class. Inventory has 15 or so classes just dedicated to keybindings. Probably need to look at how bindings work and also look at maybe how some other engines handle bindings and what we can do to make the process better. Some key press events happen depending on context so different events can be handled by the same key depending on what the player is doing. probably a way to handle bindings depending on conditions in the engine.

engine-tests/out/test/resources/gameManifest.json
skaldarnar: If a unit test is using this and the value does not matter, we should make that clearer by using something like `Gooey:DummyWorld` or similar. If it is not used, we should delete it... 
I can try to delete it and see if anything breaks in the tests...

templates/override.cfg
skaldarnar: So, should we reference the world generator that might not be present here, or stick with the sample builder world thingy that still lives in the engine repo?

jdrueckert:
image

That's the new dependency graph when checking out this branch and recursing `CoreSampleGameplay`

Cervator:
Oh, for the change notes with the next release it'll be important to point out old saves will likely _not_ work, unless tweaked a bit to address some of the very base stuff that moved around (not even sure that'll work, but it isn't something i'd personally worry about

Devs will need to do some extra cleanup in existing workspaces too, like delete any stray files from the formerly embedded modules, clear out any jars in `modules/` etc. Stuff we can monitor and help fix via chat :-)

engine/src/main/java/org/terasology/input/internal/BindCommands.java
keturn: I am a person who was glad to see the button for switching to a dvorak keyboard layout in the config screen! I just say that as +1 to interest for Cervator's "does anyone use this?" point, but I know it does nothing to address your concerns about its implementation.

(I have _ambitions_ for input configuration stuff myself, but I won't get to act on them for a while and this PR is not the place.)

Cervator: Feedback appreciated @keturn - would indeed love to see somebody actually using one of those layouts help make the implementation a little better 😁

PR #3947, opened 2020-05-10
feat(JOML): Migrate BlockChange

kBlaszczyk:
This PR migrates `BlockChange` to JOML and adjusts its usage throughout the engine.

### DIFF
--- a/engine/src/main/java/org/terasology/world/internal/WorldProviderCoreImpl.java
+++ b/engine/src/main/java/org/terasology/world/internal/WorldProviderCoreImpl.java
@@ -78,7 +78,7 @@ public class WorldProviderCoreImpl implements WorldProviderCore {
 
     private final List<WorldChangeListener> listeners = Lists.newArrayList();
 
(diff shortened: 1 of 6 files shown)


engine/src/main/java/org/terasology/world/internal/WorldProviderCoreImpl.java
```diff
@@ -201,9 +200,9 @@ public Block setBlock(Vector3ic worldPos, Block type) {
             Vector3i blockPos = ChunkMath.calcBlockPos(JomlUtil.from(worldPos));
             Block oldBlockType = chunk.setBlock(blockPos, type);
             if (oldBlockType != type) {
-                BlockChange oldChange = blockChanges.get(worldPos);
```
kBlaszczyk, line 204: Note:
This was an error in the previous joml-migration commit. `blockChanges` uses TeraMath-vectors as keys, but here a JOML-vector is being used to find an element.
The TeraMath- and JOML-vector classes are not hash-equal.

PR #3948, opened 2020-05-10
feat(JOML) replace perspective and ortho with JOML impl

pollend:
joml already implements perspective and ortho projections so we can default to that implementation over the one provide in MatrixUtil.


engine/src/main/java/org/terasology/rendering/cameras/OrthographicCamera.java
```diff
@@ -89,7 +89,7 @@ public void updateMatrices(float fov) {
             return;
         }
 
-        projectionMatrix = MatrixUtils.createOrthogonalProjectionMatrix(left, right, top, bottom, zNear, zFar);
+        projectionMatrix = new Matrix4f().ortho(left,right,top,bottom,zNear,zFar).transpose();
```
kBlaszczyk, line 92: The top and bottom parameters need to be switched.
```suggestion
        projectionMatrix = new Matrix4f().ortho(left,right,bottom,top,zNear,zFar).transpose();
```

pollend:
@kBlaszczyk I added unit test based off the original behavior and verified that the old and current behavior are the same.

httpdigest:
I was also wondering why you always work with the transposed matrix? The difference to the "normal" OpenGL way of having `matrix*vector` convention with column-vectors is that with the transposed matrices your columns are rows, you need to "left-multiply" matrices (e.g. `model * view * projection` instead of the OpenGL convention `projection * view * model`) and you need to left-multiply vectors `vector * matrix` instead of `matrix * vector`. Why do you do this? I assume you can shave off some cycles by just omitting that transposition everywhere.

kBlaszczyk:
> I was also wondering why you always work with the transposed matrix?

The problem is that the math library we are replacing was using row major matrices. So in order to cause as little damage and confusion as possible, we are transposing the joml matrices to keep the calculations unchanged.
I think (or hope?) that as soon as we are fully migrated, we'll adjust the multiplication order throughout the engine to get rid of `transpose()` calls and do things the OpenGL/JOML way.

engine-tests/src/test/java/org/terasology/rendering/cameras/PerspectiveCameraTest.java
```diff
@@ -55,24 +56,28 @@ public void testPerspectiveCamera() {
...
         camera.updateMatrices();
 
-        Matrix4f viewMatrix = new Matrix4f(5.0273395f, 0.0f, 0.0f, 0.0f
-                , 0.0f, 8.937492f, 0.0f, 0.0f
-                , 0.0f, 0.0f, -1.00004f, -0.20000401f
-                , 0.0f, 0.0f, -1.0f, 0.0f);
+        Matrix4f viewMatrix = new Matrix4f(-5.253E-1f, 0.000E+0f, 8.509E-1f, 0.000E+0f,
+            8.509E-4f, 1.000E+0f, 5.253E-4f, 0.000E+0f,
+            -8.509E-1f, 1.000E-3f, -5.253E-1f, 0.000E+0f,
+            -0.000E+0f, -0.000E+0f, -0.000E+0f, 1.000E+0f);
```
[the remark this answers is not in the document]
keturn: (I'm glad you moved those line-starting commas, though. 🙂 I don't think start-with-comma is a standard this repository follows.)

engine-tests/src/test/java/org/terasology/rendering/cameras/PerspectiveCameraTest.java
```diff
@@ -55,24 +56,28 @@ public void testPerspectiveCamera() {
...
-                ,-0.96200305f,0.0014514439f,-0.27303496f,-0.0f,
-                0.0013962948f,0.99999887f,3.962953f,-0.0f,
-                0.0f,0.0f,0.0f,1.0f);
+            , 0.0f, 8.937492f, 0.0f, 0.0f
+            , 0.0f, 0.0f, -1.00004f, -0.20000401f
+            , 0.0f, 0.0f, -1.0f, 0.0f);
+        Matrix4f normViewMatrix = new Matrix4f(-.525322f, 0.0f, 0.8509036f, -0.0f
+            , 8.509035E-4f, 0.9999995f, 5.25322E-4f, -0.0f,
+            -0.85090315f, 0.001f, -0.5253218f, -0.0f,
+            0.0f, 0.0f, 0.0f, 1.0f);
```
keturn, line 70: Here we've got what looks to be the same values as in `viewMatrix`, _almost._

Are those the same numbers, but displayed a little differently due to binary-floating-point vs decimal representations? Or is the change of 0.01% significant? Could this have been a copy of viewMatrix instead?

I don't believe it's necessary to remove _all_ numbers from tests (in a math library, even!), but I'm failing to tell signal from noise here.

engine-tests/src/main/java/org/terasology/testUtil/TeraAssert.java
```diff
@@ -65,6 +66,34 @@ public static void assertEquals(Vector3f expected, Vector3f actual, float error)
...
+
+            org.junit.jupiter.api.Assertions.assertEquals(expected.m20(), actual.m20(), error, errorMessageSupplier);
+            org.junit.jupiter.api.Assertions.assertEquals(expected.m21(), actual.m21(), error, errorMessageSupplier);
+            org.junit.jupiter.api.Assertions.assertEquals(expected.m22(), actual.m22(), error, errorMessageSupplier);
+            org.junit.jupiter.api.Assertions.assertEquals(expected.m23(), actual.m23(), error, errorMessageSupplier);
+
+            org.junit.jupiter.api.Assertions.assertEquals(expected.m30(), actual.m30(), error, errorMessageSupplier);
+            org.junit.jupiter.api.Assertions.assertEquals(expected.m31(), actual.m31(), error, errorMessageSupplier);
+            org.junit.jupiter.api.Assertions.assertEquals(expected.m32(), actual.m32(), error, errorMessageSupplier);
+            org.junit.jupiter.api.Assertions.assertEquals(expected.m33(), actual.m33(), error, errorMessageSupplier);
```
keturn, line 93: I bet there's a way we can make this assertion method easier to read _and_ give more useful output on fail. I think you're following established precedent here, so I certainly won't hold up a review on that, but maybe a little side project for sometime.

engine-tests/src/test/java/org/terasology/rendering/cameras/PerspectiveCameraTest.java
keturn: On a second reading, I see this test _does_ have that camera.setOrientation with the 0.001 radian pitch and yaw, so maybe those tiny differences _are_ significant?

engine-tests/src/test/java/org/terasology/rendering/cameras/PerspectiveCameraTest.java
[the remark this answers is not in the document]
pollend: used some random position and tested it with the old code and then moved it back to this. the test might not be necessary.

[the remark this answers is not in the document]
keturn: we talked about this on discord and I agreed that if these methods are just calling through to joml's orthographic and perspective methods, we don't learn much by testing the arithmetic ourselves.

PR #3952, opened 2020-05-12
chore(groovyw): Add command to list local modules

skaldarnar:
Adds a flag '--local' that can be added to `groovyw module list` in
order to only list modules that are present locally as source.

The output of this command is machine readable such that it can be used
easily to perform some command for all modules.

For instance, the following Bash command will reset and update all
modules:

```sh
for m in $(./groovyw module list --local)
do
    pushd modules/$m
    git reset --hard HEAD
    git pull
    popd
done
```

### DIFF
--- a/config/groovy/util.groovy
+++ b/config/groovy/util.groovy
@@ -153,23 +153,29 @@ switch (cleanerArgs[0]) {
         break
 
     case "list":
-        ListFormat listFormat = determineListFormat(cleanerArgs)
         String[] availableItems = common.retrieveAvailableItems()
         String[] localItems = common.retrieveLocalItems()
         String[] downloadableItems = availableItems.minus(localItems)
-        println "The following items are available for download:"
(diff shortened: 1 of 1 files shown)


skaldarnar:
Is there some good library of sorts to write a CLI tools with Groovy? I've been working with some JS code recently where you easily get some command line parser for (even advanced) CLI tools...

Overall, before adding _more_ to the groovy scripting we should bring some better structure to it 🙈 (at least my 2 cents on the topic)

Cervator:
Good question - I'm unsure, but that'd be a useful bit of research.

Coincidentally I'm working on it a bit for a work project, but unsure where that's going to go. Will keep an eye on potential structural improvements and bring them back here if relevant :-)

keturn:
[https://ajalt.github.io/clikt/](Clikt) for command line parsing?

PR #3953, opened 2020-05-12
fix: allow game to quit without waiting for registry updater thread

### LINKED ISSUE #3940: process does not quit when game closes, some threads still live
I've just quit Terasology by the exit button on the main menu, and the window closed but the java process did not end.

While I was typing this up, it _eventually_ closed, about eight minutes after initially logging the "Shutting down Terasology" message.

Discovered in the process of investigating https://github.com/MovingBlocks/TerasologyLauncher/issues/563

Release 36. Or 81. Or 3.2.0. Which of these numbers is the important one? See attached [Terasology.log](https://github.com/MovingBlocks/Terasology/files/4602386/Terasology.log)

Also: 
[Thread Dump](https://github.com/MovingBlocks/Terasology/files/4602388/thread-dump.txt)

### Computer details

Linux, JDK is liberica-11.

keturn:
This fixes #3940. The thread that was holding things up there was this one that's updating the mod registry.

### How to test

* Start Terasology. Push the _Exit Terasology_ button quickly, before everything has time to settle.
* Make sure the window not only closed, but the process has quit as well. i.e. you see "process finished with exit code 0" in intellij or "build successful" and a return to the command line from the gradle task.

- [ ] use a more convenient and readable thread runner

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
@@ -15,9 +15,11 @@
  */
 package org.terasology.rendering.nui.layers.mainMenu.advancedGameSetupScreen;
 
+import ch.qos.logback.classic.pattern.TargetLengthBasedClassNameAbbreviator;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.codehaus.plexus.util.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


keturn:
The initial version of this I've provided here is deliberately verbose.

Not that I want creating threads to be easy, but I certainly don't want the right way to be this much more difficult than the wrong way.

The three qualities we've got here are:

1. returns a Future
2. is a [daemon thread](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Thread.html)
3. has a name that tells us something useful about what it's doing or where it was spawned

and I threw in the abbreviator for good measure, but 

```java
org.terasology.rendering.nui.layers.mainMenu.advancedGameSetupScreen.AdvancedGameSetupScreen#initialise
```

_is_ a really long string.

keturn:
Side note: Is it surprising that this is running as soon as the MainMenuScreen is initialized? The user may never visit the AdvancedGameSetupScreen.

keturn:
oh, that distracted me from what I was saying about my deliberately-verbose implementation. That's a straw man, posted here in the hope of generating discussion that will point toward an existing utility, or some agreement about where to add this as a new utility function.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/advancedGameSetupScreen/AdvancedGameSetupScreen.java
```diff
@@ -130,8 +132,11 @@ public void onOpened() {
     @Override
     public void initialise() {
         setAnimationSystem(MenuAnimationSystems.createDefaultSwipeAnimation());
-        remoteModuleRegistryUpdater = Executors.newSingleThreadExecutor()
-                .submit(moduleManager.getInstallManager().updateRemoteRegistry());
+        remoteModuleRegistryUpdater = Executors.newSingleThreadExecutor(
+                new ThreadFactoryBuilder()
+                        .setNameFormat(new TargetLengthBasedClassNameAbbreviator(36).abbreviate(getClass().getName()) + "-%d")
```
DarkWeird, line 137: "ModuleRegistryUpdater"  sounds better and more ralate then "AdvancedGameSetupScreen"

pollend:
Just tested this with and without the PR and the game exits correctly when closing.

PR #3954, opened 2020-05-13
add IntelliJ configuration

keturn:
I was working on #3951 and came to the conclusion that it's counterproductive to try to manually tweak gradle and Intellij configs to sync their compiler-output directories.

Wondering “what if I don't do that?” led me to “what if we don't use `./gradlew idea` _at all?_”

So I let IntelliJ import the gradle project and save its project configuration.

What needs to be done to confirm that this is no less broken than the `gradlew idea` version?

## Contains

Support for IntelliJ IDEA 2020.1. Some of it might be backwards-compatible with 2019 versions, but I recommend only trying to support 2020.1+

### An `.idea/` directory

with the project configuration, including features from [the prior `ide.gradle`](https://github.com/MovingBlocks/Terasology/blob/v3.2.0/config/gradle/ide.gradle):

* Checkstyle
* Copyright template
* Terasology-specific annotations marked as entry points
* all the Run Configs.

as well as:

* A startup task that runs `gradlew extractConfigs extractNatives copyInMissingTemplates` so they'll be in place right away.
* Code style setting to not use `*` imports with fewer than 999 imports. (yes, there's really no way to turn it off, setting it to a ridiculously high number is what everyone recommends.)

#### Omitted:

* git stuff. I haven't tested this extensively, but I think recent intellij figures out git roots pretty well on its own?
* compiler "build project automatically" setting. Seems to be stored with the workspace settings, much easier for us to not try to mess with those.

### gitignore changes

* removes some IntelliJ-IDEA ignores, because we want the .idea directory tracked now. The IDE seems to also think you don't want configs tracked if you ignore `.ipr` files, though this uses the directory-based format instead of IPRs.
* Some general clean-up. I think we all have a "somewhat recent git version" compared to that comment from 2015.
* A few things are ignored in `.idea/.gitignore`. I started writing notes in there and it turned in to [a lot of words about how to store .idea in git](https://keturn.net/blog/posts/2020/05/git-add-dot-idea-share-intellij/). I can cross-post that to the forum or wiki or something if that would help.

## How to Test

Things to try, each from a clean checkout (no old build results, no old IntelliJ configs):

* make sure `./gradlew game` works without ever opening it in IntelliJ.
* open in IntelliJ 2020.1
  * agree to the Import Gradle prompt
  * make sure the TerasologyPC run configuration works without ever manually running gradle.
* open in IntelliJ 2020.1
  * agree to the Import Gradle prompt
[...]

PR #3708, opened 2019-07-23
Collective behavior trees

casals:
### Contains
- Additions and small changes in the `org.terasology.logic.behavior` package.

### Objective
To create a collective beehavior structure capable of interpreting a single behavior tree for a group of actors. This was necessary due to some restrictions found in the original group model while working on the last tasks of Milestone #2. The scenario is discussed in [the GSoC Milestone 2 blog post](https://casals.io/code/gsoc-reaching-second-milestone/).

### Changes

The following classes were added to the engine:

- `org.terasology.logic.behavior.core.CollectiveBehaviorTreeRunner`

- `org.terasology.logic.behavior.DefaultCollectiveBehaviorTreeRunner`

- `org.terasology.logic.behavior.CollectiveInterpreter`

- `org.terasology.logic.behavior.CollectiveBehaviorComponent`

Also: 

- the interface `Callback` was extracted from org.terasology.logic.behavior.DefaultBehaviorTreeRunner (not being used yet)

- copy constructors were added to the classes `CollectiveInterpreter` and `Interpreter`. This change is related to the task involving persistence of a behavior tree execution state for a single entity.

### Usage

- Install the module [WildAnimalsMadness] ([PR 31 for WildAnimals](https://github.com/Terasology/WildAnimals/pull/31) must be merged)

- Create/load a game using the module WildAnimalsMadness

- Execute the command `groupTestThree`

PR #3955, opened 2020-05-14
chore: remove stray compiler outputs

keturn:
accidentally added in #3708.

found while working on #3954 but I figured I'd make a separate PR for this for ease of reviewing.


DarkWeird:
@keturn  is ```out```  directory generated by IDE?

keturn:
Yep

DarkWeird:
> Yep

Then need add gitignore rule, isn't it?

keturn:
Oh, that would make sense to include in this same PR, yes. Added.

PR #3958, opened 2020-05-14
fix: world gen should use the configuration provided by the user

skaldarnar:


engine/src/main/java/org/terasology/persistence/internal/EntitySetDeltaRecorder.java
```diff
@@ -67,7 +67,9 @@ public void onEntityComponentChange(EntityRef entity, Class<? extends Component>
             EntityDelta entityDelta = getOrCreateEntityDeltaFor(entity);
             Component component = entity.getComponent(componentClass);
             Component componentSnapshot = componentLibrary.copy(component);
```
skaldarnar, line 69: this call to `.copy` fails and `componentSnapshot` is `null`... not sure this check should be in here or not, but we face similar a similar issue one step later in `NetworkSystemImpl` ...

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/GameManifestProvider.java
```diff
@@ -48,55 +50,91 @@ private GameManifestProvider() {
...
         GameManifest gameManifest = new GameManifest();
-        if (StringUtils.isNotBlank(universeWrapper.getGameName())) {
-            gameManifest.setTitle(universeWrapper.getGameName());
-        } else {
-            gameManifest.setTitle(GameProvider.getNextGameName());
+
+        try {
+            addModulesTo(gameManifest, new DependencyResolver(moduleManager.getRegistry()), config);
+        } catch (DependencyResolutionFailedException e) {
+            logger.error(e.getMessage());
```
keturn, line 69: you can send the exception object itself to the logger without the getMessage, and it will recognize it as an exception and do its exception-formatting thing.

but that can be very verbose, if it's configured to do full stacktraces and whatnot, so disregard this comment if you intentionally want to keep it brief.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/GameManifestProvider.java
```diff
@@ -48,55 +50,91 @@ private GameManifestProvider() {
...
+            return null;
         }
 
+        gameManifest.setTitle(getGameTitle(universeWrapper.getGameName()));
+        gameManifest.setSeed(getSeed(universeWrapper));
+
+        // This is multiplied by the number of seconds in a day (86400000) to determine the exact  millisecond at which the game will start.
+        long startTime = (long) (WorldTime.DAY_LENGTH * WorldTime.NOON_OFFSET);
```
keturn, line 77: I know this PR moved this comment, it's not responsible for how it was written in the first place, but since it's here:

I saw it before and was left confused. Is something dependent on knowing or setting the start time to millisecond precision? what?

engine/src/main/java/org/terasology/world/generation/BaseFacetedWorldGenerator.java
```diff
@@ -83,6 +84,15 @@ public WorldConfigurator getConfigurator() {
         return configurator;
     }
 
+    @Override
+    public void setWorldConfigurator(WorldConfigurator configurator) {
+        //TODO: figure out what to do with other world configurator classes...
+        //      maybe this should not be exposed on WorldGenerator at all...
+        if (configurator instanceof FacetedWorldConfigurator) {
+            this.configurator = (FacetedWorldConfigurator) configurator;
+        }
```
keturn, line 93: should this log a warning? (at least until something else is figured out)

engine/src/main/java/org/terasology/persistence/internal/EntitySetDeltaRecorder.java
keturn: do we know anything yet about why that fails? or still needs research?

keturn:
Is the EntitySetDeltaRecorder question the reason this is still a draft? Or is there additional stuff needed to fix the issue?

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/GameManifestProvider.java
skaldarnar: Thanks for pointing this out :upside_down_face: As I'm aware of the fact that I can simply pass `e` directly I _guess_ I did this on purpose because the output would have been too verbose :thinking: not sure though :see_no_evil:

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/GameManifestProvider.java
skaldarnar: To be honest: I have no idea :roll_eyes: I hope not...

engine/src/main/java/org/terasology/world/generation/BaseFacetedWorldGenerator.java
skaldarnar: If we keep that in, probably yes... although I don't like interfaces that depend on being implemented using shabby type casts... :confused: I still hope that we'll find a better way to do this with a better understanding of what is happening here...

skaldarnar:
> Is the EntitySetDeltaRecorder question the reason this is still a draft? Or is there additional stuff needed to fix the issue?

I think the delta recorder was the first place where it failed, but there were other listeners in the queue to be notified... We have to double check that claim, though.

skaldarnar:
This was a WIP PR with some rambling here and there. Can be closed.

Issue #3846, opened 2020-03-07
New Gradle approach seems to accept newer than max version module dependencies


Cervator:
### What you were trying to do

Build modules in the Jenkins after the big Gradle upgrade in #3835. I figured a bunch would fail as the version bump from v2 to v3 in the engine plus its embedded modules should mean a bunch of modules should fail - by default we don't accept a new major version bump to satisfy a minimum requirement on the prior major version. 

### What actually happened

Mostly everything went fine, including locally in a source workspace.

### How to reproduce

* Have a module A depend on module B at version v1.0.0
* Bump the available version of B to v2.0.0
* Module A should no longer have its dependencies satisfied

You can see the intended behavior in the in-game module browser which will mark A in red text if B is out of bounds on its dependency requirement. But Gradle happily accepts it right now. Not really a harmful "bug" right now but something we should adjust for 👍 

Out of scope: if in the future with better release management we'd want to allow a new major as acceptable with some sort of automated disclaimer to the user.

Maybe of interest to @DarkWeird (thanks for all the Gradle work again!)

DarkWeird:
It is "+" replacement time! :D
Seems, It is change for 3 lines  of gradle code :D
(Using version "2+" instead version "+" for deps)

PR #3959, opened 2020-05-16
spike: update gradle and get overly ambitious about build dependencies

keturn:
### Contains

* upgrades gradle from 6.0 to 6.4.1
* changes to build.gradle files to address deprecation warnings

### How to test

make sure stuff still builds in gradle and your IDE

- [ ] merge #3954


keturn:
are there things that are subtle and testable? would this be a good way to test them? https://docs.gradle.org/current/userguide/test_kit.html

a note for later. haven't dug in to it yet.

DarkWeird:
You can stuck on module's dependencies.
```compile``` configuration is deprecated (pre 7.0)
Modules use it for deps from ```module.txt```
if you want replace it for ```implementation```, then you must write all dependencies as direct....
Or  you can create new configuration similar ```compile```

keturn:
this branch is some highly experimental work-in-progress stuff right now, but I've tried to separate out the commits so maybe they can be cherry-picked even if the experiment doesn't end up so well

keturn:
I crossed a line somewhere and the original issue title is no longer accurate, the current state of the branch is far more ambitious than "fix deprecated uses"

keturn:
Okay, how do I rein in this branch so we can get some work merged before spiraling in to an endless rewrite?

I think there is value in a PR that does the deprecation-fixing and upgrading. Including the new groovy-wrapper.jar is key for the upgrade to stay compatible.

The cutoff is _before_ I started trying to rewire the inter-project dependencies and resolution strategies.

Also need to know how to check this kind of thing for regressions. Need to make sure we can still build a release from it.

Do we ship any other artifacts from this? are there any consumers of the engine.jar?

DarkWeird:
engine.jar used by single-module build and releases. It is work via maven repo.

P.S. about interproject deps: #3846(fyi :))

keturn:
Okay, I've pushed all my outstanding work in progress to this branch, and am calling that the end of this experiment. I'll be picking pieces of it in to a new, more manageable branch.

keturn:
Another thing I learned along the way:

I was trying to make a gradle Project such that you could depend on it to (transitively) depend on all the modules sub-projects. My first idea was to make that a `java-library`, though I had doubts about how well that would work since it didn't really have a compilation source set of its own.

I later learned that gradle has another type of thing, a “[Java Platform](https://docs.gradle.org/current/userguide/java_platform_plugin.html)” which might be more fitting. It's just a dependency/constraint container, it doesn't have artifacts of its own. (They compare it to a Maven BOM, if that means anything to you.)

I hadn't put that in to practice yet but I think it's something to try for the next iteration.

PR #3962, opened 2020-05-17
remove ConcurrentMapChunkCache

pollend:
I removed the ConcurrentMapChunkCache since it just seems to just wrap a map unless we have some plans to support different types of cache. the other thing we can do is Also use the same implementation for the RemoteChunkProvider since that implementation also uses a chunkCache and that just uses a Maps.newHashMap(). At the moment there doesn't really seem to be a need to implement more then one type of cache. 

For chunkFinalizer I removed the instance that gets passed in at startup and just grab an instance out of the Supplier. when purgeWorld is called; this will trigger another instance from the Supplier. this would make sense since the only other instance just declares new LightMergingChunkFinalizer and LightMergingChunkFinalizer::new.


engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -112,38 +111,23 @@
 
     private BlockManager blockManager;
     private ExtraBlockDataManager extraDataManager;
-    private final ChunkCache chunkCache;
-    private final Supplier<ChunkFinalizer> chunkFinalizerSupplier;
+    private final Map<Vector3i, Chunk> chunkCache = Maps.newConcurrentMap();
```
skaldarnar, line 114: I guess this needs to be JOML-ified now?

engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -112,38 +111,23 @@
...
     private BlockEntityRegistry registry;
 
+    //TODO: Michael Pollind -- if we have more then one finalizer use  Supplier<ChunkFinalizer[]> and ChunkFinalizer[]
+    private final Supplier<ChunkFinalizer> chunkFinalizerSupplier;
     private ChunkFinalizer chunkFinalizer;
 
-    //TODO Remove this old constructor at the end of the chunk overhaul
```
skaldarnar, line 121: when is the "end of the chunk overhaul"? 🤔

engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -112,38 +111,23 @@
 
     private BlockManager blockManager;
     private ExtraBlockDataManager extraDataManager;
-    private final ChunkCache chunkCache;
-    private final Supplier<ChunkFinalizer> chunkFinalizerSupplier;
+    private final Map<Vector3i, Chunk> chunkCache = Maps.newConcurrentMap();
     private BlockEntityRegistry registry;
 
+    //TODO: Michael Pollind -- if we have more then one finalizer use  Supplier<ChunkFinalizer[]> and ChunkFinalizer[]
+    private final Supplier<ChunkFinalizer> chunkFinalizerSupplier;
```
skaldarnar, line 118: What do we need to store the supplier as member here? When is it used?

skaldarnar (commented):
Hey @pollend, is this still relevant, or rather, should we try to get this merged? I'm always a fan of removing code, so belated 👍 from here 😉

DarkWeird:
This PR not actual after ChunkRefactoring part 2 :p

pollend:
I can re-base these changes on to the latest develop.

pollend:
@DarkWeird there you go. looks like we can just remove ConcurrentMapChunkCache

DarkWeird (approved):

PR #3967, opened 2020-05-18
feat(JOML): BlockRegion replacement for Region3i

pollend:


engine-tests/src/test/java/org/terasology/math/BlockRegionTest.java
```diff
@@ -0,0 +1,165 @@
...
+        for (Vector3ic pos : BlockRegionIterable.region(region).build()) {
+            assertTrue(expected.contains(pos), "unexpected position: " + pos);
+            expected.remove(pos);
+        }
+
+        assertEquals(0, expected.size(), "All vectors provided");
+    }
+
+    @Test
+    public void testSimpleIntersect() {
```
kBlaszczyk, line 123: The usage of `unionBlock()` in these intersection tests is absolutely irrelevant. You should use the min-max constructor and adjust the expected values where necessary.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -0,0 +1,243 @@
...
+    @Override
+    public BlockRegion transform(Matrix4fc m) {
+        return (BlockRegion) super.transform(m);
+    }
+
+    public BlockRegion transform(Matrix4fc m, BlockRegion dest) {
+        return (BlockRegion) super.transform(m, dest);
+    }
+
+    public BlockRegion unionBlock(int x, int y, int z, BlockRegion dest) {
```
kBlaszczyk, line 175: unionBlock() adds +1 to the max value. I assume that is to make sure that the specified vector is fully enclosed into the bounding box. Is that correct? I'm wondering why the same thing isn't done for the min vector. If my region min is at (10, 10, 10) and I'm unionizing with (5, 5, 5), don't I expect the new min to be at (4, 4, 4)? Why would this only be relevant for max?
This method should also have some Javadoc on it.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -0,0 +1,243 @@
...
+
+            maxX = Integer.MIN_VALUE;
+            maxY = Integer.MIN_VALUE;
+            maxZ = Integer.MIN_VALUE;
+        }
+        return this;
+    }
+
+    public BlockRegion intersection(BlockRegion other, BlockRegion dest) {
+        dest.minX = Math.max(minX, other.minX);
```
kBlaszczyk, line 212: Why are we using `Math.min/max()` here but not above in `unionBlock()`?

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -0,0 +1,243 @@
...
+import org.joml.AABBi;
+import org.joml.Math;
+import org.joml.Matrix4fc;
+import org.joml.Rectanglei;
+import org.joml.RoundingMode;
+import org.joml.Vector3i;
+import org.joml.Vector3ic;
+import org.terasology.entitySystem.entity.EntityRef;
+
+public class BlockRegion extends AABBi {
```
kBlaszczyk, line 27: I don't like that this class extends `AABBi` and yet overrides every method to add a simple cast.
If it's done like that, you could as well not extend AABBi and have `BlockRegion` use an `AABBi` object internally instead.
This would also make the `BlockRegion` interface independent from `AABBi`.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -0,0 +1,243 @@
...
+
+    public boolean testBlock(Vector3ic pos) {
+        return testBlock(pos.x(), pos.y(), pos.z());
+    }
+
+    public boolean testBlock(int x, int y, int z) {
+        return x >= minX && y >= minY && z >= minZ && x < maxX && y < maxY && z < maxZ;
+    }
+
+    public boolean isValid() {
```
kBlaszczyk, line 194: Why did you drop `isEmpty()`? Wouldn't it still be useful?

engine/src/main/java/org/terasology/world/block/BlockRegionIterable.java
```diff
@@ -0,0 +1,134 @@
...
+                }
+                current.set(next);
+                return next;
+            }
+        };
+    }
+
+    @Override
+    public void forEach(Consumer<? super Vector3ic> consumer) {
+
```
kBlaszczyk, line 111: Purposefully empty? Wouldn't it be better to throw something like "UnsupportedOperationException"?

engine/src/main/java/org/terasology/world/block/BlockRegion.java
pollend: a block is from the corner to +1 so the max range is increased by 1. then it also makes it a bit funny when you setMax because in this case your are specific just the max corner of the AABB region and not the block.

kBlaszczyk: Is the purpose of this to emulate the `createFromMinMax()` behavior of `Region3i`? (because that was similar)
I don't really get why the definition of a block being "an AABB with max + 1" is useful at all.

PR #3968, opened 2020-05-18
feat: Particle pool rendering

kBlaszczyk:
### Contains

This PR improves the rendering of particle pools.
Instead of rendering each particle individually, all particles in a particle pool are rendered at once.
The particle quads are generated on the GPU using a geometry shader. The rendering result should be practically identical to the previous implementation.

I added support for geometry shaders in `GLSLShader`. The class demanded version 1.2 shaders to be used. Since my implementation relies on OpenGL 3.3, I also added support for arbitrary shader versions. 1.2 is still used as the default in case no version is provided in a shader.
Note that the current implementation requires the version definition ("#version xxx") to be the very first line of the shader file.

I turned `ParticlePool` into a "drawable", which means it can now be rendered like a mesh. Note that this makes the class dependent on OpenGL.

### How to test

Destroy some blocks and verify that the block particles are being rendered as expected. You can also use the WeatherManager module and take a look at the weather particles.


engine/src/main/java/org/terasology/particles/ParticlePool.java
```diff
@@ -225,6 +240,74 @@ public void storeTemporaryDataAt(final int index, final int rawMask) {
         }
     }
 
+    /**
+     * Initializes this particle pool for rendering.
+     * The method should be called right after object creation.
+     */
+    public void initRendering() {
```
kBlaszczyk, line 247: The only reason for me to not do this automatically in the constructor is that the `glMethods` are breaking the unit tests, because they run without an OpenGL context.

engine/src/main/java/org/terasology/rendering/opengl/GLSLShader.java
```diff
@@ -257,23 +261,17 @@ private void registerAllShaderPermutations() {
         Set<Set<ShaderProgramFeature>> allPermutations = Sets.powerSet(availableFeatures);
 
         for (Set<ShaderProgramFeature> permutation : allPermutations) {
+            int featureHash = ShaderProgramFeature.getBitset(permutation);
+
             int fragShaderId = compileShader(GL20.GL_FRAGMENT_SHADER, permutation);
             int vertShaderId = compileShader(GL20.GL_VERTEX_SHADER, permutation);
-
-            if (compileSuccess(fragShaderId) && compileSuccess(vertShaderId)) {
```
kBlaszczyk, line 263: The compilation validation moved into `compileShader()`.

DarkWeird:
Looks great, pretty, fast.

Add new required capablities there (OpenGL33, etc):
[LwjglGraphics.java#L261](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/subsystem/lwjgl/LwjglGraphics.java#L261)
```java
 private void checkOpenGL() {
        boolean[] requiredCapabilities = {
                GLContext.getCapabilities().OpenGL12,
                GLContext.getCapabilities().OpenGL14,
                GLContext.getCapabilities().OpenGL15,
                GLContext.getCapabilities().OpenGL20,
                GLContext.getCapabilities().OpenGL21,   // needed as we use GLSL 1.20
.....
```

skaldarnar:
@kBlaszczyk @DarkWeird is anything of this influenced by the Render DAG or CoreRendering? If not, I guess this is good to be merged?

kBlaszczyk:
> @kBlaszczyk @DarkWeird is anything of this influenced by the Render DAG or CoreRendering? If not, I guess this is good to be merged?

DAG is high level rendering, while my changes concern low level rendering. I _think_ there isn't much, if any, interference.

skaldarnar:
I just tested this by merging latest latest `develop` in and starting a game. Particles are rendering fine when hitting blocks and stuff, so I'm merging this now :+1: 

Thanks for the work @kBlaszczyk :green_heart:

PR #3971, opened 2020-05-20
feat: disable transient dependencies

jdrueckert:
- missing direct dependencies on modules formerly added transiently by other dependencies will now result in compile errors

### DIFF
--- a/templates/build.gradle
+++ b/templates/build.gradle
@@ -111,7 +111,7 @@ configurations.all {
 }
 configurations {
     terasologyModule
-    compile.extendsFrom terasologyModule
+    implementation.extendsFrom terasologyModule
 }
 
 // Set dependencies. Note that the dependency information from module.txt is used for other Terasology modules


templates/build.gradle
```diff
@@ -111,7 +111,7 @@ configurations.all {
 }
 configurations {
     terasologyModule
-    compile.extendsFrom terasologyModule
+    implementation.extendsFrom terasologyModule
```
keturn, line 114: Is the `terasologyModule` configuration doing anything at this point?

I don't see any other references to it. What would happen if we drop it entirely and put the module dependencies straight on `implementation` instead of adding them one place and then extending from it?

https://github.com/MovingBlocks/Terasology/blob/bd5670f9299766551fcd14859c938058371444b5/templates/build.gradle#L124

DarkWeird: We can. This for easy involve special configurations for modules (a bit overengineered)

Cervator:
Oh, and I also am pushing a bump to version 3.3.0 for the engine (plus tests & BSG), as we somehow got the 3.2.1 into the `master` branch already somehow. We're well into minor release territory over patch anyway :-)

Also still forgetting the commit naming convention most the time, will try to do better!

Modules I tweaked as part of this (I realize some are essentially obsolete, but cleeeaann):

* AdditionalStructureTemplates
* AdditionalWorlds
* BlockDetector
* ChangingBlocks
* CoreBlocks - making PR
* CoreItems - making PR
* DynamicCities
* FluidComputerIntegration
* PlantPack
* ThroughoutTheAges
* TutorialBehaviors
* TutorialDynamicCities
* TutorialWorldGeneration

PR #3973, opened 2020-05-21
feat: Only show "augmentation" modules initially

skaldarnar:
This will only show modules from the "augmentation" category in the
advanced game setup to slim down the overwhelming list a bit.

- Game mode modules should be selected via the game play drop down even
  before entering detailed setup
- Game modes often come with their default world generator tailored for
  the game mode, thus fiddling with that is also an "advanced" use case
- Special modules are ... special
- Libraries only make sense in combination with other modules anyways
- Assets may be part of that list in addition to augmentation modules...

image


skaldarnar:
@stefaniamak FYI

PR #3974, opened 2020-05-21
Fixes for #3954

DarkWeird:
Fixes for #3954

* misc.xml to git ignore - saves Project JDK
* jarRepositories to git ignore - regenerates/generates by idea
* Choose shortenClasspath used by java8 and java9+


keturn:
Ah, I wondered if the repositories file was managed by gradle. If it is, and it updates as it syncs with gradle about which repositories the project uses, then yes, it is appropriate to ignore.

as for shortenClassPath: We can certainly set this back to "default" if that is necessary information. I thought it looked like a good setting to reduce the clutter in the run's output window. Is that information you need to refer to on a regular basis?

DarkWeird:
>as for shortenClassPath: We can certainly set this back to "default" if that is necessary information. I thought it looked like a good setting to reduce the clutter in the run's output window. Is that information you need to refer to on a regular basis?
I'm don't understand you question.

keturn:
We had [a lengthy chat about this in #stabilization](https://discord.com/channels/270264625419911192/696830442065625118/713123216071196752), I'll try to summarize here for the record:

1. We agreed it is okay for this IDEA configuration to use JDK 11. (related: #3975)
2. We'd like to avoid removing `misc.xml` entirely, because we'd lose the bit about annotation entry points. (I don't think https://github.com/JetBrains/gradle-idea-ext-plugin offers any clean alternative for setting those.)
3. the bellsoft liberica JDK isn't necessary for engine and the PC facade, they do not involve javafx. We'll change project-jdk-name from `liberica-11` to `11`, which is the default thing IDEA will suggest if it finds a system installation of openjdk-11.

I think we still have an outstanding concern that other plugins the developer may have installed will write things to other `component` sections of misc.xml?

If the things they write are suitable values for the project, I'm okay with adding those sections as they come. But if they frequently change, or the plugin is mistakenly storing stuff in there that's not really project related (e.g. should be workspace or ~/.config), or if someone else's IDEA removes any sections for plugins not _currently_ installed, then we really can't keep misc.xml directly in version control.

keturn:
As for the ["All in NameGenerator" run config](https://github.com/MovingBlocks/Terasology/pull/3974/commits/e958df4e7f5ac5c4d7959e1f840a250a3e304a58#diff-9e6dfc869b767f0ef1f380a8bb7046c0L2):

@Cervator, I know you said it's good to have an example here of how to do a run config for all the tests in a module, but it does look odd if you don't have NameGenerator installed 😄. Should we change this to BuilderSimpleGameplay instead?

or is BSG special in some way that means it would not be a good example? (because it is included in the repo)

Cervator:
@keturn I'm fine removing the NameGenerator test run config 👍 

BSG doesn't have any test, and very little actual code to test. It isn't really special, it is just a minimal example of running the game. I guess we could try to validate that it can load its dirt and grass blocks? Heh :-)

As for `misc.xml` there's precedent for the weirdness of first checking in a file, then separately `.gitignoring` it - at least in the past I think it _did_ work, even if it was kinda odd.

DarkWeird:
> We had [a lengthy chat about this in #stabilization](https://discord.com/channels/270264625419911192/696830442065625118/713123216071196752), I'll try to summarize here for the record:
> 
>     1. We agreed it is okay for this IDEA configuration to use JDK 11. (related: #3975)
> 
>     2. We'd like to avoid removing `misc.xml` entirely, because we'd lose the bit about annotation entry points. (I don't think https://github.com/JetBrains/gradle-idea-ext-plugin offers any clean alternative for setting those.)
> 
>     3. the bellsoft liberica JDK isn't necessary for engine and the PC facade, they do not involve javafx. We'll change project-jdk-name from `liberica-11` to `11`, which is the default thing IDEA will suggest if it finds a system installation of openjdk-11.
> 
> 
> I think we still have an outstanding concern that other plugins the developer may have installed will write things to other `component` sections of misc.xml?
> 
> If the things they write are suitable values for the project, I'm okay with adding those sections as they come. But if they frequently change, or the plugin is mistakenly storing stuff in there that's not really project related (e.g. should be workspace or ~/.config), or if someone else's IDEA removes any sections for plugins not _currently_ installed, then we really can't keep misc.xml directly in version control.

Arch Linux: IDEA's download jdk feature not work.
all JDKs in pattern "OpenJDK <Version>"

DarkWeird:
ehhhh.
Jetbrains have build instruction with "configure JDK" :(
https://github.com/JetBrains/intellij-community

.idea/.gitignore
```diff
@@ -18,6 +18,13 @@ workspace.xml
...
+/libraries
 
 # too much churn with teresology-modules each having their own repo
 /vcs.xml
+
+# generated by idea
+/jarRepositories.xml
+/uiDesigner.xml
+/modules.xml
+/modules
```
keturn, line 30: ah, yes, not surprised we need to exclude `modules/`, I think I remember needing that in a different project.

this one so far I haven't seen anything show up in there except for `.iml` files, which are already ignored. other things end up in there too?

.idea/misc.xml
```diff
@@ -12,5 +12,8 @@
     </writeAnnotations>
   </component>
   <component name="ExternalStorageConfigurationManager" enabled="true" />
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="liberica-11" project-jdk-type="JavaSDK" />
+  <component name="ProjectPlainTextFileTypeManager">
+    <file url="file://$PROJECT_DIR$/README.markdown" />
```
keturn, line 16: PlainTextFileType? Does this disable inspections and rendering for the README?

keturn: Yes, it seems to, if I restart after pulling this and open README, I no longer have proper Markdown support. :disappointed:

DarkWeird: yep.
I not noticed change...
I changed readme... IDEA throws error to me (Needs javafx's webview for rendering) and didn't want open file.
will fixed

.idea/.gitignore
DarkWeird: i don't remember what there on my system.
See that:
https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

PR #3771, opened 2019-10-20
[WIP] Add FullRotationFamily - placed blocks always "facing" the player

skaldarnar:

PR #3978, opened 2020-05-23
feat: Ceiling BlockFamily

kBlaszczyk:
This PR adds a new rotation family that causes blocks to always face the player, supporting upright and upside down block placement.

The family behaves similar to the `HorizontalFamily`, but puts blocks upside down in case they are being attached to the bottom side of another block. Its purpose is to be used for stairs and other things that need to be attached to the ceiling upside down. Unfortunately, it is currently impossible actually use this block family for stairs, since stairs are a block shape, and the engine doesn't support assigning block families to shapes. Addressing this problem is another issue though.

Implementing this block families behavior was previously not possible, due to limitations in the `BlockFamily` interface. Its `getBlockForPlacement()` method was supposed to determine the orientation of a new block based on the attachment side, and an additional "direction side".
These don't provide enough information to have a block face the player in every scenario, in which this would be the desired behavior. I fixed this by replacing the "direction side" parameter with a `Vector3f`, intended to be used for passing the player's viewing direction.
This gives the block family more information to determine the block orientation. The previously passed "direction side" can be derived from the viewing direction, doing so is now the block family's responsibility.

### Test
Assign the new block family to a furnace by adding the following line to the furnace.block json file.
`"family": "ceilingSupportingHorizontal",`

Add furnaces to your inventory using `give furnace 10` and attach them to different block sides. Verify they behave as described above.

### Notable
This PR is based on this one #3771, it's fixing misbehavior and is supposed to replace the other.

~~The 'Furnishings'-module is incompatible to this PR.~~
~~The fix can be found here: [#1](https://github.com/Terasology/Furnishings/pull/1).~~


engine/src/main/java/org/terasology/world/block/family/BlockFamily.java
```diff
@@ -42,12 +43,12 @@
...
-     * @param location            The location where the block is going to be placed.
-     * @param attachmentSide      The side of the block which this block is being attached to, e.g. Top if the block is being placed on the ground
-     * @param direction           A secondary direction after the attachment side that determines the facing of the block.
+     * @param location          The location where the block is going to be placed
+     * @param attachmentSide    The side of the block which this block is being attached to, e.g. Top if the block is being placed on the ground
+     * @param viewingDirection  The player's viewing direction
      * @return The appropriate block
      */
-    Block getBlockForPlacement(Vector3i location, Side attachmentSide, Side direction);
+    Block getBlockForPlacement(Vector3i location, Side attachmentSide, Vector3fc viewingDirection);
```
skaldarnar, line 51: At least the method signature changes, so we can spot breaking families easily. I don't see an easy way to make this work in a backwards compatible way, so let's do this step! 

I'll check the Omega workspace for a full list of breaking block families 🙈

engine/src/main/java/org/terasology/world/block/family/FreeformFamily.java
```diff
@@ -75,7 +79,10 @@ public FreeformFamily(BlockFamilyDefinition blockFamilyDefinition, BlockBuilderH
 
 
     @Override
-    public Block getBlockForPlacement(Vector3i location, Side attachmentSide, Side direction) {
+    public Block getBlockForPlacement(Vector3i location, Side attachmentSide, Vector3fc viewingDirection) {
+        Vector3f attachmentSideNormal = new org.joml.Vector3f(attachmentSide.direction());
+        Side direction = ChunkMath.getSecondaryPlacementDirection(JomlUtil.from(viewingDirection), JomlUtil.from(attachmentSideNormal));
```
skaldarnar, line 84: This is a bit more clunky than what you've done `HorizontalFamily`. What's the difference?

```java
Side secondaryDirection = Side.inDirection(-viewingDirection.x(), 0, -viewingDirection.z());
```

As I cannot spot any other changes around the `Side direction` in both cases, shouldn't those two be the same?

engine/src/main/java/org/terasology/math/Rotation.java
```diff
@@ -218,8 +218,15 @@ public int hashCode() {
...
      * @return All possible rotations for each yaw, pitch & roll combination.
+     *
+     * @deprecated use `allRotations` instead
      */
+    @Deprecated
     public static Iterable<Rotation> allValues() {
+        return allRotations();
+    }
+
+    public static Iterable<Rotation> allRotations() {
```
skaldarnar, line 229: What's this about - just better naming? Please move/copy the JavaDoc over to the non-deprecated method if you want to leave this change in.

kBlaszczyk: This is part of the old PR, it's not used or anything. I thought about removing it but since I'm aiming to replace the other PR I kept everything in. (if I'm not mistaken, it was you who committed this btw :)

engine/src/main/java/org/terasology/world/block/family/BlockFamily.java
kBlaszczyk: Alright, I'll open a seperate PR for the API change.

engine/src/main/java/org/terasology/world/block/family/FreeformFamily.java
kBlaszczyk: In `HorizontalFamily` I could improve and simplify the algorithm with the new concept. In this one I didn't want to touch the logic. So what you are seeing here is the exact same as it was originally, with the exception that `direction` is calculated within the method instead of being passed into it.

kBlaszczyk: Is the definition of the free form family to use the archetype or behave like the horizontal family?
I guess they could be using the same logic, but note that the calculation of `direction` works slightly different here than it currently does in `HorizontalFamily`.

engine/src/main/java/org/terasology/math/Rotation.java
skaldarnar: I must had a reason for that that is long forgotten now 🙄 🙈

kBlaszczyk:
@skaldarnar check out the API PR (#3987).
I'll continue with this one as soon as we got the other merged.

engine/src/main/java/org/terasology/world/block/BlockBuilderHelper.java
```diff
@@ -21,23 +21,101 @@
...
+ * A helper class for building and registering blocks with specific properties.
+ * Intended for use with Block Family related code.
+ * <p>
+ * The core features of this helper is to convert from definition data, into a Block
+ * and to register this block into the engine for use.
+ */
 public interface BlockBuilderHelper {
 
+    /**
+     * {@inheritDoc}
```
skaldarnar, line 34: What does this do? We are in an interface, where should we inherit from?

engine/src/main/java/org/terasology/math/Rotation.java
skaldarnar: Let's remove this change if not used...

engine/src/main/java/org/terasology/world/block/BlockBuilderHelper.java
```diff
@@ -21,23 +21,101 @@
...
+     *
+     * @see #constructSimpleBlock(BlockFamilyDefinition, BlockShape, String, BlockUri, BlockFamily)
+     */
     Block constructSimpleBlock(BlockFamilyDefinition definition, String section, BlockUri uri, BlockFamily blockFamily);
 
+    /**
+     * Constructs a basic block from the data provided. This block is assumed to have no rotations, and to use reasonable defaults
```
skaldarnar, line 55: What are "reasonable defaults"? What kind of defaults are we talking about/defaults for what?

kBlaszczyk: I don't know enough about block creation to answer this adequately. All kinds of things are setup using the parameters. I assume there are block properties independent of the specified parameters that would be initialized "using reasonable defaults". It's an interface method, therefore it has an abstract description. The claim is to produce a useful block without having to specify all the properties its gonna have.

engine/src/main/java/org/terasology/world/block/BlockBuilderHelper.java
kBlaszczyk: Nothing useful, I removed it.

PR #3983, opened 2020-05-24
fix: Respawn dead player after exiting the game

### LINKED ISSUE #3300: Player not dead after re-load
When the player dies and doesn't respawn before re-loading the game, they start alive but on 0 health.

### How to reproduce

* Load a world with Core enabled.
* Die.
* Exit to main menu without re-spawning.
* Re-load game.

cvennel:
### Contains

Fixes #3300

### How to test

- Load/create game with health module
- Enable hjump cheat
- Jump to your doom
- Choose "Exit Terasology"
- Re-open the game and load the same save
- Should now have full health instead of 0

NOTE: currently, trying to test using 'kill' in the cheat menu causes an OpenALException at "Setting listener orientation" when attempting to re-spawn. Use the high-jump or another method to cause player death.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
+++ b/engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
@@ -177,7 +177,7 @@ public class PlayerSystem extends BaseComponentSystem implements UpdateSubscribe
             characterComp.controller = entity;
             character.saveComponent(characterComp);
             character.setOwner(entity);
-            if (!character.hasComponent(AliveCharacterComponent.class)) {          
+            if (!character.hasComponent(AliveCharacterComponent.class)) {
                 respawnPlayer(entity);
             }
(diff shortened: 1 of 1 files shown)


cvennel:
Regarding the note about the kill cheat in the description, I opened issue #3984 for that.

engine/src/main/java/org/terasology/logic/players/PlayerSystem.java
```diff
@@ -177,8 +177,8 @@ private void restoreCharacter(EntityRef entity, EntityRef character) {
             characterComp.controller = entity;
             character.saveComponent(characterComp);
             character.setOwner(entity);
-            if (!character.hasComponent(AliveCharacterComponent.class)) {
-                character.addComponent(new AliveCharacterComponent());
+            if (!character.hasComponent(AliveCharacterComponent.class)) {
```
skaldarnar, line 180:
```suggestion
            if (!character.hasComponent(AliveCharacterComponent.class)) {
```

PR #3985, opened 2020-05-24
Update languages label

AlexHladin:
### Contains

This PR adds a few labels in Ukrainian and Russian translations.

### How to test


engine/src/main/resources/assets/i18n/menu_ru.lang
```diff
@@ -3,6 +3,7 @@
     "add-server": "Добавить",
     "add-server-tip": "Можно нажать ENTER, чтобы скопировать имя в поле адреса!",
     "advanced-usage": "Расширенное использование!",
+    "advanced": "Расширенный",
```
DarkWeird, line 6: I think better : "Расширенные". As part of "Расширенные настройки игры"(Advanced game settings)

engine/src/main/resources/assets/i18n/menu_ru.lang
```diff
@@ -97,6 +98,7 @@
     "error-downloading-server-list": "Ошибка при загрузке списка серверов!",
     "exit-game": "Выйти из Terasology",
     "experimental": "Экспериментальные коррективы",
+    "extras-label": "Дополнительные настройки",
```
DarkWeird, line 101: I suggest write there: "Дополнительно". Because there Record and Replay system, credits, reports, metrics.

engine/src/main/resources/assets/i18n/menu_ru.lang
AlexHladin: ok, I've fixed it

engine/src/main/resources/assets/i18n/menu_ru.lang
AlexHladin: ok, I've fixed it

PR #3991, opened 2020-05-25
build(templates): Add JUnit 5 params dependency

skaldarnar:
Found while working on https://github.com/Terasology/Inventory/pull/10,
which uses shiny JUnit 5 parameterized tests.

### DIFF
--- a/templates/build.gradle
+++ b/templates/build.gradle
@@ -138,6 +138,7 @@ dependencies {
     }
 
     testImplementation "org.junit.jupiter:junit-jupiter-api:5.6.2"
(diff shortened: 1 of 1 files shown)


templates/build.gradle
```diff
@@ -138,6 +138,7 @@ dependencies {
     }
 
     testImplementation "org.junit.jupiter:junit-jupiter-api:5.6.2"
+    testImplementation "org.junit.jupiter:junit-jupiter-params:5.5.2"
```
keturn, line 141: oh this is using 5.6 for the other junit-jupiter stuff, so this, but yes

```suggestion
    testImplementation "org.junit.jupiter:junit-jupiter-params:5.6.2"
```

PR #3535, opened 2018-10-29
Overhauled serialization system

eviltak:
# Contents
- Clean serialization API for use in the engine and in modules
- Add serialization support for various types previously unsupported
- Modularize serialization to make it easier to add custom type serialization
- Rework serialization to serialize objects via Gson or Protobuf by changing a single line of code (see `AbstractSerializer`, `GsonSerializer` and `ProtobufSerializer`)

Closes #3490.

# Review guide
This PR is finally ready to review and merge! Since there are a lot of changes to review, I'd recommend taking a look at the user-facing API introductions/changes first:

- [`AbstractSerializer`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-2941c45404523e22f6e2065994f11747)
- [`GsonSerializer`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-816e327c5e9a4f8b10204a0e0cde698e)
- [`ProtobufSerializer`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-277310030793629bd23b8b3e84e07a4c)
- [`TypeHandlerLibrary`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-76befd1046e3bd671dffb4e3994e53fb) (renamed from `TypeSerializationLibrary`)
- [`TypeHandler`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-ed98a03252f3f4c7750bbe08c2f14ace)
- [`TypeHandlerContext`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-497284d4c2d9a118ea2c83bddfbf0506)
- [`TypeHandlerFactory`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-9e490a02ea1ddf5cfab8db828bcfdb04)
- [`SerializedName`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-3a94a5171ecc346de026ea8247d2c617)
- [`TypeInfo`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-fdccd99cf7bf0a08d54808a83a9dd717)
- [`ReflectionUtil`](https://github.com/MovingBlocks/Terasology/pull/3535/files#diff-35714e56a06ddc78003aaa818ea3d9b1) -- will probably not be used outside the engine at all, but it's still worth a look to see if I missed documenting any public methods
[...]

PR #3992, opened 2020-05-25
test(TypeHandlerLibrary): RuntimeDelegatingTypeHandler test failure

keturn:
When running tests under JDK 11, there's one engine test that fails.

o.t.persistence.typeHandling.coreTypes.RuntimeDelegatingTypeHandlerTest#testDeserializeSub

The other four test cases in RuntimeDelegatingTypeHandlerTest continue to be fine.

There were "warning: raw use of parameterized type" messages all over that test file, in trying to understand the tests, the first thing I did was clean those up.

I haven't fixed the test yet, but I did get one of the other four to fail!

The place I've ended up is here: https://github.com/MovingBlocks/Terasology/blob/37debd9b59b0d7e62435a57d699504c96a33c0b7/engine/src/main/java/org/terasology/persistence/typeHandling/coreTypes/RuntimeDelegatingTypeHandler.java#L89-L94

that `th.getClass().equals(dh.getClass())` is coming back true for the test's baseTypeHandler and subTypeHandler, so it uses the baseTypeHandler (as that was the one given to RuntimeDelegatingTypeHandler construction), but the test is written specifically to assert the subTypeHandler is used.

So, did I
* uncover a bug in RuntimeDelegatingTypeHandler.serializeNonNull?
* accidentally change (i.e. _break_) the semantics of the test when I changed things around to not use unparameterized types?
* stumble on to some other type of distinction the test should be making? (e.g. does it need the subTypeHandler to be of a distinctly _different class_ than the baseTypeHandler, not just a different instance with different type parameters).

and why were the test results different between javas 8 and 11?

I'll run some things under java 8 to see if I can find out about that, but I need input on the rest.


keturn:
okay, now that test suite behaves the same way under JDK 8 and 11, which I appreciate.

The code under test looks like it was introduced in https://github.com/MovingBlocks/Terasology/pull/3535, if that rings any bells.

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="VarargParameter" enabled="false" level="INFORMATION" enabled_by_default="false" />
```
DarkWeird, line 4: You sure?

keturn: > Reports methods taking a variable number of arguments, also known as varargs methods. Such methods are not supported under Java 1.4 or earlier JVMs. The quickfix of this inspection replaces a variable argument parameter with the equivalent array parameter. Relevant arguments in calls to the method are wrapped in an array initializer expression. 

We don't have to worry about compat with Java 1.4, so I think the only other reason to have it enabled is so that quickfix-action of replace-variable-argument-parameter-with-array shows up when you have the cursor on varargs.

we can keep it at default if that's useful, but so far I've found it to be an unhelpful suggestion.

keturn:
EvilTak provided some comments about this today: https://discord.com/channels/270264625419911192/696830442065625118/762033934925561866

keturn:
That was a big merge because it has been a _while_ and all of the TypeHandlerLibrary code move, but it seems to still apply.

keturn:
@eviltak [wrote](https://discord.com/channels/270264625419911192/696830442065625118/762052742721568840):
>  If my hunch on how the JUnit Mockito extension works is right, then simply replacing the type of the  test class constructor parameter `@Mock TypeHandler<Sub> subTypeHandler` with `@Mock SubHandler subTypeHandler` should do the trick.

No such luck. It just makes the test fail a little earlier, `deserialize` not returning a value at all. https://github.com/MovingBlocks/Terasology/blob/caf972162cd2209cc82a33afe6c96f7de0262ad7/subsystems/TypeHandlerLibrary/src/test/java/org/terasology/persistence/typeHandling/coreTypes/RuntimeDelegatingTypeHandlerTest.java#L131-L133

keturn:
After stubbing out the `SerializationSandbox` so we're not relying on how this test case builds an instance of `Reflections`, and paying a bit more attention to whether we've stubbed the `Type` or `Class<>` variant of methods, I think we have tests that pass in both Java 8 and Java 11 builds. :relieved:

keturn:
[Java Version 11](https://trello.com/c/Gn9Fmcov/178-java-version-11)

Cervator (dismissed):
Seems fine here, CI finally passes, huzzah! :-)

Discord chatter seemed pretty confident about this, I'm not at that level but can at least confirm I looked at it 👍

DarkWeird (approved):
Looks right.
Works(i love tests)

keturn:
I can't say I'm _confident_ about this, relying as it does on mockito while involving type inspection. But it Works On My Machine, and on CI.

PR #3993, opened 2020-05-26
Convert the module build script to a precompiled plugin

keturn:
### Contains

This moves `templates/build.gradle` to gradle's `buildSrc/` directory where it becomes a [precompiled script plugin](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:precompiled_plugins).

Also, it is now a Kotlin gradle script instead of a Groovy gradle script.

### How to test

* Have a workspace with some module sources.
* `groovyw module refresh` to replace their `build.gradle` files.
* make sure modules build and run as usual.
* test the _publishing_ actions for the modules, the stuff configured by config/gradle/publish.gradle.  
  I'm not sure how to test that part and it's the most likely to have broken.

### Discussion

This introduces some new tech for the build process so I've tried to keep the scope of this small. The things under consideration here are:

* a module's build.gradle becomes as short as `plugins { id("terasology-module") }`
* starting to migrate the gradle configuration to Kotlin

I tried to transliterate the groovy config to kotlin *as directly as possible.* There's little here that cleans things up, and it's not a showcase of the best way to write something in Kotlin. I figured _changing languages_ is a big enough change without other reorganization, refactoring, or redesign on top.

GitHub's diff for the PR just shows a big delete in one file and add in another, but you can see more detail in this commit: https://github.com/MovingBlocks/Terasology/pull/3993/commits/0a352a72c887e16077f95b91c6f56937b06c4b90#diff-873dc4d71db70d570c73ef0cb16a8731

It has already demonstrated some practical benefits:

* Because the only thing the gradle file in the module subdirectory does is apply the plugin, there's no longer need to `groovyw module refresh` to update module build configs. Changing it in one place in buildSrc is enough.
* Because the script plugin is precompiled, the configuration phase of big multi-project workspaces is _a lot faster._ It only has to compile that mess once after it changes, not a hundred times over for the copies of it in a hundred module directories.


DarkWeird:
Is it work  with dependencies subtitles and generating dependencies on "fly" from module.txt?

keturn:
It _should._ It's all the same logic, and all that reading of module.txt and applying it to the dependencies should still happen for each sub-project where the plugin is applied.

buildSrc/src/main/kotlin/terasology-module.gradle.kts
```diff
@@ -16,89 +16,69 @@
...
+version = moduleConfig["version"] as String
 
 // Check for an outright -SNAPSHOT in the loaded version - for ease of use we want to get rid of that everywhere, so warn about it and remove for the variable
-def undesiredSnapshotTag = version.endsWith("-SNAPSHOT")
+val undesiredSnapshotTag = version.toString().endsWith("-SNAPSHOT")
 if (undesiredSnapshotTag) {
     println("Taking off undesired -SNAPSHOT")
-    version -= "-SNAPSHOT"
-    def out = services.get(StyledTextOutputFactory).create("an-ouput")
-    out.withStyle(Style.FailureHeader).println("WARNING: Module " + project.name + " is explicitly versioned as a snapshot in module.txt, please remove '-SNAPSHOT'")
```
keturn: Oh, so there is one feature I didn't preserve. The extra styling on the `please remove -SNAPSHOT` message. I was afraid it was using some groovy-specific formatter or something.

In hindsight, it looks like it's probably not groovy-specific? So I guess I probably could put it back.  
On the other hand, it is importing from an `.internal` package.

keturn:
There is [a bit of weirdness going on with IntelliJ IDEA](https://discuss.gradle.org/t/buildsrc-precompiled-kotlin-script-problems-in-intellij/36249?u=keturn) here, but I don't think it's a showstopper.

buildSrc/src/main/kotlin/terasology-module.gradle.kts
```diff
@@ -0,0 +1,246 @@
...
+
+// TODO: Remove when we don't need to rely on snapshots. Needed here for solo builds in Jenkins
+configurations.all {
+    resolutionStrategy.cacheChangingModulesFor(0, "seconds")
+}
+
+// Set dependencies. Note that the dependency information from module.txt is used for other Terasology modules
+dependencies {
+    // Check to see if this module is not the root Gradle project - if so we are in a multi-project workspace
+    "implementation"(group = "org.terasology.engine", name = "engine", version = "+") { isChanging = true }
```
DarkWeird, line 113: you have extra quotes :D

DarkWeird:
Fyi, my dirty benchmark (many side effects, side programs, and non clear workspace)

JoshariasSurvival(some dirty speed tests):
=========== Kotlin
First run `gradle-sync`  - 2m 17s (download kotlin 2 m, compile BuildSrc - 16s, invoke kotlin everywhere - 1s )
Then `jar` task - 43s ()
Then `jar` task again - 7s (recache reflection and reinvoke `jar`s)
Then `clean jar` - 20s(hmmm)
Resync modules: 4s

=========== Groovy
First run after checkout `gradle-sync` - 13s (only invoke everywhere)
Then `jar` task - 34s (seems, something cached, or not)
Then `jar` task again - 7s (recache reflection and reinvoke `jar`s)
Then `clean jar` - 23s(hmmm)
Resync modules: 5s



Module's gradle files looks pretty now. and DRY won! :D

buildSrc/src/main/kotlin/terasology-module.gradle.kts
keturn: Cerv asked about that too. I tried taking them out, and gradle seems just fine with that, but IDEA's gradle sync broke on me at that point. :disappointed:  I asked about it on the [gradle forum](https://discuss.gradle.org/t/buildsrc-precompiled-kotlin-script-problems-in-intellij/36249?u=keturn), I'll report back if I hear anything.

DarkWeird: hmm. It is worked for me :/

keturn: Oh, that's good to know.

I hit "Invalidate Caches and Restart" and after a rebuild and a bit that part works here too. :+1:

DarkWeird: We need hit this button often, after `.idea` PR :(

PR #3997, opened 2020-05-28
fix: prevent crash entering rendering settings with no renderers

keturn:
It crashed when going through the advanced game setup and trying to enter the render settings scene.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/rendering/RenderingModuleRegistry.java
+++ b/engine/src/main/java/org/terasology/engine/module/rendering/RenderingModuleRegistry.java
@@ -16,7 +16,6 @@
 package org.terasology.engine.module.rendering;
 
 import org.terasology.context.Context;
-import org.terasology.module.Module;
 import org.terasology.module.ModuleEnvironment;
 import org.terasology.module.sandbox.API;
 import org.terasology.naming.Name;
@@ -73,7 +72,6 @@ public class RenderingModuleRegistry {
      * For use everywhere to get last ordered list of {@link ModuleRendering} instances.
      * @return list of {@link ModuleRendering} instances
      */
-    @Nullable
     public List<ModuleRendering> getOrderedRenderingModules() {
         return this.orderedModuleRenderingInstances;
     }
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/RenderingModuleSettingScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/RenderingModuleSettingScreen.java
@@ -75,10 +75,9 @@ public class RenderingModuleSettingScreen extends CoreScreenLayer implements UIS
 
         renderingModuleRegistry = context.get(RenderingModuleRegistry.class);
 
-        renderingModuleRegistry.updateRenderingModulesOrder(moduleEnvironment, subContext);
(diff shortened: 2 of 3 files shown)


engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/RenderingModuleSettingScreen.java
```diff
@@ -76,6 +78,13 @@ public void postInit() {
         renderingModuleRegistry.updateRenderingModulesOrder(moduleEnvironment, subContext);
         orderedModuleRenderingInstances = renderingModuleRegistry.getOrderedRenderingModules();
 
+        if (orderedModuleRenderingInstances == null || orderedModuleRenderingInstances.isEmpty()) {
```
DarkWeird, line 81: make  transformation "`null` -> emptyList" in `org/terasology/engine/module/rendering/RenderingModuleRegistry.java:121` instead checking on null.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/RenderingModuleSettingScreen.java
```diff
@@ -76,6 +78,13 @@ public void postInit() {
         renderingModuleRegistry.updateRenderingModulesOrder(moduleEnvironment, subContext);
```
DarkWeird, line 78: `updateRenderingModuleOrder` returns `orderdRenderingModules`
Use it directly

engine/src/main/java/org/terasology/rendering/world/WorldRendererImpl.java
```diff
@@ -216,7 +210,7 @@ private void initRenderingModules() {
             renderingModuleRegistry.updateRenderingModulesOrder(context.get(ModuleManager.class).getEnvironment(), context);
```
DarkWeird, line 210: use it directly there. (via variable)

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/videoSettings/RenderingModuleSettingScreen.java
keturn: Odd. The only reason I was getting null-warnings on that site was because this method is explicitly annotated as `@Nullable`

https://github.com/MovingBlocks/Terasology/blob/73b04de01c6d543a897cd54f12e7fa6dcfa092b7/engine/src/main/java/org/terasology/engine/module/rendering/RenderingModuleRegistry.java#L75-L76

but I don't think it is? that field has a default value, is only set in one other place that looks like it's non-nullable, and is private so I don't think anything else can null it.

skaldarnar: why would you ever return null if you could return an empty list? o.O

engine/src/main/java/org/terasology/engine/module/rendering/RenderingModuleRegistry.java
```diff
@@ -73,7 +72,6 @@ public ModuleRendering getModuleRenderingByModuleId(Name renderingModuleId) {
      * For use everywhere to get last ordered list of {@link ModuleRendering} instances.
      * @return list of {@link ModuleRendering} instances
      */
-    @Nullable
```
skaldarnar, line 76: :+1:

Issue #3996, opened 2020-05-28
deadlocked threads won't die after crash from RenderingModulesSettingScreen


keturn:
When crashing from RenderingModulesSettingScreen (crash itself to be filed in another issue), a bunch of threads don't want to go away and let the game close in peace.

Non-daemon threads include:

* Pathfinder-0
* Workboard-0
* Saving-0 — A thread calling "Saving" is maybe one we actually _do_ avoid killing? but I don't know what would be saving anything this early.
* Chunk-Generator-* 
* Chunk-Unloader-*
* Chunk-Updater-*

Additionally, this is a daemon thread, but it's waiting for something that seems like it's not coming:

```
"Thread-3" #43 daemon prio=10 os_prio=0 cpu=0.64ms elapsed=851.89s tid=0x00007f8f1a9a9800 nid=0x247c5b in Object.wait()  [0x00007f8e1a5f8000]
   java.lang.Thread.State: WAITING (on object monitor)
        at java.lang.Object.wait(java.base@11.0.7/Native Method)
        - waiting on <0x00000000a174bb78> (a net.java.games.input.LinuxDeviceThread)
        at java.lang.Object.wait(java.base@11.0.7/Object.java:328)
        at net.java.games.input.LinuxDeviceThread.run(LinuxDeviceThread.java:60)
        - waiting to re-lock in wait() <0x00000000a174bb78> (a net.java.games.input.LinuxDeviceThread)
```

see also #3953 for other non-daemon-thread talk.

Where we were when crashing:

```
	at org.terasology.rendering.nui.layers.mainMenu.videoSettings.RenderingModuleSettingScreen.postInit(RenderingModuleSettingScreen.java:112)
	at org.terasology.rendering.nui.layers.mainMenu.StartPlayingScreen.lambda$initialise$3(StartPlayingScreen.java:111)
	at org.terasology.rendering.nui.ActivatableWidget.activateWidget(ActivatableWidget.java:46)
```

### Log details and game version

this is just after the DAG branch landed in develop.

PR #3998, opened 2020-05-28
fix(concurrency): make TaskProcessor threads daemon-threads by default

keturn:
This makes all of TaskMaster's threads (the TaskProcessors) be daemon threads by default.

Daemon threads can not prevent the application from shutting down.

Related to deadlock-on-exit issues such as #3996. (All the non-daemon threads named in that issue are TaskProcessor threads.)

This is a naive patch, it applies this setting in the place unconditionally. I hope someone reviews this and explains which uses we need consider before making such a global change!

### DIFF
--- a/engine/src/main/java/org/terasology/utilities/concurrency/TaskProcessor.java
+++ b/engine/src/main/java/org/terasology/utilities/concurrency/TaskProcessor.java
@@ -1,11 +1,11 @@
 /*
- * Copyright 2013 MovingBlocks
+ * Copyright 2020 MovingBlocks
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *     https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -43,6 +43,7 @@ final class TaskProcessor<T extends Task> implements Runnable {
         boolean running = true;
         Thread.currentThread().setPriority(Thread.MIN_PRIORITY);
         Thread.currentThread().setName(name);
+        Thread.currentThread().setDaemon(true);
         while (running) {
             try {
                 T task = queue.take();


keturn:
Test failures? Is that for real @GooeyHub, or were you having a bad night? I can't see the details you've linked.

keturn:
oh, yup, those are real errors.

> This method must be invoked before the thread is started.

okay, fine, I can do it when the TaskMaster sets up its thread pool instead. Still much the same question for review, though: is there anything to consider before making that change to all TaskMaster's threads?

keturn:
the question I'm tiptoeing around here is: if I spend time on TaskManager, should I spend that time changing TaskManager, or should I deprecate the entire interface and replace its usages in engine to hold ExecutorService objects instead?

on the assumption that ExecutorService is a java standard library thing so it'll be one less unique-to-Terasology interface to learn.

Cervator:
Going to ping @casals on this one - as far as I can tell solely Pathfinding, including its Flexible counterpart, along with Behaviors use this. Admittedly it probably predates his presence but he might still have thoughts on it :-)

@kaen could be another option, particularly for the Flexible part, but may be harder to reach. Also: huh, didn't have him in the org here yet, just the Terasology org. So may not ping - same name on Discord IIRC.. Invite sent 👍

casals:
We're having some performance problems with Behaviors here and there, but I didn't notice anything deadlocked (I was not looking specifically for it though). Regardless - daemon-ing all threads shouldn't really have an impact in Behaviors. I tested the PR locally with some of the Behaviors I've been using, no visible impact (expected, given how Behaviors use the Task interface). From the PF perspective: *perhaps* something from the Flexible navgraph processing, but to little to go on at the moment (both in things using Flexible* and viable test cases). All things considered, it should be safe to give it a try/see if anything comes up in the next multiplayer session (but I doubt so).

keturn:
I'm quite surprised this worked for you in its current form! The errors that came up from the unit tests are credible: `TaskProcessor.run` is too late to be calling `thread.setDaemon()`.

The open question here, from https://github.com/MovingBlocks/Terasology/pull/3998#issuecomment-636365300 , is not whether daemon=true on these threads will have any run-time performance impact on Pathfinding or Behaviors.

The question is whether anything (including Pathfinding, but there are a number of other usages in critical engine classes) needs this particular implementation of a thread pool. If not, we can trim down the amount of code we're maintaining and the number of interfaces contributors need to learn by switching them to something closer to the standard ExecutorService interface.

casals:
> I'm quite surprised this worked for you in its current form! 
Hm. I did it in a rush, but I'm quite sure the PR was downloaded. Perhaps the old classes were cached.

> The question is whether anything (including Pathfinding, but there are a number of other usages in critical engine classes) needs this particular implementation of a thread pool.
Ah, OK. Short answer is "yes" - both referenced modules (Behaviors/PF) use the current thread pool implementation (`TaskMaster`) with the associated interface (`Task`).

keturn:
I can see that references to TaskMaster exist in the codebase.

What I'm seeking to learn is whether TaskMaster has any distinctive behavior that makes it an improvement over ExecutorService for our use cases.

casals:
Although this code predates me for a long time, when I look at `TaskMaster` what I see is "prioritization made easier through wrapping". That's the immediate benefit you're looking for. I get your point on maintaining only necessary code and sticking to standard libs, but if we consider that (i) not all of our contributors (by far) have considerable experience with multithreading; and (ii) it's been stable for almost 3y since the last commit, I'd say that it's worth keeping it - considering its size versus what it's wrapping, I don't really see the "extra learning" as a problem.

keturn:
Ah, prioritized tasks, that's a good use case for me to look at in more detail. I agree it's less obvious how to do that with the standard library than the FIFO queue.

I also agree “if it ain't broke, don't fix it” is a good guideline. I'm afraid I've muddied my own argument by bringing up TaskMaster's _interface_ (saying it's an extra thing to document and learn) while the _implementation_ is the real source of my confusion and concern.

To recap: I wanted to make a change to the threads TaskMaster uses, so

* I added a line to TaskProcessor.run. That seemed like the obvious place for it, as it's where other properties of the thread are configured.
* that turned out to _not_ be the right place. This is really no fault of TaskMaster/TaskProcessor, I simply hadn't read the Thread.setDaemon docs closely enough. It says quite plainly “This method must be invoked before the thread is started.”
* that led me to look at TaskMaster's implementation in more detail, searching for the place where it starts threads: https://github.com/MovingBlocks/Terasology/blob/b9001dabdc2ced69a18db6cb33e0ae2d9d22bedd/engine/src/main/java/org/terasology/utilities/concurrency/TaskMaster.java#L153-L156
* which has me asking _why is it like this?_ why is it immediately giving its executorService `threads` number of tasks? it seems to be deliberately sidestepping much of the functionality offered by the implementations of executors and thread pools in the standard library.
* I don't know why, and I can't figure it out from the documentation or tests provided.

The one-line addition that was the original topic of this PR could _still_ be a one-line addition, but before building on top of something I'm worried about the maintainability of, I wanted to confirm I can't deprecate & remove the thing instead.

In this environment, I don't find that the age of code is an endorsement of its fitness. I put a lot of work in to a PR on another topic where I was _very careful_ to keep compatibility during my refactoring, only to find in the end that there was no _live_ code depending on it at all.

Following your feedback, I've done a first-pass reading of the priority methods of the things that use TaskMaster with priority queues: Behavior and a couple things in terasology.world.chunks. That's taught me that there are, in fact, things that make deliberate use of the prioritizing feature.

But the difference between an executor that works with a priority queue as compared to a FIFO queue is a one-line static method of TaskMaster providing constructor arguments. I don't feel closer to understanding why TaskMaster isn't _only_ those static utility methods.

casals:
Yeah, it took me some time to get the argument on interface vs implementation, sorry. Thanks for clearing that up. :)

Following your recap/questions:

- About your question on why the implementation uses a fixed number of tasks - I can say for certain that I have absolutely no clue.
- (We do have some issues with maintained code versus live dependencies, so yeah, probably age is not the best argument here)
- I understand why prioritization is needed in Behaviors/PathFinding (and I can see how the wrapper might be helpful), but I also can't pitch on why it's built like this. 

All things considered - as long as support to prioritization is there (which we know is necessary), I don't see why not try changing the implementation. Any live and non-documented limitations (such as fixed thread count) would be identified by things breaking - and since we have an ongoing stabilization effort before the first stable release, I think it's more than justified. We can use Behaviors/Pathfinding as the most critical testing cases and see what happens with everything else.

keturn:
I think we are close to obsoleting this PR. We're expecting to remove the TaskManager class soon, and TaskProcessor probably leaves with it?

PR #3999, opened 2020-05-29
chore: set the default copyright header to something shorter

keturn:
Apache says this is a "shorter variant you may use."

https://www.apache.org/foundation/license-faq.html#Apply-My-Software

this PR makes no attempt to change _all_ the existing copyright headers, but does do a few of the high-level files to set an example.

### DIFF
--- a/.idea/copyright/TerasologyEngine_brief.xml
+++ b/.idea/copyright/TerasologyEngine_brief.xml
@@ -1,0 +1,8 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="allowReplaceRegexp" value="20\d\d" />
+    <option name="keyword" value="Copyright.*Moving\s?Blocks" />
+    <option name="notice" value="Copyright &amp;#36;today.year MovingBlocks&#10;SPDX-License-Identifier: Apache-2.0" />
+    <option name="myName" value="TerasologyEngine-brief" />
+  </copyright>
+</component>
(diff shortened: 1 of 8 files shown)


.idea/copyright/TerasologyEngine_brief.xml
```diff
@@ -0,0 +1,8 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="allowReplaceRegexp" value="20\d\d" />
+    <option name="keyword" value="Copyright.*Moving\s?Blocks" />
```
skaldarnar, line 4: Do we want to go for "The Terasology Foundation" instead of "Moving Blocks"? @Cervator ?

Cervator: Yes please - good catch 👍 

That's been a global change far down the todo list for a while now

skaldarnar:
> Even though I know the objective is to make it short I kinda wish there'd be a tiny link in there to explain what "SPDX-License..." refers to. Should we have that explained in a readme somewhere?

I think it's fine to only have the short form in the source files - a developer should be able to type ["SPDX license header"](https://is.gd/nwi4gQ) into their search engine of choice. 
First entry is https://spdx.org/licenses/ to which we _could_ link from the README, if we had a license section ^^ (see launcher has https://github.com/MovingBlocks/TerasologyLauncher#license- - that's where I would put it, if anywhere).

keturn:
We also have `/LICENSE`, and GitHub does recognize that and point its "View License" link there.

keturn:
oh, I missed something by only updating copyright headers in gradle files: Checkstyle validates headers in java files. https://github.com/MovingBlocks/TeraConfig/issues/11

PR #4004, opened 2020-05-30
fix: gracefully catch and report errors that happen during game load

keturn:
This catches errors that occur during game load.

It opens the crash reporter while sending the user back to the Terasology main menu.

### How to test

Have a crashy save file or something?

If not, you can edit the source someplace like [AwaitCharacterSpawn#step](https://api.codestream.com/p/Xt2iViTQ7y7YLbE4/_cidawS0RHyZPrtYxfaT0Q) to deliberately throw an exception.

- [x] Is this the right place to catch the error? I don't think it's the _worst_ place, as StateLoading could know that switching back to StateMainMenu is a reasonable fallback. But maybe it should be higher. wrap all of StateLoading.update? or all of GameEngine.tick?
- [x] Will need translation work ← this is one of the template checklist items but is totally true. 
  - [ ] ~Use a translatable resource for the error message here.~ needs compound messages; will address in #4045.
  - [x] Make sure the thing that shows the message uses the translation resources.


keturn:
A screenshot showing the untranslated error message with its template variable.

engine/src/main/java/org/terasology/engine/modes/StateLoading.java
```diff
@@ -243,8 +245,17 @@ public void update(float delta) {
...
-            if (current.step()) {
-                popStep();
+            try {
+                if (current.step()) {
+                    popStep();
+                }
+            } catch (Exception e) {
+                logger.error("Error while loading {}", current, e);
+                String errorMessage = String.format("Failed to load game. There was an error during %s.",
+                        current == null ? "the last part" : current.getMessage());
```
skaldarnar, line 255: If `current.getMessage()` is a key for a translateable string, you probably "just" need to push that through `TranslationSystem::translate`.

In screen layers and widget you'll get it by using `@In` annotation:

```java
    @In
    private TranslationSystem translationSystem;
```

keturn:
⚠ this error-handling PR is turning into a i18n-feature PR

since the error-handling remains small in line-count and you've approved its approach, the scope creep isn't going to be too hard to review. but they're definitely different subjects.

keturn:
hmm, yeah, that's not a great idea. Too much controversial stuff in i18n branch. I'll look for a branch-split place that leaves this error-handling thing in its partially-translated state, before it gets to the stuff that changes public APIs.

keturn:
Okay. The message is now _half_-translated. That is, the previously-translated string describing the loading stage should be shown for the current locale instead of `${blah#awaiting-chacater-spawn}`.

The _new_ error text is stuck being hardcoded. To get the whole thing together requires compound messages, which as far as I can tell, org.terasology.i18n doesn't currently support. That is too much to be in scope here and will be the subject of another issue.

engine/src/main/java/org/terasology/engine/modes/StateLoading.java
keturn: Okay, I've made it so the part of StateMainMenu that shows these messages will run them through the translation system.

PR #4005, opened 2020-05-30
chore: more descriptive error message on spawn failure

keturn:
This changes an error message from being a NoSuchElementException to

> java.lang.RuntimeException: Failed to find an acceptable spawn location.

is RuntimeException the right thing to use for this?

NoSuchElementException is a direct subclass of RuntimeException so it's probably not very wrong, but maybe we have a more fitting type for this kind of thing.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
+++ b/engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
@@ -93,7 +93,9 @@ public class PlayerFactory {
         EntityBuilder builder = entityManager.newBuilder("engine:player");
         float extraSpace = 0.5f;  // spawn a little bit above the ground
         float entityHeight = getHeightOf(builder) + extraSpace;
-        return findSpawnPos(JomlUtil.from(locationComponent.getWorldPosition()), entityHeight).get(); // TODO: Handle Optional being empty
+        return findSpawnPos(JomlUtil.from(locationComponent.getWorldPosition()), entityHeight)
+                // TODO: Handle Optional being empty
+                .orElseThrow(() -> new RuntimeException("Failed to find an acceptable spawn location."));
     }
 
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
```diff
@@ -93,7 +93,9 @@ public Vector3f findSpawnPositionFromLocationComponent(LocationComponent locatio
         EntityBuilder builder = entityManager.newBuilder("engine:player");
         float extraSpace = 0.5f;  // spawn a little bit above the ground
         float entityHeight = getHeightOf(builder) + extraSpace;
-        return findSpawnPos(JomlUtil.from(locationComponent.getWorldPosition()), entityHeight).get(); // TODO: Handle Optional being empty
+        return findSpawnPos(JomlUtil.from(locationComponent.getWorldPosition()), entityHeight)
```
skaldarnar, line 96: I think the "right" fix for this is to change the method signature to return an `Optional<Vector3f>`. Imho there is nothing _exceptional_ about a `find...` method not finding something. 
An alternative which leaves the signature as is would be to return `null` instead (another way of encoding the no spawn position was found).

I like the `Optional` better as it forces the calling site to react on the possibility of failure.

I don't think that a plain `RuntimeException` is the right thing to use here, as it is way too generic, and not really necessary. If we'd go for "exception-based programming" we'd introduce a `NoSpawnPositionFoundException` to catch further up the chain...

keturn: Oh, this is by no means a _fix._

this is a patch to a known-broken method to produce a more recognizable failure message when it breaks.

skaldarnar: Right, the TODO is still there 😅 Baby step improvements ^^

PR #4006, opened 2020-05-30
refactor(BlockItemFactory): remove code dupes; add doc; retain components

skaldarnar:
Reduces code duplication in `BlockItemFactory` and unifies the way block items are created when using `BlockItemFactory::newInstance` and `BlockItemFactory::newBuilder`.

Block item creation now also takes `RetainComponentsComponent` into account when copying components from a reference entity or block prefab to the block item. 

📘 Also, full addition of JavaDoc for `BlockItemFactory`.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/items/BlockItemFactory.java
+++ b/engine/src/main/java/org/terasology/world/block/items/BlockItemFactory.java
@@ -16,18 +16,19 @@
 package org.terasology.world.block.items;
 
 import org.terasology.entitySystem.Component;
+import org.terasology.entitySystem.ComponentContainer;
 import org.terasology.entitySystem.entity.EntityBuilder;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
-import org.terasology.entitySystem.prefab.Prefab;
 import org.terasology.logic.common.DisplayNameComponent;
 import org.terasology.logic.inventory.ItemComponent;
 import org.terasology.rendering.logic.LightComponent;
 import org.terasology.world.block.family.BlockFamily;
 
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/world/block/items/BlockItemFactory.java
```diff
@@ -85,37 +98,145 @@ public EntityRef newInstance(BlockFamily blockFamily, int quantity) {
...
+     * @param quantity item quantity (see {@link ItemComponent#stackCount}); constrained to [0...128)
+     * @return a pre-populated entity builder for a block item entity
+     */
+    public EntityBuilder newBuilder(BlockFamily blockFamily, int quantity) {
+        final ComponentContainer components =
+                blockFamily.getArchetypeBlock().getPrefab()
+                        .map(p -> ((ComponentContainer) p))
+                        .orElse(EntityRef.NULL);
+
+        return createBuilder(blockFamily, components, (byte) Ints.constrainToRange(quantity, 0, Byte.MAX_VALUE));
```
keturn, line 139: I feel like there should be a shorter form of that conversion. Oh, here we go:

https://guava.dev/releases/snapshot/api/docs/com/google/common/primitives/SignedBytes.html#saturatedCast-long-

`signedBytes.saturatedCast(quantity)`

skaldarnar:
> I'm trying not to think too much about why that method takes byte as a quantity, but this looks good. Yay docs 👍

For some reason the item quantity in `ItemComponent` is stored as byte. Previously, we slurred the int all the way down and just put it there - if you've created a new instance with a requested quantity of 128 you'd get some negative amount instead - yay for overflows 🎊 🤦

PR #4008, opened 2020-05-31
Update Groovy Wrapper, make Java 8 warning milder, and add `groovyw module init`

### LINKED ISSUE #3975: update build.gradle "compiling with a JDK newer than" warning for jdk11
We had [a talk about jdk versions](https://discord.com/channels/270264625419911192/696830442065625118/713123216071196752) today and I think we agreed it is fine to use jdk 11 for development.

Note this is _different_ than saying Terasology engine will start using Java 11 _language_ features; for the time being source compatibility remains at Java 8. This is just so the build environment won't scold people for using the current LTS release of OpenJDK.

it would have been faster to write a pull request than type up this issue if not for two things:

1. it needs documenting anyway ;)
2. the way it's currently written, it checks to see if the value is _exactly_ `JavaVersion.VERSION_1_8`, and we want the new check to allow anything 8 – 11, and that would have taken me several moments to look up.

Cervator:
Bit of a rush trying to fit all this in available time, so could probably use a quick round of cleanup / polish. Did remember to use the fancy commit naming at least! In short:

* Fixes #3975 (maybe?) by making the Java 8 check more mild. Still tied to Java 8 specifically but suggests 9-11 will be OK and just try 8 if anything seems weird
* Further improves @keturn's great effort in getting the Groovy Wrapper working with Gradle 6.4 (see https://github.com/MovingBlocks/groovy-wrapper/pull/4 and the one commit I added), now formally released as the Groovy Wrapper v2.0.0
* Adds `groovyw module init` or maybe more interestingly `groovyw module init iota|omega` which will fetch a named module set based on the Index, or default to just CoreSampleGameplay.
  * Not sure if the default is really worth it since it is the exact same as `groovyw module get CoreSampleGameplay` but I couldn't think of what else to make it do 🤔 Probably defaulting to Iota is _too_ much? Yet it doesn't do much else. Would be nice if it was just `groovyw init` and/or did more stuff

Haven't tested on different systems yet, so that'd be nice. To test:

* Make sure if in a workspace with *no* source modules that a warning gets printed in red, and if using something newer than Java 8 that a different but milder note gets logged
* Make sure that the `groovyw` command will execute even on a *clean* system (no cached Gradle wrapper files - first update required `gradlew` to run at least once before `groovyw`)
* See if `groovyw module init` or `groovyw module init iota` does its thing. Or even Omega but that may take a little while!

Thanks bunches to @keturn for sorting out the Groovy Wrapper issue in the first place, I just made it more formal.

On a related note the game handles the total lack of modules decently. Found one small UI thing that could be improved, will submit issue tomorrow

Outstanding: Update docs to mention the `init` task and the need to run something to get modules before starting dev work (at least in most cases)

Curious: Since I'm often short on time I often end up producing low quality code when I *do* manage to throw something together. If it is in need of relatively simple (but possibly slightly time consuming due to completeness) refactoring for quality reasons, would it be fair to immediately submit a quick GFI issue to get that handled later / give some junior types an opening to do something relatively easy? 🤔


config/groovy/util.groovy
```diff
@@ -220,6 +220,36 @@ switch (cleanerArgs[0]) {
...
+            if (cleanerArgs.length == 1) {
+                println "Doing a default module init - simply fetching CoreSampleGameplay without dependencies in source (same as `groovyw module get CoreSampleGameplay`)"
+                String[] targetMods = ["CoreSampleGameplay"]
+                common.retrieve(targetMods, false)
+            } else if (cleanerArgs.length == 2) {
+                def targetModuleDistro = cleanerArgs[1]
+                println "Doing an init of a given module distro: " + targetModuleDistro
+                def targetDistroURL = "https://raw.githubusercontent.com/Terasology/Index/master/distros/" + targetModuleDistro + "/gradle.properties"
+                if (!common.isUrlValid(targetDistroURL)) {
+                    println "Distro URL $targetDistroURL does not seem to be valid, typo? See https://github.com/Terasology/Index/tree/master/distros for available distros"
```
skaldarnar, line 235: If something goes wrong, this prints a very verobse message (imho)

> Distro URL https://raw.githubusercontent.com/Terasology/Index/master/distros/jota/gradle.properties does not seem to be valid, typo? See https://github.com/Terasology/Index/tree/master/distros for available distros


I would slim this down and split it up a bit:

```suggestion
                    println "[init] Invalid distribution name: '$targetModuleDistro'"
                    println "[init]     See https://github.com/Terasology/Index/tree/master/distros for available distributions
```

As follow-up improvement, we could just fetch the list of supported distros dynamically:

> ```
> [init] Invalid distribution name: "jota"
> [init]     Available distributions: [awt, iota, omega]
> ```

config/groovy/util.groovy
```diff
@@ -220,6 +220,36 @@ switch (cleanerArgs[0]) {
...
+                    break
+                }
+                String distroContent = new URL(targetDistroURL).text
+                String moduleSnippet = "extraModules="
+                int someIndex = distroContent.indexOf(moduleSnippet)
+                if (someIndex != -1) {
+                    moduleLine = distroContent.substring((someIndex + moduleSnippet.length()), distroContent.indexOf("\n", someIndex))
+                    common.retrieve(moduleLine.split(","), false)
+                } else {
+                    println "Found the distro definition, but it seemed invalid - did not contain snippet: $moduleSnippet"
```
skaldarnar, line 245:
```suggestion
                    println "[init] ERROR: Distribution does not contain key: '$moduleSnippet'"
```

Possibly also add where to report errors or target PRs against:

```groovy
                    println "[init]     Report an issue: https://github.com/Terasology/Index/issues"
```

config/groovy/util.groovy
```diff
@@ -220,6 +220,36 @@ switch (cleanerArgs[0]) {
...
+                }
+            } else {
+                println "Too many arguments! Init only does a single given module distro. Use `get` or `recurse` for multiple arbitrary modules. See `groovyw usage`"
```
skaldarnar, line 248:
```suggestion
                println "[init] Too many arguments! Usage: 'grooyw module init [distribution]'"
                println "[init]     See `groovyw usage` for more information."
```

build.gradle
```diff
@@ -64,7 +64,7 @@ import static org.gradle.internal.logging.text.StyledTextOutput.Style
 // Check for Java 8
 if(JavaVersion.current() != JavaVersion.VERSION_1_8){
     def out = services.get(StyledTextOutputFactory).create("an-ouput")
-    out.withStyle(Style.FailureHeader).println("WARNING: You're compiling with a JDK newer than Java 8, this may not be fully supported for Terasology, consider adjusting your JAVA_HOME accordingly")
+    out.withStyle(Style.FailureHeader).println("NOTE: You're compiling with a JDK newer than 8. While we believe Java 9-11 to be safe to use, newer versions than that may cause issues. If you encounter any oddities try to see if they persist with Java 8")
```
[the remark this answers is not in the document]
keturn: Yeah, we're setting that compatibility target in all the java projects. Sometimes through indirect reference to `common.gradle`, but it should still be true.

The one documented case of a java 11 failure we've found is that thing we had with Ubuntu's openjdk distribution the other day. If we do keep a warning like this (and I'm not sure we should; it will be a false positive for the many working java 11 builds), it should have a link to our [troubleshooting page](https://github.com/MovingBlocks/Terasology/wiki/Troubleshooting-Developer#java).

keturn:
:thinking: the java 11 error we know about is a _run-time error,_ not a build-time error. So we could put it in the game log like we do with the https://github.com/MovingBlocks/Terasology/blob/v3.2.0/engine/src/main/java/org/terasology/engine/subsystem/lwjgl/LwjglGraphics.java#L193

the confounding thing, though, is that those lines might end up being in the same place, since lwjgl is both the thing that will crash with video driver problems _and_ the first library we load that we know has problems with ubuntu-openjdk's dl-lookup. But maybe there's a moment between loading the library and initializing the video stuff that would let us distinguish between the two cases?

build.gradle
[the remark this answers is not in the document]
Cervator: Updated 👍 Screenie on Discord. Went with a Java > 11 warning

config/groovy/util.groovy
```diff
@@ -220,6 +220,36 @@ switch (cleanerArgs[0]) {
 
         break
 
+        // Makes sure a workspace has one or more modules matching user input, most useful for grabbing whole "module distros"
+        case "init":
+            // TODO: Move most this code into module.groovy, leave an error here if you use it for something else
+            if (cleanerArgs.length == 1) {
+                println "Doing a default module init - simply fetching CoreSampleGameplay without dependencies in source (same as `groovyw module get CoreSampleGameplay`)"
```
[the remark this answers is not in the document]
Cervator: Applied all the logging improvements, thanks for the suggestions! Those were drafts at best :)

As for `get` vs `recurse` I wanted to keep out unneeded source unless the dev wants it - really the docs could just say to use `groovyw module get|recurse X` as desired. `groovy module init` would otherwise be about the same as init iota?

In any case I'm still not sure what to really think about the default init vs just the distros

Cervator:
Okay, updated, may be ready 👍

build.gradle
```diff
@@ -62,9 +62,9 @@ import org.gradle.internal.logging.text.StyledTextOutputFactory
 import static org.gradle.internal.logging.text.StyledTextOutput.Style
 
 // Check for Java 8
-if(JavaVersion.current() != JavaVersion.VERSION_1_8){
+if(JavaVersion.current() != JavaVersion.VERSION_11){
```
skaldarnar, line 65: Aren't Java 1.8 and Java 9 `!= JavaVersion.VERSION_11`? 🧐 Did you test that with Java 8 and 9?

keturn: it's probably fine — maybe even better — to support 8 **or** 11, instead of 8–11.

9 and 10 weren't long-term releases. 10 hasn't been maintained for a year and a half by this point.

Cervator: Whoops, thought that was a `>` not a `!=` good catch :-)

Will fix and test with a few more versions, been aiming to try it on two other systems. Can adjust logging for 8 or 11. Would we want to outright warn people if they're on 9 or 10?

Cervator:
Okay all set and tested out now. Adjusted the warning piece to warn on anything other than Java 8 or 11 (the earlier check for Java 8 compatibility should catch anything older). Tested on 8, 10, 11, and 14 last time I think. Tested on a clean system (total Gradle cache wipe) to make sure `groovyw` first works. Tested `groovyw module init` and `groovyw module init iota`

I made some updates at https://github.com/MovingBlocks/Terasology/wiki/Dev-Setup#quick-start on the new `init` and `groovyw` overall, including where Iota & Omega are defined, hopefully that'll help. Probably still more we can do, but for now I'd like to get this out of the way so I can get back to Jenkins for a bit :-)

PR #4009, opened 2020-05-31
groovyw module update-all improvements

### LINKED ISSUE #3661: Enhance the `update-all` groovyw utility script further with better logging and resume capability
Essentially three small separate enhancements but maybe not separate issues:

1) Make `groovyw module update-all` log what actually gets updated per module - right now there's no actual indication if _anything_ got updated, it would be nice to see a summary with "x new files added, y files updated, z files deleted" or something. Ideally log it to the screen, but if for some reason a log file is better that could work too.

2) In case of failures either _skip_ a given module and/or allow the script to resume smartly if executed again. For instance don't re-update a module that's already been updated within the last few minutes (there's got to be something in the local git repo that indicates the last fetch or something?)

3) Respect the `.gitignore` file when it comes to directories in the modules dir to ignore, like a possible `.settings` hanging out in there for Eclipse

JGit or whatever it is we're using should have _some_ useful utility to help with some of those :-)

Possible extra just so we have it listed somewhere: support a `groovyw module set [x]` where `[x]` is a branch name (so in other words a bulk "Set all my module branches to [x]" operation. Mostly this is probably useful if you're trying to get back to the master branch (so `groovyw module set master`) - I'm unsure what it should do if you give it a branch name that doesn't exist yet. IntelliJ rarely has tried to bulk-change my branches for all nested Git repos but rarely in a useful fashion.

cvennel:
### Contains

Fixes #3661 ~~Parts 1 and 2~~ (now fixes all three parts)

### Notes:
- I would have rather done a full diff of everything that changed in an update instead of listing the file changes per commit, but grgit doesn't seem to have the diff capability on anything other than single commits until this PR is merged https://github.com/ajoberstar/grgit/pull/318

- I'm not fully sold that skipping an update is the best approach if a module has recently been updated. Running a few extra git-fetches in the event an update needs to be re-run isn't _that_ expensive in the grand scheme, and personally I'd rather be able to ensure I have the latest by re-running the pull. Maybe some sort of "force update" option would be a good compromise here? So skip updating the module unless something like -f was passed?

- ~~I didn't understand Part 3 of the issue~~

#### Skip updating module if recently updated:
`cd modules/<module name>`
`git pull`
`cd ../..`
`groovyw module update-all`

#### Simulate updating with 5 commits:
`cd modules/<module name>`
`git reset --hard HEAD~5`
`cd ../..`
`groovyw module update-all`

#### Confirm subdirectories in .gitignore do not block parent directory
in .gitignore change `.settings/` to `.settings/hide`
Confirm that .settings is no longer ignored in groovyw module update-all

### Sample output
In the sample screenshot we see:
- .Settings is ignored due to being in the .gitignore 
- CakeLie already up to date
-  CoreSampleGameplay recently updated within 2 minutes, so skip
-  Health pulls 5 commits  to update
-  ModuleTestingEnvironment already up to date

(Image updated with .Settings skipped to honor .gitignore)
update-all-output

### DIFF
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -186,12 +186,15 @@ class common {
      * Update a given item.
      * @param itemName the name of the item to update
      */
-    def updateItem(String itemName) {
+    def updateItem(String itemName, boolean skipRecentUpdates = false) {
         File targetDir = new File(targetDirectory, itemName)
         def searchString = itemName
-        if (itemName.startsWith(".")){
-            // add literal slash for regex start with '.'
-            searchString = "\\$itemName"        
+        if (!Character.isLetterOrDigit(itemName.charAt(0))){   
+            println color ("Skipping update for $itemName: starts with non-alphanumeric symbol", Ansi.YELLOW)
+            return
+        }
+        if(skipRecentUpdates){
+            println color("Continue flag!", Ansi.RED)
         }
         def inGitIgnore = false
         new File(".gitignore").eachLine{ line -> 
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -210,15 +210,9 @@ class common {
             searchString = "\\$itemName"        
         }
         for (file in gitIgnoreFiles) {
-            def relPathString = file.relativePath(new File("."))
-            def relSearchString = searchString
-            if (relPathString.length() > 3){
-                // file.relativePath adds an extra ../ for what we need in the gitignore, trim it off
-                relSearchString = relPathString.substring(3) + '/' + searchString
-            }
             def lines = file as String[]
             for (line in lines){
-                if ((line ==~ relSearchString) || (line ==~ "$relSearchString/")){
+                if ((line ==~ searchString) || (line ==~ "$searchString/")){
                     println("$itemName set to ignore in $file")
                     return true
                 }
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -248,7 +248,7 @@ class common {
                         }
                         print("\n")
                     } else {
-                        println color ("No changes found\n", Ansi.YELLOW)
+                        println color ("No changes found", Ansi.YELLOW)
                     }
                 } catch (GrgitException exception) {
                     println color("Unable to update $itemName, Skipping: ${exception.getMessage()}", Ansi.RED)
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -236,6 +236,20 @@ class common {
                 println color("uncommitted changes. Skipping.", Ansi.YELLOW)
             } else {
                 println color("updating $itemType $itemName", Ansi.GREEN)
+                File targetDirFetchHead = new File("$targetDir/.git/FETCH_HEAD")
+                Date lastUpdate = new Date(targetDirFetchHead.lastModified())
+                def recentlyUpdated = use(groovy.time.TimeCategory){
+                    def timeElapsedSinceUpdate = new Date() - lastUpdate
+                    if (timeElapsedSinceUpdate < 2.minutes){
+                        println color("Skipping update for $itemName: updated within last 2 minutes", Ansi.YELLOW)
+                        return true
+                    } else {
+                        return false
+                    }
+                }
+                if (recentlyUpdated){
+                    return
+                }
                 try {
                     def current_sha = itemGit.log(maxCommits: 1).find().getAbbreviatedId(8)
                     itemGit.pull remote: defaultRemote
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -215,6 +215,20 @@ class common {
                 println color("uncommitted changes. Skipping.", Ansi.YELLOW)
             } else {
                 println color("updating $itemType $itemName", Ansi.GREEN)
+                File targetDirFetchHead = new File("$targetDir/.git/FETCH_HEAD")
+                Date lastUpdate = new Date(targetDirFetchHead.lastModified())
+                def recentlyUpdated = use(groovy.time.TimeCategory){
+                    def timeElapsedSinceUpdate = new Date() - lastUpdate
+                    if (timeElapsedSinceUpdate < 2.minutes){
+                        println color("Skipping update for $itemName: updated within last 2 minutes", Ansi.YELLOW)
+                        return true
+                    } else {
+                        return false
+                    }
+                }
+                if (recentlyUpdated){
+                    return
+                }
                 try {
                     def current_sha = itemGit.log(maxCommits: 1).find().getAbbreviatedId(8)
                     itemGit.pull remote: defaultRemote
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -183,6 +183,23 @@ class common {
     }
 
     /**
+     * Check if an item was updated within the provided time limit
+     * @param fetchHead the item's FETCH_HEAD file in the .git directory
+     * @param timeLimit the time limit for considering something recently updated, for example: use(groovy.time.TimeCategory){ 10.minute }
+     */
+    def isRecentlyUpdated(File fetchHead, def timeLimit){
+        Date lastUpdate = new Date(fetchHead.lastModified())
+        def recentlyUpdated = use(groovy.time.TimeCategory){
+            def timeElapsedSinceUpdate = new Date() - lastUpdate
+            if (timeElapsedSinceUpdate < timeLimit){
+                return true
+            } else {
+                return false
+            }
+        }
+    }
+
+    /**
      * Update a given item.
      * @param itemName the name of the item to update
      */
@@ -192,9 +209,6 @@ class common {
         if (!Character.isLetterOrDigit(itemName.charAt(0))){   
             println color ("Skipping update for $itemName: starts with non-alphanumeric symbol", Ansi.YELLOW)
             return
-        }
-        if(skipRecentUpdates){
-            println color("Continue flag!", Ansi.RED)
         }
         def inGitIgnore = false
         new File(".gitignore").eachLine{ line -> 
@@ -237,17 +251,9 @@ class common {
             } else {
                 println color("updating $itemType $itemName", Ansi.GREEN)
                 File targetDirFetchHead = new File("$targetDir/.git/FETCH_HEAD")
-                Date lastUpdate = new Date(targetDirFetchHead.lastModified())
-                def recentlyUpdated = use(groovy.time.TimeCategory){
-                    def timeElapsedSinceUpdate = new Date() - lastUpdate
-                    if (timeElapsedSinceUpdate < 2.minutes){
-                        println color("Skipping update for $itemName: updated within last 2 minutes", Ansi.YELLOW)
-                        return true
-                    } else {
-                        return false
-                    }
-                }
-                if (recentlyUpdated){
+                def timeLimit = use(groovy.time.TimeCategory){ 10.minute }
+                if (skipRecentUpdates && isRecentlyUpdated(targetDirFetchHead, timeLimit)){
+                    println color("Skipping update for $itemName: updated within last $timeLimit", Ansi.YELLOW)
                     return
                 }
                 try {
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -265,20 +265,6 @@ class common {
                 println color("uncommitted changes. Skipping.", Ansi.YELLOW)
             } else {
                 println color("updating $itemType $itemName", Ansi.GREEN)
-                File targetDirFetchHead = new File("$targetDir/.git/FETCH_HEAD")
-                Date lastUpdate = new Date(targetDirFetchHead.lastModified())
-                def recentlyUpdated = use(groovy.time.TimeCategory){
-                    def timeElapsedSinceUpdate = new Date() - lastUpdate
-                    if (timeElapsedSinceUpdate < 2.minutes){
-                        println color("Skipping update for $itemName: updated within last 2 minutes", Ansi.YELLOW)
-                        return true
-                    } else {
-                        return false
-                    }
-                }
-                if (recentlyUpdated){
-                    return
-                }
                 try {
                     def current_sha = itemGit.log(maxCommits: 1).find().getAbbreviatedId(8)
                     itemGit.pull remote: defaultRemote
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -44,9 +44,6 @@ class common {
     /** For keeping a list of remote items that can be retrieved */
     String[] cachedItemList
 
-     /** For keeping a list of .gitignore files in the project */
-    def gitIgnoreFiles
-
     /**
      * Initialize defaults to match the target item type
      * @param type the type to be initialized
@@ -70,7 +67,6 @@ class common {
         githubTargetHome = githubDefaultHome
         targetDirectory = itemTypeScript.targetDirectory
         itemType = itemTypeScript.itemType
-        gitIgnoreFiles = getGitIgnoreFiles()
     }
 
     /**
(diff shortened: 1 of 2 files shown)


skaldarnar:
Thanks @cvennel - the sample output looks neat. I'll get round reviewing this later this eveneing (hopefully 🙈 ). Here are my initial 2 cents on the topic:

Part 3 refers to the fact that the script is blindly trying to update the "module" `.settings` even though the folder is in `.gitignore`d. 

I'm not sure what the benefit of skipping should be _in any case_ - I hope we are not optimizing for bandwidth, and pointlessly waiting for two minutes or doing the commands manually seems silly. I thought that `git pull` (doing `fetch` and `merge`) is already kind of smart to check that no updates are available.

cvennel:
Hi @skaldarnar. Thanks for the clarification on part 3. 

Regarding part 2, I agree with you, I'm not seeing the benefit of skipping as described in part 2, but it was in the issue so I wrote it up :) I have it as a separate commit to make it easy to back out if that's what we end up deciding to do. 

I'll add some more to handle part 3.

cvennel:
Ok, now the .setting module should be ignored since it's in the .gitignore. 
We also want to make sure subdirectories in the .gitignore don't block the parent directory. I'll add more details to the description and how to test

config/groovy/common.groovy
```diff
@@ -188,6 +188,24 @@ class common {
...
+        }
+        def inGitIgnore = false
+        new File(".gitignore").eachLine{ line -> 
+            // match if line is exactly same os itemName or has trailing '/' but
+            // not if has further subdirectories
+            if ((line ==~ searchString) || (line ==~ "$searchString/")){
+                inGitIgnore = true
+                return
+            }
+        }
```
skaldarnar, line 204: This should be extracted to its own method for better readability. 

Doing so will also make it easier to spot bugs or find the right location for doing improvements. We're missing that `.gitignore` files can also appear in subfolders and "add up" to each other. Not a huge deal, but ideally we could use some library to do this work for us...

config/groovy/common.groovy
```diff
@@ -215,8 +233,41 @@ class common {
...
+                    if (timeElapsedSinceUpdate < 2.minutes){
+                        println color("Skipping update for $itemName: updated within last 2 minutes", Ansi.YELLOW)
+                        return true
+                    } else {
+                        return false
+                    }
+                }
+                if (recentlyUpdated){
+                    return
+                }
```
skaldarnar, line 249: Honestly, I'd throw that out as I don't see any good reason for doing this :monocle_face: The solution works nicely, I just don't think we should do it that way...

cvennel: I was kind of hoping you would say that  :) I'll back this out.

cvennel: reverted

config/groovy/common.groovy
cvennel: Ok this now checks every `.gitignore`. I also added some output to tell you which `.gitignore` is blocking the module. See sample below:
update-all-output-all-gitignores

cvennel:
@skaldarnar, do you mind telling me what check failed?

config/groovy/common.groovy
skaldarnar: If I understand this correclty, `modules/foo` will be ignore now if mentioned in `sandbox/.gitignore`? I don't think that's how the ignored files are computed 🧐 

Some references:

- https://www.atlassian.com/git/tutorials/saving-changes/gitignore
- https://git-scm.com/docs/gitignore
- https://help.github.com/en/github/using-git/ignoring-files

skaldarnar:
The **analytics step** failed in our Jenkins pipeline: 

image

It failed with 

```
FAILURE: Build failed with an exception.

* What went wrong:
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)
```

I'm restarting the build.

config/groovy/common.groovy
cvennel: I'm not used to multiple .gitignores. I had a section computing the relative path but realized I made a mistake and was finding the relative path to the root dir instead of the modules dir. Then the build failed after backing out the relative path stuff which surprised me. Wanted to get more info on that before continuing the relative path

keturn:
I haven't been following this, just noticing bits as they come across the activity feed, but maybe https://git-scm.com/docs/git-check-ignore is interesting to you?

cvennel:
> I haven't been following this, just noticing bits as they come across the activity feed, but maybe https://git-scm.com/docs/git-check-ignore is interesting to you?

@keturn I hadn't seen that before, thanks for the new info! Unfortunately it looks like this would ignore every module since the main .gitignore has `/modules/*` in it which can't change otherwise the main repo would get all the modules stuff. 

I also don't think the .gitignores below the modules directory should be able to control each other except maybe within .settings (putting ../health in the .gitignore file in the CakeLie module shouldn't allow blocking updates for health)

So I think this is in a weird state of trying to honor the .gitignores, but since the .gitignore would ignore everything we can't trust the .gitignore. Would we be better off creating another settings file to control this like maybe a .moduleignore file in the modules dir? 

Using a dedicated module update config isn't as slick, but the only other thing I can think of at the moment is manually parsing each config file looking for an exact match of the relative path to a module in addition to the main .gitignore's /modules/* setting. This feels wrong to me since it would seem redundant to have both `/modules/*` and `/modules/.settings/` in the gitignore when in reality they serve two completely different purposes.

sorry for the long rambling message, kind of unsure which way to take this now.

Cervator:
Hey @cvennel - great work here, really appreciate seeing it and am looking forward to trying it :-)

For the `.gitignore` thing while it sounds cool to use it in a dynamic fashion, if it is getting too complex would it be an option simply to hardcode ignoring anything not starting with an alphanumeric character? That'll catch all the metadata type directories like `.settings` and about anything else I could think of. Anything starting with an actual letter (or possibly number) likely should be a module or we're being goofy.

On skipping recently updated modules I could see that maybe being handy in a mega-workspace - if you're trying to update 200 modules and a few fail from being dirty it might be nice to clean them up then just run again (I've hit that occasionally and been reluctant to run the whole thing again). Although who knows if 2 minutes would be enough for that, and trying to make that more advanced would probably again just overcomplicate things.

cvennel:
New updates:

part 1) none

part 2)For the .gitignore section, I took @Cervator's suggestion and hard-coded skipping non-alphanumeric modules since it was just getting too goofy with the `.gitignore` follow but don't follow problem.

part 3) I made skipping modules an option. By default, we'll let git try and fetch each module and progress as normal. However, if so desired, you can now run `groovy update-all --continue` which will skip anything that has been updated within 10 minutes, I think this strikes a good balance between various suggestions from @Cervator  and @skaldarnar

cvennel:
and regarding the WIP prefix, I lost track of the draft button and habits from using Gitea at work kicked in, oops :) thanks for the reminder about draft

cvennel:
> * I tried the `--continue` flag after first running an update normally, which ended up not actually needing to update anything. On the subsequent run with `--continue` oddly _two_ of the module (of about a dozen) skipped due to recent updates, while the rest just did the "updating module x" then followed by "No changes found". That's a convenience at best so even if it doesn't work perfectly it doesn't really block merging :-)

The condition will only skip updates if the .git/FETCH_HEAD has been updated, so a `git fetch` that results in no changes may not change the file and such would not cause that module to be skipped. I think in testing I had to `git pull` in the module directory because `git fetch` didn't update the file but I can't fully remember. I bet we could do something simple like just open the FETCH_HEAD while iterating through to change the system time of the last update on the file. That would cause every module you fetched to skip as well.

config/groovy/common.groovy
```diff
@@ -215,8 +250,33 @@ class common {
                 println color("uncommitted changes. Skipping.", Ansi.YELLOW)
             } else {
                 println color("updating $itemType $itemName", Ansi.GREEN)
+                File targetDirFetchHead = new File("$targetDir/.git/FETCH_HEAD")
+                def timeLimit = use(groovy.time.TimeCategory){ 10.minute }
+                if (skipRecentUpdates && isRecentlyUpdated(targetDirFetchHead, timeLimit)){
```
cvennel, line 255: I just realized in a newly cloned module the FETCH_HEAD may not exist. I need to add some sort of check to make sure the file exists before running `isRecentlyUpdated()`

keturn: reading stuff directly out of `.git/` makes me anxious.

I won't say it's _never_ a good idea. In practice, those sorts of files usually have pretty stable formats. But if you can do it through your git client — which in this case looks like is Grgit wrapping JGit — do prefer that. Then they can be the ones responsible for handling whether or not files exist and what settings are in them.

cvennel: I'm not actually reading the file, just checking the last time the file was modified by git.

cvennel:
I just added checks to make sure the fetch_head exists and now update the modified time whenever `groovyw module update-all` is run. This fixes the behavior that @Cervator noted:

> On the subsequent run with --continue oddly two of the module (of about a dozen) skipped due to recent updates, while the rest just did the "updating module x" then followed by "No changes found".

config/groovy/common.groovy
skaldarnar: I'd still vote for removing this "feature" 🤷

Cervator: So I'm not married to the concept of that feature personally, I thought it could be a cool bit of utility if it would work out easily enough, but if it is becoming too complex I'm not at all opposed to going without :-)

I just retried it, ran `groovyw module update-all` and one module updated. Ran `groovyw module update-all --continue` and all skipped. So it seemed like it worked 👍 Whether it actually is meaningful at this point I'm less sure about. 

We had more cases in the past where an `update-all` in a mega-workspace could hit _one_ module where the remote was gone (say the user had been testing some other users module and it had been deleted on GitHub) which would break the whole update. But I just tried to break a module's Git config and the process caught that gracefully, skipped it, and continued on its own.

Still, that could just be one case, I'm not sure if there might be others. It seems like it works perfectly at this point and it is optional if passing in `--continue` - maybe it'll still be useful?

config/groovy/common.groovy
```diff
@@ -182,12 +182,47 @@ class common {
         addRemote(itemName, defaultRemote, "https://github.com/${githubDefaultHome}/${itemName}.git")
     }
 
+    /**
+     * Check if an item was updated within the provided time limit
+     * @param fetchHead the item's FETCH_HEAD file in the .git directory
```
skaldarnar, line 187: This methods works independently of the fact that you pass it a `FETCH_HEAD` file. Thus, I would just call this `file` to make that clear. 

Using this method to determine whether the git state is somewhat recent is a usage specific.

config/groovy/common.groovy
```diff
@@ -182,12 +182,47 @@ class common {
...
+            }
+        }
+    }
+
     /**
      * Update a given item.
      * @param itemName the name of the item to update
      */
-    def updateItem(String itemName) {
+    def updateItem(String itemName, boolean skipRecentUpdates = false) {
```
skaldarnar, line 206: So if this is called `skipRecentUpdates` here - why is the corresponding command line flag called `--continue`?

config/groovy/common.groovy
```diff
@@ -182,12 +182,47 @@ class common {
...
+        def inGitIgnore = false
+        new File(".gitignore").eachLine{ line -> 
+            // match if line is exactly same os itemName or has trailing '/' but
+            // not if has further subdirectories
+            if ((line ==~ searchString) || (line ==~ "$searchString/")){
+                inGitIgnore = true
+                return
+            }
+        }
+        if (inGitIgnore){
```
skaldarnar, line 222: Please turn this into a method and use it inline in the check:

```suggestion
        if (isGitIgnored(itemName)) {
```

config/groovy/common.groovy
```diff
@@ -182,12 +182,47 @@ class common {
...
+    }
+
     /**
      * Update a given item.
      * @param itemName the name of the item to update
      */
-    def updateItem(String itemName) {
+    def updateItem(String itemName, boolean skipRecentUpdates = false) {
         File targetDir = new File(targetDirectory, itemName)
+        def searchString = itemName
```
skaldarnar, line 208: What is this declaration used for? Are you modifying the `searchString` somewhere?

config/groovy/common.groovy
```diff
@@ -215,8 +250,37 @@ class common {
...
                 println color("updating $itemType $itemName", Ansi.GREEN)
+                File targetDirFetchHead = new File("$targetDir/.git/FETCH_HEAD")
+                def timeLimit = use(groovy.time.TimeCategory){ 10.minute }
+                if (skipRecentUpdates && targetDirFetchHead.exists() && isRecentlyUpdated(targetDirFetchHead, timeLimit)){
+                    println color("Skipping update for $itemName: updated within last $timeLimit", Ansi.YELLOW)
+                    return
+                }
+                if (targetDirFetchHead.exists()){
+                    targetDirFetchHead.setLastModified(new Date().getTime())
+                }
```
skaldarnar, line 261: This should at least be extracted to a method for better readability.

I think this bit clutters our code, and in my opinion for no apparent reason other than "because it is technically possible". We are not using Github API to worry about rate limiting, so I don't understand why we would ever need this feature. 

@Cervator I'd rather have 20-30 lines of code less to maintain (and/or go stale)...

config/groovy/common.groovy
```diff
@@ -215,8 +250,37 @@ class common {
...
+                
                 try {
+                    def current_sha = itemGit.log(maxCommits: 1).find().getAbbreviatedId(8)
                     itemGit.pull remote: defaultRemote
+                    def post_update_sha = itemGit.log(maxCommits: 1).find().getAbbreviatedId(8)
+                        
+                    if (current_sha != post_update_sha){
+                        println color("Updating $current_sha..$post_update_sha", Ansi.GREEN)
```
skaldarnar, line 269: Could you please add a `//TODO: ..` comment referencing the grgit issue

config/groovy/common.groovy
Cervator: That's the main bit of hesitancy I've got now that my original use case doesn't result in the update halting early. If that still happened in mega-workspaces (I've hit that regularly in the past, having various GCI modules etc) then I'd still go for this for sure. But now, that I'm not sure it'll ever be needed? Conflicted 😕 

To be clear the time it takes to chug through a 200 module megaworkspace with an update-all possibly several times (if something breaks) was the motivation, less so the stress on GitHub. Although my digital treehugger self does lose a tiny bit of sleep at night over that as well :-)

skaldarnar:
Okay, let's get this merged (see pending comments) _with_ the skipping feature for now. 

In addition, bringing @cvennel 's attention to https://github.com/MovingBlocks/Terasology/issues/4035 for more build and logistics magic 😉 Ever wanted to try out Kotlin? 🤓

cvennel:
I should be able to work on this tonight after work. I had to finish up some stuff for school this weekend . And I like kotlin :) I'll take a look at #4035 too

config/groovy/common.groovy
cvennel: Good point.

config/groovy/util.groovy
```diff
@@ -117,8 +117,14 @@ switch (cleanerArgs[0]) {
     case "update-all":
         println "We're updating every $itemType"
         println "List of local entries: ${common.retrieveLocalItems()}"
-        for (item in common.retrieveLocalItems()) {
-            common.updateItem(item)
+        if (cleanerArgs.contains("--continue")) {
```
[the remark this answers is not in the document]
cvennel: The original reason I chose `--continue` is it lines up with some git commands so I thought it would be easier to remember. For example, If a git merge fails due to conflicts, you resolve the conflicts and call `git merge --continue`. I also did the two dashes convention for full words (-u is short for --set-upstream). 

I'm going change this to `-skip-recently-updated` to closer match the other flags listed in the printUsage.

config/groovy/common.groovy
cvennel: changed flag

config/groovy/common.groovy
cvennel: Oof, I mean to revert this as well. This isn't actually going to be used any more, I took out the gitignore processing in favor of just ignoring directories starting with non-alpha-numeric characters since I can't truly honor the gitignore

config/groovy/common.groovy
cvennel: This accidentally returned when I reverted a commit. At one point I was modifying the search string to find relative paths... I cut this out

config/groovy/common.groovy
cvennel: done

config/groovy/util.groovy
```diff
@@ -265,48 +266,49 @@ private void printListItemsCondensed(String[] items) {
...
+        - 'list-remotes (item)' - lists all remotes for (item) 
+        - 'refresh' - replaces the Gradle build file for all items of the given type from the latest template
+        - 'createDependencyDotFile' - creates a dot file recursively listing dependencies of given locally available module, can be visualized with e.g. graphviz
+        
+        Available flags:
+        '-remote [someRemote]' to clone from an alternative remote, also adding the upstream org (like MovingBlocks) repo as 'origin'
+            Note: 'get' + 'recurse' only. This will override an alternativeGithubHome set via gradle.properties.
+        '-simple-list-format' to print one item per row for the 'list' sub-command, even for large numbers of items
+        '-condensed-list-format' to group items by starting letter for the 'list' sub-command (default with many items)
+        '-skip-recently-updated' (Only for update-all) to skip updating modules that have already been updated within 10 minutes
```
cvennel, line 293: This is the new line, The reset is just removing the million `println` statements to replace with a multiline  string

config/groovy/util.groovy
```diff
@@ -297,50 +298,52 @@ private void printListItemsCondensed(String[] items) {
...
+    - 'list-remotes (item)' - lists all remotes for (item) 
+    - 'refresh' - replaces the Gradle build file for all items of the given type from the latest template
+    - 'createDependencyDotFile' - creates a dot file recursively listing dependencies of given locally available module, can be visualized with e.g. graphviz
+    
+    Available flags:
+    '-remote [someRemote]' to clone from an alternative remote, also adding the upstream org (like MovingBlocks) repo as 'origin'
+           Note: 'get' + 'recurse' only. This will override an alternativeGithubHome set via gradle.properties.
+    '-simple-list-format' to print one item per row for the 'list' sub-command, even for large numbers of items
+    '-condensed-list-format' to group items by starting letter for the 'list' sub-command (default with many items)
+    '-skip-recently-updated' (Only for update-all) to skip updating modules that have already been updated within 10 minutes
```
cvennel, line 327: This is the new line. The rest is removing unnecessary `println` statements in favor of a multi-line string

cvennel: to be clear, I meant line 327

config/groovy/common.groovy
cvennel: I didn't fully extract to a new method, but did make it more readable. It was looking to me like extracting wouldn't make it much more readable, let me know what you think

Cervator:
Thank you very much for the work @cvennel :-)

And for the review and support @skaldarnar 👍

PR #4010, opened 2020-05-31
feat: Improve block placement of ceiling-supporting family

kBlaszczyk:
This PR improves the block placement behavior of the `CeilingSupportingHorizontalFamily`

Previously the only way to place blocks upside down, using this block family, was to attach them to a BOTTOM side. This PR makes it possible to achieve the same thing, targeting horizontal sides.
This prevents the player from having to build a temporary support structure to simply have a ceiling available, to use for the placement of a new block.

Upside down block placement, is done if any of these conditions are met:
* the target side is a BOTTOM side
* the target side is horizontal (FRONT, BACK, LEFT, RIGHT) and the player targeted the upper half of the side when placing the block

### Test
Assign the above-mentioned block family to a furnace by adding the following line to the furnace.block json file:
`"family": "ceilingSupportingHorizontal",`

Add furnaces to your inventory using the following command:
`give furnace 10`
Attach them to different block sides and verify that the placement behaves as described above.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/family/CeilingSupportingHorizontalFamily.java
+++ b/engine/src/main/java/org/terasology/world/block/family/CeilingSupportingHorizontalFamily.java
@@ -122,13 +122,9 @@ public class CeilingSupportingHorizontalFamily extends AbstractBlockFamily {
 
     @Override
     public Block getBlockForPlacement(BlockPlacementData data) {
-        Side mainSide = Side.TOP;
-
         boolean upsideDownPlacement = data.attachmentSide == Side.BOTTOM
                 || data.attachmentSide != Side.TOP && data.relativeAttachmentPosition.y() > 0.5;
-        if (upsideDownPlacement) {
-            mainSide = Side.BOTTOM;
-        }
+        final Side mainSide = upsideDownPlacement ? Side.BOTTOM : Side.TOP;
 
         Side blockDirection = Side.inDirection(-data.viewingDirection.x(), 0, -data.viewingDirection.z());
         return blocks.get(ExtendedSide.getExtendedSideFor(mainSide, blockDirection));
--- a/engine/src/main/java/org/terasology/world/block/family/CeilingSupportingHorizontalFamily.java
+++ b/engine/src/main/java/org/terasology/world/block/family/CeilingSupportingHorizontalFamily.java
@@ -34,8 +34,8 @@ import java.util.Map;
 /**
  * Rotates a block to either be upright or upside down and always face the player.
  * <p>
- * If the block is placed onto a BOTTOM side (i.e. the ceiling), the block will be upside down and face the player horizontally.
- * Placement on any other side will cause the block to simply face the player horizontally.
+ * The block will be placed upside down, if it's either being attached to a BOTTOM side (i.e. the ceiling),
+ * or if it's being attached to the upper half of any horizontal side (LEFT, RIGHT, FRONT, BACK).
  */
 @RegisterBlockFamily("ceilingSupportingHorizontal")
 public class CeilingSupportingHorizontalFamily extends AbstractBlockFamily {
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/world/block/family/CeilingSupportingHorizontalFamily.java
```diff
@@ -122,13 +122,16 @@ private Block transformBlock(
...
 
-        if (data.attachmentSide == Side.BOTTOM) {
-            return blocks.get(ExtendedSide.getExtendedSideFor(Side.BOTTOM, blockDirection));
-        } else {
-            return blocks.get(ExtendedSide.getExtendedSideFor(Side.TOP, blockDirection));
+        boolean upsideDownPlacement = data.attachmentSide == Side.BOTTOM
+                || data.attachmentSide != Side.TOP && data.relativeAttachmentPosition.y() > 0.5;
+        if (upsideDownPlacement) {
+            mainSide = Side.BOTTOM;
         }
```
skaldarnar, line 131: Instead of declaring `mainSide` and then mutating the value you could do the check first and just do a single-time assignment:

```suggestion
        final Side mainSide = upsideDownPlacement ? Side.BOTTOM : Side.TOP;
```

engine/src/main/java/org/terasology/world/block/family/CeilingSupportingHorizontalFamily.java
```diff
@@ -122,13 +122,16 @@ private Block transformBlock(
 
     @Override
     public Block getBlockForPlacement(BlockPlacementData data) {
-        Side blockDirection = Side.inDirection(-data.viewingDirection.x(), 0, -data.viewingDirection.z());
+        Side mainSide = Side.TOP;
```
skaldarnar, line 125:
```suggestion
```

engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
```diff
@@ -112,6 +121,28 @@ public void onPlaceBlock(ActivateEvent event, EntityRef item) {
...
+    /**
+     * Returns the position at which the block side was hit, relative to the side.
+     * <p/>
+     * Example: The front side was hit right in the center.
+     * The result will be (0.5, 0.5), representing the relative hit position on the side's surface.
+     * @param hitPosition the hit position
+     * @param blockPosition the block position relative to its center (block (0, 0, 0) has block position (0.5, 0.5, 0.5))
+     * @return the 2D hit position relative to the side that was hit
+     */
+    private Vector2f getSideHitPosition(Vector3f hitPosition, Vector3f blockPosition) {
```
skaldarnar, line 133: This assumes a cubic block, right? I'm totally fine with that, but maybe we should mention that in the doc comment?

engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
```diff
@@ -112,6 +121,28 @@ public void onPlaceBlock(ActivateEvent event, EntityRef item) {
...
+        float epsilon = 0.0001f;
+        Vector3f relativeHitPosition = new Vector3f(hitPosition).sub(blockPosition);
+
+        if (Math.abs(relativeHitPosition.x) > 0.5f - epsilon) {
```
skaldarnar, line 137: I'm wondering whether we have convenience for that in TeraMath....

engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
```diff
@@ -90,8 +91,16 @@ public void onPlaceBlock(ActivateEvent event, EntityRef item) {
         Vector3i placementPos = new Vector3i(targetBlock);
         placementPos.add(surfaceSide.getVector3i());
 
+        Vector3f targetPosition = event.getTargetLocation();
+        Vector2f relativePlacementPosition;
+        if (targetPosition != null) {
+            relativePlacementPosition = getSideHitPosition(event.getHitPosition(), targetPosition);
+        } else {
+            relativePlacementPosition = new Vector2f();
+        }
```
skaldarnar, line 100: What about moving the null check into `getSideHitPosition`?

```suggestion
        Vector2f relativePlacementPosition = getSideHitPosition(event.getHitPosition(), targetPosition);
```

engine/src/main/java/org/terasology/world/block/family/CeilingSupportingHorizontalFamily.java
kBlaszczyk: Yes, much better.

engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
kBlaszczyk: I added some explanation.

engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
kBlaszczyk: I don't know, feels rather specific.
You mean a convenience for "testing if the absolute of a value is bigger than some value, considering some delta"?
There are no utility functions in TeraMath using the absolute functions, so I'm pretty sure it's not in there.
Also, I'm not sure how to name this, in case I wanted to move it to TeraMath. Any ideas?

engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
kBlaszczyk: I don't like this one.
My method wants a vector to work with, nobody should pass it null.
It is true however, that I'm cluttering up the higher level method, so I moved the null check into another private method, working on the event. I think it makes more sense that way and I hope you'll like that solution too.

engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
skaldarnar: Not necessarily with the `Math.abs` but the comparison of floating point numbers with some ε

engine/src/main/java/org/terasology/world/block/items/BlockItemSystem.java
skaldarnar: I surely do :+1:

PR #4016, opened 2020-06-01
refactor: remove dead code

skaldarnar:
As @pollend found out in #3961 there are unused classes and other dead code in the engine.

Tested in Omega workspace, can run Josharias Survival without any (new) issues.


engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ComponentClassTypeHandler.java
```diff
@@ -17,22 +17,23 @@
...
     @Override
     public Class<? extends Component> getFromString(String representation) {
-        try {
-            final Class<? extends Component> clazz = Class.forName(representation).asSubclass(Component.class);
-            return clazz;
-        } catch (ClassNotFoundException e) {
-            e.printStackTrace();
-            return null;
-        }
+        return context.getSandbox().findSubTypeOf(representation, Component.class).orElse(null);
```
keturn, line 37: This is the only thing in this PR that's not just a straight deletion. And it looks okay from here.

It _does_ change the constructor, but presumably that's the sort of thing that you'd find out pretty fast as soon as you tried to compile against it.

Actually ... where _is_ this class used? it's not in my workspace.

skaldarnar: Ah, this is actually some sneaky little code cleanup :nerd_face:

This class here existed, but was never used. However, there existed another inner class with exact same name (and purpose) in `ComponentClassTypeHandlerFactory`. As I don't like inner classes, I thought it would be better to have the correct code in this class, which is now used by the factory. 

See https://github.com/MovingBlocks/Terasology/pull/4016/files#diff-7fd63f399c356366e5f1abab7dc34ebcL40-L57

PR #4019, opened 2020-06-01
fix(docs): replace some develop instructions with links to the wiki.

keturn:
Duplicated information was becoming out-of-date.

### DIFF
--- a/.idea/dictionaries/kevint.xml
+++ b/.idea/dictionaries/kevint.xml
@@ -1,6 +1,9 @@
 <component name="ProjectDictionaryState">
   <dictionary name="kevint">
     <words>
+      <w>gameplay</w>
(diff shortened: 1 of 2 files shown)


.idea/dictionaries/kevint.xml
```diff
@@ -1,6 +1,9 @@
 <component name="ProjectDictionaryState">
   <dictionary name="kevint">
```
skaldarnar, line 2: Is that your personal dictionary (or where does `kevin?  come from)?

keturn: [this confuses everyone](https://keturn.net/blog/posts/2020/05/git-add-dot-idea-share-intellij/#idea/dictionaries) but they've been resistant to changing their implementation

Issue #4015, opened 2020-05-31
gradle reliability roadmap


keturn:
Reliable builds during development with support for combined gradle & IntelliJ IDEA.

I have a sort of roadmap for this. Maybe it's the sort of thing that should be a Milestone or have a Project board, but I'll start by writing here so it's someplace besides my head:

* [ ] [regression] #4014: transient `modules/*.jar`
* [ ] gradle `terasology-modules` plugin. started by #3993.
  * [ ] extract `cacheReflections` task. work in progress, keturn has a branch somewhere. beware #1472. also relates to #2445.
  * [ ] make terasology-module plugin not depend on `/config/gradle/publish.gradle`. Surprisingly enough, it seems to work, but it is unexpected source code organization and a blocker to the plugin being used outside this repository. https://github.com/MovingBlocks/Terasology/blob/957543938f28d703e3175e0ca6f2c5f64da0dd1d/buildSrc/src/main/kotlin/terasology-module.gradle.kts#L89
* [ ] reflections and other work with resources will be easier if engine gets rid of its `dev` source set, see https://github.com/MovingBlocks/Terasology/issues/3979#issuecomment-633094908
* [ ] make `syncAssets` task (or `copyResourcesToClasses`, as it's called in engine's build) gradle-safe. **or** make unnecessary by getting Gestalt to allow us to register more than one filesystem-directory to a gestalt-module. see #3957 & https://github.com/MovingBlocks/Terasology/blob/957543938f28d703e3175e0ca6f2c5f64da0dd1d/engine/build.gradle#L277-L278
* [ ] remove outputDir overrides on IDEA config; colliding gradle and IDEA build directories is an anti-goal.
  * [ ] in modules: https://github.com/MovingBlocks/Terasology/blob/957543938f28d703e3175e0ca6f2c5f64da0dd1d/buildSrc/src/main/kotlin/terasology-module.gradle.kts#L235-L236
  * [ ] in engine: https://github.com/MovingBlocks/Terasology/blob/957543938f28d703e3175e0ca6f2c5f64da0dd1d/engine/build.gradle#L301-L302
  * [ ] in facade: 
https://github.com/MovingBlocks/Terasology/blob/957543938f28d703e3175e0ca6f2c5f64da0dd1d/facades/PC/build.gradle#L426-L427
[...]

Issue #3979, opened 2020-05-23
gradle: merge engine-tests back in to engine directory


keturn:
The directory (and gradle) layout for engine and its tests is like this:

* /engine/src/main
* /engine-tests/src/main
* /engine-tests/src/test

instead of the standard

* /engine/src/main
* /engine/src/test

Using the standard layout would simplify the build configuration.

Dumping some [thoughts](https://discord.com/channels/270264625419911192/696830442065625118/713789711793258587) from Cervator here:

> they're kinda structured like tests. Maybe we should just move them into engine-tests , add a Gradle JavaExec to run them, and redirect the output to a file we could then probably freestyle graph in Jenkins
> does there need to be an `engine-tests` _anymore_ - many moons ago we needed that hack, but i think Gradle (and IntelliJ?) both got more advanced about how you could link together sourceSets / code modules / compile vs test
> i don't remember the details anymore but it was definitely something about being unable to have it unified because modules (maybe specifically module tests?) would get confused and things wouldn't compile / test in the right order
> maybe particularly about tests that needed the environment base classes, like HeadlessEnvironment
> worth noting is that class (and its friends IIRC) all live in the regular sourceSet inside engine-tests

Cervator:
On a related note the `dev` sourceSet inside `engine` can be simplified as well. It was kept separate for ... I dunno :) Probably just didn't seem right to have that code in the engine (it doesn't do anything for the engine) _nor_ in tests since it isn't junit based or anything, although it _feels_ kinda close to the tests, and benchmarking is sort of a test category

I propose (and have just validated that it'll work - look for and run `EntitySystemBenchmark`) moving the `org.terasology.benchmarks` package into the `main` sourceSet in `engine-tests` + remove the special handling for the `dev` sourceSet in the engine, although maybe that'll then be weird after re-unifying it since it _does_ still not fit as engine code, nor regular test code 🤔 

Once upon a time I was thinking we should have a *test* sourceSet per type of test - unit, integration, benchmark, acceptance (MTE) but maybe that'd just re-complicate things. It would really organize what kind of test something is though.

keturn:
I have found one reason for making a `-tests` project that builds an independent thing from the main library: code you want to be re-used by other tests.

If you make something like ModuleTestingEnvironment, it has dependencies that aren't otherwise part of your library's `implementation` dependencies. Your users don't want this stuff on their `implementation` classpath either, they want something they can use as a `test` dependency.

[...]

PR #4021, opened 2020-06-02
refactor: move o.t.benchmarks back in to main source set

keturn:
Moves engine/src/dev in to engine/src/main.
Removes the special `dev` SourceSet from engine's build configuration.

mentioned in #4015, #3979

### How to test

Not sure. Anyone remember why this was split out to a separate source set to begin with? I don't see any additional test failures from moving it back.

### DIFF
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -62,9 +62,13 @@ def convertGitBranch = { gitBranch ->
 // Java Section                                                                                                      //
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-// Engine for now has two source sets
 sourceSets {
-    dev
+    main.java {
+        // dev/ holds things that are neither unit tests nor application code.
+        // For notes about whether it should be its own source set, see
+        // see https://github.com/MovingBlocks/Terasology/pull/4021
+        srcDir("src/dev/java")
+    }
 
     // Adjust output path (changed with the Gradle 6 upgrade, this puts it back)
     main.java.outputDir = new File("$buildDir/classes")
@@ -78,9 +82,6 @@ configurations {
         exclude module: 'lwjgl-platform'
         exclude module: 'jinput-platform'
     }
-
-    // Beyond the standard compile "configuration" we declare one called "devImplementation" specific to the "dev" source set
-    devImplementation.extendsFrom implementation
 }
 
 // TODO: Remove when we don't need to rely on snapshots. Wonder why modules respected this set in root project, engine not so much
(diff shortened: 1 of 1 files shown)


keturn:
Oh, the dev source set has been around for a _while_: https://github.com/MovingBlocks/Terasology/commit/8e1a024bd49a081d2ec73dea37b8061efcd3b7b2

keturn:
the "it's not part of engine" argument would carry more weight if it didn't get put in the jar along with all the engine classes and included in the same reflections cache.

The gradle manual has some [opinions about when to create a custom source set](https://docs.gradle.org/current/userguide/building_java_projects.html#sec:custom_java_source_sets), but concludes

> If you’re unsure whether to create a custom source set or not, then go ahead and do so. It should be straightforward and if it’s not, then it’s probably not the right tool for the job.

My motivation for wanting to do away with it before some of the other work on the build configuration comes mostly from the fact that it _does_ end up with its outputs combined with the main source set. Reflections cache is based on `classes` output, so it seems like we should build one reflections.cache per source set, but then if you end up with two source sets each with their own reflections.cache file and you go to pack them in to the same jar, that sounds like resource collision.

and we can't name them anything other than `reflections.cache` if they're intended for use with gestalt because that's the only name by which it looks for them.

keturn:
which suggests another option: leave the sources where they are, in their separate `src/dev` and separate source set, but keep the part of this patch that source set from the reflections cache and jar.

or another other option: keep the files where they are, remove the source set, but add `src/dev` as a second source directory of the `main` source set.

Cervator:
> or another other option: keep the files where they are, remove the source set, but add src/dev as a second source directory of the main source set.

That might have been what I was fishing for in my mind the other day. Wanting to only really have one `test` sourceSet, but different directories inside to distinguish between the _types_ of tests, rather than just different packages.

Having them end up in the same jar anyway - blah. Whole new `engine-benchmarks` _project_ within the repo? Gestalt does or did that at one point.

In any case: I don't think it is necessarily worth worrying about right now. So long as we have it documented for the future, like with #3979, on what it is, why it is unsual, and what we might do with it in the future 👍

keturn:
oh, hah, did you know if you force-push **before** committing so you've just pushed what's already merged, it counts that as a nothing-to-commit PR and calls it closed?

keturn:
Okay, new pass with a less disruptive approach:

* keeps directory layout as-is
* keeps java package namespace as-is
* keeps the dev/* stuff in the engine.jar
* removes the `dev` SourceSet; gradle considers all this to be in the `main` SourceSet

and I did make sure that IDEA still recognizes that these are sources.

I know this still isn't quite ideal for what your goals about how to use those benchmarks, but I hope it's not so much of a step _away_ from it as merging the directory trees was.

PR #4022, opened 2020-06-02
gradle plugin for cacheReflections

keturn:
### Contains

This moves the definition of the `cacheReflections` task to a grade plugins.

Goals:
* be usable by our terasology-module plugin _and_ the not-exactly-a-module engine project.
* follow gradle best practices for compatibility with dependency configuration and up-to-date checks.

### How to test

That's a good question! We need better documentation about how to test the reflections cache.


keturn:
This has been a prime example of a gradle task that started with me thinking "this should be a small PR to make a task" and ended up being not-so-simple at all.

My plan was to make something that depends upon sourceSet.output.classesDir (the class files read by the reflections scanner) and puts its result to a different part of the SourceSetOutput, so it'll be included in the graph of things-gradle-does-to-compile-a-source-set-and-that-get-included-in-the-distribution.

It was a great plan until I got to the part where gradle told me I'd made a circular dependency, and I found this in gradle's implementation of SourceSetOutput:

https://github.com/gradle/gradle/blob/2ce1bc8db4b993f0dd1d8834be5b53e43f657985/subprojects/plugins/src/main/java/org/gradle/api/internal/tasks/DefaultSourceSetOutput.java#L49-L51

While looking to see if there was a bug report about that I should vote for, I came across this evidence that we're in good company: https://github.com/gradle/gradle/issues/12774

and trying to follow gradle best practices for dependency configuration is great, until you come up against these parts of gradle that were developed well before these new APIs: https://github.com/gradle/gradle/issues/12546

…

that's put a bit of a damper on things, as I _thought_ I was doing this to simplify and improve the reliability of our configuration, and it's turned out to need some messy hacks to work with these design gaps.

If I had estimated it would be this hard up-front, I would have wondered if it's even worth doing at all: how much time is actually saved by creating these caches during the build as opposed to at run time?

but now that I think I may have a set of hacks that will work, I'm swayed by the Sunk Cost Fallacy. It's come this far, might as well see it through!

DarkWeird:
NUI uses reflections for detect widgets, screen overlay(finding by .ui files's type )

PR #4023, opened 2020-06-02
fix invisible skeletalmesh

pollend:
ref: https://github.com/Terasology/CoreRendering/issues/11
ref: https://github.com/MovingBlocks/Terasology/pull/3950

verified with gooeyquest. 
unknown

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/logic/SkeletonRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/logic/SkeletonRenderer.java
@@ -248,7 +248,7 @@ public class SkeletonRenderer extends BaseComponentSystem implements RenderSyste
             skeletalMesh.material.setFloat("blockLight", 1.0f, true);
             skeletalMesh.material.setFloat3("colorOffset", skeletalMesh.color.rf(), skeletalMesh.color.gf(), skeletalMesh.color.bf(), true);
 
-            skeletalMesh.material.setMatrix4("projectionMatrix", worldRenderer.getActiveCamera().getProjectionMatrix());
+            skeletalMesh.material.setMatrix4("projectionMatrix", new org.joml.Matrix4f(worldRenderer.getActiveCamera().getProjectionMatrix()).transpose());
             skeletalMesh.material.bindTextures();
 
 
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/rendering/logic/SkeletonRenderer.java
```diff
@@ -248,7 +248,7 @@ public void renderOpaque() {
             skeletalMesh.material.setFloat("blockLight", 1.0f, true);
             skeletalMesh.material.setFloat3("colorOffset", skeletalMesh.color.rf(), skeletalMesh.color.gf(), skeletalMesh.color.bf(), true);
 
-            skeletalMesh.material.setMatrix4("projectionMatrix", worldRenderer.getActiveCamera().getProjectionMatrix());
+            skeletalMesh.material.setMatrix4("projectionMatrix", new org.joml.Matrix4f(worldRenderer.getActiveCamera().getProjectionMatrix()).transpose());
```
keturn, line 251: Since this is in the rendering path, the `new` caught my eye and I started wondering if there's a way to do this without that heap allocation.

If I'm reading this correctly, the projection matrix doesn't change that often, right? And even if it _did_ change every frame, we should plan for rendering a lot more than one skeleton per frame.

This idea sounds a little sketchy but maybe it's worth something: add a transposedProjectionMatrix accessor to the camera. that way these things that need the transposed-version can get it without each doing their own transposition of it.

pollend: there is another transpose in the perspective camera that needs to be undone. 

```
return new Matrix4f().perspective(fovY,aspectRatio,zNear,zFar).transpose();
```

keturn:
oh, I've probably overheard you having this conversation in #math-and-physics before, haven't I?

So if you've heard this before and already hashed it out and there is a Plan, don't let me derail the plan.

but it's enough of a hassle to get one round of joml PRs though; I'm worried that waiting until later makes a bigger mess and in the meantime there's a performance penalty.

that's premature optimization on my part! I haven't done the profiling and for all I know maybe the JVM can optimize this kind of thing away. But the fact that joml is so diligent about avoiding these sorts of allocations makes me think the JVM isn't _that_ good in these situations.

before I saw your edit I wasn't sure how to convey in words and figured it would be easier to show by example:

https://github.com/pollend/Terasology/compare/fix/skeletal-mesh...keturn:fix/skeletal-mesh#diff-6a3dd314d84ecfeac039ea99375ec804R137

pollend:
I think it also might not be too bad to just create a PR that corrects the transpose problem. Its most of the way there so I don't see why it can't be done now. for the moment we can merge this in so the rendering isn't broken.

PR #4025, opened 2020-06-02
fix: do not skip setting default world generator

### LINKED ISSUE #4024: unable to resolve world generator for CoreSampleGameplay

start a terasology game.

### What actually happened

new game failed with 

> Unable to resolve world generator '' - not found

### How to reproduce

* groovyw module recurse CoreSampleGameplay
* [IDEA] reimport gradle projects
* [IDEA] rebuild project
* [IDEA] run _Terasology PC_
* Singleplayer
* _Core Gameplay_ is shown as the Gameplay Template. :+1: 
* Start Playing
* :boom: 

[Terasology-init.log](https://github.com/MovingBlocks/Terasology/files/4718740/Terasology-init.log)
[Terasology-Game_1.log](https://github.com/MovingBlocks/Terasology/files/4718741/Terasology-Game_1.log)

some debugger investigation shows that the WorldInfo in InitialiseWorld.step contains only empty strings.

keturn:
Fixes #4024. :crossed_fingers: 

oh, and some fixes for raw type warnings that probably don't have anything to do with this error, I was just preemptively fixing warnings I ran in to while debugging in case they would lead me to something.

### How to test

Test gameplay modes _other than_ Josharias Survival. (JS should still work too, of course.)

Make sure you can play the game without going to the advanced setup screen.

Make sure the world type is the one chosen by the module.txt for that mode.

Try without interacting with the Gameplay Template menu at all; get the gameplay mode you're testing to be the default one by removing other types.

Try with a fresh `config.cfg` (move your old one out of the way).

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/NewGameScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/NewGameScreen.java
@@ -268,9 +268,7 @@ public class NewGameScreen extends CoreScreenLayer {
         if (defaultGameplayModule != null) {
             gameplay.setSelection(defaultGameplayModule);
 
-            if (configDefaultModuleName.equalsIgnoreCase(DEFAULT_GAME_TEMPLATE_NAME)) {
-                setDefaultGeneratorOfGameplayModule(defaultGameplayModule);
-            }
+            setDefaultGeneratorOfGameplayModule(defaultGameplayModule);
         } else {
             // Find the first gameplay module that is available.
             for (Module module : moduleManager.getRegistry()) {
--- a/engine/src/main/java/org/terasology/world/generator/plugin/DefaultWorldGeneratorPluginLibrary.java
+++ b/engine/src/main/java/org/terasology/world/generator/plugin/DefaultWorldGeneratorPluginLibrary.java
@@ -29,13 +29,13 @@ import java.util.List;
  */
 public class DefaultWorldGeneratorPluginLibrary implements WorldGeneratorPluginLibrary {
 
-    private ClassLibrary<WorldGeneratorPlugin> library;
+    private final ClassLibrary<WorldGeneratorPlugin> library;
 
     public DefaultWorldGeneratorPluginLibrary(ModuleEnvironment moduleEnvironment, Context context) {
         library = new DefaultClassLibrary<>(context);
-        for (Class entry : moduleEnvironment.getTypesAnnotatedWith(RegisterPlugin.class)) {
+        for (Class<?> entry : moduleEnvironment.getTypesAnnotatedWith(RegisterPlugin.class)) {
             if (WorldGeneratorPlugin.class.isAssignableFrom(entry)) {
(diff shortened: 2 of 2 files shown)


engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/NewGameScreen.java
```diff
@@ -268,9 +268,7 @@ public void onOpened() {
         if (defaultGameplayModule != null) {
             gameplay.setSelection(defaultGameplayModule);
 
-            if (configDefaultModuleName.equalsIgnoreCase(DEFAULT_GAME_TEMPLATE_NAME)) {
-                setDefaultGeneratorOfGameplayModule(defaultGameplayModule);
-            }
+            setDefaultGeneratorOfGameplayModule(defaultGameplayModule);
```
kBlaszczyk, line 271: Why did this become necessary?
It looks like the default generator was only used under a specific condition. Now you are potentially overwriting the generator, regardless the value of `configDefaultModuleName`.
The changes to `DefaultWorldGeneratorPluginLibrary.java` are fixing the issue on their own btw.

keturn:
> The changes to DefaultWorldGeneratorPluginLibrary.java are fixing the issue on their own btw.

They are‽ I made those changes first, so I think I would have noticed that.

And I can't quite imagine _how_ they would fix this. Those changes to not use raw types should make things a little stricter, but if they're changing that kind of behavior I'm even more confused than I thought.

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/NewGameScreen.java
keturn: Yeah, that specific condition was "is the default module Josharias Survival?"

_why_ that specific condition, that is a good question, I wish I knew.

>  Now you are potentially overwriting the generator

**over**writing it? The problem I was encountering is that it wasn't written in the first place.

Under what conditions do we _not_ want the default generator to be set?

(I think some things here could be clearer if we separated some of this world-configuration code from the UI callbacks, but that's more than I wanted to bite off at the moment.)

kBlaszczyk:
Well, I could reproduce the issue. I checked out your branch and could successfully create a new game. Then I removed that first change and could still successfully create a new game. If there isn't a weird side effect going on introduced by that change, I'd argue that the change isn't required.
Can you try verifying this in your build environment?

keturn:
There probably is a weird side effect! Did your config.cfg change?

engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/NewGameScreen.java
kBlaszczyk: I was able to reproduce the necessity of this change.

keturn:
Thanks for being so thorough in working through this one. That got tricky.

> […] a future issue when it causes some already specified world generator to be unwillingly replaced with the default one.

If you (or anyone else) _can_ come up with a scenario where that happens, do let me know. Makes me nervous to remove conditionals without a proper understanding of what they are there to guard against.

PR #4026, opened 2020-06-03
feat(JOML): Migrate MeshRenderer

pollend:
https://github.com/MovingBlocks/Terasology/pull/3950 broke compatibility with https://github.com/MovingBlocks/Terasology/pull/3865 this is a simplification that focuses on just the MeshRenderer. 

should be enough to verify by breaking blocks in the world. physics objects use the MeshRender and anything incorrect should be fairly noticeable. Carts, Items dropped, are a couple other objects that can be used to verify.  Testing this in an omega workspace and verifying it with a couple different modules should be sufficient.


engine/src/main/java/org/terasology/rendering/logic/MeshRenderer.java
```diff
@@ -218,17 +223,17 @@ private void renderEntitiesByMaterial(SetMultimap<Material, EntityRef> meshByMat
                             lastMesh.preRender();
                         }
 
-                        Matrix4f modelViewMatrix = MatrixUtils.calcModelViewMatrix(JomlUtil.from(worldRenderer.getActiveCamera().getViewMatrix()), matrixCameraSpace);
-                        MatrixUtils.matrixToFloatBuffer(modelViewMatrix, tempMatrixBuffer44);
-                        MatrixUtils.matrixToFloatBuffer(MatrixUtils.calcNormalMatrix(modelViewMatrix), tempMatrixBuffer33);
+                        modelViewMatrix.set(worldRenderer.getActiveCamera().getViewMatrix()).transpose().mul(matrixCameraSpace);
```
4Denthusiast, line 226: Why is there a `transpose` inserted here? This seems to be actually changing the logic.

pollend: the transpose is not needed anymore when I pull in the camera from upstream.

pollend:
@4Denthusiast since the camera PR was merged, I got rid of the transpose.

Cervator:
Tested a bit locally with this merged, seemed all good 👍

PR #4027, opened 2020-06-03
feat(JOML): Migrate ViewFrustum

pollend:


keturn:
Can we use [joml.FrustumIntersection](https://javadoc.io/static/org.joml/joml/1.9.25/org/joml/FrustumIntersection.html) instead of our own FrustumPlane class?

pollend:
I was looking at it a bit today and I couldn't quite get the frustrum to behave correctly. the way we have the viewFrustrum is kind of hard to follow since it reads it out of a Floatbuffer but it doesn't really need to which is kind of strange.

kBlaszczyk:
I'll have a look. I also think we should use the JOML class and technically you can simply set it up using a mat4. But I'll have to experiment with it too, to figure out how get the desired frustum out of that.

kBlaszczyk:
In regard to using the JOML class for frustum culling.
The main problem is that our view matrix is not an actual view matrix. I'm not sure what it is, but I pointed out in my Joml conversion branch, that the value we are calculating in `MatrixUtils` differs from the result of Joml's `setLookAt()` method.
We are calculating the view matrix from eye position (0, 0, 0) and then set the translation component to -eye.
When creating my own projection and view matrices, using Joml methods, I can successfully use `FrustumIntersection` for the intersection tests.
So, as we probably concluded before, it would be best to get rid of the implementations in `MatrixUtils` and use the appropriate Joml methods instead.

engine/src/main/java/org/terasology/rendering/cameras/ViewFrustum.java
```diff
@@ -147,35 +156,80 @@ public boolean intersects(AABB aabb) {
...
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Returns true if this view frustum intersects the given AABB.
+     */
+    public boolean intersects(AABBf aabb) {
```
kBlaszczyk, line 199: I think as long as `AABB` is being used, we should provide a converter method in `JomlUtil`.
That way we don't have to duplicate the logic like this.

engine/src/main/java/org/terasology/rendering/cameras/ViewFrustum.java
```diff
@@ -186,6 +240,9 @@ public boolean intersects(AABB aabb) {
 
     /**
      * Returns true if the given sphere intersects the given AABB.
+     *
+     * @deprecated This is scheduled for removal in an upcoming version method will be replaced with JOML implementation
+     *     {@link #intersects(Vector3fc, float)}.
      */
     public boolean intersects(Vector3f position, float radius) {
         for (int i = 0; i < 6; i++) {
```
kBlaszczyk, line 248: The deprecated method should redirect to the new one to get rid of duplication:
`return intersects(JomlUtil.from(position), radius);`

PR #4028, opened 2020-06-03
feat(JOML): Correct matrix output for Camera#getProjectionMatrix()

pollend:
termath and JOML matrices are transposes of each other and this PR tries to address the changes required to start migrating from one system to the other. This PR specifically focuses on correcting the ProjectionMatrix(). This is necessary to avoid additional matrix allocations to convert from one to the other. This affects the camera so any significant changes in these matrices should drastically affect rendering. 

This PR also requires merging the changes in this branch: https://github.com/Terasology/CoreRendering/pull/13

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/cameras/PerspectiveCamera.java
+++ b/engine/src/main/java/org/terasology/rendering/cameras/PerspectiveCamera.java
@@ -1,18 +1,5 @@
-/*
- * Copyright 2017 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// Copyright 2020 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
 package org.terasology.rendering.cameras;
 
 import org.joml.Matrix4f;
@@ -55,6 +42,9 @@ public class PerspectiveCamera extends SubmersibleCamera implements PropertyChan
     private DisplayDevice displayDevice;
 
     private Vector3f tempRightVector = new Vector3f();
+    //TODO: remove when projectionMatrix is corrected
+    private final Matrix4f transposeProjectionMatrix = new Matrix4f();
+
 
     public PerspectiveCamera(WorldProvider worldProvider, RenderingConfig renderingConfig, DisplayDevice displayDevice) {
         super(worldProvider, renderingConfig);
@@ -150,7 +140,8 @@ public class PerspectiveCamera extends SubmersibleCamera implements PropertyChan
 
         float aspectRatio = (float) displayDevice.getDisplayWidth() / displayDevice.getDisplayHeight();
         float fovY = (float) (2 * Math.atan2(Math.tan(0.5 * fov * TeraMath.DEG_TO_RAD), aspectRatio));
-        projectionMatrix.setPerspective(fovY, aspectRatio, getzNear(), getzFar()).transpose();
+        transposeProjectionMatrix.setPerspective(fovY, aspectRatio, getzNear(), getzFar());
+        transposeProjectionMatrix.transpose(projectionMatrix);
 
         viewMatrix = MatrixUtils.createViewMatrix(0f, bobbingVerticalOffsetFactor * 2.0f, 0f, viewingDirection.x, viewingDirection.y + bobbingVerticalOffsetFactor * 2.0f,
             viewingDirection.z, up.x + tempRightVector.x, up.y + tempRightVector.y, up.z + tempRightVector.z);
@@ -183,6 +174,12 @@ public class PerspectiveCamera extends SubmersibleCamera implements PropertyChan
         cachedReflectionHeight = getReflectionHeight();
 
         updateFrustum();
+    }
+
+    @Override
+    public Matrix4f getProjectionMatrix() {
+        //TODO: can use default superclass implementation when projectionMatrix is corrected
+        return transposeProjectionMatrix;
     }
 
     public void setBobbingRotationOffsetFactor(float f) {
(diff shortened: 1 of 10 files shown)


engine/src/main/java/org/terasology/rendering/cameras/PerspectiveCamera.java
```diff
@@ -137,23 +138,25 @@ public void updateMatrices() {
...
             return;
         }
 
         viewingDirection.cross(up, tempRightVector);
         tempRightVector.mul(bobbingRotationOffsetFactor);
 
-        projectionMatrix = createPerspectiveProjectionMatrix(fov, getzNear(), getzFar(),this.displayDevice);
+        float aspectRatio = (float) displayDevice.getDisplayWidth() / displayDevice.getDisplayHeight();
+        float fovY = (float) (2 * Math.atan2(Math.tan(0.5 * fov * TeraMath.DEG_TO_RAD), aspectRatio));
+        projectionMatrix.identity().perspective(fovY, aspectRatio, getzNear(), getzFar()).transpose();
```
keturn, line 153: does this end up being the same as joml's [perspectiveRect](https://javadoc.io/static/org.joml/joml/1.9.25/org/joml/Matrix4fc.html#perspectiveRect-float-float-float-float-boolean-org.joml.Matrix4f-)? or is the fovY we come up with different than the perspective it would set if we passed display width and height to perspectiveRect?

either way, I think we can use the [`setPerspective`](https://javadoc.io/static/org.joml/joml/1.9.25/org/joml/Matrix4f.html#setPerspective-float-float-float-float-boolean-) would let us avoid the need to reset the matrix to `identity` first.

keturn: also ... keeping the `.transpose` here but then re-transposing in the abstract `Camera.getProjectionMatrix`?

I guess the `projectionMatrix` property still needs to be transposed for some things, that's probably what you meant by things you need to correct in the camera later. But I lean toward adding an additional private field so we can store the non-transposed version instead of making a new object on every access.

engine/src/main/java/org/terasology/rendering/cameras/PerspectiveCamera.java
```diff
@@ -71,7 +72,7 @@ public boolean isBobbingAllowed() {
     @Override
     public void loadProjectionMatrix() {
         glMatrixMode(GL_PROJECTION);
-        GL11.glLoadMatrix(MatrixUtils.matrixToFloatBuffer(getProjectionMatrix()));
+        GL11.glLoadMatrix(getProjectionMatrix().get(BufferUtils.createFloatBuffer(16)));
```
keturn, line 75: If this is something that gets called every frame, I'm itching to optimize the memory access here too, but that doesn't have to be in this PR.

engine/src/main/java/org/terasology/rendering/cameras/PerspectiveCamera.java
pollend: JOMLUtil already makes new instances when it goes back and forth so I don't see too much of an issue here. If the transpose goes away then the new instance just needs to be removed. keeping an extra field seems like it will introduce a bit more work when I go back to remove it.

keturn:
Ah, if only I'd made it back to this sooner! I fear half the context has left my brain in the meantime.

Can we put that transpose call in updateMatrices like this, so it only needs to be done once per matrix update instead of once per get?

https://github.com/keturn/Terasology/commit/14be9e0b5539e227c75a14b0b70b1d50ba7f516d#diff-6a3dd314d84ecfeac039ea99375ec804R143-R144

I'm not sure where in the Camera hierarchy it needs to go. Does it need to be in the default implementation of `Camera` so it's applied to Orthographic Camera as well?

What about VR? (and are we maintaining VR or not?)

pollend:
> Ah, if only I'd made it back to this sooner! I fear half the context has left my brain in the meantime.
> 
> Can we put that transpose call in updateMatrices like this, so it only needs to be done once per matrix update instead of once per get?
> 
> [keturn@14be9e0#diff-6a3dd314d84ecfeac039ea99375ec804R143-R144](https://github.com/keturn/Terasology/commit/14be9e0b5539e227c75a14b0b70b1d50ba7f516d#diff-6a3dd314d84ecfeac039ea99375ec804R143-R144)
> 
> I'm not sure where in the Camera hierarchy it needs to go. Does it need to be in the default implementation of `Camera` so it's applied to Orthographic Camera as well?
> 
> What about VR? (and are we maintaining VR or not?)

I would rather keep it mostly the same so it is easier to correct later on. will probably make less sense much later on when we come back and correct the matrices.

keturn:
I had told myself I was going to stop trying to prematurely-optimize your PRs (...unless I actually had data to go with it, so it wouldn't be so premature), but since you added that field in response to the performance concern, it seemed relevant to bring up.

> I would rather keep it mostly the same so it is easier to correct later on.

Is that in response to the question about the other cameras, or my suggestion about where to move the transpose call?

It's a real small diff. I don't know how much more "mostly the same" one could make it. And if something changes later to remove the transpose in updateMatrices and leaves that private field unused, that should be pretty clear.

this is one of those times when I feel like there is something lost in transmission in the remote-pull-request-code-review process because it feels like we're not seeing the same thing 🙈

pollend:
Actually, It's not too bad. I'm over reacting a little.

pollend:
this will need to be merged in as well @keturn  https://github.com/Terasology/CoreRendering/pull/13

keturn:
the number of engine PRs we have going on these days that require accompanying module PRs does not leave me with happy thoughts about the engine / CoreRendering modularization split 😒 

How do we handle versioning for things like this? Does it mean that engine needs a major version bump, which that CoreRendering PR needs to depend on? Does that mean a major version bump for CoreRendering as well? Or is it only a minor version bump for CoreRendering because the public interface it exports doesn't change?

I guess we haven't actually published an engine 4.0 release yet, have we? so this could be one of the (several) things included in the engine v4 interface changes.

Cervator:
@keturn correct - engine v4 is still pending, so this is the time to break all the things then fix them :-)

We haven't previously been really hard on modules breaking their own API, and even the engine we're been soft about for a while after getting some initial practice. Alpha and all that + we've got some related tooling to do like the module release management.

So we need to be heading for somewhere better, I'm not too worried personally right now though. I'd be more eager just to finally cut down on some of all these engine PRs, assuming of course that they test out!

keturn:
skaldamar's just been talking about semver so much I keep trying to figure out how it would be applied to the changes I see and our twisty web of dependencies.

PR #4029, opened 2020-06-04
feat(JOML): Migrate ChunkMath

pollend:


engine/src/main/java/org/terasology/math/ChunkMath.java
```diff
@@ -64,35 +67,188 @@ public static int calcChunkPosZ(int z, int chunkPowerZ) {
...
     public static int calcChunkPosY(int y) {
         return calcChunkPosY(y, ChunkConstants.CHUNK_POWER.y);
     }
+
     public static int calcChunkPosZ(int z) {
         return calcChunkPosZ(z, ChunkConstants.CHUNK_POWER.z);
     }
 
+    /**
+     *
```
Cervator, line 80: Bonus for dropping in even a brief line at the start of the javadoc just to cover the method overall 👍

engine/src/main/java/org/terasology/math/ChunkMath.java
```diff
@@ -64,35 +67,188 @@ public static int calcChunkPosZ(int z, int chunkPowerZ) {
...
+     * @return
+     * @deprecated This is scheduled for removal in an upcoming version
+     *             method will be replaced with JOML implementation {@link #calcChunkPos(int, int, int, org.joml.Vector3i)}.
+     */
+    @Deprecated
+    public static Vector3i calcChunkPos(int x, int y, int z, Vector3i chunkPower) {
+        return new Vector3i(calcChunkPosX(x, chunkPower.x), calcChunkPosY(y, chunkPower.y), calcChunkPosZ(z, chunkPower.z));
+    }
+
+    //TODO: can be replaced by region3i iterator
```
Cervator, line 145: Should this be deprecated then, or briefly Javadoc'ed?

engine/src/main/java/org/terasology/math/ChunkMath.java
```diff
@@ -64,35 +67,188 @@ public static int calcChunkPosZ(int z, int chunkPowerZ) {
...
 
-    public static Vector3i calcChunkPos(int x, int y, int z, Vector3i chunkPower) {
-        return new Vector3i(calcChunkPosX(x, chunkPower.x), calcChunkPosY(y, chunkPower.y), calcChunkPosZ(z, chunkPower.z));
+    /**
+     * The position relative of the chunk with the given power ({@link ChunkConstants#POWER_X}, {@link ChunkConstants#POWER_Y}, {@link ChunkConstants#POWER_Z})
+     *
+     * <p>Chunk size is in powers of 2 (2, 4, 8, 16, ...)</p>
+     *
+     * @param pos absolute position of the block
+     * @param dest will hold the result
```
Cervator, line 156: Clarify? If `dest` is a param, and it gets used in a calculation, the result of which is returned while also matching `dest` ... why was the calculation needed? 🤔 Methinks that probably isn't accurate then?

engine/src/main/java/org/terasology/math/ChunkMath.java
```diff
@@ -64,35 +67,188 @@ public static int calcChunkPosZ(int z, int chunkPowerZ) {
...
+    public static org.joml.Vector3i calcChunkPos(Vector3ic pos, Vector3ic chunkPower, org.joml.Vector3i dest) {
+        return calcChunkPos(pos.x(), pos.y(), pos.z(), chunkPower.x(), chunkPower.y(), chunkPower.z(), dest);
+    }
+
+    /**
+     * The position relative to the size of chunk with the given chunk power
+     *
+     * <p>Chunk size is in powers of 2 (2, 4, 8, 16, ...)</p>
+     *
+     * @param x absolute x component of the block
```
Cervator, line 225: Is "component" a math term here? To me as a mundane develop it makes me think about our ES Components. If not a needed math term could it be named differently?

engine/src/main/java/org/terasology/math/ChunkMath.java
pollend: I write it to a destination variable so we can avoid an extra allocation. we use these functions in quite a few loops so reducing the hit to cache seems worth it.

engine/src/main/java/org/terasology/math/ChunkMath.java
pollend: I just follow what JOML is doing with referring to vector components. do you have a better name we can use? 

image

pollend:
@Cervator I don't think we have any tests for ChunkMah.

engine/src/main/java/org/terasology/math/ChunkMath.java
Cervator: So if that happens in a different method, I still think the documentation here is confusing. Because the logic clearly returns the result of a method call to which `dest` is passed - in the end it may well end up the same thing, but you aren't literally returning `dest` you're returning a calculation based on `dest` ?

If the goal is to avoid an extra allocation then you're achieving that in the other method. It doesn't really seem meaningful to tell the user about that here? Shouldn't you doc instead what `dest` _starts_ as, since that's what it is doing as a param?

engine/src/main/java/org/terasology/math/ChunkMath.java
Cervator: x y and z _coordinates_ ?

pollend:
dest is the vectors with the result being written to. so 
```
Vector3f temp = new Vector3f();
calcChunkPos(10,10,10,temp);
temp <- (10,10,10)
calcChunkPos(1,10,10,temp);
temp <- (1,10,10)
calcChunkPos(10,20,10,temp);
temp <- (10,20,10)
```

engine/src/main/java/org/terasology/math/ChunkMath.java
httpdigest: Mathemtically, it's more common to talk about "the components of a vector" when interpreting that vector as merely an n-tuple of numbers. It seems that when the components are interpreted as a position in some kind of coordinate system (like cartesian) then it's more common to talk about "coordinates".

keturn:
oh geez, just the amount of function overloading to support (vector, vector) vs (x, y, z, vector) and (vector, x, y, z) and float and int ...  smells like the sort of thing where it'd be really useful to have some bit of metaprogramming that could auto-generate the whole set of them for you. But I have no idea how that's done in Java.

(and I certainly wouldn't require it for PR approval. I'm just wishing out loud while thinking about how much work it must be to manually write all the variants and their docs.)

I guess the practical question is: do we need all these variants? I can imagine a use case for each of them, so if we do, then we do. I just want to know someone has checked.

engine/src/main/java/org/terasology/math/ChunkMath.java
keturn: From what I've seen of joml so far, this seems to be the common idiom. The current value of `dest` is irrelevant, it's just a mutable container for stuffing results in to.

I don't love it, because it feels like one of the bad parts of programming in C, but if the joml devs think it helps I find that all-too-believable.

(...if only Java could stack-allocate vectors. then we could just call functions, return results. wouldn't that be nice.)

engine/src/main/java/org/terasology/math/ChunkMath.java
```diff
@@ -30,79 +33,208 @@
...
+    /**
+     * The position of the chunk given the coordinate and size of chunk in powers of 2.
+     *
+     * @param pos absolute position of the block
+     * @param chunkPower the size of the chunk in powers 2
+     * @param dest will hold the result
+     * @return dest
+     */
+    public static org.joml.Vector3i calcChunkPos(Vector3fc pos, Vector3ic chunkPower, org.joml.Vector3i dest) {
+        return calcChunkPos(Math.roundHalfUp(pos.x()), Math.roundHalfUp(pos.y()), Math.roundHalfUp(pos.z()),
```
keturn, line 152: Are things pick about how these values are rounded?

I suggest making a test for it. As much for documentation as anything else, and it'll explicitly point out if this ever changes. It's one of those almost-too-simple-to-test things otherwise.

values like -1000f, -499.9f, -500.1f, -500.5f, -0f, 0f, 500.5f, 500.1f, -499.9f, 1000f
or somesuch for mix of positive, negative, above-half and below-half.

If they're _not_ picky about how they're rounded, x.intValue() might be ever-so-slightly cheaper. except that'll be always-towards-zero, which is probably worse for us than always-more-positive or always-more-negative, so, in conclusion, I'm over-thinking things again

engine/src/main/java/org/terasology/math/ChunkMath.java
Cervator: I might be dense, but then I still don't get it. Why have `dest` at all to begin with, if its value is irrelevant? Doesn't the user ultimately end up using the return value, even if that becomes the same value as `dest` in the end? Or is the user meant to subsequently use the variable they passed in as `dest` as it has changed-via-reference?

In either case the docs fail to enlighten this dummy here what's going on :-)

engine/src/main/java/org/terasology/math/ChunkMath.java
keturn: yeah, component was the standard term for this in my math/physics education. unfortunate terminology collision in an environment that works with entity-component-system so much.

engine/src/main/java/org/terasology/math/ChunkMath.java
keturn: a method that takes `dest` _and_ has a return value does have some redundancy in its signature.

returning void would be unambiguous. I expect the return value is provided for convenience so you can write things like `calculateChunkPos(pos, dest).mul(5)` instead of `calculateChunkPos(pos, dest);    dest.mul(5)`

this is a pattern that's going to show up all over the joml-using matrix and vector code, so I think documentation about the pattern would go in the wiki or the package-level docs of terasology.math, probably with references to things like

* https://github.com/JOML-CI/JOML/wiki/Design
* https://github.com/JOML-CI/JOML/wiki/Anti-Patterns-and-Best-Practices#favor-destination-arguments-over-allocating-new-objects

Cervator: So really the only reason to taking `dest` as a param is just to avoid allocating something new?

`* @param dest value ignored, but set to the same as the return value to avoid an allocation`

?

keturn: hmm, "set to the same as" makes it sound like dest and the return value are separate-but-equal things. But it's the same object.

mmm. I think I need another lunch before I try to think more about technical writing.

engine-tests/src/test/java/org/terasology/math/ChunkMathTest.java
```diff
@@ -93,4 +93,14 @@ public void testCalcChunkPos() {
         assertTrue(ChunkMath.calcChunkPos(10, 10, 10, temp).equals(0, 0, 0));
     }
 
+
+    @Test
+    public void testFloatingPointCalcChunkPos() {
```
keturn, line 98: That's some exactly-half, over-half, under-half, that's good.

Throw in at least one test with some negative coordinates, so we can tell if it's doing round-towards-zero or round-towards `-infinity`. What should that be... I guess x -0.1 is in chunk -1, along with x -63.9

x: -64 and -64.1 are chunk -2? bah, the fencepost errors are messing with me. I keep confusing myself, I should draw more diagrams.

engine-tests/src/test/java/org/terasology/math/ChunkMathTest.java
```diff
@@ -93,4 +93,14 @@ public void testCalcChunkPos() {
...
+        org.joml.Vector3i temp = new org.joml.Vector3i();
+        assertTrue(ChunkMath.calcChunkPos(31.9f, 64.1f, 32.5f, temp).equals(0, 1, 1), temp.toString());
+        assertTrue(ChunkMath.calcChunkPos(32.9f, 63.9f, 32.9f, temp).equals(1, 0, 1), temp.toString());
```
keturn, line 101: wait what ... ohhh, chunks aren't cubical, they're taller than they are wide? okay. then my previous comment about where the x-boundary for chunk -1 is was wrong.

pollend:
I tested this with an omega configuration and everything seems to be correct.

PR #4032, opened 2020-06-04
include dependency versions when defining a module's gradle dependencies

keturn:
When building a module, gradle was using the module's module.txt to get the _names_ of the module's dependencies, but it was ignoring the versions there.

This PR adds those version constraints to gradle's dependency information.

### How to test

Build modules from source!

Build modules from source that depend on other source modules in your workspace!

Throw some garbage (like git conflict markers) in your `modules.txt` file, make sure the error message is not misleading!

Build modules from source that depend on modules _not_ in your workspace to make it get their jar files from the repo!

Run your module tests!

I think this is pretty likely to break some builds. The question is, will those breakages turn out to reveal bad dependency data? or will those be signs of bugs in this PR?

### Outstanding

- [x] check error handling on malformed module.txt
- [ ] make sure modules with `isReleaseManaged: true` are handled appropriately, e.g. [CoreSampleGameplay](https://github.com/Terasology/CoreSampleGameplay/blob/develop/module.txt)


keturn:
Okay, I succeeded in building in an omega workspace. MTE tests are still pretty messed up for me, but things build.

keturn:
The astute reader may note that this adds another sixty-ish lines to the already sprawling terasology-module.gradle.kts file, instead of extracting independent pieces (like parsing of module JSON) out to more focused, manageable files.

This is because, as a precompiled script plugin with the "kts" suffix, there is gradle/kotlin-script magic going on in here that both (a) makes this whole thing usable to the gradle `plugin` block, and (b) applies this whole thing against the project that uses the plugin, so all sorts of stuff is already in scope (or evaluated against the project receiver, is maybe the more accurate way to say that).

So I've been putting off figuring out how to refactor it without breaking the automagic spell. It might turn out to be no big deal! But right now it's an unknown. One we'll have to deal with at some point.

skaldarnar (dismissed):
Came here from the bigger PR linking this, thought reviewing one hundred lines is easier than one thousand lines 🙈

keturn:
@4Denthusiast — as to why this wasn't merged on approval: It'll change how a bunch of versions are handled, which will break some dependencies as currently defined, so the thinking at the time was to roll it up with #4039 which will _also_ change version-dependency things.

That was stalled for a time to make sure the CI system would reliably tell us which things broke (e.g. #4047), but if the test flakiness is fixed work on #4039 should be able to continue.

maybe some labels on this issue should change somehow to reflect some of that?

PR #4033, opened 2020-06-05
fix(Particles): disable particles, when driver or GPU not supports OpenGL 3.3

DarkWeird:
### Contains

particles disabled when GPU or Driver not supports OpenGL 3.3 (Mesa driver issue, possible Mac OS)
Remove, when shaderland and rendering will work on OpenGL 3+


engine/src/main/java/org/terasology/engine/subsystem/lwjgl/LwjglGraphics.java
```diff
@@ -259,9 +259,6 @@ private void checkOpenGL() {
                 GLContext.getCapabilities().OpenGL15,
                 GLContext.getCapabilities().OpenGL20,
                 GLContext.getCapabilities().OpenGL21,   // needed as we use GLSL 1.20
-                GLContext.getCapabilities().OpenGL30,
```
kBlaszczyk, line 262: You have to remove the versions from `capabilityNames` below too.

PR #4038, opened 2020-06-06
Attach new Kotlin source for a Gradle build plugin used with modules

Cervator:
Fixes broken module builds after the `buildSrc` dir was added, but not made available to module.

Actually just one half of the fix, will push the other half to https://github.com/MovingBlocks/ModuleJteConfig after merging this and testing a bit more

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,4 @@
-node ("default-java") {
+node ("default-java || heavy-java") {
     stage('Checkout') {
         echo "Going to check out the things !"
         checkout scm
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@ node ("default-java") {
     stage('Build') {
         // Jenkins sometimes doesn't run Gradle automatically in plain console mode, so make it explicit
         sh './gradlew --console=plain clean extractConfig extractNatives distForLauncher'
-        archiveArtifacts 'gradlew, gradle/wrapper/*, templates/build.gradle, config/**, facades/PC/build/distributions/Terasology.zip, build/resources/main/org/terasology/version/versionInfo.properties, natives/**'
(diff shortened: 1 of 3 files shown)


Cervator:
Okay think this is done now. Had to fix another thing where the module plugin couldn't find Gradle plugins hosted by a third party. Also finally configured SpotBugs correctly while in there.

_Bizarrely_ there is a single test failure, which has varied across three branches/PRs built by 2 different Jenkins masters (Nanoware fork by jenkins.terasology.io, this one with the usual cbc url). First build was fine too in two cases but not with this PR. There's nothing in here that should be able to trigger those test failures.

* Nanoware/attachBuildSrc build 2: `testBlockingEventDuringReplay` -> `org.opentest4j.AssertionFailedError: expected: <3> but was: <0>`
* Nanoware/attachBuildSrc build 3: `testSendingEventAfterReplay()` -> `org.opentest4j.AssertionFailedError: expected: <4> but was: <0>`
  * Also happened for Nanoware/develop build 3, despite build 2 an hour and a half earlier being fine
* Terasology/this PR: `testReplayStatus` -> `org.opentest4j.AssertionFailedError: expected: <REPLAY_FINISHED> but was: <REPLAYING>`

Triggering one more Nanoware build and a rebuild of this PR and imagining each will trigger exactly one test failure 🤔 

On the plus side the module builds work fine again, and I've confirmed that the Spotbugs reporting actually running properly and being recorded does not lead to a bad commit status if problems are detected

Cervator:
I added a tiny tweak for the Jenkins migration so there's compatibility between two masters till we can retire older ones. Newest setup has more flexibility in build agent selection and capacity management.

This also passed without the mystery test failure, so it is all green now. However, a build of `develop` even _before_ this has been merged has tripped the same single test failure. Suspicious.

In any case: moving on. Actually getting close to something decent in Jenkinsland ...

Issue #4037, opened 2020-06-06
module version read by game should match version used by java package management


keturn:
Cervator uncovered the fact that modules are published with their version appearing like 0.2.0-SNAPSHOT according to the metadata on the Artifactory repository (used by gradle when retrieving dependencies), but contain a `module.txt` file with version 0.2.0 (no SNAPSHOT designation) as read by Terasology at runtime. (by the gestalt module system.)

This is, at the very least, a source of confusing and misunderstanding, and currently the prime suspect for some module resolution errors we've been troubled by.

Current plan to address this:

* no longer try to make "is this a snapshot?" decisions in the build configuration. The version in the module.txt in git will be the version used for the jar's published metadata. The module.txt inside the jar will remain unaltered from its source.

PR #4039, opened 2020-06-07
remove snapshot-or-not logic from gradle configuration

keturn:
This removes the code from the build configurations that would decide whether or not to call a module build a snapshot based on the current git branch, because #4037.

It also includes #4032, which is bad for separation of concerns, but good for reducing the number of times we have to go through the cycle of review and maybe-rebuild-a-lot-of-modules.

### How to test

build some modules,  
make sure the version in the name of the jar file matches the one in its source module.txt, whether or not it has -SNAPSHOT.

maybe run the `generatePomFile` task and look at the `<version>` element in there too.

there's also some stuff about choosing which repository we publish to that should probably be tested by someone who knows how the repositories work.

- [ ] https://github.com/MovingBlocks/Terasology/blob/ab028b2d5a8557b351e1c9a10ff051640ca8d176/config/gradle/publish.gradle#L36-L39

- [x] `cacheChangingModulesFor` setting?
  - [x] in root: https://github.com/MovingBlocks/Terasology/blob/ab028b2d5a8557b351e1c9a10ff051640ca8d176/build.gradle#L179-L182
  - [x] in engine https://github.com/MovingBlocks/Terasology/blob/ab028b2d5a8557b351e1c9a10ff051640ca8d176/engine/build.gradle#L87-L89
  - [x] in modules: https://github.com/MovingBlocks/Terasology/blob/ab028b2d5a8557b351e1c9a10ff051640ca8d176/buildSrc/src/main/kotlin/terasology-module.gradle.kts#L88-L90
- [x] snapshot determination for engine and engine-test? https://github.com/MovingBlocks/Terasology/blob/012985ccb97b6872af9d5bfdb3c589845ab98605/engine/build.gradle#L241-L243


keturn:
wait, hubris, I haven't actually tested manually putting `-SNAPSHOT` in my module.txt with this

I might have pruned too much

keturn:
Oh, okay, I did _not_ completely break the ability to build a snapshot release. :relieved:

skaldarnar:
Where to publish to should probably be derived from whether the version number is a pre-release according to Semver spec or not. I guess for now it suffices that we check whether the version ends on `-SNAPSHOT`.

I'm wondering what impact this has on our module builds? 🤔 One thing I dislike is that we recently bumped all the version of Omega modules to not include the SNAPSHOT tag anymore - should we re-add that? I think we have to, otherwise we'll need to change the version with every build...

buildSrc/src/main/kotlin/terasology-module.gradle.kts
```diff
@@ -100,21 +83,49 @@ configure<SourceSetContainer> {
...
+            version = gestaltDep.versionRange().toString()
+        )
+
+        if (gestaltDep.isOptional) {
+            // `optional` module dependencies are ones it does not require for runtime
+            // (but will use opportunistically if available)
+            compileOnly(gradleDep) { isChanging = true }
+            // though modules also sometimes use "optional" to describe their test dependencies;
+            // they're not required for runtime, but they *are* required for tests.
+            testImplementation(gradleDep) { isChanging = true }
```
skaldarnar, line 125: We should probably introduce pure test dependencies in module land... a task for another day ^^

config/gradle/publish.gradle
```diff
@@ -46,12 +33,10 @@ publishing {
                             deducedPublishRepo = "terasology"
                         }
 
-                        // Check the active Git branch and some module logic to see whether we're doing a release or snapshot
-                        String gitBranch = System.getenv()["BRANCH_NAME"]
-                        if (isMaster(gitBranch) && !shouldBypassModuleRelease()) {
-                            deducedPublishRepo += "-release-local"
-                        } else {
+                        if (project.version.toString().endsWith("-SNAPSHOT")) {
```
skaldarnar, line 36: I somehow like that this code is gone now 🙃 still being a bit afraid of what happens to module build chaos 🙈

skaldarnar (dismissed):
Overall looks good from my side, but please have a look at @DarkWeird's feedback.

jdrueckert:
@keturn @Cervator what's the status of this? Is there anything outstanding other than the "comments that make it sound like they aren't meant to stick around" that keturn wanted to reconsider?

keturn:
I don't see anything else outstanding, no. In light of Cerv's [earlier comment](https://github.com/MovingBlocks/Terasology/pull/4039#issuecomment-678582262), I'm inclined to go with the "It might be worth getting rid of it just to reduce complexity," option. Maybe that involves a note somewhere that says "use `gradlew --refresh-dependencies`" in the module development docs if we can figure out which cases need it.

(I suspect it's something like "your in-development module depends on another in-development module but you don't have the source checked out for it.")

Overall, I was leaving this on the back burner while waiting for some other break-the-world stuff to settle down, because this probably will require rebuilding modulespace. But as far as I know, the case for it is still as valid as when we came up with it, and I don't think anything we've done in the mean time has changed that. It'd be great to merge as soon as it seems practical to do so.

Cervator:
Looked at this a bit in light of recent discussion on Discord - this is looking good and I don't think any changes to the Jenkinsfile would really be needed. With this merged and the Gradle block changed from

```gradle
  String gitBranch = System.getenv()["BRANCH_NAME"]
  if (isMaster(gitBranch) && !shouldBypassModuleRelease()) {
      deducedPublishRepo += "-release-local"
  } else {
      deducedPublishRepo += "-snapshot-local"
  }
```

To

```gradle
if (project.version.toString().endsWith("-SNAPSHOT")) {
      deducedPublishRepo += "-release-local"
} else {
      deducedPublishRepo += "-snapshot-local"
}
```

Then we've got the new approach that bases target repo in Artifactory on version instead of branch.

The Jenkinsfile either old school or new school just determines _whether_ to publish at all.

Old school (engine, scripted pipeline):

```groovy
boolean specialBranch = env.BRANCH_NAME.equals("master") || env.BRANCH_NAME.equals("develop")

...

stage('Publish') {
    if (specialBranch) {
```

New school (CrashReporter, declarative pipeline)

```groovy
stage('Publish') {
    when {
        anyOf {
            branch 'master'
            branch pattern: "release/v\\d+.x", comparator: "REGEXP"
        }
    }
```

The difference there is just `master` / `develop` publish vs `master` / `release...` pattern, which I call the "single branch" approach as all snapshots **and** releases come from `master` (or eventually a renamed `main`) - it just _also_ happens to support doing the same for parallel release streams if needed, but that's a new extra not a necessary second branch set, which in most cases wouldn't exist.

We don't need that distinction yet, although we might want to stay on top of not leaking snapshots into `master` or doing releases from `develop` - just by making sure we only push the final commit to the right branch. Wouldn't even be the end if we did, the artifacts would go to the right place in Artifactory, just the Omega zips might end up building against the wrong engine version until it no longer selects by copying files attached to jobs in Jenkins.

To re-highlight how that works:

* When an Omega zip job runs it will copy _engine_ level files from the job in Jenkins that produced them.
  * A release zip will copy from the job that built the `master` branch
  * A dev build zip will copy from the job that built the `develop` branch

Only difference now would be if you pushed a release to `develop` you might get confused if a release Omega would still copy files from the `master` branch job in Jenkins, not the very latest released binaries in Artifactory. Hopefully we can sort that out and simplify soon ™️

PR #4017, opened 2020-06-01
Moved NUI into a separate library

jdrueckert:
Belongs to #2307 

Base PR introducing the **NUI extraction** by @BenjaminAmos in #3819 .

This PR adds a feature branch for joint effort to get this in.

---

## How to run ...

To test this PR with some gameplay you'll need to checkout the respective NUI extraction branches for the Iota module line-up.
Information about getting suitable versions of the modules to test with is in [the description of 4048](https://github.com/MovingBlocks/Terasology/pull/4048#issue-434043679).

Check out this PR for the engine (e.g., with Github CLI tool):
```
gh pr checkout 4017
```

### ... with TeraNUI Artifact

TeraNUI v1.x is automatically published from MovingBlocks/TeraNUI now. Thus, testing this PR should work without a local copy of TeraNUI. This will use the published artifact as specified in the engine's `build.gradle`

### ... with TeraNUI from source

Get TeraNUI as source dependency via

  ```sh
  groovyw lib get TeraNUI
  # check out v1 branch
  cd libs/TeraNUI
  git checkout release/v1.x
  ```

Other notes including **test plans** are [in 3819](https://github.com/MovingBlocks/Terasology/pull/3819#issue-358430014).

---

Co-authored-by: Benjamin Amos <benjamin.amos11@gmail.com>
Co-authored-by: Cervator <cervator@gmail.com>
Co-authored-by: Kevin Blaszczyk <OrcHound@gmail.com>
Co-authored-by: Kevin Turner <83819+keturn@users.noreply.github.com>
Co-authored-by: Michael <polli104@mail.chapman.edu>
Co-authored-by: Rostyslav Zatserkovnyi <zatserkovnyi.rostyslav@gmail.com>
Co-authored-by: Tobias Nett <skaldarnar@googlemail.com>

PR #4045, opened 2020-06-10
i18n: translatable compound messages

keturn:
This allows translatable strings to contain format arguments. It enables compound messages:

translation resources:
> "tree-punch": "You punch the {0} tree.",
> "tree-species-fir": "fir",
> "tree-species-palm": "palm",

application code:
```java
translate("${#tree-punch}", tree.getSpecies())
```

result:
> _You punch the palm tree._

(This would be a much more compelling demonstration if I knew how to talk about trees in more than one language.)

For this it uses the [MessageFormat class](https://docs.oracle.com/javase/tutorial/i18n/format/messageFormat.html) from the Java library. That also lets it support things like locale-aware date and number formatting. (and probably pluralization? but that interface looks messy and I haven't tested it.)

### How to test

The game-loading error message from #4004 has a compound message. If you know any other languages, you could provide their `.lang` files with a translation for `load-game-error`.

Oh, and check to see that the _existing_ text isn't broken. That's probably the more important thing.

- [ ] Does this break API compatibility? It adds varargs to several interfaces marked as public `@API`. The changes are _source_ compatible, because it's fine to pass zero varargs, but are they binary compatible?
- [ ] unit tests
- [ ] Will need developer documentation. We have a mishmash of the standard Java interface and homegrown formats.


engine/src/main/java/org/terasology/engine/modes/StateLoading.java
```diff
@@ -243,8 +232,16 @@ public void update(float delta) {
         EngineTime time = (EngineTime) context.get(Time.class);
         long startTime = time.getRealTimeInMs();
         while (current != null && time.getRealTimeInMs() - startTime < 20 && !gameEngine.hasPendingState()) {
-            if (current.step()) {
-                popStep();
+            try {
+                if (current.step()) {
+                    popStep();
+                }
+            } catch (Exception e) {
```
skaldarnar, line 239: Do we have more concrete information about the errors that can appear here? Catching the general `Exception` class is kind of broad, we might want to refine that if possible.

engine/src/main/java/org/terasology/engine/modes/StateMainMenu.java
```diff
@@ -136,7 +129,9 @@ public void init(GameEngine gameEngine) {
         //guiManager.openWindow("main");
         context.get(NUIManager.class).pushScreen("engine:mainMenuScreen");
         if (!messageOnLoad.isEmpty()) {
-            nuiManager.pushScreen(MessagePopup.ASSET_URI, MessagePopup.class).setMessage("Error", messageOnLoad);
+            TranslationSystem translationSystem = context.get(TranslationSystem.class);
```
skaldarnar, line 132: Check if `translationSystem` is `null`?

engine/src/main/java/org/terasology/engine/modes/StateMainMenu.java
```diff
@@ -136,7 +129,9 @@ public void init(GameEngine gameEngine) {
...
+            MessagePopup popup = nuiManager.pushScreen(MessagePopup.ASSET_URI, MessagePopup.class);
+            popup.setMessage("Error", translationSystem.translate(messageOnLoad, messageOnLoadArguments));
```
skaldarnar, line 134: Is "Error" the title of the popup? Should that be translated, too?

engine/src/main/java/org/terasology/i18n/TranslationProject.java
```diff
@@ -44,9 +31,10 @@
      * If no perfect match is found for the given locale, fallback strategies will attempt to find the closest match.
      * @param id the id of the string to translate (without project reference).
      * @param locale the target locale
+     * @param arguments {@link java.text.MessageFormat} arguments for the translation string
      * @return the translated string
      */
-    Optional<String> translate(Name id, Locale locale);
+    Optional<String> translate(Name id, Locale locale, Object... arguments);
```
skaldarnar, line 37: As you said this is definitely source compatible - I'd say time will tell if it is binary compatible 🙈

engine/src/main/resources/assets/i18n/menu.lang
```diff
@@ -221,6 +221,7 @@
     "listed-servers": "listed-servers",
     "load-delay-warning": "load-delay-warning",
     "load-game": "load-game",
+    "load-game-error": "load-game-error",
```
skaldarnar, line 224: Should the placeholder/dev translation in `menu.lang` indicate that this string is supposed to be a composed message?
```suggestion
    "load-game-error": "load-game-error ({0})",
```

engine/src/main/java/org/terasology/i18n/TranslationSystem.java
```diff
@@ -37,18 +24,21 @@
 
     /**
      * If no perfect match is found for the default locale, fallback strategies will attempt to find the closest match.
-     * @param id must match the pattern <code>${module:project#id}</code>
+     * @param id must match the pattern {@code ${module:project#id}}
+     * @param arguments {@link java.text.MessageFormat} arguments for the translation string
      * @return the translated string
      */
-    String translate(String id);
+    String translate(String id, Object... arguments);
```
skaldarnar, line 31: The NUI PR #4017 touches this file, but it does not remove it. I think the changes should be compatible without major disruption here.

engine/src/main/java/org/terasology/i18n/TranslationSystemImpl.java
```diff
@@ -90,31 +77,72 @@ public TranslationProject getProject(Uri name) {
...
+     * <p>
+     * Example:
+     * <pre>{@code
+     *     Object[] arguments = {17, "${engine:menu#credits}", new Date(16e11)};
+     *     Object[] expandedArguments = expandFormatArguments(locale, arguments);
+     *
+     *     println(expandedArguments)  // {17, "Credits", Date(16e11)}
+     * }</pre>
+     *
+     * @param otherLocale FIXME: why <em>is</em> this "other" locale and not just "locale"?
```
skaldarnar, line 108: Probably the class had a member called `locale` at some point and Checkstyle would warn if you'd overload the name in a non-setter context? 🤷‍♂️

engine/src/main/java/org/terasology/i18n/TranslationSystemImpl.java
```diff
@@ -90,31 +77,72 @@ public TranslationProject getProject(Uri name) {
...
+        SimpleUri projectUri = new SimpleUri(uri.getModuleName(), uri.getResourceName());
+        TranslationProject project = getProject(projectUri);
+        if (project != null) {
+            Optional<String> opt = project.translate(uri.getFragmentName(), otherLocale, arguments);
+            if (opt.isPresent()) {
+                return opt.get();
+            } else {
+                logger.warn("No translation for '{}'", id);
+                return "?" + uri.getFragmentName() + "?";
+            }
```
skaldarnar, line 141: gnarf, this usage of `Optional` is kind of opposite to idiomatic 🧐  Waiting for [`ifPresentOrElse`](https://docs.oracle.com/javase/9/docs/api/java/util/Optional.html#ifPresentOrElse-java.util.function.Consumer-java.lang.Runnable-) to be available in the engine...

What do you think about the following?

```suggestion
            Optional<String> maybeTranslation= project.translate(uri.getFragmentName(), otherLocale, arguments);
            if (!maybeTranslation.isPresent()) {
                logger.warn("No translation for '{}'", id);
            }
            return maybeTranslation.orElse("?" + uri.getFragmentName() + "?");
```

engine/src/main/resources/assets/i18n/menu.lang
keturn: _something_ should, whatever the something is that gets shown to the translators.

In the translation systems I've seen previously, that something is the program's "native" language (i.e. English for any project I've worked on);.

I don't know what the keys in this menu.lang map are used for.

engine/src/main/java/org/terasology/i18n/TranslationSystemImpl.java
keturn: That is another way to write it. The variable name is better, but aside from that it doesn't add any clarity for me. Using an if-statement for the logger and then `.orElse` for the return value is two ways of testing for the same condition, which is unappealing.

By its signature, this method has to unwrap the Option somehow, I don't know if it has anything to gain from trying to do functional composition.

engine/src/main/java/org/terasology/engine/modes/StateMainMenu.java
keturn: Oh, now that you mention it, I suppose it probably is.

engine/src/main/java/org/terasology/engine/modes/StateLoading.java
keturn: Being broad is kinda the point of #4004!  Oh, I failed to mention that this PR is based on that one, didn't I. Oops. Anyway, I think that's where this thread should happen.

engine/src/main/java/org/terasology/i18n/TranslationSystemImpl.java
skaldarnar: Yeah, I know 😔 I'd like to chain somewhat like this:

```java
project.translate(uri.getFragmentName(), otherLocale, arguments)
  .ifEmpty(() -> logger.warn("No translation for '{}'", id))
  .orElse("?" + uri.getFragmentName() + "?");
```

But I agree, this `if-else` is probably a better solution here for the sake of readability...

... oh, wait:

```suggestion
            project.translate(uri.getFragmentName(), otherLocale, arguments)
            		.orElseGet(() -> {
            				logger.warn("No translation for '{}'", id);
							return "?" + uri.getFragmentName() + "?";
            		});
```

keturn:
reminder: merging #4004 will make it much clearer what's actually in this branch

engine/src/main/java/org/terasology/i18n/TranslationSystem.java
```diff
@@ -39,18 +26,31 @@
...
+
+    @Override
+    default String translate(String id) {
+        return translate(id, new Object[0]);
+    }
+
+    @Override
+    default String translate(String id, Locale locale) {
+        return translate(id, locale, new Object[0]);
+    }
```
keturn, line 53: After the NUI Translate interface was moved to an external library, I had to add these default implementations that delegate to our varargs versions.

It's not a great look.

Cervator: That's not terrible :) Does that mean the third checkbox can be checked? Since we're now post-NUI extraction

keturn:
Oh, this should really be in draft state until it has those tests and docs ready.

With those in place, it wouldn't hurt to merge this. The API being compatible-enough, providing a way to do a little more with only a minor change from the current methods.

but honestly, now that NUI is split out, I think it's worth a closer look at this interface in Terasology's code and how it's split between Terasology and NUI and how internationalizable resources work in NUI. Cuz this looks kludgy, especially after the NUI split, and if there's no rush to get this in it's worth a little more thought.

PR #4046, opened 2020-06-13
make `libs/` use included builds instead of subprojects

keturn:
This treats subdirectories of `libs/` as [included builds of a composite build](https://docs.gradle.org/6.4.1/userguide/composite_builds.html) instead of [subprojects of the multi-project build](https://docs.gradle.org/6.4.1/userguide/multi_project_builds.html).

This works mostly the same but keeps them a little more independent of each other.

This  PR also renames the files that are included by `/settings.gradle` so they also end in `.settings.gradle` to distinguish them from other files that are used after the settings phase is complete.

### How to test

Could use input here from people who have used this `libs/` setup in the past!

I used it with the [extracted NUI library](https://github.com/MovingBlocks/TeraNUI/pull/2#issuecomment-643520647) but that's the only thing I've tried it on.


keturn:
This has been superseded by #4082, but it's worth checking to see if there are any differences worth combining.

Cervator:
Changes combined, thank you very much as usual 👍

PR #4050, opened 2020-06-14
feat: add BlockArea to replace Rect2i

pollend:
This implements the same pattern for BlockRegion.


engine/src/main/java/org/terasology/world/block/SquareRegion.java
```diff
@@ -0,0 +1,671 @@
...
+import org.joml.Vector2fc;
+import org.joml.Vector2i;
+import org.joml.Vector2ic;
+import org.joml.Vector3ic;
+
+/**
+ * is a bounded rectangle describing squares contained within.
+ * A {@link SquareRegion} is described and backed by an {@link org.joml.Rectanglei}
+ */
+public class SquareRegion {
```
skaldarnar, line 20: "Square" in the sense of 90° angles, or in the sense of a quadratic rectangle with all four sides of equal length?
I'd vote to call this `RectangularRegion` for clearer naming and to avoid confusion.

skaldarnar:
Thanks for renaming this to `RectangularRegion`.

Please excuse if I have asked this before, but what exactly is the purpose of `RectangularRegion` and `BlockRegion`? What do they add what the backing JOML classes don't give us?

pollend:
the BlockRegion has a bounded volume in space and an upper abstraction that describes a region containing blocks. There are specific methods I added for interacting with blocks and you can use the aabb if you want to check for lets say intersection with the player or another object. Rectangular region is just a continuation of the pattern established in BlockRegion. you kind of get the best of both one for interacting with blocks and the aabb describing an intersecting volume.

4Denthusiast:
I can sort of see what you're going for with the "+1"s and "-1"s, but I think the distinction between this and the Rectanglei it's based on is likely to cause confusion. Also, the interface is somewhat inconsistent even with itself. For example, if you create a new RectangularRegion with arguments min and max, then call getMax, you don't get back what you passed in, and if you call setMax with the same maximum it was constructed with, it changes. It would be more consistent if all of the methods and constructors that interact with the max coordinates were offset from those of the underlying Rectanglei, so that it behaved equivalently to a Rectanglei except that the methods relating to intersections of the shape with other shapes are offset relative to the methods that directly deal with coordinates.

Mostly I think that RectangularRegion (and BlockRegion too) shouldn't exist. It will be at least as much effort to avoid off-by-one errors when using them as when just directly using Rectanglei and AABBi.

pollend:
The other was around this was to just setup a utility module for things like this. This looked like a good middle ground between implementing two versions of an AABB.

pollend:
The goal of having this level of indirection for AABBi and Rectangle is to allow for consistency. BlockRegion backs a joml AABBI and RectangularRegion backs a Rectanglei. There are some adjustments we would like to have that are specific to BlockRegion (
https://github.com/MovingBlocks/Terasology/issues/4086). I think we want to have some similar methods in RectangularRegion we should apply some similar changes for consistency. There is some concern that this level of indirection would add additional confusion to how these things works in the engine.

jdrueckert:
@pollend is this still WIP as the label suggests?

skaldarnar:
@pollend I'm still confused about the name. I thought `BlockRegion` encodes two things in its name:
- `Block` - we're talking about the block-space of things, e.g., block coordinates. The block at coordinate (0, 0) occupies the 3d space between (0,0) and (1,1), with it's center being at (0.5, 0.5).
- `Region` - we're talking about a "region", i.e., a bounded volume in space. Basically, we mean AABB by it, in some way.

This leads me to the question what `RectangularRegion` encodes in its name. It does not say anything about block-space, but it uses the same word "Region". However, here the region is only 2d... I'd propose to use either pair of the following:
- 🅰️ `BlockRegion` (3d axis-aligned bounded volume in block-space) and `BlockArea` (2d axis-aligned bounded area in block-space)
- 🅱️ `BlockRegion3` and `BlockRegion2`

@4Denthusiast I think the discussion whether we should have these classes or not boils down to the same problem the JOML maintainer has with the integer-based primitives. The exact semantics is somewhat subjective, and there's probably no one-size-fits-all. 
They've recently extracted the primitives from the main library, and they try to encourage users of JOML to build their own primitives specific to their usage scenarios. I think that our block-based semantics would qualify for custom implementations (under whatever name).

pollend:
Blockarea sounds more correct. Blockregion2 lines up better semantically with the rest of the joml stuff.

jdrueckert:
> Blockarea sounds more correct. Blockregion2 lines up better semantically with the rest of the joml stuff.

@pollend So... for which one do we go? :sweat_smile:

engine/src/main/java/org/terasology/world/block/BlockArea.java
```diff
@@ -0,0 +1,622 @@
...
+/**
+ * A bounded axis-aligned rectangle of blocks.
+ * <p>
+ * A block area is described and backed by an {@link org.joml.Rectanglei}
+ *
+ * @see BlockAreas
+ */
+public class BlockArea {
+    private final Vector2i min = new Vector2i(Integer.MAX_VALUE);
+    private final Vector2i max = new Vector2i(Integer.MIN_VALUE);
```
skaldarnar, line 30: @pollend Is this too much overhead keeping the vectors? Should I instead use just 4 plain integers?

engine/src/main/java/org/terasology/world/block/BlockArea.java
```diff
@@ -0,0 +1,622 @@
...
+
+    private boolean dirty = true;
+    private Rectanglef rectangle = new Rectanglef();
```
skaldarnar, line 33: @pollend Should we compute the bounding rectangle every time something changes, or just when it is requested (lazy)? I've prepared to do the latter, not fully implemented yet, though.

engine/src/main/java/org/terasology/world/block/BlockArea.java
pollend: I would just use 4 integers.

engine/src/main/java/org/terasology/world/block/BlockArea.java
pollend: I would just update the bounds if it changes. you still end up needing to recalculate it for most cases.

4Denthusiast: I would expect this to be used sufficiently rarely that just not cacheing at all would be best.

skaldarnar: If we're going to re-compute it anyways, maybe make it more explicit? Then the caller can do whatever they want with it....

```java
Rectanglef worldArea = BlockAreas.toWorldArea(area);
```

engine/src/main/java/org/terasology/world/block/BlockArea.java
```diff
@@ -0,0 +1,508 @@
...
+        return dest.set(getSizeX(), getSizeY());
+    }
+
+    /**
+     * The number of blocks on the X axis
+     *
+     * @return number of blocks in the X axis
+     */
+    public int getSizeX() {
+        return this.maxX - this.minX;
```
pollend, line 236: for the size you need to add 1.

engine/src/main/java/org/terasology/world/block/BlockArea.java
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    /**
+     * The number of blocks on the Y axis
+     *
+     * @return number of blocks in the Y axis
+     */
+    public int getSizeY() {
+        return this.maxY - this.minY;
```
pollend, line 245: you need to add 1 for the size so its consistent with BlockRegion.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -518,6 +519,43 @@ public boolean containsPoint(Vector3fc point) {
...
+
+    /**
+     * Test whether the given {@link AABBd} lies inside the {@link BlockRegion}
+     * @param value the aabb to test
+     * @return <code>true</code> iff the given value lies inside this {@link BlockRegion}; <code>false</code> otherwise
+     */
+    public boolean containsAABB(AABBd value) {
+        return this.aabb.containsAABB(value);
+    }
+
```
pollend, line 558: lets wait on this changes after this https://github.com/MovingBlocks/Terasology/pull/4321

skaldarnar:
Now, let's also wait for https://github.com/MovingBlocks/Terasology/pull/4326 here 😉

engine/src/main/java/org/terasology/world/block/BlockRegion.java
skaldarnar: I'll remove these changes from this PR as they are not related to `BlockArea`.

skaldarnar (approved):
Yes, I think it is really good. I did a good job! Very well done, me, myself and I (_pads own shoulder_)... 👌 

Honestly, can somebody else have a quick look at this? 😅

pollend (commented):
looks good

PR #4051, opened 2020-06-14
feat(JOML): partial migrate of LocationComponent

pollend:
I migrate over LocationComponent along with some changes in Direction. The old methods are mapped onto the replacement methods and marked as deprecated. I was able to verify this with an omega workspace and nothing significant was impacted with these changes. 

 It should be easy to verify if this works correctly and at some point the internal implementation can be swapped over. The test that verify LocationComponent have been migrated over.  LocationComponent is used throughout the engine so testing this with a couple different modules should be sufficient. 

Migrated methods in LocationComponent:
- setWorldPosition(Vector3f)
- setWorldRotation(Quat4f)
- getWorldRotation(Quat4f)
- getWorldRotation()
- getWorldDirection()
- getWorldPosition(Vector3f)
- getWorldPosition()
- getLocalDirection()
- setLocalPosition(Vector3f)

I would like to set getLocalRotation and getLocalPosition as readonly. The set methods store the previous local rotation and position.  These methods are used for the LocationChangedSystem but a boolean flag seems more appropriate.


DarkWeird (changes requested):
nitpics about code :3

This is builds on omega modules space and seems, it is works right (player moving, block item falling, respawning(teleportation))

DarkWeird (approved):

PR #4059, opened 2020-06-21
Improve Simplex and Perlin noise generators

sin3point14:
### Contains

More control on Simplex and Perlin noise generators, Adds Docs to use them as Tileable noise generators

### How to test

Run these two branches of the module I made for testing, use the world generator by the name of "oki" in game
https://github.com/sin3point14/WorldGenTesting/tree/simplex-test
https://github.com/sin3point14/WorldGenTesting/tree/perlin-test

You should see flanked mountains like this
image
the "flankiness" will be controllable by changing `SurfaceProvider.gridSize` in the module


.idea/misc.xml
```diff
@@ -12,5 +12,11 @@
     </writeAnnotations>
   </component>
   <component name="ExternalStorageConfigurationManager" enabled="true" />
+  <component name="FrameworkDetectionExcludesConfiguration">
+    <file type="web" url="file://$PROJECT_DIR$" />
+  </component>
+  <component name="JavaScriptSettings">
```
skaldarnar, line 18: @keturn we'll see more and more of these unrelated changes, right? Ultimate vs Community edition, as well as plugins installed in addition to the recommended ones will add to these files if I get it right... 🙄

engine/src/main/java/org/terasology/utilities/procedural/PerlinNoise.java
```diff
@@ -27,38 +27,50 @@
...
     private final int[] noisePermutations;
+    private final int permCount;
 
     /**
      * Init. a new generator with a given seed value.
      *
      * @param seed The seed value
      */
     public PerlinNoise(long seed) {
+        this(seed, 1 << 8);
```
skaldarnar, line 38: Is there a specific reason for writing `1 << 8` instead of `256`? Is the grid dimension required to be a power of 2?

engine/src/main/java/org/terasology/utilities/procedural/PerlinNoise.java
```diff
@@ -71,9 +83,9 @@ public PerlinNoise(long seed) {
      */
     @Override
     public float noise(float posX, float posY, float posZ) {
-        int xInt = (int) TeraMath.fastFloor(posX) & 255;
-        int yInt = (int) TeraMath.fastFloor(posY) & 255;
-        int zInt = (int) TeraMath.fastFloor(posZ) & 255;
+        int xInt = Math.floorMod((int) TeraMath.fastFloor(posX), permCount);
```
skaldarnar, line 86: Why do we `floor` two times here, or what exactly does `floorMod` do?

engine/src/main/java/org/terasology/utilities/procedural/SimplexNoise.java
```diff
@@ -62,42 +62,57 @@
...
+     * @see SimplexNoise#tileable1DMagicNumber
      *
      * @param seed a seed value used for permutation shuffling
+     * @param gridDim gridDim x gridDim will be the number of squares in the square grid formed after skewing the simplices belonging to once "tile"
      */
     public SimplexNoise(long seed) {
+        this(seed, 8);
+    }
+
+    public SimplexNoise(long seed, int girdDim) {
```
skaldarnar, line 87: Typo
```suggestion
    public SimplexNoise(long seed, int gridDim) {
```

engine/src/main/java/org/terasology/utilities/procedural/SimplexNoise.java
```diff
@@ -62,42 +62,57 @@
...
+    /**
+     * Initialize permutations with a given seed and grid dimension.
+     * Supports 1D tileable noise
+     * @see SimplexNoise#tileable1DMagicNumber
      *
      * @param seed a seed value used for permutation shuffling
+     * @param gridDim gridDim x gridDim will be the number of squares in the square grid formed after skewing the simplices belonging to once "tile"
      */
     public SimplexNoise(long seed) {
+        this(seed, 8);
```
skaldarnar, line 84: Why does Perlin noise have a grid dimension of `1 << 8` and Simplex noise a grid dimension of `8`?

engine/src/main/java/org/terasology/utilities/procedural/SimplexNoise.java
```diff
@@ -154,8 +169,8 @@ public float noise(float xin, float yin) {
         float y2 = y0 - 1.0f + 2.0f * G2;
 
         // Work out the hashed gradient indices of the three simplex corners
-        int ii = i & 255;
-        int jj = j & 255;
+        int ii = i % permCount;
```
skaldarnar, line 172: In the Perlin implementation you did some combination of `floorMod` and `floor`  - why is a simple modulo operation sufficient here, and why is it not sufficient in Perlin implementation?

engine/src/main/java/org/terasology/utilities/procedural/PerlinNoise.java
sin3point14: Ah nothing it think this is copypasta effects?
Anyways fixing this

engine/src/main/java/org/terasology/utilities/procedural/PerlinNoise.java
sin3point14: `%` doesn't handle negative numbers well, for those cases `floorMod` is needed and `fastFloor` converts a float to int while always reducing it casting to `int` will give issues- `(int)-7.65 == -7` while `-8` is needed here

engine/src/main/java/org/terasology/utilities/procedural/SimplexNoise.java
sin3point14: Oh this was a bad miss, fixing this rn thanks!

engine/src/main/java/org/terasology/utilities/procedural/PerlinNoise.java
sin3point14: ok in the simplex implementation I found the better version to use here `floorToInt`, swapping it in

engine/src/main/java/org/terasology/utilities/procedural/SimplexNoise.java
sin3point14: oh I thought that here `i`, `j` are positive integers but on second look I can see that this is problematic, fixed this

sin3point14:
Also, I've tested it with the default Perlin and Simplex generators provided and all logic seems to be working fine now.
(Sorry I didn't get this idea earlier 😅 )

.idea/misc.xml
```diff
@@ -12,5 +12,11 @@
...
+    <option name="languageLevel" value="ES6" />
+  </component>
```
pollend, line 20:
```suggestion
```

PR #4063, opened 2020-07-02
doc: Add docs and contracts for AbstractValueModifiableEvent

skaldarnar:
This PR follows from https://github.com/Terasology/ClimateConditions/pull/17#discussion_r448634606 (thanks to @ktksan).

The lack of documentation makes it hard to spot quirky or unexpected behavior. In this particular case, I'd like to start a discussion to change the implementation.

Currently the result value is computed as 
```
result = max(0, <baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>
```
which is counter-intuitive for many reasons (and I don't know why Marcin implemented it this way).

At first I thought that capping the value to be positive holds for the result value in general, but that's not true - only the first set of _modifiers_ is guaranteed to result in a value greater or equal to zero, both _multipliers_ and _post-modifiers_ can move the result to negative numbers.

## Options

1. Ensure that the result value is _always non-negative_. This can be achieved in different ways and it needs to be clearly documented how the value is computed.
    ```
    result = max(0, <baseValue> + Σ <modifier> * Π <multiplier> + Σ <postModifier>)
    result = max(0, max(0, max(0, <baseValue> + Σ <modifier>) * Π <multiplier>) + Σ <postModifier>)
    ...
    ```
2. Simplify the computation without any restrictions to domain or range. All modifiers and multipliers can be zero or negative, and the result is just computed as
    ```
    result = <baseValue> + Σ <modifier> * Π <multiplier> + Σ <postModifier>
    ```
    Any system using the result value is responsible to cap it if necessary.

## Proposal

I'd propose to go for **Option 2 - Simplify** to make the usage more intuitive and fail-safe. However, I don't know whether there are any implications. However, if any system relies on the fact that the result is capped to be non-negative that's a bug even right now... 🙄

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/event/AbstractValueModifiableEvent.java
+++ b/engine/src/main/java/org/terasology/entitySystem/event/AbstractValueModifiableEvent.java
@@ -20,6 +20,9 @@ import gnu.trove.list.TFloatList;
 import gnu.trove.list.array.TFloatArrayList;
 
 /**
+ * A generic event for getting a value for a property.
+ *
+ * @see AbstractValueModifiableEvent
  */
 public abstract class AbstractValueModifiableEvent implements Event {
     private float baseValue;
@@ -48,10 +51,17 @@ public abstract class AbstractValueModifiableEvent implements Event {
         postModifiers.add(amount);
     }
 
+    /**
+     * Calculates the result value from the base value and given modifiers and multipliers.
+     * 
+     * The value is calculated base on the following formula:
+     * <pre>
+     * result = max(0, <baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>
+     * </pre>
+     *
+     * <emph>Note that the value range of result is not restricted!</emph>
+     */
     public float getResultValue() {
-        // For now, add all modifiers and multiply by all multipliers. Negative modifiers cap to zero, but negative
-        // multipliers remain.
-
         float result = baseValue;
         TFloatIterator modifierIter = modifiers.iterator();
         while (modifierIter.hasNext()) {
--- a/engine/src/main/java/org/terasology/entitySystem/event/AbstractValueModifiableEvent.java
+++ b/engine/src/main/java/org/terasology/entitySystem/event/AbstractValueModifiableEvent.java
@@ -64,7 +64,7 @@ public abstract class AbstractValueModifiableEvent implements Event {
      * <p>
      * The value is calculated based on the following formula:
      * <pre>
-     * result = max(0, <baseValue> + Σ <modifier> * Π <multiplier> + Σ <postModifier>)
+     * result = max(0, (<baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>)
      * </pre>
      *
      * <emph>The result value is guaranteed to be non-negative!</emph>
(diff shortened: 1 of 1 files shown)


ktksan:
@skaldarnar Maybe the AbstractValueModifiableEvent was created at the time of creating the GetMaxSpeedEvent and AffectJumpForceEvent.. and probably since the walk speed and jump force should be greater than zero at all times, this was done.
So I think, before venturing into option 2 all these systems should be checked for the same being used and alternate checks could be added within the respective systems

skaldarnar:
Pinging @jellysnake and @vedant-shroff based on the recent discussion on this topic 😉 

A generic type of `ModifiableValue` similar to this may be used in components that should be modified directly.

vedant-shroff:
At first glance, I believe that **Option 2 - Simplify** would be the correct approach to take, however echoing @ktksan's concerns, this would probably have a negative effect on any systems that currently have a reducing effect on any component that is currently modified with the `AbstractValueModifiableEvent`. **Option 1 - Ensure** seems like an improvement on the original approach in case of negative multipliers or post-modifiers. I believe that any component values could be used in a way such that they are positive values but when they are used to modify other components they could be used as negative modifiers, for example, a `PoisonComponent` would have a value of 5 but this would be a negative modifier for the `HealthComponent`. A large reduction in the `PoisonComponent` to a negative value would then cause an add Health to the user rather than reduce it and I think that is not intended, therefore must be limited at 0. In short, I vote for **Option 1 - Ensure**

skaldarnar:
@vedant-shroff @ktksan If you both vote for **Option 1 - Ensure** I take the point. I don't think that "some systems might depend on the current behavior" is a good very strong point against it, but I'm fine with having a simpler API in terms on contracts, regardless of which version 😉 
I'd go for **Option 1.1 - Single Max** in that case, basically capping the value at 0 once after the result is computed. 

@jellysnake any objections?

ktksan:
Just for the record: I do still support Option 2. 
All I was trying to say was to look out for such systems which depend on this (if any) and update them accordingly. 
Quoting @skaldarnar 
> However, if any system relies on the fact that the result is capped to be non-negative that's a bug even right now...

And going for option 1.1 is fine too I guess...

skaldarnar:
@vedant-shroff @ktksan @jellysnake please have a final look at this 🙃 I think this round is ready for merging - let's see whether this introduces any bugs because of some weird implementations somewhere 🙈

engine/src/main/java/org/terasology/entitySystem/event/AbstractValueModifiableEvent.java
```diff
@@ -44,31 +47,33 @@ public void add(float amount) {
...
-
-        final TFloatIterator postModifierIter = postModifiers.iterator();
-        while (postModifierIter.hasNext()) {
-            result += postModifierIter.next();
-        }
-        return result;
+        //TODO(skaldarnar): Based on a discussion in https://github.com/MovingBlocks/Terasology/pull/4063 we may want
+        // to lift the guarantee/restriction that the result value needs to be non-negative. Systems are still free to
+        // apply this restriction if needed.
+        return Math.max(0, baseValue + modifiers.sum() * product(multipliers) + postModifiers.sum());
```
skaldarnar, line 76: Thanks to @4Denthusiast for pointing this out in #4094

```suggestion
        return Math.max(0, (baseValue + modifiers.sum()) * product(multipliers) + postModifiers.sum());
```

engine/src/main/java/org/terasology/entitySystem/event/AbstractValueModifiableEvent.java
```diff
@@ -44,31 +47,33 @@ public void add(float amount) {
...
+    public void postAdd(float amount) {
         postModifiers.add(amount);
     }
 
+    /**
+     * Calculates the result value from the base value and given modifiers and multipliers.
+     * <p>
+     * The value is calculated based on the following formula:
+     * <pre>
+     * result = max(0, <baseValue> + Σ <modifier> * Π <multiplier> + Σ <postModifier>)
```
skaldarnar, line 67:
```suggestion
     * result = max(0, (<baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>)
```

PR #4066, opened 2020-07-04
always try loading classpath jars as modules

kaen:
### Contains

Fixes loading modules from classpath to resolve https://github.com/Terasology/ModuleTestingEnvironment/issues/12

### How to test

Detailed in the accompanying MTE PR: https://github.com/Terasology/ModuleTestingEnvironment/pull/16

I'm not sure why this limited to URLClassLoader previously. My guess is because ClassLoader has these methods `protected` where URLClassLoader has them public. I couldn't find any case where the system classloader is changed from the default. To my knowledge this can only typically be done via a system property, and that property didn't have any hits in the codebase either.

My only concern is that I don't know the implications of always loading modules from the classpath when they're provided. Still, it seems semantically correct that any time we explicitly put modules in the class path we also want those loaded by Terasology when it resolves dependencies.

 JoshariasSurvival loads happily with these changes in place, in addition to fixing the issue above.

### Comments

The only change here is to remove the typecheck and cast in favor of using the ClassLoader interface methods on lines 134 and 142. The rest of the changes are just fixing the indentation level.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ModuleManagerImpl.java
+++ b/engine/src/main/java/org/terasology/engine/module/ModuleManagerImpl.java
@@ -48,7 +48,6 @@ import java.io.Reader;
 import java.lang.reflect.ReflectPermission;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.net.URLClassLoader;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.Policy;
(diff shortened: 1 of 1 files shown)


Cervator:
Can confirm this and the MTE PRs fix the issue! Working MTE tests in Jenkins woo!

Cervator:
Got some comments from Immortius on the topic, including a possible future fate with the very latest Gestalt ES, which is a fair bit ahead of what remains inside Terasology - and a ways out on us dealing with it here:

> That has always been a tricky area
> The latest gestalt dropped the idea of auto-loading classpath modules and replaced it with package modules as that felt a bit more controllable
> Particularly because in some cases the code and resources for a module would end up in separate classpath entries
> But if that change works then go for it
> To expand on the changes a bit, rather than trying to pick out individual jars from the classpath, the new gestalt lets you specify a package or set of packages across the entire classpath
> although you can still replicate the existing approach - you can provide a filter that will determine what elements of the classpath fall into a module or not
> packages are just a bit easier and cross-platform

So since all seems well I'm going to merge this and the two MTE PRs. Thanks @kaen ! Working MTE tests in Jenkins at last, woo!

Goes with https://github.com/MovingBlocks/ModuleJteConfig/pull/6 to re-enable unit test parsing for modules (letting unstable builds correctly mark module PRs with a bad commit status)

PR #4072, opened 2020-07-07
WIP: feat(JOML): Migrate ChunkProvider

pollend:


engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -178,6 +191,22 @@ public ChunkViewCore getSubviewAroundChunk(Vector3i chunkPos) {
...
+    private ChunkViewCore createWorldView(BlockRegion region, Vector3i offset) {
+        Chunk[] chunks = new Chunk[region.getSizeX() * region.getSizeY() * region.getSizeZ()];
+        org.joml.Vector3i temp = new org.joml.Vector3i();
+        org.joml.Vector3i temp2 = new org.joml.Vector3i();
+        for (Vector3ic chunkPos : BlockRegionIterable.region(region).build()) {
+            Chunk chunk = chunkCache.get(JomlUtil.from(chunkPos));
+            if (chunk == null) {
+                return null;
+            }
+            temp.set(chunkPos).sub(region.getMinX(), region.getMinY(), region.getMinZ());
```
4Denthusiast, line 203: This line doesn't actually do anything. It just changes `temp`, which is never read.

engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -163,6 +167,15 @@ public ChunkViewCore getLocalView(Vector3i centerChunkPos) {
         return null;
     }
 
+    @Override
+    public ChunkViewCore getLocalView(Vector3ic centerChunkPos) {
+        BlockRegion region = new BlockRegion().setMin(centerChunkPos).addExtents(ChunkConstants.LOCAL_REGION_EXTENTS.x, ChunkConstants.LOCAL_REGION_EXTENTS.y, ChunkConstants.LOCAL_REGION_EXTENTS.z);
```
4Denthusiast, line 172: Shouldn't this just be an AABBi, as it isn't actually the block coordinates this refers to? It seems a bit semantically incorrect.

engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -178,6 +191,22 @@ public ChunkViewCore getSubviewAroundChunk(Vector3i chunkPos) {
         return null;
     }
 
+    private ChunkViewCore createWorldView(BlockRegion region, Vector3i offset) {
```
4Denthusiast, line 194: There's the same AABBi/BlockRegion issue here (and in the ChunkViewCoreImpl constructor it uses).

engine/src/main/java/org/terasology/world/internal/WorldProviderCoreImpl.java
```diff
@@ -183,6 +194,11 @@ public boolean isRegionRelevant(Region3i region) {
         return true;
     }
 
+    @Override
+    public boolean isRegionRelevant(BlockRegion region) {
+        return false;
```
4Denthusiast, line 199: Shouldn't this have an actual implementation?

PR #4075, opened 2020-07-08
fix: Increase item pickup range

4Denthusiast:
The code for extending item pickup range beyond the character collider was already present, but was broken accidentally in [#3907](https://github.com/MovingBlocks/Terasology/pull/3907/files/d8689f1710fb567d6ddfa3f2f9e6d4d09aedd768#diff-65b44f3586a2bf7259ca4594f2521535), so this requires very little change.

### DIFF
--- a/engine/src/main/resources/assets/prefabs/player/player.prefab
+++ b/engine/src/main/resources/assets/prefabs/player/player.prefab
@@ -20,11 +20,6 @@
   },
   "VisualCharacter": {
   },
-  // the CapsuleShape is used to detect item entities for pickup. The actual character collider is generated programmatically.
-  "CapsuleShape": {
-    "height" : 4,
-    "radius" : 1.5
-  },
   "Trigger": {
     "detectGroups": [
       "engine:debris",
--- a/engine/src/main/resources/assets/prefabs/player/player.prefab
+++ b/engine/src/main/resources/assets/prefabs/player/player.prefab
@@ -20,6 +20,11 @@
   },
   "VisualCharacter": {
   },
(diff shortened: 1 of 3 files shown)


kaen:
Confirmed this makes picking up the small rocks and branches in JS much easier :+1:

engine/src/main/resources/assets/prefabs/player/player.prefab
```diff
@@ -20,6 +20,11 @@
   },
   "VisualCharacter": {
   },
+  // the CapsuleShape is used to detect item entities for pickup. The actual character collider is generated programmatically.
```
skaldarnar, line 23: Hm, I'm wondering whether this additional shape can be derived from the collider (to keep them in sync).
If it is a "special" shape just for item pickup, maybe this should be part of some better-named component?

skaldarnar: Additional question: where is the character collider shape programmatically set? I see that in https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java#L491-L505 the character collider is created with values taken from the `CharacterMovementComponent`.

How can we keep a single source of truth for the (base) values?

4Denthusiast: I've now moved the pickup-radius information to the CharacterMovementComponent instead, with the same default value, which I think resolves this. There should be no difference in behaviour, but this makes it simpler for other things to alter the pickup radius later.

PR #4077, opened 2020-07-12
Derive player height from CharacterMovementComponent instead of BoxShapeComponent

4Denthusiast:
As the player prefab no longer actually has a BoxShapeComponent, the check for whether there's enough space to spawn the player should instead use the size defined in the CharacterMovementComponent, which is kept more up-to-date anyway if the player height changes.

This avoids the warning `[main] WARN  o.t.logic.players.PlayerFactory - entity org.terasology.entitySystem.entity.EntityBuilder@55158884 does not have any known extent specification - using default`, and the corresponding very minor risk of the player spawning with their head in a block if they have a greater than usual height.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
+++ b/engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
@@ -29,17 +29,12 @@ import org.terasology.entitySystem.entity.EntityBuilder;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.logic.characters.CharacterComponent;
+import org.terasology.logic.characters.CharacterMovementComponent;
 import org.terasology.logic.location.Location;
 import org.terasology.logic.location.LocationComponent;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/logic/players/PlayerFactory.java
```diff
@@ -99,30 +94,9 @@ public Vector3f findSpawnPositionFromLocationComponent(LocationComponent locatio
...
-        HullShapeComponent hull = prefab.getComponent(HullShapeComponent.class);
-        if (hull != null) {
-            AABB aabb = hull.sourceMesh.getAABB();
-            return aabb.maxY() - aabb.minY();
+        CharacterMovementComponent movementComponent = prefab.getComponent(CharacterMovementComponent.class);
+        if (movementComponent != null) {
+            return movementComponent.height;
         }
 
         logger.warn("entity {} does not have any known extent specification - using default", prefab);
```
skaldarnar, line 102: This can be a bit more specific now stating that the player entity does not have a `CharacterMovementComponent`.

Issue #3607, opened 2019-02-02
Opening the BT editor with F5 in multiplayer after spawning a few deer then trying to open the drop-down meant to hold BT entities crashes the game


Cervator:
To replicate: Run or join a multiplayer game with WildAnimals enabled. Press F5 and click the drop down in the top left.

To troubleshoot: Follow the stacktrace or consider running in debug mode with a breakpoint close to where the NPE happens. Figure out what the null is and harden the code so the game won't crash - just log a useful error instead :-)

Results in the following crash log on the client:

```
12:55:47.947 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
java.lang.NullPointerException: null
	at org.terasology.logic.behavior.nui.BehaviorEditorScreen$4.get(BehaviorEditorScreen.java:124)
	at org.terasology.logic.behavior.nui.BehaviorEditorScreen$4.get(BehaviorEditorScreen.java:121)
	at org.terasology.rendering.nui.widgets.UIDropdown.getOptions(UIDropdown.java:163)
	at org.terasology.rendering.nui.widgets.UIDropdown$1.onMouseClick(UIDropdown.java:64)
	at org.terasology.rendering.nui.internal.CanvasImpl.processMouseClick(CanvasImpl.java:243)
	at org.terasology.rendering.nui.internal.NUIManagerInternal.mouseButtonEvent(NUIManagerInternal.java:606)
	at org.terasology.rendering.nui.internal.NUIManagerInternalMethodAccess.invoke(Unknown Source)
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:531)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.sendConsumableEvent(EventSystemImpl.java:301)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:280)
	at org.terasology.entitySystem.entity.internal.BaseEntityRef.send(BaseEntityRef.java:204)
	at org.terasology.input.InputSystem.send(InputSystem.java:496)
	at org.terasology.input.InputSystem.sendMouseEvent(InputSystem.java:469)
	at org.terasology.input.InputSystem.processMouseButtonInput(InputSystem.java:212)
	at org.terasology.input.InputSystem.processMouseInput(InputSystem.java:191)
	at org.terasology.input.InputSystem.update(InputSystem.java:134)
[...]

PR #4088, opened 2020-07-17
fix: NPE in BehaviorEditor when on multiplayer client

IsaiahBlanks:
### Contains

Fixes issue [#3607 ](https://github.com/MovingBlocks/Terasology/issues/3607) that was causing NPE in BehaviorEditorScreen when joining multiplayer worlds. The BehaviorSystem was not registered in these situation because it was registered with RegisterMode.AUTHORITY, which only registers for the host in multiplayer.

### How to test

Join a mulitplayer world with WildAnimals enabled (not as a host). Spawn in some deer. Hit F5, and then try to open the dropdown menu in the top left of the screen. Game should now work as expected

### DIFF
--- a/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
+++ b/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
@@ -58,7 +58,7 @@ import java.util.Optional;
  * <p/>
  * Modifications made to a behavior tree will reflect to all entities using this tree.
  */
-@RegisterSystem(RegisterMode.AUTHORITY)
+@RegisterSystem(RegisterMode.ALWAYS)
 @Share(BehaviorSystem.class)
 public class BehaviorSystem extends BaseComponentSystem implements UpdateSubscriberSystem {
 


skaldarnar:
@IsaiahBlanks I'm sorry to be that "bad cop" 👮‍♂️  here - although this fix works in preventing the crash, it brings too many unknowns into the equation (essentially, code that is supposed to be run only in the authority system now also runs on each client). To quote @Cervator from Discord:

> imho the `ALWAYS` can be a cheap and dirty fix that may appear to help (and may even fix limited things) but carries high risk of causing more subtle bugs. If you're making an `ALWAYS` system the right answer might actually be to reassess how the system is designed in the first place and if it should be split into two separate systems with better event-based communication. That's a lot more work tho 🙂 
>
> an example of an accidental fix could indeed involve opening a UI via some system. Could bug out before because it only triggers on an authority, even if that's a headless server. Switching to `ALWAYS` may "fix" it because now the UI also opens on a client (as well as on a headless server), but really the better fix is to split the logic to where the authoritative stuff happens in one system solely on a server (without any UI work as that makes no sense) and there's a separate client system that deals with initiating UI actions
in this specific case there's behavior logic in `BehaviorSystem` that would start running on clients that didn't before, including an update loop. That seems potentially problematic, but might by accident just be thrown away. Not sure without digging deeper

I guess we have to look out for a different approach to fix #3607.

Issue #2307, opened 2016-05-01
NUI Extraction


rzats:
### Contains

A list of dependencies that need to be resolved in order to extract NUI ([org.terasology.rendering.nui](https://github.com/MovingBlocks/Terasology/tree/develop/engine/src/main/java/org/terasology/rendering/nui)) into a standalone library: [MovingBlocks/TeraNUI](https://github.com/MovingBlocks/TeraNUI). These can generally be resolved individually.
- [ ] CoreRegistry is utilised by [UIFormat](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/asset/UIFormat.java), [UISkinFormat](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/skin/UISkinFormat.java) and [PropertyProvider](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/rendering/nui/properties/PropertyProvider.java). However, since the extracted library shouldn't have a Context dependency, a different way of doing dependency injection will need to be introduced here (passing constructor parameters, for instance). Refer to #1818 for some relevant discussion.
- [ ] NUI depends heavily on [org.terasology.input](https://github.com/MovingBlocks/Terasology/tree/develop/engine/src/main/java/org/terasology/input), which in turn depends on [org.terasology.entitySystem](https://github.com/MovingBlocks/Terasology/tree/develop/engine/src/main/java/org/terasology/entitySystem). This package is currently being refactored into a gestalt library, but NUI is not expected to depend on it - a new way of handling entities will need to be developed instead.
- [ ] [org.terasology.input.BindableButton](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/input/BindableButton.java) uses a [SimpleUri](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/SimpleUri.java) as an identifier - since this class is essentially a wrapper around a gestalt-asset-core Uri string, it can potentially be moved to gestalt.
[...]

PR #4089, opened 2020-07-18
feat(FlowLayout): add option for vertical spacing

stefaniamak:
#### Summary

Enhancement of FlowLayout: the option to put vertical spacing in between the items included withing that FlowLayout.

#### Additions

Added a variable named `verticalSpacing`, which can be set within the UI file that used a `FlowLayout`. I used `RowLayout.java`'s implementation of `horizontalSpacing`.

#### Steps

1. Get an up-to-date version of the LightAndShadow module
2. Go to file `LightAndShadow/overrides/Inventory/ui/hud/inventoryHud/ui`
3. Find where the `"flowLayout"` is located at, and add the line `"verticalSpacing": 10,` _(or any number instead of `10`)_. Place the line somewhere between lines with `"type"` and `"contents"`, e.g. like this:
```JSON
"type": "flowLayout",
"id": "toolbar",
"verticalSpacing": 10,
"contents": [
```
4. Run your build and create/open a LaS game
5. Notice the spacing between the quickslot slots

#### Result

There should be spacing in between the slots, like the image below:

image

**Q:** Should I add the feature to add horizontal spacing as well, considering it is probably done similarly? If yes, should this PR be merged, and then make a new one for the other new feature?

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
@@ -20,6 +20,7 @@ import org.terasology.math.geom.Rect2i;
 import org.terasology.math.geom.Vector2i;
 import org.terasology.rendering.nui.Canvas;
 import org.terasology.rendering.nui.CoreLayout;
+import org.terasology.rendering.nui.LayoutConfig;
 import org.terasology.rendering.nui.LayoutHint;
 import org.terasology.rendering.nui.UIWidget;
 
@@ -31,6 +32,12 @@ import java.util.List;
 public class FlowLayout extends CoreLayout<LayoutHint> {
 
     private List<UIWidget> contents = Lists.newArrayList();
+
+    /**
+     * The vertical spacing between adjacent widgets, in pixels
+     */
+    @LayoutConfig
+    private int verticalSpacing;
 
     @Override
     public void addWidget(UIWidget element, LayoutHint hint) {
@@ -62,6 +69,7 @@ public class FlowLayout extends CoreLayout<LayoutHint> {
             canvas.drawWidget(widget, Rect2i.createFromMinAndSize(filledWidth, heightOffset, size.x, size.y));
             filledWidth += size.x;
             filledHeight = Math.max(filledHeight, size.y);
+            heightOffset += verticalSpacing;
         }
     }
 
@@ -83,7 +91,7 @@ public class FlowLayout extends CoreLayout<LayoutHint> {
             }
         }
         result.x = Math.max(result.x, filledWidth);
-        result.y += filledHeight;
+        result.y += filledHeight + verticalSpacing * (contents.size() - 1);
 
         return result;
     }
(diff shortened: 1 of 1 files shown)


stefaniamak:
FYI @skaldarnar & @jdrueckert

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
```diff
@@ -54,14 +62,15 @@ public void onDraw(Canvas canvas) {
...
-            if (filledWidth != 0 && filledWidth + size.x  > canvas.size().x) {
+            if (filledWidth != 0 && filledWidth + size.x > canvas.size().x) {
                 heightOffset += filledHeight;
                 filledWidth = 0;
                 filledHeight = 0;
             }
             canvas.drawWidget(widget, Rect2i.createFromMinAndSize(filledWidth, heightOffset, size.x, size.y));
             filledWidth += size.x;
             filledHeight = Math.max(filledHeight, size.y);
+            heightOffset += verticalSpacing;
```
skaldarnar, line 73: This adds a height offset after each element, but I think this should only happen in case there is a "wrap", i.e., inside the conditional above.

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
```diff
@@ -54,14 +62,15 @@ public void onDraw(Canvas canvas) {
         int heightOffset = 0;
         for (UIWidget widget : contents) {
             Vector2i size = canvas.calculatePreferredSize(widget);
-            if (filledWidth != 0 && filledWidth + size.x  > canvas.size().x) {
+            if (filledWidth != 0 && filledWidth + size.x > canvas.size().x) {
                 heightOffset += filledHeight;
```
skaldarnar, line 66: I think the vertical spacing should be added here, instead:
```suggestion
                heightOffset += filledHeight;
                heightOffset += verticalSpacing;
```

This part of the code resets the filled width and height after increasing the height offset, e.g., after wrapping the line.

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
stefaniamak: oh indeed, I'll look into adding it only when there is a wrap, and then make a commit for it!

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
stefaniamak: I see I see, that seems like it, yes. Placed it in that position at the very start (because the other `filledHeight` was there, not because I had the wrapping in mind), but among my other changes till I reach the correct solution I changed it. I'll commit the suggestion 👍

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
```diff
@@ -83,7 +92,7 @@ public Vector2i getPreferredContentSize(Canvas canvas, Vector2i sizeHint) {
             }
         }
         result.x = Math.max(result.x, filledWidth);
-        result.y += filledHeight;
+        result.y += filledHeight + verticalSpacing * (contents.size() - 1);
```
skaldarnar, line 95: Good point - that's something I missed before 👍 

I think you can only multiply with the number of elements in the content if you can make sure that there is only one element per row. However, if you make the flow are bigger, you should see that there might be multiple widgets in the same row before the automatic wrapping kicks in.

stefaniamak: Indeed, this is a similar note with the one for adding the spacing in the first place, should have thought about that. I'll look into changing `contents.size() - 1` with a variable that counts the rows of the FlowLayout!

skaldarnar: I think it might be enough to just increment `result.y` by `verticalSpacing` whenever the spacing is applied to the offset.

skaldarnar:
❗ This change affects #4017 (#2307)

PR #4090, opened 2020-07-20
feat(FlowLayout): Add option for horizontal spacing

stefaniamak:
#### Summary

Enhancement of FlowLayout: the option to put vertical spacing in between the items included withing that FlowLayout. 

#### Additions

Added a variable named `horizontalSpacing`, which can be set within the UI file that used a `FlowLayout`. Using #4089 's `verticalSpacing` as reference.

#### Steps

1. Search for `inventoryHud.ui` file within the `Inventory` module
2. Place right below the `"id": "toolbar",` the line `"horizontalSpacing": 30,`
3. Find the line `"use-content-width": true,` at the `layoutInfo"` of the `"FlowLayout"` element (line 56) and replace it with `"width": 150,`
4. Run a game of any module that uses the quickslot module (if you use a module that overrides the above file, then make the above changes there. You can check if a file gets overridden if there is a file with this path: `modules/YourModule/overrides/inventory/ui/hud/inventoryHud.ui)
5. Notice the quickslot

##### Issues

- [X] Add spacing
- [ ] ~~Remove spacing from the last element of each row~~ - not an error
- [x] Bring the hidden part of the lowest row back inside the screen
- [x] Quickslot is not stuck at its position formating

image


stefaniamak:
FYI @skaldarnar, I had a hard time pinning down how to correct the written issues, but I will be looking into them again tomorrow.

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
```diff
@@ -51,16 +58,17 @@ public void removeAllWidgets() {
     public void onDraw(Canvas canvas) {
         int filledWidth = 0;
         int filledHeight = 0;
+        int widthOffset = 0;
         int heightOffset = 0;
         for (UIWidget widget : contents) {
             Vector2i size = canvas.calculatePreferredSize(widget);
-            if (filledWidth != 0 && filledWidth + size.x  > canvas.size().x) {
+            if (filledWidth != 0 && filledWidth + size.x > canvas.size().x) {
```
skaldarnar, line 65: I think this check also needs to take the spacing into account. If we add both the spacing and the size of the next element to the currently filled width, do we exceed the canvas size?

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
```diff
@@ -72,13 +80,13 @@ public Vector2i getPreferredContentSize(Canvas canvas, Vector2i sizeHint) {
         int filledHeight = 0;
         for (UIWidget widget : contents) {
             Vector2i size = canvas.calculatePreferredSize(widget);
-            if (filledWidth != 0 && filledWidth + size.x  > sizeHint.x) {
+            if (filledWidth != 0 && filledWidth + size.x > sizeHint.x) {
```
skaldarnar, line 83: Same here.

stefaniamak:
> 
> 
> I don't think we need to do something about the "last element" here 🤔 we're not adding any exceeding spaces, and the widget flowing out of the screen is probably just because we're overflowing the FlowLayout itself - and that's nothing introduced by this PR.

_About the last element:_ Not the last element of the FlowLayout, but the last element of each line (I'll make it more clear at the description too). If we let that be added, then we have an extra spacing at the end of each line, instead of only between the elements. Then having both spacing and `position-right` `offset` together would be a real struggle.

_About overflowing:_ it does sound like a feasible reason, but I don't think it's what's happening here. I guess so because when removing the `"horizontalSpacing":` from the `UI` file, the last row is correctly within the screen.

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
```diff
@@ -72,10 +80,10 @@ public Vector2i getPreferredContentSize(Canvas canvas, Vector2i sizeHint) {
...
                 result.x = Math.max(result.x, filledWidth);
                 result.y += filledHeight;
-                filledWidth = size.x;
+                filledWidth = size.x + horizontalSpacing * (contents.size() - 1);
```
stefaniamak, line 86: This worked at first with 2 test, but the rest tests proven for this not being the correct solution.

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
stefaniamak: Considering we add the spacing right afterwards, yes that probably should include the spacing as well. Generally though, wouldn't we want for the item to be added if it fits to the row, no matter the spacing we set in between the other items? 

So what my understanding of the code is that it adds the item and the spacing (to its right) in one loop, and then go to the next item. Is that how it works? Because I still feel like there is a detail within the two methods that I still haven't understood yet.

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
```diff
@@ -121,6 +138,17 @@ public int getVerticalSpacing() {
      * @param spacing The spacing, in pixels
      * @return This {@code FlowLayout}
      */
+    public FlowLayout setHorizontalSpacing(int spacing) {
+        this.horizontalSpacing = spacing;
+        return this;
+    }
+
+    /**
+     * Sets the horizontal spacing between adjacent widgets in this {@code FlowLayout}.
```
skaldarnar, line 147: I think the docstrings are now flipped for `setHorizontalSpacing` and `setVerticalSpacing` 😅

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
skaldarnar: I think the solution for this is similar to the vertical case - just add the `horizontalSpacing` each time when adding `size.x` somewhere.

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
skaldarnar: Yeah, you're probably right - that's also where "remove from last element" comes from 🤔  so this check is indeed correct, because the "previous" gap is already contained in `filledWidth`... 👍

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
stefaniamak: And I did double check... On it 👍

engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
```diff
@@ -107,6 +115,15 @@ public Vector2i getMaxContentSize(Canvas canvas) {
     }
 
     /**
+     * Retrieves the horizontal spacing between adjacent widgets in this {@code FlowLayout}.
+     *
+     * @return The spacing, in pixels
+     */
+    public int getHorizontalSpacingSpacing() {
```
stefaniamak, line 122: Also this changed to double `SpacingSpacing` for some reason 😆

skaldarnar:
This PR also needs to be ported to [MovingBlocks/TeraNUI](https://github.com/MovingBlocks/TeraNUI) when finally extracting the content.

PR #4093, opened 2020-07-21
fixed the load latest configuration

### LINKED ISSUE #4092: Null pointer exception when trying to use load latest.
Use the load latest configuration. 
### What actually happened
Null pointer exception in the Terasology java class
### How to reproduce

* Step 1
Try to use the load latest configuration.
* Step 2
Crash...

I think 3.2.0
### Computer details
i7 7700k
GTX 2080ti
16GB Ram
NVMe 128gb SSD

Elyahu41:
### Contains

This PR fixes #4092

### How to test

For now, run the load latest configuration.

My code might not be named or commented in the best way. It could also be benficial to create tests for this specific load configuration.


Elyahu41:
I refactored the code a bit. Let me know if I need to change anything.

Cervator:
Thanks! Can confirm this works, and the fix seems safe enough. Thank you :-)

Also went ahead and added both you and @IsaiahBlanks to the game credits for your recent work, much appreciated!

PR #4094, opened 2020-07-22
feat: Add ModifiableValue type

vedant-shroff:
### Contains

Added a ModifiableValue type to the engine which can be used by Components so that Systems changing the values of these components do not affect the real values but rather add PreModifiers, multipliers and PostModifiers which can then form a mathematically correct altered value. This PR is a result of a discussion with @skaldarnar @jellysnake @ktksan and @e-aakash after https://github.com/MovingBlocks/Terasology/pull/4063#issue-443423710 .
Components must now mention whether their data is to be modified in-place / if the data is modified via an AbstractModifiableEvent / if the data members are of ModifiableValue type, in their respective Javadocs
This PR also includes a TypeHandler to handle proper deserialization and serialization of ModifiableValue from prefabs and to persist the data across saves.

### How to use

Add the ModifiableValue type as the datatype of the data members of a component so that the value of the component can be modified correctly with different modifying systems acting on the component, without changing the base value of the component. An initial value can be given via the constructor and the .getValue() method can be used to get the value of the ModifiableValue type variable at any point 

- [x] Javadoc
- [x] Testing of the type with existing components
- [x] Further discussion on immutable values/return types

### DIFF
--- a/engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
+++ b/engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
@@ -1,0 +1,29 @@
+// Copyright 2020 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.persistence.typeHandling.extensionTypes;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.terasology.persistence.typeHandling.PersistedData;
+import org.terasology.persistence.typeHandling.PersistedDataSerializer;
+import org.terasology.persistence.typeHandling.RegisterTypeHandler;
+import org.terasology.utilities.modifiable.ModifiableValue;
+
+import java.util.Optional;
+
+@RegisterTypeHandler
+public class ModifiableValueTypeHandler extends org.terasology.persistence.typeHandling.TypeHandler<ModifiableValue> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ModifiableValueTypeHandler.class);
+
+    @Override
+    public PersistedData serializeNonNull(ModifiableValue value, PersistedDataSerializer serializer) {
+        return serializer.serialize(value.getValue());
+    }
+
+    @Override
+    public Optional<ModifiableValue> deserialize(PersistedData data) {
+        return Optional.of(new ModifiableValue(data.getAsArray().getAsFloatArray().get(0)));
+    }
+}
--- a/engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
+++ b/engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
@@ -14,6 +14,14 @@ import org.terasology.utilities.modifiable.ModifiableValue;
 
 import java.util.Optional;
 
+/**
+ * The ModifiableValue type is deserialized from prefabs directly from the corresponding component values.
+ * <b>Only</b> the base value is to be specified in the prefabs, this will deserialize into a ModifiableValue type,
+ * with set preModifier=0, multiplier=1, postModifier=0.
+ * "<field_name>": <base_value>
+ * It is not recommended to set the modifiers in the prefab but it can be done via an unlabeled array.
+ * "<field_name>": [<base_value>, <pre_modifier>, <multiplier>, <post_modifier>]
+ */
 @RegisterTypeHandler
 public class ModifiableValueTypeHandler extends org.terasology.persistence.typeHandling.TypeHandler<ModifiableValue> {
 
@@ -21,8 +29,8 @@ public class ModifiableValueTypeHandler extends org.terasology.persistence.typeH
 
     @Override
     public PersistedData serializeNonNull(ModifiableValue value, PersistedDataSerializer serializer) {
-        return serializer.serialize(value.getBaseValue(), value.getPreModifiers(), value.getMultipliers(),
-                value.getPostModifiers());
+        return serializer.serialize(value.getBaseValue(), value.getPreModifier(), value.getMultiplier(),
+                value.getPostModifier());
     }
 
     @Override
@@ -32,10 +40,10 @@ public class ModifiableValueTypeHandler extends org.terasology.persistence.typeH
             if (vals.isNumberArray()) {
                 TFloatList floatList = vals.getAsFloatArray();
                 ModifiableValue modifiableValue = new ModifiableValue(floatList.get(0));
-                if (floatList.size() > 1) {
-                    modifiableValue.setPreModifiers(floatList.get(1));
-                    modifiableValue.setMultipliers(floatList.get(2));
-                    modifiableValue.setPostModifiers(floatList.get(3));
+                if (floatList.size() == 4) {
+                    modifiableValue.setPreModifier(floatList.get(1));
+                    modifiableValue.setMultiplier(floatList.get(2));
+                    modifiableValue.setPostModifier(floatList.get(3));
                 }
                 return Optional.of(modifiableValue);
             }
--- a/engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
+++ b/engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
@@ -27,25 +27,19 @@ public class ModifiableValueTypeHandler extends org.terasology.persistence.typeH
 
     @Override
     public Optional<ModifiableValue> deserialize(PersistedData data) {
-        if (!data.isArray()) {
-            return Optional.empty();
+        if (data.isArray()) {
+            PersistedDataArray vals = data.getAsArray();
+            if (vals.isNumberArray()) {
+                TFloatList floatList = vals.getAsFloatArray();
+                ModifiableValue modifiableValue = new ModifiableValue(floatList.get(0));
+                if (floatList.size() > 1) {
+                    modifiableValue.setPreModifiers(floatList.get(1));
+                    modifiableValue.setMultipliers(floatList.get(2));
+                    modifiableValue.setPostModifiers(floatList.get(3));
+                }
+                return Optional.of(modifiableValue);
+            }
         }
-
-        PersistedDataArray vals = data.getAsArray();
-
-        if (!vals.isNumberArray() || (vals.size() != 3 && vals.size()!=1)) {
-            return Optional.empty();
-        }
-
-        TFloatList floatList = vals.getAsFloatArray();
-        ModifiableValue modifiableValue = new ModifiableValue(floatList.get(0));
-
-        if (vals.size() != 1) {
-            modifiableValue.setPreModifiers(floatList.get(1));
-            modifiableValue.setMultipliers(floatList.get(2));
-            modifiableValue.setPostModifiers(floatList.get(3));
-        }
-
-        return Optional.of(modifiableValue);
+        return Optional.empty();
     }
 }
--- a/engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
+++ b/engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
@@ -3,9 +3,11 @@
 
 package org.terasology.persistence.typeHandling.extensionTypes;
 
+import gnu.trove.list.TFloatList;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.persistence.typeHandling.PersistedData;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
```diff
@@ -0,0 +1,99 @@
...
+     * <p>
+     * The value is calculated based on the following formula:
+     * <pre>
+     * result = max(0, <baseValue> + Σ <modifier> * Π <multiplier> + Σ <postModifier>)
+     * </pre>
+     *
+     * <emph>The result value is guaranteed to be non-negative!</emph>
+     */
+    public float getValue() {
+        return Math.max(0, baseValue + modifiers.sum() * product(multipliers) + postModifiers.sum());
```
skaldarnar, line 55: I'd think that we can remove the `Math.max(...)` for this class - a value on it's own can very well be negative! If the event wants to provide this guarantee it can wrap a `ModifiableValue` and perform that operation.

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public TFloatList getModifiers() {
+        return modifiers;
+    }
+
+    public void setModifiers(TFloatList modifiers) {
```
skaldarnar, line 62: I'm not sure we should offer this setter to change _all_ modifiers at once, basically overriding everything someone else did 🤔  Depending on whether we want this to be mutable or immutable we may offer a `clear()` method to reset all modifier lists, though.

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
```diff
@@ -0,0 +1,99 @@
...
+     * result = max(0, <baseValue> + Σ <modifier> * Π <multiplier> + Σ <postModifier>)
+     * </pre>
+     *
+     * <emph>The result value is guaranteed to be non-negative!</emph>
+     */
+    public float getValue() {
+        return Math.max(0, baseValue + modifiers.sum() * product(multipliers) + postModifiers.sum());
+    }
+
+    public TFloatList getModifiers() {
```
skaldarnar, line 58: I'm also not sure whether we should expose this detail to the outside - what use cases are there where it would help a consumer to know whether a modifier of `+4` is because of `[1, 3]`, `[4]`, or `[-2, 3, 0, 3]` (or any other combination...)

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Calculate the product of all values in the given list.
+     * <p>
+     * A static helper dual to {@code TFloatList#sum()}.
+     *
+     * @param multipliers the list of multipliers to calculate the product of.
+     * @return the product of all values. 1 if the list is empty.
+     */
+    private static float product(TFloatList multipliers) {
```
skaldarnar, line 90: TODO: move this to `TFloatList` ^^

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
```diff
@@ -0,0 +1,99 @@
...
+import gnu.trove.list.array.TFloatArrayList;
+
+/**
+ * A helper type to get and modify the value of a component without changing its actual value
+ * <p>
+ * The result value is guaranteed to be greater or equal to zero.
+ * Components using this type must mention so in their javadoc.
+ * </p>
+ */
+public class ModifiableValue {
```
skaldarnar, line 17: TODO: decide whether `ModifiableValue` should be mutable or immutable.

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
4Denthusiast, line 55: Shouldn't the multipliers also multiply the baseValue (i.e. `(baseValue + modifiers.sum()) * product(multipliers) ...`)?

4Denthusiast:
I think that in order for the lists of modifiers (presumably one for each system that applies a modification) to actually be useful, each system should be able to tell which of the modifiers is its own, so it can replace or remove it. Something like `Map<Name, Float>` instead of just `TFloatList`.

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
skaldarnar: obviously yes 🙈

vedant-shroff: Yep silly mistake, have fixed.

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
vedant-shroff: Method was removed and the modifiers were changed in place as they were added.

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
vedant-shroff: Gone with the mutable approach in my latest commit #8876c27

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
```diff
@@ -0,0 +1,56 @@
...
+        postModifiers += amount;
+        return this;
+    }
+
+    /**
+     * Calculates the result value from the base value and given modifiers and multipliers.
+     * <p>
+     * The value is calculated based on the following formula:
+     * <pre>
+     * result = max(0, <baseValue> + Σ <modifier> * Π <multiplier> + Σ <postModifier>)
```
4Denthusiast, line 51: This formula should be updated to match the actual value: `result = (<baseValue> + Σ <modifier>) * Σ <multiplier> + Σ <postModifier>`

vedant-shroff: Have changed the multiplers to compound now as per the formula

4Denthusiast: The rest of the formula in the comment still needs changing though: `result = (<baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>`

engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
```diff
@@ -0,0 +1,45 @@
...
+import org.terasology.persistence.typeHandling.PersistedData;
+import org.terasology.persistence.typeHandling.PersistedDataArray;
+import org.terasology.persistence.typeHandling.PersistedDataSerializer;
+import org.terasology.persistence.typeHandling.RegisterTypeHandler;
+import org.terasology.utilities.modifiable.ModifiableValue;
+
+import java.util.Optional;
+
+@RegisterTypeHandler
+public class ModifiableValueTypeHandler extends org.terasology.persistence.typeHandling.TypeHandler<ModifiableValue> {
```
skaldarnar, line 18: I'd like to have some JavaDoc here (or some documentation somewhere else) on how the `ModifiableValue` is actually persisted. What is the format to specify a modifiable in a prefab?

From looking at the code it seems like the value is persisted as "unlabeled" array:

```
"<field_name>": [<base_value>, <pre_modifier>, <multiplier>, <post_modifier>]
```

I don't think this is very intuitive, and when seeing this in a prefab I'd think of a vector or list but not of a special data class 🙈

engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
```diff
@@ -0,0 +1,45 @@
...
+                TFloatList floatList = vals.getAsFloatArray();
+                ModifiableValue modifiableValue = new ModifiableValue(floatList.get(0));
+                if (floatList.size() > 1) {
+                    modifiableValue.setPreModifiers(floatList.get(1));
+                    modifiableValue.setMultipliers(floatList.get(2));
+                    modifiableValue.setPostModifiers(floatList.get(3));
+                }
+                return Optional.of(modifiableValue);
+            }
+        }
```
skaldarnar, line 42: I'm wondering whether it would make sense to check an `else if` here for a single number. That would allow to easily specify the base value in a prefab without worrying about all the modifiers. So you could either do 

```jsonc
"myValue": 7,					// base value only
// ... or ...
"otherValaue": [3, 0, 0.9, 2]	// base and modifiers/multiplier
```

engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public Optional<ModifiableValue> deserialize(PersistedData data) {
+        if (data.isArray()) {
+            PersistedDataArray vals = data.getAsArray();
+            if (vals.isNumberArray()) {
+                TFloatList floatList = vals.getAsFloatArray();
+                ModifiableValue modifiableValue = new ModifiableValue(floatList.get(0));
+                if (floatList.size() > 1) {
```
skaldarnar, line 35: Shouldn't this check that the size is actually exactly 4? Otherwise, `floatList.get(n)` would fail...

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
```diff
@@ -0,0 +1,85 @@
...
+ * Components using this type must mention so in their javadoc so all modifiers are added correctly.
+ * </p>
+ */
+@API
+public class ModifiableValue {
+    private final float baseValue;
+
+    private float preModifiers;
+    private float multipliers;
+    private float postModifiers;
```
skaldarnar, line 20: As these are simple float values they should not be named in plural form (that makes more sense for lists or arrays):
```suggestion
    private float preModifier;
    private float multiplier;
    private float postModifier;
```

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
```diff
@@ -0,0 +1,85 @@
...
+        return preModifiers;
+    }
+
+    public float getPostModifiers() {
+        return postModifiers;
+    }
+
+    public float getMultipliers() {
+        return multipliers;
+    }
```
skaldarnar, line 72:
```suggestion
    public float getPreModifier() {
        return preModifier;
    }

    public float getPostModifier() {
        return postModifier;
    }

    public float getMultiplier() {
        return multiplier;
    }
```

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
```diff
@@ -0,0 +1,85 @@
...
+        this.preModifiers=preModifiers;
+    }
+
+    public void setMultipliers(float multipliers){
+        this.multipliers=multipliers;
+    }
+
+    public void setPostModifiers(float postModifiers){
+        this.postModifiers=postModifiers;
+    }
```
skaldarnar, line 84: I'd say these methods should have JavaDoc to encourage usage of the `preAdd`, `postAdd`, and `multiply` instead of overriding the respective value.

engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
vedant-shroff: This is handled by default as the constructor takes the base value and assigns default modifiers. The single value specified in the prefabs is deserialized into the constructor for ModifiableValue which takes only the base value. Other modifiers can be set optionally

engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java
vedant-shroff: Done :+1:

engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
vedant-shroff: Since the size can only be 1(when deserializing the prefabs) or 4(on a game load after a seed has been harvested), this would not cause any problems. However, this seems like the right thing to do in general. Have made the change.

engine/src/main/java/org/terasology/persistence/typeHandling/extensionTypes/ModifiableValueTypeHandler.java
vedant-shroff: Added Javadocs, the default way to specify in prefabs is with a single value only and then the modifiers are given default values.

PR #4095, opened 2020-07-22
feat: Add BeforeAfterEvent

ktksan:
Adds AbstractBeforeAfterEvent : Immutable event to notify a change in a certain value.
This is a generic event and more specific events implementing this event should be fired and reacted to.

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/event/BeforeAfterEvent.java
+++ b/engine/src/main/java/org/terasology/entitySystem/event/BeforeAfterEvent.java
@@ -1,0 +1,35 @@
+// Copyright 2020 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.entitySystem.event;
+
+/**
+ * Immutable event to notify a change in a certain value.
+ * NOTE: This is a generic event and a more specific Event extending this event should be fired and reacted to.
+ */
+public abstract class BeforeAfterEvent<T> implements Event {
+    protected final T oldValue;
+    protected final T newValue;
+
+    /**
+     * Creates a new notification event on change in value.
+     */
+    BeforeAfterEvent(final T oldValue, final T newValue) {
+        this.oldValue = oldValue;
+        this.newValue = newValue;
+    }
+
+    /**
+     * Returns the value prior to change.
+     */
+    public T getOldValue() {
+        return oldValue;
+    }
+
+    /**
+     * Returns the value after change.
+     */
+    public T getNewValue() {
+        return newValue;
+    }
+}


ktksan:
I was thinking about more methods which could be of help like getChange() - returns the change in value. Please comment if you think it's a good idea to add more and if any specific one comes to mind.. 
I thought about getFactor() {return newValue/ oldValue} but that would create a problem in case old value is zero ,so I dropped it.

engine/src/main/java/org/terasology/entitySystem/event/AbstractBeforeAfterEvent.java
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.entitySystem.event;
+
+/**
+ * Immutable event to notify a change in a certain value.
+ * NOTE: This is a generic event and a more specific Event extending this event should be fired and reacted to.
+ */
+public class AbstractBeforeAfterEvent implements Event {
```
skaldarnar, line 10: Did you ever work with [Generics](https://www.baeldung.com/java-generics), as in, implemented a generic class? I'd think this would be a good case where we can use that - don't fix the event to be based on float right away.

So here's my suggestion:
```suggestion
public abstract class BeforeAfterEvent<T> implements Event {
```

- make the class abstract to ensure nobody can create a `new BeforeAfterEvent("foo", "bar")` out of thin air, but has to extend it to use it.
- shorten the name to `BeforeAfterEvent` - I don't see any good value in explicitly prefixing the "Abstract" part, and the inheritance line will read a little bit cleaner
  ```java
  public class HealthChangedEvent extends BeforeAfterEvent<Float> { ... }
  ```
- make the type of the changed value generic `<T>`. this allows to reuse the same base class regardless whether the value is an integer, a float, a string, or anything else. However, this implies that we cannot offer the `getChange` on this base class.

engine/src/main/java/org/terasology/entitySystem/event/AbstractBeforeAfterEvent.java
```diff
@@ -0,0 +1,42 @@
...
+    private final float oldValue;
+    private final float newValue;
```
skaldarnar, line 12: If the class works with a generic `<T>` these values become of type `T`.
```suggestion
    protected final T oldValue;
    protected final T newValue;
```

If subclasses should be able to access `oldValue` and `newValue` those should be protected instead of private.

engine/src/main/java/org/terasology/entitySystem/event/AbstractBeforeAfterEvent.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Creates a new notification event on change in value.
+     */
+    AbstractBeforeAfterEvent(float oldValue, float newValue) {
```
skaldarnar, line 17: Same here:
```suggestion
    AbstractBeforeAfterEvent(final T oldValue, final T newValue) {
```

We can also mark them final as we are not going to change them ever in this event.

engine/src/main/java/org/terasology/entitySystem/event/AbstractBeforeAfterEvent.java
```diff
@@ -0,0 +1,42 @@
...
+     * Returns the value after change.
+     */
+    public float getNewValue() {
+        return newValue;
+    }
+
+    /**
+     * Returns the amount by which the value has been changed.
+     */
+    public float getChange() {
```
skaldarnar, line 39: If we want this convenience functionality you could introduce a 

```java
public abstract class FloatBeforeAfterEvent extends BeforeAfterEvent<Float> { ... }
```

for this purpose. 

For calculating the factor you could consider to return infinity in case the old value was zero, something like this (there's a difference between positive and negative infinity, though):

```java
    public float getFactor() {
        return oldValue != 0 ? newValue / oldValue : Float.POSITIVE_INFINITY;
    }
```

engine/src/main/java/org/terasology/entitySystem/event/AbstractBeforeAfterEvent.java
ktksan: Nope I certainly haven't worked with Generics 😅... And I did absolutely everything wrong I see 😅
And Now the component based alteration effect pr thing  (Class<C> componentClass) makes sense..
Thanks for explaining all this in such great detail!

engine/src/main/java/org/terasology/entitySystem/event/AbstractBeforeAfterEvent.java
ktksan: I see, Again thanks for explaining!
But for this PR I don't think we should introduce FloatBeforeAfterEvent just for a getChange/Factor() method... 
The whole point of using generics I think is we don't need to have separate events for different datatypes.. 
So I think we can leave those separate functionalities to the specific events that extend this event.

engine/src/main/java/org/terasology/entitySystem/event/AbstractBeforeAfterEvent.java
ktksan: Ah yes. I should've at least gotten that right 😔

engine/src/main/java/org/terasology/entitySystem/event/AbstractBeforeAfterEvent.java
skaldarnar: > And I did absolutely everything wrong

That's not true - basically everything was there, and it made it easy for me to add all those comments about fancy features :sweat_smile: I hope you'll learn one thing or the other, and please don't hesitate if you have questions (and please ask about things like `Class componentClass` if they are unclear :wink: )

ktksan:
These are some of the potential candidates which could use BeforeAfterEvent:
CameraTargetChangedEvent
Health - ChangeHealth/MaxHealth/Damaged/Restored
InventorySlotChangedEvent
InventorySlotStackSizeChangedEvent
LocationChangedEvent
OnBiomeChangedEvent
PlayerTargetChangedEvent
new ClimateCondition PRs

NOTE:
1.This list includes events I could find after a bit of searching from only the modules which I have in my workspace...
2.Some of these events can be straight up replaced while some (very few) of them have a lot of other attributes along with the newValue and oldValue and the old/newValue s are used multiple times in different methods inside the event... so in those cases we may choose to leave them as they are at the moment...

PR #4097, opened 2020-07-24
feat(AGS): by default show only locally present modules

jdrueckert:
- this includes modules present as source _and_ as .jar
- https://github.com/MovingBlocks/Terasology/pull/3973 set the default to augmentation which is a perfect match for the player-workflow, but as long as the main workflow is dev/test and not player, local makes more sense


jdrueckert:
pinging @agent-q1 for review as they were the latest one to fall into the augmentation filter trap :sweat_smile:

PR #4098, opened 2020-07-25
fix: Actually register InputSystem as system

skaldarnar:
There were a couple of NPEs with Dialogs because of this. The dependencies for the InputSystem were not injected via `@In` because the system was never registered as such. This simple change should solve this issue.

### DIFF
--- a/engine/src/main/java/org/terasology/input/InputSystem.java
+++ b/engine/src/main/java/org/terasology/input/InputSystem.java
@@ -25,6 +25,7 @@ import org.terasology.engine.subsystem.config.BindsManager;
 import org.terasology.engine.subsystem.config.BindsSubsystem;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.systems.BaseComponentSystem;
+import org.terasology.entitySystem.systems.RegisterSystem;
 import org.terasology.input.cameraTarget.CameraTargetSystem;
 import org.terasology.input.device.ControllerAction;
 import org.terasology.input.device.KeyboardAction;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/input/InputSystem.java
```diff
@@ -64,6 +65,7 @@
  * In addition to raw keyboard and mouse input, the system handles Bind Buttons and Bind Axis, which can be mapped
  * to one or more inputs.
  */
+@RegisterSystem
```
Cervator, line 68: I wonder if this should be set to Client? It'll otherwise default to Always which means it would also run on a headless server, which seems silly for input handling.

Which makes me wonder what it was doing _without_ the annotation in the first place 🤔 I'm tempted to put a couple fingers in my ears and walk away slowly going "lalalalala" pretending to never having seen this ... 😁

skaldarnar: I got the impression that we should double-check all `@RegisterSystem` usages anyways 🙈 Having it `ALWAYS` register wouldn't hurt that much, would it?

skaldarnar:
Merging to unblock @stefaniamak working on Dialog changes for LaS.

engine/src/main/java/org/terasology/input/InputSystem.java
Cervator: Probably not, but the class _does_ have an `update` that'll run every tick, pointlessly, since servers don't process input. Likely it is minor and probably even has been an issue for ages. So not likely much to worry about - we should hunt whichever performance issues actually manifest themselves rather than theoretical ones :-)

(and yeah we should totally check all the systems - if no sooner then maybe when we re-integrate Gestalt, as any detailed effort until then might end up wasted if everything changes too much anyway ...)

PR #4064, opened 2020-07-03
fix(GameProvider): prevent creating game with duplicate name

DarkWeird:
### Contains
Fixes #3982 

Custom game name works like Default game name prefix ("Game") now.
If you create game with name as previously created new game will have name "<previous> <index>"
If previously created game name already have index, then that index will be incremented.

### How to test

1. create game with name "X"
2. exit to main menu
3. create game with name "X" again
4. check save games. You have save games "X" and "X 1" now.

PR #4105, opened 2020-07-29
Feat/create latest

### LINKED ISSUE #4061: Create a game world directly, skipping the UI
Create a new game from a gameplay template every time it is run..
There is already a "Load Latest" config that loads the least played game but this isn't very helpful in case of WorldGen related work as a previously loaded world won't be showing any changes that may have been done. A freshly created world is needed to observe the changes.
Currently one has to create a new game using the gameplay template and often it is required to required `hspeed` and `flight` commands to enable easy observation.
So, automating this would result in faster dev cycles

Relevant conversation on Discord- https://discordapp.com/channels/270264625419911192/702893102481932308/725449783291936869

Admin edit: See the parameter `-loadlastgame` passed to the game by the "TerasologyPC (load latest)" run configuration in IntelliJ. That argument is parsed in the `Terasology` class in the PC facade and passed onward to the engine. A similar approach would likely work for `recreateLastGame` or so - just take the default configuration (game must have been run once with a world created) and pass it through, bypassing/automating the game UI just like load latest

Elyahu41:
### Contains

Fixes #4061 

### How to test

Run the "create latest" configuration

Should probably add comments/javadoc for these configurations. (load lastest and create latest)


.idea/runConfigurations/TerasologyPC__create_latest_.xml
```diff
@@ -0,0 +1,12 @@
...
+    <option name="MAIN_CLASS_NAME" value="org.terasology.engine.Terasology" />
+    <module name="Terasology.facades.PC.main" />
+    <option name="PROGRAM_PARAMETERS" value="-homedir -noCrashReport -createlastgame" />
+    <shortenClasspath name="MANIFEST" />
+    <option name="VM_PARAMETERS" value="-Xms256m -Xmx1536m" />
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
```
Elyahu41, line 12: This is the run configuration.

Elyahu41:
I think you should first review my other PR which is #4093 , because this code will only run if that PR is merged first.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -150,18 +153,30 @@ public static void main(String[] args) {
...
+                    engine.initialize();
+                    engine.getFromEngineContext(ThreadManager.class).submitTask("createLastGame", () -> {
                         GameManifest gameManifest = getLatestGameManifest();
                         if (gameManifest != null) {
-                            engine.changeState(new StateLoading(gameManifest, NetworkMode.NONE));
+                            GameManifest newlyCreatedGM = gameManifest;
+                            Random random = new Random();
+                            String randomNumber = String.valueOf(random.nextInt());
+                            newlyCreatedGM.setTitle("New Created " + gameManifest.getSeed() + "(" + randomNumber + ")");
+                            engine.changeState(new StateLoading(newlyCreatedGM, NetworkMode.NONE));
```
Elyahu41, line 173: Setting the title to something else will rebuild the world with the same seed.

Cervator: This overall looks reasonable, but I wonder if we could name the game more cleanly - if you've got the previous game manifest does that by chance include that save's _name_ and if so could we just do "Recreated '[old game name]'" or so? Maybe with an incrementing number rather than a random one?

Just recently https://github.com/MovingBlocks/Terasology/pull/4064 was merged which was trying to harden the choice of which increment would get added at the end of a game name, including _clever_ cases where a user tries to trick the system (by for instance adding an increment manually that would clash with something). Mind taking a look at that and seeing if we have any ideal options? :)

Elyahu41:
@Cervator I'll probably create a new PR with a fresh branch. As for renaming the newly created file, I was having truble with an unending loop. But, I'll look into it and see if I can do it better.

Cervator:
@Elyahu41 - thanks! Yeah new PR with a fresh branch is fine. Take your time 👍

PR #4107, opened 2020-07-29
refactor(FlowLayout): Use same code path for drawing and calculating size

skaldarnar:
With this PR the Flow layout uses the same code path to compute the preferred size and to actually draw the widgets on screen. @stefaniamak faced a couple of quirks where the computed size was different from the layout chosen by `onDraw` when adding horizontal and vertical spacing (#4089 and #4090).

In addition, this PR fixes a few Checkstyle rules and adds more JavaDoc, partially taken from the documentation at https://terasology.org/TeraNUI/layouts/flow-layout.html.

I've only tested this PR with Light and Shadow locally. Should do another round of testing with other game modes to see if everything still works as expected.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
@@ -35,18 +35,6 @@ public class FlowLayout extends CoreLayout<LayoutHint> {
     private List<UIWidget> contents = Lists.newArrayList();
 
     /**
-     * Whether the directional flow of this layout goes from left-to-right and right-to-left.
-     * <p>
-     * The children are laid out from left-to-right by default (false), aligned at the left border of the canvas. If
-     * this toggle is explicitly enabled (true) the children are laid out right-to-left, aligned at the right border of
-     * the canvas.
-     * <p>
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
```diff
@@ -63,41 +87,64 @@ public void removeAllWidgets() {
...
-            if (filledWidth != 0 && filledWidth + size.x > sizeHint.x) {
-                result.x = Math.max(result.x, filledWidth);
-                result.y += filledHeight + verticalSpacing;
-                filledWidth = size.x + horizontalSpacing;
-                filledHeight = size.y;
+
+            if (widthOffset != 0 && widthOffset + horizontalSpacing + size.x <= boundingSize.x) {
+                // place widget in the current row
+                widthOffset += horizontalSpacing;
             } else {
```
skaldarnar, line 130: Only wrap when at least one element was placed:
```suggestion
            } else if (widthOffset != 0) {
```

skaldarnar:
This was merged as https://github.com/MovingBlocks/TeraNUI/pull/10 directly into the extracted TeraNUI.

PR #4108, opened 2020-07-30
feat(FlowLayout): Add option for right-to-left align

stefaniamak:
### Summary

Enhancement of FlowLayout: option to change the left-to-right align of included items that it has, to a right-to-left align.

### Additions

Added a variable named `rightToLeftAlign`, by which the module chooses which type of aligning to use at its FlowLayout.

#### Before the use of `rightToLeftAlign`

Before reaching the presented changes, I had changed the code as for it to apply only left-to-right alignment. Afterwards, I incorporated the boolean and combined the before and after. Seeing the merged version right away might be tricky to get what is for which alignment, thus, you can check:

 - left-to-right  code here:
https://github.com/MovingBlocks/Terasology/blob/7a86cd19a2f436740c87513166a58acf091897f7/engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java#L65-L80
 - right-to-left code here:
```java
        int filledWidth = canvas.size().x;
        int filledHeight = 0;
        int heightOffset = 0;
            filledWidth -= size.x;
            if (filledWidth != canvas.size().x - size.x && filledWidth < 0) {
                heightOffset += filledHeight + verticalSpacing;
                filledWidth = canvas.size().x - size.x;
                filledHeight = 0;
            }
            canvas.drawWidget(widget, Rect2i.createFromMinAndSize(filledWidth, heightOffset, size.x, size.y));
            filledWidth -= horizontalSpacing;
            filledHeight = Math.max(filledHeight, size.y);
        }
    }
}
```

Basically:
-  anchor point of first item in a row
     - **left-to-right** anchor point is 0
     - **right-to-left** anchor point is the canvas width minus the widget length
 - wrapping
     - **left-to-right** wraps up when the widget's width exceeds the remaining space at the row
     - **right-to-left** wraps up when the next widget's anchor point is below 0 (can't start from outside the FlowLayout!)

### How to test

1. Check https://github.com/Terasology/LightAndShadow/pull/141
2. Start a LAS game
3. Notice the quickslot's items _(should get the 1st result screenshot below)_

_or_ if you don't want to test it with LightAndShadow, you can:

1. Find a `.ui` file that uses a FlowLayout - in which are added **at least** 2 items/widgets
2. add the line `"rightToLeftAlign": true,` right below the line `"type": "flowLayout",`
3. run the module you applied the changes at, and check that flowLayout's included item's aligning. Result should be that items are added from right-to-left instead of the original left-to-right

I am not that fond of how many times I check if `rightToLeftAlign` boolean is true, but I wanted to avoid repeating the rest of the code as much as possible.

#### Result
image image

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layouts/FlowLayout.java
@@ -35,8 +35,13 @@ public class FlowLayout extends CoreLayout<LayoutHint> {
     private List<UIWidget> contents = Lists.newArrayList();
 
     /**
-     * The alignment of widgets; true = left-to-right align, no change / false = right-to-left
-     * This variable shall be set true only within the ui file that the FlowLayout is used
+     * Whether the directional flow of this layout goes from left-to-right and right-to-left.
+     * <p>
+     * The children are laid out from left-to-right by default (false), aligned at the left border of the canvas. If
+     * this toggle is explicitly enabled (true) the children are laid out right-to-left, aligned at the right border of
+     * the canvas.
+     * <p>
+     * This toggle can be set programmatically or in {@code .ui} files that use the Flow layout.
      */
     @LayoutConfig
     private boolean rightToLeftAlign = false;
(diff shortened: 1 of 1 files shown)


jdrueckert:
Updated the description to switch right-to-left and left-to-right.
The original alignment is left-to-right: WidgetA -> Widget B -> Widget C
The new alignment option is to change this to right-to-left: Widget C <- Widget B <- Widget A

PR #4113, opened 2020-08-02
fix: Increase connection timeout for clients as heavier module setups can take a while

Cervator:
Our JS (now including ClimateConditions, WildAnimals, etc) play test server was taking too long for clients to successfully connect before timing out. This seems to fix it.


Cervator:
We still hit timeouts after this, so I bumped further via https://github.com/MovingBlocks/Terasology/commit/0216049cad6ce2b8bab7763fbb21da6c2eca7242 - now it affects two different timeout types, both connectivity and chunk gen related.

PR #4117, opened 2020-08-06
feat: new run config to recreate the most recently launched game auto…

### LINKED ISSUE #4061: Create a game world directly, skipping the UI

Elyahu41:
### Contains

Fixes  #4061

### How to test

Run the "recreate latest" configuration

Should probably add comments/javadoc for these configurations. (load lastest and create latest)

### DIFF
--- a/.idea/runConfigurations/TerasologyPC__recreate_latest_.xml
+++ b/.idea/runConfigurations/TerasologyPC__recreate_latest_.xml
@@ -1,0 +1,12 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="TerasologyPC (recreate latest)" type="Application" factoryName="Application">
+    <option name="MAIN_CLASS_NAME" value="org.terasology.engine.Terasology" />
+    <module name="Terasology.facades.PC.main" />
+    <option name="PROGRAM_PARAMETERS" value="-homedir -noCrashReport -createlastgame" />
+    <shortenClasspath name="MANIFEST" />
+    <option name="VM_PARAMETERS" value="-Xms256m -Xmx1536m" />
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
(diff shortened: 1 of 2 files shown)


.idea/runConfigurations/TerasologyPC__recreate_latest_.xml
```diff
@@ -0,0 +1,12 @@
...
+    <option name="MAIN_CLASS_NAME" value="org.terasology.engine.Terasology" />
+    <module name="Terasology.facades.PC.main" />
+    <option name="PROGRAM_PARAMETERS" value="-homedir -noCrashReport -createlastgame" />
+    <shortenClasspath name="MANIFEST" />
+    <option name="VM_PARAMETERS" value="-Xms256m -Xmx1536m" />
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
```
Elyahu41, line 12: run configuration

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -150,13 +153,32 @@ public static void main(String[] args) {
...
+                            if (!title.startsWith("New Created")) { //if first time run
+                                newlyCreatedGM.setTitle("New Created " + title + " 1");
+                            } else { //if not first time run
+                                String newTitle = title.substring(0, title.length() - 2);
+                                //-2 to skip the last number added
+                                char fileNumber = title.charAt(title.length() - 1);
+                                fileNumber++;
+                                newlyCreatedGM.setTitle(newTitle + " " + fileNumber);
+                            }
+                            engine.changeState(new StateLoading(newlyCreatedGM, NetworkMode.NONE));
```
Elyahu41, line 181: I was able to get it to increment instead of a random number, I couldn't get parentheses or brackets around the number, so I just added it to the end. This code shouldn't run into any problems, unless someone creates a world with a name that starts with "New Created" and end with a number. Which I doubt will happen.

Elyahu41:
Noticed that the char datatype will stop at 9 and then go into unwanted numbers and symbols.

PR #4118, opened 2020-08-06
feat(JOML): migrate color

pollend:
toVector3f, toVector3i and toVector4f are not used in any of the modules under omega or the engine. If its not used then It would probably be ok to just remove the methods entirely.  I also update some of the documentation at the same time.  

Any reason why we would want to keep these methods?


Cervator:
This looks reasonable and I agree it looks unused, no reason to not just remove.

However, in this case it is a NUI class and since we're trying to extract that I wouldn't bother with cleanups in the code here, but would recreate this change against https://github.com/MovingBlocks/TeraNUI/blob/master/nui/src/main/java/org/terasology/nui/Color.java instead

PR #4122, opened 2020-08-11
perf: simplified check for rendering ready.

### LINKED ISSUE #4007: optimize RenderableWorldImpl.queueVisibleChunks
I did a little profiling, and a great deal of the CPU usage and memory allocations I see every tick go to RenderableWorldImpl.queueVisibleChunks.

That's made up of
* isChunkVisible
* isChunkVisibleReflection
* and mostly isChunkValidForRender

I believe it spends more CPU trying to decide which chunks to render than it does actually rendering the things in the chunks. #1886 is an excellent analysis of just how many chunks need to be loaded and sent to the client before the client may render things, but what it _doesn't_ discuss is how much of this work is done every tick.

RenderableWorld-profile z

(hmm, I wonder if there's a place to host profile output so it's web-viewable.)

I'm curious to see if the `isChunkVisible` methods will change in performance after ViewFrustum gets joml-ified.

Optimizing isChunkValidForRender will take something else. https://github.com/MovingBlocks/Terasology/blob/d9fdb02b1ed3a1fa8276a239aacdb6282da911ad/engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java#L380-L385

it seems like that getWorldViewAround call is not particularly cheap, and all it does with it in the end is check whether or not it's `null` and throws it away!

DarkWeird:
### Contains

Remove unnecessary check for rendering ready.
Removed condition saves 10ms-15ms. (activity "Queueing Visible Chunks" takes ~18ms)
Removed condition check that near chunks ready too. It's useless because LightMerger(chunk final loading) made it valid for rendering early.

Fixes #4007 

### How to test
Try compare FPS changes :(
Or compare "Queueing Visible Chunks" activity changes.

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
@@ -380,10 +380,6 @@ class RenderableWorldImpl implements RenderableWorld {
         return chunk.isReady();
     }
 
-    private boolean areSurroundingChunksLoaded(RenderableChunk chunk) {
-        return worldProvider.getWorldViewAround(chunk.getPosition()) != null;
-    }
-
     private boolean isChunkVisibleFromMainLight(RenderableChunk chunk) {
         return isChunkVisible(shadowMapCamera, chunk); //TODO: find an elegant way
     }
--- a/engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
@@ -377,7 +377,7 @@ class RenderableWorldImpl implements RenderableWorld {
     }
 
     private boolean isChunkValidForRender(RenderableChunk chunk) {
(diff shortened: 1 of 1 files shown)


keturn:
I'd been wondering about ways of making that check cheaper. I didn't think to ask if it could be removed entirely!

I guess, in the absence of anyone who can speak up as to what sort of edge case that is guarding against, this is a "if it needed to be there (and the failure mode isn't obvious), there should be a test for it" sort of thing.

DarkWeird:
@keturn I must to proove that edge case not exists now then. :3 otherwise I can to cache for it.
P.S.
I think that it os outdated after introduce LighMerger(needs dig in commit history)

@Cervator this code by manu3D :(

pollend:
welp, I guess you will need to become the expert @DarkWeird .

Cervator:
Going to take a chance on this and merge it, since it seemingly tests out yet any hidden meaning might remain in the dark without somebody truly able to speak to it. Better have less code and keep an eye out than leave in mystery code and wonder if we're missing major performance boosts for no reason? 🤷

PR #4123, opened 2020-08-12
Remove negative value capping from Abstract Value Modifiable Event

ktksan:
Removes negative Value capping in Abstract Value Modifiable Event. With this PR, the event shall support all values both positive and negative. And the new formula for calculating the result value would now be:
>  result = (<baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>

instead of the original : 

> result = max(0, (<baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>)


skaldarnar:
This could probably make use of [ModifiableValue](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/utilities/modifiable/ModifiableValue.java) as backing type in the event. 

... off to vacation 🥾 🌲 ⛺

ktksan:
@skaldarnar Have a great vacation!
(In case you do read this.. I had a doubt)
I'm not sure how ModifiableValue would work in here though.. since modifiableValue has its own separate list of multipliers and modifiers, and while trying to use this event, the result value calculation would require the multipliers and modifiers in the event and not the variable.... and I just feel it could get a bit complicated to understand if we try to use modifiableValue in here and try to map the modifiers and multipliers in there...

ktksan:
@Cervator the way to test this PR is slightly complicated. All of the current uses of AbstractValueModifiableEvent that I know of use only positive values anyway (GetMaxSpeedEvent, AffectJumpForceEvent - both of which don't have any negative value calculations anyway).
The only place where the negative value has been used is the BTS which was till now using a work around. So..., the only way to test this PR is with the attached AffectBodyTemperature PR which gets rid of the workaround.. And even in there after getting rid of the work-around some conditions have to be met to test the negative value in particular... the environment temperature needs to be lower than the body temperature. When this happens, as long as the player is in an area where the  env temp is lower than the body temp the bodytemp won't change because of the negative base value for the event but as soon as the player goes into a region with higher temp, the body temp starts increasing as it should.

ktksan:
Also I haven't really worked with unit tests... will have to look into - how to make one..😅

casals:
I get the logic and the code changes seem sound, but still couldn't test it locally (I tried following your last comment). Unit tests are important, but since it requires getting rid of the workaround you mentioned, perhaps detailed testing instructions (step-by-step) are in order.

ktksan:
Detailed Testing Instructions : https://github.com/Terasology/ClimateConditions/pull/29#issuecomment-674875942

PR #4124, opened 2020-08-14
feat(JOML): add orientation for Rotation

pollend:
This is a simple change for the orientation of the Rotation. This is just an api change that should help with migrating changes for orientation. there are no direct impacts at the moment. I also updated the test case to test for consistency.

### DIFF
--- a/engine-tests/src/main/java/org/terasology/testUtil/TeraAssert.java
+++ b/engine-tests/src/main/java/org/terasology/testUtil/TeraAssert.java
@@ -16,6 +16,7 @@
 package org.terasology.testUtil;
 
 import com.google.common.collect.Lists;
+import org.joml.Quaternionfc;
 import org.joml.Vector2fc;
 import org.joml.Vector3fc;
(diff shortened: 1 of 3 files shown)


DarkWeird (approved):
looks good.

pollend:
I think this looks good other then the naming I had concerns with for rotation.

jdrueckert (changes requested):
Compilation fails with 
```> Task :engine-tests:compileJava
/home/jenkins/agent/workspace/Terasology_engine_PR-4124/engine-tests/src/main/java/org/terasology/testUtil/TeraAssert.java:145: error: method assertEquals(Quaternionfc,Quaternionfc,float) is already defined in class TeraAssert
    public static void assertEquals(Quaternionfc expected, Quaternionfc actual, float error) {
                       ^
```
@pollend @DarkWeird Please have another look.

pollend:
@jdrueckert I had to create a new PR since this commit is before bullet and NUI.

PR #4136, opened 2020-08-24
feat migrate nui canvas for Colorc

pollend:
The changes in TeraNui need to be updated in Terasology.

issue: https://github.com/MovingBlocks/TeraNUI/pull/19


engine/src/main/java/org/terasology/rendering/assets/font/FontMeshBuilder.java
```diff
@@ -18,6 +18,7 @@
 import com.google.common.collect.Maps;
 import org.terasology.math.geom.Vector3f;
 import org.terasology.nui.Color;
+import org.terasology.nui.Colorc;
```
DarkWeird, line 21: `Color` is unused now?

Cervator: Fixed 👍 Same in another couple spots

PR #4101, opened 2020-07-28
feat(JOML): Migrate side#getVector3i() to direction()

pollend:
I migrated getVector3i to direction.  direction better describes the a side then getVector3i but this is an API change. There is no direct impact to modules and the change is pretty straightforward. The other way to go about this is change the API and resolve it across all the modules.

Issue #3559, opened 2018-11-27
Game crashes when block next to stacked fences is destroyed


ebackhaus:
### What you were trying to do
Place fence below floating fence then destroy a connected block to place another fence.

### What actually happened
Game crashes when the block connected to the stacked fence is destroyed.

### How to reproduce

* Place fence
* Destroy block below fence
* Place fence below floating fence
* Destroy block connected to the stacked fence

![game crashes when block next to stacked fences is destroyed](https://user-images.githubusercontent.com/42822402/49118645-c1ff7680-f26a-11e8-9d00-c10ddeba4dd2.gif)

### Log details and game version
v2.1.0 - Alpha 11

### Computer details
Windows 10 64-bit
RAM 12 GB

portokaliu:
I just tried this on version 78, and it seems fences break all physics in the game when you destroy a block adjacent to a fence which has an item near it.

Repro: Dig a hole in between other blocks, don't pick up the dropped block, place fence in dug out area, dig any adjacent area on the sides.

Result: No block will have a rigidbody _except_ for the block you just placed. 

Observations: It does not seem to happen when the block is surrounded by only 3 other blocks on the sides. Same for destroying blocks above or below while surrounded by less than 4 blocks on the sides.

Fun fact: after destroying that block while in ghost mode, you're able to _see_ the air block (with the in-game thing that tells you block names)

jellysnake:
@pollend could this be caused by some interactions in the Block Family code?
I suppose testing to see if any blocks using a similar family definition also break it.

pollend:
potentially but I can't really think of a situation where this would happen. you can try using git-bisect.

Qwertygiy:
Reproduced with a fresh build today; issue is still active.

Here is the log results upon breaking the block: https://pastebin.com/fjJsdX40
The second of the two errors is repeated every tick until the game is paused.

Once the game is exited and reloaded, physics seem to function normally once again.

Qwertygiy:
I've discovered why this happens.

The fence cross shape (when it connects to 4 blocks) has 3 colliders.

The fence T shape (when it connects to 3 blocks) has 2 colliders.

The block entity that is interacting with the fence when it changes shape winds up looking for that third collider, which is now outside the bounds of the array.

I tested this by adding an extra collider to the T shape, and the error no longer occurred. But, of course, that isn't a proper solution.

[...]

PR #4139, opened 2020-08-28
backport changes from old bullet branch

pollend:
This is a backport of my changes from my old bullet branch I need to change how the binary gets loaded but with the optimization flag this is pretty much on par with JBullet. We need to do some more testing and judge if this could work into the existing line up. So I moved the library path to be defined as `org.terasology.librarypath`. 

issue: https://github.com/MovingBlocks/Terasology/pull/3809

this migrates some methods to JOML and these PR's should make the process easier:
https://github.com/MovingBlocks/Terasology/pull/4130
https://github.com/MovingBlocks/Terasology/pull/4124
https://github.com/MovingBlocks/Terasology/pull/4101
https://github.com/MovingBlocks/Terasology/pull/4051


build.gradle
```diff
@@ -73,6 +73,8 @@ repositories {
     mavenCentral()
     mavenLocal()
 
+    mavenLocal()
```
Cervator, line 76: Dupe, already there a couple lines up

engine-tests/src/test/java/org/terasology/world/block/shape/BlockShapeTest.java
```diff
@@ -44,46 +42,43 @@
...
 
         BulletConvexHullShape bulletConvexHullShape = (BulletConvexHullShape) shape;
 
-        ObjectArrayList<javax.vecmath.Vector3f> points = ((com.bulletphysics.collision.shapes.ConvexHullShape) bulletConvexHullShape.underlyingShape).getPoints();
-        for (int x = 0; x < points.size(); x++) {
-            fuzzVectorTest(test[x], VecMath.from(points.get(x)));
-
-        }
-
+//        Vector3f[] points = ((ConvexHullShape) bulletConvexHullShape.underlyingShape).getVertices();
```
Cervator, line 78: Commented out? Better to either leave a TODO for why / how to restore, or just take it out if it is no longer relevant.

engine/build.gradle
```diff
@@ -110,6 +110,8 @@ dependencies {
     api group: 'com.miglayout', name: 'miglayout-core', version: '5.0'
     implementation group: 'de.matthiasmann.twl', name: 'PNGDecoder', version: '1111'
 
+    compile group: 'org.terasology', name: 'JNBullet', version: '0.3.2-SNAPSHOT'
```
Cervator, line 113: Being that this is one of our internal deps shouldn't it be back in the group where `tera-bullet` was? Also, `compile` is deprecated - should either be `implementation` or `api`

engine/src/main/java/org/terasology/engine/subsystem/lwjgl/BaseLwjglSubsystem.java
```diff
@@ -15,6 +15,8 @@
  */
 package org.terasology.engine.subsystem.lwjgl;
 
+import com.badlogic.gdx.physics.bullet.Bullet;
+import com.badlogic.gdx.physics.bullet.BulletBase;
```
Cervator, line 19: Just looking on GitHub, but it doesn't seem like this import gets used? Also is a LWJGL subsystem the best place for physics? It might be, but it the physics isn't strictly related to LWJGL.

engine/src/main/java/org/terasology/logic/characters/KinematicCharacterMover.java
```diff
@@ -459,8 +459,9 @@ private boolean moveHorizontal(Vector3f horizMove, CharacterCollider collider, V
         }
         boolean horizontalHit = false;
         Vector3f normalizedDir = horizMove.normalize();
-        if(!normalizedDir.isFinite())
+        if(!normalizedDir.isFinite()) {
```
Cervator, line 462: Oh boy, we still had bare `if`s? Yikes :-)

Trivial: One more space needed here

engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
```diff
@@ -96,59 +96,55 @@
 /**
  * Physics engine implementation using TeraBullet (a customised version of JBullet).
  */
+@RegisterSystem
 public class BulletPhysics implements PhysicsEngine {
+    public static final int AABB_SIZE = Integer.MAX_VALUE;
+
+    public static final float SIMD_EPSILON = 1.1920929E-7F;
```
Cervator, line 103: Wut. This constant's name is probably sensible enough, but most people probably won't know what that means anyway (I'm one of them). Might be good with with a one-liner javadoc to explain this field?

engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
```diff
@@ -429,31 +450,30 @@ public boolean hasCharacterCollider(EntityRef entity) {
...
+            other.activate(true);
+        }
+        discreteDynamicsWorld.removeCollisionObject(ghost);
     }
 
     @Override
     public float getEpsilon() {
-        return BulletGlobals.SIMD_EPSILON;
+        //TODO: figure out how access this from libgdx
+        return 1.19209290e-07f;
```
Cervator, line 471: How does this relate with the constant in this class that's also about Epsilon, and in fact shares the _same_ name as the old version of the getter? Same thing? This seems a _very_ magic number

engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
```diff
@@ -308,13 +327,13 @@ public boolean updateRigidBody(EntityRef entity) {
             return false;
         } else if (rigidBody != null) {
             float scale = location.getWorldScale();
-            if (Math.abs(rigidBody.rb.getCollisionShape().getLocalScaling(new Vector3f()).x - scale) > BulletGlobals.SIMD_EPSILON
+            if (Math.abs(rigidBody.rb.getCollisionShape().getLocalScaling().x - scale) > this.getEpsilon()
```
Cervator, line 330: Use the getter, or the constant directly?

engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
```diff
@@ -597,7 +620,11 @@ private synchronized void processQueuedBodies() {
         }
         while (!removalQueue.isEmpty()) {
             BulletRigidBody body = removalQueue.poll();
+            if(body.isDisposed) {
```
Cervator, line 623: Trivial: code style (missing space) - more of them too, won't comment on each, but may want to run the Checkstyle plugin on each modified file to catch these :-)

engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
```diff
@@ -646,62 +677,66 @@ private void removeRigidBody(BulletRigidBody body) {
      * <br><br>
      * TODO: Flyweight this (take scale as parameter)
```
Cervator, line 678: Is this TODO still valid? If so maybe could it be expanded on a bit to make it clear what it means?

engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
```diff
@@ -710,72 +745,78 @@ private PairCachingGhostObject createCollider(Vector3f pos, ConvexShape shape, s
...
-                for (PersistentManifold manifold : manifolds) {
+                for (int y = 0; y < manifolds.size(); y++) {
+
+                    btPersistentManifold manifold = manifolds.atConst(y);
                     for (int point = 0; point < manifold.getNumContacts(); ++point) {
-                        ManifoldPoint manifoldPoint = manifold.getContactPoint(point);
-                        if (manifoldPoint.getDistance() < 0) {
+                        btManifoldPoint manifoldPoint = manifold.getContactPoint(point);
+                        if (manifoldPoint.getDistance() < 0.f) {
+                            Vector3f a1 = new Vector3f();
```
Cervator, line 791: More meaningful variable names than `a1` 2 and 3? Seeing more, than those three as well

engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
```diff
@@ -790,128 +831,151 @@ public boolean needsCollision(BroadphaseProxy proxy0) {
...
-        public void setTransform(org.terasology.math.geom.Vector3f location, org.terasology.math.geom.Quat4f orientation) {
-            rb.getWorldTransform(pooledTransform);
-            pooledTransform.origin.set(VecMath.to(location));
-            pooledTransform.setRotation(VecMath.to(orientation));
-            rb.proceedToTransform(pooledTransform);
+        public void setLocation(Vector3f location) {
+            Matrix4f translation = rb.getWorldTransform();
+            Quaternionf quaternion = new Quaternionf().setFromUnnormalized(translation);
+
+//            Quat4f quaternion = new Quat4f();
```
Cervator, line 919: More commented out code - either remove if it is no longer needed or put a comment / TODO to explain why it is still there please :-)

Seeing at least one more of these, will skip commenting on any more

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletConvexHullShape.java
```diff
@@ -15,47 +15,58 @@
...
     }
 
     @Override
     public Vector3f[] getVertices() {
-        return convexHullShape.getPoints().stream()
-                .map(VecMath::from)
-                .toArray(Vector3f[]::new);
+        Vector3f[] verts = new Vector3f[convexHullShape.getNumPoints()];
+        for(int x = 0; x < convexHullShape.getNumPoints(); x++){
+            Vector3f p = convexHullShape.getScaledPoint(x);
```
Cervator, line 67: Any reason this isn't just one line instead of two? Assigning this `Vector3f` to `verts` next

engine/src/main/java/org/terasology/physics/bullet/shapes/BulletCylinderShape.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2018 MovingBlocks
```
Cervator, line 2: Outdated copyright statements (new one can be generated from within IntelliJ)

engine/src/main/java/org/terasology/physics/engine/PhysicsEngineManager.java
```diff
@@ -39,6 +41,6 @@
      * @return The created {@link PhysicsEngine} instance.
      */
     public static PhysicsEngine getNewPhysicsEngine(Context context) {
-        return new BulletPhysics(context.get(WorldProvider.class));
+        return new BulletPhysics();
```
Cervator, line 44: This seems like a change that might be meaningful to explain: as per the _name_ of this method it could make perfect sense to return a new physics engine, but what did adding the `WorldProvider` before achieve that isn't needed now? And if we don't need the `Context` here should that be removed as a method param?

engine/src/main/java/org/terasology/physics/engine/PhysicsSystem.java
```diff
@@ -146,7 +147,7 @@ public void updateRigidBody(OnChangedComponent event, EntityRef entity) {
 
     @ReceiveEvent(components = {BlockComponent.class})
     public void onBlockAltered(OnChangedBlock event, EntityRef entity) {
-        physics.awakenArea(event.getBlockPosition().toVector3f(), 0.6f);
+        physics.awakenArea(new org.joml.Vector3f(JomlUtil.from(event.getBlockPosition())), 0.6f);
```
Cervator, line 150: Unnecessarily qualified class? I see the old `Vector3f` import get removed, so is this needed? Haven't asked my IDE tho

engine/src/main/java/org/terasology/physics/engine/PhysicsEngineManager.java
```diff
@@ -21,6 +21,8 @@
 import org.terasology.physics.shapes.CollisionShape;
 import org.terasology.physics.shapes.CollisionShapeFactory;
 import org.terasology.world.WorldProvider;
+import org.terasology.world.block.BlockManager;
```
Cervator, line 24: Mystery new imports?

engine/src/main/java/org/terasology/physics/engine/VoxelWorldSystem.java
```diff
@@ -0,0 +1,145 @@
...
+import org.terasology.world.chunks.event.BeforeChunkUnload;
+import org.terasology.world.chunks.event.OnChunkLoaded;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+import static org.terasology.physics.bullet.BulletPhysics.AABB_SIZE;
+
+@RegisterSystem
+public class VoxelWorldSystem extends BaseComponentSystem {
```
Cervator, line 58: This _very_ much looks like the sort of class that could use some solid javadoc :-) And would it make sense to make the `@RegisterSystem` explicitly `ALWAYS` ? If that's intended

engine/src/main/java/org/terasology/physics/engine/VoxelWorldSystem.java
```diff
@@ -0,0 +1,145 @@
...
+            wrapper = new VoxelCollisionAlgorithmWrapper(ChunkConstants.SIZE_X, ChunkConstants.SIZE_Y, ChunkConstants.SIZE_Z);
+            worldShape = new btVoxelShape(wrapper, new Vector3f(-AABB_SIZE, -AABB_SIZE, -AABB_SIZE), new Vector3f(AABB_SIZE, AABB_SIZE, AABB_SIZE));
+
+            Matrix4f matrix4f = new Matrix4f();
+            btDefaultMotionState blockMotionState = new btDefaultMotionState(matrix4f);
+
+            blockConsInf = new btRigidBodyConstructionInfo(0, blockMotionState, worldShape, new Vector3f());
+            rigidBody = new btRigidBody(blockConsInf);
+            rigidBody.setCollisionFlags(btCollisionObject.CollisionFlags.CF_STATIC_OBJECT | rigidBody.getCollisionFlags());
+            short mask = (short) (~(StandardCollisionGroup.STATIC.getFlag() | StandardCollisionGroup.LIQUID.getFlag()));
```
Cervator, line 90: This and other bits of code in this class is kinda ... wild. And unusual. It might make sense if you're familiar with native physics code but if not I'd find this pretty opaque. More comments at least, perhaps?

engine/src/main/java/org/terasology/physics/engine/VoxelWorldSystem.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @ReceiveEvent(components = {WorldComponent.class})
+    public void onNewChunk(OnChunkLoaded chunkAvailable, EntityRef worldEntity) {
+        Vector3i chunkPos = chunkAvailable.getChunkPos();
+        Chunk chunk = chunkProvider.getChunk(chunkPos);
+        ByteBuffer buffer = ByteBuffer.allocateDirect(2 * (ChunkConstants.SIZE_X * ChunkConstants.SIZE_Y * ChunkConstants.SIZE_Z));
+        buffer.order(ByteOrder.nativeOrder());
+        short[] entries = new short[ChunkConstants.SIZE_X * ChunkConstants.SIZE_Y * ChunkConstants.SIZE_Z];
+        for (int x = 0; x < ChunkConstants.SIZE_X; ++x) {
```
Cervator, line 127: Really not sure what this does, but it looks expensive. And is it intentional to use the prefix version of `++` ? That sticks out as unusual and therefore bug prone? Especially if not explained well.

engine/src/main/java/org/terasology/physics/engine/VoxelWorldSystem.java
```diff
@@ -0,0 +1,145 @@
...
+                    int index = (z * ChunkConstants.SIZE_X * ChunkConstants.SIZE_Y) + (x * ChunkConstants.SIZE_Y) + y;
+                    Block block = chunk.getBlock(x, y, z);
+                    short id = block.getId();
+                    tryRegister(block);
+                    entries[index] = id;
+
+                }
+            }
+        }
+        for (int k = 0; k < entries.length; k++) {
```
Cervator, line 139: Why `k`

pollend:
This was mostly to just get the ball rolling again. so some of the changes that @eviltak  made a while ago are kind of in a funny state with this. umm, not sure hat exactly to do with some of these changes. The abstraction is in a funny middle ground.

engine/src/main/java/org/terasology/physics/engine/PhysicsEngineManager.java
DarkWeird: world used by LiquidWorldWrapper and PhysicsWorldWrapper.
is handles "static body collision" (?)
I don't known how this doing now :D
candidates for check

DarkWeird (changes requested):
It is looks pretty for me.

DarkWeird (approved):

Cervator:
With https://github.com/MovingBlocks/JNBullet/pull/9 and natives from there this now works on my Mac 👍 

Also needed https://github.com/Terasology/Inventory/pull/26

Cervator:
Can confirm I've now tested this on both Mac and Windows, with the shiny new v1 release of JNBullet, complete with natives built for all three OSes via Jenkins! Hooked up work Mac as a static build agent and JNBullet's `Jenkinsfile` now correctly builds most the stuff on Linux then just the Mac natives on a different agent, stashing the file needed from there, and eventually cramming it all into a zip that goes into Artifactory. Testing includes multiplayer with railgunning galore and that part seems alright!

Upon closer testing however 
it does seem like there might be some quirks:

- [ ] Unit test fails are legit, the same 17 fail locally as well (I particular wonder about unit tests in modules as well)
- [ ] If you place a free standing fence pole on a flat surface there's plenty of space normally to target and hit the ground immediately below it. With the physics from here that doesn't work - if you target the fence block directly it takes damage, but any attempt to target through the blockspace causes damage to seemingly nothing. Block below is unaffected
- [ ] I placed a fence pole then broke adjacent dirt to break the dirt under the fence pole. Was able to partly move past along the sides of the fence block which might be okay, but oddly after doing it a few times i suddenly and immediately died of physical damage 🤔 Not sure if new with the physics or if that's a rare manifestation of the quirk where if you rub up against corners or something it might kill you ... (on the plus side I couldn't replicate #3559!). Tried a second time and didn't die that time.
- [ ] If you try to push against a fence pole you might get weirdly stuck on it - can be somewhat hard to replicate and I'm not sure if it is new. In the below screenie I'm actually trying to move forward. If I keep slowly looking further and further to the left I'll eventually escape and push forward. I would expect to be slowly sliding since the fence post is a hard corner 

image

keturn:
Under Ubuntu 20.04 with Adopt OpenJDK 11.0.8, the Terasology.zip from Jenkins does look like it loads JNBullet:

```
Found the natives dir: …/natives
16:36:30.611 [main] INFO  c.b.gdx.physics.bullet.NativeSupport - Loading JNBullet Library:bullet
16:36:30.636 [main] INFO  c.b.gdx.physics.bullet.NativeSupport - 	Loaded from java.library.path: …/natives/linux/libbullet-linux-amd64.so
```

but that distribution contains no modules and the runtime does not seem to retrieve them by itself, so I can't say anything about how the gameplay works.

It did show the main menu, there was music and sound, and the usual log messages from the input system, so from that I think the usual jinput + lwjgl + openal libraries were not disrupted by this.

keturn:
It also succeeds in getting past the "Loading JNBullet" part of the logs under the current release candidate for OpenJDK 15. (Java VM: OpenJDK 64-Bit Server VM, version: 15+36-1562)

JDK 15 fails to load lwjgl, but that failure isn't particular to this branch, it happens in develop too.

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
```diff
@@ -155,6 +156,9 @@
      *                   audio and input subsystems.
      */
     public TerasologyEngine(TimeSubsystem timeSubsystem, Collection<EngineSubsystem> subsystems) {
+        // configure native paths
+        PathManager.getInstance();
```
DarkWeird, line 160: Why there?
What are means booleans?

engine/src/main/java/org/terasology/physics/bullet/EntityMotionState.java
```diff
@@ -16,47 +16,55 @@
...
+     * @param entity The entity to relate this motion state to and set the LocationComponent of.
      */
     EntityMotionState(EntityRef entity) {
+        super();
         this.entity = entity;
     }
 
     @Override
-    public Transform getWorldTransform(Transform transform) {
+    public void getWorldTransform(Matrix4f transform) {
```
DarkWeird, line 53: Are we using returns  somewhere? in modules?
getWorldTransform with returns void - sounds strange..

engine/src/main/java/org/terasology/engine/TerasologyEngine.java
pollend: when the native paths are defined I also set the java paths. Just calling an instance of PathManager should be enough to set this up.

engine/src/main/java/org/terasology/physics/bullet/EntityMotionState.java
pollend: this syncs changes between the LocationComponent and the bullet rigid body. this is for the JNI layer and its not exposed to engine.

Cervator (approved):

Cervator:
Merged everything locally and did some additional spot testing. Checks out, beyond the usual sorts of quirks - some dropped items don't really fall over in a natural way, rail carts go in the right _direction_ again now, but as usual don't really work well for non-hosting players in multiplayer. And the rail gun carnage is glorious 👍 

Will merge module PRs after engine build is going. Thanks for hanging in there on this one @pollend!

PR #4141, opened 2020-09-06
fix: Gradley weirdness about a Directory/DirectoryProperty thing

Cervator:
Was causing `gradlew distPCZip` to fail on my Win10 system, despite seemingly working fine in Jenkins, with some "invalid file path" thing as the new `distsDirectory` (replacement for deprecated `distsDi`) would turn into this long awkward property thing, not a directory. I guess maybe somehow on Linux that could translate into a valid directory? Dunno. I figure this'll work on both.

Also let a `.idea/misc.xml` change slip in that keeps toggling in my workspace for some reason. Lets see if it flips back for anybody else

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -12,5 +12,5 @@
     </writeAnnotations>
   </component>
   <component name="ExternalStorageConfigurationManager" enabled="true" />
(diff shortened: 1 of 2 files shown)


keturn:
Do you have any record of what the "invalid" value was?

Issue #4146, opened 2020-09-11
jopenvr.jar contains outdated version of JNA


keturn:
I tried to add a feature involving version 4.5.2 of JNA Platform, but it failed at runtime with

> Exception in thread "main" java.lang.NoSuchMethodError: 'java.util.List com.sun.jna.platform.unix.Resource$Rlimit.createFieldsOrder(java.lang.String[])'

I believe that when JNA Platform went to extend the Structure class from JNA, it found the one in `engine/libs/jopenvr.jar` instead of the one provided by the gradle dependency.

PR #4147, opened 2020-09-12
feat: set limits for maximum memory use.

keturn:
We've had some memory leaks where the game seems to take up far more of the system's memory than the amount allowed for Java's maximum heap size.

This uses some operating-system-specific functions to set limits on how much memory the process can allocate.

Because this is so system-dependent, I'm introducing this as a developer/debugger feature rather than intending it for production use.

(Players won't want the game taking runaway amounts of memory either, but getting consistent support for this and setting expectations for how it works would be a bigger project. Possibly would want to do it in Launcher, as some of these controls are better set before even launching the process.)

### Contains

Linux support only.

* Sets [RLIMIT_DATA](https://www.man7.org/linux/man-pages/man3/getrlimit.3p.html#DESCRIPTION) to limit the maximum amount of memory for the process's data segment.
* Sets [`oom_score_adj`](https://www.kernel.org/doc/html/latest/filesystems/proc.html#proc-pid-oom-adj-proc-pid-oom-score-adj-adjust-the-oom-killer-score) to give the out-of-memory reaper encouragement to choose this process for termination when the system is out of memory as a whole.

### How to test

On Linux:

check `/proc/PID/limits`

* test default
* test with low limit, see out-of-memory error
* test with high limit

On other platforms:

* make sure game starts as normal, i.e. that the Linux-only code didn't ruin things for everyone else.

- [x] Make limits configurable instead of hard-coded
- [x] Blocked by #4146 for the upgrade of JNA Platform


engine/build.gradle
```diff
@@ -96,7 +96,7 @@ dependencies {
     api group: 'org.codehaus.plexus', name: 'plexus-utils', version: '1.5.6'
 
     // Java magic
-    implementation group: 'net.java.dev.jna', name: 'jna-platform', version: '4.2.2'
+    implementation group: 'net.java.dev.jna', name: 'jna-platform', version: '4.5.2'
```
keturn, line 99: There's a new major version of JNA 5.x out since October 2018, but the 4.5 series is sufficient to provide the [Resource](https://javadoc.io/doc/net.java.dev.jna/jna/4.5.2/com/sun/jna/platform/unix/Resource.html) API and I thought I could get away with doing less testing if I stuck to the same major version.

Manually keeping versions in sync between here and facade's build.gradle is a bother. Perhaps after clearing the jna-in-jopenvr obstacle we can revisit this and see if we can avoid that duplication.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -477,4 +470,26 @@ private static int getLastNumber(String str) {
...
+        }
+    }
+
+    private static void adjustOutOfMemoryScore(int adjustment) {
+        Path procFile = Paths.get("/proc", "self", "oom_score_adj");
+        try {
+            Files.write(procFile, String.valueOf(adjustment).getBytes(),
+                    StandardOpenOption.WRITE, StandardOpenOption.TRUNCATE_EXISTING);
+        } catch (IOException e) {
+            e.printStackTrace();
```
keturn, line 492: Placeholder error handling, do something better here.

keturn:
Related research note: The other mechanism Linux provides for setting these sorts of constraints is through [cgroups (Control Groups)](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html). But I _think_ `setrlimit` is sufficient for this purposes without getting in to cgroups.

4Denthusiast:
If the process would otherwise exceed this limit, will it just crash, or will Java try garbage-collection?

keturn:
If the program is allocating memory on the heap and there's not enough free space within the maximum heap size set (`-Xmx` setting), that's when the JVM will try garbage collection.

The process limits should be set higher than that max JVM heap size.

In testing this I've seen Java throw an OutOfMemoryError sometimes, but other times it's not been so descriptive.

It probably depends on if the thing failing to allocate the memory is Java or if it's in native code (e.g. LWJGL or OpenAL).

keturn:
The JNA in jopenvr was removed in #4169, clearing the way for this. It also upgraded JNA so this PR can drop that part.

But since this adds command-line options, I'll say it's blocked by #4157. It's not _technically_ blocked but I'd rather get picocli merged first and then switch this to use picocli, as opposed to making the picocli PR do more work for this one.

keturn:
unblocked!!! now that #4157 gave us a better way to add command-line args

keturn:
[stricter limits for memory use on Linux](https://trello.com/c/m8IsoQan/159-stricter-limits-for-memory-use-on-linux)

facades/PC/build.gradle.kts
```diff
@@ -68,6 +69,16 @@ dependencies {
 
     // TODO: Consider whether we can move the CR dependency back here from the engine, where it is referenced from the main menu
     implementation(group = "org.terasology.crashreporter", name = "cr-terasology", version = "4.1.0")
+
+    implementation(platform("org.junit:junit-bom:5.7.1")) {
+        // junit-bom will set version numbers for the other org.junit dependencies.
+    }
+    implementation("org.junit.jupiter:junit-jupiter-api")
+    implementation("org.junit.jupiter:junit-jupiter-params")
```
keturn, line 77: FIXME: these are supposed to be test dependencies!

facades/PC/src/main/java/org/terasology/engine/DataSizeConverter.java
```diff
@@ -0,0 +1,56 @@
...
+package org.terasology.engine;
+
+import picocli.CommandLine;
+
+import java.math.BigDecimal;
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class DataSizeConverter implements CommandLine.ITypeConverter<Long> {
```
keturn, line 13: FIXME: docstring

keturn (commented):
This needs a few finishing touches, but it's working as expected.

DarkWeird (approved):
Nice! Works! OOM happens fast.

image

checked at:
default - works
256m - oom at splashscreen
2G - oom at game starting
8G - works

Notice strange thing: 
i runs game with -Xmx1536m.. (default at runconfiguration for TerasologyPC) and game takes more memory then maximum.
I suspect this is cause of this pr.

Issue #4126, opened 2020-08-16
Substantial performance issues in latest JS, possibly spawning at 0,0,0 in multiplayer


Cervator:
### What you were trying to do

Play test recent JoshariasSurvival with @ktksan's awesome new GSOC biome stuff, including enabling of WildAnimals and some temperature maps.

After initial play testing also lagged wildly at times even in single player https://github.com/Terasology/WildAnimals/pull/58 was submitted to cut excessive deer spawn. This appears to have helped _some_

Timeouts may also have been helped _some_ by #4113 - you really need to be patient in some cases trying to join!

### What actually happened

We had a variety of problems just trying to get a server stable, extensive load times, timeout failures trying to spawn in the world, massive lag spikes on the server with players trying to join/leave, to the point where a hosting client might completely stall for a minute or two while a client tries to join.

A weird consequence appears to be when you _do_ finally spawn you might be embedded deep in the ground _on the client's  view_ of the world - but the server thinks you're perfectly fine and happy at spawn, able to move around and everything (as observed by a host player) despite not moving at all on the client (because stuck in solid ground). Maybe this could be a desync between host and client, with the client having partially loaded without blocks to land on, falling for a while until chunks caught up and froze the player in place?

Edit: Got more info on that case, the 2nd player appeared stuck at exactly 0, 0, 0 - we've had bugs like that in the past. Something about a `LocationComponent` not initializing or attaching right to the 2nd player?

### How to reproduce

* Unsure - but seen both from source playing JS lately, as well as while testing [Omega dev build 1094](http://jenkins.terasology.org/job/DistroOmega/1094/).
* Host a multiplayer game of JS and survive loading it (may take a while)
* Try to join with a second client - may take longer, may freeze the host for a period of time
* After a long freeze for the host player everything may then work perfectly fine, no lag, can see the 2nd client fine, despite that player being embedded deep in the ground in their game window

### Log details and game version

 [Omega dev build 1094](http://jenkins.terasology.org/job/DistroOmega/1094/)

Some log snippets from near the end of the 2nd client's log (host log was unremarkable) - highlights: 

* Weird "unavailable entity" logs
* Odd warnings from DAG node stuff
* `ClimateMapDisplaySystem` is part of ktksan's changes and may just not be wired up right for multiplayer yet

```
...
[...]

PR #4148, opened 2020-09-13
fix(performance): optimize ModuleEnvironmentSandbox.getSubTypeIdentifier

keturn:
Always use the class's fully-qualified Java class name as its type identifier, as ReflectionsSandbox does.

The previous implementation led to calling loadClass on *all* of Component's subclasses every time the serializer asked for this type identifier. We encountered scalability problems with that in #4126.

### How to test

This changes identifiers used by serializers for pretty much everything, which could be a big deal!

Those are used for networking and save games.

So while it could break everything, if some things seem to be working, I don't have ideas of particular edge cases to test for.

Ideas welcome here.


Cervator:
Can confirm that this seems to resolve the performance issue, nice work @keturn! 🎉 

In host/client + client multiplayer I noticed no lag at all for either client _but_ it still took a full 2 minutes for player 2 to join the first time, so there's still something squirrely going on somewhere.

Player 1 could craft perfectly fine both in-hand and via workbench. Player 2 hit a known issue where crafting a crude weapon in a workbench leaves the output item invisible until picked up. Also the durability overlay on an item appears to have gone buggy and takes up way more space (down from the usual space, roughly a full inventory slot's size total) than it is meant to.

On trying to relog player 2 I hit the other issue where I logged in with the camera stuck at 0,0,0. Still was actually in the right location, could break blocks there and everything, but on the 2nd client itself you're stuck in the ground helplessly.

engine/src/main/java/org/terasology/persistence/typeHandling/reflection/SerializationSandbox.java
```diff
@@ -46,7 +33,9 @@
      * @param <T>      The base type whose subtype needs to be identified.
      * @return The unique identifier for {@code subType}.
      */
-    <T> String getSubTypeIdentifier(Class<? extends T> subType, Class<T> baseType);
+    default <T> String getSubTypeIdentifier(Class<? extends T> subType, Class<T> baseType) {
+        return subType.getName();
```
keturn, line 37: I think there are no longer any implementations of this that use `baseType` at all.

Should we change the interface and remove that parameter entirely?

If so, is this still a subtypeIdentifier, or is it just its Identifier?

Cervator: Probably just keep simplifying, if we need something like this in the future we could put it back on. Might still be good to hear from @eviltak 

Additionally serialization and/or the whole ECS likely would majorly change when reintegrated as libraries anyway, so we don't have to aim for perfection right now

Cervator:
Not sure why GitHub didn't notice me merging this - even shows the related merge commit in the activity list 🤔 

Closing this as merged - tested several additional rounds of multiplayer separately with CoreSampleGameplay, ManualLabor, and JoshariasSurvival and submitted two new issues to cover any leftovers: #4152 and #4154

Many thanks @keturn ! Blocker resolved, but sure more stuff we can fix. I didn't see the spawn at 0,0,0 even once during testing at least :-)

Cervator:
Looks like I was missing the move commit locally for some reason despite having fetched 🤔 Okay trying to merge a second time, maybe then this will auto-close as merged.

PR #3969, opened 2020-05-19
Upgrading LWJGL to version 3

DarkWeird:
### Contains

LWJGL change:
 * Windowing system - GLFW
 * Input system - GLFW
 * GL methods, which use buffers (+suffix)
 * Some GL Extensions moved

We can replace jinput library to GLFW for joystics(controllers) after merge.

### Related PRs

https://github.com/MovingBlocks/TeraNUI/pull/24 - after extraction NUI
https://github.com/MovingBlocks/SplashScreen/pull/4 - GLFW splash screen's classes.  (Old Swing screen crashes LWJGL3's OpenGL initialization - deep side effect)
https://github.com/Terasology/CoreRendering/pull/15 - Update CoreRendering for lwjgl3

### How to test

Checkout https://github.com/Terasology/CoreRendering/pull/15.
Run and play as usual, check :)

Platforms:
- [x] - MacOs
- [ ] - Windows
- [x] - Linux(X)
- [ ] - ~~Linux(Wayland) - non goal, GLFW provides support for Wayland~~   LWJGL3 haven't proper code. possibly in next release

Checklist:
 * Input system:
    * Keyboard:
        * keys working, keybindings working
        * chat and console receive chars as usual  - your language work, special symbols work, uppercase correct
    * Mouse:
        * position in NUI work correctly
        * scroll works
        * mouse acceleration  works correctly
 * Windowing system:
    * Windows resizing works
    * Fullscreen works
    * Pseudo fullscreen works
 * Rendering works, not glitches
 * OpenVR works (I hope)
 * Natives publishes in distribution
 * Check switchable graphics on MacOS

[bonus] measure performance on discrete graphics

### Outstanding before merging

- [x] Return ```SkySphere```, LWJGL 3 lost ```glu``` packages
- [x] Merge GLFW's char and key callbacks products... or separate NUIs key events
      Its works separatelly now. Generates two independent ```KeyboardEvent``` streams. 
      [!] You are moving when type in chat or console in-game!
- [x] Repair type adapter for configuration (Resolution not saves)
- [x] ~~Repair ```LwjglPortlet``` (for TeraEd, or use another way)~~
- [x] Repair framelimit
- [x] Fix ```Resolution``` in settings gui (used Object's toString)
- [x] Expose LWJGL 3 to modules
- [x] Saving window position
- [x] Refactor Debug/Error outputs for LWJGL 3
- [x] Overall refactoring
[...]

PR #4157, opened 2020-09-22
chore[facade]: use picocli for processing command line options

keturn:
This PR switches PC facade's handling of command-line options to use [picocli](https://picocli.info).

The hope is that it improves input validation, provides better feedback to users, and makes it easier to implement new command-line options.

### How to test

test out all the command line options!

- [ ] What do we do about Launcher? https://github.com/MovingBlocks/TerasologyLauncher/issues/589
- [x] re-add extended help message with examples
- [x] update https://github.com/MovingBlocks/Terasology/blob/develop/docs/Playing.md
- [x] make sure `--homedir --no-crash-report` isn't a problem (it was using '--no-crash-report` as the value of homedir!)
    - filed upstream as https://github.com/remkop/picocli/issues/1197
    - local workaround in place: fail if homedir starts with `-`


.idea/compiler.xml
```diff
@@ -2,7 +2,17 @@
...
+        <outputRelativeToContentRoot value="true" />
+      </profile>
+      <profile name="Gradle Imported" enabled="true">
+        <outputRelativeToContentRoot value="true" />
+        <processorPath useClasspath="false">
+          <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/info.picocli/picocli-codegen/4.5.1/cafc543418928d5b76370409348d40a76d9c5b6a/picocli-codegen-4.5.1.jar" />
+          <entry name="$USER_HOME$/.gradle/caches/modules-2/files-2.1/info.picocli/picocli/4.5.1/ff913b736b6ffc7cdf8144401b02a501554ba0c5/picocli-4.5.1.jar" />
+        </processorPath>
+        <module name="Terasology.facades.PC.main" />
+      </profile>
```
keturn, line 15: This doesn't seem good. Are we going to have to gitignore compiler.xml?

probably not much loss there but it does set bytecodeTargetLevel

keturn:
I ended up moving everything currently in Terasology (the class with the static main method) to a new class (TerasologyCommand), but it looks like I didn't have to do that. I could I have left that main as static but switched everything else to non-static.

Those aren't two useful classes, it just moved stuff around and left a one-line static method in the other class, so I think I'll shift it back so it's a smaller change overall.

[that's a recap from discord, adding it here as a comment as I haven't got to actually do it yet.]

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -91,390 +30,20 @@
...
-    }
-
-    private static int getLastNumber(String str) {
-        int positionOfLastDigit = getPositionOfLastDigit(str);
-        if (positionOfLastDigit == str.length()) {
-            // string does not end in digits
-            return -1;
-        }
-        return Integer.parseInt(str.substring(positionOfLastDigit));
+        new CommandLine(new TerasologyCommand()).execute(args);
```
skaldarnar, line 46: 👍 I like ^^

facades/PC/src/main/java/org/terasology/engine/TerasologyCommand.java
```diff
@@ -0,0 +1,417 @@
...
+        if (path == null) {
+            path = Paths.get("logs");
+        }
+
+        LoggingContext.initialize(path);
+    }
+
+
+    private static void printUsageAndExit() {
+        // TODO: Add all this stuff back in to the help text.
```
skaldarnar, line 272: This is then added via the picocli library, isn't it? I don't think we need to copy word-by-word, but just make sure that users can figure out what options/flags are available.

facades/PC/src/main/java/org/terasology/engine/TerasologyCommand.java
```diff
@@ -0,0 +1,417 @@
...
+ * </table>
+ * <br><br>
+ * When used via command line an usage help and some examples can be obtained via:
+ * <br><br>
+ * terasology -help    or    terasology /?
+ * <br><br>
+ */
+
+@CommandLine.Command(name = "terasology")
+public class TerasologyCommand implements Callable<Integer> {
```
skaldarnar, line 90: Does the `Callabale` move this to a different thread? Will this cause even more issues with porting to LWJGL 3 and MacOS (see #3969 )?

keturn: I don't *think* so. I think picocli just uses Callable<Integer> as an interface that's easy to define for use with its `.execute(it)` method, so it can run it, handle exceptions, and get a process return code at the end. But it'll be a good thing to double-check.

keturn:
It's working in IntelliJ IDEA but the build-with-command-line-gradle fails. :disappointed: 

```
:facades:PC:game FAILED
Error: Unable to initialize main class org.terasology.engine.Terasology
Caused by: java.lang.NoClassDefFoundError: picocli/CommandLine$Model$ArgSpec
```

changing the picocli dependency type to `implementation` or `compileOnly` doesn't help.

Not sure what the deal is. Is it related to the picolci-codegen or annotation processing or something?

keturn:
It's something in the difference between the `game` and `run` tasks in :facades:PC. `run` is working better.

facades/PC/src/main/java/org/terasology/engine/TerasologyCommand.java
keturn: I confirmed this. `currentThread()` returns the same value inside the `call` method as it does in the static `main`.

.idea/compiler.xml
keturn: added to gitignore and filed https://youtrack.jetbrains.com/issue/IDEA-251986

keturn:
Oh heck. I just realized that if we change the way command-line options are spelled, Launcher needs to know.

We could add aliases for the old spellings. It wouldn't be difficult.

Thing is, I like the idea of following the POSIX conventions for these, and in that convention prefixing options with a single dash means you're using the short-form single-letter options. That's a potential source of confusion I'd like to avoid.

I'd prefer to make this explicitly-breaking-change instead, get it over with.

But then what does Launcher do?

Is it willing to break compatibility with previous releases just as quickly, or do we need to teach it to use old-style options if the version number is below a cutoff point?

pollend:
At the moment the whole engine is going through a large split and change so I wouldn't be too worried about breaking API changes if it makes things work better. I also don't want to have stuff be bagged down by older changes for the sake of consistency since I don't think we have any obligations to keep things consistent.

keturn:
The blocker to getting this merged is the `game` task: its classpath is missing something, something the `run` task _isn't_ missing.

And if we're working on the `game` task and its variants, it seems like a good time to address the code duplication going on between them.

For that, we could use some documentation on our requirements for those tasks.

* **Run Game** (start a process that brings up the game window etc etc) is the obvious one.
* **Build all classes and assets for `engine` and `facade`** should be a dependency, I don't think there's any disagreement about that.

I think the part where questions come up is around modules. Currently `game` adds these dependencies:

```kotlin
    // Dependencies: natives + all modules & the PC facade itself (which will trigger the engine)
    dependsOn(":extractNatives")
    dependsOn(":moduleClasses")
    dependsOn("classes")
```

Where `:moduleClasses` is a task of the root project that depends on the `classes` tasks of everything under `/modules/*`.

And `game` has this to say about its classpath:

```kotlin
    // Classpath: PC itself, engine classes, engine dependencies.
    //     Not modules or natives since the engine finds those.
    classpath(sourceSets["main"].output.classesDirs)
    classpath(sourceSets["main"].output.resourcesDir)
    classpath(project(":engine").sourceSets["main"].output.classesDirs)
    classpath(project(":engine").sourceSets["main"].output.resourcesDir)
    classpath(project(":engine").configurations.runtimeClasspath)
```

from reading that, I expect:

* `game` depends on compiling the `.class` files of all `modules/*`
  * but not their `.jar`s
  * a compilation failure of _any_ module will be treated as a failure to provide a dependency of the task, and game will not launch
* the classpath does **not** include anything from `/modules` (neither `/modules` itself nor `/modules/**/build/**`)
* the classpath *does* include `*.class` files and resources from from `/engine/build`

[continued…]

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -91,53 +79,72 @@
...
+    @Option(names = "--permissive-security")
+    private boolean permissiveSecurity;
+
+    @Option(names = "--save-games", defaultValue = "true", negatable = true, description = "Enable new save games")
+    private boolean saveGamesEnabled;
+
+    @Option(names = "--server-port", description = "Change the server port")
+    private Integer serverPort;
+
+    @Option(names = "--config", description = "Override default config")
```
keturn, line 122: I changed this from `overrideDefaultConfig` to simply `config` because I believed it was overriding the location of the default configuration file, and I thought that a `config` pointing to a file was clear enough for that.

but now I think this maybe specifies an **additional** config file that contains overrides. Not something used **instead** of the default config. If that's true, it should get its "override" name back.

keturn:
Follow-up questions to [earlier](https://github.com/MovingBlocks/Terasology/pull/4157#issuecomment-704658753):

We habitually tell people to run `gradlew jar game`. The `game` task specifically depends on building the `.class` files, _not_ the `.jar`s. Does the fact that we tell people to build the jars mean:
* that the `game` task should really have a `jar` dependency?
* or that the `jar` was left out on purpose because gestalt module loader _should_ be able to work with the classes alone, but we've broken module loading somehow so it doesn't?

There's a category of module development workflow requirements that I'd like to have spelled out, not only for this but for any time we make changes to the gradle + IntelliJ build process.

Things like:
* after you make changes to the source, does it rebuild everything you expect it to? (and what are those things?)
* when is the reflections cache built and used, and when is it not?
* which things can be reloaded at runtime with gestalt? a module's assets? its classes? new classes?
* do the debugger's breakpoints work?
* does the IDE + debugger's class-reloading work?

DarkWeird:
Gradle jar game - it is old behaviour (before my changes to build.gradle) it is seems was related with build dir or reflections.cache. dont remember

Gestalt-module works fine with classes and jars(idk which priorities)


If you change source - then rebuilding only changef module and it's dependents.
If we make api module for apis then recompile only implement module(gradle 5+ used in android builds mostly)

Reflections - builds at compile time and used by gestalt-module. If reflections.cache not found - then it gathering at runtime.
Current behavior: reflection gather before jar and cleanup after jar.. then reflections on source modules gathering at runtime.

Better: gather reflection when classes(or compileJava) task done and uptodate it when classes uptodate.


Reloading:
Only few assets type possible to reload(see code) (we have special method invoke on tick somewhere) (gestaltv7 have special manager with reloading)
You cannot reload classes during debugging only method's body changes (see jvm debug reloading limitation) (but you can use some thing like DCEVM or JRebel)

Debugging with gradle works fine.

keturn:
## classes

Do we keep the the `modules/**/build/**` directories off the runtime classpath only because it makes things tidier (it would make the classpath a _lot_ longer), or also because including them there would break module loading or the security model or something?

## reloading

On asset reloading: There's this [gestalt demo video](https://youtu.be/Abzi0WdC9oM) showing changing textures. Is it sufficient to just write to their source locations, without having `jar` or `syncAssets` tasks rerun?

I went looking for docs on asset reloading. I found a brief note about it here: https://github.com/MovingBlocks/gestalt/wiki/Gestalt-Asset-Core-Quick-Start#reload-assets-changed-on-disk
though I'm having trouble finding a class with that name used by Terasology. I think that might be its gestalt v7 name.

the invocation to reload on tick is in [TerasologyEngine.tick](https://github.com/MovingBlocks/Terasology/blob/0d50ea5cc559d44b29ff19f890d11c9645076ca4/engine/src/main/java/org/terasology/engine/TerasologyEngine.java#L457)

## reflections

> Reflections - builds at compile time and used by gestalt-module. If reflections.cache not found - then it gathering at runtime.
> Current behavior: reflection gather before jar and cleanup after jar.. then reflections on source modules gathering at runtime.
> Better: gather reflection when classes(or compileJava) task done and uptodate it when classes uptodate.

That “better” option is one of those things where I expect we could do it for just gradle, but I'm worried about getting consistent behavior when mixed with IntelliJ IDEA's compilation process.

# Scope

A lot of these notes are only very loosely related to this picocli PR, but they are related in that we're specifying the things that need to keep working when we tinker with these JavaExec tasks and their classpaths.

I'll want to organize this information and put it someplace. A page on MovingBlocks/Terasology/wiki I guess?

DarkWeird:
## Classes
We cannot use gradles build dir definition (build/classes/main) and not copy resources in classes dir  because gestalt-module/gestalt-assets use only one dir from source sets.

## Reflections

I hope we can drop reflections and use Annotation Processing, when gestalt DI will done. Then idea and gradle processes will be consistent

## Scope

Terasology - everytime code research :D

Yeah, plz do it!

keturn:
> We cannot use gradles build dir definition (build/classes/main) and not copy resources in classes dir because gestalt-module/gestalt-assets use only one dir from source sets.

then how did that asset-reloading demo work? was the editor saving to the classes directory? or was there a auto-refreshing copy-resources-to-classes task running in the background?

DarkWeird:
Copying resources in classes build directory.

Gestalt-module can load only one directory as source module.

Normal gradle build directories include separate direcories for resources and source sets.

Cervator:
I still haven't had time to dig into this, although I did just bump into the `run` vs `game` tasks, wondering why we still had two - thought one was renamed into the other years ago. Sooo ... I took it out (#4216). Looked like `game` was more complete anyway. Are we still missing something or is that unique to this rework?

`gradlew jar game` is a mix of habit and speed (just three characters woo!). I don't know if it has ever been required (although reflections being involved in the past rings a bell - but only coincidentally). It really should just be `gradlew game` but ... gremlins and superstitions

keturn:
>  still haven't had time to dig into this, although I did just bump into the run vs game tasks, wondering why we still had two - thought one was renamed into the other years ago.

Ah, that might well be true about the renaming. I (re)introduced the `run` task while I was doing the big gradle upgrade (for 6.4 & .idea, probably?). That was one on the areas where I started down that path of these changes to facades/PC/build.gradle and then realized it was biting off more than I could chew in that PR.

> Looked like game was more complete anyway.

In what sense? It might have been overlooked among all the comments posted to this issue, but note [here](https://github.com/MovingBlocks/Terasology/pull/4157#issuecomment-701709067) that the question that started all that discussion is that the `run` tasks _works_ after I add the picocli dependency and the `game` task doesn't.

If you're able to describe what `game` provides that is lacking from `run`, that would make the specification I was looking for to be able to work on these tasks without breaking someone's workflow.

> It really should just be gradlew game but ... gremlins and superstitions

I am against keeping gremlins in source code. You can make a module for gremlins in-game if you want. 😉

As for superstitions, please document them here or in their own issues as appropriate. Then we can do some mythbusters on 'em. :no_entry_sign: :ghost:

keturn:
Hooray! With #4238 fixed, this is no longer blocked. That is, there's still a bit of work to do (see "Outstanding before merging"), but it's no longer blocked on classpath errors.

jdrueckert:
Listed outstanding issues are:
- What do we do about Launcher? MovingBlocks/TerasologyLauncher#589
- re-add extended help message with examples
- update https://github.com/MovingBlocks/Terasology/blob/develop/docs/Playing.md

Questions: 
1. Do we need to do something about the Launcher at this point? Is this PR in any way breaking something for the launcher? If not, it should not block this and can be done as a follow-up
2. Re-add sounds like there were help messages with examples before, but I don't see you removing any 🤔 where are they or where have they gone?
3. Does updating the docs only affect for instance `--` instead of `-` or is there any explanatory text that needs to be added or updated?

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -196,153 +219,121 @@ private static void setupLogging() {
...
-            } else if (arg.startsWith(OVERRIDE_DEFAULT_CONFIG)) {
-                System.setProperty(Config.PROPERTY_OVERRIDE_DEFAULT_CONFIG, arg.substring(OVERRIDE_DEFAULT_CONFIG.length()));
-            } else {
-                recognized = false;
+    private void handleLaunchArguments() throws IOException {
+        if (homeDir != null) {
+            logger.info("homeDir is {}", homeDir);
+            if (homeDir.toString().startsWith("-")) {
+                // picocli shouldn't let this happen, but see
+                // https://github.com/remkop/picocli/issues/1197
```
skaldarnar, line 306: This has been fixed and is available in picocli v4.5.2 - do we still need this workaround?

skaldarnar:
> Do we need to do something about the Launcher at this point? Is this PR in any way breaking something for the launcher? If not, it should not block this and can be done as a follow-up

Yes, we won't be able to launch releases containing these changes from the launcher (we make use of the `--homedir` flag, for instance). The required fix is similar to https://github.com/MovingBlocks/TerasologyLauncher/commit/2a85337ba8c38f6f768ac2f0772649f3af5efa57 as it depends on the release version which variant of flags we need.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
keturn: Correct, we can probably do without it now.

There are some edge cases involving what happens when you spell flags wrong, but going with the default behavior for that is a reasonable choice.

keturn:
> 2. Re-add sounds like there were help messages with examples before, but I don't see you removing any 🤔 where are they or where have they gone?

This stuff https://github.com/MovingBlocks/Terasology/blob/3c5845ec103c15454d441a60bfe155c59335bfc5/facades/PC/src/main/java/org/terasology/engine/Terasology.java#L278-L295

> 3. Does updating the docs only affect for instance `--` instead of `-` or is there any explanatory text that needs to be added or updated?

Looks like it's only the spelling of the options. I think how it's showing their use there is still the same.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
keturn: this has been removed.

immortius:
> > We cannot use gradles build dir definition (build/classes/main) and not copy resources in classes dir because gestalt-module/gestalt-assets use only one dir from source sets.
> 
> then how did that asset-reloading demo work? was the editor saving to the classes directory? or was there a auto-refreshing copy-resources-to-classes task running in the background? maybe it's worth pinging @immortius for info about that demo.

I was editing the assets within the built location (so in the directory module) - not in the source location.

I would note that latest Gestalt has the technical underpinnings to support modules composed of multiple directories, although the module scanner doesn't create them.

keturn:
> Yes, we won't be able to launch releases containing these changes from the launcher […] Doing this change should be fairly simple, I can prepare that to have it ready +1

@skaldarnar, have you started a Launcher branch for this somewhere?

keturn:
I've tried to collect the notes we made here about what we expect from build results and classpaths and whatnot on https://github.com/MovingBlocks/Terasology/wiki/Supported-Development-Workflows

keturn:
That was a potentially messy merge for PC/build.gradle.kts but I think it ended up okay :crossed_fingers:

skaldarnar:
I did not forget about this ... I swear! Made a note to update the launcher sometime in the not-so-far future.

keturn:
There's been a new picocli release in the meantime.

One thing that might be of interest is picocli using [`Optional`](https://github.com/remkop/picocli/releases/tag/v4.6.0#4.6.0-java-util-optional).

.idea/misc.xml
```diff
@@ -1,15 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="EntryPointsManager">
-    <list size="5">
-      <item index="0" class="java.lang.String" itemvalue="org.terasology.entitySystem.event.ReceiveEvent" />
-      <item index="1" class="java.lang.String" itemvalue="org.terasology.entitySystem.systems.RegisterSystem" />
```
skaldarnar: Why is `RegisterSystem` not needed (anymore)? Is the class considered used in any case?

keturn: I was re-checking these at some point and removing the class-level annotation didn't result in any inspection warning popping up. 🤷

keturn:
Launcher supports this now, but I added a new bug in the process. So this is waiting on https://github.com/MovingBlocks/TerasologyLauncher/issues/651 in order for Launcher to be releasable.

PR #4161, opened 2020-09-30
feat(JOML): migrate target system

pollend:
this is a migration of the target system since JOMLUtil is not needed any more in this case.

### DIFF
--- a/engine/src/main/java/org/terasology/input/cameraTarget/CameraTargetSystem.java
+++ b/engine/src/main/java/org/terasology/input/cameraTarget/CameraTargetSystem.java
@@ -17,14 +17,15 @@
 package org.terasology.input.cameraTarget;
 
 import com.google.common.base.Objects;
+import org.joml.RoundingMode;
+import org.joml.Vector3f;
+import org.joml.Vector3i;
 import org.terasology.config.Config;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.logic.players.LocalPlayer;
 import org.terasology.math.JomlUtil;
 import org.terasology.math.TeraMath;
-import org.terasology.math.geom.Vector3f;
-import org.terasology.math.geom.Vector3i;
 import org.terasology.physics.CollisionGroup;
 import org.terasology.physics.HitResult;
 import org.terasology.physics.Physics;
@@ -34,7 +35,6 @@ import org.terasology.rendering.world.WorldRenderer;
 import org.terasology.world.BlockEntityRegistry;
 import org.terasology.world.block.BlockComponent;
 
-import java.math.RoundingMode;
 import java.util.Arrays;
 
 /**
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/terasology/input/internal/BindableButtonImpl.java
```diff
@@ -158,7 +159,10 @@ public boolean updateBindState(Input input,
                 }
                 if (!keyConsumed) {
                     buttonEvent.prepare(id, ButtonState.DOWN, delta);
-                    buttonEvent.setTargetInfo(target, targetBlockPos, hitPosition, hitNormal);
+                    buttonEvent.setTargetInfo(target,
+                        targetBlockPos == null ? null : JomlUtil.from(targetBlockPos),
+                        hitPosition == null ? null : JomlUtil.from(hitPosition),
+                        hitNormal == null ? null : JomlUtil.from(hitNormal));
```
jdrueckert, line 165: Would it make sense to have `JomlUtil.from(null)` return `null` instead?

engine/src/main/java/org/terasology/input/cameraTarget/CameraTargetSystem.java
```diff
@@ -144,7 +144,7 @@ private void updateFocalDistance(HitResult hitInfo, float delta) {
         if (hitInfo.isHit()) {
             Vector3f playerToTargetRay = new Vector3f();
             //calculate the distance from the player to the hit point
-            playerToTargetRay.sub(JomlUtil.from(hitInfo.getHitPoint()), localPlayer.getViewPosition());
+            hitInfo.getHitPoint().sub(JomlUtil.from(localPlayer.getViewPosition()), playerToTargetRay);
```
jdrueckert, line 147: I guess/hope that "rotation" is on purpose?
So TeraMath's `x.sub(y, z)` equals joml's `y.sub(z, x)`?

engine/src/main/java/org/terasology/input/internal/BindableButtonImpl.java
```diff
@@ -176,7 +180,10 @@ public boolean updateBindState(Input input,
                 }
                 if (!keyConsumed) {
                     buttonEvent.prepare(id, ButtonState.UP, delta);
-                    buttonEvent.setTargetInfo(target, targetBlockPos, hitPosition, hitNormal);
+                    buttonEvent.setTargetInfo(target,
+                        targetBlockPos == null ? null : JomlUtil.from(targetBlockPos),
+                        hitPosition == null ? null : JomlUtil.from(hitPosition),
+                        hitNormal == null ? null : JomlUtil.from(hitNormal));
```
jdrueckert, line 186: see above

engine/src/main/java/org/terasology/input/internal/BindableButtonImpl.java
```diff
@@ -205,7 +212,10 @@ public void update(EntityRef[] inputEntities,
                 boolean consumed = triggerOnRepeat(delta, target);
                 if (!consumed) {
                     buttonEvent.prepare(id, ButtonState.REPEAT, delta);
-                    buttonEvent.setTargetInfo(target, targetBlockPos, hitPosition, hitNormal);
+                    buttonEvent.setTargetInfo(target,
+                        targetBlockPos == null ? null : JomlUtil.from(targetBlockPos),
+                        hitPosition == null ? null : JomlUtil.from(hitPosition),
+                        hitNormal == null ? null : JomlUtil.from(hitNormal));
```
jdrueckert, line 218: see above

jdrueckert:
oh and this PR is not up-to-date with current develop

engine/src/main/java/org/terasology/input/cameraTarget/CameraTargetSystem.java
pollend: for termath x.sub(y,z) --> x = y - z where joml it's  x-y = z. z is the dest in this case.

jdrueckert:
@pollend please add null check inside jomlutil function

jdrueckert (approved):

PR #4166, opened 2020-10-01
feat: remove Has/Cancel/Add DelayedActionEvent

pollend:
These action events are not used anywhere in omega. seems like a good time to remove them from the engine. There might be some older modules using this but, at this point its probably just better to rework/rebuild them at this point.


Cervator (approved):
Can confirm, looks fine to me. Reminded me of some funny naming issues elsewhere but that's not relevant to merging this or not.

PR #4168, opened 2020-10-01
feat(JOML): add orientation for Rotation

pollend:
this is a squash from the change in: https://github.com/MovingBlocks/Terasology/pull/4124

This is a simple change for the orientation of the Rotation. This is just an api change that should help with migrating changes for JOML. this adds an orientation method that returns the Quaternionf instead of Quat4f (termath) . there are no direct impacts at the moment and should work in omega since I only deprecate the old method. I also updated the test case to test for consistency.


Cervator (approved):
Seems to test out fine with some minor poking around in CoreSampleGameplay and GooeysQuests 👍

PR #4172, opened 2020-10-01
feat(JOML): migrate action event

pollend:
this migrates ActivateEvent/ActivationRequest/ActivationPrediction along with changes to get the game to run.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/events/ActivationPredicted.java
+++ b/engine/src/main/java/org/terasology/logic/characters/events/ActivationPredicted.java
@@ -15,10 +15,10 @@
  */
 package org.terasology.logic.characters.events;
 
+import org.joml.Vector3f;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.event.AbstractConsumableEvent;
 import org.terasology.logic.location.LocationComponent;
-import org.terasology.math.geom.Vector3f;
 
 /**
  */
@@ -76,16 +76,24 @@ public class ActivationPredicted extends AbstractConsumableEvent {
 
     public Vector3f getTargetLocation() {
         LocationComponent loc = target.getComponent(LocationComponent.class);
-        if (loc != null&& !Float.isNaN(loc.getWorldPosition().x)) {
-            return loc.getWorldPosition();
+        if (loc != null) {
+            Vector3f result = loc.getWorldPosition(new Vector3f());
+            if (result.isFinite()) {
+                return result;
+            }
         }
         return null;
     }
 
     public Vector3f getInstigatorLocation() {
         LocationComponent loc = instigator.getComponent(LocationComponent.class);
-        if (loc != null&& !Float.isNaN(loc.getWorldPosition().x)) {
-            return loc.getWorldPosition();
+        if (loc != null) {
+            Vector3f result = loc.getWorldPosition(new Vector3f());
+            if (result.isFinite()) {
+                return result;
+            }
+            result.set(0, 0, 0);
+            return result;
         }
         return new Vector3f();
     }
--- a/engine/src/main/java/org/terasology/logic/common/ActivateEvent.java
+++ b/engine/src/main/java/org/terasology/logic/common/ActivateEvent.java
@@ -15,11 +15,11 @@
  */
 package org.terasology.logic.common;
 
+import org.joml.Vector3f;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.event.AbstractConsumableEvent;
 import org.terasology.logic.characters.events.ActivationRequest;
 import org.terasology.logic.location.LocationComponent;
-import org.terasology.math.geom.Vector3f;
 
 /**
  */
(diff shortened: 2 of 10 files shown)


engine/src/main/java/org/terasology/logic/common/ActivateEvent.java
```diff
@@ -84,16 +84,24 @@ public int getActivationId() {
 
     public Vector3f getTargetLocation() {
         LocationComponent loc = target.getComponent(LocationComponent.class);
-        if (loc != null&& !Float.isNaN(loc.getWorldPosition().x)) {
```
jdrueckert, line 87: Do we know why this originally only checked x for being finite?

engine/src/main/java/org/terasology/logic/common/ActivateEvent.java
```diff
@@ -84,16 +84,24 @@ public int getActivationId() {
...
+            if (result.isFinite()) {
+                return result;
+            }
         }
         return null;
     }
 
     public Vector3f getInstigatorLocation() {
         LocationComponent loc = instigator.getComponent(LocationComponent.class);
-        if (loc != null&& !Float.isNaN(loc.getWorldPosition().x)) {
```
jdrueckert, line 95: see above

engine/src/main/java/org/terasology/logic/common/ActivateEvent.java
```diff
@@ -84,16 +84,24 @@ public int getActivationId() {
...
-            return loc.getWorldPosition();
+        if (loc != null) {
+            Vector3f result = loc.getWorldPosition(new Vector3f());
+            if (result.isFinite()) {
+                return result;
+            }
+            result.set(0, 0, 0);
+            return result;
```
jdrueckert, line 104: Why do we return [0,0,0] vector here but `null` above in `getTargetLocation()`?

engine/src/main/java/org/terasology/logic/characters/events/ActivationPredicted.java
```diff
@@ -76,16 +76,24 @@ public int getActivationId() {
 
     public Vector3f getTargetLocation() {
         LocationComponent loc = target.getComponent(LocationComponent.class);
-        if (loc != null&& !Float.isNaN(loc.getWorldPosition().x)) {
```
jdrueckert, line 79: Do we know why this originally only checked x for being finite?

engine/src/main/java/org/terasology/logic/characters/events/ActivationPredicted.java
```diff
@@ -76,16 +76,24 @@ public int getActivationId() {
...
+            if (result.isFinite()) {
+                return result;
+            }
         }
         return null;
     }
 
     public Vector3f getInstigatorLocation() {
         LocationComponent loc = instigator.getComponent(LocationComponent.class);
-        if (loc != null&& !Float.isNaN(loc.getWorldPosition().x)) {
```
jdrueckert, line 87: see above

engine/src/main/java/org/terasology/logic/characters/events/ActivationPredicted.java
```diff
@@ -76,16 +76,24 @@ public int getActivationId() {
...
-            return loc.getWorldPosition();
+        if (loc != null) {
+            Vector3f result = loc.getWorldPosition(new Vector3f());
+            if (result.isFinite()) {
+                return result;
+            }
+            result.set(0, 0, 0);
```
jdrueckert, line 95: Why do we return [0,0,0] vector here but `null` above in `getTargetLocation()`?

engine/src/main/java/org/terasology/logic/common/ActivateEvent.java
pollend: An assumption that if x is nan then the whole vector is nan. Method from joml does not use this assumption.

engine/src/main/java/org/terasology/logic/common/ActivateEvent.java
pollend: I have no idea why we do this. It's kind of strange. Don't want to think about it too much. We can look at it in another pass.

jdrueckert: Okay, different question: Do we need this at all if we have the `return new Vector3f()` at the end?
```
            result.set(0, 0, 0);
            return result;
```

jdrueckert: Yeah I see that, but isn't a cleared vector equal to an empty vector?
So the code path leads to the return of the empty vector (`return new Vector3f()` in l.106) anyway, so there's no need to explicitly clear and return at that point, is there?

pollend: I'm assuming there is some math that will add it or something and a nan vector will throw a wrench into things. Better to clear it. Basically the same as passing out a new vector.

jdrueckert: I see okay. Fair point.

jdrueckert (approved):

PR #4174, opened 2020-10-02
add type handlers for Vector2/3/4fc and Vector2/3/4ic TypeHandlers and basic test cases for Event 

pollend:
here are some additional test cases for event serialization. I have it setup for Vector3/2/4fc at the moment but more test cases should be setup. just to verify something when I attempt to convert some of the events over from BaseVector to Vector*fc.


eviltak (approved):
Looks good to me!

PR #4175, opened 2020-10-02
chore[build]: convert facades/PC to gradle to kts

keturn:
This converts the gradle config from groovy to kotlin.

This PR is intended to be a near line-for-line port. Hold any further changes for another branch that won't be so cluttered by the conversion.

blocking #4157, as it needs to make changes to these JavaExec tasks.

### How to test

Run `gradlew :facades:PC:tasks` for the list of tasks.

* test  the **Terasology run tasks**
  * See [Playing#Configuring a Server](docs/Playing.md#configuring-a-server) for information about how to configure the server settings from the gradle command line.
* test the **Terasology dist tasks**
  * dists should include jar files for all the modules you have sources checked out for in `/modules/`


facades/PC/build.gradle.kts
```diff
@@ -30,396 +28,370 @@ dateTimeFormat.timeZone = TimeZone.getTimeZone("UTC")
...
+val versionFileName = "VERSION"
+val versionBase by lazy { File(templatesDir, "version.txt").readText().trim() }
+val displayVersion = versionBase
 
-    // Version related
-    startDateTimeString = dateTimeFormat.format(new Date())
-    versionFileName = 'VERSION'
-    versionBase = new File(templatesDir, "version.txt").text.trim()
-    displayVersion = versionBase
-}
```
keturn: This isn't strictly line-for-line compatible because I only kept _some_ of these as gradle "extra" properties. Most of them I left as local variables.

I don't believe it's a problem for the normal build environment, since there's nothing else in the build that interacts with them, but maybe it could be a concern if something like Jenkins has an alternate configuration that sets or uses them differently?

keturn:
Reviewing by file on GitHub tells you the `.gradle.kts` is an entirely new file, but if you review by commit it'll actually show a diff.

keturn:
marking as Draft while I fix a couple things with the distribution tasks

facades/PC/build.gradle.kts
```diff
@@ -30,396 +28,370 @@ dateTimeFormat.timeZone = TimeZone.getTimeZone("UTC")
...
 
 // Jenkins-Artifactory integration catches on to this as part of the Maven-type descriptor
-group = 'org.terasology.facades'
+group = "org.terasology.facades"
 
 dependencies {
-    implementation project(':engine')
-    implementation group: 'org.reflections', name: 'reflections', version: '0.9.10'
+    implementation(project(":engine"))
+    implementation(group = "org.reflections", name = "reflections", version = "0.9.10")
```
skaldarnar, line 86: On a side note: https://github.com/ronmamo/reflections is available in 0.9.12 now.

facades/PC/build.gradle.kts
```diff
@@ -30,396 +28,370 @@ dateTimeFormat.timeZone = TimeZone.getTimeZone("UTC")
...
+tasks.named<Jar>("jar") {
     manifest {
         //TODO: Maybe later add the engine's version number into here?
-        def manifestClasspath = configurations.runtimeClasspath.collect { it.getName() }.join(" ")
-        attributes("Main-Class": mainClassName)
-        attributes("Class-Path" : manifestClasspath)
-        attributes("Implementation-Title": "Terasology-" + project.name)
-        attributes("Implementation-Version": env.BUILD_NUMBER + ", " + env.GIT_BRANCH + ", " + env.BUILD_ID)
+        attributes["Main-Class"] = mainClassName
+        attributes["Class-Path"] = configurations["runtimeClasspath"].map { it.name }.joinToString(" ")
```
skaldarnar, line 97: 👍 I like how this can be inlined now 😃

keturn:
I confirmed that the `server` task writes a config.cfg with the values from `-PextraModules` and such

and that the dist tasks make their zip files including module jars.

there might be some things involving version files or various `BUILD_*` environment variables that didn't get exercised in my development environment here.

facades/PC/build.gradle.kts
```diff
@@ -30,396 +28,370 @@ dateTimeFormat.timeZone = TimeZone.getTimeZone("UTC")
...
 
-    args "-homedir", "-noCrashReport"
-    jvmArgs "-Xmx1536m"
-    workingDir rootDir
+    args = listOf("-homedir", "-noCrashReport")
+    jvmArgs = listOf("-Xmx1536m")
+    workingDir = rootDir
 
-    dependsOn rootProject.moduleClasses
+    dependsOn(":moduleClasses")
```
skaldarnar, line 118: Hm, so everything else seems to become more (type-)safe, while you reference the dependency via a plain string there? Is there another way to do this in the Kotlin DSL?

facades/PC/build.gradle.kts
```diff
@@ -30,396 +28,370 @@ dateTimeFormat.timeZone = TimeZone.getTimeZone("UTC")
...
+    dependsOn("classes")
 
     // Run arguments
     main = mainClassName
     workingDir = rootDir
-    args "-homedir"
-    jvmArgs ["-Xmx1536m"]
+    args = listOf("-homedir")
+    jvmArgs = listOf("-Xmx1536m")
```
skaldarnar, line 139: Would be nice to make this adjustable in the future via some kind of command line argument...

facades/PC/build.gradle.kts
```diff
@@ -30,396 +28,370 @@ dateTimeFormat.timeZone = TimeZone.getTimeZone("UTC")
...
+configure<IdeaModel> {
     module {
         // Change around the output a bit
         inheritOutputDirs = false
-        outputDir = file('build/classes')
-        testOutputDir = file('build/testClasses')
+        outputDir = file("build/classes")
+        testOutputDir = file("build/testClasses")
     }
 }
```
skaldarnar: Do we still need this with the new idea setup introduced recently?

facades/PC/server.build.gradle
```diff
@@ -0,0 +1,43 @@
+// Copyright 2020 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+// This is still in Groovy because the embedded Groovy has a JsonBuilder.
+import groovy.json.JsonBuilder
+
+
+// By delaying this task to doLast (the << bit) we don't get the headless server dir set up unless actually wanting it
```
skaldarnar, line 8: Is this comment still up to date for the code? There is no notion of `<<` or `doLast` in the code (neither before nor after the refactoring)...

skaldarnar (commented):
Looked over the changes and can confirm this is more or less a direct translation from Groovy to Kotlin DSL. Left a few comments, so please have a look.

I did not test the changes locally, though. Still setting up my system after hardware upgrade...

facades/PC/build.gradle.kts
keturn: It has been for a while, but there are some [widely reported issues](https://github.com/ronmamo/reflections/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) with 0.9.12 (and some for 0.9.11) so I've been afraid to mess with it.

facades/PC/build.gradle.kts
keturn: There's some tension going on in gradle-config-land. In one direction, you have Kotlin making things more type-safe. In the other direction, Gradle has this configuration-avoidance initiative to do lazy evaluation for certain types of things so it doesn't have to figure out the _entire_ multi-project configuration tree before it's able to run a single task. That's _discouraged_ direct links between sub-projects in many ways.

on second thought, I'm not sure how much the configuration-avoidance API has to do with this specific example.

from Kotlin's perspective, these are instances of Project. Project, as a Class, doesn't have a property named `moduleClasses`. build.gradle.kts files _configure_ projects but they do not define _new classes_ for each of them. Things like `rootProject.moduleClasses` only work with dynamic run-time attribute-lookup trickery.

We certainly could say something along the lines of `rootProject.tasks.get("moduleClasses")` but that's not any different than what we have here in terms of providing assurance that "moduleClasses" is a safe reference.

...and also the Gradle API is really big and provides a bunch of ways to get references to things and I certainly haven't learned everything about what's supposed to be the current best practice.

keturn: Long-term, the “safe” way involves not knowing about the _tasks_ of other projects at all. You declare a dependency like "I depend on things from `"org.terasology.engine:engine"` and need assets of type `<Jar>` that conform to `bytecode_level=JAVA_1_8`" or something like that, the dependency-resolver does its thing and checks to see if that project exports anything that matches those criteria.

facades/PC/build.gradle.kts
keturn: no but no matter how many times people have asked me that none of them have ever removed it :wink:

facades/PC/server.build.gradle
keturn: I agree with your reading here.

I don't know whether that means it's just an obsolete comment that should be removed, or if there was a regression and the code lost some desired functionality.

facades/PC/build.gradle.kts
DarkWeird: It can affect gestalt source module loading when you compile and run it via idea.
(And it compability with gradle build, it use same output dir, instead standart build/<sourceSet or resource>/classes)

keturn: facades/PC/ doesn't have modules in it, so for the stuff that applies only to this sub-project we don't have to worry about the gestalt loader.

facades/PC/server.build.gradle
skaldarnar: At this point we don't know (anymore), so let's remove this.
```suggestion
```

skaldarnar (approved):
I cannot start the server, but starting the game and building the distribution works fine! :+1:

PR #4179, opened 2020-10-04
tweak sort for validating chunk n^2 cost

pollend:
this sort for chunk relevance adds an n^2 cost to the search. since it will try to sort the current chunks against all possible nearby chunks. this function only add the chunk from the region if its valid or creates a new chunk if its not available. This same problem is also apparent with  `ChunkTaskRelevanceComparator`. Some better strategies would need to be setup if there are a lot of relevant entities. 

```
 Chunk chunk = chunkProvider.getChunk(pos);
                            if (chunk != null) {
                                region.checkIfChunkIsRelevant(chunk);
                            } else {
                                chunkProvider.createOrLoadChunk(pos);
                            }
```

### DIFF
--- a/engine/src/main/java/org/terasology/world/chunks/internal/ChunkRelevanceRegion.java
+++ b/engine/src/main/java/org/terasology/world/chunks/internal/ChunkRelevanceRegion.java
@@ -58,7 +58,7 @@ public class ChunkRelevanceRegion {
     }
 
     public Vector3i getCenter() {
-        return new Vector3i(center);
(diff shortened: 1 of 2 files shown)


pollend:
It should be lower then n^2. I exit the for loop early if the distance is at 0. Probably want to think about some better strategies to managing these queries.

DarkWeird (approved):
Yeah, it needs to rethink for algorithm. 
It is enough now, when a bit  regions used

PR #4186, opened 2020-10-07
feat(joml): prepare `BlockComponent` for migration

jdrueckert:


pollend:
Is the plan to remove the get and set methods afterwards but other then that looks good to me.

jdrueckert:
@pollend once everything that depends on / uses `BlockComponent` is migrated to joml, the plan would be to remove the deprecated ones. We can also discuss removing all of that altogether as suggested in the deprecation comment from 2018.
For the switch to joml I found the getter/setter rather useful tbh. Now it doesn't break everything at once :sweat_smile:

pollend (approved):

PR #4188, opened 2020-10-09
chore(joml): remove `dest` param from `get*` methods

jdrueckert:
- return constant, e.g. `Vector3fc` instead of `Vector3f`
- method naming will require refactoring after removal of deprecated methods but the behaviour on caller side can stay the same

### DIFF
--- a/engine/src/main/java/org/terasology/logic/location/LocationComponent.java
+++ b/engine/src/main/java/org/terasology/logic/location/LocationComponent.java
@@ -19,6 +19,7 @@ import com.google.common.collect.Lists;
 import org.joml.Quaternionf;
 import org.joml.Quaternionfc;
 import org.joml.Vector3fc;
+import org.joml.Vector3i;
 import org.terasology.entitySystem.Component;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.math.Direction;
@@ -134,7 +135,7 @@ public final class LocationComponent implements Component, ReplicationCheck {
     /**
      * @return
      * @deprecated This is scheduled for removal in an upcoming version method will be replaced with JOML implementation
-     *     {@link #getLocalDirection(org.joml.Vector3f)}.
+     *     {@link #getLocalDirectionConst()}.
      */
     @Deprecated
     public Vector3f getLocalDirection() {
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/terasology/logic/location/LocationComponent.java
```diff
@@ -184,21 +184,17 @@ public Vector3f getWorldPosition() {
...
-    /**
-     * get the world position
-     *
-     * @param dest will hold the result
-     * @return dest
-     */
-    public org.joml.Vector3f getWorldPosition(org.joml.Vector3f dest) {
+    // TODO: rename to `getWorldPosition` after removal of deprecated method
+    public org.joml.Vector3fc getWorldPositionConst() {
+        org.joml.Vector3f dest = new org.joml.Vector3f();
```
pollend, line 197: same comment as above.

engine/src/main/java/org/terasology/logic/location/LocationComponent.java
```diff
@@ -237,9 +233,10 @@ public Vector3f getWorldDirection() {
...
     }
 
-
-    public org.joml.Vector3f getWorldDirection(org.joml.Vector3f dest) {
-        return dest.set(Direction.FORWARD.asVector3f()).rotate(JomlUtil.from(getWorldRotation()));
+    public org.joml.Vector3fc getWorldDirectionConst() {
+        org.joml.Vector3f dest = new org.joml.Vector3f();
+        dest.set(Direction.FORWARD.asVector3f()).rotate(JomlUtil.from(getWorldRotation()));
+        return dest;
     }
```
pollend, line 240: same comment above

engine/src/main/java/org/terasology/world/chunks/internal/ChunkImpl.java
```diff
@@ -113,8 +114,9 @@ public Vector3i getPosition() {
     }
 
     @Override
-    public org.joml.Vector3i getPosition(org.joml.Vector3i dest) {
-        return dest.set(chunkPos.x,chunkPos.y,chunkPos.z);
```
pollend, line 117: wondering if it makes sense to use the method as is like this and then optimize it in post. Vector3ic is preferred this dest is not necessary. the constant type could just be wrapped into another vector as follows below. dest does not add anything else since there is nothing else on the back-end beyond just returning the position. Might make sense to keep it like this and drop it at the end when this whole class is migrated over and deprecate the staging methods here. 

```
new Vector3i(block.getPosition());
new Vector3i().add(block.getPosition());
```

PR #4191, opened 2020-10-11
feat(JOML): joml migrate wavefront format

pollend:
This is a fairly simple migration for .obj wavefront object. There are a couple wavefront objects littered around in module space and in the core engine. The player hand is fairly easy to see and verify that it works correctly in the game.


DarkWeird (approved):
Player hand looks like usual.
All in CoreSampleGameplay looks like usual

PR #4199, opened 2020-10-13
fix(nui): `InputSettings:onClose` out-of-bounds crash

jdrueckert:
- previously if any of the `getBinds()` calls returned a non-null but empty list, the game would crash due to an out-of-bounds error

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings/InputSettingsScreen.java
+++ b/engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings/InputSettingsScreen.java
@@ -120,16 +120,20 @@ public class InputSettingsScreen extends CoreScreenLayer {
         }
 
         UISlider mouseSensitivity = new UISlider("mouseSensitivity");
-        mouseSensitivity.bindValue(BindHelper.bindBeanProperty("mouseSensitivity", inputDeviceConfiguration, Float.TYPE));
+        mouseSensitivity.bindValue(BindHelper.bindBeanProperty("mouseSensitivity", inputDeviceConfiguration,
+                Float.TYPE));
         mouseSensitivity.setIncrement(0.025f);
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/rendering/nui/layers/mainMenu/inputSettings/InputSettingsScreen.java
```diff
@@ -307,9 +317,18 @@ public void onClosed() {
...
+            bindsConf.getBinds(new SimpleUri("engine:tabbingUI")).stream().findFirst().ifPresent(input -> {
+                TabbingManager.tabForwardInput = input;
+            });
+            bindsConf.getBinds(new SimpleUri("engine:tabbingModifier")).stream().findFirst().ifPresent(input -> {
+                TabbingManager.tabBackInputModifier = input;
+            });
+            bindsConf.getBinds(new SimpleUri("engine:activate")).stream().findFirst().ifPresent(input -> {
+                TabbingManager.activateInput = input;
+            });
+        }
```
jdrueckert, line 331: That's the actual fix, the rest is simply formatting.
An alternative here would be to return null. I chose to do nothing instead as there might've been a valid configuration before.

skaldarnar: We talked about this offline - it's not clear to neither of us what the semantic operation here is. Is it "reset" to default values? What should happen if there is no value in the binding config?

As @jdrueckert said this will do nothing in case the binding list is empty. An alternative would be to explicitly set the input to `null`. I don't know who can make this call - @BenjaminAmos ?

```java
bindsConf.getBinds(new SimpleUri("engine:tabbingUI")).stream().findFirst().orElse(null)
```

skaldarnar (approved):

DarkWeird (approved):

PR #4204, opened 2020-10-18
feat(JOML): migrate ChunkMath#getChunkRegionAroundWorldPos

pollend:
I wrote tests cases for the termath version of ChunkMath#getChunkRegionAroundWorldPos verified it against this when I re-implemented. This is just a method migration so there is no direct impact on omega.


engine/src/main/java/org/terasology/math/ChunkMath.java
```diff
@@ -472,6 +481,20 @@ public static Region3i getChunkRegionAroundWorldPos(Vector3i pos, int extent) {
         return Region3i.createFromMinMax(minChunk, maxChunk);
     }
 
+    /**
+     * get chunks contained within a center point and extent
+     * @param pos the world position
+     * @param extent the extent
+     * @return chunk region
+     */
+    public static BlockRegion getChunkRegionAroundWorldPos(Vector3ic pos, int extent) {
```
DarkWeird, line 490: I did understand this as "get chunk region around block world position" at first. 
Confusing param description.
Chunk haven't not world position, isn't it?
```suggestion
public static BlockRegion getChunkRegionAroundChunk(Vector3ic chunkPos, int extent) {
```

pollend: that's not a chunkPos, its a global position with some added extent then a chunk pos is calculated from that. Its the chunks that intersect a block position with an extent.

DarkWeird: Ah, calcChunkPos. Not noticed

DarkWeird (approved):

PR #4210, opened 2020-10-20
fix: add fluid system collider for bullet

pollend:
@4Denthusiast can you verify if this is correct. The fluid is a separate colliding objects to check against. 
https://github.com/Terasology/Fluid/pull/21


engine/src/main/java/org/terasology/physics/engine/VoxelFluidWorldSystem.java
```diff
@@ -0,0 +1,157 @@
...
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+import static org.terasology.physics.bullet.BulletPhysics.AABB_SIZE;
+
+/**
+ * Manages voxel shape for fluid and updates collision state between Bullet and Terasology
+ */
+@RegisterSystem
+public class VoxelFluidWorldSystem extends BaseComponentSystem {
```
4Denthusiast, line 47: I've been trying to maintain a consistent distinction between "fluid", the substances registered with the FluidRegistry, which can fill containers and pipes, and "liquid", blocks which have the `liquid` attribute. This is liquid, not fluid.

engine/src/main/java/org/terasology/physics/engine/VoxelFluidWorldSystem.java
```diff
@@ -0,0 +1,157 @@
...
+     * update voxel info for the wrapper for the associated block id
+     *
+     * @param block the block
+     */
+    private void tryRegister(Block block) {
+        short id = block.getId();
+        if (!registered.contains(id)) {
+            if (block.isLiquid()) {
+                btCollisionShape shape = ((BulletCollisionShape) block.getCollisionShape()).underlyingShape;
+                btVoxelInfo info = new btVoxelInfo(false,
```
4Denthusiast, line 98: This first argument specifies whether the block should be raytracable. This has to be set to true, otherwise the liquid collider can't actually be used. I tested this PR in its current state with the liquid placement PR, and liquids can't actually be targeted with fluid containers.

engine/src/main/java/org/terasology/physics/engine/VoxelFluidWorldSystem.java
```diff
@@ -0,0 +1,157 @@
...
+
+            Matrix4f matrix4f = new Matrix4f();
+            btDefaultMotionState blockMotionState = new btDefaultMotionState(matrix4f);
+
+            blockConsInf = new btRigidBody.btRigidBodyConstructionInfo(0, blockMotionState, worldShape, new Vector3f());
+            rigidBody = new btRigidBody(blockConsInf);
+            rigidBody.setCollisionFlags(btCollisionObject.CollisionFlags.CF_STATIC_OBJECT | rigidBody.getCollisionFlags()); // voxel world is added to static collision flag
+            short mask = (short) StandardCollisionGroup.LIQUID.getFlag(); // interacts with anything but static and
+            // liquid
+            discreteDynamicsWorld.addRigidBody(rigidBody, physics.combineGroups(StandardCollisionGroup.WORLD), mask); // adds rigid body to world
```
4Denthusiast, line 82: Could you describe what exactly the mask and the rigid body's collision group are for? As I understand it, the collision group specified on line 82 is the group that this mesh is itself a part of, so that if a raytrace is requested, this group determines whether this mesh is included, therefore it should be LIQUID, not WORLD, and the mask determines what other groups this collides with, which should probably just be nothing (0), because liquids don't actually collide. I expect it would still be possible to make something collide with this by including LIQUID in _its_ collision mask (e.g. potentially a duck or a player wearing boots of water-walking or something).

engine/src/main/java/org/terasology/physics/engine/VoxelWorldSystem.java
```diff
@@ -100,17 +100,20 @@ public void initialise() {
...
         short id = block.getId();
-        if (!registred.contains(id)) {
-            btCollisionShape shape = ((BulletCollisionShape) block.getCollisionShape()).underlyingShape;
-            btVoxelInfo info = new btVoxelInfo(shape != null && block.isTargetable(),
-                shape != null && !block.isPenetrable(), id, shape, block.getCollisionOffset(),
-                block.getFriction(), block.getRestitution(), block.getFriction());
-            wrapper.setVoxelInfo(info);
-            registred.add(id);
+        if (!registered.contains(id)) {
+            if (!block.isLiquid()) {
```
4Denthusiast, line 109: This `block.isLiquid()` check isn't actually necessary. Liquids generally have the penetrable property and don't have the targetable property anyway, and in the odd case that they don't (e.g. FlowingLiquids.DebugLiquid), it would be more consistent to just leave them in this collision mesh anyway.

4Denthusiast:
This PR is superseded by https://github.com/MovingBlocks/Terasology/pull/4221, and is no longer needed.

PR #4212, opened 2020-10-21
fix(gradle-server): restore missing suffix for jvmarg -Xmx

DarkWeird:
### Contains

Fix jvmArgs for `server` gradle task. 
Bug prevent server to start with error:
```
Error occurred during initialization of VM
Too small maximum heap
```
introduced in #4175

### How to test

1. Checkout it
2. run `server` task.
3. check that server started (any output from server)

### DIFF
--- a/facades/PC/build.gradle.kts
+++ b/facades/PC/build.gradle.kts
@@ -255,7 +255,7 @@ tasks.register<JavaExec>("server") {
     main = mainClassName
     workingDir = rootDir
     args = listOf("-headless", "-homedir=$localServerDataPath")
-    jvmArgs = listOf("-Xmx1536")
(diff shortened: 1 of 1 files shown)


skaldarnar (approved):

keturn:
Thank you for this, @DarkWeird. It was sloppy of me. I had been thinking that it was too small because other reasons, not because I had made a typo to only allocate 1,536 _bytes._  :sob:

DarkWeird:
> Thank you for this, @DarkWeird. It was sloppy of me. I had been thinking that it was too small because other reasons, not because I had made a typo to only allocate 1,536 _bytes._  :sob: 

Yeah I thought that too at first. 
I began to suspect later

PR #4217, opened 2020-10-26
build: add extract natives for jinput

pollend:
extract natives for jinput.


skaldarnar (approved):
These changes look reasonable enough to fix some problems with `jinput` not being present :sweat_smile: I've quickly corrected some copy-paste typo, let's get this merged!

PR #4218, opened 2020-10-26
feature(controllers): replaced Jinput controller lib with LWJGL3 (GLFW) controller lib

DarkWeird:
### Contains

Replacing Jinput lib with GLFW lib for controllers.

### How to test

1. Start terasology with pluged controller.
2. start game
3. try buttons and sticks on controller, it must works as with jinput previously.
4. unplug and plug controller during game. 

P.S. check buttons GUIDE(idk what is it), LEFT_THUMB, RIGHT_THUMB(click on stick).
There can be invalid mapping.

- [ ] Test mappings for axes and buttons from GLFW to Terasology.
- [ ] Implement joystick handling. (currently only gamepads)


pollend:
we should probably also drop net.java.game.jinput if were not going to use it. I believe this is the only dependency in the engine with jinput.

DarkWeird:
> we should probably also drop net.java.game.jinput if were not going to use it. I believe this is the only dependency in the engine with jinput.

Dropped

pollend (approved):

skaldarnar:
Tried with xbox 360 wireless controllers (i) connected to SteamLink, and (ii) connected to Windows 10 pc directly. I'm not sure where the problem is, but the experience was not really ... pleasing 😕

- I could not spot a difference in behavior between current `develop` and this PR
- The "key mapping" was off when connected via Steam Link. The buttons and sticks are mostly mapped to keys (like ENTER or PAGE_UP), the left joystick only works in vertical direction as MOUSEWHEEL_UP/MOUSEWHEEL_DOWN
- Even though the driver software correctly identifies the controllers and buttons in Windows, Terasology did not 🙈 

tl;dr: Doesn't look like this PR makes it worse, but it wasn't really working either...

pollend:
Slight improvement since we don't need to depend on jinput

DarkWeird:
@skaldarnar 
Are you ok with this this PR? - no differences between in functionally, but without jinput

PR #4219, opened 2020-10-26
fix(nui-input): access to `KeyboardDevice`

jdrueckert:


Cervator (approved):
I coincidentally ran into the same issue and had already validated the fix locally, although mine was subtly different (package level). I like this better although in both cases I wonder if we aren't just missing a bunch of other formerly `@API` things in NUI now extracted with the annotation doing nothing. We should probably look through and either cover all API classes from NUI in the whitelist or find a better approach somehow 🤔 

Tested again with this merged, all is well 👍

PR #4223, opened 2020-10-29
build: log Java version and JDK path with gradle warning.

### LINKED ISSUE #4209: "WARNING: Compiling with a JDK not 8 nor 11" warning should print more info about what JDK it's using
Someone new to the project was trying to build it from source and they got the `Compiling with a JDK not 8 nor 11` warning, as seen in this log: https://pastebin.com/na70rGh4

but when they do `java -version` at the command line, it reports OpenJDK 11.0.8, which _is_ a JDK 11.

That warning could help by printing more details about which JDK gradle is running under, as it may be using some configuration that has it using a different thing that what the shell resolves `java` as on the command line.

Details it would be useful to have include:

* JDK version number
* other JDK build info (i.e. the part that shows if it's from AdoptOpenJDK or built by another distro)
* the absolute path to the executable it's using

The code involved is here:

https://github.com/MovingBlocks/Terasology/blob/5d87630d6331f240585e9353c7c3a739c7a172ef/build.gradle#L51-L55

dhruvil-93:
Log Java version and JDK path with gradle warning about non-compatible JDK version used while compilation.
Resolves #4209


keturn (changes requested):
Yeah, I think this is basically what we're going for!

Can you include the `java.vendor` property too?

also FYI, `.gradle` files use the Groovy language, and you can do [string interpolation](https://docs.groovy-lang.org/latest/html/documentation/#_string_interpolation) with `${}` like

```groovy
prinln("blah blah: ${JavaVersion.current()} and stuff")
```

as an alternative to adding strings with `+` that's often easier to read.

keturn (approved):
Looks like it should do the trick :+1:

skaldarnar (approved):

PR #4227, opened 2020-10-31
chore: Archive COLLADA format

Cervator:
See https://trello.com/c/Xj9Iks10/41-consider-reviving-collada-format - in short COLLADA was never quite finished and never used and now with `glTF` in place we're trying to clean up.

I checked with @mkienenb who originally contributed the format (and many other things of both esoteric and useful flavors!) and it was nice to hear a sign of life, but yeah this isn't useful right now and isn't likely to be revived in the near future. I still prepared a Trello backlog card for it and will point at this PR in case anybody some day wants to revive the support. Juuuust in case since I do dislike losing potentially useful stuff. This way we can say we've just parked the code on a nice comfy farm somewhere up the river where it can live happily in retirement until called upon again.

I think I got all the related files and nothing should be affected by this going missing. Did some assorted spot testing with no obvious problems.


skaldarnar (approved):
LGTM

Cervator:
[Consider reviving COLLADA format](https://trello.com/c/Xj9Iks10/41-consider-reviving-collada-format)

PR #4228, opened 2020-10-31
feat(JOML): migrate block placement

pollend:


jdrueckert:
Requires changes in
- [ ] Minimap
- [ ] StructuralResources
- [ ] Furnishings
- [ ] Rails
- [ ] MultiBlock
- [ ] IRLCorp
- [ ] Smithing
- [ ] MobileBlocks
- [ ] ModularComputers
- [ ] FlowingLiquids
- [ ] Fluid
- [ ] StructureTemplates
- [ ] ItemPipes
- [ ] Climbables
- [ ] ChangingBlocks
- [ ] DynamicCities
- [ ] GrowingFlora
- [ ] AnotherWorldPlants

engine/src/main/java/org/terasology/world/block/entity/placement/BlockPlacingSystem.java
```diff
@@ -34,6 +38,9 @@
 
     @ReceiveEvent(components = {WorldComponent.class}, priority = EventPriority.PRIORITY_TRIVIAL)
     public void placeBlockInWorld(PlaceBlocks event, EntityRef world) {
-        worldProvider.setBlocks(event.getBlocks());
+        //   worldProvider.setBlocks(event.getBlocks());
```
jdrueckert, line 41: do we still need this? if so why is it commented out? if not, why is it still there :yum:

jdrueckert (changes requested):

engine/src/main/java/org/terasology/world/block/entity/placement/BlockPlacingSystem.java
pollend: looks like the same line but commented out :P

pollend: opps

jdrueckert (approved):

PR #4125, opened 2020-08-15
chore(settings): disable identity service login ui features

Ankhell:
### Contains

Added new prop "overrideChildEnabledProp" to AbstractWidget to avoid overriding of child enabled property during initialization (In this exact case - by UIBox), modified UIFormat to process enabled field after Layout content initialization, disabled Identity Service Login features in UI

Contributes to #4104

PR #4229, opened 2020-11-01
fix(menu): remove notification to configure identity service

### LINKED ISSUE #4104: Disable Identity Service Login
We've got an increasing amount of questions and reports regarding the **Identity Service Login** from players trying out the game. While the backing service (https://github.com/MovingBlocks/terasology-key-server) is in maintenance mode and not working correctly, we should disable the login option from the UI to not confuse players.

ℹ️   Fixing the Identity Service itself is a different issue!

I believe this is the part of the UI that renders information for the Identity Service:
https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/resources/assets/ui/menu/playerMenuScreen.ui#L64-L104

Related forums post: https://forum.terasology.org/threads/unable-to-login-to-terasology-multiplayer.2314/

skaldarnar:
Follow-up to #4125 

This PR removes the notification (top right of the main menu) redirecting the user to the settings menu to set up the identity storage (the respective fields were disabled in #4125). In addition, no notification message will be logged in the bottom left of the menu if the identity service token is not set and the user could not be logged in.

Closes #4104


jdrueckert (approved):
Looks good. However I'm wondering whether we really need to remove the label and its translations? We do want to revive the identity service after all.

skaldarnar:
> However I'm wondering whether we really need to remove the label and its translations? We do want to revive the identity service after all.

Do we need to? Probably not. On the other hand, I doubt that anybody will pick up that exact same label when re-implementing the whole thing. Instead of removing the translations and the log messages I could have commented it, leaving the i18n strings in place. I would guess this would just become technical debt sooner or later without anyone noticing...

PR #4231, opened 2020-11-02
Migrate Omega zip packager jobs to new Jenkins and overhaul user-displayed version details everywhere

Cervator:
## Overview

This is the engine-side part of finally migrating the omega dist packaging jobs from the old Jenkins (jenkins.terasology.org) to the new Jenkins (jenkins.terasology.io), so we can switch over the launcher to source builds from the new location and eventually decommission the old Jenkins entirely.

There's a fair bit more to unpack here than the relatively few changes suggest, but for the main part here it is really just triggering a second job to do the Omega zip after the engine job has published. If not interested in the gory technical details can skip most the below.

The second change is reworking the pattern of what we display to an end user when it comes to release/version details, which has historically been all over the place. Now it is fairly consistent and should be possible to link between different places more easily

After this is merged alongside maybe two more PRs then the launcher should then be able to pick up two new package lines with nice clean names.

* TODO: InfraPlayground PR with the new Omega job for Terasology (the [Nanoware one is here as DSL](https://github.com/MovingBlocks/InfraPlayground/blob/master/jenkins/jobDSL/Nanoware/omega.groovy) and [here as an actual job](http://jenkins.terasology.io/teraorg/job/Nanoware/job/Omega/)
* TODO: Index PR with a Jenkinsfile that the Omega job will actually run. This is the [old Omega freestyle job](http://jenkins.terasology.org/job/DistroOmega/) converted into a [simple Jenkinsfile](https://github.com/Nanoware/Index/blob/develop/distros/Jenkinsfile)

## Gory details - version numbers

This slowly turned into a bit of a wasps' nest over the years, as few things have really been in sync / map well to each other.

* The jobs in Jenkins come in two sets, release ("stable") builds and dev ("unstable") builds with the engine alone jobs containing no modules and the omega jobs come with full modules, each of them with their own _build_ number, based on the job in Jenkins. So for instance "1103" is the latest "dev build" _with_ the extra modules in Omega, and there are _three_ others (the engine build number for dev Omega 1103 for instance is actually 2317). These numbers reset if the jobs in Jenkins are ever remade and in the grand scheme should _not_ be meaningful on their own. Additional which builds are stable vs unstable actually varies widely, names or lines aside.
* The engine itself has a version number, currently `4.1.0-SNAPSHOT`, using SemVer just like modules and every other actual individual software component. The engine tests, PC facade, etc also have such numbers, but luckily those aren't really used anywhere. During the engine build a `versionInfo.properties` file is generated _if_ we're building within Jenkins.
* On GitHub when we do a release we're really releasing the _game_ not the _engine_ even though that's really all that's left in this repo (we moved out the remaining modules here). So engine `4.1.0` isn't the same as "The game package with some modules and the 4.1.0 engine" - it needs its own name/number. So years ago, probably around when we went from pre-alpha to alpha I just arbitrarily threw a number after "Alpha" meaning our most recent release on GitHub is "Alpha 16" and this is used exactly nowhere else (other than occasional announcements that said release is out)
* The launcher loads numbers from Jenkins, grouped on some package lines we define that map with the job lines in Jenkins. This is another spot where we've pretty much just made up names, that then pair with the build numbers from Jenkins that _shouldn't_ be meaningful. Right now for instance we've got "Nightly" and "Lite" in there which again doesn't match anything else anywhere. All 4 job lines are represented (dev builds are "nightly", non-omega are "lite")
* When you actually load the game there's a version string on the main menu, which loads some details originally determined during the Gradle build and stored in a `versionInfo.properties` file we cram into the engine jar file and also attach in Jenkins. However, this again barely matches with everything and isn't clear at all.

### Jenkins

On the surface Jenkins remains roughly the same as before

* URLs naturally vary, both from the base URL and from mostly everything being multi-branch pipeline jobs now. In _theory_ we could have just gone with more plain pipeline builds for the omega packager, but this is nice and easy plus consistent. This does mean there'll be both a `master` and `develop` branch in the Index repo now as well, when we've only ever had the one `master` branch before. Seems like we should be following that workflow anyway so if we ever need to rebuild a release omega it won't fetch modules added in `develop` since last (it _will_ however still grab the very latest version of every module, but that's a fight for another day and I have thoughts on it)
* The latest versions of Jenkins abstract away the SCM systems, so Git is no longer a natural default exposed via env variables. This seems OK and we can still get at that data if we really need it, but I don't think it is really in use. Technically the `VERSION` file is meant to contain the branch, but I'm not sure that _ever_ gets looked at, and while it is referenced in both the engine and PC build files I think it is only used in the facade. Got one more commit to clean up and maybe look at this just a little more.
* The `versionInfo.properties` file is now also attached to the Omega job (it wasn't previously, just living inside the engine jar), so ideally the launcher could get at it there (more easily?) and not need the bare engine jobs at all.

### Engine

Several changes here, whole diff in fact since this is the engine repo after all!

* Assorted bits of cleanup / outdated stuff, like removing the two Git vars no longer exposed by default in Jenkins, and swapping in the `jobName` fully instead. Also for some reason the SNAPSHOT vs not deduction was in the engine plus engine-tests even though it is explicit in there, so in some cases we could get `-SNAPSHOT-SNAPSHOT` in the version stuff nobody reads
* Added the informal alpha increment number to the weird old `templates/version.txt` which previously just held "alpha" - we need to start maintaining this now but that isn't exactly a huge effort yet should make a huge difference visually in places. Not really sure why that was ever considered a "template" but this is truly ancient stuff.
* Updating the version detail calculating logic that'll be used in the game UI (see its own section below)
* Updating the version details that'll be used in `MANIFEST.MF` not that anybody ever looks at it (did you know: it actually held the build number _twice_ for some reason?)
* The Jenkinsfile now will trigger the Omega job to run on the publish stage otherwise succeeding (no real need to wait for analytics as we don't use them to gate anything yet, that may change in the future). This is hard-wired between `master` + `develop` and the corresponding Omega jobs, including Terasology vs Nanoware which needs to be adjusted before this is merged. That can be deduced but tired and getting late :-)
* Keying off `JOB_NAME` now instead of `BUILD_URL` for calculating fancy version details for later user display - both set by Jenkins, no idea why I picked the URL in the past, job name seems more useful

### GitHub 

When we do a manual release (until we finally fully automate that somehow) we need to be sure to also update the `templates/version.txt` file as part of the prep for the _next_ release (akin to version bumping and re-attaching `-SNAPSHOT`) so that we call the releases on GitHub the same thing as we've codified it to be in the engine repo. This probably should go on some wiki page I've got at the top of my tongue but late + tired. So: TODO

### Launcher

These are the current package lines in the launcher:

image

I don't remember exactly what the launcher pulls and has access to, I can dig around some if needed. But I do remember we attached the `versionInfo.properties` file in Jenkins so it would be publicly accessible, so hopefully the launcher can use it to parse out further details.

My suggestion would then be the following (if we can easily add package lines without them all using the same base URL, so we can have both new and old entries for a while):

* Add two new lines that solely represent the Omega packages in the new Jenkins (the engine zip alone has no modules so it isn't very useful anymore). Call them "Terasology Release" and "Terasology Preview" - do away entirely with "stable" and "unstable" and "dev build" and "nightly" and "lite" - just those two things. I nearly went with "snapshot" since I'm a dev before I remembered to think like a player who wouldn't necessarily know and shouldn't care what a "snapshot" is, the dev builds are _previews_ of the next release, simple as that.
* Grab the `displayVersion` field from the `versionInfo.properties` file attached in Jenkins for every build, use that as the _primary_ meaningful name for the entries associated with the package line, paired with the actual job build number, which may reset over time, but will never repeat when paired with the display name of the build (we'd aim to tie any future resets with game releases or something if needed). Instead of "Terasology (nightly)" + "1103" You'd have "Terasology Preview" + "Alpha 17 build 1103" (actually it'd be 2317 but ... bear with me the answer to that is elsewhere, and really the idea is to make these numbers meaningless to where it doesn't matter _which_ is there, just that higher number for a given alpha version is newer)
* Useful URLs:
  * http://jenkins.terasology.io/teraorg/job/Nanoware/job/Omega/job/develop/lastSuccessfulBuild/buildNumber - automatically shows the most recent build number, the package line in the launcher would start here then go backwards `x` numbers until it either has a nice list of entries or runs out (id goes below 1). Naturally needs to be adjusted from "Nanoware" to "Terasology" when the change is live. I believe this is how it works as-is, just need the updated URL
    * Note that this should _always_ be "lastSuccessful" as the Omega packager has no unit tests. It either succeeds or fails, there's no unstable to worry about like with the engine job
  * http://jenkins.terasology.io/teraorg/job/Nanoware/job/Omega/job/develop/2/artifact/distros/omega/build/distributions/TerasologyOmega.zip - game zip for a given build 
  * http://jenkins.terasology.io/teraorg/job/Nanoware/job/Omega/job/develop/2/artifact/distros/omega/versionInfo.properties - info prop file for a given build
  * For the Release line instead of the Preview line swap `develop` to `master` (no example as the `master` build for Nanoware fails for some inexplicable reason, I might blow it up and let it recreate)

It should be possible to test this against the Nanoware develop job + develop Omega right now in the launcher, assuming we can add such a new package line entry for testing.

### Game UI

Here's what the current main menu looks like when it comes to the version label:

2020-11-01 19_55_29-Terasology _ Alpha

This is not very useful. The "2317" is really the only thing that ties it to something and that's the _engine_ build number from Jenkins for the dev build line. Wha? This isn't even what you get if you use the "nightly" Omega line in the launcher, since that's a _different_ build number, namely 1103 from the Omega dev job.

Here's what you'd get with the changes, although the `null` would be replaced by the engine job's build number if built in Jenkins (you get this by setting the `JOB_NAME` env var to `Terasology/engine/develop` locally then running `gradlew createVersionInfoFile game` - I skip setting the build number when testing locally)

2020-11-01 19_59_24-Terasology Alpha

That's a little better. The highlight is you've got an Alpha 17 preview build. Alright, that's a good baseline. If we pretend this was built in place of the current dev build + Omega it would list "engine v4.1.0-SNAPSHOT, dev build 2317..."

Still not quite perfect as the _launcher_ would still list the build number as 1103. We could in theory update that during the Omega packaging job by going in and updating the `versionInfo.properties` file inside the engine jar. But the engine doesn't really know or (should?) care whether it was built standalone or repackaged via Omega? Dunno.

If it was built from the master branch / release Omega it would just be "ALPHA 17" with a similar detail string (minus the snapshot and "dev" bits). Still would have the engine build number instead of the omega build number though.

For other cases we get stuff like this:

2020-11-01 20_34_05-Terasology Alpha

Anything that isn't either the main master or develop branch builds yet is still built out of Jenkins (or faked locally via `JOB_NAME`) gets that "Special:" decoration with the exact job path it came out of.

Finally if you don't have `JOB_NAME` set you're likely running from source, or you ran a manual build for some odd reason. You get a simplified label since we can't really be sure what's going on so we just call it "Custom"

image

For the absolute vast majority of time a regular player should only ever see the "ALPHA 17" or "ALPHA 17 Preview" examples on a single line, with a second line of details just in case (that should probably be shrunk and made less prominent, maybe even get moved somewhere else). All the other cases are for devs.

## Extras

**Likely future change:** We should probably move the concept of the game package naming/versioning entirely out of the engine repo into the Index repo or whatever it becomes. For right now it is just easier to keep it in the engine since it works and gets shoved into the engine jar which then gets used by the omega packager. This would likely pair well with a better definition of the module / game distribution lines than prop files in a folder somewhere. As part of the GooeyCLI work I started looking at this, including module lines that don't get built but form useful groupings (like "all tutorials"), but that needs more work, and might go to a different format than poor old .properties (JSON or YAML)

**Potential subtle complication:** if an omega packaging job is executed _manually_ it'll copy engine artifacts from the last _successful_ build - as in, no broken unit tests. If we have unstable builds (flakey engine tests) the packager would grab the most recent _fully_ successful engine build. In theory this should be a very rare occurrence since that would only ever happen if we manually wanted to rebuild an omega zip alone, rather than just run a fully new engine build. Full build is more the recommended approach, and historically I've tried to find a quick simple PR to merge just to retrigger a proper Omega (or just manually executed the engine build itself). But there might be a rare time when we're trying to test a replacement module real quick, where we in that perfect storm of conditions could end up with a slightly older engine build than expected. 

**Here be dragons:** This PR will likely fail or do really weird things when tested in Jenkins

**Good First Issue:** If this approach looks reasonable then we could make it a little prettier by splitting the version label and putting the fine details of the version on line two into a separate label with a smaller font size. Easy GFI. TODO.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/BlockRegion.java
+++ b/engine/src/main/java/org/terasology/world/block/BlockRegion.java
@@ -443,9 +443,9 @@ public class BlockRegion {
      */
     public Vector3f center(Vector3f dest) {
         return dest.set(
-                (aabb.maxX - aabb.minX) / 2.0f,
-                (aabb.maxY - aabb.minY) / 2.0f,
-                (aabb.maxZ - aabb.minZ) / 2.0f
+                aabb.minX + ((aabb.maxX - aabb.minX) / 2.0f),
+                aabb.minY + ((aabb.maxY - aabb.minY) / 2.0f),
+                aabb.minZ + ((aabb.maxZ - aabb.minZ) / 2.0f)
         );
     }
 


skaldarnar:
> The `versionInfo.properties` file is now also attached to the Omega job (it wasn't previously, just living inside the engine jar), so ideally the launcher could get at it there (more easily?) and not need the bare engine jobs at all

> I don't remember exactly what the launcher pulls and has access to, I can dig around some if needed. But I do remember we attached the versionInfo.properties file in Jenkins so it would be publicly accessible, so hopefully the launcher can use it to parse out further details.

That's pretty neat. I was just playing around with collecting the SemVer for our releases across different Jenkins jobs, this would make it a lot easier! It currently doesn't use that information, but that's a simple change indeed.

> My suggestion would then be the following (if we can easily add package lines without them all using the same base URL, so we can have both new and old entries for a while):

I'm working on a refactoring that will allow to easily consume game releases from several places. Having a unique way (SemVer) of identifying things would help a lot. I'm looking for an identifying tuple of `(semver, profile)` where the profile is the "codeline" or "release type", i.e., whether we're talking about Omega, the bare engine, or whatever in-between there might be in the future.

> automatically shows the most recent build number, the package line in the launcher would start here then go backwards x numbers until it either has a nice list of entries or runs out (id goes below 1). Naturally needs to be adjusted from "Nanoware" to "Terasology" when the change is live. I believe this is how it works as-is, just need the updated URL 

We currently just use the REST API for the job to get the list of builds in one batch and scan that for successful builds which also have artifacts attached. I'd like to keep it that way, instead of grabbing one build at a time and counting backwards manually.

Here's what I get with my local testing setup in mid-refactoring land:

```log
19:37:20.964 [Launcher init thread] INFO  o.t.launcher.LauncherInitTask - Fetching game releases ...
19:37:20.979 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - fetching releases from 'http://jenkins.terasology.io/teraorg/job/Nanoware/job/Omega/job/develop//api/json?tree=builds[actions[causes[upstreamBuild]]{0},number,result,artifacts[fileName,relativePath],url,changeSet[items[msg]]],upstreamProjects[name]'
19:37:21.865 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - Derived SemVer for OMEGA@2+NIGHTLY: 	4.1.0-SNAPSHOT+2
19:37:21.867 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - Derived SemVer for OMEGA@1+NIGHTLY: 	null
19:37:21.867 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - fetching releases from 'http://jenkins.terasology.io/teraorg/job/Nanoware/job/Terasology/job/develop//api/json?tree=builds[actions[causes[upstreamBuild]]{0},number,result,artifacts[fileName,relativePath],url,changeSet[items[msg]]],upstreamProjects[name]'
19:37:22.974 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - Derived SemVer for ENGINE@4+NIGHTLY: 	4.1.0-SNAPSHOT+4
19:37:23.417 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - Derived SemVer for ENGINE@3+NIGHTLY: 	4.1.0-SNAPSHOT+3
19:37:23.552 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - Derived SemVer for ENGINE@2+NIGHTLY: 	4.1.0-SNAPSHOT+2
19:37:23.675 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - Derived SemVer for ENGINE@1+NIGHTLY: 	3.3.0-SNAPSHOT+1
19:37:23.675 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - fetching releases from 'http://jenkins.terasology.io/teraorg/job/Nanoware/job/Omega/job/master//api/json?tree=builds[actions[causes[upstreamBuild]]{0},number,result,artifacts[fileName,relativePath],url,changeSet[items[msg]]],upstreamProjects[name]'
19:37:23.948 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - Derived SemVer for OMEGA@1+STABLE: 	null
19:37:23.948 [Launcher init thread] DEBUG o.t.l.r.NanowareJenkinsRepositoryAdapter - fetching releases from 'http://jenkins.terasology.io/teraorg/job/Nanoware/job/Terasology/job/master//api/json?tree=builds[actions[causes[upstreamBuild]]{0},number,result,artifacts[fileName,relativePath],url,changeSet[items[msg]]],upstreamProjects[name]'
19:37:24.226 [Launcher init thread] INFO  o.t.launcher.LauncherInitTask - ENGINE@1+NIGHTLY
```

## Questions

- **Why do you want to display the user a version such as "Alpha 17" instead of a Semver?** I'd like to understand why we try to maintain an additional version here, instead of just carrying over the engine version to the Omega release? Is it important that we are still "alhpa"? Do you think a Semver is too complicated?
	I'd prefer to display a Semver in the launcher, for instance `3.5.1` or `4.1.0` for "stable" releases, and (a slightly modified) Semver for previews, e.g., `4.2.0 Preview (202011081947)` (which would be `4.2.0-SNAPSHOT+202011081947` under the hood). That way we would get the order of releases from semver semantics.

	Also, I'd like to change the way we display the releases in the launcher. Instead of the having 4 entries in the dropdown I'd like to only have two for the main release lines (Omega, Lite (engine)) and just list all releases for the respective code line _in order_. (plus adding an option to only show stable versions/hide preview versions).
- **In the old Jenkins we also had the changelog to collect and display in the launcher. Is that still available in the new setup?**

Cervator:
> Why do you want to display the user a version such as "Alpha 17" instead of a Semver

A few reasons, but neither of them particularly deep or permanent :-)

* Players don't care about SemVer, or necessarily understand it. Sure, they can just compare numbers and see what's newer (especially a small little corner tag in the app showing the "technical version"). But "Alpha 16" vs "Alpha 17" is probably a more meaningful human comparison than "engine v3.1.0" vs "engine v3.2.1": "Oh, okay, Alpha 17 is 'one newer'"
* Engine version may theoretically not change at all for a new release, if for some reason we need to just change some modules (most likely case: a quick re-release due to some bad modules)
* Omega != engine - it really should have its own release number IMHO, probably tracked in the Index repo, which in turn explicitly points out what engine version, module versions, etc, are dependencies for that Omega release. That could itself be a SemVer number, but then the point about player meaningfulness still stands. That seems like a future discussion, wasn't something I wanted to try throwing code at right now.

I'm not particularly attached to any of those points, they're just my arguments in favor of why it is the way it is right now.

Launcher listings: I don't think "Lite" makes any sense anymore. You literally can't enter a world that way, without yourself adding modules, possibly via the semi-broken in-game module downloader. It just serves as a point of potential confusion to the user now. Additionally I'm not sure what actual player would want that - it seems more like a modder/dev thing.

_Unless_ "Lite" turns into the Iota module set or something instead. But then that just may get into more naming confusion, plus size-wise there isn't that much of a difference - 70mb vs 150mb). I think merely having the two entries to start with (Release vs Preview) would be ideal, possibly with different lines hidden by default (Iota if it makes sense to distribute it via launcher, kinda skeptical, experimental branches, etc)

> Changelog

Yep: http://jenkins.terasology.io/teraorg/job/Nanoware/job/Terasology/job/develop/changes for instance (different path of course for the live game's release and preview tracks)

Great to see the progress and options, and I'm sure we'll have more of them, just the sooner we can get past this point the sooner we can destroy the old Jenkins in a maelstrom of fire 😁

skaldarnar:
Thanks for taking your time to explain this.

> Players don't care about SemVer, or necessarily understand it. Sure, they can just compare numbers and see what's newer (especially a small little corner tag in the app showing the "technical version"). But "Alpha 16" vs "Alpha 17" is probably a more meaningful human comparison than "engine v3.1.0" vs "engine v3.2.1": "Oh, okay, Alpha 17 is 'one newer'"

I think AAA games conditioned players quite well to understand semver by now. Nearly all patch notes I'm aware of refer to (more or less) a semver, for instance [Minecraft 1.16.2](https://www.minecraft.net/en-us/article/minecraft-java-edition-1-16-2), [Valorant 1.11](https://playvalorant.com/en-us/news/game-updates/valorant-patch-notes-1-11/), or [Witcher 3 1.30](https://en.cdprojektred.com/news/patch-1-30-changelog/). However, major releases (not necessarily in the sense of semver, though) often come with catchy names or titles like "The Aquatic Update" or "The Cave Update". 
Preview releases of the next major update can often be found as "release candidates" which can be expressed in semver as `2.0.0-rc.1`. These release candidates count up from _1_ to _n_ regardless of the underlying build number.

> Engine version may theoretically not change at all for a new release, if for some reason we need to just change some modules (most likely case: a quick re-release due to some bad modules)

Yep, totally agreed. I take back my suggestions to completely bind the Omega release version to the engine version. Especially the case of "re-release" should IMHO be a patch release (`4.0.1` is fixing a broken `4.0.1` build), so this version part should be available for exactly this use case.

> Omega != engine - it really should have its own release number IMHO, probably tracked in the Index repo, which in turn explicitly points out what engine version, module versions, etc, are dependencies for that Omega release. That could itself be a SemVer number, but then the point about player meaningfulness still stands. That seems like a future discussion, wasn't something I wanted to try throwing code at right now.

I like the idea of having dedicated versions for the release distributions we provide. Where these numbers should start and how they may look like is a completely different topic - and probably quite complicated 🙈 First thing to keep in mind is that we have "press coverage" that we released Terasology v4, so changing the version scheme should take this into account somehow. The v4 comes from the engine version, so we can probably keep that in some way.

Technically the Index repo could define an aribitrary semver for the released distributions, e.g., `0.17.0` for Omega "alpha 17", but this does not contain any information on engine or modules. We could derive a _module lineup version_ from the included modules (or rather, the version bumps of the included modules) in some way, but that leaves us with two semvers to identify the release (engine and module lineup) - meh.
I'd like to propose a version scheme that takes both into account, but melts it into a single Semver:

```
<engine.major>.<# of module changes>.<patch>
```

- `<engine.major>` - Simply the major version of the engine. We don't care about the exact version in the distribution (at least in the displayed version), but still couple our omega release to the engine release.
- `<# of module changes>` - The first release with a new major engine version would start at _0_, and every major or minor module version bump would increase this number by one (where "every version bump" is every bump within release builds, i.e., the next release build could contain multiple updated modules but still only bump this version by one).
	Not sure about minor engine changes that don't require any module updates (new features) - they would also have to increase this number in case nothing else changed...
- `<patch>` - Any patches or fixes applied to the distribution. Both engine patches and module patches contribute to this version. 

So `4.1.0` would be the second "stable" release with engine v4, and `5.42.3` would be the 3rd attempt to get the 42nd release with engine v5 to work properly. The same version scheme would work for other distributions/line-ups.

> Launcher listings: I don't think "Lite" makes any sense anymore. You literally can't enter a world that way, without yourself adding modules, possibly via the semi-broken in-game module downloader. It just serves as a point of potential confusion to the user now. Additionally I'm not sure what actual player would want that - it seems more like a modder/dev thing.

We can remove the "Lite" option from the launcher easily, no problem with that. However, if we ever fix the "semi-broken in-game module downloader" I think it would be a viable option to still offer it for players to just grab what they want (instead of pre-bundling everything). At some point, the whole update/upgrade topics will come up again (update single modules, update the engine, ...). Let's postpone that...

skaldarnar:
@Cervator is there a way to get the version info into Jenkins metadata instead of attaching it as a file (or in addition to)? Fetching the builds overview via the REST API is quite fast, but doing round trips for fetching the version file for every build slows down the launcher startup noticably. The more info we can get with a single request the better (we can use some way of filtering directly in the request to even reduce to only the stuff we actually want).

I'll see whether I can get the changes directly, but doing _one_ more request sounds fine, too.

Cervator:
>  get the version info into Jenkins metadata instead of attaching

Well, _maybe_ somehow, I'm not sure. We probably don't want to tie build packages _more_ to Jenkins, if our longer term goal is to support build packages somewhere else? In fact, maybe what we should do is go in a different direction and make the meta server also store an index for launcher packages? It has an API, stores stuff in a database, is tool agnostic, etc ... skip ever having the launcher calculate anything, make the meta server worry about it and just pull a catalog. That'd be a step toward arbitrary build packages hosted elsewhere too.

That'd be more work so maybe it would be worth just accepting a bit of a performance drag right now to be able to move forward and iterate more on speedier more rich launcher metadata later? 

> I think AAA games conditioned players quite well to understand semver by now.

I suspect this might be a hard call. Veteran gamers like us? Sure - you don't need to be a developer to sort out numbers. But school kids? We might be biased due to our immersion in the topic. Minecraft is a bit of a special case due to the _intense_ focus on modding - admittedly also a goal of ours, we're more similar to that than the polished AAA games that likely next to nobody pays attention to numbers for.

> catchy names or titles like "The Aquatic Update" or "The Cave Update".

I like these and would love to get there, but they're probably _too_ far away from numbers right now, until we have a more cohesive game and update pattern. I figured maybe by the time we go Beta (or even release) we'd call the overall game v1.0.0 and reassess naming by then.

> derive a module lineup version from the included modules

That one makes me itch a little, since suddenly we're using a non-SemVer approach in the SemVer pattern. So some places are SemVer, while others look like it but mean something else :-)

_Maybe_ we could use a game-level SemVer but use the decorators after `x.y.z` to indicate the game line / module updates? That could potentially be the best of both worlds. A meaningful number that both respects SemVer but tells a bit more of a story. Although I think it'd be nice to still have a more "human" name for the main releases.

* `x` - breaks save compatibility with past worlds (backwards incompatible)
* `y` - there are new things (one or more in _any_ component)
* `x` - bug fixes only (literally every changed component only incremented at the patch level)

> We can remove the "Lite" option from the launcher

I do think that for now we should remove it, that doesn't mean we couldn't bring it back later, or perhaps more likely Iota? Along with Omega or the other letters for a few predefined lines. Or better names yet (Iota -> Lite or what not). Just so long as we can default to something very clear that a user can just click a button for, get, and play. Consistency is key and right now we need a bunch of that until we can doc and display our things better.

Everything above is just quick opinions, honestly as long as we're moving forward I'm fine with whatever approach as long as it gets the attention it needs 👍 

On a somewhat related note: If we end up writing an outright "Omega book of materials" style thing for a given release rather than just a list of module names in a prop file that would be helpful in parsing out _actual_ changes in modules since a previous point (even multiple releases). At that point rather than just a vague number meaning "42 modules changed for this release!" you could give it whatever number and provide collapsible change notes for the exact diff. Kinda like writing out a pinned version/dependency file for some languages.

jdrueckert:
Notes from today's org meeting:

snapshots are overwritten 
-> content changes, not useful for reproducible builds
-> for reproducible builds we’ll need release management for all modules
default version resolution: always use latest (including snapshots)
gradle can “pin” (lock) versions
checked-in lock file
idea 1: have module builds update pin lock file
idea 2: only pin released modules, use gradle dependency resolution

jdrueckert:
whoops sorry. accidentally closed this while trying to test the release process. the changes that were on `Nanoware:develop` now live in `topic/packaging-and-versions`

Issue #4180, opened 2020-10-04
Visual artifacts on crosshair


Gizmordo:
### What happened
Visual artifacts on crosshair
![unknown](https://user-images.githubusercontent.com/65704439/95005283-63bd2080-05c4-11eb-88b1-fa2219c81948.png)

# Game version
37

Cervator:
Can confirm. May be a fairly easy fix if digging around in the engine for the asset / UI element leads to anything obvious. Also may want to try checking out some really old game versions to see if it was introduced around a given period of time, perhaps with some relevant sounding PRs having been merged around that time.

aRealCyborg:
From what I can tell this bug has started in version 37 in my test.
Version 36, 35, 25 I remember didn't have them.

jdrueckert:
Unable to reproduce on Linux in Terasology Nightly 1103 and latest iota from source.

Cervator:
Took off a couple labels as this is likely to be GPU/OS/driver related - surprisingly enough. Had hoped it was just a quirky texture :-(

To summarize some more bits on chat

* Could reproduce on multiple Win10 systems with Nvidia cards (including with the same 1103 binary that didn't show it for Niru on Linux)
* Niru's system for the test was Radeon on Linux - did not show the issue
* @DarkWeird saw it on Linux + intel chip and might have a bit more insight - " i run git bash with nvidia and run ts from it.. and it is works! We have problem with hybrid graphics on windows"

DarkWeird:
Nvidia window 10 - contains issue
Intel windows 10 - not contains issue.

DarkWeird:
Another notebook:
Nvidia(proprietary driver) linux -  contains issue.
Intel linux - not contains issue.

DarkWeird:
I have idea about it issue.

I checked atlas with crosshair.
![gui atlas texture](https://raw.githubusercontent.com/MovingBlocks/Terasology/develop/engine/src/main/resources/assets/textures/gui.png)
See on on crosshair and texture on left from it. 
Left texture have border with color equals to crosshair artifact color.

This is issue can be problem with math rounding or texture interolation at driver/gpu level.

Suggest to check Atlas offset calculation

skaldarnar:
I did a brief round of investigation today:

- artifact is present with nvidia (linux/windows)
- everything is fine with Intel(R) HD Graphics 530 (macbook)

[...]

PR #4232, opened 2020-11-02
chore: change texture file for 'engine:gui' atlas (crosshair-only)

skaldarnar:
Related to #4180 , but does not fix it.

Most parts of the old `gui.png` were not used at all, so I'm moving the crosshair part into a smaller texture file only containing crosshair-related textures.

crosshair-artefact
_You can still see the artefact on the left of the charge bar (red pixels)._


engine/src/main/resources/assets/atlas/gui.atlas
```diff
@@ -1,13 +1,12 @@
...
+    "textureSize": [220, 22],
     "grid": {
         "tileSize": [22, 22],
-        "gridDimensions": [9, 1],
-        "gridOffset": [24, 23],
+        "gridDimensions": [10, 1],
         "tileNames": [
             "crosshair", "crosshairCharge1", "crosshairCharge2", "crosshairCharge3", "crosshairCharge4",
-            "crosshairCharge5", "crosshairCharge6", "crosshairCharge7", "crosshairCharge8"
+            "crosshairCharge5", "crosshairCharge6", "crosshairCharge7", "crosshairCharge8", ""
```
DarkWeird, line 9: why we have empty tile?

DarkWeird (approved):

engine/src/main/resources/assets/atlas/gui.atlas
skaldarnar: Because I made the texture file 220x22 with 10 tiles... I could also go for 198x22 and have only 9 tiles :shrug:

PR #4233, opened 2020-11-03
feature(subsystems): extract DiscordRPCSubSystem

DarkWeird:
### Contains

New gradle-modules folder - `subsystems`

Extracted DiscordRPC subsystem - used by `PCFacade` only.

DiscordRPC was used in PlayerConfig directly - implemented config property observer. (like in rendering configs for runtime graphics settings changing)

### How to test

Required:
* Discord desktop

TestCases:
1. run game.
2. enable DIscord in settings
3. start game 
4. check status in Discord
5. rename your player name
6. check changes in discord (seems)

- [ ] improve subproject.settings.gradle and convert it in kotlin
- [ ] make configs composite( migrate to FlexableConfigs) - for removing discord config property from engine
- [x] (can be my dirty workspace) fix error with logs


engine/src/main/java/org/terasology/config/PlayerConfig.java
```diff
@@ -16,15 +16,18 @@
...
-import org.terasology.engine.subsystem.rpc.DiscordRPCSubSystem;
 import org.terasology.nui.Color;
 import org.terasology.rendering.nui.layers.mainMenu.settings.CieCamColors;
 import org.terasology.utilities.random.FastRandom;
 import org.terasology.utilities.random.Random;
+import org.terasology.utilities.subscribables.AbstractSubscribable;
+
+import java.util.List;
+
+public class PlayerConfig extends AbstractSubscribable {
```
skaldarnar, line 27: I'd like to take the chance to put some docstring here. Will take a closer look at this file to answer the following questions:
- what does it do? what kind of data does it hold?
- how does subscription work? how should that be used?
- any general notes?

@DarkWeird feel free to share some thoughts on this, I'll take care of making the Englishz fancy 🤓 🙃

subsystems/DiscordRPC/build.gradle.kts
```diff
@@ -0,0 +1,12 @@
+plugins {
+    java
+    `java-library`
```
skaldarnar, line 3: Gradle confuses me 🧐  Do you use backticks, single quotes, or double quotes here in the plugin section?

subsystems/DiscordRPC/build.gradle.kts
```diff
@@ -0,0 +1,12 @@
+plugins {
```
skaldarnar, line 1: Let's make sure that there is a README in `subsystems/DiscordRPC` to explain how this works and how it is supposed to be used. Maybe we should even have a README in the `subsystems` folder itself to explain the separation. Would be a good place to the diagrams you shared in Discord, too 😉

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCSubSystem.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.terasology.engine.subsystem.rpc;
+package org.terasology.engine.subsystem.discordrpc;
```
skaldarnar, line 16: Should this package name still have `engine` in it?
```suggestion
package org.terasology.subsystem.discordrpc;
```

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCSystem.java
```diff
@@ -13,9 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
```
skaldarnar, line 14: Let's update all the license headers in the extracted subsystem to use the shorter two-line license.

subsystems/subprojects.settings.gradle
```diff
@@ -0,0 +1,13 @@
+// This magically allows subdirs in this subproject to themselves become sub-subprojects in a proper tree structure
+new File(rootDir, 'subsystems').eachDir { possibleSubprojectDir ->
+    if (!possibleSubprojectDir.name.startsWith(".")) {
+        def subprojectName = 'subsystems:' + possibleSubprojectDir.name
+        println "Processing facade $subprojectName, including it as a sub-project"
```
skaldarnar, line 5:
```suggestion
        println "Including '$subprojectName' as a sub-project"
```

subsystems/subprojects.settings.gradle
```diff
@@ -0,0 +1,13 @@
...
+        include subprojectName
+        def subprojectPath = ':' + subprojectName
+        def subproject = project(subprojectPath)
+        subproject.projectDir = possibleSubprojectDir
+    } else {
+        println "Ignoring entry $possibleSubprojectDir as it starts with ."
```
skaldarnar, line 11:
```suggestion
        println "Ignoring hidden folder '$possibleSubprojectDir'"
```

skaldarnar (changes requested):
I really appreciate this small PR, makes reviewing so much fun 😊  Changes look very reasonable to me, I haven't tested it yet, though.

As written in the comments, I would like to take the opportunity to improve our docstrings here. I'll likely make some commits for that directly on this branch 🤓

subsystems/DiscordRPC/build.gradle.kts
DarkWeird: It is kotlin.
You can write methods with spaces too(useful for test names).

skaldarnar: 💡

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCSubSystem.java
DarkWeird: it is still part of engine yet.

skaldarnar: as in: still in the _engine repo_? or still technically in the engine? I don't mind too much, was just wondering...

skaldarnar (approved):

skaldarnar:
Successfully tested this on my Mac. It didn't work on my Linux Mint system (neither before nor after the extraction to a subsystem, so I suspect this is a different issue).

facades/PC/build.gradle.kts
```diff
@@ -111,6 +101,22 @@ configurations {
...
+        description = "for fetching modules for running a server"
+    }
+    implementation {
+        extendsFrom(configurations["subsystem"])
+    }
+}
+
+dependencies {
+    implementation(project(":engine"))
+    "subsystem"(project(":subsystems:DiscordRPC"))
```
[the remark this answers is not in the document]
DarkWeird: Yeah, it is configuration.
In groovy i can omit `"`. 
But it is kotlin. and this configuration from this file, not from plugins.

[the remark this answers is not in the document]
keturn: It would be nice to complete #4157 before introducing new complications to the facade build.

I think I would have stuck with the built-in `implementation` Configuration for the subsystems here instead of making a custom Configuration for them.

I can start to imagine why it might be nice to separate them. For example, if we had a lot of Subsystems, and a bunch of other direct dependencies, and we had some troubleshooting use case where we needed to see the Subsystem dependencies independently and there was a nice diagnostic report for that —

but I don't expect that to be the case, and I'm concerned it's not worth the added complexity to the build file and the loss of edit-time type-checking on the custom Configuration name. Are there other benefits to it I have overlooked?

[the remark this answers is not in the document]
DarkWeird: I don't think that picocli will produce many conflicts.

New configuration introduced for classpath at run tasks (you made building classpath from engine, it's dependencies and facade )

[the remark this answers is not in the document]
keturn: > you made building classpath from engine, it's dependencies and facade

I didn't! But yes my name is probably on all the lines because of the change to `.kts`.

> I don't think that picocli will produce many conflicts.

Hmm, I see I was unclear when I referred to picocli's issue. The bug that is blocking the picocli PR is that picocli isn't included in the runtime classpath of the `game` task (and others). That is why I have been asking all [these](https://github.com/MovingBlocks/Terasology/pull/4157#issuecomment-704658753) [questions](https://github.com/MovingBlocks/Terasology/pull/4157#issuecomment-716830609) about why the heck those tasks are written the way they are.

But this is not really in the scope of picocli, and I believe you have run in to the same problem here, let us separate it to another issue: #4238 

My objection is that these tasks already have a bunch of configuration in them that nobody seems to be able to explain, and this change makes the facade build configuration more complicated, not less.

So it's not that this conflicts with picocli, exactly. It conflicts with trying to fix the bug-that-blocks-picocli.

subsystems/DiscordRPC/build.gradle.kts
```diff
@@ -0,0 +1,12 @@
...
+    java
+    `java-library`
+}
+
+apply(from = "$rootDir/config/gradle/publish.gradle")
```
keturn, line 6: What does this need from `publish.gradle`?

do we expect subsystems to be published as independent artifacts to the jar repository?

DarkWeird: I think publishing is possible in future.

Really i can use common. Because i needs repositories from it

skaldarnar: Even though it might be possible in the future we should probably not include it right now. Using `common` might indeed be a better option.

skaldarnar (dismissed):
@keturn are you okay with merging this? https://github.com/MovingBlocks/Terasology/pull/4157 is still a draft PR and I don't know whether your questions have been sufficiently answered. We can put that topic on the agenda for the next org meeting to drive it forward.

If it is not too much of an issue that this PR introduces a new configuration for subsystem I'd like to merge this before it gets stale or we forget about it...

subsystems/DiscordRPC/build.gradle.kts
keturn: Ah, okay. Maybe still something to review later, but `common` doesn't worry me as much as `publish`. 👍

keturn:
> @keturn are you okay with merging this? #4157 is still a draft PR

See https://github.com/MovingBlocks/Terasology/pull/4233#discussion_r519235302

In short, yes I am concerned about the addition of the Configuration. Merging #4157 itself isn't necessary; #4238 is a more focused issue relevant to the facade build config without the other picocli distractions.

DarkWeird:
@skaldarnar 
suggest merge this and i will take  #4238 as next task

skaldarnar (approved):

PR #4235, opened 2020-11-06
feature(autoconfig): migrate SystemConfig

DarkWeird:
### Contains

Migration SystemConfig to AutoConfig

Must merge with (https://github.com/Terasology/ModuleTestingEnvironment/pull/28)

### How to test
1. start game
2. change system settings
3. restart game(with exit from game)
4. Should works fine, without warnings about configs and settings must changed after restart.


engine/src/main/java/org/terasology/config/SystemConfig.java
```diff
@@ -16,99 +16,63 @@
...
-        this.locale = locale.toLanguageTag();
-    }
+    public final Setting<Long> dayNightLengthInMs = setting(
+            type(Long.class),
+            defaultValue(1800000L),
+            constraint(new NumberRangeConstraint<>(0L, Long.MAX_VALUE, false, false))
+    );
+    public final Setting<Integer> maxThreads = setting(
+            type(Integer.class),
+            defaultValue(Runtime.getRuntime().availableProcessors() - 1),
```
skaldarnar, line 41: Where do these default values come from?

engine/src/main/java/org/terasology/config/SystemConfig.java
```diff
@@ -16,99 +16,63 @@
...
+    );
+    public final Setting<Integer> maxSecondsBetweenSaves = setting(
+            type(Integer.class),
+            defaultValue(60),
+            constraint(new NumberRangeConstraint<>(0, Integer.MAX_VALUE, false, false))
+    );
+    public final Setting<Integer> maxUnloadedChunksPercentageTillSave = setting(
+            type(Integer.class),
+            defaultValue(40),
+            constraint(new NumberRangeConstraint<>(0, Integer.MAX_VALUE, false, false))
```
skaldarnar, line 52: If this is about percentage, shouldn't the range constraint be [0...100]?
```suggestion
            constraint(new NumberRangeConstraint<>(0, 100, false, false))
```

engine/src/main/java/org/terasology/config/SystemConfig.java
```diff
@@ -16,99 +16,63 @@
...
+    );
+    public final Setting<Boolean> debugEnabled = setting(
+            type(Boolean.class),
+            defaultValue(true)
```
skaldarnar, line 56: I think this should be false - normal players should not see the debug info by default. (Assuming this is about the in-game debug view...)

skaldarnar (commented):
Looks good overall - @eviltak would be happy about this :smile: 

Did you test this with an Omega workspace? Are there module changes required?

engine/src/main/java/org/terasology/config/SystemConfig.java
DarkWeird: This value from recommendations for threading(Google) and i saw it in issue about FlexableConfig.
(Default value in config is 2)

All other value from config.

engine/src/main/java/org/terasology/config/SystemConfig.java
DarkWeird: Oh yeah... 
I  enable it on fresh install always

DarkWeird:
> Did you test this with an Omega workspace? Are there module changes required?

Yeah this changes applied with omega workspace.

Affects only MTE module.

Compiles, runs.

engine/src/main/java/org/terasology/config/SystemConfig.java
```diff
@@ -16,99 +16,63 @@
...
-    public boolean isMonitoringEnabled() {
-        return monitoringEnabled;
-    }
-
-    public void setMonitoringEnabled(boolean monitoringEnabled) {
-        this.monitoringEnabled = monitoringEnabled;
-    }
-
-    public boolean isWriteSaveGamesEnabled() {
-        String property = System.getProperty(SAVED_GAMES_ENABLED_PROPERTY);
```
eviltak, line 85: I don't know how important the overriding of this setting is, but if this is something we encounter frequently it might be worth looking into adding an overriding feature into `AutoConfig` or the setting UI. For now we might get away with setting this as the default value of the `Setting`, but we'll need expert opinion on how this case should be handled.

eviltak (changes requested):
Looks great to me! As @skaldarnar said, I am really happy to see this, thanks for the work @DarkWeird! This PR also helps identify some potential improvements, like adding a DSL of some sort for `NumberRangeConstraint` as well.

engine/src/main/java/org/terasology/config/SystemConfig.java
DarkWeird: I have some code about this locally.
Extends DSL with:
systemProperty(String, Function<String,V>)
Where function - converter,
V - settings type

DarkWeird:
Add SystemProperty value source for settings.

If SystemProperty presents:
    Set value inhibited with warn.
    Get value returns SystemProperty.

DarkWeird:
@eviltak 
Is it better now?

pollend:
@DarkWeird I rebased these changes. what is the state of this and should this be considered for merging.

engine-tests/src/test/java/org/terasology/config/flexible/internal/SettingImplTest.java
```diff
@@ -37,40 +24,81 @@
...
+
+        @Test
+        void testSystemPropertyValueNotPresent() {
+            System.getProperties().remove(TEST_CONFIG_SYSTEM_PROPERTY);
+            assertEquals(50, setting.get());
+        }
+
+        @Test
+        void testDoesNotSetValue() {
+            assertFalse(setting.set(25));
```
pollend, line 94: isn't this a valid settings change. what are you trying to test @DarkWeird?

pollend (changes requested):

engine/src/main/java/org/terasology/config/flexible/internal/SettingBuilder.java
```diff
@@ -66,8 +55,16 @@
     SettingBuilder<T> description(String description);
 
     /**
-     * Builds the {@link Setting} with the components that have already been specified and
-     * returns it.
+     * Specfies the system property of the {@link Setting} being created.
+     *
+     * @param systemProperty This system property which will rewrite value.
+     * @param converter converter for systemProperty: String -> T
```
skaldarnar, line 61: **nitpick (non-blocking):** To me _converter_ often implies that the conversion can work in both directions, and that it can be between arbitrary types. In this case (converting from `String -> T`) I'd call it a _decoder_ function (the dual function `T -> String` would be the encoder).

engine/src/main/java/org/terasology/config/flexible/internal/SettingImpl.java
```diff
@@ -137,13 +143,23 @@ public T getDefaultValue() {
 
     @Override
     public T get() {
+        T systemPropertyValue = getSystemPropertyValue();
+        if (systemPropertyValue != null) {
+            return systemPropertyValue;
+        }
```
skaldarnar, line 149: **thought** Should we have a (private) method to check whether this setting is a system property? This might make the semantics of this method a bit more obvious:

```java
if (isSystemProperty()) {
  return getSystemPropertyValue();
} else {
	return value;
}
```

Is it intended that we return the `value` in case the system property converter yielded a `null `value here?

DarkWeird: No. System property should rewrite value only.

engine/src/main/java/org/terasology/config/flexible/AutoConfigManager.java
```diff
@@ -48,6 +50,7 @@
 
     public AutoConfigManager(TypeHandlerLibrary typeHandlerLibrary) {
         this.typeHandlerLibrary = typeHandlerLibrary;
+        typeHandlerLibrary.addTypeHandler(Locale.class, new LocaleTypeHandler());
```
eviltak, line 53: Since `Locale` is a SDK class, can we add this type handler by default to every type handler library (and hence move it into one of the initialization methods in `TypeHandlerLibrary`)?

engine/src/main/java/org/terasology/config/flexible/typehandling/LocaleTypeHandler.java
```diff
@@ -0,0 +1,23 @@
...
+package org.terasology.config.flexible.typehandling;
+
+import org.terasology.persistence.typeHandling.StringRepresentationTypeHandler;
+
+import java.util.Locale;
+
+/**
+ * Locale type handler for {@link org.terasology.config.SystemConfig#locale}
+ */
+public class LocaleTypeHandler extends StringRepresentationTypeHandler<Locale> {
```
eviltak, line 13: Similar to above -- since `Locale` is a SDK class, it might be better to move this to a subpackage in `org.terasology.persistence.typeHandling`.

eviltak (changes requested):
I think adding system property support to settings is a good idea. I do think that we can go a step further by not limiting ourselves to system properties only and adding generic "override" support. Rather than store a system property name and a converter, I propose that we make settings store a single `override`, which (for a `Setting<T>`) is simply a closure of the form `Supplier<Optional<T>>` (i.e. `() -> Option<T>`). If `override` returns an empty `Optional`, then we simply use the value stored in the setting. Otherwise, we "override" the internally stored value with the one returned by `override`.

With this `override`, `SettingImpl.get()` can be implemented as a one-liner of the form `this.override.get().orElse(value)`. Assuming that we add a corresponding `SettingArgument.override` factory method, the DSL to override a setting with a system property would look as follows:

```java
public final Setting<Boolean> writeSaveGamesEnabled = setting(
        type(Boolean.class),
        defaultValue(true),
        override(() -> Optional.ofNullable(System.getProperty(SAVED_GAMES_ENABLED_PROPERTY))
			.map(Boolean::parseBoolean)
);
```

It's a bit more verbose, but it allows for us to support the use of other overrides, such as the presence of a command line argument, environment variable, etc.

engine/src/main/java/org/terasology/config/flexible/typehandling/LocaleTypeHandler.java
eviltak: `* Serializes objects of type {@link java.util.Locale}` --  no need to reference the particular use case for which this type handler was created.

DarkWeird:
> I think adding system property support to settings is a good idea. I do however think that we can go a step further by not limiting ourselves to system properties only and adding generic "override" support. Rather than store a system property name and a converter, I propose that we make settings store a single `override`, which (for a `Setting<T>`) is simply a closure of the form `Supplier<Optional<T>>` (i.e. `() -> Option<T>`). If `override` returns an empty `Optional`, then we simply use the value stored in the setting. Otherwise, we "override" the internally stored value with the one returned by `override`.
> 
> With this `override`, `SettingImpl.get()` can be implemented as a one-liner of the form `this.override.get().orElse(this.value)`. Assuming that we add a corresponding `SettingArgument.override` factory method, the DSL to override a setting with a system property would look as follows:
> 
> ```java
> public final Setting<Boolean> writeSaveGamesEnabled = setting(
>         type(Boolean.class),
>         defaultValue(true),
>         override(() -> Optional.ofNullable(System.getProperty(SAVED_GAMES_ENABLED_PROPERTY))
> 			.map(Boolean::parseBoolean))
> );
> ```
> 
> It's a bit more verbose, but it allows for us to support the use of other overrides, such as the presence of a command line argument, environment variable, etc.

Nice idea! how i didn't think about this?

engine/src/main/java/org/terasology/config/flexible/internal/SettingBuilder.java
DarkWeird: I don't like idea to write something in sysProps at runtime.
It are going to problems :D I saw systems, where general config source - is  sysProps. 
then we catch problem like CoreRegistry + parallel tests.. it was worst.

engine/src/main/java/org/terasology/config/flexible/internal/SettingBuilder.java
```diff
@@ -55,13 +56,11 @@
     SettingBuilder<T> description(String description);
 
     /**
-     * Specfies the system property of the {@link Setting} being created.
-     *
-     * @param systemProperty This system property which will rewrite value.
-     * @param converter converter for systemProperty: String -> T
-     * @return The builder object.
+     * Specfies the override f the {@link Setting} being created.
```
eviltak, line 59: Spelling: `of` instead of `f`

eviltak (changes requested):
Looks good to me (besides the one character fix)! If we have confirmation that `SystemConfig` works in practice, we can go ahead and merge. This will break existing configs, so do we want to come up with some sort of migration tool or should we just include this change in the next major release?

DarkWeird:
> Looks good to me (besides the one character fix)! If we have confirmation that `SystemConfig` works in practice, we can go ahead and merge. This will break existing configs, so do we want to come up with some sort of migration tool or should we just include this change in the next major release?

This configs not changed by hands. defaults values is equals to `default.cfg`'s values.

pollend (approved):

jdrueckert (approved):

engine/src/main/java/org/terasology/config/SystemConfig.java
keturn: On single-core systems, this comes out to zero. :boom:

DarkWeird: Ok. Will fixed.... Where are you found single core? VM?

PR #4241, opened 2020-11-09
Fix not killed process.

### LINKED ISSUE #4240: Game process does not finish
When exiting the game some thread is still running and does not finish/exit. This leaves the game process in a weird state, where the game windows is already closed, but the process is still running indefinitely. I can observe this locally when running from source, but this also happens when starting releases via the launcher (see https://github.com/MovingBlocks/TerasologyLauncher/issues/584). 

@DarkWeird already investigated this a bit and tracked it down to being `Thread-8` - whatever that means 😅

DarkWeird:
### Contains

Fixes #4240

### How to test

1. Run game
2. create/load world
3. Exit from TS.
4. Game process should finished.


skaldarnar (approved):
Tested on my Linux Mint, works fine!

PR #4245, opened 2020-11-12
Extract  "run game" task configurations

### LINKED ISSUE #4238: game tasks should include facade's dependencies on the runtime classpath

Add a dependency to facades/PC, i.e. `picocli` in #4157.

### What actually happened

Gradle resolves the dependency and it compiles correctly, but when I go to start the game with `gradlew game` (`:facades:PC:game`), it cannot find the dependency in its classpath.

### How to reproduce

see e.g. the current state of #4157's branch

### Implementation thoughts

There is a lot of duplication between the various JavaExec tasks in facades/PC's gradle file. To make it more maintainable and improve confidence that when we fix this issue, we fix it for _all_ those tasks, it would be helpful to refactor or change the implementation to not require so many duplicated lines in them.

DarkWeird:
### Contains

Reworking `classpath`, `args`, `jvmargs` handlers for game run tasks.
Extract common part of game run tasks.
Also direct dependencies in classpath on running now.

Fixes #4238

### How to test

1. Run every run game task. works like should. (except profile on liberica JDK 11)

2. Run  run game task on MacOS

### DIFF
--- a/facades/PC/build.gradle.kts
+++ b/facades/PC/build.gradle.kts
@@ -111,7 +111,21 @@ configurations {
     }
 }
 
+// Used for all game configs.
+val commonConfigure : JavaExec.()-> Unit = {
+    dependsOn(":extractNatives")
+    dependsOn(":moduleClasses")
+    dependsOn("classes")
+
+    // Run arguments
+    main = mainClassName
+    workingDir = rootDir
+
+    classpath(sourceSets["main"].runtimeClasspath)
+}
+
 tasks.register<JavaExec>("game") {
+    commonConfigure()
     description = "Run 'Terasology' to play the game as a standard PC application"
     group = "terasology run"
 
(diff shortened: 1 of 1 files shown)


facades/PC/build.gradle.kts
```diff
@@ -129,109 +123,45 @@ tasks.register<JavaExec>("game") {
     main = mainClassName
     workingDir = rootDir
 
+    classpath(sourceSets["main"].runtimeClasspath)
```
skaldarnar, line 126: so that's ... all .. we need?

DarkWeird: Yeah.
Facade classes + dependencies.

facades/PC/build.gradle.kts
```diff
@@ -111,16 +111,10 @@ configurations {
     }
 }
 
-tasks.register<JavaExec>("game") {
-    description = "Run 'Terasology' to play the game as a standard PC application"
+// Used for all game configs.
+val commonConfigure : JavaExec.()-> Unit = {
```
keturn, line 115: Use function-definition syntax here instead of assigning an anonymous function to this variable?

```suggestion
fun JavaExec.commonConfigure(): Unit {
```

I _think_ that's what you were going for. I'm a little rusty.

facades/PC/build.gradle.kts
keturn: and I expect this is the default classpath of the `run` task.

What I don't know is how to take advantage of those defaults for making and customizing several different tasks. Last time I looked, `run` wasn't its own Class of task, it was a customized JavaExec a lot like this.

Either way, I think this is a big improvement! We can update it later if we figure out how bring it closer to defaults, and we won't have to worry about finding all the duplicates if/when we do.

keturn (approved):
I haven't had a chance to test this, but I approve from a code-maintainability perspective. 👍

facades/PC/build.gradle.kts
DarkWeird: Yeah. This is default parameter in unknown place.


JavaExec have default as null there.
(It can be used for run external java code)

skaldarnar (approved):
I just tested this within my Iota workspace after doing a `git clean -fdx`.

The game starts up fine with 
```
gradlew clean jar game
```

Issue #4243, opened 2020-11-11
Investigate / Fix closing of AFK screen


jdrueckert:
After https://github.com/MovingBlocks/Terasology/pull/3571 was merged, the AFK mode is triggered as expected and the AFK screen opens after some time of doing nothing.

Whenever I faced this, the screen displayed "Press anything to disable the AFK mode", however whatever I pressed, it didn't close. I was able to walk around, but everything else (looking around, switching inventory slots, opening the inventory) didn't work. Only when I pressed `Esc` the screen closed.

On checking the PR I found that [the screen string is set depending on the values of `afkTime` in comparison to `AfkClientSystem.AFK_FREEDOM`](https://github.com/MovingBlocks/Terasology/pull/3571/files#diff-db5b3ce9480f76496c06faa0f0ac09f65001884c679573385af7daf166af20a5R52). If `afkTime` is smaller or equal `AfkClientSystem.AFK_FREEDOM` it shows "ESCAPE", otherwise "anything".

[`AfkClientSystem.AFK_FREEDOM` is a constant set to `5 * 1000`](https://github.com/MovingBlocks/Terasology/pull/3571/files#diff-94b175a07fc04ef3ec8b9e60680d78b53bb0741eb75aa77e7adf8f0bae00d43eR31).
In my tests, `afkTime` was always higher and usually at around 60000, which makes me wonder whether maybe it should be compared to [`AfkClientSystem.AFK_TIMEOUT`, a constant set to `60 * 1000`](https://github.com/MovingBlocks/Terasology/pull/3571/files#diff-94b175a07fc04ef3ec8b9e60680d78b53bb0741eb75aa77e7adf8f0bae00d43eR34) instead.

As a quick fix, I simply [inverted the comparison](https://github.com/MovingBlocks/Terasology/commit/f6d87a6a50a45c967d9249383ed13f782120bff5).

@iHDeveloper any information on the following questions would be much appreciated:

1. Why do we differentiate between "ESCAPE" and "anything" here?
1. What is the semantic of the time ranges `t <= AfkClientSystem.AFK_FREEDOM`, `AfkClientSystem.AFK_FREEDOM < t <= AfkClientSystem.AFK_TIMEOUT` and `AfkClientSystem.AFK_TIMEOUT < t`?

jdrueckert:
FYI @skaldarnar

jdrueckert:
Closed by #4248

PR #4248, opened 2020-11-14
fix: remove afk screen

pollend:
removed the afk screen and moved the console messages to logging instead. having an popup show up every minute or so is kind of obnoxious. probably just enough to mark the player as afk rather then also notify them that they are afk. I don't think there is really a need to convey this to player in the first place. in single player this popup shows up constantly so that needs to be addressed as well. it would need to be disabled in singe player since afk is not really relevant in this case. mostly to signal other players that the given player is not available to interact with. 

ref: https://github.com/MovingBlocks/Terasology/pull/3571


engine/src/main/java/org/terasology/logic/afk/AfkClientSystem.java
```diff
@@ -77,14 +79,11 @@ public void onCommand() {
         component.afk = !component.afk;
         entity.addOrSaveComponent(component);
         if (component.afk) {
-            nuiManager.pushScreen(SCREEN_URL, AfkScreen.class).setAfkClientSystem(this);
-            nuiManager.closeScreen(CONSOLE_SCREEN_URL);
             enableDiscord();
-            console.addMessage("[AFK] You are AFK now!");
+            logger.info("[AFK] You are AFK now!");
```
skaldarnar, line 83: I think some form of confirmation would be nice... Console is the text box in the bottom left, right? I think we could also print to the dev console by returning a String from this command method.

engine/src/main/java/org/terasology/logic/afk/AfkClientSystem.java
```diff
@@ -77,14 +79,11 @@ public void onCommand() {
...
         } else {
-            nuiManager.closeScreen(SCREEN_URL);
             disableDiscord();
-            console.addMessage("[AFK] You are no longer AFK!");
+            logger.info("[AFK] You are no longer AFK!");
```
skaldarnar, line 86: Same here, application logs are not visible to the player.

skaldarnar (changes requested):

skaldarnar (approved):

jdrueckert:
Closes #4243

pollend:
is there any other changes we want to make @skaldarnar.

skaldarnar (approved):

PR #4249, opened 2020-11-15
feat(JOML): migrate LocalPlayerSystem

pollend:
Here is a migration of the localplayersystem. make sure you can log in a the camera behaves correctly both in single and multiplayer. I think there is some aspects for the Camera that can be improved. we should make a better distinction for get and set operators for the camera especially for quaternions. we have some cases where we getOrientation() and also apply a transform. Other then that I think this is fairly straightforward change.


jdrueckert:
Singleplayer works fine, on multiplayer though the camera of the joining player is shaking horribly (and constantly) :thinking:

pollend:
not sure I can't reproduce it?

jdrueckert:
@pollend probably not related to your change after all. when playing JS this week, this happened to me without your PR checked out.
We'll need to investigate this, but it shouldn't block your PR.

jdrueckert (approved):

PR #4250, opened 2020-11-16
feat(rendering): Adjust minimal amount of daylight at night

skaldarnar:
I'm not sure whether this is just fixing symptoms or the actual root cause... Any input is appreciated. I played around with different values, see screenshot comparison below. The overall brightness is heavily influenced by the combination of video settings, so it's hard to tell whether we can find a one-size-fits-all value.

terasology_night-brightness

I also found that we decide on whether to use the "night exposure" values in CoreRendering based on whether the daylight value is 0. This might need adjustment with this change (as daylight will never be 0 afterwards).

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/backdrop/Skysphere.java
+++ b/engine/src/main/java/org/terasology/rendering/backdrop/Skysphere.java
@@ -107,7 +107,7 @@ public class Skysphere implements BackdropProvider, BackdropRenderer {
         float daylight = 1.0f;
 
         if (angle < 24.0f) {
-            daylight = 1.0f - (24.0f - angle) / 24.0f;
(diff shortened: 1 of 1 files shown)


4Denthusiast:
We should get rid of that night exposure thing anyway (setting the current night settings to be on all the time). It looks silly when you're underground and suddenly all of your torches get brighter just because the sun set outside.

4Denthusiast:
If I have eye adaptation turned off, I think 0.25 looks best. If I have it turned on, 0.15 seems like plenty. I think that we should take eye adaptation on to be the default, but I actually had it disabled before testing this, because it looks bad in some cases. I'm not sure whether actually making the night brightness higher when eye adaptation is off would be a good idea. Anyway, this PR clearly improves things, so I approve it.

4Denthusiast (approved):

PR #4253, opened 2020-11-17
feat: start first game day at sunrise

jdrueckert:
- thus the night won't seem that long and the player has more time to prepare for it


engine/src/main/java/org/terasology/world/time/WorldTime.java
```diff
@@ -39,6 +39,11 @@
      */
     float NOON_OFFSET = 0.50f;
 
+    /**
+     * The offset used by the game to start at sunrise
+     */
+    float SUNRISE_OFFSET = 0.3f;
```
skaldarnar, line 45: Could you please add an analogous `SUNSET_OFFSET`, please?

skaldarnar (approved):

Issue #4135, opened 2020-08-23
Harden against "Cannot set a changed component with a null component variable"


Cervator:
### What you were trying to do

Various things - this is a somewhat common error

One replication routine: while testing #4131 I would fly up in the air with WildAnimals enabled and place a deer (`spawnPrefab deer`) perilously close to an edge of some sort, be it a handy tree or a constructed plank, then undermine said support so the poor deer falls to its demise. This would crash a server without #4131 merged. With that fix in place I'd just expect a nice calm and absolutely humane and painless deer death.

### What actually happened

Get an angry stacktrace probably as an involved entity, event, or scheduled regen tick tries to do something past a point where they probably aren't relevant anymore

My closest suspicion would be that some sort of entity cleanup isn't fully happening. But as noted we've seen this regularly in different cases so it is hard to say if some hardened code or improved cleanup would resolve it. But at least there's an easy replicate scenario. Well, easy, unless your heart breaks at the thought of mercilessly sending a deer to the great beyond.

### How to reproduce

* Run game with a Perlin world and WildAnimals active (may have to be hosting multiplayer? But no second client needed)
* Put a deer on an edge high above the ground somehow (find/make a good spot then point right above it and `spawnPrefab deer`)
* Dig out blocks the peaceful innocent little deer is standing on, you monster!
* Watch the log for errors and heartbreak

### Log details and game version

I didn't notice at first that this may have spanned an auto-save trigger? Maybe that's involved somehow?

```
2:39:22.712 [main] INFO  o.t.p.i.ReadWriteStorageManager - Auto Saving - Creating game snapshot
22:39:22.725 [main] INFO  o.t.p.i.ReadWriteStorageManager - Auto Saving - Snapshot created: Writing phase starts
22:39:22.870 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
java.lang.RuntimeException: Cannot set a changed component with a null component variable
	at org.terasology.persistence.internal.EntityDelta.setChangedComponent(EntityDelta.java:34)
	at org.terasology.persistence.internal.EntitySetDeltaRecorder.onEntityComponentChange(EntitySetDeltaRecorder.java:70)
	at org.terasology.persistence.internal.EntitySetDeltaRecorder.onEntityComponentAdded(EntitySetDeltaRecorder.java:61)
	at org.terasology.persistence.internal.ReadWriteStorageManager.onEntityComponentAdded(ReadWriteStorageManager.java:514)
[...]

PR #4254, opened 2020-11-17
fix: reverse order of notification in PojoEntityManager

skaldarnar:
Possibly a fix for #4135 

**Problem:** The `PojoEntityManager` handles the life cycle of components added to and removed from entities. It notifies two kinds of subscribers about these changes, (1) registered systems receive `OnAddedComponent`/`OnActivatedComponent` or `OnChangedComponent`, and (2) internal systems (such as the delta recorder) are directly notified. 
The error of #4135 appears if the component that was just added is directly removed in reaction to any of the life cycle events mentioned in (1).

**Solution:** I assume that the internal systems that get directly notified won't change the state of the entity but are read-only system (e.g., for storing the state, or recording the delta info). Therefore, I think it is safe to notify those before sending events via the event system.

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityManager.java
+++ b/engine/src/main/java/org/terasology/entitySystem/entity/internal/PojoEntityManager.java
@@ -552,6 +552,16 @@ public class PojoEntityManager implements EngineEntityManager {
         Preconditions.checkNotNull(component);
         Optional<Component> oldComponent = getPool(entityId).map(pool -> pool.getComponentStore().put(entityId, component));
 
+        // notify internal users first to get the unobstructed views on the entity as it is at this moment.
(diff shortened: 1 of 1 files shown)


pollend (approved):

Cervator (approved):
Tested this out against the replication notes in #4135 and can confirm the (benign?) stacktrace no longer happens, which seems to be a good sign. Wonder how other ways to trigger it will fare. Code change seems reasonable, good spelunking work.

Hard to guess at whether this'll impact anything unexpected, after years of various things being coded with various assumptions - one day it'd be great if we could have unit tests covering the actual expected lifecycle cases and so on, but that's probably a post-gestalt-reintegration task.

Going to merge this in the hope that it will do good things and if anything bad happens we'll catch it later 👍

PR #4256, opened 2020-11-20
fix: InputSystem#getInputsForBindButton was throwing NPE

skaldarnar:
I think the only usage of that API wrapper method (in Omega) is in Dialogs where it is used to get the associated bind for an action to display it as a hint to the user.

https://github.com/Terasology/Dialogs/blob/5a0073966afc66f18bc1d7220c7b49c8c053d9c2/src/main/java/org/terasology/dialogs/DialogSystem.java#L190

This was throwing a NPE for a while, resulting in the hint not showing up. This PR fixes that.

### DIFF
--- a/engine/src/main/java/org/terasology/input/InputSystem.java
+++ b/engine/src/main/java/org/terasology/input/InputSystem.java
@@ -12,7 +12,6 @@ import org.terasology.engine.SimpleUri;
 import org.terasology.engine.Time;
 import org.terasology.engine.subsystem.DisplayDevice;
 import org.terasology.engine.subsystem.config.BindsManager;
-import org.terasology.engine.subsystem.config.BindsSubsystem;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.entitySystem.systems.RegisterSystem;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/input/InputSystem.java
```diff
@@ -554,13 +552,15 @@ public void drainQueues() {
...
+     * {@link org.terasology.config.BindsConfig#getBinds(SimpleUri)}.
+     * <p>
+     * TODO: Restored for API reasons, may be duplicating code elsewhere. Should be reviewed.
      *
      * @param bindId the ID.
      * @return a list of keyboard/mouse inputs that trigger the binding.
      */
     public List<Input> getInputsForBindButton(SimpleUri bindId) {
-        return bindsSubsystem.getInputsForBindButton(bindId);
+        return bindsManager.getBindsConfig().getBinds(bindId);
```
skaldarnar, line 564: This is the actual bug fix, all other changes are just IntelliJ's formatting...

engine/src/main/java/org/terasology/input/InputSystem.java
```diff
@@ -78,9 +77,6 @@
     @In
     private CameraTargetSystem targetSystem;
 
-    @In
-    private BindsSubsystem bindsSubsystem;
```
skaldarnar, line 82: The `bindsSubsystem` was `null` here anyways, and the only place this was used could easily be replaced with access to the super interface of `BindsManager`.

jdrueckert (approved):

PR #4258, opened 2020-11-21
feat(MutableComponentContainer): Handle 'null' values in [update|upsert]Component

skaldarnar:
Handle cases where the transformation/upsert function returns `null`. This was previously not well defined, now we state what the behavior is in the Javadoc and change the implementation to handle these cases gracefully.

**Update Component** I encountered a case where I wanted to update a component, and possibly remove it in the end if some condition is met. I thought returning null in this case might be a good way to indicate that this component should be removed from the container.

**Upsert Component** Analog to the update case, returning a null value from the upsert function indicates that no component should be added to the container.

This is used already in https://github.com/Terasology/Notifications/tree/develop (if this PR get's rejected I'll adapt the module code).

### DIFF
--- a/engine/src/main/java/org/terasology/entitySystem/MutableComponentContainer.java
+++ b/engine/src/main/java/org/terasology/entitySystem/MutableComponentContainer.java
@@ -92,7 +92,7 @@ public interface MutableComponentContainer extends ComponentContainer {
      *     item.updateComponent(ItemComponent.class, MyItemHandler::updateItem);
      *     }
      * </pre>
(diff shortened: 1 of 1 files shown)


jdrueckert (approved):

DarkWeird (approved):
Look nice. Idea too.

jdrueckert:
`ci/cloudbees/checkstyle` result doesn't update.
[Restarted job](http://jenkins.terasology.io/teraorg/job/Terasology/job/engine/job/PR-4258/3/execution/node/26/log/) succeeded.

PR #4259, opened 2020-11-21
feat(JOML): migrate `InputSystem` and events

jdrueckert:


pollend (approved):

engine/src/main/java/org/terasology/input/internal/BindableButtonImpl.java
```diff
@@ -213,9 +212,9 @@ public void update(EntityRef[] inputEntities,
                 if (!consumed) {
                     buttonEvent.prepare(id, ButtonState.REPEAT, delta);
                     buttonEvent.setTargetInfo(target,
-                        targetBlockPos == null ? null : JomlUtil.from(targetBlockPos),
-                        hitPosition == null ? null : JomlUtil.from(hitPosition),
-                        hitNormal == null ? null : JomlUtil.from(hitNormal));
+                        targetBlockPos == null ? null : targetBlockPos,
+                        hitPosition == null ? null : hitPosition,
+                        hitNormal == null ? null : hitNormal);
```
pollend, line 217: I don't think you need the null check anymore. I did that to avoid a crash from JomlUtil.from.

pollend (changes requested):

pollend (approved):

PR #4263, opened 2020-11-21
feat(JOML): migrate block selection component

pollend:
This is just a blind migration of the block selection component and I'm not exactly sure how this fits into other modules. I only really see this used in MasterOfOreon. I do remember migrating some similar type of selection logic for StructureTemplates. wonder if we are duplicating some similar logic for two systems that would be better fit into a single setup. There is also the matter of migrating and verifying that this is correct with MasterOfOreon. 

- [ ] Behaviors
- [ ] Maze
- [ ] MasterOfOreon


engine/src/main/java/org/terasology/logic/selection/LocalPlayerBlockSelectionByItemSystem.java
```diff
@@ -93,13 +94,16 @@ public void onCamTargetChanged(CameraTargetChangedEvent event, EntityRef entity)
...
-
-            blockSelectionComponent.currentSelection = Region3i.createFromCenterExtents(new Vector3i(targetLocation.x, targetLocation.y, targetLocation.z),
-                    new Vector3i(region.sizeX()/2, 0, region.sizeZ()/2));
+            BlockRegion region = blockSelectionComponent.currentSelection;
+            blockSelectionComponent.currentSelection = new BlockRegion()
+                .union(
+                    Math.roundUsing(targetLocation.x, RoundingMode.FLOOR),
+                    Math.roundUsing(targetLocation.y, RoundingMode.FLOOR),
+                    Math.roundUsing(targetLocation.z, RoundingMode.FLOOR))
+                .addExtents(region.getSizeX() / 2, 0, region.getSizeZ() / 2);
```
jdrueckert, line 106:
```suggestion
            blockSelectionComponent.currentSelection = new BlockRegion(
                    Math.roundUsing(targetLocation.x, RoundingMode.FLOOR),
                    Math.roundUsing(targetLocation.y, RoundingMode.FLOOR),
                    Math.roundUsing(targetLocation.z, RoundingMode.FLOOR))
                .expand(region.getSizeX() / 2, 0, region.getSizeZ() / 2);
```

jdrueckert (changes requested):

skaldarnar (dismissed):

skaldarnar:
Superseded by https://github.com/MovingBlocks/Terasology/pull/4336

PR #4269, opened 2020-11-24
Drop creation of unnecessary array on new chunk event

hajdam:
### Contains

Optimization - There seems to be unnecessary creation of array for each call of new chunk methods.
While it's just 130kB of data per call and GC should be able to clean it simply with next call, coping there seems be not necessary. This could lead to spikes of memory use and slight delay when lot of new chunks is processed.

My experience with project is limited. There might be some reason to use array if getBlock must be called by Z-axis first, but I didn't recognize it.

### How to test

Check if chunks are loaded properly and at the same speed (or faster).


jdrueckert:
Looks valid from my point of view, but I'd like a second opinion from @pollend

DarkWeird:
I think better summon @4Denthusiast 
this is his code.

4Denthusiast (commented):
I actually just copied this from VoxelWorldSystem to VoxelLiquidWorldSystem. I don't know why it was done like this in the first place. The code changes look good. I haven't actually tested this, but assuming someone has done that (checking that block collision shapes still work as normal), I approve this.

jdrueckert (approved):
Tested out in `CoreSampleGameplay`, chunks still load fine :+1:

engine/src/main/java/org/terasology/physics/engine/VoxelLiquidWorldSystem.java
```diff
@@ -146,20 +146,15 @@ public void onNewChunk(OnChunkLoaded chunkAvailable, EntityRef worldEntity) {
...
             ByteBuffer.allocateDirect(2 * (ChunkConstants.SIZE_X * ChunkConstants.SIZE_Y * ChunkConstants.SIZE_Z));
         buffer.order(ByteOrder.nativeOrder());
-        short[] entries = new short[ChunkConstants.SIZE_X * ChunkConstants.SIZE_Y * ChunkConstants.SIZE_Z];
-        for (int x = 0; x < ChunkConstants.SIZE_X; x++) {
-            for (int y = 0; y < ChunkConstants.SIZE_Y; y++) {
-                for (int z = 0; z < ChunkConstants.SIZE_Z; z++) {
-                    int index = (z * ChunkConstants.SIZE_X * ChunkConstants.SIZE_Y) + (x * ChunkConstants.SIZE_Y) + y;
+        for (int z = 0; z < ChunkConstants.SIZE_Z; z++) {
+            for (int x = 0; x < ChunkConstants.SIZE_X; x++) {
+                for (int y = 0; y < ChunkConstants.SIZE_Y; y++) {
```
skaldarnar, line 151: What is the reason the order here changed from x, y, z to **z, x, y**?

skaldarnar (approved):

engine/src/main/java/org/terasology/physics/engine/VoxelLiquidWorldSystem.java
hajdam: This is how the data are organized in the Buffer (see. index variable on removed line 153 which has Y as 1+ increment)

pollend:
looks straightforward to me. I just implemented it like that for convince when I was implementing this.

pollend (approved):

jdrueckert:
`ci/cloudbees/error` didn't update correctly to display the successful last rerun

PR #4271, opened 2020-11-24
feat(discord-rpc): improve the system

iHDeveloper:
### Contains
This improvement aims to fix bugs in the system.
And, it improves communication between the game and Discord RPC to provide full access to its features.

image
_[Source: **Discord Developer Documentation**](https://discord.com/developers/docs/rich-presence/how-to#updating-presence-update-presence-payload-fields)_

Currently, the system supports `state` that can be changed dynamically using `DiscordRPCSubSystem#setState`.
And, `details` is also supported but its value is fixed as `Name: PlayerName`. It can't be changed.

This PR is going to implement other features such as `partySize`, `partyMax`, etc...

#### Discussion Points
- Turning the `default` to be dynamic to have different details from different gameplay such as `Currently in JoshariasSurvival`, etc...

- Since the engine has a networking infrastructure I'm thinking that we can implement the secrets (`join` and `spectate`)
in a way to provide the modules the ability to receive events like `DiscordAskToJoinEvent` and `DiscordSpectateEvent`.
Instead of each module dealing with it in a raw way.

_**Comment about what do you think of the points above?**_ 🤔 

#### Bugs
- **Game detection is broken**
The game assumes the player is in the lobby while in reality, the player is in-game.
- **Not thread-safe**
The Discord RPC connection is being handled in a separate thread to avoid conflicts with the game.
The system isn't thread-safe which might result in unexpected behavior.

### How to test
- Fetch the PR commits
`git fetch origin pull/4271/head:<localBranchName>`
- Checkout to the PR's branch
`git checkout <localBranchName>`
- Run the game and make sure you have `Discord Rich Presence` turned on.
- Join a game to trigger your game state from `In Lobby` to `In Game`

**NOTE:**
In terms of unit testing, I'm not sure how I would implement it for this system. 😅 

- [x] Fix: Game detection is broken (solved by #4274)
- [x] Fix: Not thread-safe (47bae74)
- [x] Feat: Ability to change `details` (48d1502)
- [ ] Feat: Implement party size
- [ ] Feat: Implement secrets
- [ ] Test the RPC
- - [ ] Single Player
- - [ ] Multiplayer


subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
```diff
@@ -0,0 +1,250 @@
...
+    private int tries;
+
+    private boolean enabled;
+    private boolean waiting;
+    private boolean connectedBefore;
+    private boolean connected;
+    private boolean autoReconnect;
+
+    public DiscordRPCThread() {
+        thread = new Thread(this);
```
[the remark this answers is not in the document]
iHDeveloper: I thought of using the `java.util.Timer`. But, that would make handling connection to Discord client harder to do.

Because there are special cases like:
- If the player doesn't have the Discord client we need to try to reconnect then gave up on trying.
- If the player has the Discord client then we connect to it and keep the connection alive. Even if the player disabled the RPC option in the player settings.
- When the player toggles the RPC option in the player settings we just have to send `null` to the Discord client. Instead of closing the connection. And re-connecting again when enabled.

pollend:
@iHDeveloper what is the status on this?

iHDeveloper:
> @iHDeveloper what is the status on this and what is left to be implemented?

Working on it. I'm debugging the party size commit. And, I'm not sure if the secrets should be implemented in this PR. Or, maybe later.

Cervator:
>  I'm not sure if the secrets should be implemented in this PR

@iHDeveloper I would suggest keeping the PRs about as small as could be, for more frequent merges. Can always keep going in additional PRs :-)

Every checkbox in the bullets above could pretty much be its own PR, although if multiple finish around the same time anyway and overlap at all then grouping them may make sense.

iHDeveloper:
This PR is ready to be reviewed and merged into the codebase. \o/ 😄 

I discussed with @Cervator and decided not to implement the secrets feature in this PR.
And, I'm going to keep it to be implemented in the future when needed.

I tested this PR on both **Singleplayer** and **Multiplayer (Hosting only)** with the latest build from the develop branch.
I didn't have the ability to test this PR by joining a server. But, I'm going to assume that it works fine on the server.

.idea/checkstyle-idea.xml
```diff
@@ -10,7 +10,7 @@
         <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
         <entry key="location-2" value="PROJECT_RELATIVE:$PROJECT_DIR$/config/metrics/checkstyle/checkstyle.xml:Terasology" />
         <entry key="property-2.samedir" value="config/metrics/checkstyle" />
-        <entry key="scan-before-checkin" value="false" />
+        <entry key="scan-before-checkin" value="true" />
```
skaldarnar, line 13: What kind of scans does this include? 

@keturn to the rescue? :upside_down_face:

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCSubSystem.java
```diff
@@ -38,281 +31,136 @@
...
+        config = rootContext.get(Config.class);
+
+        if (config.getPlayer().isDiscordPresence()) {
+            thread.enable();
+        } else {
+            logger.info("Discord RPC is disabled! No connection is being made during initialization.");
+            thread.disable();
         }
-        logger.info("Discord RPC >> Disconnected!");
+        thread.start();
```
skaldarnar, line 68: Since we log the start of initializing as INFO, should we also load the success when the thread was started (analog to the case where the integration is disabled).

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
```diff
@@ -0,0 +1,284 @@
...
+import com.jagrosh.discordipc.IPCClient;
+import com.jagrosh.discordipc.IPCListener;
+import com.jagrosh.discordipc.entities.RichPresence;
+import com.jagrosh.discordipc.exceptions.NoDiscordClientException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class DiscordRPCThread implements IPCListener, Runnable {
+    private static final Logger logger = LoggerFactory.getLogger(DiscordRPCThread.class);
+    private static final long DISCORD_APP_CLIENT_ID = 515274721080639504L;
```
skaldarnar, line 15: Is this a random number or something assigned to us in some way?

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
```diff
@@ -0,0 +1,284 @@
...
+    @Override
+    public void onDisconnect(IPCClient client, Throwable t) {
+        connected = false;
+        logger.info("Discord RPC lost connection: Disconnected!");
+    }
+
+    @Override
+    public void run() {
+        while (true) {
+            logger.info("Waiting for auto-connecting...");
```
skaldarnar, line 131: How much will this actually log on the INFO level? I'm cautious when I see `logger.info` in combination with loops or frequently called methods as this might spam the log.

skaldarnar (commented):
Code looks good overall, left a few comments. I haven't tested this, though.

.idea/checkstyle-idea.xml
iHDeveloper: Honestly, I didn't notice this file is added into the commit stage. 😅 
But, I can assume it happened because I enabled the Checkstyle scan before VCS check-in (aka before commit).

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
iHDeveloper: This random represents the client ID of the Discord application. Discord client needs this ID to fetch information and assets about the application to display it to the rich presence.

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
iHDeveloper: The purpose of the loop is to keep the thread alive. It's rare for this auto-connect log to appear.

It will appear by one of these conditions below:
- Once on the start
- Every time the player disables the rich presence from the player settings (if the player doesn't have the Discord client)
- When the thread tries to find the Discord client but still couldn't connect to it.

.idea/checkstyle-idea.xml
Cervator: Yes please - while that's possibly a good practice it is probably better if we leave it up to individual contributors for now, not all of whom may even have the Checkstyle plugin installed (I don't remember if that's a default one) :-)

Probably would be a different project to actually get the Checkstyle config and docs to the point where we could consider doing this for real. We'd want to make sure we have good docs on what a false positive is, for instance, or that might confuse contributors and lead to unrelated "fixes" in PRs we then have to ask to have undone 😅

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
Cervator: Yeah I think that might be associated with our registered app on Discord, which I originally set up. Nowadays the app is owned by a team with several of us having access 👍

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
Cervator: Likewise a bit concerned here, it doesn't seem overly far fetched that a connectivity issue could appear then suddenly the log is being filled by junk due to a decidedly optional extra that you could play without perfectly fine otherwise :-)

In other words: if the Discord integration breaks or is in a bad state it shouldn't be _too_ loud about it - ideally just log once per "incident" however we define that

.idea/checkstyle-idea.xml
skaldarnar: Then we should find a way to ignore these change by default (put `.idea` into the git-ignored list?). I think if a user changes this setting, it will automatically reflect in a change in this checked-in config file.... =/

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
skaldarnar: Is that _sensitive information_ (like an access token) or more of a global identifier?

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
skaldarnar: So what happens if the discord servers are down, or the player just does not have an active internet connection?

.idea/checkstyle-idea.xml
keturn: This looks like a topic for another venue. For the purposes of this PR, if it's not obviously the right setting, go ahead and revert the change to this file.

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
Cervator: Just an identifier 👍 There's also a secret but I don't think it is needed for anything in this case.

Incidentally you should have access to it via https://discord.com/developers/teams/781638052501520417 although it looks like you might have gotten an invite and not accepted it yet (Nira is in)

.idea/checkstyle-idea.xml
iHDeveloper: Alright I will revert the change and try to avoid adding those unrelated files by accident 😅

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
iHDeveloper: Yep as Cervator said it's an identifier so it's not sensitive information. And, about the secret, it's mostly used in OAuth2 applications which I think we don't need soon.

I agree with what Cervator said. And about the secret, it's mostly used in OAuth2 applications which I don't think we will need soon.

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCThread.java
iHDeveloper: The Discord client hosts a local IPC server for the applications to connect it.
When the application connects it exchanges information with the server to identify the application.
And then the application can send RPC information every x amount of time to update it.
The Discord client decides which application RPC will be shown depends on different factors.

If something happens in the Discord servers the only affected side is going to be the Discord client.
The Discord Client runs the IPC server when the client starts. And it doesn't stop until the client is being shutdown.
And until that happens it's going to stay listening to application RPCs whether the Discord server is down or not.

subsystems/DiscordRPC/src/main/java/org/terasology/engine/subsystem/discordrpc/DiscordRPCSubSystem.java
iHDeveloper: The purpose of the thread is to be alive all the time until the game shuts down. But if the user disabled the RPC the thread will be in the sleeping mode until something changes and notifies it. When the thread gets notified it tries to connect to the IPC server. If it fails to connect 5 times at maximum then the thread will go sleep again.

.idea/checkstyle-idea.xml
iHDeveloper: Alright I will revert the change and try to avoid adding those unrelated files by accident 😅

iHDeveloper:
- Pull with rebase to resolve conflict with `develop` branch
- Remove the change from `.idea/checkstyle-idea.xml`

skaldarnar:
:confused: Wanted to briefly test this on my Linux machine, ending up with an exception...

```
22:47:14.806 [main] INFO  o.t.e.s.d.DiscordRPCSubSystem - Initializing...
22:47:14.808 [DISCORD-RPC-THREAD] INFO  o.t.e.s.discordrpc.DiscordRPCThread - Waiting for auto-connecting...
22:47:14.810 [DISCORD-RPC-THREAD] INFO  o.t.e.s.discordrpc.DiscordRPCThread - Waiting for enabling...
22:47:14.810 [DISCORD-RPC-THREAD] INFO  o.t.e.s.discordrpc.DiscordRPCThread - Waiting for connection...
22:47:14.810 [DISCORD-RPC-THREAD] INFO  o.t.e.s.discordrpc.DiscordRPCThread - Connecting to Discord RPC...
Exception in thread "DISCORD-RPC-THREAD" java.lang.RuntimeException: org.newsclub.net.unix.AFUNIXSocketException: No such file or directory (socket: /run/user/1000/discord-ipc-0)
        at com.jagrosh.discordipc.entities.pipe.Pipe.createPipe(Pipe.java:167)
        at com.jagrosh.discordipc.entities.pipe.Pipe.openPipe(Pipe.java:67)
        at com.jagrosh.discordipc.IPCClient.connect(IPCClient.java:116)
        at org.terasology.engine.subsystem.discordrpc.DiscordRPCThread.run(DiscordRPCThread.java:170)
        at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: org.newsclub.net.unix.AFUNIXSocketException: No such file or directory (socket: /run/user/1000/discord-ipc-0)
        at org.newsclub.net.unix.NativeUnixSocket.connect(Native Method)
        at org.newsclub.net.unix.AFUNIXSocketImpl.connect(AFUNIXSocketImpl.java:136)
        at org.newsclub.net.unix.AFUNIXSocket.connect(AFUNIXSocket.java:107)
        at org.newsclub.net.unix.AFUNIXSocket.connect(AFUNIXSocket.java:98)
        at com.jagrosh.discordipc.entities.pipe.UnixPipe.<init>(UnixPipe.java:46)
        at com.jagrosh.discordipc.entities.pipe.Pipe.createPipe(Pipe.java:163)
        ... 4 more
```

Discord integration works in principle, at least Spotify manages to display my activity nicely...

skaldarnar (approved):

Issue #4200, opened 2020-10-14
JOML: Question about `Region3i` >>> `BlockRegion`


skaldarnar:
I was looking at `Region3i`/`BlockRegion` (JOML) tonight and I'm wondering how we gonna approach this. I started in PolyWorlds world generator (for some completely unrelated reason) and stumbled upon `teramath.Region3i::createFromCenterExtents(center, extents)`.

So I wanted to add a similar way to create a `BlockRegion`. Instead of adding more and more constructors or convenience methods I went for a utility class `BlockRegions` (plural form). The implementation for the creator method was simple enough (even simpler than with TeraMath):

```java
public static BlockRegion fromCenterAndExtents(Vector3ic center, Vector3ic extents) {
    Vector3ic min = center.sub(extents, new Vector3i());
    Vector3ic max = center.add(extents, new Vector3i());
    return new BlockRegion(min, max);
}
```

An equivalent implementation is the following:
```java
return new BlockRegion(center, center).addExtents(extents.x(), extents.y(), extents.z());
// when adding BlockRegion::addExtents(Vector3ic extents) this becomes even simpler
return new BlockRegion(center, center).addExtents(extents);
```

This, in turn, made me wonder whether we really need these helper methods, or whether that replacement is simple enough. Or whether we need a different kind of convenience methods, to be able to write something like
```java
BlockRegions.point(center).addExtents(extents);
// ... instead of ...
BlockRegions.fromCenterAndExtents(center, extents);
```

pollend:
maybe BlockRegion could also just accept a single value.

```
new BlockRegion(center).addExtents(extent);
```

skaldarnar:
Yep, I agree that the builder pattern allows for readable code here. I do, however, prefer the static methods with descriptive names over overloaded constructors. Doesn't matter too much though, we agree on the builder-like mutations 👍 

One thing I'm particularly worried about for BlockRegion is that point-like AABBs are **invalid** in JOML - we need to be careful that the semantics work out in the, e.g., for `containsBlock` 

https://github.com/MovingBlocks/Terasology/blob/a829df4c8802441c66078b00cb1f3f9f9ef4b6ad/engine/src/main/java/org/terasology/world/block/BlockRegion.java#L398-L406

I would assume that 

```java
assertTrue(new BlockRegion(center).containsBlock(center))
```

pollend:
that's also a really good point. I think the default behaviour for new BlockRegion(p,p) is also same.
[...]

PR #4272, opened 2020-11-24
feat(BlockRegion): add BlockRegion helpers

jdrueckert:
### Changes

- add `createFromMinAndMax` and `createFromCenterAndExtents` helpers
- deprecate BlockRegion constructors from min and max in favor of new helpers
- add tests for `BlockRegion` and `BlockRegionIterable` 
- add helper methods to create iterables for BlockRegions

### Module PRs

- https://github.com/Terasology/DynamicCities/pull/73
- https://github.com/Terasology/ClimateConditions/pull/38
- https://github.com/Terasology/StructureTemplates/pull/47
- https://github.com/Terasology/CombatSystem/pull/61
- https://github.com/Terasology/MetalRenegades/pull/117

---

Co-authored-by: Tobias Nett <skaldarnar@googlemail.com>
Co-authored-by: Michael Pollind <mpollind@gmail.com>

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/BlockRegion.java
+++ b/engine/src/main/java/org/terasology/world/block/BlockRegion.java
@@ -1,18 +1,5 @@
-/*
- * Copyright 2020 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// Copyright 2020 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
 package org.terasology.world.block;
 
 import org.joml.AABBf;
@@ -44,7 +31,6 @@ public class BlockRegion {
     public final AABBi aabb = new AABBi();
 
     public BlockRegion() {
-        this.setMin(0, 0, 0).setMax(0, 0, 0);
     }
 
     public BlockRegion(BlockRegion source) {
--- a/engine/src/main/java/org/terasology/world/block/BlockRegion.java
+++ b/engine/src/main/java/org/terasology/world/block/BlockRegion.java
@@ -33,8 +33,8 @@ import org.joml.Vector3ic;
 import org.terasology.entitySystem.entity.EntityRef;
 
 /**
- * is a bounded box describing blocks contained within.
- * A {@link BlockRegion} is described and backed by an {@link AABBi}
+ * is a bounded box describing blocks contained within. A {@link BlockRegion} is described and backed by an {@link
+ * AABBi}
  */
 public class BlockRegion {
 
@@ -44,6 +44,7 @@ public class BlockRegion {
     public final AABBi aabb = new AABBi();
 
     public BlockRegion() {
+        this.setMin(0, 0, 0).setMax(0, 0, 0);
     }
 
     public BlockRegion(BlockRegion source) {
@@ -54,10 +55,18 @@ public class BlockRegion {
         aabb.set(source);
     }
 
+    /**
+     * Deprecated in favor of {@link org.terasology.world.block.BlockRegions#createFromMinAndMax(Vector3ic, Vector3ic)}
+     */
+    @Deprecated
     public BlockRegion(Vector3ic min, Vector3ic max) {
         this(min.x(), min.y(), min.z(), max.x(), max.y(), max.z());
     }
 
+    /**
+     * Deprecated in favor of {@link org.terasology.world.block.BlockRegions#createFromMinAndMax(Vector3ic, Vector3ic)}
+     */
+    @Deprecated
     public BlockRegion(int minX, int minY, int minZ, int maxX, int maxY, int maxZ) {
         this.setMin(minX, minY, minZ).setMax(maxX, maxY, maxZ);
     }
@@ -82,22 +91,28 @@ public class BlockRegion {
     public Vector3i getMax(Vector3i dest) {
         return dest.set(aabb.maxX - 1, aabb.maxY - 1, aabb.maxZ - 1);
     }
+
     /**
      * the maximum coordinate of the second block x
+     *
      * @return the minimum coordinate x
      */
     public int getMaxX() {
         return this.aabb.maxX - 1;
     }
+
     /**
      * the maximum coordinate of the second block y
+     *
      * @return the minimum coordinate y
      */
     public int getMaxY() {
         return this.aabb.maxY - 1;
     }
+
     /**
      * the maximum coordinate of the second block z
+     *
      * @return the minimum coordinate z
      */
     public int getMaxZ() {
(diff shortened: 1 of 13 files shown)


jdrueckert:
As proposed in #4200

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -44,6 +44,7 @@
     public final AABBi aabb = new AABBi();
 
     public BlockRegion() {
+        this.setMin(0, 0, 0).setMax(0, 0, 0);
```
skaldarnar, line 47: This changes the semantics of `new BlockRegion()` from being some invalid infinite something (the empty region, so to say) to the non-empty BlockRegion containing exactly (0,0,0).

This allows to express `createFromCenterAndExtents` conveniently via
```java
new BlockRegion().translate(center).addExtents(extents)
```

Not sure whether the previous semantics was used somewhere, or whether we need the empty region to be a thing. We could add that to `BlockRegions`:

```java
public static BlockRegion empty() {
  return new BlockRegion(new AABBi());
}
```

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -54,10 +55,18 @@ public BlockRegion(AABBi source) {
         aabb.set(source);
     }
 
+    /**
+     * Deprecated in favor of {@link org.terasology.world.block.BlockRegions#createFromMinAndMax(Vector3ic, Vector3ic)}
+     */
+    @Deprecated
```
skaldarnar, line 61: (_copying from Discord for the record_)

I think it is somewhat confusing to have some variants as constructor `new BlockRegion(min, max)` and some as constructor method `BlockRegions.createFromCenterAndExtents(center, extents)`. Especially if another variant like `BlockRegions.createFromMinAndSize(min, size)` has the same argument types...

Most of the implementations are pretty neat one-liners which may be inlined instead. For consistency I'd also inline the two constructor variants that were deprecated.

Only culprit we found is create a BlockRegion from center and extents when given floating point coordinates. One needs to be careful with rounding in that case...

```java
    public static BlockRegion createFromCenterAndExtents(Vector3fc center, Vector3fc extents) {
        Vector3f min = center.sub(extents, new Vector3f());
        Vector3f max = center.add(extents, new Vector3f());

        return new BlockRegion().setMin(new Vector3i(min, RoundingMode.CEILING)).setMax(new Vector3i(max,
                RoundingMode.FLOOR));
    }
```

engine/src/main/java/org/terasology/world/block/BlockRegions.java
```diff
@@ -0,0 +1,59 @@
...
+     * extents}
+     *
+     * @return new block region
+     */
+    public static BlockRegion createFromCenterAndExtents(Vector3fc center, Vector3fc extents) {
+        Vector3f min = center.sub(extents, new Vector3f());
+        Vector3f max = center.add(extents, new Vector3f());
+
+        return new BlockRegion().setMin(new Vector3i(min, RoundingMode.CEILING)).setMax(new Vector3i(max,
+                RoundingMode.FLOOR));
```
pollend, line 57: for the min you also want to floor the value.

if your min is (.001,.001,.001) ceiling would be (1,1,1) which would be one block extra. .001 would be within (0,0,0)

pollend (changes requested):

engine/src/main/java/org/terasology/world/block/BlockRegion.java
pollend: a static method to define an empty blockregion seems kind of strange. It makes more sense to me to have the initial state as undefined. `setMin(0, 0, 0).setMax(0, 0, 0)` would yield a region that is 1x1x1 units in size. min is 0,0,0 and max is 1,1,1

engine/src/main/java/org/terasology/world/block/BlockRegion.java
pollend: Its a little confusing to move the actual declaration to a separate static class. might be better to move the static implementations into BlockRegion so they are easily found in one place and leave the default constructor for min and max since that directly describes how a region is defined.

engine/src/main/java/org/terasology/world/block/BlockRegions.java
skaldarnar: Hm, but doing a floor on the min value would cause the region to grow because of rounding. Consider the region `R` where `R_min = (-0.5, -0.5, -0.5)` and `R_max = (0.5, 0.5, 0.5)`. I'd think that `R.containsPoint(0, 0, 0)` is true, but false for all other points (with integer coordinates). 

If we would _floor_ the min values the region `R'` would span the AABB from (-1, -1, -1) to (0, 0, 0) and `R'.contains(-1, -1, -1)` would now hold. I think the caller should make sure that the rounded values are "correct", possibly by adding a (small) margin.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
jdrueckert: @pollend From what I read in #4200 you agreed to what we implemented :thinking: 
Why the change of heart?
On what grounds should creation from min and max be the default constructor?

pollend: I was thinking about it a bit, but maybe it's not as confusing as I thought it would be.

skaldarnar: One more thing is discoverability from user's perspective. Following up to @jdrueckert's question on why initialization from min and max is the default, I'd argue that from center and extents or from min and size are equally valid ways to create a BlockRegion. 

The pattern with the plural form `BlockRegions` is also used in other places in the Java world (like `Lists` or `Streams` from guava), but there are also cases like `Stream.of` from plain Java... 

I would go for the following line-up (I'd prefer `BlockRegions` but I could also live with `BlockRegion`):

```java
BlockRegion r = new BlockRegion(); // empty region
BlockRegion r = otherRegion.copy()

BlockRegion r = BlockRegions.createFromMinAndMax(min, max)
BlockRegion r = BlockRegions.createFromCenterAndExtents(center, extents)
BlockRegion r = BlockRegions.createFromMinAndSize(min, size)
BlockRegion r = BlockRegions.createFromPoint(point)
```

engine/src/main/java/org/terasology/world/block/BlockRegion.java
skaldarnar: hm, I probably don't get how these AABBs and regions work exactly :monocle_face: 

I thought that `setMin(0, 0, 0).setMax(0, 0, 0)` would create a region of size 0x0x0, containing a single point (0,0,0). This is basically a **point**.

With `setMin(0, 0, 0).setMax(0, 0, 1)` you'd create a region of size 0x0x1 containing 2 points. This is a **line**.

With `setMin(0, 0, 0).setMax(0, 1, 1)` you'd create a region of size 0x1x1 containing 4 points. This is a **plane** (surface).

With `setMin(0, 0, 0).setMax(1, 1, 1)` you'd create a unit cube of size 1x1x1 containing 8 points. This is a **volume**.

Is this understanding wrong?

skaldarnar: Another interpretation/visualization of `setMin(0, 0, 0).setMax(0, 0, 0)` would be to think of the region that contains exactly the block at position (0,0,0), i.e., the region going from min (-.5, -.5, -.5) to max (.5, .5, .5) and being of size 1x1x1.

skaldarnar: That explanation helps - I'll make sure to add that in the JavaDoc of BlockRegion for further reference.

pollend (approved):

skaldarnar (approved):

PR #4278, opened 2020-11-25
fix: prevent NPE during movement interpolation

skaldarnar:
**Problem:** Killing a mob in a multi-player game as connected client lead to a NPE in the `CharacterMovementSystemUtility`. I suspect this is because we try to interpolate the movement of the mob even though we already removed components (like the `CharacterMovementComponent`) from the entity. 

**Solution:** Make use of `updateComponent` functional API to only update the respective component if it is present. Otherwise, nothing will happen. 

**Question:** This (maybe) only fixes a symptom. The interpolation is called for entries in a field called `playerStates`, but those actually contain NPC entities such as the deer or sheep. Basically, all entities with `CharacterMovementComponent`, `LocationComponent`, and `AliveCharacterComponent` are considered for movement prediction. Is this intended? The entity is supposed to be removed from `playerStates` on destruction - so why do we still try to interpolate movement when the entity was just killed?

### DIFF
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterMovementSystemUtility.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterMovementSystemUtility.java
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.terasology.logic.characters;
 
+import com.google.common.collect.Lists;
 import org.joml.Quaternionf;
 import org.joml.Vector3f;
 import org.terasology.entitySystem.entity.EntityRef;
@@ -89,7 +90,11 @@ public final class CharacterMovementSystemUtility {
             return movementComponent;
         });
 
-        setPhysicsLocation(entity, newPos);
+        // BulletPhysics requires the entity to have both these components. This is not clear from the interfaces we're
+        // using, but the exception thrown in 'BulletPhysics#createCharacterCollider' is pretty self-explanatory...
+        if (entity.hasAllComponents(Lists.newArrayList(CharacterMovementComponent.class, LocationComponent.class))) {
+            setPhysicsLocation(entity, newPos);
+        }
     }
 
     public void setToExtrapolateState(EntityRef entity, CharacterStateEvent state, long time) {
--- a/engine/src/main/java/org/terasology/logic/characters/CharacterMovementSystemUtility.java
+++ b/engine/src/main/java/org/terasology/logic/characters/CharacterMovementSystemUtility.java
@@ -1,18 +1,5 @@
-/*
- * Copyright 2015 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


Cervator:
Short answer: I _think_ that yes the prediction is intended for all moving entities. The exact sequence probably is somewhat brittle, we hit another related issue fixed with #4131. 

Longer answer: Anything making the process _less_ brittle is probably a good thing, even if it may just be fixing a symptom. In the other PR we fixed something by not removing dead entities until the end of movement prediction. Is that better or worse than trying to do it mid-prediction? Not really sure. But it seems like hardening the update calls might help in either situation? 

I could imagine this is another spot where over time we've just ended up with code that might look like "Huh, how did all these pieces seem reasonable together?" when it was really one step at a time over a long while. A full and true cleanup might avoid the symptom entirely, until more changes happen incrementally again. Hardening stuff in either case seems like a good idea, might end up letting us go longer without crashing even if some less performant code sneaks in. Unless it becomes a bottleneck it may not be worth the added attention.

Cervator (dismissed):
Approach looks plenty reasonable to me 👍 Haven't tested it though

skaldarnar:
Hm, just gave it another test run, now it crashed with this:

```stacktrace
23:28:39.332 [main] ERROR o.terasology.engine.TerasologyEngine - Uncaught exception, attempting clean game shutdown
java.lang.IllegalArgumentException: Expected an entity with a Location component and CharacterMovementComponent.
	at org.terasology.physics.bullet.BulletPhysics.createCharacterCollider(BulletPhysics.java:540)
	at org.terasology.physics.bullet.BulletPhysics.getCharacterCollider(BulletPhysics.java:446)
	at org.terasology.logic.characters.CharacterMovementSystemUtility.setPhysicsLocation(CharacterMovementSystemUtility.java:129)
	at org.terasology.logic.characters.CharacterMovementSystemUtility.setToInterpolateState(CharacterMovementSystemUtility.java:92)
	at org.terasology.logic.characters.ServerCharacterPredictionSystem.setToTime(ServerCharacterPredictionSystem.java:244)
	at org.terasology.logic.characters.ServerCharacterPredictionSystem.restoreToPresent(ServerCharacterPredictionSystem.java:266)
	at org.terasology.network.internal.NetClient.processEvents(NetClient.java:538)
	at org.terasology.network.internal.NetClient.processReceivedMessages(NetClient.java:422)
	at org.terasology.network.internal.NetClient.update(NetClient.java:220)
	at org.terasology.network.internal.NetworkSystemImpl.update(NetworkSystemImpl.java:337)
	at org.terasology.engine.subsystem.common.NetworkSubsystem.preUpdate(NetworkSubsystem.java:52)
	at org.terasology.engine.TerasologyEngine.tick(TerasologyEngine.java:472)
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:442)
	at org.terasology.engine.TerasologyEngine.run(TerasologyEngine.java:418)
```

jdrueckert (approved):

PR #4285, opened 2020-11-28
fix: tweak lighting for over exposure problem

pollend:
during mid day the lighting seems to get heavily over-saturated to the point where a lot of the scene is heavily washed out. 

current:
image

modified:
image

image

### DIFF
--- a/engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
+++ b/engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
@@ -186,7 +186,7 @@ float calcDayAndNightLightingFactor(float daylightValue, float daylight) {
 }
 
 vec3 calcSunlightColorDeferred(float daylightValue, float diffuseLighting, float ambientIntensity, float diffuseIntensity, vec3 ambientColor, vec3 diffuseColor) {
-    vec3 daylightColorValue = vec3(ambientIntensity) + diffuseLighting * diffuseIntensity * diffuseColor;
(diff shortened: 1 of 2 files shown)


pollend:
image

image

engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
```diff
@@ -186,7 +186,7 @@ float calcDayAndNightLightingFactor(float daylightValue, float daylight) {
 }
 
 vec3 calcSunlightColorDeferred(float daylightValue, float diffuseLighting, float ambientIntensity, float diffuseIntensity, vec3 ambientColor, vec3 diffuseColor) {
-    vec3 daylightColorValue = vec3(ambientIntensity) + diffuseLighting * diffuseIntensity * diffuseColor;
+    vec3 daylightColorValue = max(vec3(ambientIntensity),diffuseLighting * diffuseIntensity * diffuseColor);
```
skaldarnar, line 189: So now the `ambientIntensity` is a threshold rather than a base value? Was the ambient intensity used to get this bloom effect? 

I think having everything more evenly lit is quite nice, I'm just wondering whether it might be a bit too dull...

skaldarnar (approved):

engine/src/main/resources/org/terasology/include/globalFunctionsFragIncl.glsl
pollend: yea, the colors are starting to look a bit dull. we need to really look at the lighting and rework the shaders. I kind of prefer this over the intense saturation.

PR #4287, opened 2020-11-28
add standard and performant options for blockregion iteration

pollend:
Here is a tweak to how the iterator works. I kind of wanted to keep the type handling to avoid bugs. implementation is a bit more complex but I think this should work well.


skaldarnar:
Closing as manually integrated with https://github.com/MovingBlocks/Terasology/pull/4272

PR #4293, opened 2020-11-30
feat: replace deprecated md5 assets with new gltf floating cube

jdrueckert:
- please note that in contrast to the old md5 assets, the new gltf asset does not have animations


skaldarnar (approved):
tested in local multiplayer, could see white default cubes for the players :+1:

PR #4294, opened 2020-11-30
chore: remove some of the most noisy debug logs

skaldarnar:
**Problem:** There was a lot of noise in our logs, especially on DEBUG level which is used by IntelliJ as default. This was particularly spamming from `update()` methods.

**Solutions:** Remove those debug logs, which often look like `println`-debug statements to keep track of where the execution flow goes without the need to actually step through everything in the debugger.

### DIFF
--- a/facades/PC/src/main/resources/logback.xml
+++ b/facades/PC/src/main/resources/logback.xml
@@ -48,5 +48,6 @@
   <logger name="org.terasology" level="${logOverrideLevel:-info}" />
   <logger name="org.reflections.Reflections" level="${logOverrideLevel:-error}" />
   <logger name="com.snowplowanalytics" level="${logOverrideLevel:-error}" />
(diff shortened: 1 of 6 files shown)


facades/PC/src/main/resources/logback.xml
```diff
@@ -48,5 +48,6 @@
   <logger name="org.terasology" level="${logOverrideLevel:-info}" />
   <logger name="org.reflections.Reflections" level="${logOverrideLevel:-error}" />
   <logger name="com.snowplowanalytics" level="${logOverrideLevel:-error}" />
+  <logger name="io.netty" level="${logOverrideLevel:-warn}" />
```
skaldarnar, line 51: This suppresses the netty DEBUG message with some exception, at least when run via Gradle.

jdrueckert (approved):

PR #4298, opened 2020-12-01
fix: reduce volume and overmodulation of main theme

jdrueckert:
- reduce volume and overmodulation of main theme
- use 0.5 as default music volume
- remove deprecated audio settings from default.cfg as AudioSettings use AutoConfig


skaldarnar (commented):
Hm, why does Audacity blow up the file size so much? Might that be the reason for the current quality (small file size)? File size after this PR is 5.7MB....

```+4.35 MB (440%) engine/src/main/resources/assets/music/MenuTheme.ogg ```

I don't think that this is a problem per se, but I think we should avoid tracking "big" files in git or think about something like https://git-lfs.github.com/

skaldarnar:
I tried to reproduce the changes with Audacity and could achieve the following file sizes:
- OGG quality :one: 1.3M
- OGG quality :three: 1.6M
- OGG quality :five: 2.0M (default)

skaldarnar (approved):

PR #4299, opened 2020-12-03
fix: dispose physic's natives, stop threads.

DarkWeird:
### Contains

Free resources for MTE.
related to https://github.com/Terasology/ModuleTestingEnvironment/issues/30

dispose physic's natives(holds TerasologyEngine)
stop threads(holds TerasologyEngine)

### How to test

1. Run MTE test with client connection in loop
2. measure memory usage.

### DIFF
--- a/engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
+++ b/engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
@@ -1,18 +1,5 @@
-/*
- * Copyright 2016 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
```diff
@@ -589,7 +580,7 @@ private RigidBody newRigidBody(EntityRef entity) {
     private void removeCollider(btCollisionObject collider) {
         discreteDynamicsWorld.removeCollisionObject(collider);
 //        collider.getCollisionShape().dispose();
-//        collider.dispose();
```
pollend, line 592: can we remove these commented lines

pollend (changes requested):

skaldarnar (approved):

PR #4296, opened 2020-12-01
feat(JOML): migrate BlockRegionComponent logic

pollend:
Fairly straightforward migration of BlockRegionComponent. along with some thoughts from looking at how the logic works: https://github.com/MovingBlocks/Terasology/issues/4295 . I also need separate PR's in these modules for this API change: 

- [ ] [Furnishings](https://github.com/Terasology/Furnishings/pull/4)
- [ ] [Health](https://github.com/Terasology/Health/pull/55)
- [ ] [IRLCorp](https://github.com/Terasology/IRLCorp/pull/29)
- [ ] [LightAndShadow](https://github.com/Terasology/LightAndShadow/pull/176)
- [ ] [Machines](https://github.com/Terasology/Machines/pull/42)
- [ ] [MultiBlock](https://github.com/Terasology/MultiBlock/pull/11)
- [ ] [Rails](https://github.com/Terasology/Rails/pull/63)
- [ ] [Smithing](https://github.com/Terasology/Smithing/pull/11)
- [ ] [StructureTemplates](https://github.com/Terasology/StructureTemplates/pull/48)
- [ ] [WoodAndStone](https://github.com/Terasology/WoodAndStone/pull/63)
- [ ] [WorkstationCrafting](https://github.com/Terasology/WorkstationCrafting/pull/17)

```bash
gh pr checkout 4296
for m in Furnishings Health Smithing WoodAndStone WorkstationCrafting StructureTemplates MultiBlock Machines; do
    pushd modules/$m
    git fetch
    git checkout feature/joml-migrate-blockregion
    popd
done
```

PR #4301, opened 2020-12-06
fix(BlockRegion): correct center calculation for BlockRegion

pollend:
this center method was only calculating the extent of a region. this produced off center structures in MetalRenegades. 

offcenter


skaldarnar (approved):
For some reason I cannot push changes to this PR, so I've added some tests to #4296 in https://github.com/MovingBlocks/Terasology/pull/4296/commits/2b7dfcf95e803d95b7ecda68ba621f3c67f8c810.

The center calculation was off for block regions spanning across 0, i.e, min component being negative and max component being positive for some dimension.

PR #4307, opened 2020-12-09
ci(subsystems): Make TypeHandlerLibrary publishable with engine's version

DarkWeird:
### Contains

Fix single module building in Jenkins.

### How to test

1. invoke `gradle publish`
2. check that you have artifact `org.terasology:TypeHandlerLibrary:<engine version>` in repo

- [ ] Check how it works in jenkins ()

### DIFF
--- a/subsystems/TypeHandlerLibrary/build.gradle.kts
+++ b/subsystems/TypeHandlerLibrary/build.gradle.kts
@@ -8,7 +8,7 @@ plugins {
 
 apply(from = "$rootDir/config/gradle/publish.gradle")
 
-group = "org.terasology"
+group = "org.terasology.subsystems"
 version = project(":engine").version
 
 dependencies {
--- a/subsystems/TypeHandlerLibrary/build.gradle.kts
+++ b/subsystems/TypeHandlerLibrary/build.gradle.kts
@@ -6,7 +6,10 @@ plugins {
     `java-library`
 }
 
-apply(from = "$rootDir/config/gradle/common.gradle")
+apply(from = "$rootDir/config/gradle/publish.gradle")
+
+group = "org.terasology"
+version = project(":engine").version
 
 dependencies {
     implementation("org.slf4j:slf4j-api:1.7.21")


subsystems/TypeHandlerLibrary/build.gradle.kts
```diff
@@ -6,7 +6,10 @@ plugins {
     `java-library`
 }
 
-apply(from = "$rootDir/config/gradle/common.gradle")
+apply(from = "$rootDir/config/gradle/publish.gradle")
```
jdrueckert, line 9: why do we replace this here? 🤔 
I would guess that
a) we used `common.gradle` for something and
b) `publish.gradle` does (should) not duplicate what's in `common.gradle`

DarkWeird: `common.gradle` provides:
0. repositories for artifact resolving
1. pmd,spotbug,checkstyle etc

`publish.gradle` - provides:
0. `common.gradle` logic 
1. publish to our maven repository logic.

We should publish this subsystem in external repo for exposing this subsystem to single module builds. (aka make it accessable in engineless builds)

subsystems/TypeHandlerLibrary/build.gradle.kts
```diff
@@ -6,7 +6,10 @@ plugins {
...
+
+group = "org.terasology"
```
jdrueckert, line 11: Should subsystems be published to group "org.terasology.subsystems"?

jdrueckert: Examples from artifactory:

engine
```
<dependency>
    <groupId>org.terasology.engine</groupId>
    <artifactId>engine</artifactId>
    <version>4.1.0</version>
    <type>pom</type>
</dependency>
```

PC facade
```
<dependency>
    <groupId>org.terasology.facades</groupId>
    <artifactId>PC</artifactId>
    <version>4.1.0</version>
    <type>pom</type>
</dependency>
```

So maybe for type handler library subsystem we could opt for
```
<dependency>
    <groupId>org.terasology.subsystems</groupId>
    <artifactId>TypeHandlerLibrary</artifactId>
    <version>4.1.0</version>
    <type>pom</type>
</dependency>
```

subsystems/TypeHandlerLibrary/build.gradle.kts
Cervator: Also: `publish.gradle` effectively _extends_ `common.gradle` to avoid said overlap. This just means THL goes from being a generic Java project to one that gets published, which sounds about right based on the PR title (we just need to make sure we're on top of whatever added version tag it uses)

subsystems/TypeHandlerLibrary/build.gradle.kts
Cervator: Concur, `org.terasology.subsystems` sounds great 👍

jdrueckert, line 11:
```suggestion
group = "org.terasology.subsystems"
```

subsystems/TypeHandlerLibrary/build.gradle.kts
```diff
@@ -6,7 +6,10 @@ plugins {
...
+version = project(":engine").version
```
Cervator, line 12: Ah there we go, there's the version link, which is fine if they'll always be in sync, which seems reasonable :-)

Cervator (approved):
Haven't outright tested, but just eyeballing it looks good, with the quick group change thrown in 👍

Cervator (approved):

PR #4311, opened 2020-12-11
Add borders to atlas subtextures

### LINKED ISSUE #4180: Visual artifacts on crosshair

4Denthusiast:
Hopefully fixes #4180 by slightly reducing the size of the area that's actually used for each region within an atlas. I can't actually reproduce that issue in the first place (it seems to depend on which graphics card is used), so someone else will have to test whether it actually does fix the issue. Also, the constant BORDER_SIZE was picked pretty arbitrarily, so testing how large that needs to be would also be useful. It should be less than 1/the resolution of any texture used, but also high enough that it actually fixes the issue (if it's so near to 0 that it's effectively like it's 0 due to rounding errors, this PR will just have no effect).


skaldarnar (commented):
Hm, I cannot reprodcue the issue on my Windows machine anymore 🤔 (might have updated my GPU drivers in the meantime...). I'll try to remember to test this on my Linux later...

What I can confirm is that these changes don't seem to have any visual impact otherwise ✔️

DarkWeird (approved):
Check on my linux laptop. Nvidia driver. not updated :)

It is works!

PR #4321, opened 2020-12-13
fix!(BlockRegion): Rewrite of BlockRegion

### LINKED ISSUE #4086: BlockRegion intersection tests incorrect.
All of the BlockRegion.intersects* methods with floating-point arguments and the BlockRegion.containsPoint(Vector3fc) method are incorrect. The actual position of the corner of a BlockRegio is not (region.aabb.minX, region.aabb.minY, region.aabb.minZ) but (region.aabb.minX-0.5, region.aabb.minY-0.5, region.aabb.minZ-0.5), as points in the world with integer coordinates are actually at the centres of the blocks with the corresponding integer coordinates, not the corner. This means that the region covered by the BlockRegion is not the same as the region covered by its AABB.

Currently none of these BlockRegion methods seem to be actually used, so this does not manifest as an actual bug. Ideally it shouldn't be left in this state though, and the methods should be either corrected or removed.

QuentinAndre11:
### Contains

Fixes #4086 
BlockRegion intersection tests were based on the center of the blocks instead of their corner, and we fixed it.

### How to test

We added unit tests, you can run it or add some tests if you think that some are missing. It's impossible to test it in the game because these functions seem to be currently unused. 

These methods might be removed because currently unused,

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/BlockRegion.java
+++ b/engine/src/main/java/org/terasology/world/block/BlockRegion.java
@@ -7,7 +7,6 @@ import org.joml.AABBi;
 import org.joml.Intersectionf;
 import org.joml.LineSegmentf;
 import org.joml.Math;
-import org.joml.Matrix4f;
 import org.joml.Matrix4fc;
 import org.joml.Planef;
 import org.joml.Rayf;
@@ -51,9 +50,6 @@ public class BlockRegion {
      * The z coordinate of the maximum corner.
      */
     private int maxZ = Integer.MIN_VALUE;
-
-    private AABBf bounds = new AABBf();
-    private boolean isDirt = true;
 
     public BlockRegion() {
     }
@@ -158,7 +154,6 @@ public class BlockRegion {
      * @return the minX
      */
     public BlockRegion maxX(int x) {
-        this.isDirt = true;
         this.maxX = x;
         return this;
     }
@@ -169,7 +164,6 @@ public class BlockRegion {
      * @return the minY
      */
     public BlockRegion maxY(int y) {
-        this.isDirt = true;
         this.maxY = y;
         return this;
     }
@@ -180,7 +174,6 @@ public class BlockRegion {
      * @return the minZ
      */
     public BlockRegion maxZ(int z) {
-        this.isDirt = true;
         this.maxZ = z;
         return this;
     }
@@ -191,7 +184,6 @@ public class BlockRegion {
      * @return the minX
      */
     public BlockRegion minX(int x) {
-        this.isDirt = true;
         this.minX = x;
         return this;
     }
@@ -202,7 +194,6 @@ public class BlockRegion {
      * @return the minY
      */
     public BlockRegion minY(int y) {
-        this.isDirt = true;
         this.minY = y;
         return this;
     }
@@ -213,7 +204,6 @@ public class BlockRegion {
      * @return the minZ
      */
     public BlockRegion minZ(int z) {
-        this.isDirt = true;
         this.minZ = z;
         return this;
     }
@@ -297,24 +287,20 @@ public class BlockRegion {
         this.maxX = source.maxX;
         this.maxY = source.maxY;
         this.maxZ = source.maxZ;
-        this.isDirt = true;
         return this;
     }
 
     //TODO: 1.9.26 has a constant interface for aabbf
-    public AABBf getBounds() {
-        if(isDirt) {
-            return bounds.setMin(
-                this.minX - .5f,
-                this.minY - .5f,
-                this.minZ - .5f
-            ).setMax(
-                this.maxX + .5f,
-                this.maxY + .5f,
-                this.maxZ + .5f
-            );
-        }
-        return bounds;
+    public AABBf getBounds(AABBf dest) {
+        dest.minX = minX - .5f;
+        dest.minY = minY - .5f;
+        dest.minZ = minZ - .5f;
+
+        dest.maxX = minX + .5f;
+        dest.maxY = minY + .5f;
+        dest.maxZ = minZ + .5f;
+
+        return  dest;
     }
 
     /**
@@ -327,7 +313,6 @@ public class BlockRegion {
         this.minX = min.x();
         this.minY = min.y();
         this.minZ = min.z();
-        isDirt = true;
         return this;
     }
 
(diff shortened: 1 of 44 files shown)


pollend (changes requested):
I think it would be ideal to drop the AABBi object and replace it with an AABBf to just cache the floating point position. replace AABBi with 6 integer values for the min and max. for getMax you don't need to +1 and -1 for that calculation anymore since AABBi doesn't represent the bounds of the region anymore.  we also want to avoid declaring temporary objects in this way to avoid unnecessary GC. 

image

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -25,20 +25,39 @@
...
+    /**
+     * The y coordinate of the maximum corner.
+     */
+    public int maxY = Integer.MIN_VALUE;
     /**
-     * AABB region that backs a BlockRegion
+     * The z coordinate of the maximum corner.
      */
-    public final AABBi aabb = new AABBi();
+    public int maxZ = Integer.MIN_VALUE;
```
skaldarnar: @pollend Should those really be public? I'm a bit confused right now, as I thought we agreed on Slack with @4Denthusiast that we should ensure that invalid min/max combinations should throw exceptions. With these values publicly accessible we cannot guarantee that...

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -792,36 +1026,50 @@ public BlockRegion addExtents(int extentX, int extentY, int extentZ, BlockRegion
...
+
+        dest.maxX = Math.roundUsing(this.maxX + extentX, RoundingMode.CEILING);
+        dest.maxY = Math.roundUsing(this.maxY + extentY, RoundingMode.CEILING);
+        dest.maxZ = Math.roundUsing(this.maxZ + extentZ, RoundingMode.CEILING);
 
-        dest.aabb.maxX = Math.roundUsing(this.aabb.maxX + extentX, RoundingMode.CEILING);
-        dest.aabb.maxY = Math.roundUsing(this.aabb.maxY + extentY, RoundingMode.CEILING);
-        dest.aabb.maxZ = Math.roundUsing(this.aabb.maxZ + extentZ, RoundingMode.CEILING);
         return dest;
     }
```
pollend, line 1039: I'm pretty sure this logic is incorrect @4Denthusiast

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -297,24 +287,20 @@ public BlockRegion set(BlockRegion source) {
...
-        }
-        return bounds;
+    public AABBf getBounds(AABBf dest) {
+        dest.minX = minX - .5f;
+        dest.minY = minY - .5f;
+        dest.minZ = minZ - .5f;
+
+        dest.maxX = minX + .5f;
+        dest.maxY = minY + .5f;
+        dest.maxZ = minZ + .5f;
```
4Denthusiast, line 301: maxZ, not minZ (and similarly for the other two).

engine/src/main/java/org/terasology/world/block/BlockRegion.java
skaldarnar: Agreed (and I've added a test for that). But I'm also not sure whether there's a value in this method anymore, now that we treat integer coordinates as block centers. I think the following should hold (and could be the implementation):

```java
public BlockRegion addExtents(float extentX, float extentY, float extentZ, BlockRegion dest) {
	return addExtents(Math.roundUsing(extentX, RoundingMode.FLOOR), 
		Math.roundUsing(extentY, RoundingMode.FLOOR), 
		Math.roundUsing(extentZ, RoundingMode.FLOOR), dest);
}
```

skaldarnar:
@QuentinAndre11 Sorry for "hijacking" this PR 😅 Your initial PR was right on point - highly appreciated! We just realized there is a lot more work needed to get this right 🙈 So, we just took the liberty to add more and more commits ourselves...

Please feel free to participate in this coding party. At this point, more tests are appreciated, as well as careful review of all the changes here.

pollend (approved):

engine/src/main/java/org/terasology/world/block/regions/BlockRegionComponent.java
```diff
@@ -8,15 +8,19 @@
...
 /**
+ *
  */
 public class BlockRegionComponent implements Component {
+    /**
+     * May be null.
+     */
     @Replicate
-    public BlockRegion region = new BlockRegion();
+    public BlockRegion region;
```
skaldarnar, line 18: I think this will have some implications, as it could not be null before.

I haven't double-checked where this component is used/accessed.

pollend: hmm, so an empty blockmap is null.

engine/src/main/java/org/terasology/world/block/BlockRegions.java
```diff
@@ -53,7 +56,7 @@ public static BlockRegion createFromMinAndMax(int minX, int minY, int minZ, int
      * @return new block region
      */
     public static BlockRegion createFromCenterAndExtents(Vector3ic center, Vector3ic extents) {
-        return new BlockRegion().union(center).addExtents(extents);
+        return new BlockRegion(center).extend(extents);
```
skaldarnar, line 59: I'm not sure this is intuitive - while writing tests for this, I was wondering whether 
```java
BlockRegions.createFromCenterAndExtents(new Vector3i(), new Vector3i(0,0,0))
```
is valid or not, and what it should return. In the given implementation it _is valid_, as the extents are added to the center block, while the utility method might be interpreted in a way that "extents" refers to the final extents...

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -19,317 +19,384 @@
...
 
-    public BlockRegion(BlockRegion source) {
-        aabb.set(source.aabb);
+    /**
+     * Creates an empty block region with invalid minimum/maximum corners.
+     * <p>
+     * {@link #isValid()} will return {@code false} for an empty block region created via this constructor.
+     */
+    @Deprecated
+    BlockRegion() {
```
skaldarnar, line 62: This is not public anymore, so there is hopefully no way to create invalid BlockRegions anymore. However, this comes in very handy when redcuing a list of positions in the `encompassing` helper methods, that's why I left it in...

Not sure what to do with it...

pollend: I wouldn't mark it as deprecated though.

PR #4326, opened 2020-12-20
refactor!(BlockRegion): full API overhaul

skaldarnar:
Follow up to https://github.com/MovingBlocks/Terasology/pull/4321


engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -116,6 +143,11 @@ public BlockRegion set(BlockRegion source) {
...
     }
 
+    /**
+     * Create a fresh copy of this block region.
+     *
+     * @see BlockRegion#BlockRegion(BlockRegion)
+     */
     public BlockRegion copy() {
         return new BlockRegion(this);
     }
```
pollend, line 153: this seems kind of redundant. do we need this?

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -528,9 +577,12 @@ public BlockRegion union(EntityRef blockRef) {
     // ---------------------------------------------------------------------------------------------------------------//
 
     /**
-     * calculate the BlockRegion that is intersected between another region
+     * Compute the intersection of this region with the {@code other} region.
+     * <p>
+     * NOTE: If the regions don't intersect this region will become invalid!
```
skaldarnar, line 582: This bit is a bit nasty, as this is currently the only way how to get an invalid region I think...

pollend (changes requested):

engine/src/main/java/org/terasology/world/block/BlockRegion.java
pollend: I don't think you can avoid everything.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
skaldarnar: While writing tests I thought that I could read

```java
region.copy().doThis().doThat()
```

way better than
```java
new BlockRegion(region).doThis().doThat()
```

Might be just personal taste, but I thought this would not hurt...

jdrueckert: I'm guessing @pollend meant the comment rather than the method itself.
In order to get this in so we can align the other joml PRs with it and make progress, I'll dismiss your review @pollend but please let's talk about this on Discord if you want and if applicable do a follow-up PR.

jdrueckert (dismissed):

engine/src/main/java/org/terasology/world/block/BlockRegions.java
```diff
@@ -16,98 +17,134 @@
...
+     * <i>empty region</i> by other means, e.g., by using {@code null} or {@link java.util.Optional}.
+     * <p>
+     * This region may be used as initial value for reductions on block regions. For instance, to compute the union (the
+     * region that encompasses all regions in some collection {@code regions}) the following Stream-API snippet can be
+     * used:
+     * <pre>
+     *     BlockRegion union = regions.stream().reduce(BlockRegions.INVALID, BlockRegion::union, BlockRegion::union);
+     * </pre>
+     */
+    public static final BlockRegion INVALID = new BlockRegion();
```
skaldarnar, line 33: I've asked about this on Discord: https://discord.com/channels/270264625419911192/701818347238588561/790546593911799808

skaldarnar: Note: this either has to be a `Supplier<BlockRegion>` or should return a `BlockRegionc` so that it cannot be modified.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
pollend: It's not too bad, I think it should be fine as is then. Just don't want there to be too many variations. Some of the convenience ends up being more confusing. might not be recognizable that copy means producing a new duplicate of the original. I don't think `new BlockRegion(a)` is any more convenient then `a.copy()`.  If we do something like then then be prepared to also do this for other situations. rather not having something purely unique only to BlockRegion if that makes sense.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -210,31 +245,32 @@ public BlockRegion minZ(int z) {
...
      *
-     * @param minX the x coordinate of the first block
-     * @param minY the y coordinate of the first block
-     * @param minZ the z coordinate of the first block
-     * @return this
+     * @return this region (after modification)
+     * @throws IllegalArgumentException if any dimension is greater than the respective component of the max
+     *         corner
      */
     public BlockRegion setMin(int minX, int minY, int minZ) {
```
pollend, line 275: I just realized this precondition might be a problem if you setMin above max and then aftewards you setmax. The final region is going to be valid but you're going to hit this exception. Its an extra burden to require the min and max to be set once.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -267,9 +303,10 @@ public int getMaxX() {
...
     /**
-     * set the maximum coordinate of the second block x
+     * Set the maximum x-coordinate of this region.
      *
-     * @return the minX
+     * @return this region (after modification)
+     * @throws IllegalArgumentException if {@code x} is smaller than the minimum x-coordinate
      */
     public BlockRegion maxX(int x) {
         Preconditions.checkArgument(x >= this.minX || this.minX == Integer.MAX_VALUE);
```
pollend, line 312: similar problem here.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -413,34 +455,58 @@ public Vector3i getSize(Vector3i dest) {
...
-     *
-     * @return number of blocks in the Y axis
+     * @see #sizeX()
+     */
+    public int getSizeX() {
+        return sizeX();
+    }
+
+    /**
+     * The number of blocks on the y axis.
```
pollend, line 472: There is really no need to offer both. it's either getSizeX or sizeX().

pollend (changes requested):

engine/src/main/java/org/terasology/world/block/BlockRegion.java
skaldarnar: I actually did re-expose the constructors for min/max and single block position, as well as the copy-constructor taking another region. What's still left on the utility class are constructors for `fromMinAndSize` and `fromCenterAndExtents`. 

I can remove the `copy` method, but then I'd also go for removing the utility class, as
```java
fromMinAndSize(min, size) == new BlockRegion(min).setSize(size);

fromCenterAndExtents(center, extents) == new BlockRegion(center).expand(extents);
```

It would be nice to make `BlockRegion implements Iterable`, but then we'd need to throw out on of the iterables, too (preferably keep in the in-place variant, as you can implement the other on caller side).

engine/src/main/java/org/terasology/world/block/BlockRegion.java
skaldarnar: Well, I thought we agreed with @4Denthusiast that we are trying for a design where a BlockRegion is (almost always) valid - which means that you have to pay attention to details like this. In this case you know that min and max will work out in the end, but what happens if the calls are split over several methods or even different systems handling it?

We already have `set(BlockRegion)`, maybe we also need `set(minX, minY, minZ, maxX, maxY, maxZ)` which would solve your problem, I guess?

engine/src/main/java/org/terasology/world/block/BlockRegion.java
skaldarnar: Yep, will remove `sizeX()` again as per comment on Discord: "only use `minX` (without `get` for coordinates and/or main attributes?)".

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,550 +114,397 @@
...
-    /**
-     * set the maximum coordinate of the second block x
-     *
-     * @return the minX
-     */
-    public BlockRegion maxX(int x) {
-        Preconditions.checkArgument(x >= this.minX || this.minX == Integer.MAX_VALUE);
-        this.maxX = x;
-        return this;
+    public BlockRegion addMin(int dx, int dy, int dz) {
```
jdrueckert, line 293: @pollend would you be okay with `addToMin`?

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,550 +114,397 @@
...
     }
 
-    /**
-     * The number of blocks on the Y axis
-     *
-     * @return number of blocks in the Y axis
-     */
-    public int sizeY() {
-        return this.maxY - this.minY + 1;
+    public BlockRegion addMax(int dx, int dy, int dz) {
```
jdrueckert, line 371: @pollend would you be okay with `addToMax`?

engine/src/main/java/org/terasology/world/block/BlockRegion.java
jdrueckert: Or alternatively `moveMin`?

engine/src/main/java/org/terasology/world/block/BlockRegion.java
jdrueckert: Or alternatively `moveMax`?

jdrueckert (commented):
Still not totally happy with `addMin` and `addMax`.
Left a comment with two alternative proposals. As I know that @skaldarnar doesn't care, I hope @pollend can tell me which he would prefer if any.

pollend:
Addtomin and addtomax sounds better.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -76,15 +80,32 @@
         this.maxZ = maxZ;
     }
```
skaldarnar, line 81: This check is out of sync with `set(...)`. Can we just use 'set` in the constructor?

```java
    public BlockRegion(int minX, int minY, int minZ, int maxX, int maxY, int maxZ) {
      this.set(minX, minY, minZ, maxX, maxY, maxZ);
    } 
```

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,550 +114,397 @@
...
-     */
-    public BlockRegion minY(int y) {
-        Preconditions.checkArgument(y <= this.maxY || this.maxY == Integer.MIN_VALUE);
-        this.minY = y;
-        return this;
+    @Override
+    public BlockRegion minX(int x, BlockRegion dest) {
+        Preconditions.checkArgument(x <= this.maxX || this.maxX == Integer.MIN_VALUE);
+        dest.minX = x;
+        return dest;
```
skaldarnar, line 245: This implementation is wrong!
```suggestion
    public BlockRegion minX(int x, BlockRegion dest) {
        return dest.set(x, minY, minZ, maxX, maxY, maxZ);
```

skaldarnar: Alternatively, we could only offer these methods to modify individual components on the mutable class.

jdrueckert: Guess that's a question for @pollend

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,550 +114,397 @@
...
-     * sets the minimum block for this {@link BlockRegion}
-     *
-     * @param minX the x coordinate of the first block
-     * @param minY the y coordinate of the first block
-     * @param minZ the z coordinate of the first block
-     * @return this
-     */
-    public BlockRegion setMin(int minX, int minY, int minZ) {
+    @Override
+    public BlockRegion setMin(int minX, int minY, int minZ, BlockRegion dest) {
```
skaldarnar, line 275: Should this be called `setMin` or `min`, analogous to `minX`/`minY`?

engine/src/main/java/org/terasology/world/block/BlockRegionc.java
```diff
@@ -0,0 +1,799 @@
...
+     * @throws IllegalArgumentException if extending this region would result in any non-positive dimension
+     */
+    //TODO: why do we offer float variants here?
+    default BlockRegion expand(float dx, float dy, float dz, BlockRegion dest) {
+        return this.expand(
+                Math.roundUsing(dx, RoundingMode.FLOOR),
+                Math.roundUsing(dy, RoundingMode.FLOOR),
+                Math.roundUsing(dz, RoundingMode.FLOOR),
+                dest);
+    }
```
skaldarnar, line 479: I'm not sure this is correct :thinking: I think we are missing the `0.5f` margin to each side, e.g., 
```java
// bounding box: (.5, .5, .5)..(1.5, 1.5, 1.5)
new BlockRegion(1,1,1).expand(0.5f, 0.5f, 0.5f) == new BlockRegion(0,0,0, 2,2,2)
```

```suggestion
        return this.expand(
                Math.roundUsing(dx + 0.5f, RoundingMode.FLOOR),
                Math.roundUsing(dy + 0.5f, RoundingMode.FLOOR),
                Math.roundUsing(dz + 0.5f, RoundingMode.FLOOR),
                dest);
    }
```

jdrueckert: I don't understand why we need `expand` to accept float values in the first place.
Can a `BlockRegion` contain only parts of blocks? If not I don't see a use case in which we would need to support float input.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
pollend: Why do you need to have a dest for a mutation like this. Just modify the direct object. Simpler to set it to another object and then set min and max.

skaldarnar: Hm, can we clarify in which situations we want to have the variants with explicit destinations and in which we don't want/need them? I don't see a clear pattern here :see_no_evil:

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,550 +104,349 @@
...
-     *
-     * @return the minY
-     */
-    public BlockRegion minY(int y) {
-        Preconditions.checkArgument(y <= this.maxY || this.maxY == Integer.MIN_VALUE);
-        this.minY = y;
-        return this;
+    @Override
+    public BlockRegion minX(int x, BlockRegion dest) {
+        return dest.set(x, this.minY, this.minZ, this.maxX, this.maxY, this.maxZ);
```
pollend, line 233: A setter like this just looks kind of strange to me. all it does is write to the dest with a substitution of a single value.  I don't really seeing this adding much to the API. 

```
BlockRegion a = new BlockRegion(0,0,0);

BlockRegion b = new BlockRegion(a).minY(-1);
vs
BlockRegion b = a.minY(-1, new BlockRegion(BlockRegion.INVALID));
```

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,550 +104,349 @@
...
-        Preconditions.checkArgument(minX <= this.maxX || this.maxX == Integer.MIN_VALUE);
-        Preconditions.checkArgument(minY <= this.maxY || this.maxX == Integer.MIN_VALUE);
-        Preconditions.checkArgument(minZ <= this.maxZ || this.maxX == Integer.MIN_VALUE);
-        this.minX = minX;
-        this.minY = minY;
-        this.minZ = minZ;
-        return this;
+    @Override
+    public BlockRegion setMin(int minX, int minY, int minZ, BlockRegion dest) {
+        return dest.set(minX, minY, minZ, this.maxX, this.maxY, this.maxZ);
```
pollend, line 260: I would drop this as well.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,550 +104,349 @@
...
-        dest.minZ = minZ - .5f;
+    // -- size -------------------------------------------------------------------------------------------------------//
 
-        dest.maxX = maxX + .5f;
-        dest.maxY = maxY + .5f;
-        dest.maxZ = maxZ + .5f;
+    @Override
+    public BlockRegion setSize(int x, int y, int z, BlockRegion dest) {
+        return dest.set(this.minX, this.minY, this.minZ, this.minX + x, this.minY + y, this.minZ + z);
+    }
```
pollend, line 350: I would drop this as well. just as easy to just do `new BlockRegion(a).setSize(3,3,3)` or `b.set(a).setSize(3,3,3)` vs `a.setSize(3,3,3, new BlockRegion(BlockRegion.INVALID));`

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,550 +104,349 @@
...
-     * Get the block coordinate minimum corner.
-     *
-     * @param dest will hold the result
-     */
-    public Vector3i getMin(Vector3i dest) {
-        return dest.set(minX, minY, minZ);
+    @Override
+    public BlockRegion minZ(int z, BlockRegion dest) {
+        return dest.set(this.minX, this.minY, z, this.maxX, this.maxY, this.maxZ);
     }
```
pollend, line 252: I would drop this as well.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,550 +104,349 @@
...
-     * @return the minimum coordinate z
-     * @deprecated use {@link #minZ()}
-     */
-    @Deprecated
-    public int getMinZ() {
-        return this.minZ;
+    @Override
+    public BlockRegion minY(int y, BlockRegion dest) {
+        return dest.set(this.minX, y, this.minZ, this.maxX, this.maxY, this.maxZ);
     }
```
pollend, line 243: I would drop this as well.

pollend (changes requested):
This seems a bit overdone. the API looks a bit bloated with all the variations. Something about the setters in particular just seem kind of extraneous.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
skaldarnar: > A setter that sets the value of another object is kind of strange?

I'm REALLY confused by now o.O At first I was confused by this pattern in JOML, than you and others explained to me why that's a good thing. Now, it seems the other way around, and we should probably not offer those `dest` variants at all...

In all cases, these two expressions are equivalent:
```java
a.op(b, dest) == dest.set(a).op(b)
```

pollend: Drop the dest variations for the setters. They're simple enough for the most part not to need it. I guess that can be debatable.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -294,132 +311,171 @@ public int maxZ() {
...
-    public BlockRegion union(int x, int y, int z, BlockRegion dest) {
-        dest.minX = Math.min(this.minX, x);
-        dest.minY = Math.min(this.minY, y);
-        dest.minZ = Math.min(this.minZ, z);
-
-        dest.maxX = Math.max(this.maxX, x);
-        dest.maxY = Math.max(this.maxY, y);
-        dest.maxZ = Math.max(this.maxZ, z);
-        return dest;
-    }
```
pollend: why did you drop this?

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -294,132 +311,171 @@ public int maxZ() {
...
+        return this.union(other.minX(), other.minY(), other.minZ()).union(other.maxX(), other.maxY(), other.maxZ());
     }
 
     // ---------------------------------------------------------------------------------------------------------------//
 
-    @Override
-    public Optional<BlockRegion> intersect(BlockRegionc other, BlockRegion dest) {
-        dest.minX = Math.max(minX, other.minX());
-        dest.minY = Math.max(minY, other.minY());
-        dest.minZ = Math.max(minZ, other.minZ());
```
pollend: this one is also fine.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -294,132 +311,171 @@ public int maxZ() {
...
-    @Override
-    public BlockRegion translate(int x, int y, int z, BlockRegion dest) {
-        dest.minX = this.minX + x;
-        dest.minY = this.minY + y;
-        dest.minZ = this.minZ + z;
-        dest.maxX = this.maxX + x;
-        dest.maxY = this.maxY + y;
-        dest.maxZ = this.maxZ + z;
-        return dest;
-    }
```
pollend: you can keep this one to.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -428,14 +484,20 @@ public BlockRegion translate(Vector3ic vec) {
 
     // -- extend -----------------------------------------------------------------------------------------------------//
 
-    @Override
-    public BlockRegion expand(int dx, int dy, int dz, BlockRegion dest) {
-        return dest.set(this.minX - dx, this.minY - dy, this.minZ - dz,
-                this.maxX + dx, this.maxY + dy, this.maxZ + dz);
-    }
```
pollend: this one is also fine.

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -444,8 +506,16 @@ public BlockRegion expand(Vector3ic extents) {
 
     // -- transform --------------------------------------------------------------------------------------------------//
 
-    @Override
-    public BlockRegion transform(Matrix4fc m, BlockRegion dest) {
```
pollend: you can keep this one to.

pollend:
```
BlockRegionc rd = new BlockRegion(0,0,0,0,0,0);

rd.translate(2,2,2,new BlockRegion(BlockRegion.INVALID));
new BlockRegion(rd).translate(2,2,2);

BlockRegion temp = new BlockRegion();
for(int x = 0; x < 10; x++) {
   rd.translate(x,0,0,temp);
  
   rd.setX(1, temp);

   temp.set(rd).setX(1); 
}
```

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -93,392 +103,309 @@
...
      */
     public BlockRegion setSize(int x, int y, int z) {
-        Preconditions.checkArgument(x > 0);
-        Preconditions.checkArgument(y > 0);
-        Preconditions.checkArgument(z > 0);
-        this.maxX = this.minX + x - 1;
-        this.maxY = this.minY + y - 1;
-        this.maxZ = this.minZ + z - 1;
-        return this;
+        return this.setMax(this.minX + x, this.minY + y, this.minZ);
```
pollend, line 393:
```suggestion
        Preconditions.checkArgument(x > 0);
        Preconditions.checkArgument(y > 0);
        Preconditions.checkArgument(z > 0);
        return this.setMax(this.minX + x - 1, this.minY + y - 1, this.minZ + z - 1);
```

pollend (changes requested):

engine/src/main/java/org/terasology/world/block/BlockRegion.java
skaldarnar: That's right 👍 However, we don't need the preconditions here as we are calling `setMax` which will check conditions (so no need to duplicate the code).

engine/src/main/java/org/terasology/math/ChunkMath.java
```diff
@@ -575,67 +574,67 @@ public static BlockRegion getEdgeRegion(BlockRegion region, Side side, BlockRegi
...
+                    region.minY(),
+                    region.minZ()
                 ).setMax(
-                    region.getMaxX(),
-                    region.getMaxY(),
-                    region.getMaxZ()
+                    region.maxX(),
+                    region.maxY(),
+                    region.maxZ()
                 );
```
pollend, line 638: for these you need to use `set` the dest can become invalid between setMin and setMax

pollend (changes requested):

engine/src/main/java/org/terasology/math/ChunkMath.java
skaldarnar: Done 👍

pollend (approved):

PR #4328, opened 2020-12-22
fix(publish): don't try to figure out the repository to publish to until after the version is set

keturn:
From [#logistics](https://discord.com/channels/270264625419911192/633681387995660303/790717924422844466):

> oddly project.version is not working in publish.gradle for the engine (returns "unspecified")

This PR moves that code in to an `afterEvaluate` block to give it a chance to load the version first. This is the solution offered by the [gradle manual](https://docs.gradle.org/6.7.1/userguide/publishing_maven.html#publishing_maven:deferred_configuration).

### How to test

`gradlew :engine:publish` is the task,  
the condition is when a `publishRepo` gradle property is **not** set.

[fixme: add more details about how to test and what to look for]

### DIFF
--- a/config/gradle/publish.gradle
+++ b/config/gradle/publish.gradle
@@ -25,15 +25,22 @@ publishing {
                     deducedPublishRepo = "terasology"
                 }
 
-                if (project.version.toString().endsWith("-SNAPSHOT")) {
-                    deducedPublishRepo += "-snapshot-local"
-                } else {
-                    deducedPublishRepo += "-release-local"
-                }
+                afterEvaluate {
+                    // depends on project.version, so must be delayed until project is configured
+                    if (!project.version || project.version == "unspecified") {
+                        throw new GradleException("Project ${project} does not have a version set yet: ${project.version}")
+                    }
 
-                logger.info("The final deduced publish repo is {}", deducedPublishRepo)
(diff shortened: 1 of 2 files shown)


keturn:
The cumulative diff is messy in GitHub due to a refactor. The functional change is in commit 0d635e23bd63f747831418859605dd630d5aa742.

Cervator (approved):
Can confirm the fix from local testing 👍 Thanks!

keturn:
Wondering how to test this, I imagined there would be some sort of gradle report to run that would show its repositories. But I don't see it showing up in `components` or `outgoingVariants` or any of the other reports.

I did `gradlew :engine:publish --info --dry-run`

and there are log lines in there that say
> The final deduced publish repo is terasology-snapshot-local

but it doesn't show the URL, and it's only there because we have added that log line ourselves, and it happens during configuration phase so the one for `:engine` is mixed in with alllll the other sub-projects (which is a lot if you have modules checked out).

Cervator:
I just tested it by running the publish normally knowing it would fail from no credentials, heh. Brute force, simple, works (by failing successfully!)

Would indeed be nice to see all the target URLs nicely lined up. At one point it did at least get logged in a nice block when a publish succeeded, but I haven't seen that for a little while (?) and that doesn't really help troubleshoot failures either.

PR #4329, opened 2020-12-22
DISCUSSION: API for BlockRegion

skaldarnar:
Related to #4321 and #4326.

### Design Goals 
- consistent API, similar to JOML
- support all known use cases (minimum) / prepare for future use cases
- (almost) every instance of a `BlockRegion` is valid
- modifying `BlockRegion` so that it becomes invalid is prevented by checks, throwing exceptions
- no splits over multiple files/locations

### Content
- Propose an interface for `BlockRegion`
- Remove `BlockRegions` utility class
- Remove `BlockRegionIterable` (to be inlined in `BlockRegion`, only one variant)


engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public BlockRegion expand(int dx, int dy, int dz);
+    public BlockRegion expand(Vector3ic extents);
+
+    // -- world -----------------------------------------------------------------------------------------------------//
+    public AABBf getBounds(AABBf dest);
+    public Vector3f center(Vector3f dest);
+
+    // -- IN-PLACE MUTATION -----------------------------------------------------------------------------------------//
+    // -- union -----------------------------------------------------------------------------------------------------//
```
skaldarnar, line 93: I removed the union with `EntityRef` as that felt quite arbitrary to be added only here. 

The need to have something like this is probably caused by lack of convenience on `ComponentContainer`.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,145 @@
...
+    // -- size ------------------------------------------------------------------------------------------------------//
+    public int getSizeX();
+    public int getSizeY();
+    public int getSizeY();
+    public Vector3i getSize(Vector3i dest);
+
+    public BlockRegion setSize(int sizeX, int sizeY, int sizeZ);
+    public BlockRegion setSize(Vector3ic size);
+
+    public int size();                                              // what to call this? size, volume, blockCount, ...
```
skaldarnar, line 83: This question still holds - what should this be called?

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,145 @@
...
+import org.joml.Vector3fc;
+import org.joml.Vector3i;
+import org.joml.Vector3ic;
+import org.terasology.entitySystem.entity.EntityRef;
+
+import java.util.Iterator;
+import java.util.Optional;
+
+public interface IBlockRegion {
+    public static BlockRegion invalid();
```
skaldarnar, line 22: Something like this required to build the union over multiple "things" (positions, regions, ...) into a new region (initial value).

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,145 @@
...
+
+    // -- CONSTRUCTORS ----------------------------------------------------------------------------------------------//
+    BlockRegion();
```
skaldarnar, line 25: I kept this non-public to avoid creation of invalid regions without typing in "invalid" to do so.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,145 @@
...
+    public BlockRegion(int minX, int minY, int minZ, int maxX, int maxY, int maxZ);
+    public BlockRegion(Vector3ic min, Vector3ic max);
+    public BlockRegion(int x, int y, int z);
+    public BlockRegion(Vector3ic block);
+    public BlockRegion(BlockRegion source)
+
+    // -- ITERABLE -------------------------------------------------------------------------------------------------//
+    public Iterator<Vector3fc> iterator();
```
skaldarnar, line 33: This is the in-place variant, which may require changes on the caller site if they persist the positions:

```java
for (Vector3ic p: region) {
	Vector3i pos = new Vector3i(p);
    // do stuff
}
```

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,145 @@
...
+    public Vector3f center(Vector3f dest);
+
+    // -- IN-PLACE MUTATION -----------------------------------------------------------------------------------------//
+    // -- union -----------------------------------------------------------------------------------------------------//
+    public BlockRegion union(int x, int y, int z);
+    public BlockRegion union(Vector3ic pos);
+    public BlockRegion union(BlockRegion other);
+
+    // -- intersect -------------------------------------------------------------------------------------------------//
+    public Optional<BlockRegion> intersect(BlockRegion other);      // this can leave the region in invalid state!
```
skaldarnar, line 99: This is the only operation that can lead to an invalid state, I think...

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,142 @@
...
+    public BlockRegion maxY(int y);
+
+    public int maxZ();
+    public BlockRegion maxZ(int z);
+
+    public Vector3i getMax(Vector3i dest);
+
+    public BlockRegion setMax(int x, int y, int z);
+    public BlockRegion setMax(Vector3ic max);
+    public BlockRegion translateMax(int dx, int dy, int dz);
```
pollend, line 71: Why do we need translation for min and max. Operation is easily achieved with getting the min with with vector and doing an add.  Just rename it to addMin/addMax

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,142 @@
...
+
+    // -- intersect -------------------------------------------------------------------------------------------------//
+    public Optional<BlockRegion> intersect(BlockRegion other);      // this can leave the region in invalid state!
+
+    // ---------------------------------------------------------------------------------------------------------------//
+    public BlockRegion translate(int dx, int dy, int dz);
+    public BlockRegion translate(Vector3ic vec);
+
+    // -- expand -----------------------------------------------------------------------------------------------------//
+    public BlockRegion expand(int dx, int dy, int dz);
```
pollend, line 103: Can we keep with aabb naming. I think it's addExtent.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
pollend: Volume is more descriptive then size.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
pollend: No reason to have a method for invalid.  Either make the constructor methods public or define as a constant in blockregion.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,142 @@
...
+import org.joml.Vector3f;
+import org.joml.Vector3fc;
+import org.joml.Vector3i;
+import org.joml.Vector3ic;
+import org.terasology.entitySystem.entity.EntityRef;
+
+import java.util.Iterator;
+import java.util.Optional;
+
+public interface IBlockRegion extends Iterable<Vector3ic> {
```
pollend, line 21: Just use BlockRegionc. Keep it similar to joml.

pollend (changes requested):

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
```diff
@@ -0,0 +1,142 @@
...
+    // -- CONSTRUCTORS ----------------------------------------------------------------------------------------------//
+    BlockRegion();
+    public BlockRegion(int minX, int minY, int minZ, int maxX, int maxY, int maxZ);
+    public BlockRegion(Vector3ic min, Vector3ic max);
+    public BlockRegion(int x, int y, int z);
+    public BlockRegion(Vector3ic block);
+    public BlockRegion(BlockRegion source)
+
+    // -- ITERABLE -------------------------------------------------------------------------------------------------//
+    public Iterator<Vector3ic> iterator();
```
pollend, line 33: What kind of iteration is this?

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
skaldarnar: Making `new BlockRegion()` would undermine the design principle to ensure that regions are valid.

Having a static `BlockRegion.INVALID` without an immutable interface would allow to change `BlockRegion.INVALID` itself - which means it would not be a constant...

That's why I went for a static method here.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
skaldarnar: I also considered `addMax` (and potentially `subMax`) as alternatives here. I'm fine with renaming it :+1: 

The use case I had in mind here was to extend a given region upwards by a few blocks:
```java
region.addMax(0,2,0)
```
as opposed to
```java
region.setMax(region.maxX(), region.maxY() + 2, region.maxZ())
// .. or when instantiating a temporary Vector3i
region.setMax(region.getMax(new Vector3i()).add(0,2,0))
```

I think this is quite verbose and the use case of extending only in a specific direction seems reasonable to me.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
skaldarnar: Yeah, this interface is not meant to be actually in the code - it was just the easiest way to give us something to comment on specific lines :wink: 

`BlockRegionc` would also just contain a subset of these methods (accessing data and performing checks, I guess).

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
skaldarnar: See https://github.com/MovingBlocks/Terasology/pull/4329#discussion_r547258485

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
pollend: That makes sense. I don't think moving a corner is really a translation of a blockregion. It's more of a matter of semantics.

jdrueckert: > It's a translation but add is simpler to understand.

Is it though? What about the ambiguity of add a max and add _to_ a max?

jdrueckert: > I don't think moving a corner is really a translation of a blockregion.

It's not a translation of a blockregion, but a translation of the max point.

skaldarnar: > I don't think moving a corner is really a translation of a blockregion
Yep, it's rather only translating one of the corners...

> What about the ambiguity of adding a (max) point
Hm, I'd think this would only be ambiguous if block regions would have a notion of adding things to it (under that name).

pollend: Sounds strange to me to refer to it as a translation. I guess it makes sense.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
pollend: BlocRegionc is a read-only interface.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
pollend: C in blockregionc is constant. So invalid is guarded by the interface.

engine/src/main/java/org/terasology/world/block/IBlockRegion.java
pollend: might make sense to just have a method return an iterable if we want to have both variations

pollend:
ref: https://github.com/MovingBlocks/Terasology/pull/4326

PR #4330, opened 2020-12-22
chore(gradle): remove mavenLocal from project repositories

keturn:
Gradle's user guide [discourages the use of mavenLocal](https://docs.gradle.org/6.7.1/userguide/declaring_repositories.html#sec:case-for-maven-local). The section heading is “The case for mavenLocal()” but the text is thoroughly _against_ its use whenever possible.

I believe this project is now working sufficiently well with [included builds](https://github.com/MovingBlocks/Terasology/blob/5cfb0c66f5457d16cffe1246a697314ccc1cd328/libs/subprojects.settings.gradle#L4) that it no longer requires mavenLocal —  
but I hope for input on that, because I don't know everyone's workflows.

### Contains

Removes `mavenLocal` as a place to check for dependencies.

### How to test

Test your workflows that depend on in-development versions of libraries, e.g. TeraNUI or gestalt.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,7 +74,6 @@ repositories {
     // External libs - jcenter is Bintray and is supposed to be a superset of Maven Central, but do both just in case
     jcenter()
     mavenCentral()
-    mavenLocal()
 
     // MovingBlocks Artifactory instance for libs not readily available elsewhere plus our own libs
     maven {
--- a/config/gradle/common.gradle
+++ b/config/gradle/common.gradle
@@ -1,18 +1,5 @@
-/*
- * Copyright 2020 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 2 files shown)


skaldarnar (approved):
This screwed up my local build of couple of times in the past. I think adding `mavenLocal()` should be a means for developers if they try out fancy new stuff (e.g., testing out some PR to a third party lib).

This can easily be stated in a PR description, or even committed to the PR with a note to remove this line again before merging.

Thus, LGTM! :+1:

pollend:
can we just comment it out with a comment instead of removing the line. @Cervator

Cervator:
We could, but on the other hand in the rare time we might want this again whomever probably can just add it back in. If it is there but commented out it may be tempting to comment it back in to see if something works then forget it and accidentally slip it back into a commit? The two options aren't very far apart but I'm fine just letting it go for now :-)

Cervator (approved):

keturn:
Oh, I'm happy with this resolution, though I did mean to do the comment-thing as per request.

This is the documentation that I wrote for it https://github.com/MovingBlocks/Terasology/wiki/Using-Locally-Developed-Libraries

PR #4334, opened 2020-12-27
chore: remove SubSampledNoise2D/3D

pollend:
I have an omega workspace and it appears we are not using SubSampleNoise2D/3D anywhere so I'm pretty sure we can just drop this from the engine.


skaldarnar (approved):
Sounds good to me, they have been deprecated in favour of `SubSampledNoise` for some time now.

PR #4335, opened 2020-12-27
feat(JOML): migrate chunkview and tests cases

pollend:
Migrate chunkview as post refactor with : https://github.com/MovingBlocks/Terasology/pull/4326

### Module Changes

- [ ] https://github.com/Terasology/FlowingLiquids/pull/24


engine/src/main/java/org/terasology/world/internal/ChunkViewCoreImpl.java
```diff
@@ -211,24 +203,24 @@ public void setExtraData(int index, int blockX, int blockY, int blockZ, int valu
...
-        Vector3i minPos = new Vector3i(region.min());
+    public void setDirtyAround(BlockRegionc region) {
+        Vector3i minPos = region.getMin(new Vector3i());
         minPos.sub(1, 1, 1);
-        Vector3i maxPos = new Vector3i(region.max());
+        Vector3i maxPos = region.getMax(new Vector3i());
         maxPos.add(1, 1, 1);
 
         Vector3i minChunk = ChunkMath.calcChunkPos(minPos, chunkPower);
         Vector3i maxChunk = ChunkMath.calcChunkPos(maxPos, chunkPower);
```
skaldarnar, line 220: Isn't this the same as:

```java
Vector3i tmp = new Vector3i();
BlockRegion expandedRegion = new BlockRegion(region).expand(1, 1, 1);

Vector3i minChunk = ChunkMath.calcChunkPos(expandedRegion.getMin(tmp), chunkPower);
Vector3i maxChunk = ChunkMath.calcChunkPos(expandedRegion.getMax(tmp), chunkPower);
```

This might get a bit esoteric here, but I'd love to write something like:
```java
BlockRegion chunkRegion = region.expand(1, 1, 1).map(corner -> ChunkMath.calcChunkPos(corner, chunkPower);
```

And is there an alternative to `ChunkMath.calcChunkPos(pos, power)` in `Chunks`? I at least don't see one which takes vectors for both arguments when looking at `Chunks.toChunkPos`.

skaldarnar: On a second look, we introduced `toChunkRegion(BlockRegion region, Vector3ic chunkPower, BlockRegion dest)`, which I guess can (or even should) be used here:

```java
        BlockRegion expandedRegion = new BlockRegion(region).expand(1, 1, 1);
        for (Vector3ic pos : Chunks.toChunkRegion(expandedRegion, chunkPower, expandedRegion) {
            chunks[pos.x() + offset.x + chunkRegion.getSizeX() * (pos.z() + offset.z)].setDirty(true);
        }
```

skaldarnar (changes requested):

skaldarnar:
I wanted to test with JS (for FlowingLiquids) but cannot start a world...

skaldarnar (approved):

PR #4339, opened 2020-12-27
feat(JOML): migrate SubSampledNoise for BlockArea

pollend:
migrate SubSampleNoise noise to use BlockArea.


engine/src/main/java/org/terasology/utilities/procedural/SubSampledNoise.java
```diff
@@ -65,13 +67,41 @@ public float noise(float x, float y) {
...
+     */
+    @Deprecated
     public float[] noise(Rect2i region) {
         Rect2i fullRegion = determineRequiredRegion(region);
         float[] keyData = getKeyValues(fullRegion);
         float[] fullData = mapExpand(keyData, fullRegion);
         return getSubset(fullData, fullRegion, region);
     }
 
+    public float[] noise(BlockArea region) {
```
skaldarnar, line 85: As this does not modify the `region` if I see it correctly, this should be 
```suggestion
    public float[] noise(BlockAreac area) {
```

Also, let's try to call all `BlockArea`s `area` instead of `region` to avoid confusion with `BlockRegion`.

engine/src/main/java/org/terasology/utilities/procedural/SubSampledNoise.java
```diff
@@ -121,6 +173,22 @@ public float noise(float x, float y) {
...
+                int actualX = x * sampleRate + fullRegion.minX();
+                int actualY = y * sampleRate + fullRegion.minY();
+                fullData[x + y * xDim] = source.noise(zoom.x * actualX, zoom.y * actualY);
+            }
+        }
+
+        return fullData;
+    }
+
     private Rect2i determineRequiredRegion(Rect2i region) {
```
skaldarnar, line 192: Is this method still used?

skaldarnar (changes requested):

engine/src/main/java/org/terasology/utilities/procedural/SubSampledNoise.java
pollend: its an internal method so its used within the class. Am I misunderstanding the question?

skaldarnar: Hm, I think I'm confused by the duplicate code. Instead of keeping the old implementation on the TeraMath types we could completely remove it and only migrate the public entry point for this class, which is the `public float[] noise(...)` function.

engine/src/main/java/org/terasology/utilities/procedural/SubSampledNoise.java
```diff
@@ -65,13 +67,41 @@ public float noise(float x, float y) {
...
+     * @return
+     * @deprecated This is scheduled for removal in an upcoming version method will be replaced with JOML implementation
+     *     {@link #noise(BlockAreac)}.
+     */
+    @Deprecated
     public float[] noise(Rect2i region) {
         Rect2i fullRegion = determineRequiredRegion(region);
         float[] keyData = getKeyValues(fullRegion);
         float[] fullData = mapExpand(keyData, fullRegion);
         return getSubset(fullData, fullRegion, region);
```
skaldarnar, line 82: I'd implement this now in terms of the new `BlockArea`:
```suggestion
        return noise(new BlockArea(region.minX(), region.minY(), region.maxX(), region.maxY()));
```

skaldarnar (changes requested):

skaldarnar (approved):

PR #4341, opened 2020-12-27
feat(JOML): migrate to Rectanglei nui.animation

pollend:
here is a migration of some animation logic for rectanglei. from what I can tell the only screen that uses this is the debug dropdown terminal in the game. so this should be pretty simple to verify and test.


skaldarnar (approved):
The drop-down terminal just works as before -LGTM!

PR #4347, opened 2020-12-29
build(facades.PC): tidy distribution tasks

keturn:
This is a follow-up to #4343. I'm not aware of any open issues fixed by this, but it makes some build dependencies more consistent with the way #4343 manages them and addresses some TODOs that have been in the gradle files for a while.

- Removes `RemoteModuleGatherer`, the code that puts **`.jar` files in `modules/`** during gradle dependency resolution. It's unnecessary under the modules-on-the-classpath strategy used by #4343, figuring out a better way to provide those dependencies without breaking the sandbox is #4375.
- Removes PC facade targets `distApp` and `distModules`, `distPCZip`, they are unused.
  - the gradle-standard `distTar` and `distZip` tasks are supported.
- Keeps `distForLauncher` so we don't need to change any of Launcher's assumptions about the zip file layout.
- Removes some overrides to gradle/IntelliJ build directories that were counterproductive.

### Caveat

If you build a dist while you have module subprojects present, they will get lumped in on the classpath and not separated.

I think that's not a showstopper because our release builds are made without modules and then a separate process adds them to the distribution later.

### How to test

- make some zips using `:facades:PC:distZip` and `distForLauncher`; make sure they are something you can unzip and start up (outside your development workspace) using the script included in its `bin/`
- in-game look at Extras / Credits and make sure it displays the credits file.

- [ ] merge gradle upgrade #4439 
- [ ] do the start scripts need to be customized? (the only arg the old start scripts added was `-Xmx`)


Cervator:
>  if it was important for those jar files to be in that directory

So I take it the modules jars from the Gradle cache gets used instead? Two main concerns there:

* Potentially confusing - especially if anybody ever wants to manually move jars files in and out of there, expecting them to be picked up at runtime by the engine (unless of course comparable source modules are present). Likewise with the files in there + source dirs you sort of know exactly what you got, if using stuff out of the Gradle cache you'd have to look elsewhere (Gradle logging, IntelliJ project structure ..)
* _Probably_ a non-issue, but jobs in Jenkins using an engine dependency out of the Gradle cache dir has led to multiple frustrating issues due to execution starting with a working directory there - I dunno if that might happen again in any odd cases (probably mostly from alternative execution starting points like unit tests or weird utility classes)

> versionInfo.properties

Used at runtime by the engine to show metadata, but if the file ends up in the same place by some other means that should be fine. It is also used in Jenkins / by the launcher, but likely same story. 

Can test by faking some of the Jenkins env vars to trigger the file creation then look at the version label on the main menu

> Credits.md

Used to power the in-game credits screen, which we used to update manually in a duplicate file, which was a pain. Now we just copy that file to where the engine can later read it. So probably same story - so long as it ends up in the same place we should be good 👍 

Test by doing whatever thing that triggers the copy, then visiting the credits screen in-game. Without doing the thing that triggers the copy you just get a blank credits screen (or only part is missing? I forget) - if that becomes automatic with this then yay 🎉 

So I think this PR may well be enough to where we can drop `distForLauncher` - although in that case we for sure need to remember to update Jenkins :-)

keturn:
> So I take it the modules jars from the Gradle cache gets used instead? […]
>
> Potentially confusing - especially if anybody ever wants to manually move jars files in and out of there, expecting them to be picked up at runtime by the engine.

Well, I didn't remove any of that runtime code. So if you drop files in there it should behave as it does now. I'm not clear on what the use cases are that involve that.

and if we _don't_ want the gradle cache used for module runtime dependencies, then we have to put #4343 on hold and talk about requirements around that. That's where the dependency sources changed. What's being removed in this PR is the now-redundant .jar copying.

engine/build.gradle
```diff
@@ -216,26 +214,31 @@ group = 'org.terasology.engine'
 println "Version for $project.name loaded as $version for group $group"
 
 // This version info file actually goes inside the built jar and can be used at runtime
-task createVersionInfoFile {
-    inputs.property('dateTime', startDateTimeString)
-    onlyIf { env.BUILD_URL != null }
```
keturn, line 221: @Cervator This task had an "only if $BUILD_URL is set" condition on it. Is that important to keep? i.e. is a properties file with a bunch of missing entries worse than no properties file at all?

keturn:
What is this code in distForLauncher doing?

https://github.com/MovingBlocks/Terasology/blob/b1ff0bbfdfdee1beabf1c85a292c5635655682a6/facades/PC/build.gradle.kts#L265-L269

It looks like it's copying this resource out from engine and putting it in facade's resources, under the same resource path? 🤨

engine/build.gradle
Cervator: So that's the guard that means it normally only runs in Jenkins, which sets the env vars that get read. As for whether an incomplete file is worse than empty - unsure. It might list gibberish in the main menu? The file also gets attached in the open in Jenkins so the launcher can read it, but I'm not sure about the details there

Cervator:
>  I'm not clear on what the use cases are that involve that.

One perhaps contrived scenario: using the in-game module downloader will drop the retrieved jar files into `modules/` - if you now have a conflict between a hidden file from the Gradle cache vs the visible file in the module dir which one wins?

I don't know why somebody would use the module downloader in a source workspace though. I would worry that there are use cases we'll miss.

> have to put #4343 on hold and talk about requirements around that

That makes sense - and is probably partly why it is the way it is. I do remember in the long long ago also messing with the module dependencies and ending up in the Gradle cache as the "easy" option, but that having drawbacks. I can't really dig a lot of good details out of my memory though :-(

Is it not easily possible to capture the transitive dependencies but redirect them into a copy to `modules/` ?

> What is this code in distForLauncher doing?

I swear there's a reason for this but ... 😬 

I tweaked it not too long ago: https://github.com/MovingBlocks/Terasology/commit/00f44d1bdb421a6c0872b5b14a3e994050f8b07c#

But that was just from a Gradle update changing paths slightly. I don't remember why the copy is in _both_ engine and PC. Could it simply be copy paste duplication from years ago?

I kinda wish we had a Nanoware test line in the launcher, so we could merge this there, build some game zips in Jenkins, make sure they list everything fine in the UI, see them show up in the launcher and work fine. That'd be ideal. Then try to take out a thing and see if it still works.

keturn:
> One perhaps contrived scenario: using the in-game module downloader will drop the retrieved jar files into modules/ - if you now have a conflict between a hidden file from the Gradle cache vs the visible file in the module dir which one wins?

Okay, _that_ sort of thing is what I was afraid of when asking why the classpath stuff in the PC facade's build is the way it is: https://github.com/MovingBlocks/Terasology/pull/4157#issuecomment-708713746

I haven't been planning around how the runtime module downloader provides dependencies because I can't remember having seen it work in recent years.

If there's a way the game can satisfy its dependencies at runtime, there's no need to have gradle pull them at build time. But I've been assuming that's a "later" problem — that we need to be able to get dependencies resolving consistently when we have the advantages of a full dev environment before we worry about how our own package manager works.

Current momentum is to keep going down this "leave the dependencies to Gradle; that's what it's good at" course, but if you've been down this road before and think we are racing to our ***DOOM*** then we need to back up to before #4245 and course-correct.

Cervator:
In-game module downloader is at least semi-broken, yeah. Another similar thing is the server auto-downloader when a client is missing modules that the server is running. Which might also be semi-broken? Or I just keep bumping into quirky situations where the engine tries to download a different version of itself.

It may be less about the game at runtime and more about how running from a source workspace can be quirky. Normal running from a zip will be fine.

Really, we could magic-wand-fix this by telling developers "thou shalt use `groovyw module recurse`" and put big fat warnings on the `get` and `init` variants to say they are _not_ responsible for dealing smoothly with dependencies.

Or we just go this way for source workspaces (Gradle manages all dependencies, don't do _any_ jars in `modules/`) then let runtime from zip deal with pure jar files there.

I dunno. My head hurts.

keturn:
I've made some progress on the distribution-related stuff. Just need to check in with the Launcher code and see what it expects to find in distributions. At this point I'm still hoping to not require any Launcher changes.

keturn:
It looks like launcher doesn't care very much about the structure of the distribution.

GameStarter is what has the info about where to find the main class: https://github.com/MovingBlocks/TerasologyLauncher/blob/59f8a31d552bec05958d26f484bec9a418762b6a/src/main/java/org/terasology/launcher/game/GameStarter.java#L52-L53

It expects `libs/Terasology.jar`. (It uses `-jar` rather than a class name.)

It doesn't use the startup scripts, and it doesn't need the distribution to contain a file with version metadata.

I should probably also look at whatever thing it is that grabs the distForLauncher and combines it with a set of modules to make the "omega" package. 🤔

Cervator:
> I should probably also look at whatever thing it is that grabs the distForLauncher and combines it with a set of modules to make the "omega" package

That one is a non-issue. The script lives at https://github.com/Terasology/Index/blob/master/distros/distros.gradle and just reads in a flat list of modules from https://github.com/Terasology/Index/blob/master/distros/omega/gradle.properties. It uses Gradle to fetch the latest versions out of Artifactory ... which isn't ideal and takes a while. There might just be some permalink for "latest" we could use instead, or ideally we'd move on to that book-of-materials approach, and probably avoid Gradle entirely as we aren't really dealing with dependencies at that point.

It even explicitly cuts out transitive dependencies, I dunno what it might do it we forget to add a new dependency to the prop file - would Gradle find it and surprise us later when we realize the prop file doesn't match what actually ships? I'm pretty sure I just did that originally because it was convenient, not correct or particularly fast :-)

We still could run some reports and check, but really cramming some jar files into a zip doesn't seem like a thing that _requires_ Gradle.

The use or non-use of the version metadata file for the launcher is probably just an option (I raised it on chat some time ago) and I guess it isn't used yet and might never be. Even if then I figure we could just deal with a semi-empty version file in logic rather than include / not-include it like some sort of sign. Simple is good and all that 👍

keturn:
https://github.com/Terasology/Index/blob/7ade39080a194834d0cb97bb3877bf8d1a2d77d1/distros/distros.gradle#L67-L68

I guess this does expect the .zip distribution to be exactly `Terasology.zip`, not `Terasology-4.2.0.zip` or any other variant.

It starts with that, and then dumps more stuff in to `modules/` and _maybe_ `libs/`.

Okay, I think those are the details I was checking for.

Cervator:
> I guess this does expect the .zip distribution to be exactly Terasology.zip

It does, but honestly that's just a shortcut that served the need at the time. I think some zip file naming prop snuck into the distro prop files already, but it may not be in use everywhere - we should be able to just start using something like that and not hard code the name anywhere

keturn:
other significant detail:

gradle's default distribution-builder puts everything inside a containing folder with the name of the project so things are like `Terasology-4.2.0/libs/foo.jar`, whereas our current dist tasks put `libs` and everything else at the top level of the archive.

Personally, I like having the containing folder so unzipping an archive doesn't barf out a bunch of files in to your current directory, but I think that's out-of-scope for this ticket because it would require coordination with Index and Launcher.

keturn:
I've tried to collect the notes we made here about what we expect from build results and classpaths and whatnot on https://github.com/MovingBlocks/Terasology/wiki/Supported-Development-Workflows

keturn (commented):
I think this is ready to take out of Draft.

facades/PC/build.gradle.kts
```diff
@@ -217,103 +239,78 @@ tasks.register<Copy>("createVersionFile") {
...
+    val launcherLibraryDirectory = "libs"
+
+    this.with(distributions.getByName("main").contents {
+        eachFile {
+            val pathSegments = relativePath.segments
+
+            when (pathSegments[0]) {
+                defaultLibraryDirectory -> {
+                    // Redirect things from lib/ to libs/
+                    val tail = pathSegments.sliceArray(1 until pathSegments.size)
```
Cervator, line 260: This looks pretty fancy. I trust the comment but have no idea how the magic works.

facades/PC/build.gradle.kts
```diff
@@ -69,22 +67,27 @@ val displayVersion = versionBase
 
 
 application {
+    applicationName = "Terasology"
     mainClass.set(extra.get("mainClassName") as String)
 }
 
-// Adjust as the Gradle 6 upgrade changed this path a bit
-sourceSets {
```
Cervator, line 76: I guess this doesn't get in the way of anything? Like runtime hot reloading or stuff like that. Not really sure what this would affect. Assets changing at runtime is one thing but unsure if that'd be affected.

Cervator (changes requested):
This looks pretty good, lots of nice cleanup 👍 

Tried to test locally. Game runs fine with `gradlew jar game` as usual. Module picker screen showed the classpath-borne modules fine _and_ the `engine-tests` piece as a module for some reason:

image

Under module details both tests and the regular engine were shown, but as a metadata piece there that might be fine and natural, but a bit odd to have the engine tests as a pickable module.

Tried to run a `gradlew distForLauncher` here on my Win10 which worked fine and ran fast (I guess since I did already do a `jar` but it made a somewhat slim package:

image

I see `Terasology.bat` and `Terasology` for Linux both in the `bin` dir - would it be possible / better to have them at the root? It is a fairly small package otherwise, and while it might be more natural on Linux to have them in `bin/` the Windows executables were in the root previously.

Trying to run the game with the bat did not work:

```
The input line is too long.
The syntax of the command is incorrect.
```

Not sure if that's because I had done a `groovyw module get CoreSampleGameplay` so I could test running from source, that appears to have added those jars to the generated classpath as well as dropping their jars into the `libs` directory for some reason (but I realize normally there just wouldn't be any modules - although it could be nice if that still worked for making a local dist, say if you're a casual contributor and trying to share with a friend)

Leaving it at that for now, could perhaps test a bit further but am running out of time for the day :-)

facades/PC/build.gradle.kts
keturn: facade's sources don't contain any gestalt modules, so we don't have to be as controlling when we build this as we do when we build modules.

facades/PC/build.gradle.kts
keturn: I was disappointed with how verbose my Kotlin solution for this was. This

https://github.com/MovingBlocks/Terasology/blob/9d84f8ee66ba268d0a49ce48dc8f4287027cf3d8/facades/PC/build.gradle.kts#L259-L261

would be something like this in Python:

```python
relativePath = RelativePath(true, launcherLibraryDirectory, *pathSegments[1:])
```

but I didn't figure out a way to do that kind of open-ended slice in Kotlin. I tried just `.drop()`ing the first segment, but that left me with a List instead of an Array and I seem to need Array to use the spread operator with the varargs argument.

keturn:
> a somewhat slim package

is there supposed to be more in there? oh, that's right, I was comparing a list from the current implementation and this one, that's what set me off on the whole `libs` vs `lib` thing. That took a _while_ to straighten out and I didn't go back to checking the diff after I was done.

>  in the bin dir - would it be possible / better to have them at the root? 

hmm, I'll have a look at that. I know there's configuration to change the name `bin` to something else, but that's not quite the same as not making a directory for them at all.

> Trying to run the game with the bat did not work:
>
> > The input line is too long.

Ugh. This is https://github.com/gradle/gradle/issues/1989
Well, there are plenty of comments from the last six or seven years with workarounds we can try. Using a wildcard seems like the most straightforward thing, as I don't expect the ordering of things in that generated classpath to be significant. (famous last words)

>  I had done a `groovyw module get CoreSampleGameplay` so I could test running from source, that appears to have added those jars to the generated classpath as well as dropping their jars into the libs directory for some reason

Yep, since #4343 included modules as runtime dependencies, they get packaged as runtime dependencies.
and then you get the test stuff in there because some modules have an "optional" dependency on things they require for their tests, but module.txt doesn't have a way to separate test dependencies from runtime dependencies.

keturn:
On the "line is too long" classpath thing:

Oh, one of the alternate approaches for this is putting the classpath in the jar manifest instead of on the command line. That's probably why we put a classpath in a jar manifest!

https://github.com/MovingBlocks/Terasology/blob/9d84f8ee66ba268d0a49ce48dc8f4287027cf3d8/facades/PC/build.gradle.kts#L209-L212

…

gradle does _so many things,_ java projects are its home turf, this seems like a very standard java feature and it's clearly a use case that comes up a lot. How does it not have an option for this built-in??? This is one of the sections of our config that I've been giving the side-eye, thinking "this looks like too much implementation detail, there's probably a standard implementation to use instead." Only in this case, there isn't one.

I'm not even sure I trust the gradle plugins I'm seeing for it. Looks like one of those things that someone puts in a plugin every few years in a "it works for me" state, but there's no continuity as gradle changes or people find edge cases. Bother.

Okay, enough venting for now.

- Terasology.jar's manifest already contains the classpath. (I'm concerned whether everything is escaped properly, but that can wait.)
- we want start scripts that
  - do **not** attempt to include a classpath in a command line or an environment variable. At best, it will be redundant, and it's likely to break as we've seen.
  - **do** launch java with `-jar`.

This puts a real big hole in my "we're simplifying things by using the default start script!" argument, but I do still like what the gradle-provided scripts bring in honoring the `JAVA_HOME` environment variables and the like.

Shall we store patch files for these scripts, as [launcher does](https://github.com/MovingBlocks/TerasologyLauncher/tree/e5e42fbb2679ea8594e7d71e9a1d3396681b3457/buildres/scripts)? except I think I want to patch the templates, not the end result.

This is a rabbit hole. Before I get lost in it any further, I should check: if you build the dist _without modules,_ is the classpath short enough? If that's the case we can wait on this. though it might depend on the length of your `APP_HOME`?

keturn:
> is there supposed to be more in there? oh, that's right, I was comparing a list [of files contained in the zip] from the current implementation and this one

I've checked this again and it looks like the thing Cerv noted about having the executables in the top level instead of `bin/` is the only difference. That's a bit of good news.

Cervator:
Same failure even without modules, yeah it is the app home. I was starting out of `D:\Dev\Terasology\Git_WS\GradleMagic\Terasology\facades\PC\build\distributions\Terasology\bin` which is a tad longish.

Starting out of `D:\Terasology\bin` worked fine 👍 

And yeah looks like the main (only?) change is that the executables used to be at the root, not in a `bin/`

Although - I will say that I tended to edit the really short executables when doing play test prep and such to tweak the memory settings, maybe add the `-override` flag and other small things. The standard executable script is a little, uh, heavy for that. You could just pass the parameters, of course, it was just handy to edit the script once, snapshot the play test server, then restore it later and reuse the modified script. That's a pretty small convenience though, could just _make_ a script that calls the script, I guess. 

Or could we perhaps leave an optional line at the very top of the generated script for plugging in _standard_ extra arguments? If that isn't immediately obvious then don't worry about it.

> I'm not even sure I trust the gradle plugins I'm seeing

Agreed - hopefully we can do without

> This puts a real big hole in my "we're simplifying things by using the default start script!" argument, but I do still like what the gradle-provided scripts bring in honoring the JAVA_HOME environment variables and the like.

Curiously, would this at all interfere with the launcher attempting to force the use of the JRE it ships with?

> Shall we store patch files for these scripts

I guess? Or the template thing. No real preference there.

keturn:
> Curiously, would this at all interfere with the launcher attempting to force the use of the JRE it ships with?

When Launcher starts Terasology, it starts java directly. Bypasses these scripts entirely.

keturn:
I got it set up to use the jar manifest and not put a huge `CLASSPATH=` line in the scripts.

I also have the scripts in the top-level directory, which works fine for distForLauncher but for the standard `distZip` target it overshoots and puts them outside the `Terasology-4.2.0/` container directory.

I thought I was going to fix that last little thing before putting it up for re-review, but I didn't get to that yet this week so I figured I ought to at least note it in this status update.

facades/PC/build.gradle.kts
```diff
@@ -3,14 +3,15 @@
 
 // The PC facade is responsible for the primary distribution - a plain Java application runnable on PCs
 
+import Terasology_dist_gradle.ValidateZipDistribution
```
keturn, line 6: That's not pretty.

We're using [precompiled script plugins](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:precompiled_plugins) to cut down on the amount of config we have to use. It sets the gradle plugin name nicely, but this import path is gross. I wonder if that's something we can change without giving up its convenience.

Or is there a way we can have our Task class auto-imported like the core gradle Tasks seem to be?

keturn (commented):
Now with tests!

How should we have CI run these tests? Unfortunately they're not nicely packaged tests that will result in a junit.xml file, the tasks just fail if there's a problem.

Make the general `test` task depend on `testDist`, or change Jenkinsfile to add a `testDist` step?

They do take a bit to run, because they involve zipping up everything (including dependencies), which is why I'm a little hesitant to add them to `test`.

facades/PC/build.gradle.kts
keturn: hmm. not seeing much to do about this. I can put a namespace in front of it like `org.terasology.foo.` but it still involves that `_gradle` part.

I'm inclined to leave it be until we get to the point where we're packaging our gradle plugins for distribution for use outside this multiproject.

buildSrc/src/main/kotlin/terasology-dist.gradle.kts
```diff
@@ -0,0 +1,40 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+@Suppress("UnstableApiUsage")
+open class ValidateZipDistribution @Inject constructor(objects: ObjectFactory) : DefaultTask() {
```
keturn, line 5: It would be worthwhile to add a bit of documentation for this Task subclass.

facades/PC/build.gradle.kts
keturn: try [`@JvmName`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-name/)

keturn: hmm, will try later, not worth changing generated filenames at this point in the PR, as we already have another PR waiting on this one.

Cervator (dismissed):
I left my big review at https://github.com/MovingBlocks/Terasology/pull/4454#pullrequestreview-581987662 which I figure to also include this one - I'd just merge the other one to get it all in one go 👍

keturn:
Oh, I recommend merging this separately from that.The grade 6.8 upgrade and the reimplementation of `distForLauncher` with its  change to `.bat` for the Windows distribution  is more than sufficient for one PR, and don't need to wait for finishing up the module dependency stuff.

but I'll look at what that review has to say and see how much of it applies to this

keturn:
From @Cervator's review:

> Finally and another functional one - out of curiosity: with the new Terasology.bat being generated, does it include setting default memory stats like the old one did?

Nope, there's an open note about that here. It doesn't set `-Xmx` and the JVM goes with whatever its heart wants or something.

That is one functional difference, but since the hard-coded 1.5 GB wasn't exactly a one-size-fits-all value, I didn't know if it was worth carrying over to this version.

facades/PC/build.gradle.kts
```diff
@@ -200,12 +200,41 @@ tasks.register<JavaExec>("server") {
...
+    archiveFileName.set("Terasology.jar")
+
+    manifest {
+        //TODO: Maybe later add the engine's version number into here?
+        attributes["Main-Class"] = mainClassName
+        // A classpath in the manifest avoids the problem of having to put a classpath on the command line and
+        // "line is too long" errors: https://github.com/gradle/gradle/issues/1989
+        attributes["Class-Path"] = configurations["runtimeClasspath"].joinToString(" ") { it.name }
+        attributes["Implementation-Title"] = "Terasology-" + project.name
+        attributes["Implementation-Version"] = """${env["BUILD_NUMBER"]}, ${env["GIT_BRANCH"]}, ${env["BUILD_ID"]}"""
```
keturn, line 221: https://github.com/MovingBlocks/Terasology/pull/4454#discussion_r569121967

GIT_BRANCH no longer used for this

PR #4348, opened 2020-12-29
feat(JOML): migrate BulletPhysics

pollend:
Here is the migration of the LocationComponent logic in BulletPhysics. I'm wondering about all these finite checks if we should both with it. I'm wondering if it would be better to add a Precondition to LocationComponent to check for this and drop all these finite checks littering the engine.

### DIFF
--- a/engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
+++ b/engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
@@ -324,7 +324,7 @@ public class BulletPhysics implements PhysicsEngine {
         RigidBodyComponent rb = entity.getComponent(RigidBodyComponent.class);
         BulletRigidBody rigidBody = entityRigidBodies.get(entity);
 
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/physics/bullet/BulletPhysics.java
```diff
@@ -324,7 +324,7 @@ public boolean updateRigidBody(EntityRef entity) {
         RigidBodyComponent rb = entity.getComponent(RigidBodyComponent.class);
         BulletRigidBody rigidBody = entityRigidBodies.get(entity);
 
-        if (location == null || Float.isNaN(location.getWorldPosition().x)) {
```
skaldarnar, line 327: We don't have the precondition check yet - do you think it's safe to remove the finite check here?

pollend: It doesn't actually use the worldPosition the actual position update happens in EntityMotionState.

skaldarnar (approved):

PR #4351, opened 2020-12-29
feat(JOML): migrate BlockEntitySystem

pollend:
This is a pretty straightforward migration of the BlockEntitySystem.

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
+++ b/engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
@@ -15,7 +15,9 @@
  */
 package org.terasology.world.block.entity;
 
+import org.joml.RoundingMode;
 import org.joml.Vector3f;
+import org.joml.Vector3i;
 import org.joml.Vector3ic;
 import org.terasology.audio.AudioManager;
 import org.terasology.audio.StaticSound;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
```diff
@@ -88,12 +87,12 @@ public void doDestroy(DoDestroyEvent event, EntityRef entity, ActAsBlockComponen
     @ReceiveEvent(priority = EventPriority.PRIORITY_LOW)
     public void doDestroy(DoDestroyEvent event, EntityRef entity, BlockComponent blockComponent) {
         commonDestroyed(event, entity, blockComponent.block);
-        worldProvider.setBlock(blockComponent.position, blockManager.getBlock(BlockManager.AIR_ID));
+        worldProvider.setBlock(blockComponent.getPosition(new Vector3i()), blockManager.getBlock(BlockManager.AIR_ID));
```
skaldarnar, line 90: This makes we wonder whether `BlockComponent` should just expose `Vector3ic getPosition()` for read-only access 🤔

skaldarnar (approved):

engine/src/main/java/org/terasology/world/block/entity/BlockEntitySystem.java
pollend: that's ideally what I would like to do.

skaldarnar: I wouldn't do it in this PR, but I might just make another PR for that today (after getting through some more JOML PRs 🙃 )

LGTM as is 👍

PR #4353, opened 2020-12-29
Mark new chunks ready before triggering OnChunkLoaded events.

4Denthusiast:
Because the `OnChunkLoaded` event is created in world generation threads, rather than the main thread, it usually executes after `processReadyChunk` is finished, but occasionally it doesn't. The chunk only becomes readable (i.e. `getChunk` and `getBlock` will start working inside its bounds) when `chunk.markReady` is called, which used to happen at the end of `processReadyChunk`. This meant that occasionally, the event handler for `OnChunkLoaded` would not actually be able to read that chunk. This led to issues like that `NullPointerException` in `VoxelWorldSystem` and some chunks being invisible, both of which should now be fixed. (A similar issue sometimes still happens sometimes when a world is unloaded: a chunk's `OnChunkLoaded` event may not run until after the chunk has been unloaded, but that matters less because it's being unloaded anyway.)

Moving the entirety of `processReadyChunk` to the main thread would likely make this area less error prone, but I'm a lot less confident that that change wouldn't break anything else. The small change that I have made has the virtue that what happens with the change is pretty much what used to happen already most of the time, so we can be very confident that it makes things more correct.

### DIFF
--- a/engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
+++ b/engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
@@ -197,6 +197,7 @@ public class LocalChunkProvider implements ChunkProvider {
             return; // TODO move it in pipeline;
         }
         chunkCache.put(chunk.getPosition(), chunk);
+        chunk.markReady();
         //TODO, it is not clear if the activate/addedBlocks event logic is correct.
         //See https://github.com/MovingBlocks/Terasology/issues/3244
         ChunkStore store = this.storageManager.loadChunkStore(chunk.getPosition());
@@ -244,7 +245,6 @@ public class LocalChunkProvider implements ChunkProvider {
             worldEntity.send(new OnChunkGenerated(chunk.getPosition()));
         }
         worldEntity.send(new OnChunkLoaded(chunk.getPosition()));
-        chunk.markReady();
     }
 
     private void generateQueuedEntities(EntityStore store) {
--- a/engine/src/main/java/org/terasology/world/chunks/remoteChunkProvider/RemoteChunkProvider.java
+++ b/engine/src/main/java/org/terasology/world/chunks/remoteChunkProvider/RemoteChunkProvider.java
@@ -80,12 +80,12 @@ public class RemoteChunkProvider implements ChunkProvider {
                                 .collect(Collectors.toSet())
                 ))
                 .addStage(ChunkTaskProvider.create("", chunk -> {
-                    listener.onChunkReady(chunk.getPosition());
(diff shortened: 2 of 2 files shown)


engine/src/main/java/org/terasology/world/chunks/remoteChunkProvider/RemoteChunkProvider.java
```diff
@@ -81,12 +81,12 @@ public RemoteChunkProvider(BlockManager blockManager, LocalPlayer localPlayer) {
                 ))
                 .addStage(ChunkTaskProvider.create("", chunk -> {
                     listener.onChunkReady(chunk.getPosition());
-                    worldEntity.send(new OnChunkLoaded(chunk.getPosition()));
```
keturn, line 84: You've moved the call to the worldEntity until after markReady, but what of the `listener.onChunkReady` before it? Should `onChunkReady` also be called only after the chunk is marked ready?

engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -197,6 +197,7 @@ private void processReadyChunk(final Chunk chunk) {
             return; // TODO move it in pipeline;
         }
         chunkCache.put(chunk.getPosition(), chunk);
+        chunk.markReady();
```
keturn, line 200: This one took a little more reading than RemoteChunkProvider because of the number of lines in this function, but I think I believe this. We should mark the chunk as ready before sending any of those events to blocks and entities in the chunk, or the loaded/generated events about the chunk itself.

That wouldn't be the case if something operating on that collection of event mappings is a precondition for chunk readiness -- but I don't see any evidence of that here. It just fires off a bunch of events, there are no conditions on any of them being handled any particular way.

4Denthusiast: Also, because this isn't in the main thread, the event handlers aren't guaranteed to run before the chunk is marked ready anyway (and they usually don't).

engine/src/main/java/org/terasology/world/chunks/remoteChunkProvider/RemoteChunkProvider.java
4Denthusiast: You're right. That also looks like it should be moved.

4Denthusiast: The only actual implementation of `onChunkReady` is in `ServerImpl`, and that does assume that the chunk can already be written to.

keturn (approved):
:+1:

keturn:
Seems good. I'll merge once the CI checks agree.

PR #4365, opened 2021-01-02
chore: Couple minor tweaks - logic hardening and memory

Cervator:
Part of some minor observations from recent testing.

* Guards against NPE in some behavior tree code, briefly discussed with @casals as a case where some state might get lost. Ideally we'd also hook up a special monitor for this via telemetry to get more intel about when it happens, how
* Increases default memory when running from Gradle - flying around in settings like Metal Renegades does not appreciate being memory starved. Goes beyond the 32-bit JVM capacity, but we honestly could have probably done that years ago, playing the game on 32-bit is one thing, developing it a whole other thing.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
+++ b/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
@@ -102,6 +102,12 @@ public class BehaviorSystem extends BaseComponentSystem implements UpdateSubscri
         Iterable<EntityRef> entities = entityManager.getEntitiesWith(BehaviorComponent.class);
         for (EntityRef entity : entities) {
             BehaviorComponent behaviorComponent = entity.getComponent(BehaviorComponent.class);
+            // NPE observed in the past, suspected to be about loss of behavior state. Hopefully one skip is OK then resume next tick?
+            // TODO: Highlight this log entry to the telemetry system to gather better data over time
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
```diff
@@ -102,6 +102,12 @@ public void update(float delta) {
         Iterable<EntityRef> entities = entityManager.getEntitiesWith(BehaviorComponent.class);
         for (EntityRef entity : entities) {
             BehaviorComponent behaviorComponent = entity.getComponent(BehaviorComponent.class);
+            // NPE observed in the past, suspected to be about loss of behavior state. Hopefully one skip is OK then resume next tick?
+            // TODO: Highlight this log entry to the telemetry system to gather better data over time
+            if (behaviorComponent.interpreter == null) {
```
skaldarnar, line 107: Oh, the behavior stuff is still part of the engine o.O

skaldarnar (approved):

engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
Cervator: I don't really remember why that particular piece is still in there 🤔

Issue #4362, opened 2021-01-01
Game crash on `updateSkeleton` - out of bounds a bit deeper in


Cervator:
### What you were trying to do

Test MetalRenegades. Flew fast with `ghost` + `hspeed`

Maybe we can harden or better log some of the related code?

### What actually happened

Crashed after less than a minute - something seemingly went wrong when a creature spawned and needed to be updated?

Seemingly hit some ugly chunk errors afterward as well, unless there's a possibility that was the trigger - only one millisecond later than the skeleton update, and likely in different threads?

### How to reproduce

* Start MetalRenegades
* `hspeed` and fly around fast
* Get lucky and crash with this message?

### Log details and game version

`develop` branch current as of when this issue was submitted

```
...
21:13:37.251 [main] INFO  o.t.logic.console.ConsoleImpl - [CONSOLE] Ghost mode toggled
21:13:38.310 [main] INFO  o.t.logic.console.ConsoleImpl - [CONSOLE] High-speed mode activated
21:13:41.176 [main] ERROR o.t.p.t.TypeHandlerLibrary - Unsupported field: 'DynamicCities:TreeFacet.relData'
21:13:41.264 [main] WARN  o.t.l.c.ServerCharacterPredictionSystem - Received too much input from EntityRef{id = 522, netId = 0, prefab = 'WildAnimals:sheep'}, dropping input. Delta difference: 39.0
21:13:41.988 [main] ERROR o.terasology.engine.TerasologyEngine - Uncaught exception, attempting clean game shutdown
java.lang.ArrayIndexOutOfBoundsException: 2410
        at gnu.trove.impl.hash.TLongHash.indexRehashed(TLongHash.java:233)
        at gnu.trove.impl.hash.TLongHash.index(TLongHash.java:219)
        at gnu.trove.map.hash.TLongObjectHashMap.get(TLongObjectHashMap.java:212)
        at org.terasology.persistence.internal.EntitySetDeltaRecorder.getOrCreateEntityDeltaFor(EntitySetDeltaRecorder.java:83)
        at org.terasology.persistence.internal.EntitySetDeltaRecorder.onEntityComponentChange(EntitySetDeltaRecorder.java:67)
        at org.terasology.persistence.internal.ReadWriteStorageManager.onEntityComponentChange(ReadWriteStorageManager.java:519)
        at org.terasology.entitySystem.entity.internal.PojoEntityManager.notifyComponentChanged(PojoEntityManager.java:738)
[...]

PR #4368, opened 2021-01-03
fix: Move the final stage of chunk loading back to the main thread.

4Denthusiast:
The final stage of chunk loading, where the chunk is actually added to the ChunkProvider, its entities are created, and the OnChunkLoad event runs, needs to happen on the main thread in order to avoid concurrency issues, because it affects data that other systems running on the main thread have access to. This reverts a little bit of #4140, which moved this onto the chunk loading threads in the first place.

I have actually seen some specific issues that this fixes (e.g. OnChunkLoad event handlers running after that chunk has already been unloaded), but because the relevant code (`processReadyChunk`) could overlap with anything depending on the exact timing, the problem could potentially manifest in many different ways.

### DIFF
--- a/engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
+++ b/engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
@@ -92,6 +92,7 @@ public class LocalChunkProvider implements ChunkProvider {
     private static final Logger logger = LoggerFactory.getLogger(LocalChunkProvider.class);
     private static final int UNLOAD_PER_FRAME = 64;
     private final EntityManager entityManager;
+    private final BlockingQueue<Chunk> readyChunks = Queues.newLinkedBlockingQueue();
     private final BlockingQueue<TShortObjectMap<TIntList>> deactivateBlocksQueue = Queues.newLinkedBlockingQueue();
     private final Map<Vector3i, Chunk> chunkCache;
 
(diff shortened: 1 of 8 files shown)


engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -92,6 +92,7 @@
     private static final Logger logger = LoggerFactory.getLogger(LocalChunkProvider.class);
     private static final int UNLOAD_PER_FRAME = 64;
     private final EntityManager entityManager;
+    private final BlockingQueue<Chunk> readyChunks = Queues.newLinkedBlockingQueue();
```
DarkWeird, line 95: I think `preparedChunks` sounds better.(open question)
ReadyChunks is confusing name.

DarkWeird (approved):
Nice.
Not tested.
seems all right.

engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
Cervator: `readiedChunks` maybe? To me readied and prepared would mean about the same, unless we're using either "ready" or "prepare" as stage names or something during chunk gen :-)

Cervator (approved):
Tested! Seems to fix #4362 🎉 

No real review comments, just another naming option. Nothing else sticks out and this fixes stuff so am eager to merge 👍

Cervator:
Went ahead and merged this since I could confirm it fixes #4362 which is fairly severe. Left a note in #4369 (similar chunk area bug) about possibly tweaking the naming bit with a link to the review comment here. Thanks! 👍

PR #4370, opened 2021-01-03
fix: remove reflection for gson deserilization in GsonTypeHandlerAdapter

pollend:
these interfaces don't use `serializationContext` so I just dropped them from deserilization and serilization and use the methods directly. 
```
 this.serializer = (src, typeOfSrc, context) ->
                ((GsonPersistedData) typeHandler.serialize(src, new GsonPersistedDataSerializer()))
                .getElement();

        this.deserializer = (json, typeOfT, context) ->
                typeHandler.deserializeOrNull(new GsonPersistedData(json));
```

### DIFF
--- a/engine/src/main/java/org/terasology/persistence/typeHandling/gson/GsonTypeHandlerAdapter.java
+++ b/engine/src/main/java/org/terasology/persistence/typeHandling/gson/GsonTypeHandlerAdapter.java
@@ -20,18 +20,13 @@
 package org.terasology.persistence.typeHandling.gson;
 
 import com.google.gson.Gson;
-import com.google.gson.JsonDeserializationContext;
-import com.google.gson.JsonDeserializer;
 import com.google.gson.JsonElement;
-import com.google.gson.JsonSerializationContext;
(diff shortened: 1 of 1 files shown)


eviltak (approved):

eviltak (approved):

eviltak:
The change seems very reasonable to me, and is what should have been done in the first place. As long as functionality isn't broken (I imagine the only place this is likely to be used is the UI file loader), we can go ahead and merge.

jdrueckert:
I tested it in omega and I didn't see any obvious issues. On the contrary, it fixed the
```
19:06:23.792 [main] ERROR o.t.rendering.nui.asset.UIFormat - Failed to deserialize field skin of notificationOverlay
java.lang.IllegalArgumentException: Cannot find field com.google.gson.Gson.deserializationContext
```
issue and others of this sort in my "old" omega. In a completely fresh I didn't run into those anyway anymore, but there too I don't see any obvious issues.

Not entirely sure, though, how to thoroughly check that the functionality is not broken @eviltak

PR #4371, opened 2021-01-03
Fix LocalChunkProviderTest

4Denthusiast:
Update LocalChunkProviderTest for #4368.

(I haven't actually run the test locally, because just using the PR auto-test thingy seemed more convenient, so I'll have to see how it goes.)


Cervator (approved):
Tests pass 👍

Issue #4154, opened 2020-09-19
Substantial delay in connecting a player to a multiplayer server running JoshariasSurvival


Cervator:
### What you were trying to do

Connect normally with a 2nd client to an already running game server without extensive load times. Varies based on modules enabled.

A hosting client doesn't take an inordinate amount of time standing up the server and connecting to it. Issue does not present in plain CoreSampleGameplay.

With ManualLabor enabled (identified as a culprit in #4126) connecting took a moderate 30-45 seconds for player 2. With the rest of JS enabled it went to a full 3 minutes in one test.

### What actually happened

Logging in for JoshariasSurvival_did_ succeed after #4148 was merged, and didn't slow down the host, however it took a full 2-3 minutes for no apparent constructive reason. @keturn has done some initial research noted on #4126 and on Discord (in #stabilization I think) that points to more potential issues like what #4148 fixes, that might be rooted in how Gestalt handles some stuff. So this may be a fix needed over there (either as a point release for the 5.x line or eventually by upgrading to 7.x)

Progress bar may hold some hints as to which parts are causing the delays. Asset scanning and system initialization seemed two big ones. Just trying to look real quick it seemed like about 45s for asset scanning. 1m for initializing systems for the first player to join a headless server. Think initialize systems took longer with a hosting client + client.

### How to reproduce

* Host a server via game client running JoshariasSurvival (or configure `override.cfg` and run a headless server)
* Wait for the server to be fully online and connect to it with a second client
* Observe the loading time, progress bar, and stages of game logging

keturn:
Recapping some of my ramblings from #stabilization:

Watching the connection/loading process with a profiler does reveal some slow things.

When I looked at some of those things (which do relate to asset scanning and initialization), I found some places I believe have potential for optimization.

The hitch is that I'm new to these tools for profiling multi-threaded Java applications, and I'm new to those parts of the source, and I have _very little_ surety that the things I was looking at are the same things responsible for making multiplayer so much dramatically worse than it was a release or two ago. After all, it's not like gestalt v5 changed in that time.

(though now that I mention it, it's possible we did change something about the gestalt dependency in the TeraNUI branch.)

Bisecting the history to discover when the issue started would be very useful. Not only would it provide more confidence that we were looking at the right problem, I think it would do wonders for morale and maintainability to be able to say "this is what changed and this is the impact it had" to replace my current vague dread that _something_ changed but no-one can say what.

[...]

PR #4144, opened 2020-09-07
[Breaking all] Gestalt v7 re-integration....

DarkWeird:
### Contains
 Re-integration gestaltv7

 * Changing engine module to package module. ( classpath modules dropped)
 * engine moved in self package (`org.terasology.engine` now)
 * create syntetic package module  for nui (easier then write all in whitelist)
 * Update gestalt to 7.0.5
 * Update gson to 2.8.5
 * Update TeraNui to 2.0.0
 * Update Guava to 29.0-jre
 * Update Reflections to org.terasology:reflections:0.9.12-MB (gestalt build it)
 * Reimplemented Module download (gestalt drop `BaseModule` class)
 * Restore Module loading on gestalt v7
 * Use AutoreloadingAssetManager (seems asset must reload easier now)
 * Replace condition `Module#isSourceModule` to checking Module's files source == `DirectoryFileSource` - must be equal ( affected WorldGeneratorManager)
 * add Gestalt's classes to whitelist
 * remove duplication classes from `org.terasology.input`(`org.terasology.engine.input` now). We have this classes in TeraNUI
 * move `cacheReflections` task on after `classes` task (partial resolve https://github.com/MovingBlocks/Terasology/issues/4126#issuecomment-692232675) (reflections gathering for running from sources now)
 * remove `cleanReflections` after `jar` task
 * renamed `cacheReflections`'s artifact `reflections.cache` to `manifest.json` and serializer changed to Json for using gestaltv7's default values

Fixes #2445

### Testing
1. Check annotation marked classes reloading (e.g. WorldGenerator)
   1. Run game from idea.
   2. create game, exit
   3. create dummy world generator in active module
   4. create game with advanced settings, 
   5. try found your new world generator

2. Check assets reloading
   1. Run game from idea. (in debug mode)
   2. create game
   3. change any asset of active module
   4. recompile classes 
   4. exit game and load game
   5. asset must be edited in game.

3. Check module downloading 
   1. Run game 
   2. Press New game
   3. Press Create game
   4. Press Advanced game setup
   5. try download any module.

4. Fast check for 70% sure that game works
   1. run `gradlew test`
   2. try found failed tests ;)

5. Check guava 29.0-jre :
   1. Play with MasterOfOreon ( minions uses guava's changed Futures for pathfinding)
   2. make them to use pathfinding

6.* Check what everything works fine:
   1. Play with all modules
   2. check all features
   3. if you found bug, be sure that bug from this PR :)

* - impossible

### Outstanding before merging
[...]

PR #4372, opened 2021-01-03
Bugfix/remove classpath

pollend:
this reverts the changes in: https://github.com/MovingBlocks/Terasology/commit/8befab34221667074a43a897e88939128a34fff3

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -190,10 +190,14 @@ def terasologyModules() {
 }
 
 // Helpers that do magic things after having dependencies attached below
+task moduleClasses
 task moduleJars
 
 // This magically makes everything work - without this the desired module projects returned have no tasks :-(
 gradle.projectsEvaluated {
+    // Note how "classes" may indirectly trigger "jar" for module dependencies of modules (module compile dependency)
+    moduleClasses.dependsOn(terasologyModules().classes)
+
     // This makes it work for a full jar task
     moduleJars.dependsOn(terasologyModules().jar)
 }
--- a/facades/PC/build.gradle.kts
+++ b/facades/PC/build.gradle.kts
@@ -92,7 +92,7 @@ dependencies {
 
     // TODO: Consider whether we can move the CR dependency back here from the engine, where it is referenced from the main menu
     implementation(group = "org.terasology.crashreporter", name = "cr-terasology", version = "4.1.0")
-    
+
 }
 
 // Instructions for packaging a jar file for the PC facade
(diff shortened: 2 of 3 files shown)


keturn:
Here's my current take on this:

This is not a build configuration error.

Some things shifted around and now single-player games (in certain dev environments) are running in to the performance problem we identified in multiplayer games in #4154 back in September.

For next steps, I suggest:
 * optimize this code querying subtypes in ModuleEnvironmentSandbox and/or gestalt's ModuleClassLoader. There are some [preliminary ideas about that](https://github.com/MovingBlocks/Terasology/issues/4154#issuecomment-705382655) written down already.
 * update to current gestalt (#4144), and _then_ optimize that stuff, because some of it probably changes with the new version anyway.

In the meantime, the stopgap workaround is:
* don't have a ton of modules subprojects checked out in your workspace. The performance problem climbs sharply with the number of module directories.

pollend:
I'm just using the branch to see what is different with the configuration. I don't think its a query problems. something is strange where the classloaders take a really long time to load the classes.

keturn:
yeah, the note I had on that issue that I don't think is subtype-specific is

> the other thing in ModuleClassLoader.loadClass I'm suspicious of is that I think a class loader should check to see if the class is already loaded before calling up the chain, as described for https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ClassLoader.html#loadClass(java.lang.String,boolean)
>
> but I'm not sure ModuleClassLoader checks that first.
>
> I don't yet understand the ModuleClassLoader thing of each classloader having a parent but ModuleClassLoaders also have a baseClassLoader, so I'm really not sure what order things happen in.

pollend:
> yeah, the note I had on that issue that I don't think is subtype-specific is
> 
> > the other thing in ModuleClassLoader.loadClass I'm suspicious of is that I think a class loader should check to see if the class is already loaded before calling up the chain, as described for https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ClassLoader.html#loadClass(java.lang.String,boolean)
> > but I'm not sure ModuleClassLoader checks that first.
> > I don't yet understand the ModuleClassLoader thing of each classloader having a parent but ModuleClassLoaders also have a baseClassLoader, so I'm really not sure what order things happen in.

`AccessController.doPrivileged(new ObtainClassloader(clazz));` seems to be the main bottleneck in the chain.

pollend:
with this branch the classloader isn't chained to another module class loader.

keturn:
on the security (not performance) impacts of the classpath thing: #4375

keturn:
why does this show up when running from IntelliJ but not gradle? Likely clue:

gradle is putting the _jar_ files on the classpath, but IntelliJ is giving the classes _directory_.

It's not _wrong_ to use the directories here. It might help module and asset hot-reloading speed in development and saves a build step. gradle might well do the same thing if the configuration changes a bit.

PR #4377, opened 2021-01-05
feat(TypeHandler): Add GenericMapTypeHandler for `Map<K, V>`

skaldarnar:
I wanted to use a `Map<Vector2ic, EntityRef>` in some component, and figured that this is currently not possible without writing a custom type handler. 
As converting the keys to strings is quite error-prone I thought that a generic handler for `Map<K, V>` where both `K` and `V` are arbitrary objects that can be serialized on their own would be a nice addition.

This generic map type handler encodes a generic map as list of objects representing key-value-pairs:
```jsonc
[
  { "_key": { /* encoding of K */ }, "_value": {/* encoding of V */ } },
  { "_key": { /* encoding of K */ }, "_value": {/* encoding of V */ } },
  ...
]
```

For maps with `String` keys the old type handler is used, which generates a straight-forward representation of the map as JSON object:

```jsonc
{
  "foo": { /* encoding of V */ },
  "bar": { /* encoding of V */ },
  ...
}
```


subsystems/TypeHandlerLibrary/src/main/java/org/terasology/persistence/typeHandling/coreTypes/GenericMapTypeHandler.java
```diff
@@ -0,0 +1,63 @@
...
+                .map(entry -> serializeEntry(entry, serializer))
+                .collect(Collectors.toList());
+        return serializer.serialize(entries);
+    }
+
+    private PersistedData serializeEntry(Map.Entry<K, V> entry, PersistedDataSerializer serializer) {
+        PersistedData key = keyHandler.serialize(entry.getKey(), serializer);
+        PersistedData value = valueHandler.serialize(entry.getValue(), serializer);
+
+        Map<String, PersistedData> jsonEntry = Maps.newLinkedHashMap();
```
DarkWeird, line 38: We haven't json there.

subsystems/TypeHandlerLibrary/src/main/java/org/terasology/persistence/typeHandling/coreTypes/GenericMapTypeHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public Optional<Map<K, V>> deserialize(PersistedData data) {
+        if (!data.isArray()) {
+            return Optional.empty();
+        }
+
+        Map<K, V> result = Maps.newLinkedHashMap();
+
+        for (PersistedData entry : data.getAsArray()) {
+            final Optional<K> key = keyHandler.deserialize(entry.getAsValueMap().get("_key"));
```
DarkWeird, line 55: Use Constant for names.
Also i don't think that underscore prefix is functional

skaldarnar: What do you mean by _"functional"_?

I'll change it real quick, but would still like to understand this 😉

DarkWeird: I imagine json with this map...
Many underscores in every object.

Functional:
Js have underscore library.
Underscored names used as metadata or technical data (js, json. Kinda request time, or request id) as private fields(C/C# - legacy from highlightless past)

DarkWeird (approved):

subsystems/TypeHandlerLibrary/src/main/java/org/terasology/persistence/typeHandling/coreTypes/factories/MapTypeHandlerFactory.java
```diff
@@ -50,6 +48,19 @@
...
+        if (String.class.equals(keyType)) {
+            return Optional.of((TypeHandler<T>) new StringMapTypeHandler<>(valueTypeHandler));
+        } else {
+            Optional<TypeHandler<?>> declaredKeyTypeHandler =
+                    context.getTypeHandlerLibrary().getTypeHandler(keyType);
+            TypeInfo<?> keyTypeInfo = TypeInfo.of(keyType);
+            @SuppressWarnings({"unchecked"})
+            TypeHandler<?> keyTypeHandler = new RuntimeDelegatingTypeHandler(
+                    declaredKeyTypeHandler.orElse(null),
+                    keyTypeInfo,
```
keturn, line 60: I'm wondering about the SuppressWarnings here. I guess it was probably copied from valueTypeHandler above.

keturn:
> This generic map type handler encodes a generic map as list of objects representing key-value-pairs:

Can we encode each entry as a tuple instead of having a map for each?

subsystems/TypeHandlerLibrary/src/main/java/org/terasology/persistence/typeHandling/coreTypes/GenericMapTypeHandler.java
```diff
@@ -0,0 +1,63 @@
...
+                .collect(Collectors.toList());
+        return serializer.serialize(entries);
+    }
+
+    private PersistedData serializeEntry(Map.Entry<K, V> entry, PersistedDataSerializer serializer) {
+        PersistedData key = keyHandler.serialize(entry.getKey(), serializer);
+        PersistedData value = valueHandler.serialize(entry.getValue(), serializer);
+
+        Map<String, PersistedData> jsonEntry = Maps.newLinkedHashMap();
+        if (!key.isNull()) {
```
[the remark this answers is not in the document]
skaldarnar: Hm, good question 🤔  I think the difference here is between `null` as used directly in Java, and `isNull()` as in the given data for the key could not be persisted. Here, I assumed it `isNull` check is for the latter, meaning that we could not persist the data itself. This assumption might be wrong.

subsystems/TypeHandlerLibrary/src/main/java/org/terasology/persistence/typeHandling/coreTypes/GenericMapTypeHandler.java
```diff
@@ -0,0 +1,66 @@
...
+            return Optional.empty();
+        }
+
+        Map<K, V> result = Maps.newLinkedHashMap();
+
+        for (PersistedData entry : data.getAsArray()) {
+            final Optional<K> key = keyHandler.deserialize(entry.getAsValueMap().get(KEY));
+            final Optional<V> value = valueHandler.deserialize(entry.getAsValueMap().get(VALUE));
+
+            key.ifPresent(k -> result.put(k, value.orElse(null)));
```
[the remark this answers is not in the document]
skaldarnar: Also a good question (which I think holds for other type handlers as well...). The nested nature of the maps and other collections yields two-fold results: the whole thing cannot be serialized (`Optional.empty()`) vs only parts cannot be serialized. 
Maybe we should print a warning in the type handler if that happens. What kind of data would we need in that case?

subsystems/TypeHandlerLibrary/src/main/java/org/terasology/persistence/typeHandling/coreTypes/factories/MapTypeHandlerFactory.java
skaldarnar: I guess we cannot get type safety for the generics in here 🤷‍♂️

skaldarnar:
@keturn [PersistedData](https://github.com/MovingBlocks/Terasology/blob/develop/subsystems/TypeHandlerLibrary/src/main/java/org/terasology/persistence/typeHandling/PersistedData.java) is quite limited in what it offers. There are no _pairs_ or _tuples_ we could use. 
We could potentially use an array with two elements instead of an object with two fields, such that the JSON encoding would become:
```jsonc
[
  [ /* key encoding */, /* value encoding */ ],
  [ /* key encoding */, /* value encoding */ ],
  ...
]
```

subsystems/TypeHandlerLibrary/src/main/java/org/terasology/persistence/typeHandling/coreTypes/GenericMapTypeHandler.java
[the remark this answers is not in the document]
keturn: I guess for deserialization there are times when you would want to do a forgiving lossy-loading thing like this. Maybe the server admin removed a module and you want to take your chances with just dropping that part of the data structure while loading the rest of it.

(I haven't yet learned which use cases we're talking about here. Is this for configuration files _and_ save files _and_ the wire protocol?)

There are other times when I'd want a strict-mode, when there's no _good_ reason for anything to fail to deserialize and it might as declare the entire operation invalid.

For debugging, you want the raw key and value as well as some way to _find_ it if the key name isn't very descriptive and it's nested somewhere deep. I don't think `deserialize` is being given that context here so that kind of location information might be beyond the scope of this PR.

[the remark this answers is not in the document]
eviltak: @keturn that is actually something I was thinking about during the initial serialization overhaul, and I came to the same conclusions as you. The reason I didn't implement a strict-mode was because at the time I deemed it to be too much effort for its benefits. Definitely would be nice to have though -- maybe even just as a standalone "prefab validator".

@skaldarnar I would definitely log a warning/error here -- that is what all other type handlers also do when they cannot find something that they need. Something like `"Missing '" + KEY + "' field"` if the `KEY` field is missing (and the same if the `VALUE` field is absent). We do not *have* to log a warning if there are other redundant fields present (other type handlers do not), but you could if you wanted to.

eviltak (dismissed):

DarkWeird (approved):

PR #4384, opened 2021-01-09
feat(JOML): add JOML API for `world.propagation.*` (partial)

skaldarnar:
Contributes to #3832

This keeps the existing APIs available but deprecates them. There should be no module changes required. 

- chore(world): remove unused AbstractChunkView.java (Omega scope)
- feat(JOML): add JOML API for `world.propagation.*` (partial migration)
- test(JOML): migrate tests for `world.propagation.*

PR #4389, opened 2021-01-10
feat(Chunks): add in-place modifier for `Chunks#toChunkRegion`

skaldarnar:
### Contains

This is a follow up to #4384 to allow for in-place mutation of `BlockRegion` in `Chunks` math operations. 

This allows to create or recieve a `BlockRegion` (mutable) that should be modified in-place by the chunk calculation. Instead of 

```java
BlockRegion region = new BlockRegion(...);
for (Vector3ic pos : Chunks.toChunkRegion(region, region)) {
    // do stuff
}
```

we could inline `region` as it is used implicitly as both input and output:

```java
for (Vector3ic pos : Chunks.toChunkRegion(new BlockRegion(...))) {
    // do stuff
}
```

- [ ] add test cases
- [ ] add javadoc
- [ ] check `Chunks` for consistency

### DIFF
--- a/engine/src/main/java/org/terasology/world/chunks/Chunks.java
+++ b/engine/src/main/java/org/terasology/world/chunks/Chunks.java
@@ -85,7 +85,6 @@ public final class Chunks {
         return toChunkPos(z, CHUNK_POWER.z());
     }
 
-
     /**
      * The position of the chunk given the coordinate and size of chunk in powers of 2.
      *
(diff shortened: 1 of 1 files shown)


pollend:
I'm wondering if there are any problems with source to destination but min modifies min and max modifies max so I think this should be fine.

pollend (approved):

skaldarnar:
> I'm wondering if there are any problems with source to destination but min modifies min and max modifies max so I think this should be fine

This was actually what I was wondering about when asking about this on Discord. If I understood the answers by you and httpdigest correctly the methods taking a `dest` should guarantee that the modification of `dest` is the last thing that happens, and that is is safe to use the same input as source and destination.

I agree that this needs tests.

pollend:
do we still want this. the in place modifiers are already in Chunks.

Issue #3832, opened 2020-02-16
Follow-ups after JOML adoption


Cervator:
With #3811 we embark on the journey to adopt JOML instead of our aging TeraMath library. It won't be without bruises for such a major change, however! Creating this issue to capture some future todos to follow back up on, either when somebody is able to dig in or after the final engine PR finishes the journey (and likely rolls into module land). I'm just noting a few impressions here, @pollend would probably be better at detailing them over time 👍 

## Backlog

- [x] https://github.com/Terasology/Minesweeper/pull/20

## Drafts

- [x] Draft: feat(JOML): migrate and clean up code (https://github.com/Terasology/SegmentedPaths/pull/15)
- [ ] [19 PRs on JOML primitives](https://github.com/search?q=org%3ATerasology+is%3Apr+is%3Aopen+joml+primatives&type=Issues) (to be migrated to use `joml-geometry`) 

## Done

<details>
<summary>Details</summary>

- [x] https://github.com/MovingBlocks/Terasology/pull/4404
- [x] https://github.com/Terasology/CoreRendering/pull/26
- [x] https://github.com/Terasology/Inventory/pull/31
- [x] https://github.com/Terasology/WorkstationCrafting/pull/22
- [x] https://github.com/Terasology/WeatherManager/pull/20
- [x] https://github.com/Terasology/Caves/pull/14
- [x] https://github.com/Terasology/SurfaceFacets/pull/9
- [x] https://github.com/Terasology/AnotherWorld/pull/20
- [x] https://github.com/Terasology/AdditionalRails/pull/44
- [x] https://github.com/Terasology/BiomesAPI/pull/9
- [x] https://github.com/Terasology/ClimateConditions/pull/44
- [x] https://github.com/Terasology/Climbables/pull/5
- [x] https://github.com/Terasology/CustomOreGen/pull/10
- [x] https://github.com/Terasology/Equipment/pull/131
- [x] https://github.com/Terasology/Fluid/pull/29
- [x] feat(JOML): migrate block networks (https://github.com/Terasology/BlockNetwork/pull/15)
- [x] feat(JOML): migrate chunkview and tests cases (https://github.com/MovingBlocks/Terasology/pull/4335)
- [x] ~~WIP: feat(JOML): migrate aabb for joml mesh (https://github.com/MovingBlocks/Terasology/pull/4190)~~
[...]

PR #4391, opened 2021-01-10
feat(JOML): add JOML API for BatchPropagator (partial migration)

skaldarnar:
This keeps the existing APIs available but deprecates them. There should be no module changes required.

I recommend to have a look at the individual commits to review this migration. 
I'd like to highlight the changes in 3de1db4 (refactoring) and 4dec17d (performance optimization).

If you prefer, I can split this into more PRs.

---

Contributes to #3832
Follow-up to #4384

### DIFF
--- a/engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
+++ b/engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
@@ -23,12 +23,14 @@ import org.terasology.math.ChunkMath;
 import org.terasology.math.Side;
 import org.terasology.world.block.Block;
 import org.terasology.world.block.BlockRegion;
+import org.terasology.world.block.BlockRegionc;
 import org.terasology.world.chunks.Chunks;
 import org.terasology.world.chunks.LitChunk;
 
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 
 /**
  * Batch propagator that works on a set of changed blocks Works for a single given propagation ruleset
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
```diff
@@ -272,9 +261,9 @@ private void increase(Vector3i position, byte value) {
      * @param position The position to set at
      * @param oldValue The original value at the position
      */
-    private void reduce(Vector3i position, byte oldValue) {
+    private void reduce(Vector3ic position, byte oldValue) {
         if (oldValue > 0) {
-            reduceQueues[rules.getMaxValue() - oldValue].add(position);
+            reduceQueues[rules.getMaxValue() - oldValue].add(new Vector3i(position));
```
skaldarnar, line 266: Make a copy when storing the vector in local persistence.

engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
```diff
@@ -284,27 +273,27 @@ private void reduce(Vector3i position, byte oldValue) {
      * @param position The position to propagate form
      * @param value The value to propagate out
      */
-    private void queueSpreadValue(Vector3i position, byte value) {
+    private void queueSpreadValue(Vector3ic position, byte value) {
         if (value > 1) {
-            increaseQueues[rules.getMaxValue() - value].add(position);
+            increaseQueues[rules.getMaxValue() - value].add(new Vector3i(position));
```
skaldarnar, line 278: Make a copy when storing the vector in local persistence.

jdrueckert (approved):

pollend (approved):

engine/src/main/java/org/terasology/world/propagation/StandardBatchPropagator.java
pollend: that should avoid bugs if the value is modified outside :+1:

Issue #4393, opened 2021-01-10
world loading takes an extremely long time with high CPU when many source modules are present


keturn:
### What you were trying to do

Start or join a game.

### What actually happened

It hangs here **[TODO: insert screenshot]** for minutes (or much longer?) using high CPU (one core? or all of them?).

### How to reproduce

1. Clone engine repo
2. ./groovyw module init omega
3. ./gradlew clean jar game
4. Select and start CoreSampleGameplay and observe that the game does not start but is stuck in loading phase

[reproduction plan provided by @jdrueckert in #4392]

### Discussion and Causes

Some earlier notes on this are on #4372, but that was a prototype PR that has since been dropped.

> This is not a build configuration error.
>
> Some things shifted around and now single-player games (in certain dev environments) are running in to the performance problem we identified in multiplayer games in **#4154** back in September.
>
> For next steps, I suggest:
> * optimize this code querying subtypes in ModuleEnvironmentSandbox and/or gestalt's ModuleClassLoader. There are some [preliminary ideas about that](https://github.com/MovingBlocks/Terasology/issues/4154#issuecomment-705382655) written down already.
> * update to current gestalt (#4144), and _then_ optimize that stuff, because some of it probably changes with the new version anyway.
>
> In the meantime, the stopgap workaround is:
> * don't have a ton of modules subprojects checked out in your workspace. The performance problem climbs sharply with the number of module directories.

It's interesting that JDR encountered this from gradle, as previously Michael and I had seen it happen when running from IntelliJ but _not_ gradle. I hypothesized that might be because of this:

> gradle is putting the jar files on the classpath, but IntelliJ is giving the classes directory.
>
> It's not wrong to use the directories here. It might help module and asset hot-reloading speed in development and saves a build step. gradle might well do the same thing if the configuration changes a bit.

jdrueckert:
I would be highly interested in an analysis where this comes from. It must be a recent thing as I recall that only a few days ago I was able to start various gameplays fine from my source omega workspace.

> update to current gestalt (#4144), and then optimize that stuff

Optimistically estimated, it's summer before we're done with JOML _and_ gestalt v7 conversion... Which means we can neither work towards a release nor do any kind of playtesting or even in-game testing of PRs until then. IMO that's very bad as we're practically converting stuff for JOML and the gestalt v7 migration blind-folded.

> don't have a ton of modules subprojects checked out in your workspace

That's just not an option for conversions such as we're currently doing for JOML.

[...]

PR #4394, opened 2021-01-10
Fix: Make the chunk relevance region iterator actually terminate

4Denthusiast:
Only return each position from the iterator once, rather than repeatedly returning it, possibly forever, or possibly busy-looping until the chunk is complete. Fix for #4385.


jdrueckert (approved):

jdrueckert:
Fixes #4393

PR #4395, opened 2021-01-10
feat(JOML): migrate `BatchPropagator`

skaldarnar:
- feature(JOML): migrate `InternalLightProcessor`
- chore: remove unused vector
- feat(JOML): migrate `SunlightRegenBatchPropagator`
- refactor: remove duplicate code
- refactor: move `populateMinAdjacent2D` from `ChunkMath` to `BatchPropagator`

---

Contributes to #3832
Follow-up to #4391


engine/src/main/java/org/terasology/world/propagation/SunlightRegenBatchPropagator.java
```diff
@@ -191,90 +178,78 @@ private void reduce(Vector3i position, byte oldValue) {
...
-                            strength++;
-                        }
-                        pos.y--;
+                    initialDepth = Math.max(Chunks.SUNLIGHT_REGEN_THRESHOLD - start, adjDepths[depthIndex]);
+                }
+                byte strength = (byte) Math.min(Chunks.MAX_SUNLIGHT, start + initialDepth - Chunks.SUNLIGHT_REGEN_THRESHOLD + 1);
+                for (int i = initialDepth; i <= depth[depthIndex]; ++i) {
+                    sunlightPropagator.propagateFrom(toChunk.chunkToWorldPosition(x, Chunks.SIZE_Y - i - 1, z, pos), strength);
+                    if (strength < Chunks.MAX_SUNLIGHT) {
+                        strength++;
```
pollend, line 222: so you moved the iterator out since its replicated in both conditions

engine/src/main/java/org/terasology/world/propagation/BatchPropagator.java
```diff
@@ -55,21 +45,86 @@
...
+    }
+
+    /**
+     * Populates a target array with the minimum value adjacent to each location, including the location itself.
+     *
+     * @param source
+     * @param target
+     * @param populateMargins Whether to populate the edges of the target array
+     */
+    default void populateMinAdjacent2D(int[] source, int[] target, int dimX, int dimY, boolean populateMargins) {
```
pollend, line 88: I was thinking static but this might work better. allows for different implementations. :+1:

pollend (dismissed):

engine/src/main/java/org/terasology/world/propagation/SunlightRegenBatchPropagator.java
skaldarnar: yep, I started with doing small changes, bringing arguments in the same order, inlining some parts, and suddenly it was highlighted as duplicate :sweat_smile:

engine/src/main/java/org/terasology/world/propagation/BatchPropagator.java
skaldarnar: I don't know whether we actually want different implementations? Maybe we should keep it static and only open that up if we actually find a use case for that?

skaldarnar:
Rebased this PR onto `develop` after base PR was merged. No further changes.

PR #4399, opened 2021-01-11
perf(optimize): optimize iterator for block region 

pollend:
I removed some of the conditions to speed up the iterator for block regions.


skaldarnar (approved):
Trusting in our test cases for the iterator, I'll approve this 🙈

PR #4400, opened 2021-01-11
perf: tweak simplex noise with FMA

pollend:
FMA combines a muliplication and an addin operation.  

https://en.wikipedia.org/wiki/FMA_instruction_set


engine/src/main/java/org/terasology/utilities/procedural/SimplexNoise.java
```diff
@@ -307,42 +308,42 @@ public float noise(float xin, float yin, float zin) {
         float z3 = z0 - 1.0f + 3.0f * G3;
 
         // Work out the hashed gradient indices of the four simplex corners
-        int ii = Math.floorMod(i, permCount);
-        int jj = Math.floorMod(j, permCount);
-        int kk = Math.floorMod(k, permCount);
+        int ii = java.lang.Math.floorMod(i, permCount);
```
skaldarnar, line 311: How I would like the rename-imports for Scala here 🙃

skaldarnar (commented):
How much do we gain through this change? I think it's still okay to figure out what is happening, although it _is_ obfuscating the exact operation a bit. 

I'd like to hear @4Denthusiast's opinion on this before merging 🤓

4Denthusiast:
This looks like it makes sense, especially given that noise generation can be one of the more expensive steps in world generation, except the benchmarks Michael showed on Discord show only a 1.6% improvement. I don't think that that's worth the reduced readability.

pollend:
Fma is also a bit more accurate. Not sure if it optimizes those instructions on the backend. We should tests this with some other machine because some processors might have better support then others.

4Denthusiast:
> Fma is also a bit more accurate.

Avoiding floating-point rounding errors here also doesn't seem necessary. This algorithm isn't numerically unstable, therefore the errors will only be a few ULPs (around 10^-7). If we convert that to, for example, its effect on the ElevationFacet (using the usual world generator), that would be still less than 10^-3 blocks, nowhere near enouch to cause actual problems.

DarkWeird:
@pollend 

I have bad news for you:
```
Benchmark                  Mode  Cnt   Score    Error  Units
FMABenchmark.xyDotFma      avgt    5  33.474 ± 19.786  ns/op
FMABenchmark.xyDotPlain    avgt    5  33.735 ± 25.661  ns/op

FMABenchmark.xyzDotFma     avgt    5  36.770 ± 28.444  ns/op
FMABenchmark.xyzDotPlain   avgt    5  29.548 ±  7.644  ns/op

FMABenchmark.xyzwDotFma    avgt    5  34.359 ± 31.066  ns/op
FMABenchmark.xyzwDotPlain  avgt    5  31.685 ± 18.044  ns/op
```

Using BellSoft Liberica JDK 11 (we use it for game, bundled with launcher)

Code:
```
    @Benchmark
    public float xyDotPlain(StateObject obj) {
        return obj.gx * obj.x + obj.gy * obj.y;
    }

    @Benchmark
    public float xyzDotPlain(StateObject obj) {
        return obj.gx * obj.x + obj.gy * obj.y + obj.gz * obj.z;
    }

    @Benchmark
    public float xyzwDotPlain(StateObject obj) {
        return obj.gx * obj.x + obj.gy * obj.y + obj.gz * obj.z + obj.gw * obj.w;
    }

    @Benchmark
    public float xyDotFma(StateObject obj) {
        return Math.fma(obj.gx, obj.x, obj.gy * obj.y);
    }

    @Benchmark
    public float xyzDotFma(StateObject obj) {
        return Math.fma(obj.gx, obj.x, Math.fma(obj.gy, obj.y, obj.gz * obj.z));
    }

    @Benchmark
    public float xyzwDotFma(StateObject obj) {
        return Math.fma(obj.gx, obj.x, Math.fma(obj.gy, obj.y, Math.fma(obj.gz, obj.z, obj.gw * obj.w)));
    }
```

PR #4405, opened 2021-01-16
feat(JOML): migrate BlockLifecycleEvent

pollend:


engine/src/main/java/org/terasology/world/block/BlockLifecycleEvent.java
```diff
@@ -16,14 +16,18 @@
 package org.terasology.world.block;
 
 import gnu.trove.list.TIntList;
+import org.jetbrains.annotations.NotNull;
```
skaldarnar, line 19: remove this, please

engine/src/main/java/org/terasology/world/block/BlockLifecycleEvent.java
```diff
@@ -32,11 +36,46 @@ public BlockLifecycleEvent(TIntList positions, BlockEntityRegistry registry) {
         this.positions = positions;
     }
 
-    public Iterable<Vector3i> getBlockPositions() {
-        return () -> new BlockPositionIterator(positions, registry);
+    @NotNull
```
skaldarnar, line 39: remove this

engine/src/main/java/org/terasology/world/block/BlockLifecycleEvent.java
```diff
@@ -32,11 +36,46 @@ public BlockLifecycleEvent(TIntList positions, BlockEntityRegistry registry) {
...
+            private final Vector3i current = new Vector3i();
+            private int index = 3;
+            @Override
+            public boolean hasNext() {
+                return next != null;
+            }
+
+            @Override
+            public Vector3ic next() {
+                current.set(next);
```
skaldarnar, line 59: This will throw a NPE when there's no `next` element, right? Should we throw a `NoSuchElementException` instead, as suggested in the docs for `Iterator`?

engine/src/main/java/org/terasology/world/block/BlockLifecycleEvent.java
```diff
@@ -32,11 +36,46 @@ public BlockLifecycleEvent(TIntList positions, BlockEntityRegistry registry) {
         this.positions = positions;
     }
 
-    public Iterable<Vector3i> getBlockPositions() {
-        return () -> new BlockPositionIterator(positions, registry);
+    @NotNull
+    @Override
+    public Iterator<Vector3ic> iterator() {
```
skaldarnar, line 41: What exactly does this iterator do?

The `TIntList positions` is flattened list of positions, right? So if this event affects
```
[(0,0,0), (1,2,3), (4,5,6)]
```
then `positions` would be:
```
[0,0,0, 1,2,3, 4,5,6]
```

But whether a block listed in `positions` is actually returned depends on whether there's a _permanent block entity_ registered or not? 
Doesn't this lead to a mismatch between the size as returned by `blockCount` and the actual amount of positions returned by the iterator?

engine/src/main/java/org/terasology/world/block/BlockLifecycleEvent.java
```diff
@@ -32,11 +36,46 @@ public BlockLifecycleEvent(TIntList positions, BlockEntityRegistry registry) {
...
+
+            @Override
+            public Vector3ic next() {
+                current.set(next);
+                fetchNext();
+                return current;
+            }
+
+            private void fetchNext() {
+                while (index < positions.size() - 2) {
```
skaldarnar, line 65: If the condition `!registry.hasPermanentBlockEntity(next)` does not hold for any position, then we will just iterate to the end and set `next = null`. If it does hold for some of the positions, then we only iterate over a subset of the positions.

So, it could be that `event.blockCount() == 6` but `for (Vector3ic pos: event)` only iterates over `4` blocks? :thinking:

skaldarnar (changes requested):
I don't get what this event is about, and how it should behave. I see that you did some changes in order to correct the behavior, but I'm still not sure whether it really is correct behavior after all...

engine/src/main/java/org/terasology/world/block/BlockLifecycleEvent.java
skaldarnar: Maybe we should not inline the iterator here, but rather make it explicit that it will only iterator over `blockPositionsWithNonPermanentEntities`.

engine/src/main/java/org/terasology/world/block/BlockLifecycleEvent.java
```diff
@@ -16,14 +16,18 @@
...
-import org.terasology.world.block.internal.BlockPositionIterator;
+
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
 
 /**
  */
-public abstract class BlockLifecycleEvent implements Event {
+public abstract class BlockLifecycleEvent implements Event, Iterable<Vector3ic> {
```
skaldarnar, line 30:
```suggestion
//FIXME: There is a mismatch between `blockCount` and iterated blocks.
//       I don't get what this event is about, and how it should behave. The `TIntList positions` is a flattened list
//       of positions. So if this event affects
//          [(0,0,0), (1,2,3), (4,5,6)]
//       then `positions` would be:
//          [0,0,0, 1,2,3, 4,5,6]
//       But whether a block listed in positions is actually returned depends on whether there's a permanent block
//       entity registered or not? Doesn't this lead to a mismatch between the size as returned by blockCount and 
//       the actual amount of positions returned by the iterator? So, it could be that 
//          event.blockCount() == n 
//       but 
//          for (Vector3ic pos: event) { ... }
//       only iterates over m < n blocks? Maybe we should not inline the iterator here, but rather make it explicit
//       that it will only iterator over `blockPositionsWithNonPermanentEntities`.
public abstract class BlockLifecycleEvent implements Event, Iterable<Vector3ic> {
```

skaldarnar (approved):

PR #4401, opened 2021-01-11
feature(JMH): Migrate benchmarks to JMH

DarkWeird:
### Contains

Benchmark migration to JMH.
pros:
1. Easy to write
2. microbenchmarking standart.
3. integrate with gradle(custom)
cons:
1. not own solution :)

### How to test

1. just run `gradlew jmh`

<details>
<summary>Details</summary>

```
Benchmark                                                                 Mode Cnt    Score   Error Units
entitySystem.EntityCreateBenchmark.createEntity                           avgt   5 2190.730±568.068 ns/op
entitySystem.IterateComponentsBenchmark.iterateMultipleComponent          avgt   5  455.723± 97.854 ns/op
entitySystem.IterateComponentsBenchmark.iterateSingleComponent            avgt   5  481.269±135.628 ns/op
reflectFactory.ReflectionFactoryBenchmark.byteCodeConstructor             avgt   5   30.081±3.518 ns/op
reflectFactory.ReflectionFactoryBenchmark.byteCodeFieldAccessGet          avgt   5    6.028±0.054 ns/op
reflectFactory.ReflectionFactoryBenchmark.byteCodeFieldAccessSet          avgt   5    5.874±1.046 ns/op
reflectFactory.ReflectionFactoryBenchmark.byteCodeGetterSetterAccessGet   avgt   5    6.552±0.384 ns/op
reflectFactory.ReflectionFactoryBenchmark.byteCodeGetterSetterAccessSet   avgt   5    3.122±0.062 ns/op
reflectFactory.ReflectionFactoryBenchmark.directConstructor               avgt   5   29.112±0.993 ns/op
reflectFactory.ReflectionFactoryBenchmark.directFieldGet                  avgt   5    4.978±0.014 ns/op
reflectFactory.ReflectionFactoryBenchmark.directFieldSet                  avgt   5    4.013±0.364 ns/op
reflectFactory.ReflectionFactoryBenchmark.directGetterSetterGet           avgt   5    5.457±0.516 ns/op
reflectFactory.ReflectionFactoryBenchmark.directGetterSetterSet           avgt   5    0.969±0.070 ns/op
reflectFactory.ReflectionFactoryBenchmark.reflectionConstructor           avgt   5   35.013±3.501 ns/op
reflectFactory.ReflectionFactoryBenchmark.reflectionFieldAccessGet        avgt   5    9.162±0.500 ns/op
[...]

PR #4406, opened 2021-01-16
build(jmh): git-ignore generated code

skaldarnar:
Fixup of #4401

I had hoped this would also help with SpotBugs going crazy, but nope...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@
 !.gitignore
 !.github
 
+# Ignore generated code
+/engine/src/jmh/generated
+
 # Ignore subdirs with their own Git roots. PC Facade and Core module are bundled with the main repo.
 extensions
 /facades/*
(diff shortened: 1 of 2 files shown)


pollend (approved):

Cervator (commented):
Oddly, I can't seem to actually find any files at `/engine/src/jmh/generated` - they seem to go to a couple directories under `build/` which is already ignored. The magic trick to avoid the warnings seems to have been https://github.com/MovingBlocks/TeraConfig/pull/15

I usually test by version bumping the config locally, generate the zip, shove it into Artifactory, then bump the engine dependency locally, but I guess that changed with a GitHub action setup. I don't see a place to put the version, and am guessing that's automatic now if you trigger a release by pushing a tag to the repo?

Cervator:
I added a commit to bump the TeraConfig dependency to this PR's branch and checked the outcome in Jenkins. Better!

image

Will try to do the squash merge and hopefully this'll turn out alright.

PR #4408, opened 2021-01-16
feat(JOML): migrate WorldProvider

skaldarnar:
- chore: remove unused imports
- feat(JOML): migrate WorldProvider

Contributes to #3832

## Module PRs

- [ ] https://github.com/Terasology/FlexiblePathfinding/pull/16


engine/src/main/java/org/terasology/world/internal/WorldProviderCoreImpl.java
```diff
@@ -153,25 +155,15 @@ public ChunkViewCore getLocalView(Vector3ic chunkPos) {
...
+        return chunkProvider.getSubviewAroundChunk(JomlUtil.from(chunk));
     }
 
     @Override
     public boolean isBlockRelevant(int x, int y, int z) {
         return chunkProvider.isChunkReady(ChunkMath.calcChunkPos(x, y, z));
     }
 
-    @Override
-    public boolean isRegionRelevant(Region3i region) {
```
skaldarnar, line 166: Has been replaced with https://github.com/MovingBlocks/Terasology/pull/4408/files#diff-34e089d05ce776ea17ccc885b61184ad54b78b4cf427bb7db40dd01593668616R104-R106

engine-tests/src/main/java/org/terasology/MapWorldProvider.java
```diff
@@ -83,27 +82,16 @@ public boolean isBlockRelevant(int x, int y, int z) {
         return false;
     }
 
-    @Override
-    public boolean isRegionRelevant(Region3i region) {
```
skaldarnar, line 87: Has been replaced by https://github.com/MovingBlocks/Terasology/pull/4408/files#diff-34e089d05ce776ea17ccc885b61184ad54b78b4cf427bb7db40dd01593668616R104-R106

engine/src/main/java/org/terasology/world/internal/WorldProviderWrapper.java
```diff
@@ -42,89 +40,54 @@ public WorldProviderWrapper(WorldProviderCore core, ExtraBlockDataManager extraD
...
-
     @Override
     public boolean isBlockRelevant(Vector3fc pos) {
-        return isBlockRelevant(new org.joml.Vector3i(pos, org.joml.RoundingMode.HALF_UP));
-    }
-
-    @Override
-    public Block setBlock(Vector3i pos, Block type) {
-        return core.setBlock(pos, type);
+        return isBlockRelevant(new Vector3i(pos, RoundingMode.HALF_UP));
```
pollend, line 50: why is it rounding half up?

pollend (changes requested):

engine/src/main/java/org/terasology/world/internal/WorldProviderWrapper.java
skaldarnar: I just took what it was before :thinking: 

Do we have a clear understanding which side is inclusive and which is exclusive?

pollend: actually its correct I think.

engine/src/main/java/org/terasology/world/internal/WorldProviderCoreImpl.java
```diff
@@ -67,7 +67,7 @@
 
     private final List<WorldChangeListener> listeners = Lists.newArrayList();
 
-    private final Map<Vector3i, BlockChange> blockChanges = Maps.newHashMap();
+    private final Map<org.terasology.math.geom.Vector3i, BlockChange> blockChanges = Maps.newHashMap();
```
pollend, line 70: we can't migrate this?

skaldarnar: hm... Let me double check how much this actually affects, but because of type erasure it is not possible to do this trick with the second method in the interface and forwarding with default implementation.
Migrating this method would also mean to migrate all callers at the same time...

skaldarnar: Yep, there's a long rat tail following that change. I'd like to do that in a follow-up PR.

It will also affect other places in the code, such as `world.block.structure.*`....

pollend: that sounds fine. we can do it in another pass.

pollend (approved):

PR #4409, opened 2021-01-17
Migrate Omega package from old Jenkins to new, with some minor cleanup / label improvements

Cervator:
### Contains

Newer replacement for #4231. Updates the engine repo just a little to prepare for a move to the new Jenkins which uses pipeline jobs, folders, and other fun things. Also includes the versioning updates noted in the older PR - I fully acknowledge that's a temporary state that'll likely change again, but I have scoped this down to where it is primarily just the migration plus a bit of assorted cleanup / clarifications. Hope that's OK until we come up with bigger and better methods for versioning and packaging.

### How to test

See #4231 - primarily tested in Jenkins, but won't show full "live" results until after it is merged. See for instance http://jenkins.terasology.io/teraorg/job/Nanoware/job/Omega/job/develop/ to see a new school Omega zip having been packaged. It runs and shows the new version label stuff on the main menu, but has plenty of module quirks (JS and something else was red, not really important for this PR)

Additionally http://jenkins.terasology.io/teraorg/job/Terasology/job/Omega/job/PR-10/ should actually also work - but will miss the version label updates till the engine `develop` branch is updated. PR Omega zips! For the Index repo, anyway - we could probably even map those to engine PRs somehow. Right now anything not `master` or `develop` for either Terasology or Nanoware will default to using the regular engine job's `develop` branch for the build harness

- [x] Jobs in Jenkins must be ready - http://jenkins.terasology.io/teraorg/job/Terasology/job/Omega/
- [x] With the Omega dist job being a multi-branch pipeline now any change to any branch _there_ will also trigger a build (packaging) of an Omega zip. That in theory would result in an identical set of version info for the game client, yet different modules (particularly if the commit to the Index repo was a change to the module lineup). That may just be a thing to ignore for now until we have a better packaging approach yet. The most obvious alternative would have the Omega packager update the engine's `versionInfo.properties` which may not be desired. Same thing also happens on a manually triggered rebuild.
  - One option here could just be to exclude `master` in the Index repo. Extra dev builds are fine, really just want to avoid extra release builds. A release build for the engine will still trigger a release build for Omega automatically. Created issue to handle: https://github.com/MovingBlocks/InfraPlayground/issues/17
- [x] Merge https://github.com/Terasology/Index/pull/11 - but unsure how now. It targets the `develop` branch, which has an open PR to the `master` branch, so it might be bad to squash-merge the one into the other 🤔 

### Outstanding _after_ merging

- [ ] Update release management page in wiki to indicate the tiny step of bumping the alpha increment in `templates/version.txt` till we entirely replace that setup with something else.
- [ ] Reset the Nanoware `develop` and `master` branches - I used both of them and did local testing for the branch sensitive stuff. Topic branch is still there as well but needed extra bits in the `Jenkinsfile` to use
- [ ] Also reset the master branch for Nanoware/Index and again only update it when we do the next game Omega release
- [ ] After merging this **only** push module lineup changes to the `develop` branch for Terasology/Index, until the next game release. At that point update the `master` branch which should trigger the first ever release Omega from the new Jenkins (don't want to merge into and update `master` until then). See https://github.com/Terasology/Index/pull/10
- [ ] After merging this and being sure we have good zips in the new Jenkins we need to update the launcher to pick up stuff from there instead of the old Jenkins. @skaldarnar noted https://github.com/MovingBlocks/TerasologyLauncher/pull/609 should have some stuff for that and it shouldn't be much work to hook back up
- [ ] After *that* is done and the launcher no longer needs the _old_ Jenkins we are about ready to decommission it - there are a last few tiny jobs I could probably prep real quick, but this effort is the biggest reason that poor old butler still lives
- [ ] Consider the priority and desire for an Omega PR paired with an engine PR rather than using our standard `develop` build for the engine
- [ ] Flip the config for http://jenkins.terasology.io/teraorg/job/Terasology/job/Omega/ back to ignore branches used as origin for a PR. Since `develop` is being PRed to `master` the actual _branch_ is being ignored, which we don't want. One-time quirk.
- [ ] Unrelated but discovered through testing here: Consider the impact of the Jenkins master becoming a bottleneck for reporting results to during a build storm

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,6 +34,17 @@ node ("heavy-java") {
         } else {
             println "Running on a branch other than 'master' or 'develop' bypassing publishing"
         }
+
+        // Trigger the Omega dist job to repackage a game zip with modules
+        if (env.JOB_NAME.equals("Terasology/engine/develop")) {
+            build job: 'Terasology/Omega/develop', wait: false
+        } else if (env.JOB_NAME.equals("Terasology/engine/master")) {
+            build job: 'Terasology/Omega/master', wait: false
+        } else if (env.JOB_NAME.equals("Nanoware/Terasology/develop")) {
+            build job: 'Nanoware/Omega/develop', wait: false
+        } else if (env.JOB_NAME.equals("Nanoware/Terasology/master")) {
+            build job: 'Nanoware/Omega/master', wait: false
+        }
     }
     stage('Analytics') {
         sh "./gradlew --console=plain check spotbugsmain javadoc"
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -31,18 +31,8 @@ ext {
     startDateTimeString = dateTimeFormat.format(new Date())
     versionInfoFileDir = new File(buildDir, 'classes/org/terasology/version')
     versionInfoFile = new File(versionInfoFileDir, 'versionInfo.properties')
-    versionFileName = 'VERSION'
     versionBase = new File(templatesDir, "version.txt").text.trim()
     displayVersion = versionBase
-}
-
-def convertGitBranch = { gitBranch ->
-    if (gitBranch != null) {
-        // Remove "origin/" from "origin/develop"
-        gitBranch.substring(gitBranch.lastIndexOf("/") + 1)
-    } else {
-        ""
-    }
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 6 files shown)


keturn (approved):
Looks fine from here! Jenkinsfile changes look sensible. :+1: on the gradle changes. Java stuff looks good and low-risk in any case.

Cervator:
Sorted out most the follow-ups, including the most awkward one: not wanting to trigger Omega release builds by simply updating the lineup or anything else in the Index repo's master branch. Found a special property that with a bit of extra Job DSL magic could be automagically applied to the Omega job which will solely exclude Git Triggers for the `master` branch job will still _making_ the job so it can be triggered by others (either us running it manually or the engine release build triggering it)

PR #4410, opened 2021-01-17
doc: add conventional commits to contributing guidelines

skaldarnar:
Following up the discussion on how our git workflow should look like, and how we can easily create a changelog from the merged commits between releases, I'd like to bring this up for discussion.

This draft is based on https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#commit, just do have a starting point...


.github/CONTRIBUTING.md
```diff
@@ -17,40 +16,103 @@ We prefer questions and support requests be posted in the [forum](http://forum.t
...
+```
+
+### Revert
+If the commit reverts a previous commit, it should begin with `revert: `, followed by the header of the reverted commit. In the body it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.
+
+### Type
+Must be one of the following:
+
+* **build**: Changes that affect the build system or external dependencies (example scopes: gradle, git, idea)
+* **ci**: Changes to our CI configuration files and scripts (example scopes: Jenkins)
```
Cervator, line 62: Would suggest "logistics" instead. That way it can count any sort of meta-work that's about the project and its tooling. CI is somewhat narrow in meaning, IMHO, specifically being for instance the `Jenkinsfile` - changes to Gradle aren't necessarily specific to CI, for instance. Utility scripting, template files, `.gitignore` are other examples.

.github/CONTRIBUTING.md
```diff
@@ -17,40 +16,103 @@ We prefer questions and support requests be posted in the [forum](http://forum.t
...
-1. **Optional:** Following the single line description (ideally no more than 70 characters long) should be a blank line followed by a list with the details of the commit.
-1. **Optional:** Add `Fix/Fixing/Fixes/Close/Closing/Refs #issue` if you want to mark the issue mentioned as complete in the issue tracker when the change is merged.
+Samples:
+```
+feat(world): add JOML API to WorldProvider
+```
+
+There are currently a few exceptions to the "use package name" rule:
+
+> :construction: What are our exceptions? Non-code changes (build, ci, ...)?
```
Cervator, line 82: I wonder if package would be too rigid, especially if a change affects multiple. Could it be more loosely defined concepts, like "JOML", then we can provide a list of initial common ones to pick from or come up with your own if it is different and unique, yet makes up a category likely to see more activity?

.github/CONTRIBUTING.md
```diff
@@ -17,40 +16,103 @@ We prefer questions and support requests be posted in the [forum](http://forum.t
...
+
+### Subject
+The subject contains a succinct description of the change:
+
+* use the imperative, present tense: "change" not "changed" nor "changes"
```
Cervator, line 87: This is good advice for nice consistency, although I worry a bit about contributors with a rough grasp of English. That's probably a pretty minor concern tho

.github/CONTRIBUTING.md
```diff
@@ -17,40 +16,103 @@ We prefer questions and support requests be posted in the [forum](http://forum.t
...
+* don't capitalize the first letter
+* no dot (.) at the end
+
+### Body
+Just as in the **subject**, use the imperative, present tense: "change" not "changed" nor "changes".
+The body should include the motivation for the change and contrast this with previous behavior.
+
+### Footer
```
Cervator, line 95: Even though my grasp of English _is_ pretty close to native, I'm mildly confused here between subject body and footer 😅 

I figured Subject is essentially the first sentence / line, some tools like Git Kraken has a separate line for that. Then Body is the rest of the message. Is Footer just meant to be a general reference to the _bottom_ line of a commit message, or is it something more formal I'm just not familiar with?

Cervator (approved):
Looks pretty good to me 👍 

Fairly minor comments added, none critical. It feels _a bit_ like this is a lot of thought for commit messages, compared to what the level of attention I expect a really casual contributor would pay to the topic. Not to say we shouldn't have high standards - I just wonder how high adherence to said standards would be, outside of raising additional barriers before approval.

Traditionally I've asked a time or two if some less-than-critical part of a PR is so-so (description of the overall PR, commit message, commit history, comments/javadoc etc), but then I tend to let it go in favor of moving on. That's just a personal thing / observational though.

.github/CONTRIBUTING.md
```diff
@@ -17,40 +16,103 @@ We prefer questions and support requests be posted in the [forum](http://forum.t
...
+* **docs**: Documentation only changes
+* **feat**: A new feature
+* **fix**: A bug fix for the game
+* **perf**: A code change that improves performance of the game
+* **refactor**: A code change that neither fixes a bug nor adds a feature
+* **style**: Changes that do not affect the meaning of the code (white-space, formatting, copyright, etc)
+* **test**: Adding missing tests or correcting existing tests
+
+### Scope
+The scope should be the name of the package(s) affected (as perceived by the person reading the changelog generated from commit messages).
```
keturn, line 72: This is one I've gone back and forth on over time. There certainly use cases it serves well, and then there are other things like `git annotate` where it would just be showing you the name of the package you already know you're looking at.

and that information _can_ be automatically derived from the filenames present in the commit.

.github/CONTRIBUTING.md
```diff
@@ -17,40 +16,103 @@ We prefer questions and support requests be posted in the [forum](http://forum.t
...
+<BLANK LINE>
+<body>
+<BLANK LINE>
+<footer>
+```
+
+The **header** is mandatory and the **scope** of the header is optional.
+
+Any line of the commit message cannot be longer 100 characters! This allows the message to be easier
+to read on GitHub as well as in various git tools.
```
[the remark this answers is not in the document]
skaldarnar: Unfortunately, it is :disappointed: although it is not the 50 or 80 character limit that's usually restricting, it's nevertheless a good practice to not let the title get too long..

image

I see this more as a guidance than a hard limit...

.github/CONTRIBUTING.md
skaldarnar: yeah, I'm not sure here either... I'd like to see this as an optional part we can fill in if it makes sense. For instance, for migrating JOML, it might be something like
```
refactor(world.block): migrate FooEvent and BarComponent
```
which narrows down the affected package without the need to look up the content of the respective commit/PR. It also can help to narrow down the semantic change in a PR from the follow-up changes necessary to adjust...

.github/CONTRIBUTING.md
skaldarnar: I think it's more or less the bottom line (except, it can also be multiple lines). In the Angular project they use a double empty line to separate it, I think. The whole point here is that all references to issues or PRs that are closed/fix should be in one place at the bottom.

**"Good"**
```
refactor!: adjust the API of FancyEvent

This migrates the API to work with S instead of T, and renames it accordingly.

BREAKING: FancyEvent.foo was renamed to FancyEvent.bar
Closes #123
```

**Bad**
```
refactor!: adjust the API of FancyEvent

This migrates the API to work with S instead of T (BREAKING). We close #123 this way.

Renames the API methods accordingly.
```

As we don't do full automation on this I don't think it is important that we define the technical details, but having the important information that can't be expressed in the subject in a single, easy to find place seems like a good thing to me.

.github/CONTRIBUTING.md
skaldarnar: Yep. At this point I would like to emphasize that I would like to ensure these conventions on the PR titles. And there, we could easily make sure to rephrase the title before merging. 

The reason why I think this will help us is that it is a step in a direction where we rely less on the inidividual commits for the changelog while still keeping it neat and organized in Github. Be it release drafter or some custom tool to show the merged PR between two tags, I personally would look less at commits.

If we have trouble finding a good title for a PR because it does so many things at once then I'm heavily in favor of not merging that PR, but instead splitting it up. 

(also pinging @keturn FYI)

.github/CONTRIBUTING.md
skaldarnar: Hm, I'm a bit hesitant to use "arbitrary" names here. We already have some confusion whether something is 
```
build: fix the gradle script
```
or 
```
fix(gradle): correct script
```
or
```
chore(gradle): fix the script
```

To my understanding this _scope_ should describe the part of the code base that has changed/is affected, the topic (such as JOML) would be mentioned in the subject line. I'd also organize these topics rather via labels or boards in Github.

Maybe we can limit the scope to the _top-level package_ (or max one level more), so you'd have `feat(world)` or `feat(world/chunks)`, but not something lengthy like `feat(world/chunks/localChunkProvider)` taking up the whole line...

.github/CONTRIBUTING.md
skaldarnar: Gradle changes would fall into the `build` category above... would you rather merge those two into a single entry `logistics`?

```
logistics(gradle): ... 
logistics(git): ...
logistics(idea): ...
logistics(jenkins): ...
```

.github/CONTRIBUTING.md
```diff
@@ -17,40 +16,103 @@ We prefer questions and support requests be posted in the [forum](http://forum.t
...
+
+### Revert
+If the commit reverts a previous commit, it should begin with `revert: `, followed by the header of the reverted commit. In the body it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.
+
+### Type
+Must be one of the following:
+
+* **build**: Changes that affect the build system or external dependencies (example scopes: gradle, git, idea)
+* **ci**: Changes to our CI configuration files and scripts (example scopes: Jenkins)
+* **docs**: Documentation only changes
```
keturn, line 63: I think all of these are nouns and this is the only one that's plural. Should it be `doc` for consistency?

keturn (commented):
I think I've caught up here. Looks like there have been some answers to clarifying questions but no big shifts.

I'm okay with this. I think having _some_ standards for commit messages is sensible, especially the subject (first line).

In my opinion, putting expectations of structure on human-authored content stored as a non-editable free-form text field is asking for trouble. But this isn't _too_ bad.

If we do want to stick to a strict taxonomy for the `type` field then I will probably need a pre-commit hook or IntelliJ plugin, cuz I will forever be forgetting which ones we're using and how they're spelled.

.github/CONTRIBUTING.md
keturn: Reviewing this list to pick the right type and scope, I'm noticing that I can be doing "build" (gradle) things that _also_ could be categorized as doc/fix/perf/refactor/style/test.

Which makes gradle feel more like a scope than a type?

On the other hand, it seems like a good idea to be able to separate build stuff from things that would be relevant to people who only use the distribution and don't do source builds. (doesn't really apply to engine as it is today, but in principle.)

skaldarnar: I agree, and I think it is up to us to define what we want to achieve by these guidelines (for PR titles). The _types_ and _scopes_ are a bit mixed up (e.g., you could argue about `doc: fix typo` vs `fix: correct typo in doc`, although this might exaggerating).

This is my personal view on the these _types_ in a bit more detail:
- **build** and **ci** affect the build process either for local development or our automation in the "cloud". This may have an impact on the final game (as seen by players), but is mostyl under the hood. Obviosuly, you could do `perf` or `fix` changes for these _types_ ...
- **feat**, **fix**, **perf** are noticable by players and thus have a direct impact on our final product. They may be more or less noticable in practice.
- **test** indicates that we worked on tests - which is nothing visible to players. This can be tricky, as updating/fixing a test often also requires to fix behavior in the code. In this case, I'd go for a `test: ...` commit message, but use **fix: ...** as PR title.
- **doc** if we only udpate Javadoc, or update documents living in the repository (e.g., in NUI)
- **refactor** and **style** should not change the behavior of the code and should not have an impact on anything... We currently use `chore` in a similar fashion, maybe that would fit better? What do you think?

@keturn @Cervator @jdrueckert what do you think?

Cervator: I might be too distant from this to really have a strong opinion I'd want people to listen to - so please don't take this too seriously. I thought the base conventional commits bit was interesting, going particularly deep or even using the (something) approach to indicate finer details is beyond the point that really concerned me. Not even truly sure the first thing (feat fix etc) is _super_ meaningful in my head, it was just novel to have _some_ sort of convention, I guess? Probably why I'm reluctant to wrap a whole lot of rigidity around it all, with a more laissez faire style fitting my level of interest.

That's just me though - I'm not saying that can't be useful, or even hugely impactful particularly if you fully control the environment and can count on everybody using the convention perfectly to the point where you can really wrap some cool automation around it. In our case I just worry the overhead from the extra process won't be worth the benefit. So in the end it is nice if there's a little more consistency to commits, but I spend so very little time looking at them anyway ... 🤷

PR #4415, opened 2021-01-19
feat(JOML): migrate Mesh AABBf

pollend:
- [ ] https://github.com/Terasology/CombatSystem/pull/72
- [ ] https://github.com/Terasology/GooKeeper/pull/39
- [ ] https://github.com/Terasology/Rails/pull/69

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/AABBRenderer.java
+++ b/engine/src/main/java/org/terasology/rendering/AABBRenderer.java
@@ -16,11 +16,12 @@
 
 package org.terasology.rendering;
 
+import org.joml.Vector3f;
+import org.joml.Vector4f;
 import org.lwjgl.opengl.GL11;
+import org.terasology.joml.geom.AABBf;
+import org.terasology.joml.geom.AABBfc;
 import org.terasology.logic.players.LocalPlayer;
-import org.terasology.math.AABB;
-import org.terasology.math.geom.Vector3f;
-import org.terasology.math.geom.Vector4f;
 import org.terasology.module.sandbox.API;
 import org.terasology.registry.CoreRegistry;
 
@@ -52,16 +53,16 @@ public class AABBRenderer implements BlockOverlayRenderer {
     private int displayListSolid = -1;
     private Vector4f solidColor = new Vector4f(1f, 1f, 1f, 1f);
 
-    private AABB aabb;
+    private AABBf aabb = new AABBf();
 
-    public AABBRenderer(AABB aabb) {
-        this.aabb = aabb;
+    public AABBRenderer(AABBfc aabb) {
+        this.aabb.set(aabb);
     }
 
     @Override
-    public void setAABB(AABB from) {
+    public void setAABB(AABBfc from) {
         if (from != null && !from.equals(this.aabb)) {
-            this.aabb = from;
+            this.aabb.set(from);
             dispose();
         }
     }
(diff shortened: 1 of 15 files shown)


engine/src/main/java/org/terasology/rendering/AABBRenderer.java
```diff
@@ -90,8 +91,9 @@ public void render() {
         CoreRegistry.get(ShaderManager.class).enableDefault();
 
         glPushMatrix();
-        Vector3f cameraPosition = CoreRegistry.get(LocalPlayer.class).getViewPosition();
-        glTranslated(aabb.getCenter().x - cameraPosition.x, -cameraPosition.y, aabb.getCenter().z - cameraPosition.z);
+        Vector3f cameraPosition = CoreRegistry.get(LocalPlayer.class).getViewPosition(new Vector3f());
+        Vector3f center = aabb.center(new Vector3f());
+        glTranslated(center.x - cameraPosition.x, -cameraPosition.y, center.z - cameraPosition.z);
```
skaldarnar, line 96: Looking at this, I think the API for AABBs is inconsistent here. To get the min or max vector, you'll have to do separate lookups without the option to pass in a destination vector.

```
Vector3i min = new Vector3i(aabb.minX(), aabb.minY(), aabb.minZ())
```

However, for the `center` it is exactly the other way around - we can only access the center as vector, but not the individual components...

skaldarnar: Also, we are creating a new vector here on every render call - should we instead reuse a tmp vector instead?

engine/src/main/java/org/terasology/rendering/AABBRenderer.java
```diff
@@ -102,8 +104,9 @@ public void renderSolid() {
...
```
skaldarnar, line 109: I'd like to avoid this code duplication, but that's probably something for another day...

engine/src/main/java/org/terasology/world/chunks/internal/ChunkImpl.java
```diff
@@ -341,14 +343,13 @@ public int chunkToWorldPositionZ(int z) {
...
     @Override
-    public AABB getAABB() {
-        if (aabb == null) {
-            Vector3f min = getChunkWorldOffset().toVector3f();
-            Vector3f max = ChunkConstants.CHUNK_SIZE.toVector3f();
+    public AABBfc getAABB() {
+        if (!aabb.isValid()) {
+            org.joml.Vector3f min = JomlUtil.from(getChunkWorldOffset().toVector3f());
+            org.joml.Vector3f max = new org.joml.Vector3f(Chunks.CHUNK_SIZE);
             max.add(min);
```
skaldarnar, line 350:
```suggestion
            org.joml.Vector3f max = min.add(new org.joml.Vector3f(Chunks.CHUNK_SIZE));
```

skaldarnar (dismissed):

skaldarnar (changes requested):
I'm pretty sure something is off here...
Terasology-210119195902-1080x805
Terasology-210119195903-1080x805

engine/src/main/java/org/terasology/rendering/AABBRenderer.java
pollend: sounds like we need to modify joml-ext

engine/src/main/java/org/terasology/world/chunks/internal/ChunkImpl.java
```diff
@@ -346,8 +346,7 @@ public int chunkToWorldPositionZ(int z) {
     public AABBfc getAABB() {
         if (!aabb.isValid()) {
             org.joml.Vector3f min = JomlUtil.from(getChunkWorldOffset().toVector3f());
-            org.joml.Vector3f max = new org.joml.Vector3f(Chunks.CHUNK_SIZE);
-            max.add(min);
+            org.joml.Vector3f max = min.add(new org.joml.Vector3f(Chunks.CHUNK_SIZE));
```
skaldarnar, line 349: I must have been drunk when suggestion and commiting this 🙈 This obviously changes the value of `min` instead of computing a new `max` ... 🤦 

```suggestion
            org.joml.Vector3f max = new org.joml.Vector3f(Chunks.CHUNK_SIZE).add(min);
```

skaldarnar (approved):

PR #4417, opened 2021-01-20
feat: parallelize tests in `engine`

DarkWeird:
### Contains

1. `Junit 5` run tests in parallel
2. `Javassist` update to 3.27.0-GA - supports java 11.

All tests not marked via `TestResourceLock#CORE_REGISTRY` runs in parallel. (simple tests)

### How to test

1. run `gradlew :engine-tests:test`
    tests should passed.
2. run `engine-tests`'s tests from IDEA

- [X] parallelize MTE tests. (CANCEL. PR not affects module testing)
- [X] fix GameNameTest. - flacky


DarkWeird:
stuck at `Gradle test` fixed.

keturn:
I may have mis-stepped during one of those merge conflict resolutions. In hindsight, I think you were collapsing some setupClass/setup methods to just one `@BeforeEach setup`, presumably because it's easier to parallelize when the test has no `@BeforeAll` on it?

but there may have been one or two of those I resolved the wrong way

engine-tests/src/main/java/org/terasology/PathManagerMocker.java
```diff
@@ -0,0 +1,66 @@
...
+            throw new RuntimeException("Cannot create PathManager", e);
+        }
+    }
+
+    /**
+     * Mock {@link PathManager#getInstance()}.
+     * <p>
+     * AHTUNG! This method using {@code black magic} for JVM bytecode manipulation.
+     * <p>
+     * Redirects calls to {@link #redirectedGetInstance(InvocationOnMock)}
```
keturn, line 45: oh, I think I see why. ugh.

Will https://github.com/MovingBlocks/gestalt/pull/87 let us make PathManager injectable?

Can it have both process-wide and thread-local contexts?

keturn (commented):
I've been afraid of this one because I don't know how much of our stuff has global state or is thread-safe, but now that I'm reviewing it, the changes this PR makes look okay.

One thing I do wonder is: are there clear guidelines for when a test author should use the resource lock?

It's called `CORE_REGISTRY` here. Since we're phasing out CoreRegistry, is it also necessary for Context, or is the way we've been using Context safer?

keturn:
only two failed tests, which is not bad for that merge. The error is an NPE in a setup method. Might be the one I was [worried about](https://github.com/MovingBlocks/Terasology/pull/4417#issuecomment-796189648).

engine-tests/src/main/java/org/terasology/PathManagerMocker.java
DarkWeird: Yeah. It should make PathManager injectable.

Threadlocal context not implemented yet at DI

DarkWeird:
Resourcelock should been used when one of "resource" class is using during test. 
Tte do it automatical.

skaldarnar:
@DarkWeird @keturn what's the state here? Do we see a way to get this through soonish :tm: or should we shelve this PR for now?

DarkWeird:
> @DarkWeird @keturn what's the state here? Do we see a way to get this through soonish :tm: or should we shelve this PR for now?

Abandoned at my side.
I don't want to introduce random code place deadlocks(no one lock, repeatable at my machine) without explicit cause... but idk how to fix it.

Manjaro linux, liberica jdk 11.

PR #4421, opened 2021-01-23
feat(JOML): migrate StorageManager

pollend:


DarkWeird (approved):
Omega - compiles, works

Game loads and creates correctly.
No one exception or something like this.

PR #4424, opened 2021-01-23
feat(JOML): migrate Texture

pollend:
Requires https://github.com/MovingBlocks/joml-ext/pull/17

### DIFF
--- a/engine/src/main/java/org/terasology/engine/subsystem/headless/assets/HeadlessTexture.java
+++ b/engine/src/main/java/org/terasology/engine/subsystem/headless/assets/HeadlessTexture.java
@@ -16,12 +16,11 @@
 package org.terasology.engine.subsystem.headless.assets;
 
 import com.google.common.collect.Lists;
-import org.terasology.joml.geom.Rectanglef;
-import org.terasology.joml.geom.Rectanglei;
 import org.joml.Vector2i;
 import org.terasology.assets.AssetType;
 import org.terasology.assets.ResourceUrn;
-import org.terasology.math.JomlUtil;
+import org.terasology.joml.geom.Rectanglef;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/terasology/rendering/assets/texture/Texture.java
```diff
@@ -18,13 +18,13 @@
...
 import org.terasology.assets.ResourceUrn;
-import org.terasology.math.geom.Rect2f;
+import org.terasology.joml.geom.Rectanglef;
 
 /**
  */
 public abstract class Texture extends TextureRegionAsset<TextureData> {
 
-    public static final Rect2f FULL_TEXTURE_REGION = Rect2f.createFromMinAndSize(0, 0, 1, 1);
+    public static final Rectanglef FULL_TEXTURE_REGION = new Rectanglef(0,0,1,1);
```
jdrueckert, line 27:
```suggestion
    public static final Rectanglef FULL_TEXTURE_REGION = new Rectanglef(0, 0, 1, 1);
```

skaldarnar, line 27: Why not use `Rectanglefc` here as type?

engine/src/main/java/org/terasology/engine/subsystem/headless/assets/HeadlessTexture.java
```diff
@@ -85,7 +84,7 @@ public Texture getTexture() {
 
     @Override
     public Rectanglef getRegion() {
```
skaldarnar, line 86: I think this should ideally return `Rectanglefc`, but this is rooted somewhere in NUI...

skaldarnar (approved):

PR #4427, opened 2021-01-24
Render even chunks that are next to unloaded chunks.

4Denthusiast:
The mesh of each chunk may depend on the contents of the adjacent chunks, for example, the lighting of a block face that is on the edge of a chunk depends on the light value in the adjacent chunk. Before this PR, chunks would only be rendered if the adjacent chunk was loaded, and the mesh would be perfect straight away. With this PR, every chunk is rendered as soon as it loads, as if adjacent unloaded chunks were full of dark air, then if the adjacent chunk is loaded, the mesh is replaced to take into account the new information.

This has several effects:
- Chunks generally become visible faster. This is particularly noticable when first loading a world when world generation is slow, as the first chunk that's created (generally the one containing the player) can render as soon as it's ready, rather than having to wait for all 26 chunks around it.
- There should never be a situation where a chunk is loaded but not visible.
- The ratio of visible chunks to loaded chunks is higher, so low view distances should be more efficient for the amount of the world that's visible.
- Sometimes chunk meshes have to be generated more times. This is a slight decrease in efficiency, but I don't think chunk mesh generation is actually that expensive.
- The edge of the loaded region of the world is usually opaque (at least underground) rather than transparent, as the sides of blocks facing into unloaded chunks are rendered. If a chunk is unloaded, then the meshes of adjacent chunks may not be updated, so there may be holes in the surface in those cases, but it looks like this is rare. This makes it harder for the player to see caves and things that shouldn't be visible if the world is loading slowly.
- I expect that it will make debugging slightly easier in some cases, as it will be easier to see which chunks are loaded.

When testing this, it subjectively seemed like world generation, and chunks becoming available, was faster.

I've tested this with FlowingLiquids and not noticed any problems in that case either (as that's the only module I know that has a non-trivial interaction with chunk mesh generation).

### DIFF
--- a/engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
+++ b/engine/src/main/java/org/terasology/rendering/world/RenderableWorldImpl.java
@@ -25,6 +25,7 @@ import org.slf4j.LoggerFactory;
 import org.terasology.config.Config;
 import org.terasology.config.RenderingConfig;
 import org.terasology.engine.subsystem.lwjgl.GLBufferPool;
+import org.terasology.joml.geom.AABBi;
 import org.terasology.math.JomlUtil;
 import org.terasology.math.TeraMath;
 import org.terasology.monitoring.PerformanceMonitor;
@@ -110,6 +111,12 @@ class RenderableWorldImpl implements RenderableWorld {
                 Collections.sort(chunksInProximityOfCamera, new ChunkFrontToBackComparator());
             } else {
                 logger.warn("Warning: onChunkLoaded called for a null chunk!");
+            }
+        }
+        for (Vector3ic pos : new BlockRegion(chunkCoordinates).expand(1, 1, 1)) {
+            Chunk chunk = chunkProvider.getChunk(pos);
+            if (chunk != null) {
+                chunk.setDirty(true);
             }
         }
     }
--- a/engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
+++ b/engine/src/main/java/org/terasology/world/chunks/localChunkProvider/LocalChunkProvider.java
@@ -173,9 +173,6 @@ public class LocalChunkProvider implements ChunkProvider {
         Chunk[] chunks = new Chunk[region.sizeX() * region.sizeY() * region.sizeZ()];
         for (Vector3i chunkPos : region) {
             Chunk chunk = chunkCache.get(chunkPos);
-            if (chunk == null) {
-                return null;
-            }
             chunkPos.sub(region.minX(), region.minY(), region.minZ());
             int index = TeraMath.calculate3DArrayIndex(chunkPos, region.size());
             chunks[index] = chunk;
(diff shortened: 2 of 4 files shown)


DarkWeird (approved):
Nice! Really chunks loading feel  faster.

I suspect that because chunks able to rendering faster.

Also a catch NPE in `Lost` (i think - not blocker)
I think it is because I stuck in chunk's unloading walls with `ghost` is on.

NPE:

```
11:26:38.165 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
	at org.terasology.lost.LevelSpawnSystem.onBiomeChange(LevelSpawnSystem.java:77)
	at org.terasology.lost.LevelSpawnSystemMethodAccess.invoke(Unknown Source)
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:536)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.sendStandardEvent(EventSystemImpl.java:292)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:283)
	at org.terasology.entitySystem.entity.internal.BaseEntityRef.send(BaseEntityRef.java:205)
	at org.terasology.biomesAPI.BiomeManager.checkBiomeChangeEvent(BiomeManager.java:144)
	at org.terasology.biomesAPI.BiomeManagerMethodAccess.invoke(Unknown Source)
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:536)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.sendStandardEvent(EventSystemImpl.java:292)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:283)
	at org.terasology.entitySystem.entity.internal.BaseEntityRef.send(BaseEntityRef.java:205)
	at org.terasology.logic.characters.KinematicCharacterMover.walk(KinematicCharacterMover.java:650)
	at org.terasology.logic.characters.KinematicCharacterMover.updatePosition(KinematicCharacterMover.java:571)
	at org.terasology.logic.characters.KinematicCharacterMover.step(KinematicCharacterMover.java:103)
	at org.terasology.logic.characters.ServerCharacterPredictionSystem.stepState(ServerCharacterPredictionSystem.java:203)
	at org.terasology.logic.characters.ServerCharacterPredictionSystem.onPlayerInput(ServerCharacterPredictionSystem.java:150)
	at org.terasology.logic.characters.ServerCharacterPredictionSystemMethodAccess.invoke(Unknown Source)
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:536)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.sendStandardEvent(EventSystemImpl.java:292)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:283)
	at org.terasology.entitySystem.entity.internal.BaseEntityRef.send(BaseEntityRef.java:205)
	at org.terasology.logic.players.LocalPlayerSystem.processInput(LocalPlayerSystem.java:194)
	at org.terasology.logic.players.LocalPlayerSystem.update(LocalPlayerSystem.java:155)
	at org.terasology.engine.modes.StateIngame.update(StateIngame.java:171)
	at org.terasology.engine.TerasologyEngine.tick(TerasologyEngine.java:496)
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:459)
	at org.terasology.engine.TerasologyEngine.runMain(TerasologyEngine.java:435)
	at org.terasology.engine.TerasologyEngine.run(TerasologyEngine.java:401)
```

PR #4429, opened 2021-01-24
feat(JOML): remove use of termath methods from Material

pollend:
there are no more references for any of these methods so we can safely remove them from the engine.


DarkWeird (approved):
Nice.
Omega builds. on develops.
But what about old modules?

PR #4447, opened 2021-01-31
feat(JOML): remove deprecated methods from OnBlockItemPlaced

pollend:


keturn (approved):
does what it says on the tin

PR #4448, opened 2021-01-31
feat(JOML): remove MatrixUtils

pollend:
There are no more instance of MatrixUtils so we can remove this.


keturn (approved):
The removed class was deprecated. Looks like fair game to me. 🪦 MatrixUtils

Issue #4367, opened 2021-01-02
Investigate and enhance in-game module downloader, keeping extra attention on jar files in the modules directory


Cervator:
With #4343 we "lose" the ability _at development time_ to use jar files from `modules/`  in favor of having Gradle resolve binary dependencies and load them onto the classpath from the Gradle cache directory instead.

Running the game from an actual game zip should still work the same - load jar files from the modules dir. You can also simply use `groovyw module recurse x` to get everything as source and you _should_ get the source dir to take precedence over any jars in the Gradle cache.

However, in at least one case there's a thing that relies on jar files being placed in `modules/` - the in-game module downloader, which _can_ be used to fetch jar files at runtime, which will then sit around in your modules dir possibly looking confusing, if they're not actually being used. The module downloader also seems to be rarely used, and could probably use some validation and fixes.

Here are some possible things that could be done on this topic - a mix of dev workspace or game zip contexts

- [ ] Validation: Check on the module downloader and see if it actually works from a game zip (start a server from one dir, connect from a client in another dir with a module jar missing, does it download correctly and work?), possibly in a state _before_ #4343
- [ ] Feature: Find a new way to support jar file modules at run-time for a dev workspace?
- [ ] Feature: Do _not_ try to actually update to a newer version of the _engine_ jar ... that's not really hot swappable, but can trigger if for instance you're trying to connect to a server running a release version yet you've got the next dev version which _probably_ should still work even at a slightly newer engine version
- [ ] Feature: Better react when a module is added at runtime. At present for instance key mappings are not re-initialized so any unique key bindings from a new module are missing until the next game restart
- [ ] Validate: Check that a source module directory does actually win in precedence over a jar - either one in the Gradle cache or even one from `modules/` if support is re-added to use from there at runtime (from a dev workspace)
- [ ] Feature: Allow the _distribution_ of a module from a developer workspace - at present this doesn't work as you can't transfer a source module to another user connecting to your running server. Not sure if it would work using a module jar from the build dir somehow?
- [ ] Validate: Really perhaps more a part of #4343 but it would be nice to be sure we have good game logging or alternatives (like `gradlew dependencies`) that makes it clear which exact module code is being loaded from where

keturn:
Some clarification:

#4343 **adds** more gradle involvement in resolving module dependencies, and along with that makes sure they get on the runtime classpath when we do `gradlew game` or run from the IDE. Any run-time module-scanning code that is done by the engine or gestalt is still as it was.

[...]

PR #4454, opened 2021-02-01
build(PC): move modules back off the classpath so the sandbox works

### LINKED ISSUE #4375: dev workspace: modules on the java classpath breaks the sandbox
Despite my best efforts to find things out _before_ merging build changes (#4343), there are some things that are only coming out in the aftermath.

Excerpts from a recent discussion in [#gestalt](https://discord.com/channels/270264625419911192/714313228271747152/795526922490609676):

>  If the modules are in the main classloader, they're not really modules at all at that point. Just code included in terasology.
[…]
> So for actual release use, module code (and resources) are not on the base classloader, except the engine module or any other built-in module.
> Ideally development runs would reflect this
> partially because it would help detect security related issues
> and partially because some subtle behavior differences from having a slurry of module code in the base classloader

I think two things come from this:

1. Empirically confirm how broken things are. This might be as straightforward as: edit a module so it imports something it shouldn't, run `gradlew game` and _Run TerasologyPC_ with that as a local subproject, see if it succeeds at importing that thing (and maybe invoking a method on it or something).
    I think there _may_ be a possibility that modules still get loaded by the sandboxed loader despite the fact that they would be available to the default loader if it asked for them directly, but that needs research to find out if it's true.
2. Decide if we're willing to compromise on this module security _when running in a dev workspace._ (If there's a vulnerability, it probably applies to all subprojects present in `/modules/` _and_ all their dependencies, whether present as subprojects or fetched as artifacts.)

-- 

keturn's understanding of how this interacts with the security model:

The thing that prevents Terasology Modules from importing code that would give them capabilities they're not supposed to have is the classloader used; when loading the module's class, it handles the requests for the dependencies of that class, and may deny the requests that are not approved.

If the module's class is loaded by the classloader that loads the trusted code (for engine & facade), no such restrictions are applied. If the module's class is on the classpath, it _may_ be loaded by that classloader.

keturn:
Changing facade's dependencies so all the modules still get their dependencies without being on the classpath.

Adds a new `cacheModules` directory that automatically-resolved jars get put in, kept separate from the `modules/` source directory.

Fixes #4375.

### How to test

- [x] merge parent #4347. 
- [ ] #4463 is not required, but would remove some noise from this review.
  - [ ] if #4463 is rejected, re-integrate `serverModules`
- [x] build module cache as an output of `:modules` project instead of defining the task in `:facades:PC`


keturn:
Current status: The `provideModuleDependencies` task is a good proof of concept of how to do the jar-gathering thing without the DependencyResolutionListener hooks. Works nicely under gradle. IntelliJ doesn't realize it's supposed to build the modules at all.

It might be worth looking up how IntelliJ _does_ make its dependency graph out of the gradle projects. I'm assuming it's only checking the `main` `runtimeClasspath` or something like that, so when we put the module dependencies in a non-`main` gradle Configuration it doesn't take them into account.

I guess we can probably come at this from two directions: the gradle side or the IntelliJ side.

From the IntelliJ side, we could edit the run configs so they have to do some additional build-the-modules thing before they start.

From the gradle side, we could include the modules in main runtimeClasspath, but then have the exec tasks use some other classpath without the modules?

I spent a while rummaging around in the intellij-gradle sources. I am far from understanding how it works, but this test suite is informative: https://github.com/JetBrains/intellij-community/blob/211.5361/plugins/gradle/testSources/org/jetbrains/plugins/gradle/importing/GradleDependenciesImportingTest.java

The tests include non-default Configurations and SourceSets, so maybe we don't need to put everything in `main` after all. In which case it's a matter of how we declare the dependency on that moduleCache, or it's our use of `platform`. (I don't see any test cases using platform there.)

Cervator:
Going to also link this to #4367 for relevance, since I guess if we have a _different_ module directory that might be a good place to direct the in-game module downloader to in some fashion when running from source, especially if it can sort out dependencies (am realizing I never tried using the in-game downloader to grab something with dependencies - no idea what would happen then)

As for testing - ran this branch with `gradlew distForLauncher` in a workspace without modules, was then able to run the distributed game fine, very smooth - even if that just tests the dists overall. Added all Omega modules as jars and game still ran, although naturally something crashed because JOML ;-) 

Maybe we could get TutorialMinimalEngineDemo some quick love in the near term, only two errors or so in there, and it is my quick goto when moduleland is broken. But me trying to bumble through a fix in the narrow little window I found tonight didn't really go anywhere good. So actually testing from source or breaking the sandbox I haven't been able to do just yet.

(Not that TMED would let me test that without hacking in a simple violation in its module - even Sample needs some sort of world and an inventory to be able to retrieve and trigger the crash blocks ...)

buildSrc/src/main/kotlin/org/terasology/gradology/module_deps.kt
```diff
@@ -0,0 +1,86 @@
...
+        if (it is UnresolvedDependencyResult) {
+            logger.warn("Dependency {} of {} not resolved:", it.attempted, it.from, it.failure)
+        } else {
+            logger.warn(
+                "Dependency {} of {} not resolved: Unexpected result class {}\n{}",
+                it.requested,
+                it.from,
+                it::class,
+                it
+            )
```
keturn, line 65: some auto-formatting operation? I didn't mean to make this log statement take so many times.

buildSrc/src/main/kotlin/terasology-dist.gradle.kts
```diff
@@ -0,0 +1,74 @@
...
+        zipFile.set(taskProvider.flatMap { it.archiveFile })
+    }
+
+    /**
+     * Assert two objects are equal.
+     *
+     * @see kotlin.test.assertEquals
+     */
+    fun assertEquals(expected: Any, actual: Any, message: String? = null) =
+        kotlin.test.assertEquals(expected, actual, message)
```
keturn, line 55: can we export these kotlin.test functions without adding them to the class like this?

keturn: https://github.com/MovingBlocks/Terasology/pull/4454/commits/512e8cfe201c98a494ea965a542024b2ccf307d2 works for this use case

Cervator (approved):
I'm still not going to claim that I understand all the new Kotlin-based Gradle magic, or pretend I've really been able to spend the time on this that it deserves, but ... I did at least spend _some_ time reviewing this now (the whole thing including the two related Gradle PRs), and testing it too. Really nice work 👍 And yes, this length review for this kind of thing does _not_ count as thorough from me ... 😅 

Testing: I've tried this both from source, from a dist zip, and as an Omega zip built in Jenkins via http://jenkins.terasology.io/teraorg/job/Terasology/job/Omega/job/dists/3/ - I ran locally with a bare workspace containing only TutorialMinimalEngineDemo (with @skaldarnar's quick JOML fix applied, ty again!). A `gradlew clean distForLauncher` gave me a clean game zip that ran the game fine even though it didn't bundle the module jar at all (not sure if that should fail if something classpathy tries to write to a manifest somewhere, but seemed fine). Didn't have any modules to actually get into a world with, of course.

I then did a `gradlew jar game` and validated that TMED worked fine normally, then added `System.out.println("I crash!");` to `HeldItemSystem.onGiveItemToCharacterHoldItem` to force a sandbox violation when you "break" a block, ran the game again, and successfully crashed with a sandbox violation. Grabbed the module jar (and the CoreRendering one I found in `modules/` which was probably old, I see one in `moduleCache` as well), stuffed them into the unzipped dist zip under `modules/` then ran the game again and successfully crashed on the sandbox again. Fixed TMED, re-jarred, tested again, no crashing 👍 

Also tried multiplayer but TMED wasn't a fan there, something about the held item system and player 2 which is no big surprise and no concern. Host was fine.

Tested with the [special Omega zip from Jenkins](http://jenkins.terasology.io/teraorg/job/Terasology/job/Omega/job/dists/3/artifact/distros/omega/build/distributions/TerasologyOmega.zip) and I guess the main menu label shows fine, just may be a stray quirk in the manifest version in a jar as noted with a line comment. Ran JS + Sample, which broke over JOML, okay fine, I'm sure it "works" normally .. tried with Core + Sample and game launched fine, the `CrashGameBlock` caught its error without fully crashing (not new) while `ReallyCrashGameBlock` did in fact fully crash the game (and now you know why the second one was made later ...)

I like that if you run the game via bat on Windows now you always get the process to retain its state in the command window logging stuff there, can terminate with a break, and so on. Running via exe on CLI orphaned the process. Which I guess still happens if you just double click the bat? Let me try that if I remember ... oh ya, okay, so launch via bat gets you both the `cmd` window with logging _and_ the game window, but both get legit icons. Alright cool. I guess that was one "benefit" with the exe which was also a curse. It was a little cleaner, but also hid more things you might care about (or fail to exit fully without any obvious indication). `cmd` window poofs if you exit cleanly that way.

Potentially trivial request that may not be worth it but: could we perhaps use `cachedModules` rather than `moduleCache` ? Now trying to auto-complete "modules" ends up auto-completing the other one ... 😅 

Less trivial but more mysterious: Really not sure what to do with the tests (got a couple line comments on those too). These seem like critical dist tests, yet the regular unit tests, analytics, etc, actually run _after_ we'd done a dist and attached it in Jenkins as a completed build. Yet this seems like a thing you'd want to run before that. And eventually we _should_ let analytics and unit tests fail the build.

Maybe the solution is to solely do a `classes` step early on, then in a Jenkins parallel start all the analytics and tests, including this one. At the very end after all that if health indicators a fine then do the dist, attach the artifacts, actually _record_ the analytics and tests, and finish. There's a bit more to that related to workspace reuse and where which output files may go to get sent back to Jenkins (maybe a quick `stash` thing)

That's probably a task for something other than this PR though. I'd be thrilled to merge it as-is (ship it! And its buddies), or with whatever tweaks may come from this review.

Finally and another functional one - out of curiosity: with the new `Terasology.bat` being generated, does it include setting default memory stats like the old one did?

`java -Xms128m -Xmx1536m -jar libs/Terasology.jar "$@"` (well, Linux executable, but anyway)

I see `TERASOLOGY_OPTS` mentioned in the bat, which is mildly amusing to me, for some reason, in a cool way. Could we literally set memory stats, -headless, -homedir, etc in an env var on, say, the play test server and everything is peachy each launch?

In any case:

:shipit: 🚀 💎 👐

facades/PC/build.gradle.kts
```diff
@@ -69,78 +47,39 @@ val displayVersion = versionBase
...
-
 // Base the engine tests on the same version number as the engine
 version = project(":engine").version
 logger.info("PC VERSION: {}", version)
 
 // Jenkins-Artifactory integration catches on to this as part of the Maven-type descriptor
 group = "org.terasology.facades"
 
+configurations {
+    register("serverModules") {
```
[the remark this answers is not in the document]
keturn: I haven't tested these server tasks very much at all, no. Just a bit during the conversion to .kts to make sure gradle they did something that looked reasonable at a glance.

If they're not used, I'd be happy to drop them. There's a fair amount of code in here that's specific to supporting that extraModules property and writing the config file.

facades/PC/build.gradle.kts
```diff
@@ -182,30 +122,58 @@ tasks.register<Sync>("setupServerModules") {
...
+    archiveFileName.set("Terasology.jar")
+
+    manifest {
+        //TODO: Maybe later add the engine's version number into here?
+        attributes["Main-Class"] = mainClassName
+        // A classpath in the manifest avoids the problem of having to put a classpath on the command line and
+        // "line is too long" errors: https://github.com/gradle/gradle/issues/1989
+        attributes["Class-Path"] = configurations["runtimeClasspath"].joinToString(" ") { it.name }
+        attributes["Implementation-Title"] = "Terasology-" + project.name
+        attributes["Implementation-Version"] = """${env["BUILD_NUMBER"]}, ${env["GIT_BRANCH"]}, ${env["BUILD_ID"]}"""
```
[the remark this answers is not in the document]
keturn: ah, good catch, I may have merged it correctly for the version file but not the jar manifest here or something like that

facades/PC/build.gradle.kts
```diff
@@ -216,103 +184,122 @@ tasks.register<Copy>("createVersionFile") {
...
     }
 }
 
-tasks.register<Zip>("distPCZip") {
-    group = "terasology dist"
-    dependsOn("distApp")
-    dependsOn("distModules")
-    from("${distsDirectory.get().asFile}/app")
-    archiveFileName.set("Terasology.zip")
+tasks.register<ValidateZipDistribution>("testDistZip") {
```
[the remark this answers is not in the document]
keturn: nope! distForLauncher makes this zip with the nonstandard directory names.

in the future I think it'll be easier to make Launcher accept the layout distZip uses, and then we can get rid of the StartScriptClasspathRewriter mess that is required to support distForLauncher.

buildSrc/src/main/kotlin/org/terasology/gradology/exec.kt
```diff
@@ -0,0 +1,67 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.gradology
```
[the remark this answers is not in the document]
keturn: I needed a package namespace for this stuff and kept getting stuck. Having a subdirectory named "gradle" seems like it would risk being confused with the "/gradle" directory that holds the wrapper in all gradle projects, and other words like "build" have similar problems

buildSrc/src/main/kotlin/terasology-dist.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+ * Make assertions about the content of a zip file.
+ *
+ * Do make sure you put any inspection of the file in the task's *execution phase,* not in its *configuration.*
+ * i.e. use a `doLast` block.
+ *
+ * TODO: Produce proper test output instead of just throwing an exception and failing the build.
+ *   Probably with [TestKit](https://docs.gradle.org/current/userguide/test_kit.html)?
+ */
+@Suppress("UnstableApiUsage")
+open class ValidateZipDistribution @Inject constructor(objects: ObjectFactory) : DefaultTask() {
```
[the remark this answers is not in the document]
keturn: I see you figured this out when you got to the part where it's used

keturn:
I merged #4347, which makes the diff of this vs develop a more manageable size.

facades/PC/build.gradle.kts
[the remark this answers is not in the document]
keturn: addressed in #4347

keturn (commented):
I've now got the dependency from PC facade to the module cache in a way I think is _Gradle Approved._ :elephant: 

I want to do a little testing of how it behaves in IntelliJ. If that goes okay, I'm ready to call this good.

.gitignore
```diff
@@ -72,7 +72,7 @@ Thumbs.db
 /configs/
 /screenshots/
 /logs/
-/moduleCache/
+/cacheModules/
```
Cervator, line 75: `cache` or `cached` ?

buildSrc/src/main/kotlin/org/terasology/gradology/attributes.kt
```diff
@@ -0,0 +1,11 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.gradology
+
+import org.gradle.api.Named
+import org.gradle.api.Project
+
+inline fun <reified T: Named> Project.namedAttribute(value: String) = objects.named(T::class.java, value)
```
Cervator, line 9: Maybe a comment or two (is it still Javadoc in Kotlin? 🤔 ) to help explain these kinds of things. This looks pretty intense to me :-)

config/gradle/common.gradle
```diff
@@ -16,6 +16,8 @@ apply plugin: 'com.github.spotbugs'
 apply plugin: 'jacoco'
 apply plugin: 'org.sonarqube'
 
+apply plugin: 'terasology-repositories'
```
Cervator, line 19: 🎉

Cervator (approved):
Still looks fine to me, couple trivial comments :-)

buildSrc/src/main/kotlin/org/terasology/gradology/attributes.kt
keturn: (it's [KDoc](https://kotlinlang.org/docs/reference/kotlin-doc.html).)

This one in particular is lifted straight out of the gradle examples. It's kinda weird, someone writing the docs realized you don't really want to type `objects.named(LibraryElements::class.java, name)` every time, so they made this helper... but didn't include it in the api.

config/gradle/common.gradle
keturn: the `$RESOLUTION_REPO` and `-PalternativeResolutionRepo` things I haven't tested. There's some chance of some URI type-conversion hiccup happening there, but probably not and it should be a quick fix if there is.

facades/PC/build.gradle.kts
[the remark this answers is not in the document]
keturn: moved to #4463

keturn:
SHIP IT

There have been some small changes since Cerv's last review, but I'm confident enough in them to move ahead without a re-review. I'll at least wait for CI to finish. It would be way too embarrassing if one of my "small" changes broke that.

## Test Plan

* `groovyw get CoreSampleGameplay Explosives Sample` # note this is **not** `recurse`
* `./gradlew clean extractConfig extractNatives`
* launch game. One of
  * for testing **gradle:**
    * `./gradlew game`
  * for testing **IntelliJ:** 
    * Click the **🔄 Reload All Gradle Projects** button in the Gradle Tool Window. If you don't do this (or restart IntelliJ) after checking out new things to `modules/`, it might not have recognized them.
    * Run **TerasologyPC** configuration
* assert game starts
* assert Advanced Game Setup screen shows CoreSampleGameplay and Sample modules **and** all their dependencies, and they are not error-color-red.
* activate CoreSampleGameplay and Sample, start a new game
* get the gun from Explosives (it's in the second chest CoreSampleGameplay puts in the inventory of new players), make sure it has its shoot-and-destroy-block effects. (There's a possible failure condition where the module loads assets but not code, in which case you can have a gun that doesn't shoot.)
* `give ReallyCrashGameBlock 1` in the terasology console
* place the ReallyCrashGameBlock
* assert the game really crashes when you **Use** it.

### Troubleshooting

If things don't go as expected, a few things to check are:

* did Sample and Explosives actually build their code?
  * `fdfind -Ig '*.class' modules/`
* did the modules cache get filled?
  * `ls cachedModules`  # you should see `.jar` files for dependencies here, like `CoreRendering` and `Inventory`. There should _not_ be jars for modules we have source for in `modules/`, or for non-terasology-module things.
* For IntelliJ: look in **Project Structure / Modules /** `facades/PC/main` **/ Dependencies** and see if our source modules are included in that list, e.g. `Terasology.modules.Explosives.main`. 

IntelliJ Project Structure window

PR #4456, opened 2021-02-02
feat(JOML): migrate and cleanup EntityAwareWorldProvider

pollend:


engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
```diff
@@ -118,14 +116,14 @@ public void shutdown() {
...
-            EntityRef blockEntity = getBlockEntityAt(JomlUtil.from(pos));
+            EntityRef blockEntity = getBlockEntityAt(pos);
             Block oldType = super.setBlock(pos, type);
             final Set<Class<? extends Component>> retainComponents =
                     Optional.ofNullable(blockEntity.getComponent(RetainComponentsComponent.class))
                             .map(retainComponentsComponent -> retainComponentsComponent.components)
                             .orElse(Collections.emptySet());
             if (oldType != null) {
-                updateBlockEntity(blockEntity, JomlUtil.from(pos), oldType, type, false, retainComponents);
+                updateBlockEntity(blockEntity, pos, oldType, type, false, retainComponents);
```
keturn, line 126: a bunch of removing `JomlUtil.from`, which makes me feel like it's moving in the right direction :+1:

engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
```diff
@@ -150,35 +148,29 @@ public Block setBlock(Vector3ic pos, Block type) {
...
-                updateBlockEntity(blockEntity, JomlUtil.from(position), oldType, type, false, Sets.newHashSet(components));
+                updateBlockEntity(blockEntity, position, oldType, type, false, Sets.newHashSet(components));
             }
             return oldType;
         }
         return null;
     }
 
-    private void updateBlockEntity(EntityRef blockEntity, Vector3i pos, Block oldType, Block type,
+    private void updateBlockEntity(EntityRef blockEntity, Vector3ic pos, Block oldType, Block type,
```
keturn, line 173: +1 to using the constant type where we can

engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
```diff
@@ -189,76 +181,55 @@ private void updateBlockEntity(EntityRef blockEntity, Vector3i pos, Block oldTyp
...
         blockEntity.send(changedEvent);
     }
 
     @Override
-    public EntityRef setPermanentBlockEntity(Vector3i blockPosition, EntityRef blockEntity) {
+    public EntityRef setPermanentBlockEntity(Vector3ic blockPosition, EntityRef blockEntity) {
         if (GameThread.isCurrentThread()) {
             EntityRef oldEntity = getExistingBlockEntityAt(blockPosition);
-            blockEntityLookup.put(blockPosition, blockEntity);
+            blockEntityLookup.put(new Vector3i(blockPosition), blockEntity);
```
keturn, line 196: I'm having a hard time working out where it's appropriate to add allocations. Why do we copy blockPosition here? Is it because the Vector3ic we get isn't really immutable, that's only a promise that _this method_ doesn't mutate it?

engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
```diff
@@ -377,7 +348,7 @@ private void updateBlockEntityComponents(EntityRef blockEntity, Block oldType, B
 
     private EntityRef createBlockEntity(Vector3ic blockPosition, Block block) {
         EntityBuilder builder = entityManager.newBuilder(block.getPrefab().orElse(null));
-        builder.addComponent(new LocationComponent(new org.joml.Vector3f(blockPosition)));
+        builder.addComponent(new LocationComponent(new Vector3f(blockPosition)));
```
keturn, line 351: seems like we could use a LocationComponent(Vector3ic), but maybe that's out-of-scope for this PR

engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
```diff
@@ -460,37 +417,37 @@ public void onActivateBlock(OnActivatedComponent event, EntityRef entity) {
...
     }
 
     @ReceiveEvent(components = {BlockRegionComponent.class})
     public void onBlockRegionActivated(OnActivatedComponent event, EntityRef entity) {
         BlockRegionComponent regionComp = entity.getComponent(BlockRegionComponent.class);
         blockRegions.put(entity, regionComp.region);
-        for (org.joml.Vector3ic pos : regionComp.region) {
-            blockRegionLookup.put(new org.joml.Vector3i(pos), entity);
+        for (Vector3ic pos : regionComp.region) {
+            blockRegionLookup.put(new Vector3i(pos), entity);
```
keturn, line 430: and again with the copying of the key.

if it is a pattern we want to keep, would it be worth making a Map class with some `Map.putWithKeyCopy()` or some such?

engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
pollend: yea, Vector3ic is a promise that the variable is not mutated. so you have to create a new Vector3i so another party doesn't modify the vector.

engine/src/main/java/org/terasology/world/BlockEntityRegistry.java
```diff
@@ -39,20 +37,7 @@
...
-     *
-     * @param blockPosition absolute position of the block
-     * @return The block entity for the location if it exists, or the null entity
-     * @deprecated This is scheduled for removal in an upcoming version
-     *             method will be replaced with JOML implementation {@link #getExistingBlockEntityAt(Vector3ic)}.
-     **/
-    @Deprecated
-    EntityRef getExistingBlockEntityAt(Vector3i blockPosition);
-
+    EntityRef setPermanentBlockEntity(Vector3ic blockPosition, EntityRef bockEntity);
```
jdrueckert, line 40:
```suggestion
    EntityRef setPermanentBlockEntity(Vector3ic blockPosition, EntityRef blockEntity);
```

engine/src/main/java/org/terasology/world/BlockEntityRegistry.java
```diff
@@ -39,20 +37,7 @@
      * @param bockEntity    The new entity to set
```
jdrueckert, line 37:
```suggestion
     * @param blockEntity    The new entity to set
```

jdrueckert (changes requested):

engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
skaldarnar: I don't think this would be worth it, because we cannot easily augment this on existing `Map` implementations.

skaldarnar (approved):

jdrueckert (approved):

engine/src/main/java/org/terasology/world/internal/EntityAwareWorldProvider.java
keturn: you could in Kotlin 😉

Short of that, though, these lookup maps are private members of the class, so they could switch to being some implementation without much disruption.

pollend: welp we're not using kotlin. @keturn you will just have to keep dreaming.

PR #4457, opened 2021-02-02
feat(JOML): migrate RotationTest

pollend:


keturn (approved):
okay, sure, I'll believe that rotation.orientation.transform is a replacement for rotation.getQuat4f.rotate.

PR #4458, opened 2021-02-02
test(JOML): migrate and simplify PojoPrefabManagerTest

pollend:


keturn (commented):
This looks like a straightforward type migration, but I am confused what is going on in practice.

This change to tests is not accompanied by any change to the code under test

so, uh

what happens if you remove those two lines instead of changing them? how do they contribute to the test?

pollend:
looks like its only using AssetManager and the test uses a test object.

skaldarnar (approved):

PR #4459, opened 2021-02-02
feat(JOML): remove tera-assert

pollend:
remove usage of TeraAssert. not quite sure what to do with #assertEqualsContent. is there a clean alternative we could adopt that could streamline some of this.


engine-tests/src/test/java/org/terasology/entitySystem/OwnershipHelperTest.java
```diff
@@ -67,6 +74,21 @@ public void testListsOwnedEntities() {
         EntityRef ownerEntity = entityManager.create(ownerComp);
 
         OwnershipHelper helper = new OwnershipHelper(entityManager.getComponentLibrary());
-        assertEqualsContent(Lists.newArrayList(ownedEntity), helper.listOwnedEntities(ownerEntity));
+        assertEqualsContent(helper.listOwnedEntities(ownerEntity), Lists.newArrayList(ownedEntity));
     }
+
+    //TODO: needs to be replaced with alternative
+    public static <T> void assertEqualsContent(Collection<? extends T> expected, Collection<? extends T> actual) {
```
keturn, line 81: [assertIterableEquals](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertIterableEquals(java.lang.Iterable,java.lang.Iterable))

but for this specific case, you're looking at a list you expect to have exactly one element, so you could also write an assertion that it is of size 1 and that its element has the expected value.

engine-tests/src/test/java/org/terasology/persistence/serializers/VectorTypeSerializerTest.java
```diff
@@ -61,61 +50,17 @@ public void setup() {
...
-        });
-
-        assertEquals(o.v1, new org.joml.Vector3f(11.5f, 13.15f, 3), .00001f);
-        assertEquals(o.v2, new org.joml.Vector2f(12f, 13f), .00001f);
-        assertEquals(o.v3, new org.joml.Vector4f(12, 12.2f, 3f, 15.5f), .00001f);
-
-    }
-
-    @Test
-    public void testProtobufSerializeRemapped() throws IOException {
```
keturn, line 100: Why drop these serialization tests?

keturn (changes requested):

engine-tests/src/test/java/org/terasology/persistence/serializers/VectorTypeSerializerTest.java
pollend: they were for mapping between termath and joml. since we don't really have any strong backwards support it doesn't hurt to drop it.

skaldarnar (approved):

PR #4460, opened 2021-02-02
chore[IntelliJ]: exclude compiler.xml from version control

keturn:
Its values are obtained from the gradle configuration.

See https://youtrack.jetbrains.com/issue/IDEA-251986

cherry-picked from c1d63f5914c8bf3425db924961ceec7f6e49db12


pollend (approved):
seems like a pretty sensible change.

DarkWeird (approved):

PR #4462, opened 2021-02-03
fix: remove EntityBuffer from scaled world generation methods

4Denthusiast:
This prevents a crash in MetalRenegades when LOD chunks are enabled, caused by an entity provider trying to use facets that the world generator isn't able to produce scaled versions of.

I don't know what the entity provider is actually used for, so it would be good for someone who does to check that that is still working properly in non-LOD chunks with this PR.


engine/src/main/java/org/terasology/world/generation/WorldImpl.java
```diff
@@ -67,9 +67,6 @@ public void rasterizeChunk(CoreChunk chunk, EntityBuffer buffer, float scale) {
         for (WorldRasterizer rasterizer : scalableWorldRasterizers) {
             ((ScalableWorldRasterizer) rasterizer).generateChunk(chunk, chunkRegion, scale);
         }
-        for (EntityProvider entityProvider : entityProviders) {
-            entityProvider.process(chunkRegion, buffer);
```
DarkWeird, line 71: You can remove `EntityBuffer buffer` parameter, because it not used.
This object used only for entity generation during world generation.

skaldarnar (approved):

skaldarnar (approved):

PR #4463, opened 2021-02-03
build(PC): remove unused tasks for running headless server from source

keturn:
In the PC facade build config, removes tasks `server`, `setupServerModules`, `setupServerConfig` at the suggestion of https://github.com/MovingBlocks/Terasology/pull/4454#discussion_r569120726

Those setup tasks have a significant amount of code so removing them will slim the build config down considerably.


keturn:
Cerv clarifies:

> the headless run config in IntelliJ and the server JavaExec in Gradle without all the awkward extras that try to generate a config.cfg for you really should be enough. The only real thing those extras (setupServerConfig + setupServerModules) did was allow you to not have to hand-configure override.cfg - which is so much easier at this point
> those are the bits i meant to say could probably go away, not the tasks.register<JavaExec>("server") thing itself. It could just hard code the -homedir=... arg and include the flag for override.cfg instead, much like the IntelliJ run config

keturn (commented):
Updated to leave the `server` exec task intact.

facades/PC/build.gradle.kts
```diff
@@ -105,37 +98,9 @@ tasks.register<RunTerasology>("permissiveNatives") {
...
-    from(configurations.named("serverModules"))
-    into(File(rootProject.file(localServerDataPath), "modules"))
-}
-
 // TODO: Make a task to reset server / game data
 tasks.register<RunTerasology>("server") {
     description = "Starts a headless multiplayer server with data stored in [project-root]/$localServerDataPath"
-    dependsOn("setupServerConfig")
-    dependsOn("setupServerModules")
     args("-headless", "-homedir=$localServerDataPath")
```
Cervator, line 104: You could add `-overrideDefaultConfig=override.cfg` here to match the headless run config for IntelliJ 👍

Cervator (approved):
Yay less mess! One bonus comment left that would be an enhancement, previously to test headless via Gradle we'd either need to tweak something manually ahead of time or start the server once, edit the generated `config.cfg` and restart.

PR #4466, opened 2021-02-06
fix(ModuleManager): fix improperly-escaped spaces in module path

### LINKED ISSUE #4407: Space in path to modules on Windows (at least) can cause game crash when running from source
Split out from #4392

Run the game (for some sort of testing) from a directory like `D:\Dev\Terasology\Git_WS\Pristine Omega\Terasology` using `gradlew jar game`

Alternatively run from source with an OS username containing a space, meaning there's a space in the path to the default location of the Gradle cache, with a jar file used from there

### What actually happened

💥 

### How to reproduce

* Have a workspace either with a space in its path or use a Gradle cache from a location with a space in the path
* Run the game, such as with `gradlew jar game`

One snippet from a run involving the Gradle cache by a user reporting the issue on GitHub:

`java.lang.IllegalArgumentException: Module does not exist at: C:\Users\Naman%20Tayal\.gradle\caches\modules-2\files-2.1\org.terasology.modules\Potions\1.4.0-SNAPSHOT\66b633908549499a1cc563af4574f976c60e4dbe\Potions-1.4.0-SNAPSHOT.jar`

Snippet from me trying in an Omega workspace with a space in the root directory name:

`java.lang.IllegalArgumentException: Module does not exist at: D:\Dev\Terasology\Git_WS\Pristine%20Omega\Terasology\modules\JoshariasSurvival\build\libs\JoshariasSurvival-1.1.0-SNAPSHOT.jar`

### Computer details

Win10 for me, some form of Windows from the Discord user reporting the problem.

keturn:
Fixes #4407 

Honestly, after all the trouble we just went through when we unwittingly added modules to the classpath, I don't even know why we have a `loadModulesFromClasspath` method that runs by default. But we do, so this is a fix for it.

### How to test


engine/src/main/java/org/terasology/engine/module/ModuleManagerImpl.java
```diff
@@ -126,7 +125,7 @@ public ModuleManagerImpl(Config config, List<Class<?>> classesOnClasspathsToAddT
     /**
      * Overrides modules in modules/ with those specified via -classpath in the JVM
      */
-    private void loadModulesFromClassPath() {
+    void loadModulesFromClassPath() {
```
Cervator, line 128: Intentional package scope change! Unicorn spotted in the wild 🦄

Cervator (approved):
Yet another small and useful PR, with tests even, yay! 😍 

Can confirm Gradle launching the game from source works with a space in the path on Win10, and new tests pass 👍

engine/src/main/java/org/terasology/engine/module/ModuleManagerImpl.java
skaldarnar: @jdrueckert will approve opening this up for testing :nerd_face:

skaldarnar (approved):

PR #4467, opened 2021-02-06
build(PC): add gradle command-line option for max heap size

keturn:
Adds a gradle command line option to all RunTerasology tasks to set the maximum java heap size for the process.

e.g.
```
./gradlew game --max-heap 8G
```

appears in
```
./gradlew help --task game
```


Cervator (approved):
Huh. So easy and neat. Is this a superpower from doing our own Gradology? I remember looking for cleaner args for Gradle ages ago and it just went back to the whole "just pass weird `-pSomeArg=Blah` things and capture them as properties within Gradle ..."

skaldarnar (approved):

keturn:
yeah, you [have to make a Task subclass to do it](https://docs.gradle.org/6.8.2/userguide/custom_tasks.html#limitations):

> Support for declaring command line options currently comes with a few limitations.
>
> * Command line options can only be declared for custom tasks via annotation. There’s no programmatic equivalent for defining options.

PR #4472, opened 2021-02-06
perf: avoid string operations when validating visible blocks

pollend:
This logic is called heavily when generating meshes for chunks. in this burn down chart you can see a pretty significant amount of the time is spent converting a blockURI to a string. 

image


engine/src/main/java/org/terasology/rendering/primitives/BlockMeshGeneratorSingleShape.java
```diff
@@ -149,7 +150,7 @@ private boolean isSideVisibleForBlockTypes(Block blockToCheck, Block currentBloc
             return true;
         }
 
-        if (blockToCheck.getURI().toString().equals("engine:unloaded")) {
+        if (blockToCheck.getURI().equals(BlockManager.UNLOADED_ID)) {
```
skaldarnar, line 153: comparing against a constant is way better anyways. and if this ends up doing less work that's even better! :+1:

skaldarnar (approved):

4Denthusiast:
Oops, sorry. I just assumed that the URI had the string stored in it anyway.

PR #4476, opened 2021-02-06
feat(JOML)!: drop Region3i

jdrueckert:


engine/src/main/java/org/terasology/math/JomlUtil.java
```diff
@@ -189,24 +189,6 @@ public static AABBf from(AABB aabb) {
...
-
-
-    public static Region3i from(BlockRegion aabb) {
-        if (aabb == null) {
-            return null;
-        }
-        return Region3i.createFromMinMax(JomlUtil.from(aabb.getMin(new Vector3i())), JomlUtil.from(aabb.getMax(new Vector3i())));
-    }
-
-
```
pollend, line 209: there is no need to drop this. the whole class will be removed.

pollend (approved):

engine/src/main/java/org/terasology/math/JomlUtil.java
skaldarnar: I think there is, because `Region3i` was deleted :wink:

skaldarnar (approved):

PR #4477, opened 2021-02-07
build(MTE): allow MTE to configure the logger.

keturn:
This adds slf4j to the runtime of tests and allows MTE to do logback things.

Required for https://github.com/Terasology/ModuleTestingEnvironment/pull/37

I expect this PR is low risk, though it does raise the notion that we should probably package MTE differently.


keturn:
Conflicts? Oh, from the junit upgrade. Right.

skaldarnar (approved):

PR #4474, opened 2021-02-06
feat(JOML): migrate tests

pollend:
This should finish up the rest of the test cases for Terasology.

PR #4478, opened 2021-02-07
feat(JOML): remove legacy typeHandlers

pollend:
This drops all the typehandlers for termath. the tests are migrated over in this PR (https://github.com/MovingBlocks/Terasology/pull/4474)


skaldarnar:
Is it safe to remove them already? Or would this break stuff as long as we still use those types? 

@DarkWeird any opinion on that?

DarkWeird:
> Is it safe to remove them already? Or would this break stuff as long as we still use those types? 
> 
> @DarkWeird any opinion on that?

It will break already serialized data.(old world loading, server-client connection if different versions)

Serialization will use default class serialization for legacy tera math

skaldarnar:
> It will break already serialized data.(old world loading, server-client connection if different versions)

I think this is only true if we would try to load the data as legacy types, right? I think we have new type handlers for JOML that will be able to parse the data into the new types just fine (which should allow to even load old worlds :thinking: )

Anyways, we don't guarantee that our worlds are compatible between releases... :see_no_evil:

skaldarnar (approved):

PR #4492, opened 2021-02-07
feat(JOML): remove usage of JOMLUtil and migrate remaining code

pollend:
I have other PR that address the few dependencies left and this can all be removed in one go after all that.


pollend:
looks like setSize has a problem: https://github.com/MovingBlocks/joml-ext/compare/bugfix/setSize?expand=1

jdrueckert (approved):

PR #4493, opened 2021-02-08
build: add Allure reports

keturn:
This makes it so you can use [Allure](http://allure.qatools.ru/) to look at your local test results.

Just something to try out.

It reads the results of JUnit5 output generated by the `test` tasks.

Running `gradle :allureMultiServe` starts allure's web server and opens a browser window pointing to it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ blacklist.json
 terasology.jfr
 local.properties
 *.hprof
+
+# Allure error dumps
+ajcore.*.txt
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,4 @@
-// Copyright 2020 The Terasology Foundation
+// Copyright 2021 The Terasology Foundation
 // SPDX-License-Identifier: Apache-2.0
 
 // Dependencies needed for what our Gradle scripts themselves use. It cannot be included via an external Gradle file :-(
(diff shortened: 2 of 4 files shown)


keturn:
Hmm. Has some better aesthetics than the html report that is gradle's default, but I don't see it including the test logs/output anywhere. That's a critical omission if I actually want to use it for debugging instead of only an aggregate overview.

Screenshot_2021-02-07 Allure Report

Cervator:
Cool find at least. I've wondered if there were nicer ways to look at the results locally vs in Jenkins

DarkWeird:
@Cervator 
Jenkins have plugin for showing allure reports

I is looked same..

@keturn 
Allure have some other features, which not usable for unit tests.
Like steps, and attachment.
I think to using allure for unit test  is overdo.

Gradle's or junit's Html reports shows that you see in allure without atracments,steps,epic and etc too

keturn:
Yeah, Allure is probably overkill. I was wondering if there were any better options for local viewing of test results that I might like better than gradle's current default.

I guess even when I use gradle as the test runner instead of IntelliJ, IntelliJ might still be able to parse its outputs...

DarkWeird:
Intellij can parse gradle's and junit's test results outputs. they have special listeners.

about gradle's reports. you can see it at `buildDir/reports/tests/*`:
image
image
image

DarkWeird:
Or Junit's test results in `buildDir/test-results/*`
it is Junit4's specification. in format XML -  this results used by jenkins for rendering results.

PR #4496, opened 2021-02-08
test: upgrade to JUnit 5.7.1

keturn:
and mockito 3.7.7.

I expect this is low-risk.

### DIFF
--- a/engine-tests/build.gradle
+++ b/engine-tests/build.gradle
@@ -58,7 +58,7 @@ dependencies {
     implementation group: 'org.reflections', name: 'reflections', version: '0.9.10'
 
     // Test lib dependencies
-    implementation("org.junit:junit-bom:5.7.1") {
+    implementation(platform("org.junit:junit-bom:5.7.1")) {
         // junit-bom will set version numbers for the other org.junit dependencies.
     }
     implementation("org.junit.jupiter:junit-jupiter-api")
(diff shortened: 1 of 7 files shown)


engine-tests/build.gradle
```diff
@@ -58,15 +58,16 @@ dependencies {
...
+        // junit-bom will set version numbers for the other org.junit dependencies.
+    }
+    implementation("org.junit.jupiter:junit-jupiter-api")
+    implementation("org.junit.jupiter:junit-jupiter-params")
+    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine")
 
-    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.5.2")
+    implementation("org.mockito:mockito-junit-jupiter:3.7.7")
 
-    implementation group: 'junit', name: 'junit', version: '4.12'
```
keturn, line 69: I did remove this dependency on old JUnit 4 from engine-test. I expect it to fail loudly if that was a bad idea.

(JUnit 4 _is_ still in the terasology-modules deps in case any of the modules haven't converted yet.)

keturn:
oh well this would be a problem:

```

* What went wrong:
Could not determine the dependencies of task ':engine-tests:checkstyleTest'.
> Could not resolve all task dependencies for configuration ':engine-tests:testCompileClasspath'.
   > Could not resolve org.junit:junit-bom:5.7.1.
```

keturn:
running the tests with gradle makes it pretty obvious where the problems are, but the quick check I did with IntelliJ only left me overconfident.

pollend (approved):

PR #4502, opened 2021-02-09
Feature: Extraction LightMerger usage in post-ready step.

DarkWeird:
### Contains

* LightMerger runs after chunk is ready now.
* Simplified ChunkProcessingPipeline.
* Fixed tests
* Faster chunk delivery to ready.. especially for tests! (should not stuck at chunk generation)

### How to test

1. Run tests
2. Run Singleplayer game
2.1. fly on `hspeed`, check how fast chunk loadings.
3. Run Multiplayer
3.1. fly on `hspeed`, check how fast chunk loadings.


skaldarnar:
When I try to start a new game with just Core gameplay it crashes with:

```
20:22:09.333 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
java.lang.IllegalArgumentException: Cannot find field java.util.concurrent.CompletableFuture$AsyncSupply.task
	at org.terasology.utilities.ReflectionUtil.readField(ReflectionUtil.java:474)
	at org.terasology.world.chunks.pipeline.ChunkProcessingPipeline.lambda$unwrappingComporator$0(ChunkProcessingPipeline.java:70)
	at java.base/java.util.concurrent.PriorityBlockingQueue.siftUpUsingComparator(PriorityBlockingQueue.java:377)
	at java.base/java.util.concurrent.PriorityBlockingQueue.offer(PriorityBlockingQueue.java:488)
	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1347)
	at java.base/java.util.concurrent.CompletableFuture.asyncSupplyStage(CompletableFuture.java:1714)
	at java.base/java.util.concurrent.CompletableFuture.supplyAsync(CompletableFuture.java:1931)
	at org.terasology.world.chunks.pipeline.ChunkProcessingPipeline.invokeGeneratorTask(ChunkProcessingPipeline.java:107)
	at org.terasology.world.chunks.localChunkProvider.LocalChunkProvider.createOrLoadChunk(LocalChunkProvider.java:122)
	at org.terasology.world.chunks.localChunkProvider.RelevanceSystem.lambda$addRelevanceEntity$0(RelevanceSystem.java:184)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at java.base/java.util.stream.SortedOps$RefSortingSink.end(SortedOps.java:395)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:485)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
	at org.terasology.world.chunks.localChunkProvider.RelevanceSystem.addRelevanceEntity(RelevanceSystem.java:178)
	at org.terasology.logic.players.PlayerSystem.addRelevanceEntity(PlayerSystem.java:197)
	at org.terasology.logic.players.PlayerSystem.onConnect(PlayerSystem.java:146)
	at org.terasology.logic.players.PlayerSystemMethodAccess.invoke(Unknown Source)
	at org.terasology.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:536)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.sendStandardEvent(EventSystemImpl.java:292)
	at org.terasology.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:283)
	at org.terasology.entitySystem.entity.internal.BaseEntityRef.send(BaseEntityRef.java:205)
	at org.terasology.network.internal.NetworkSystemImpl.connectClient(NetworkSystemImpl.java:882)
	at org.terasology.network.internal.NetworkSystemImpl.joinLocal(NetworkSystemImpl.java:326)
	at org.terasology.engine.modes.loadProcesses.SetupLocalPlayer.step(SetupLocalPlayer.java:45)
	at org.terasology.engine.modes.StateLoading.update(StateLoading.java:237)
	at org.terasology.engine.TerasologyEngine.tick(TerasologyEngine.java:496)
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:459)
	at org.terasology.engine.TerasologyEngine.runMain(TerasologyEngine.java:435)
	at org.terasology.engine.TerasologyEngine.run(TerasologyEngine.java:401)
20:22:09.353 [main] ERROR o.t.engine.modes.StateLoading - Error while loading org.terasology.engine.modes.loadProcesses.AwaitCharacterSpawn@11c40a65
java.lang.IllegalArgumentException: Cannot find field java.util.concurrent.CompletableFuture$AsyncSupply.task
	at org.terasology.utilities.ReflectionUtil.readField(ReflectionUtil.java:474)
	at org.terasology.world.chunks.pipeline.ChunkProcessingPipeline.lambda$unwrappingComporator$0(ChunkProcessingPipeline.java:70)
	at java.base/java.util.concurrent.PriorityBlockingQueue.siftUpUsingComparator(PriorityBlockingQueue.java:377)
	at java.base/java.util.concurrent.PriorityBlockingQueue.offer(PriorityBlockingQueue.java:488)
	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1347)
	at java.base/java.util.concurrent.CompletableFuture.asyncSupplyStage(CompletableFuture.java:1714)
	at java.base/java.util.concurrent.CompletableFuture.supplyAsync(CompletableFuture.java:1931)
	at org.terasology.world.chunks.pipeline.ChunkProcessingPipeline.invokeGeneratorTask(ChunkProcessingPipeline.java:107)
	at org.terasology.world.chunks.localChunkProvider.LocalChunkProvider.createOrLoadChunk(LocalChunkProvider.java:122)
	at org.terasology.world.chunks.localChunkProvider.RelevanceSystem.updateRelevance(RelevanceSystem.java:134)
	at org.terasology.world.chunks.localChunkProvider.RelevanceSystem.update(RelevanceSystem.java:219)
	at org.terasology.engine.modes.loadProcesses.AwaitCharacterSpawn.step(AwaitCharacterSpawn.java:50)
	at org.terasology.engine.modes.StateLoading.update(StateLoading.java:237)
	at org.terasology.engine.TerasologyEngine.tick(TerasologyEngine.java:496)
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:459)
	at org.terasology.engine.TerasologyEngine.runMain(TerasologyEngine.java:435)
	at org.terasology.engine.TerasologyEngine.run(TerasologyEngine.java:401)
```

skaldarnar:
:rocket: The chunks indeed load faster, but ...

- ... it feels like too much is loaded at once, causing a short freeze/frame drop :see_no_evil: 
- ... there is a "shadow" border at the edge of the region of loaded chunks that does not go away. It retracts farther away if I move in that direction, but then just the next line of chunks is just shadowy...
	Terasology-210210232308-1973x1031

DarkWeird:
Freeze happened, 
because final part processing at main thread with event sending takes many time, 
because processed in place, because it is main thread.

Shadowy chunks.
 Seems this is border at non-lod chunks.
 Yeah. Light merging cannot process this chunks, because not enough data for it.
But strange that this chunks not received light data from internal light grnerator...

DarkWeird:
Oh.

VoxelLiquidVWorldSystem and VoxelWorldSystem take too many time for handle onLoadChunk events.

especially now, when chunks delivering faster.

engine/src/main/java/org/terasology/world/chunks/pipeline/ChunkTask.java
```diff
@@ -7,10 +7,6 @@
 import org.terasology.utilities.concurrency.Task;
 import org.terasology.world.chunks.Chunk;
 
-/**
- * @deprecated Use {@link org.terasology.world.chunks.pipeline.stages.ChunkTask} instead
- */
-@Deprecated
```
skaldarnar, line 13: So we reverted the deprecation? Was the other `ChunkTask` recently introduced, or has it been around for some time but was never put to "special" use (justifying this deprecation)?

engine/src/main/java/org/terasology/world/chunks/ChunkLightMergingSystem.java
```diff
@@ -0,0 +1,75 @@
...
+import org.terasology.world.chunks.event.OnChunkLoaded;
+import org.terasology.world.propagation.light.LightMerger;
+
+import java.util.Set;
+
+/**
+ * System which will process light merging after chunk is ready.
+ */
+@RegisterSystem
+public class ChunkLightMergingSystem extends BaseComponentSystem {
```
skaldarnar, line 25: Oh, we don't even hook that up to the magic chunk generation off-thread process, but it is just a regular system :nerd_face: neat!

engine/src/main/java/org/terasology/world/chunks/ChunkLightMergingSystem.java
```diff
@@ -0,0 +1,75 @@
...
+        processNearbyChunks(chunkLoaded.getChunkPos());
+    }
+
+    private void processNearbyChunks(Vector3ic chunkPos) {
+        for (Vector3ic candidateForLightMerging : new BlockRegion(chunkPos).expand(1, 1, 1)) {
+            if (lightMergedChunks.contains(candidateForLightMerging)) {
+                continue;
+            }
+            processCandidateChunk(candidateForLightMerging);
+            lightMergedChunks.add(candidateForLightMerging);
```
skaldarnar, line 47: Need to create a copy of the vector here, because it will be reused in the next iteration of the loop :wink: 
```suggestion
            lightMergedChunks.add(new Vector3i(candidateForLightMerging));
```

engine/src/main/java/org/terasology/world/chunks/ChunkLightMergingSystem.java
```diff
@@ -0,0 +1,75 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.world.chunks;
+
+import com.google.common.collect.Sets;
+import org.joml.Vector3ic;
```
skaldarnar, line 7:
```suggestion
import org.joml.Vector3i;
import org.joml.Vector3ic;
```

engine/src/main/java/org/terasology/world/chunks/ChunkLightMergingSystem.java
```diff
@@ -0,0 +1,75 @@
...
+        BlockRegion around = new BlockRegion(candidateChunkPosition).expand(1, 1, 1);
+        Chunk[] chunks = new Chunk[LightMerger.LOCAL_CHUNKS_ARRAY_LENGTH];
+        int index = 0;
+        for (Vector3ic position : around) {
+            if (!loadedChunkPos.contains(position)) {
+                return;
+            }
+            chunks[index++] = chunkProvider.getChunk(position);
+        }
+        threadManager.submitTask("Chunk Light Merging", () -> new LightMerger().merge(chunks));
```
skaldarnar, line 66: are we sure that `chunks` contains elements at this point? Should we check that `index > 0`? Does it need at least 1 chunk, or 2, or ...?

engine/src/main/java/org/terasology/world/chunks/CoreChunk.java
```diff
@@ -56,12 +56,30 @@
      */
     Block getBlock(int x, int y, int z);
 
+    /**
+     * Returns block's id at given position relative to the chunk.
+     *
+     * @param pos Position of the block relative to corner of the chunk
+     * @return Block at given position
+     */
+    short getBlockId(Vector3ic pos);
```
skaldarnar, line 65: Should we directly add a default implementation unwrapping the vector here?

```suggestion
    default short getBlockId(Vector3ic pos) {
        return getBlockId(pos.x(), pos.y(), pos.z());
    }
```

engine/src/main/java/org/terasology/world/chunks/LodChunk.java
```diff
@@ -101,6 +101,16 @@ public Block getBlock(int x, int y, int z) {
         throw new UnsupportedOperationException(UNSUPPORTED_MESSAGE);
     }
 
+    @Override
+    public short getBlockId(Vector3ic pos) {
+        throw new UnsupportedOperationException(UNSUPPORTED_MESSAGE);
+    }
```
skaldarnar, line 107: Not needed if we use a default implementation.

```suggestion
```

engine/src/main/java/org/terasology/world/chunks/internal/ChunkImpl.java
```diff
@@ -142,6 +142,16 @@ public final Block getBlock(int x, int y, int z) {
         return blockManager.getBlock(id);
     }
 
+    @Override
+    public short getBlockId(Vector3ic pos) {
+        return getBlockId(pos.x(),pos.y(),pos.z());
+    }
```
skaldarnar, line 148: Not needed if we use a default implementation.

```suggestion
```

skaldarnar (changes requested):
I'm not really the best reviewer for this, I guess (peeking at @4Denthusiast :roll_eyes: ), but I found at least one little bug with iterating over `BlockRegion`. 
Please have a look at the comments :nerd_face:

engine/src/main/java/org/terasology/world/chunks/ChunkLightMergingSystem.java
DarkWeird: Yeah. If any chunk is unavailable - exit in loop before

jdrueckert:
Tried this out and the chunks seemed to load very slowly... it took quite a while and then a bunch of them popped up... usually it felt like they pop up one by one :thinking:

skaldarnar:
I'm moving this to the next milestone now (@DarkWeird I hope that's okay, I'd like to get a new Omega build out, and the rest is working reasonable stable...).

DarkWeird:
> Tried this out and the chunks seemed to load very slowly... it took quite a while and then a bunch of them popped up... usually it felt like they pop up one by one thinking

every chunk ready is faster, because lightmerger(a huge time eater) take a lot of time. (also internal chunkprocessing sorting was eliminated and compilcated wait logic was removed too)
then it is only feels. :)
and yeah it looks like load bunch... because RelativesSystem request only bunches of chunks in one tick.
 then this chunks loads in executor concurrently.. 
 then bunch of chunks(all or part of chunks) ready onetime and became ready on main thread.

pollend:
@DarkWeird testing this a but sometimes restarting the world makes the whole world black.

DarkWeird:
Restarting - world purge?

pollend:
something isn't quite working.

image

DarkWeird:
> something isn't quite working. I'm also seeing some very strange lag time between chunks start showing up. 
> 
> image
> 

Lag is physics's onChunkLoading event handler.

Very strange about black.. maybe you needs to update your reflections cache?

pollend:
Some of the chunks are loading in this way and the other half are fine.

DarkWeird:
> Some of the chunks are loading in this way and the other half are fine.

No. It is not about some of the chunks.
It bunch of them at once. (Iterate 100+ chunks iterate every chunk per block)
I try load them with limits.. it was worster

skaldarnar:
The lag spikes don't seem to be that big anymore (at least I did not notice them that much) but you can still end up with **black/unlit areas**.

image

This happened after I've waited for the LOD chunks to generate, and then moved into some direction...

DarkWeird:
> The lag spikes don't seem to be that big anymore (at least I did not notice them that much) but you can still end up with **black/unlit areas**.
> 
> image
> 
> This happened after I've waited for the LOD chunks to generate, and then moved into some direction...

Just wait. Idk why this works so bad.

ChunkLightMergingSystem checks around chunks on every chunk loading for possible to start light merging...
Maybe because working threads have the Lowest priority....

engine/src/main/java/org/terasology/world/chunks/ChunkLightMergingSystem.java
```diff
@@ -0,0 +1,80 @@
...
+        BlockRegion around = new BlockRegion(candidateChunkPosition).expand(1, 1, 1);
+        Chunk[] chunks = new Chunk[LightMerger.LOCAL_CHUNKS_ARRAY_LENGTH];
+        int index = 0;
+        for (Vector3ic position : around) {
+            if (!loadedChunkPos.contains(position)) {
+                return false;
+            }
+            chunks[index++] = chunkProvider.getChunk(position);
+        }
+        threadManager.submitTask("Chunk Light Merging", () -> new LightMerger().merge(chunks));
```
4Denthusiast, line 70: This doesn't look thread safe. It's changing the light values for a chunk that's already accessible elsewhere, so another system could be updating the lighting at the same time. Also, it looks like you don't mark the chunk dirty anywhere after light merging happens, so it may never be re-meshed (depending on the order of operations).

4Denthusiast (commented):
The changes to the world collision systems seems pretty irrelevant, and it's annoying to have that mixed into the same PR. I can't find any obvious problems that would cause the mostly unlit chunks shown in the screenshots. Internal light merging is clearly still happening, otherwise there wouldn't be that light at the bottom of the chunks, but somehow it's starting off dark, rather than starting off lit at the top like it should be.

I haven't had a look at everything yet. In particular, I haven't read through all of the changes to `ChunkProcessingPipeline`. @DarkWeird Could you summarise what those changes are meant to do (except directly removing the light merging step)? It looks like a lot has been removed. Does that reduce the functionality of it? If it removes the ability to have steps that require multiple chunks at once, I would expect it could be simplified much more than this.

engine/src/main/java/org/terasology/world/chunks/ChunkLightMergingSystem.java
DarkWeird: Yeah it not thread safe, but it costs low price if error happens there. 
About dirting chunks. It  marks implicitly.(somewhere during merging)

DarkWeird:
@4Denthusiast 
1. Light merging as separate system which works at chunks's  onLoaded event
2. Simplify pipeline with saves interface..(almost)
3. Changes to physics system because chunks began incoming to it too fast (freezes main thread, because needs to process many chunks at main thread):
3.1. 
4. You can load only one chunk if you require instead 27(3x3x3) chunks when needs only one.

DarkWeird:
@4Denthusiast 

About unlit.
I don't know how internalLightProcessor works exactly.
I don't change it.
Mostly lightning data at early stage receives from world generator or internalLightProcessor.

Mostly lightning data at all - LightMerger.

Unlit border happens because lightmerger is not processed those chunks. But it cannot process them, because not enough chunks (next chunks is lod chunk already)

pollend:
> @4Denthusiast
> 
> About unlit.
> I don't know how internalLightProcessor works exactly.
> I don't change it.
> Mostly lightning data at early stage receives from world generator or internalLightProcessor.
> 
> Mostly lightning data at all - LightMerger.
> 
> Unlit border happens because lightmerger is not processed those chunks. But it cannot process them, because not enough chunks (next chunks is lod chunk already)

What are the main issues that make lighting inefficient. just shuffling things around does not resolve the problem does it :?.

DarkWeird:
@pollend 
Seems i saw old field in protobuf, which used for light data..

Hmm. I suspect that internalLightProcesdor works  incorrect.

Also.. provide facet data to game can helps make lightning better... Quality change

pollend:
> @pollend
> Seems i saw old field in protobuf, which used for light data..
> 
> Hmm. I suspect that internalLightProcesdor works incorrect.
> 
> Also.. provide facet data to game can helps make lightning better... Quality change

one thing I would like is to have a dynamic light facet for sunlight and just move the light column up the chain until it hits an unloaded chunk and then move the up the next visible block. so we don't have areas underground that are lit because a chunk was momentarily exposed to the sky.

4Denthusiast:
> What are the main issues that make lighting inefficient. just shuffling things around does not resolve the problem does it :?.

The issue that this PR fixes is that you need to load all the chunks around a chunk before it can become visible. That's sort of an efficiency thing in that it reduces the amount of woth necessary before a given amount of chunks become visible.

4Denthusiast:
I'm not happy with just merging this with the concurrency issues left in. Although I agree that the issues would be mostly minor, and probably rare, they would also likely be really confusing. Others who have actually tested this PR have commented that it causes the game to have noticable lag spikes, and fixing the concurrency issue would make that even worse. Possibly that could be improved by improving the efficiency of the lighting calculations (which Michael thinks is very possible), but I think overall this change may just not be worth making, so I suggest just closing the PR unmerged.

PR #4512, opened 2021-02-13
chore(behavior,logic): minor code cleanup

skaldarnar:
Some minor code cleanup based on the boy scout rule to leave the code in a better state than before :nerd_face: Mainly reduced log spam with some additional changes to not catch runtime exceptions.

### DIFF
--- a/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
+++ b/engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
@@ -1,18 +1,5 @@
-/*
- * Copyright 2017 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/terasology/logic/behavior/BehaviorSystem.java
```diff
@@ -75,15 +60,14 @@
...
-            try {
-                Optional<BehaviorTree> asset = assetManager.getAsset(uri, BehaviorTree.class);
-                asset.ifPresent(behaviorTree -> trees.add(behaviorTree));
-            } catch (RuntimeException e) {
-                logger.info("Failed to load behavior tree asset {}.", uri, e);
+            Optional<BehaviorTree> asset = assetManager.getAsset(uri, BehaviorTree.class);
+            if (asset.isPresent()) {
+                trees.add(asset.get());
+            } else {
+                logger.warn("Failed to load behavior tree asset {}.", uri);
```
pollend, line 68: avoiding the exception should help with performance.

pollend (approved):

PR #3655, opened 2019-04-04
Organized server-side and client-side debug text with Compass Nav (debug)

Haxsen:
### Contains

Fix for #1942

### How to test

Press F3 in-game.

Issue #1942, opened 2015-09-20
Part of debug text originates from server?


Cervator:
I just noticed that seemingly the memory stats for the headless server rather than the connected client shows in the F3 debug stats. Showed 4 GB max memory which is what the server was launched with, but not what my client was running.

It makes sense that *some* of the debug text matches the server, like your coordinates. Other stuff like the timings surely match the client. We might want to take a closer look there :-)

Cervator:
On a not-really-related note but I'm not sure yet if it is just me and maybe it'll be encountered while testing this. I found that on some logins on a headless server the camera would be off vs my monkey head. 

First I was looking at myself from above/behind my head so looking down meant I was looking *through* my head. Then on another login I seemingly disconnected entirely from my avatar, which stayed where I was when I quit / was disconnected as the server was stopped abruptly at that point. Another attempt had no weirdness at all and then I was behind my head again.

![terasology-150920164210-1152x720](https://cloud.githubusercontent.com/assets/1063833/9982939/f6e6ae86-5fba-11e5-84bb-c97d24fc0773.jpg)

AWildBeard:
@Cervator I've been playing with this for the past few days off and on. And I don't think that it's a problem with the game reading the memory from the server, I think it is a problem with 

` gradlew game -D-XmxWHATEVER `

not passing the values properly, **or them getting dropped somewhere farther down the line**...
What I've tried (RESULT answer is from the in-game F3 menu)

As a control, I just ran 

- `./gradlew game`

> RESULT: max memory: 3559.5

Then for the actual tests:

- `./gradlew game -Xms512m -Xmx1024m`           

>  RESULT: max memory: 3559.5

- `./gradlew game -D-Xmx1024m`

> RESULT: max memory: 3559.5


- `./gradlew game -D-Xmx6144m`

> RESULT: max memory: 3559.5

    ...
    More of the same...

Log output: (cat Terasology-init.log | grep Max.)
...
[main] INFO  o.terasology.engine.TerasologyEngine - Max. Memory: 3559 MiB
...


Here's why I think this:
I made a little test file that uses the exact same static method that both the "TerasologyEngine" and "DebugOverlay" is using; (Forgive the spacing, github apparently hates me :P)

`public class test1 {
	public static void main(String[] args) {
		System.out.println(Runtime.getRuntime().maxMemory() / 1048576.0);
      }
}`

Executed this file (After compile of course)

- java test1

... And you guessed it :D .. The output is -->

[...]

PR #4513, opened 2021-02-13
feat: improve debug overlay with compass and additional heaps (MovingBlocks/Terasology#1942)

pollend:
I moved in Orientation from CommonWorld so will have to address the duplication.

image

Reference PR: https://github.com/MovingBlocks/Terasology/pull/3655
Reference Issue: https://github.com/MovingBlocks/Terasology/issues/1942

Co-authored-by: Hassan Ali <33906649+Haxsen@users.noreply.github.com>


4Denthusiast:
The heap space information seems like something that would be more appropriate as one of the things that you can toggle through (F4 I think) rather than being in the main debug information area.

pollend:
> The heap space information seems like something that would be more appropriate as one of the things that you can toggle through (F4 I think) rather than being in the main debug information area.

you have a pretty good point. umm what should we call the view?

jdrueckert (commented):
What do these "Eden Space", "Old Gen", "Survivor Space" hep allocation variants mean?
And is our x-axis the wrong way round? :thinking: seems like if I walk east, x decreases, while if I walk west it increases... that somehow seems the wrong way round...

pollend:
@jdrueckert https://stackoverflow.com/questions/2129044/java-heap-terminology-young-old-and-permanent-generations

umm, what do you think is wrong with the x axis?

pollend:
> What do these "Eden Space", "Old Gen", "Survivor Space" hep allocation variants mean?
> And is our x-axis the wrong way round? thinking seems like if I walk east, x decreases, while if I walk west it increases... that somehow seems the wrong way round...

just realized we are using a right handed coordinate system so. 

+x is left +y is up and z+ is forward. 

image

jdrueckert:
> just realized we are using a right handed coordinate system so.
> 
> +x is left +y is up and z+ is forward.

yes this is exactly what caught my eye. but I guess that's something quite deeply rooted then?
and thanks for the link about the heap terminology, that helped 👍

jdrueckert (approved):

PR #4515, opened 2021-02-14
feat: add prototype render graph overlay

pollend:
Prototype for a screen to display render graph nodes. I kind of blew up the other PR but built these changes from a stash so this should be the same. there is a lot of space for improvement. basically just a list with visited nodes and their children. 

Extras > Developer Tools > Render Graph

image
 

Refrence PR: https://github.com/MovingBlocks/Terasology/pull/3351

Co-authored-by: Mike Kienenberger <mkienenb@gmail.com>


mkienenb:
I'm going to pass on reviewing this.   It was out of my area of expertise in the first place, and I barely remember what I was even working on after this much time has passed.

pollend:
relevant discussion from discord

```
Cervator [UTC-5]02/27/2021
right, which one was it?
Michael P [UTC - 7]02/27/2021
https://github.com/MovingBlocks/Terasology/pull/4515
GitHub
feat: add prototype render graph overlay by pollend · Pull Request ...
Prototype for a screen to display render graph nodes. I kind of blew up the other PR but built these changes from a stash so this should be the same. there is a lot of space for improvement. basica...

eeeh needs polish
DarkWeird02/27/2021
tree like structure is ugly.
graph like ... i cannot place nodes properly

Michael P [UTC - 7]02/27/2021
yea

Cervator [UTC-5]02/27/2021
so should we close the old one, link to it for inspiration, and hope for some nice polish to arrive sometime?

Michael P [UTC - 7]02/27/2021
umm was thinking about closing both

DarkWeird02/27/2021
because our opengl too old and needs to update all?

Michael P [UTC - 7]02/27/2021
its a mess of old a newer features of opengl might have unintended effects

Cervator [UTC-5]02/27/2021
would it be worth at least merging what we have since it does visualize it, even if poorly, then improve it when able? Rather than just throw it all out

Michael P [UTC - 7]02/27/2021
yea we could do just that @Cervator [UTC-5]
its in a menu off to the side

Cervator [UTC-5]02/27/2021
alrighty, lets work toward that then. Updating the PR from base and will try it out

Skaldarnar [UTC+1]02/27/2021
question is if anybody is using it in the current state. if not, I'd rather have less code in the repo we have to maintain...
keeping a visible reference to the old PR(s) in some issue on the matter to not forget what we did there, in case somebody has time to grab this up in the future and make it really usable...

Cervator [UTC-5]02/27/2021
that's fine too, we can add it to the salvage list. It isn't merged yet so nobody can use it, it is new functionality :sweat_smile: question is if we often wonder what the state of the DAG is?

4D enthusiast02/27/2021
Being able to see a visual summary of the render graph is certainly handy. There are a few versions online I think. I don't know how up-to-date those are, and it would be nice to have the documentation generated directly from the code rather than relying on people updating external documentation.

```

PR #4516, opened 2021-02-14
feat(block): Add 'Blocks' utility class to convert world to block coordinates

skaldarnar:
- [x] introduce `Blocks` utility class
- [x] add tests to ensure correctness

### DIFF
--- a/engine-tests/src/test/java/org/terasology/world/block/BlocksTest.java
+++ b/engine-tests/src/test/java/org/terasology/world/block/BlocksTest.java
@@ -5,16 +5,33 @@ package org.terasology.world.block;
 
 import org.joml.Vector3f;
 import org.joml.Vector3i;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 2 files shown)


engine-tests/src/test/java/org/terasology/world/block/BlocksTest.java
```diff
@@ -0,0 +1,20 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class BlocksTest {
+
+    @Test
+    public void testToBlockPos() {
+        assertEquals(Blocks.toBlockPos(new Vector3f(.2f, .2f, .2f), new Vector3i()), new Vector3i(0, 0, 0));
+        assertEquals(Blocks.toBlockPos(new Vector3f(-.2f, -.2f, -.2f), new Vector3i()), new Vector3i(0, 0, 0));
+        assertEquals(Blocks.toBlockPos(new Vector3f(.56f, 0, 0), new Vector3i()), new Vector3i(1, 0, 0));
```
jdrueckert, line 18: if we have this, should we then also have a test for Vector3f(1.44f, 0, 0) equalling Vector3i(1, 0, 0)?

jdrueckert (approved):

PR #4521, opened 2021-02-16
chore: remove usage of fixed function pipline methods

pollend:
this removes usage of the fixed function pipeline methods from Chunk. the chunk only renders its own local position but maybe there are ways to address the transforms. glMatrixMod/glLoadIdentity/glpushMatrix etc . some of the active state stuff is called for every chunk draw so were adding a lot of additional state changes to chunk rendering. some of this is pretty limited to opengl so maybe moving these to utilities of some kind. I would imagine vulkan spec is a lot different but I think opengl is can be emulated from vulkan. some things to consider for more work in this regards.  

https://www.khronos.org/opengl/wiki/Legacy_OpenGL

PR: https://github.com/Terasology/CoreRendering/pull/35


DarkWeird:
This shaders should be in CoreRendering... isn't it?

Yeah vulkan not use fttransform. GLSL3-4 too.

pollend:
that's what I thought to @DarkWeird but apparently it lives int he engine instead :?.

DarkWeird:
@pollend 
Someone forgot to move it. seems.

pollend:
> @pollend
> Someone forgot to move it. seems.

lets start here and then migrate the shaders later. I think I'll get rid of the fttransform stuff first then slowly replace things as I go.

DarkWeird (approved):

PR #4522, opened 2021-02-17
build: move more implementation details about gestalt out of terasology-module

keturn:
Another in this series of clean-up work for terasology-module.kts. No change in functionality expected, these changes are made in hopes that putting things in named methods makes the code less tangled and more readable.


pollend:
@keturn nice, beautiful gradle code :D

pollend (approved):

PR #4524, opened 2021-02-19
fix(rendering): clipping plane on LoD 0

jdrueckert:
Noticed a black something in the center of my screen after setting LoD from 10 to 0. Seemed to be on the "back plane" as it disappeared behind loaded chunks. When setting LoD to 1, it didn't show anymore.

@PS-Soundwave assumed that the far clipping plane might be pulled in too far on lowest LoD and suggested this quickfix. Turned out fine, although there seems to be a larger issue, that @PS-Soundwave plans to write up (please also link it here) and investigate :+1: 

Co-authored-by: Soundwave <PS-Soundwave@users.noreply.github.com>


PS-Soundwave (approved):
Good to merge. I'll link the follow-on issue/PR when it's written.

PR #4528, opened 2021-02-21
chore!: limit API for BlockComponent (1/2)

pollend:
the basic gist is to reduce the access for BlockComponent. the state of this component is managed BlockEntitySystem and I would like to avoid having this get modified by another system. its really difficult to track down a bug with this component if everyone can potentially write something to BlockComponent. Its possible to end up in a state where one system modified the location and another system reads out that location. The only way to change this would be to do this through a the WorldProvider. It might be possible to have the location change listen for the component change and update the world state to match the BlockComponent but that is something to consider in the future.


engine/src/main/java/org/terasology/world/block/BlockComponent.java
```diff
@@ -12,9 +12,9 @@
  */
 public final class BlockComponent implements Component {
     @Replicate
-    Vector3i position = new Vector3i();
+    protected Vector3i position = new Vector3i();
     @Replicate
-    public Block block;
+    public Block block; // use getter
```
skaldarnar, line 17: This currently has 69 usages in Omega, with two write accesses, one not coming from this class:

image

How are we planning to do these follow ups?

Before merging this we should get the write access sorted out so that we can eventually set this field non-public.

skaldarnar:
```suggestion
    public Block block;
```

engine/src/main/java/org/terasology/world/block/BlockComponent.java
```diff
@@ -12,9 +12,9 @@
  */
 public final class BlockComponent implements Component {
     @Replicate
-    Vector3i position = new Vector3i();
+    protected Vector3i position = new Vector3i();
     @Replicate
```
skaldarnar, line 16:
```suggestion
    /**
     * @deprecated use {{@link #getBlock()}} instead.
     */
    @Deprecated
    @Replicate
```

engine/src/main/java/org/terasology/world/block/BlockComponent.java
```diff
@@ -39,13 +47,11 @@ public Vector3ic getPosition() {
      *
      * @param dest will hold the result
      * @return dest
+     * @deprecated  use {@link #getPosition()}
      */
+    @Deprecated
     public Vector3i getPosition(Vector3i dest) {
```
skaldarnar, line 53: This is currently used in 130 places... :roll_eyes: 

image

skaldarnar:
I just learned that IntelliJ has **refactor this > encapsulate fields** which can do exactly this replacement in a few clicks :astonished: Amazing! :nerd_face: 

image

engine/src/main/java/org/terasology/world/block/BlockComponent.java
skaldarnar: We could potentially replace the write access with the typical thing you do with immutable types: create a new instance and use that instead:

In `EntityAwareWorldProvider.java`
```java
BlockComponent newBlockComponent = new BlockComponent(type, oldBlockComponent.getPosition());
blockEntity.saveComponent(newBlockComponent);
```

pollend:
are we going to create a pile of PR and just correct this in one go @skaldarnar.

skaldarnar:
> are we going to create a pile of PR and just correct this in one go @skaldarnar.

I think we should do that for making the `block` field unaccessible, yes. Could also push this refactoring directly to the modules, I trust IntelliJ with that 🤓

engine/src/main/java/org/terasology/world/block/BlockComponent.java
```diff
@@ -12,7 +12,7 @@
  */
 public final class BlockComponent implements Component {
     @Replicate
-    Vector3i position = new Vector3i();
+    protected Vector3i position = new Vector3i();
     @Replicate
     public Block block;
```
skaldarnar, line 17: I'm not changing this yet to not break Omega. We can easily avoid breaking commit here by making the getter available first, then migrating all modules, and only then restricting the visibility.

skaldarnar (approved):

PR #4529, opened 2021-02-21
fix: use system property instead of ENV to determine user home

skaldarnar:
@Cervator reported on Discord that "it may not work in some cases on Windows". 

The problematic part is in the new `Jvm` class, where
```java
String userHome = System.getenv("HOME"); 
```
eventually leads to an NPE "as my Win10 simply doesn't have an env var by that name". 

@Cervator suggested that Windows uses a different ENV variable `HOMEPATH` for this, 
and I remembered that there is also a [system property] `user.home` we could use here.

[system property]: https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#getProperties()

### DIFF
--- a/engine/src/main/java/org/terasology/utilities/Jvm.java
+++ b/engine/src/main/java/org/terasology/utilities/Jvm.java
@@ -21,7 +21,7 @@ public final class Jvm {
     public static void logClasspath(Logger aLogger) {
         String interestingGroup = "org.terasology";
         String projectRoot = PathManager.getInstance().getInstallPath().toString();
-        String userHome = System.getenv("HOME");
+        String userHome = System.getProperty("user.home");
         String indent = "  ";
         int elidedCount = 0;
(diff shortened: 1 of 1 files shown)


skaldarnar:
@keturn was there any specific reason you did not use the property that I'm missing with this fix?

DarkWeird (approved):

keturn:
> was there any specific reason you did not use the property that I'm missing with this fix?

Not particularly. I think when I was looking up how to do this I ran in to some posts about how [that property doesn't work cross-platform either](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4787931), but it looks like that was fixed in Java 8.

I knew the environment variable probably wouldn't be right in Windows, but I didn't think about it very hard and I was hoping it would just not replace anything for that, I didn't consider it might crash! Thanks for the fix.

PR #4531, opened 2021-02-21
fix: save world gen facet information in manifest

pollend:
I saved the world config in the game manifest and read that out when the world entity is added to the world. Ideally we would pass it to the world entity but there are quite a few steps in between GameManifestProivder and when it gets fed to worldentity. this is kind of hack to get this to work correctly. Now if you save a world then reload the world the world generation setting should come over on the reload. in the old case when reloading the world, the game will default back to the default world config.


jdrueckert (dismissed):
Tests out fine :+1:

engine/src/main/java/org/terasology/persistence/internal/ReadWriteStorageManager.java
```diff
@@ -369,6 +371,14 @@ private void addGameManifestToSaveTransaction(SaveTransactionBuilder saveTransac
         for (WorldInfo worldInfo: worlds) {
             gameManifest.addWorld(worldInfo);
         }
+
+        WorldGenerator worldGenerator = CoreRegistry.get(WorldGenerator.class);
+        if (worldGenerator != null) {
+            WorldConfigurator worldConfigurator = worldGenerator.getConfigurator();
+            Map<String, Component> params = worldConfigurator.getProperties();
+            gameManifest.setModuleConfigs(worldGenerator.getUri(), params);
```
keturn, line 379: ❔  so this is doing a thing to make sure that the WorldConfigurator properties are current in the GameManifest when it is saved.

but how does GameManifest know how to save its new `moduleConfigs` field? I didn't see a change to a GameManifestSerializer or anything like that.

pollend: its written to the worldEntityRef.

engine/src/main/java/org/terasology/engine/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -92,13 +92,14 @@ public boolean step() {
             // Also set all the components to the world entity.
             WorldConfigurator worldConfigurator = worldGenerator.getConfigurator();
             Map<String, Component> params = worldConfigurator.getProperties();
+
             for (Map.Entry<String, Component> entry : params.entrySet()) {
                 Class<? extends Component> clazz = entry.getValue().getClass();
-                Component comp = config.getModuleConfig(generatorUri, entry.getKey(), clazz);
+                Component comp = gameManifest.getModuleConfig(generatorUri, entry.getKey(), clazz);
```
keturn, line 98: The thing that bothered me about this (and drove me to try #4584 to see if I was understanding things correctly) is that it seems like this removes the only use of `Config` in here.

and I don't see anything in this PR that takes values out of Config and puts them in GameManifest.

is there someplace else that is moving values from Config to WorldConfigurator before they are copied from WorldConfigurator to GameManifest?

engine/src/main/java/org/terasology/persistence/internal/ReadWriteStorageManager.java
```diff
@@ -369,6 +371,14 @@ private void addGameManifestToSaveTransaction(SaveTransactionBuilder saveTransac
         for (WorldInfo worldInfo: worlds) {
             gameManifest.addWorld(worldInfo);
         }
+
+        WorldGenerator worldGenerator = CoreRegistry.get(WorldGenerator.class);
```
keturn, line 375: wait, a GameManifest has worlds. the line right above this one just made sure of that. so why do we need to pull _another_ WorldGenerator out of CoreRegistry and add it to _different_ fields?

skaldarnar:
@pollend could you try to resolve the conflicts here, please? I think we should try to get this in now that proposals have been submitted and there's some time for other things again ^^

pollend:
@skaldarnar there was no conflicts when I merged it locally :confused:

jdrueckert:
> @skaldarnar there was no conflicts when I merged it locally thinking

yeah GitHub sometimes has issues for some reasons :shrug: 

@pollend there's still open questions from @keturn that were not responded to yet :wink:

pollend:
> > @skaldarnar there was no conflicts when I merged it locally thinking
> 
> yeah GitHub sometimes has issues for some reasons shrug
> 
> @pollend there's still open questions from @keturn that were not responded to yet wink

umm wouldn't quite know how to approach it since its in a complete other place in the codebase.   

I like some of the changes here: https://github.com/MovingBlocks/Terasology/pull/4584 do you want to factor them in here @keturn .

keturn:
> I like some of the changes here: #4584 do you want to factor them in here

We could just merge that one on its own, but there's at least one open question there too: https://github.com/MovingBlocks/Terasology/pull/4584#discussion_r610365018

jdrueckert:
[Save world config across reloads](https://trello.com/c/ur6Lv508/67-save-world-config-across-reloads)

engine/src/main/java/org/terasology/persistence/internal/ReadWriteStorageManager.java
keturn: because though WorldInfo has a field named `worldGenerator`, that is only the URI of the world generator class, not an initialized WorldGenerator instance we can read configuration information from.

(Unless that URI contains all the generator's configuration encoded as URL parameters, but I am pretty sure that is not the case.)

So we still need to add this world configuration information to the manifest. I think it would be good to have it integrated with the worlds field that is already there somehow. Would it be appropriate to expand the WorldInfo class to have the config information? Is there some other data class that serves that need better? I feel like there are a dozen classes with `World` in the name and I haven't sorted them out yet.

pollend: there is an absurd amount of unecessary juggling here that is a bit maddening to look at.

jdrueckert (approved):

PR #4533, opened 2021-02-21
chore: Updated default ports, server for logstash

agent-q1:
Doesn't include changes to collector ports

### Contains

Updates default values of logstash server and ports.

### How to test

Make sure to delete your config.cfg file before testing as it may have cached values from previous runs of the game.
From the main screen , first click on the "Extras" button and then the "Metrics Menu" button.   You should now see a "send error logs to server" checkbox on the bottom right. Click on it and a popup screen should open. The textboxes should already be filled in with the values **Address : telemetry.terasology.com** and **port : 9000.**


jdrueckert (approved):

pollend (approved):

Cervator (approved):
Tests out! I was able to enable and send my logs to the telemetry server _and_ somehow find it in the Discover section still not really knowing well how to use that piece 😅

Issue #1450, opened 2014-12-21
Ensure that engine ComponentSystems are initialized first


msteiger:
I want to use setup a `DelayManager` timer in a ComponentSystem's `initialize()` method. This however fails, because all module systems are initialized before engine systems.

The engine module is registered first in `ModuleManager`, but `TableModuleRegistry` does not preserve insertion order. `ModuleEnvironment` enumerates modules ordered by dependencies, but since engine is not part of the chain, it can appear last.

I see two options to solve this:
- Somehow preserve insertion order so that engine will always be enumerated first.
- Add the engine module as a dependency to all modules. This makes more sense imho, because they really depend on it.

PR #4540, opened 2021-02-22
refactor(ModuleManager): factor initiliaztion code to multiple methods

keturn:
Refactoring in anticipation of changes for gestalt v7.

I think these are all strict refactors, minimal testing required.

Open to feedback about where these methods should be ordered in the class file.

- [x] Merge parent #4539, which I based this on because that moved the ModuleManager class file. The diff on this will be clearer once that is merged.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ModuleManager.java
+++ b/engine/src/main/java/org/terasology/engine/module/ModuleManager.java
@@ -70,6 +70,30 @@ public class ModuleManager {
 
         metadataReader = newMetadataReader();
 
+        Module engineModule = loadEngineModule(classesOnClasspathsToAddToEngine);
+
+        registry = new TableModuleRegistry();
+        registry.add(engineModule);
+
+        loadModulesFromClassPath();
+
+        ModulePathScanner scanner = new ModulePathScanner(new ModuleLoader(metadataReader));
+        scanner.getModuleLoader().setModuleInfoPath(TerasologyConstants.MODULE_INFO_FILENAME);
+        scanner.scan(registry, pathManager.getModulePaths());
+
+        DependencyInfo engineDep = new DependencyInfo();
+        engineDep.setId(engineModule.getId());
+        engineDep.setMinVersion(engineModule.getVersion());
+        engineDep.setMaxVersion(engineModule.getVersion().getNextPatchVersion());
+
+        registry.stream().filter(mod -> mod != engineModule).forEach(mod -> mod.getMetadata().getDependencies().add(engineDep));
+
+        setupSandbox();
+        loadEnvironment(Sets.newHashSet(engineModule), true);
+        installManager = new ModuleInstallManager(this, masterServerAddress);
+    }
+
+    private Module loadEngineModule(List<Class<?>> classesOnClasspathsToAddToEngine) {
         Module engineModule;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/engine/module/ModuleManager.java
```diff
@@ -79,16 +79,20 @@ public ModuleManager(String masterServerAddress, List<Class<?>> classesOnClasspa
 
         loadModulesFromApplicationPath(pathManager);
 
+        ensureModulesDependOnEngine(engineModule);
```
keturn, line 82: I wondered why this is important, and found https://github.com/MovingBlocks/Terasology/issues/1450

It's not a worry that the `engine` module wouldn't be loaded without it. It's about ordering: some things run in an order derived from the dependency tree, and we want to make sure engine is at the root of it.

skaldarnar:
```suggestion
        // I wondered why this is important, and found #1450
        // It's not a worry that the engine module wouldn't be loaded without it. 
        // It's about ordering: some things run in an order derived from the dependency 
        // tree, and we want to make sure engine is at the root of it.
        ensureModulesDependOnEngine(engineModule);
```

engine/src/main/java/org/terasology/engine/module/ModuleManager.java
```diff
@@ -125,6 +134,18 @@ public ModuleManager(Config config, List<Class<?>> classesOnClasspathsToAddToEng
         this(config.getNetwork().getMasterServer(), classesOnClasspathsToAddToEngine);
     }
 
+    private ModuleMetadataJsonAdapter newMetadataReader() {
+        final ModuleMetadataJsonAdapter metadataJsonAdapter;
+        metadataJsonAdapter = new ModuleMetadataJsonAdapter();
```
skaldarnar, line 139: Why not do this in one line?
```suggestion
        final ModuleMetadataJsonAdapter metadataJsonAdapter = new ModuleMetadataJsonAdapter();
```

engine/src/main/java/org/terasology/engine/module/ModuleManager.java
```diff
@@ -68,13 +68,40 @@ public ModuleManager(String masterServerAddress) {
...
+        loadModulesFromApplicationPath(pathManager);
+
+        ensureModulesDependOnEngine(engineModule);
+
+        setupSandbox();
+        loadEnvironment(Sets.newHashSet(engineModule), true);
+        installManager = new ModuleInstallManager(this, masterServerAddress);
+    }
+
+    private void ensureModulesDependOnEngine(Module engineModule) {
```
skaldarnar, line 89:
```suggestion
    /**
     * I wondered why this is important, and found MovingBlocks/Terasology#1450.
     * It's not a worry that the engine module wouldn't be loaded without it. 
     * It's about ordering: some things run in an order derived from the dependency 
     * tree, and we want to make sure engine is at the root of it.
     */   
    private void ensureModulesDependOnEngine(Module engineModule) {
```

engine/src/main/java/org/terasology/engine/module/ModuleManager.java
skaldarnar: Added this as comment to `ensureModulesDependOnEngine` for later reference.

skaldarnar (changes requested):

skaldarnar (approved):

PR #4543, opened 2021-02-25
feat(ModuleManager): do not load all modules in the classpath by default

keturn:
As we were reminded in #4375, putting Terasology modules on the classpath is usually _not_ what we want to do. Additionally, the [draft of ModuleManager for gestalt v7](https://github.com/MovingBlocks/Terasology/blob/f987e3b57c2db0f3d883c200e64befdb9d58b765/engine/src/main/java/org/terasology/engine/core/module/ModuleManagerImpl.java#L56) doesn't load classpath modules at all.

This PR disables running `loadModulesFromClasspath` _unless_  the environment variable `terasology_load_classpath_modules` is set to `true`.

### How to test

* No expected impact on the shipped game.
* _Might_ impact running the game in dev workspaces. I think it _shouldn't,_ but I could be surprised here.
* Impacts automated tests. Check
  - tests in this repo (mostly `:engine-test:test`, but run all tests while we're at it.)
  - tests in module repos
    - tests using MTEExtension (ModuleTestingEnvironment)
    - non-MTE unit tests

- [ ] fix MTE?


keturn:
This is a bit rough around the edges, but it demonstrates two things so far:

* If we don't load modules from classpath, then `engine-test` must load its module itself. Here we restore that function to engine-test's ModuleManagerFactory (which I removed in #4479 because it was redundant while we were loading all classpath modules).
* MTE tests all break with dependency resolution errors, because currently their module dependencies are given on the classpath.

My proposed plan is to move the `loadModulesFromClasspath` stuff in to MTE, taking it out of ModuleManager entirely (including the environment variable feature toggle), and hopefully that's sufficient for MTE to run.

If MTE figures out a different way to load things later, it can switch over then.

keturn:
I've started a PR for the MTE side of this: https://github.com/Terasology/ModuleTestingEnvironment/pull/41

keturn:
Notes from [#gestalt](https://discord.com/channels/270264625419911192/714313228271747152/814973227776081980):

> **keturn:** Anyone have an inkling of why this addAll(registry.filter(Module::isOnClasspath)) is in ModuleManger.loadEnvironment? As far as I can tell, it's been that way for approximately forever. https://github.com/MovingBlocks/Terasology/blob/6a182d5a5bff47edce612d54788e556f3bcddd50/engine/src/main/java/org/terasology/engine/module/ModuleManager.java#L226-L228
>  **keturn:** well, a bunch of tests fail when I comment out, so that's something. in StorageManagerTest and EntityAwareWorldProviderTest.
> **Immortius:** Classpath modules are necessarily always part of the environment. They would be included for class lookup support I imagine.
> **Immortius:** and they might not be explicitly pulled in as dependencies.

The StorageManagerTest and EntityAwareWorldProvideTest both use [TerasologyTestingEnvironment](https://github.com/MovingBlocks/Terasology/blob/develop/engine-tests/src/main/java/org/terasology/TerasologyTestingEnvironment.java), which is a thing like ModuleTestingEnvironment. That means ModuleManagerFactory's helper that included the "unittest" module in the environment didn't apply.

keturn:
I think this is waiting on the MTE side, that it hasn't hurt anything in real released-game-mode, though I admit I haven't done much in the way of play testing on it.

I was going to move some of this code out of ModuleManager entirely and over to MTE, but that has become Complicated. One reason that it's not just the same to do it in one place as the other is the ordering, I think: When ModuleManager loads classpath modules, it does it _before_ it loads from PathManager's module paths, and that makes a difference in which one is considered the duplicate and "skipped."

Is it time to give up on the "move it to MTE" approach and stick with what I started with here, some sort of configurable setting?

It could even stay an environment variable. Not exactly the idiomatic way to communicate options between Java components, but that way the test runner could set it before it even starts the process, or MTEExtension could set it before it builds the engine, whichever.

That might be a good step that is closest to the current status quo but makes some incremental progress.

keturn:
If that's the plan, then this refactoring of PathManager.updateDirs is tangential. I don't mind keeping it but if I'm not meddling with PathManager.modPaths then it's not related to what we need to finish up here: https://github.com/MovingBlocks/Terasology/pull/4543/commits/fef2727944cd34988c40ce6c3566af1c9c584b16#diff-804c3035013492aac35a62a255b23085a08dc3f28d94a238f2a76ea9f0c3ea69

The part that currently needs cleaning up is in ModuleManager, where `attemptToLoadModule` and `loadClasspathModule` are doing nearly the same thing, but `attempt`'s caller is throwing everything in the classpath at it and just wants to see what sticks, whereas `loadClasspathModule`'s caller has a specific location and depends on a specific module being loaded, so it sets `lenient=false` on its loader and is less forgiving of nulls and exceptions.

I look at those two and think "there must be duplication to remove here," but maybe it's better to not bother in this case. gestalt-v7 will be changing their implementation details anyway. Maybe just fixing their names is good enough?

keturn:
Rewound this one a bit and moved the PathManager refactors to a new PR.
Closed MTE#41, will open a new one that uses the ModuleManager option added here.

keturn:
> It could even stay an environment variable. Not exactly the idiomatic way to communicate options between Java components, but that way the test runner could set it before it even starts the process, or MTEExtension could set it before it builds the engine, whichever.

_Or so you thought!_

Here in Java land, we [don't allow setting environment variables!](https://stackoverflow.com/questions/318239/how-do-i-set-environment-variables-from-java)

um. shucks. thwarted. `System.setProperty` then, for the in-process use?

keturn:
I think this is in a working state now. MTE needs https://github.com/Terasology/ModuleTestingEnvironment/pull/43 or it'll start encountering a lot more dependency resolution failures when this default changes.

There's also a thing in here to help with the TypeAnnotationScanner error: https://github.com/MovingBlocks/Terasology/pull/4543/commits/fa23509b672dc990f0a90b5b0bb82cb1fe8a97bd#diff-65e880602beee8aaa3da789489b79761e7e127439b3fa59d0a6185aa0d940d4f

engine/src/main/java/org/terasology/engine/module/ModuleManager.java
```diff
@@ -166,33 +186,74 @@ void loadModulesFromClassPath() {
                 System.getProperty("java.class.path").split(System.getProperty("path.separator", ":"))
         ).map(Paths::get).collect(Collectors.toList());
```
DarkWeird, line 187: If you want to prepare this to `gestalt v7` you shouldn't to use `Path`  class.
`gestalt v7` don't use `Path` class, because it  not exists in Android.

DarkWeird (approved):
Works, Looks nice.

PR #4546, opened 2021-02-27
feat(logs): improve prefab deserialization logging

jdrueckert:


engine/src/main/java/org/terasology/entitySystem/prefab/internal/PrefabFormat.java
```diff
@@ -50,7 +37,10 @@ public PrefabData load(ResourceUrn resourceUrn, List<AssetDataFile> inputs) thro
         try (BufferedReader reader = new BufferedReader(new InputStreamReader(inputs.get(0).openStream(), Charsets.UTF_8))) {
             EntityData.Prefab prefabData = EntityDataJSONFormat.readPrefab(reader);
             if (prefabData != null) {
-                logger.info("Attempting to deserialize prefab {} with inputs {}", resourceUrn, inputs);
+                if (!prefabData.hasName()) {
+                    prefabData = prefabData.toBuilder().setName(resourceUrn.toString()).build();
+                }
+                logger.info("Deserializing prefab {} with inputs {}", resourceUrn, inputs);
```
keturn, line 43: There's a lot of this in the log at start-up. Is it useful as `info` or may we bump it down to `debug`?

engine/src/main/java/org/terasology/persistence/serializers/PrefabSerializer.java
```diff
@@ -126,6 +114,7 @@ public PrefabData deserialize(EntityData.Prefab prefabData) {
     public PrefabData deserialize(EntityData.Prefab prefabData, List<EntityData.Prefab> deltas) {
         Module context = ModuleContext.getContext();
         PrefabData result = new PrefabData();
+        List<String> unknownComponents = new ArrayList<>();
```
keturn, line 117: issue: this PR adds this but I don't see it used?

engine/src/main/java/org/terasology/persistence/serializers/PrefabSerializer.java
```diff
@@ -167,7 +156,7 @@ private void applyComponentChanges(Module context, EntityData.Prefab prefabData,
                     }
                 }
             } else if (componentData.hasType()) {
-                logger.error("Prefab contains unknown component '{}'", componentData.getType());
+                logger.warn("Prefab '{}' contains unknown component '{}'", prefabData.getName(), componentData.getType());
```
keturn, line 159: question: I see you changed this from `error` to `warn`. Is it not an error to reference unknown components? What would be an example of this? (I see a lot of this sort of thing in the logs, but it doesn't seem to come up in #stabilization.)

keturn (changes requested):

keturn:
I was doing similar things for components yesterday: https://github.com/MovingBlocks/Terasology/pull/4543/commits/46394fd911f6018bb92fa0bb4c767ebd02baec56

engine/src/main/java/org/terasology/entitySystem/prefab/internal/PrefabFormat.java
jdrueckert: good point :+1:

engine/src/main/java/org/terasology/persistence/serializers/PrefabSerializer.java
jdrueckert: nice catch... leftover of a previous attempt

engine/src/main/java/org/terasology/persistence/serializers/PrefabSerializer.java
jdrueckert: I believe it should be a warning, as we mention it but continue with normal execution.
It can be components that no longer exist or components that are not available in the selected set of modules. It could have the effect that some system logic is not executed because event listeners expect a component to be present, but that would be desired behaviour of our modular ECS architecture I think.

keturn: This sounds plausible. I don't know much about how optional dependencies and interactions work in this architecture.

I worry that continuing with normal execution teaches people to ignore warnings and makes it harder to find what's relevant when hunting down problems.

but whether to continue execution is well outside the scope of this PR, and I guess I can agree that if we're willing to continue execution it is more like a "warning" than an "error."

engine/src/main/java/org/terasology/persistence/serializers/PrefabSerializer.java
```diff
@@ -1,34 +1,22 @@
...
 import org.terasology.entitySystem.metadata.ComponentMetadata;
 import org.terasology.entitySystem.prefab.Prefab;
 import org.terasology.entitySystem.prefab.PrefabData;
 import org.terasology.module.Module;
-import org.terasology.engine.module.ModuleContext;
 import org.terasology.persistence.typeHandling.TypeHandlerLibrary;
 import org.terasology.protobuf.EntityData;
+import org.terasology.utilities.Assets;
 
+import java.util.ArrayList;
```
keturn, line 19: nitpick: recommend _Optimize Imports_ under IntelliJ Commit Options, or if you're not using IntelliJ to commit, _Settings/Editor/General/Auto Import/Optimize imports on the fly_.

keturn (approved):

engine/src/main/java/org/terasology/persistence/serializers/PrefabSerializer.java
skaldarnar: I think continuing execution is "usually" what we want to do - we should not hard-crash the game (or drop back to menu) too easily imho. 

I think the current setup does not take into account that some component may come from an optional dependency, and thus does not need to be there in all cases.. It would be nice to error out on expected components being unavailable, but being silent about cases where this is not a big deal... way beyond the scope of this, as said. 

I'm wondering whether there's something about that in here: https://github.com/MovingBlocks/Terasology/pull/3236

PR #4547, opened 2021-02-27
chore(RegisterMods): make sure thread has a name and is set to daemon

keturn:
All our threads should have names to help us during debugging.

Most threads should also have daemon=true so they don't prevent the process from closing if they get stuck.

This fixes those things for loadProcesses.RegisterMods.

Like its predecessor #3953, this PR is something of a straw man, as this is a ridiculous amount of boilerplate code to copy every time we want to queue an asynchronous task. Suggestions encouraged.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/modes/loadProcesses/RegisterMods.java
+++ b/engine/src/main/java/org/terasology/engine/modes/loadProcesses/RegisterMods.java
@@ -1,21 +1,10 @@
-/*
- * Copyright 2013 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// Copyright 2021 The Terasology Foundation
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/engine/modes/loadProcesses/RegisterMods.java
```diff
@@ -87,8 +78,12 @@ public boolean step() {
                 }
 
                 EnvironmentSwitchHandler environmentSwitchHandler = context.get(EnvironmentSwitchHandler.class);
-                applyModulesThread = new Thread(() -> environmentSwitchHandler.handleSwitchToGameEnvironment(context));
-                applyModulesThread.start();
+
+                applyModulesThread = Executors.newSingleThreadExecutor(
```
DarkWeird, line 82: suggestion: Reuse same executor or shutdown executor on done or extends `ThreadManager` to return `Future`

engine/src/main/java/org/terasology/engine/modes/loadProcesses/RegisterMods.java
```diff
@@ -44,7 +35,7 @@
 
     private final Context context;
     private final GameManifest gameManifest;
-    private Thread applyModulesThread;
+    private Future<?> applyModulesThread;
```
DarkWeird, line 38: suggestion: Rename it.

DarkWeird (changes requested):

jdrueckert:
Closing this as the code changes are no longer relevant with the current transition to Reactor.
The need for naming threads is now tracked in https://github.com/MovingBlocks/Terasology/issues/5005.
The aspect of configuring threads as daemons was added as a TODO for the transition to reactor in https://github.com/MovingBlocks/Terasology/issues/4798

PR #4553, opened 2021-03-01
build: move `buildSrc` to `build-logic`

keturn:
As a developer, when I add gestalt to my `libs/` workspace, IntelliJ IDEA collapses with a message about being unable to resolve `gestalt-module`.

I filed this as [IDEA-261518](https://youtrack.jetbrains.com/issue/IDEA-261518) and the recommendation from the gradle folks is to replace `buildSrc` with an included build.

Understanding why involves some gradle history.

Gradle made `buildSrc` a place to put code that's not on the same level as your (multi-) project build. It's the code that needs to build _before_ it figures out how to build the rest of your project.

Then came along composite builds, which let you include independent builds and use their results in your root project.

That has a lot of overlap with what `buildSrc` is for!

As composite builds become more useful, `buildSrc` becomes more of a relic, as it's less widely applicable. So they're having some trouble keeping feature parity between buildSrc and included builds. The plan is _eventually_ to make `buildSrc` just another included build, but it sounds like they expect that to take a while.

The current recommendation is to use an included build, even though it's not quite a drop-in replacement. That's why this PR has some more stuff creeping back in to the root project's `buildscript`: buildSrc used to add those to the environment by default.

For further reading, see https://github.com/gradle/gradle/issues/14713#issuecomment-708456079

### How to test

build stuff.

check out stuff to `libs/`, as described in [wiki:Using Locally Developed Libraries](https://github.com/MovingBlocks/Terasology/wiki/Using-Locally-Developed-Libraries).


keturn:
Some notes from Vampire on [gradle slack](https://gradle-community.slack.com/archives/CAHSN3LDN/p1614633001101000?thread_ts=1614455346.085700&cid=CAHSN3LDN) on why we need to explicitly list our `build-logic` project in the root project's `buildscript` classpath:

> Ah, well, if you use legacy apply that could also be part of the reason.
> If you have buildSrc the result is automatically added to the class path of all build scripts.
> If you use a composite build, then not.
> The plugins blocks and buildscript { dependencies { ... } } blocks are searched for plugins and dependencies in the composite build and only what is used is built and then added to the build scripts where it is used.
> This way for example you don't need to recompile all your build scripts just because one line in buildSrc changed.
> But that means if you use apply, you also have to add a build script dependency

The "legacy apply" he's referring to is here: https://github.com/MovingBlocks/Terasology/blob/e691ffb0875e2e261a9842175b622d38f94167b2/config/gradle/common.gradle#L19

having our `terasology-module` plugin include `common.gradle` groovy file from the root project which applies the `terasology-repositories` plugin is apparently too much for the default buildscript scanner to figure out.

DarkWeird (approved):
Works with TeraNUI lib.
runs `gradlew server`

PR #4557, opened 2021-03-04
chore: upgrade guava to 30.1 from 23.0

keturn:
There was one function that had been deprecated long enough that it was no longer available and we had to switch to its new name.

A few other functions are no longer `@Beta` and we got to remove some "unstable api" tags from their call sites.

Seems to be working fine.

You can see info about how it resolved dependencies for guava in a [build scan](https://scans.gradle.com/s/gmljvt3zrah2i/dependencies?focusedDependency=WzIsMSw0MjEsWzIsMSxbNDIxXV1d&focusedDependencyView=versions&toggled=W1syXSxbMiwxXV0).

### DIFF
--- a/engine-tests/src/test/java/org/terasology/rendering/nui/layers/mainMenu/savedGames/GameProviderTest.java
+++ b/engine-tests/src/test/java/org/terasology/rendering/nui/layers/mainMenu/savedGames/GameProviderTest.java
@@ -1,18 +1,5 @@
-/*
- * Copyright 2018 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


DarkWeird:
There using guava's `Futures`, which has some differences in api:
https://github.com/Terasology/Behaviors/pull/69

skaldarnar (approved):
Should we also update the version in [TeraNUI](https://github.com/MovingBlocks/TeraNUI/search?q=guava)?

keturn:
Might as well, before the next TeraNUI release, but it probably won't impact the Terasology build much. Unless it reveals that TeraNUI has been using some long-deprecated bit of API.

PR #4559, opened 2021-03-05
chore(engine-test): use a temp directory instead of a Shrinkwrap archive

keturn:
allows removing the dependency on Shrinkwrap.

How are there so many of these???

### DIFF
--- a/engine-tests/src/main/java/org/terasology/HeadlessEnvironment.java
+++ b/engine-tests/src/main/java/org/terasology/HeadlessEnvironment.java
@@ -1,23 +1,7 @@
-/*
- * Copyright 2018 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
 package org.terasology;
 
-import org.jboss.shrinkwrap.api.ShrinkWrap;
-import org.jboss.shrinkwrap.api.nio.file.ShrinkWrapFileSystems;
-import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.assets.management.AssetManager;
@@ -104,8 +88,9 @@ import org.terasology.world.time.WorldTime;
 import org.terasology.world.time.WorldTimeImpl;
 
 import java.io.IOException;
-import java.nio.file.FileSystem;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.attribute.FileAttribute;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


engine-tests/src/test/java/org/terasology/persistence/internal/StorageManagerTest.java
```diff
@@ -116,12 +100,12 @@ static void createFolder() throws IOException {
     @BeforeEach
     public void setup() throws Exception {
         super.setup();
-        JavaArchive homeArchive = ShrinkWrap.create(JavaArchive.class);
-        FileSystem vfs = ShrinkWrapFileSystems.newFileSystem(homeArchive);
-        PathManager.getInstance().useOverrideHomePath(temporaryFolder.toPath());
+        Path tempHome = Files.createTempDirectory("terasology-smt", (FileAttribute<?>[]) null);
+        tempHome.toFile().deleteOnExit();
+        PathManager.getInstance().useOverrideHomePath(tempHome);
```
keturn, line 105: this extends TerasologyTestingEnvironment, which already does this in `beforeAll`. But I guess because these are storage tests, it wants to change to a clean directory `beforeEach` instead?

engine-tests/src/test/java/org/terasology/config/flexible/AutoConfigManagerTest.java
```diff
@@ -51,9 +37,9 @@
 
     @BeforeEach
     public void setUp() throws Exception {
-        final JavaArchive homeArchive = ShrinkWrap.create(JavaArchive.class);
-        final FileSystem vfs = ShrinkWrapFileSystems.newFileSystem(homeArchive);
-        PathManager.getInstance().useOverrideHomePath(vfs.getPath(""));
+        Path tempHome = Files.createTempDirectory("terasology-env", (FileAttribute<?>[]) null);
+        tempHome.toFile().deleteOnExit();
+        PathManager.getInstance().useOverrideHomePath(tempHome);
```
keturn, line 42: suggestion: as this is a test class and not some abstract base class, this could use [`@TempDir`](https://junit.org/junit5/docs/current/user-guide/#writing-tests-built-in-extensions-TempDirectory).

engine-tests/src/main/java/org/terasology/TerasologyTestingEnvironment.java
```diff
@@ -71,9 +56,9 @@
 
     @BeforeAll
     public static void setupEnvironment() throws Exception {
-        final JavaArchive homeArchive = ShrinkWrap.create(JavaArchive.class);
-        final FileSystem vfs = ShrinkWrapFileSystems.newFileSystem(homeArchive);
-        PathManager.getInstance().useOverrideHomePath(vfs.getPath(""));
+        Path tempHome = Files.createTempDirectory("terasology-tte", (FileAttribute<?>[]) null);
```
keturn, line 59: Would [`@TempDir`](https://junit.org/junit5/docs/current/user-guide/#writing-tests-built-in-extensions-TempDirectory).work here too? I guess if the `@BeforeAll` annotation works here, `@TempDir` should too.

engine-tests/src/main/java/org/terasology/HeadlessEnvironment.java
```diff
@@ -291,14 +276,11 @@ protected void setupModuleManager(Set<Name> moduleNames) throws Exception {
...
-     * @throws IOException ShrinkWrap errors
-     */
     @Override
     protected void setupPathManager() throws IOException {
-        final JavaArchive homeArchive = ShrinkWrap.create(JavaArchive.class);
-        final FileSystem vfs = ShrinkWrapFileSystems.newFileSystem(homeArchive);
-        PathManager.getInstance().useOverrideHomePath(vfs.getPath(""));
+        Path tempHome = Files.createTempDirectory("terasology-env", (FileAttribute<?>[]) null);
+        tempHome.toFile().deleteOnExit();
+        PathManager.getInstance().useOverrideHomePath(tempHome);
```
keturn, line 283: I had to double-check to make sure this isn't a HeadlessEnvironment used by servers.

It's in `engine-tests`, so I _think_ it is separate.

engine-tests/src/main/java/org/terasology/ModuleEnvironmentTest.java
```diff
@@ -37,9 +23,9 @@
 
     @BeforeEach
     public void before() throws Exception {
-        final JavaArchive homeArchive = ShrinkWrap.create(JavaArchive.class);
-        final FileSystem vfs = ShrinkWrapFileSystems.newFileSystem(homeArchive);
-        PathManager.getInstance().useOverrideHomePath(vfs.getPath(""));
+        Path tempHome = Files.createTempDirectory("terasology-met", (FileAttribute<?>[]) null);
```
keturn, line 26: Would [`@TempDir`](https://junit.org/junit5/docs/current/user-guide/#writing-tests-built-in-extensions-TempDirectory).work here too? I guess if the `@BeforeEach` annotation works here, `@TempDir` should too.

keturn:
oh hey there Jenkins, that's some failing tests showing up

keturn:
Oh. A bunch of NPEs because it doesn't like that I tried to pass null as the list of attributes. That's easily explicable, at least.

engine-tests/src/main/java/org/terasology/HeadlessEnvironment.java
DarkWeird: yep it is separate

DarkWeird (approved):
Use `@TempDir` at beforeEach and don't worry.

PR #4562, opened 2021-03-05
refactor: move DiscordRPC to org.terasology.subsystem

keturn:
to indicate it is not in engine itself.

following notes about package namespaces on #4560.


keturn:
We can do moves like this without much apprehension, right? Because there are not things out in module-space referring to its classes?

pollend:
do we need to move the packages for type-handler and gson .

pollend (approved):

PR #4564, opened 2021-03-05
chore: upgrade reflections to 0.9.12-MB from 0.9.10

keturn:
Using the org.terasology fork.

This is the version that gestalt-v7 uses. It might also work fine with our current one? Not too sure. Making a branch to see what CI thinks.

It's also possible that not all of these build files need the reflections dependency. e.g. do the facades make any direct use of it?

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -20,7 +20,7 @@ repositories {
 
 dependencies {
     // Needed for caching reflected data during builds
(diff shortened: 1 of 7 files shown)


pollend (approved):

pollend:
reflections seem fine and I can load a couple different game-play templates without problem.

pollend (approved):

PR #4573, opened 2021-03-12
test: unittest module may be in a jar

keturn:
Fixes #4570

at least, gets some tests past ModuleManagerFactory.create, but they may still have other problems.

PR #4574, opened 2021-03-13
test: HeadlessEnvironment loads modules from classpath

keturn:
Tests using engine-test's [HeadlessEnvironment](https://github.com/MovingBlocks/Terasology/blob/ecb5c56a3b2c3e80caef64524b01f5c76fc4b2ed/engine-tests/src/main/java/org/terasology/HeadlessEnvironment.java#L104) have a lot in common with the issues in https://github.com/Terasology/ModuleTestingEnvironment/pull/41, so they could use the same sort of treatment, which is "turn on classpath module loading until we have a better idea."

This change to HeadlessEnvironment is a more general solution than that in https://github.com/Terasology/Pathfinding/pull/57, possible because HeadlessEnvironment directly creates its own ModuleManager so we don't need to try to indirectly influence ModuleManager creation by something in the global environment.

This PR is built on #4573 because they both modify ModuleManagerFactory. 

### How to test

Run tests involving HeadlessEnvironment or WorldProvidingHeadlessEnvironment. Check the [Unhealthy Modules](http://jenkins.terasology.io/teraorg/job/Terasology/job/Modules/view/Unhealthy/) list on Jenkins, try running those tests under gradle locally, and see if this branch makes the number of failures go down or change in useful ways.

### DIFF
--- a/engine-tests/src/main/java/org/terasology/HeadlessEnvironment.java
+++ b/engine-tests/src/main/java/org/terasology/HeadlessEnvironment.java
@@ -256,7 +256,7 @@ public class HeadlessEnvironment extends Environment {
         TypeRegistry.WHITELISTED_CLASSES = ExternalApiWhitelist.CLASSES.stream().map(Class::getName).collect(Collectors.toSet());
         context.put(TypeRegistry.class, typeRegistry);
 
-        ModuleManager moduleManager = ModuleManagerFactory.create();
+        ModuleManager moduleManager = ModuleManagerFactory.create(true);
         ModuleRegistry registry = moduleManager.getRegistry();
 
         DependencyResolver resolver = new DependencyResolver(registry);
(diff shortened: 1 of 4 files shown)


keturn:
Some self-review notes:

ModuleManager constructor proliferation and ModuleManagerFactory: Wow, languages that don't have default values for function arguments are kind of a mess, aren't they?

I think we literally have more constructors than callers at this point.

I should know better than to commit anything saying "this is temporary," so I'll say this instead: I'm not dedicated to this interface for ModuleManager. I hope we can do some pruning during or after migrating to gestalt-v7 and we know what we want from it.

doLoadModulesFromClasspath's ridiculous number of ways of setting a flag: ditto on the "not committed to this." The environment variable was my first idea, as it'd be easy to set when running the entire game, not just for tests. But if nobody's needed to use that, we can drop that part.

and if we're lucky, we get to remove the option entirely if our gestalt-v7 implementation finds a way to load module dependencies without this everything-in-the-classpath approach.

pollend (approved):
I see similar logic in this PR: https://github.com/MovingBlocks/Terasology/pull/4573

this should be fine right?

PR #4578, opened 2021-03-14
chore: migrate into engine

pollend:
moved packages for gestalt. 

ref: https://github.com/MovingBlocks/Terasology/pull/4504


keturn:
don't forget `engine-test`! and removing the idea packagePrefix settings from both gradle files as described in https://github.com/MovingBlocks/Terasology/pull/4504#issuecomment-798782598

pollend:
uh oh @keturn umm I'll get right to that

DarkWeird (approved):
Omega - builds
Multiplayer works with MetalRenegates (tested: login and walk around) (server and client)
Singleplayer works with `JoshariasSurvival` (creating world, walk around)

PR #4579, opened 2021-03-14
test(modules): set default test case timeout to four minutes

keturn:
This should cause Jupiter to interrupt if any test case or setup method takes longer than four minutes to run.


keturn:
fixes https://github.com/Terasology/ModuleTestingEnvironment/issues/25

skaldarnar (approved):

PR #4581, opened 2021-03-15
ModuleManager: ensure engine module is always present in environment

antag99:
Should fix crash upon returning to menu from in-game state. I think the commit that initially caused this bug was 8ae06f0df3233422d397da23119b7bf97d8e99d4, but I have not tested the revision before.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/module/ModuleManager.java
+++ b/engine/src/main/java/org/terasology/engine/module/ModuleManager.java
@@ -67,6 +67,7 @@ public class ModuleManager {
     private ModuleEnvironment environment;
     private final ModuleMetadataJsonAdapter metadataReader;
     private final ModuleInstallManager installManager;
+    private Module engineModule;
 
     public ModuleManager(String masterServerAddress) {
         this(masterServerAddress, Collections.emptyList());
@@ -77,7 +78,7 @@ public class ModuleManager {
 
         metadataReader = newMetadataReader();
 
-        Module engineModule = loadEngineModule(classesOnClasspathsToAddToEngine);
+        engineModule = loadEngineModule(classesOnClasspathsToAddToEngine);
 
         registry = new TableModuleRegistry();
         registry.add(engineModule);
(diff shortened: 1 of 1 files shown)


keturn (approved):
approved, pending a successful build on CI when JCenter comes back online.

There is some related clean-up that _could_ be done, but since this is fixing a severe regression I won't ask anyone to take extra time to do that before merging this.

(I'm thinking of other places in ModuleManager that use engineModule. e.g. `ensureModulesDependOnEngine` doesn't need this passed in as an argument if it's on the class. There might be others.)

Other thoughts: Where are these places that are creating an environment without the engine?
and why, for instance, does it happen when going from the game _back_ to the menu, but not the _first_ time we see the menu on startup?

antag99:
> There is some related clean-up that _could_ be done, but since this is fixing a severe regression I won't ask anyone to take extra time to do that before merging this.
> 
> (I'm thinking of other places in ModuleManager that use engineModule. e.g. `ensureModulesDependOnEngine` doesn't need this passed in as an argument if it's on the class. There might be others.)

Yeah, somewhat messy but I left these the way they were in case someone later wants to remove the engineModule field again.


> Other thoughts: Where are these places that are creating an environment without the engine?
> and why, for instance, does it happen when going from the game _back_ to the menu, but not the _first_ time we see the menu on startup?

I believe this is the cause: https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/modes/StateIngame.java#L147

I think bugs like this could be prevented by having the same code run to setup the module environment regardless of whether it is done upon initialization of the engine or when restoring the environment after leaving the in-game state.

keturn:
yep, that `loadEnvironment` passing the empty set would do it.

I've been noticing some patterns in the way ModuleManager is used — especially `loadEnvironment` — and I think we could come up with an API that's less error-prone.

but that's one of the things I'm holding off on until after we find out what changes we make to ModuleManager for gestalt-module v7.

keturn:
build succeeded! merged! Thanks for digging in to this!

PR #4583, opened 2021-03-16
Add utility methods to BlockRegion and Chunks

sin3point14:
### Contains

Add utility functions for TutorialWorldGeneration

### How to test

Build with https://github.com/Terasology/TutorialWorldGeneration/pull/19

### DIFF
--- a/engine/src/main/java/org/terasology/world/block/BlockRegion.java
+++ b/engine/src/main/java/org/terasology/world/block/BlockRegion.java
@@ -87,6 +87,21 @@ public class BlockRegion implements BlockRegionc {
     }
 
     /**
+     * Creates a new region from center and extent
+     */
+    static public BlockRegion createFromCenterExtents(Vector3i center, int extents) {
+        Vector3i extentVec = new Vector3i(extents, extents, extents);
+        return new BlockRegion(center.sub(extentVec, new Vector3i()), center.add(extentVec, new Vector3i()));
+    }
+
+    /**
+     * Creates a new region from min and size
+     */
+    static public BlockRegion createFromMinAndSize(Vector3i min, Vector3ic size) {
+        return new BlockRegion(min, min.add(size, new Vector3i()));
+    }
+
+    /**
      * Creates a new region containing the single block given by the coordinates.
      */
     public BlockRegion(int x, int y, int z) {
--- a/engine/src/main/java/org/terasology/world/chunks/Chunks.java
+++ b/engine/src/main/java/org/terasology/world/chunks/Chunks.java
@@ -325,8 +325,8 @@ public final class Chunks {
      * @param worldPos world position
      * @return the modified {@code worldPos} vector
      */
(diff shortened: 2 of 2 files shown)


engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -86,6 +86,21 @@ public BlockRegion(Vector3ic min, Vector3ic max) {
         this.set(min.x(), min.y(), min.z(), max.x(), max.y(), max.z());
     }
 
+    /**
+     * Creates a new region from center and extent
+     */
+    static public BlockRegion createFromCenterExtents(Vector3i center, int extents) {
+        Vector3i extentVec = new Vector3i(extents, extents, extents);
+        return new BlockRegion(center.sub(extentVec, new Vector3i()), center.add(extentVec, new Vector3i()));
```
pollend, line 94:
```suggestion
        return new BlockRegion(center).expand(extents, extents,extents);
```

engine/src/main/java/org/terasology/world/block/BlockRegion.java
```diff
@@ -86,6 +86,21 @@ public BlockRegion(Vector3ic min, Vector3ic max) {
...
+    }
+
+    /**
+     * Creates a new region from min and size
+     */
+    static public BlockRegion createFromMinAndSize(Vector3i min, Vector3ic size) {
+        return new BlockRegion(min, min.add(size, new Vector3i()));
```
pollend, line 101:
```suggestion
        return new BlockRegion(min).expand(size);
``````

**Review by pollend — CHANGES_REQUESTED** [review]

File: engine/src/main/java/org/terasology/world/chunks/Chunks.java [hunk 3]
```diff
@@ -325,8 +325,8 @@ public static Vector3i toRelative(Vector3ic worldPos, Vector3i dest) {
      * @param worldPos world position
      * @return the modified {@code worldPos} vector
      */
-    public static Vector3i toRelative(Vector3i worldPos) {
-        return toRelative(worldPos, worldPos);
+    public static Vector3i toRelative(Vector3ic worldPos) {
+        return toRelative(worldPos, new Vector3i(worldPos));
```
skaldarnar, line 329: We've been taking the decision to no longer hide those defensive copies of vectors we were doing all over the place. This, along with other methods, deliberately takes a mutable vector because it will modify it in-place. 

Always allocating a new vector here would take away the benefit of being able to reuse temporary vectors.

skaldarnar (changes requested):
I agree with @pollend here - there's no need to add those utility methods anymore, as createing `BlockRegion`s with the given properties can easily be achieved as in the suggestions.

Thus, I'd suggest to just close this PR.

@sin3point14 please ask more questions if something is unclear, much of this is related to the whole JOML migration and everything we learned during all those PRs.

pollend:
There are no actionable changes here so I'm closing this PR.

PR #4584, opened 2021-03-17
refactor(CreateWorldEntity): aggressive refactoring

keturn:
I tried to review #4531, but I could not wrap my head around where data was coming and going in CreateWorldEntity, so I put it through the wringer.

This was _mostly_ pure refactors, even if I did stop making a commit for each one. But there is one significant twist:

Instead of applying settings from Config to the WorldConfigurator _and_ the worldEntity in one pass,  
it sets things in the WorldConfigurator and _then_ uses the values from the WorldConfigurator on the worldEntity.

that makes sense, right? is there someplace else that does anything similar? that is, hopefully something is using those WorldConfigurator properties?


engine/src/main/java/org/terasology/engine/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -52,64 +55,101 @@ public String getMessage() {
 
     @Override
     public boolean step() {
+        if (worldEntityExists()) {
+            EntityRef worldEntity = getExistingWorldEntity();
+            configureChunkProvider(worldEntity);
+            useConfigurationOfCurrentWorld(worldEntity);
+        } else {
+            EntityRef worldEntity = createWorldPoolsAndEntity();
+            configureChunkProvider(worldEntity);
```
keturn, line 64: is it important that configureChunkProvider happens before we do other stuff with the WorldConfigurator, or may we move it down out of this branch to the common path?

engine/src/main/java/org/terasology/engine/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -52,64 +55,101 @@ public String getMessage() {
...
-                Component comp = worldEntity.getComponent(clazz);
-                if (comp != null) {
-                    worldConfigurator.setProperty(entry.getKey(), comp);
-                }
+    private void useConfigurationOfCurrentWorld(EntityRef worldEntity) {
+        // replace the world generator values from the components in the world entity
+        WorldConfiguratorConfigurator configuratorConfigurator = new WorldConfiguratorConfigurator() {
+            @Override
+            public <T extends Component> T getComponentOfWorldEntity(String key, Class<T> clazz) {
+                return worldEntity.getComponent(clazz);
```
keturn, line 83: I tried to use a lambda but it said something about target parameter being generic?

would love it if this didn't take five lines or so. I miss Kotlin.

engine/src/main/java/org/terasology/engine/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -52,64 +55,101 @@ public String getMessage() {
...
+        Iterator<EntityRef> worldEntityIterator = entityManager.getEntitiesWith(WorldComponent.class).iterator();
+        EntityRef worldEntity = worldEntityIterator.next();
+        worldEntityIterator.forEachRemaining(w -> logger.warn("Ignored extra world {}", w));
+        return worldEntity;
+    }
+
+    private EntityRef createWorldPoolsAndEntity() {
+        EntityManager entityManager = context.get(EntityManager.class);
+
+        entityManager.createWorldPools(gameManifest);
```
keturn, line 115: what even are world pools? we don't seem to use them when creating the world entity.

also, this is the only place that uses GameManifest, which is sus.

engine/src/main/java/org/terasology/engine/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -52,64 +55,101 @@ public String getMessage() {
...
+
+        EntityRef worldEntity = entityManager.create();
+        worldEntity.addComponent(new WorldComponent());
+        NetworkComponent networkComponent = new NetworkComponent();
+        networkComponent.replicateMode = NetworkComponent.ReplicateMode.ALWAYS;
+        worldEntity.addComponent(networkComponent);
```
keturn, line 121: I was wondering if this should use a Component Builder, but it didn't seem to make this easier to read and I didn't need to otherwise mess with this block, so I left it alone.

keturn:
Another observation: All a WorldConfigurator is is a bag of String -> Component properties. https://github.com/MovingBlocks/Terasology/blob/dcb5ec6b523e5da80556ca928824a55ec168337a/engine/src/main/java/org/terasology/world/generator/WorldConfigurator.java#L18-L24

so if a Config or GameManifest or Entity is _also_ a bag of String -> Component properties, they could even _be_ a WorldConfigurator themselves.

and then maybe something like

```java
class WorldConfigurator {
    void import(WorldConfigurator other) {
        other.getPropeties.forEach((k, v) -> { this.setProperty(k,v) });
    }
}
```

keturn:
also -- and I'm not saying all of this should go in this PR, I'm just collecting notes at this point -- GameManifest has worlds: https://github.com/MovingBlocks/Terasology/blob/dcb5ec6b523e5da80556ca928824a55ec168337a/engine/src/main/java/org/terasology/game/GameManifest.java#L119

which know their WorldGenerator, so we should probably be pulling WorldGenerators (and their Configurators) out of there instead of the Context?

engine/src/main/java/org/terasology/engine/modes/loadProcesses/CreateWorldEntity.java
pollend: The concept was you can store entities by chunk.

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -52,64 +55,101 @@ public String getMessage() {
...
     public boolean step() {
+        if (worldEntityExists()) {
+            EntityRef worldEntity = getExistingWorldEntity();
+            configureChunkProvider(worldEntity);
+            useConfigurationOfCurrentWorld(worldEntity);
+        } else {
+            EntityRef worldEntity = createWorldPoolsAndEntity();
+            configureChunkProvider(worldEntity);
+            WorldConfigurator worldConfigurator = useConfigurationFromConfig();
+            configurateWorldEntity(worldConfigurator, worldEntity);
```
pollend, line 66: I think I would opt to combine these two methods. useConfigurationFromConfig is just retrieving the config from context.

pollend (changes requested):
my only concern is the heavy use of context at some point this will have to be untangled where we pull context out from the implementation. maybe assign everything to protected variables to untangle the implementation just a bit.  you can do this in the step phase since I have no idea what state of the context is in with initialization.

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -31,12 +33,13 @@
...
-import java.util.Map;
 
 /**
  */
 public class CreateWorldEntity extends SingleStepLoadProcess {
 
+    private static final Logger logger = LoggerFactory.getLogger(CreateWorldEntity.class);
+
     private final Context context;
     private final GameManifest gameManifest;
```
pollend, line 44: would you also note that this GameManifest is a completely different game manifest from

GameManifestProvider.java

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/CreateWorldEntity.java
keturn: You've discovered what I was aiming at during this refactoring!

Yes, my goal was to have a function that provides the appropriately configured WorldConfigurator, so I could figure out what would be different when changing where it gets that information as in #4531.

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/CreateWorldEntity.java
keturn: 😖

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -52,64 +55,101 @@ public String getMessage() {
...
+        worldEntity.addComponent(networkComponent);
+        return worldEntity;
+    }
+
+    private void configureChunkProvider(EntityRef worldEntity) {
+        context.get(ChunkProvider.class).setWorldEntity(worldEntity);
+    }
+
+    private void configurateWorldEntity(WorldConfigurator worldConfigurator, EntityRef worldEntity) {
+        worldConfigurator.getProperties().forEach((key, component) -> {
```
keturn, line 130: "configurate" is the sort of joke name that was _supposed_ to get replaced before the final draft.

keturn (commented):
Stepping away from this for a couple days and looking at it again, I don't hate this as a refactoring.

There are some places that are clearly calling out for some follow-up steps, but the one thing that bothers me about the readability is the excessive amount of boilerplate used in defining a `getComponentOfWorldEntity` method.

Is there a more elegant way to write that?

Would there be if we were using a more recent Java?

keturn:
> my only concern is the heavy use of context at some point this will have to be untangled where we pull context out from the implementation.

What do we expect it to be replaced by?

things attached to the GameManifest passed to the LoadProcess?

things injected as arguments to the `step` method?

engine/src/main/java/org/terasology/engine/modes/loadProcesses/CreateWorldEntity.java
keturn: by making the signature less precise and saying it returns `Component` instead of specifically `T`, java is now willing to let me use a lambda here.

pollend (dismissed):

jdrueckert:
@keturn IIRC you wanted to take another look at this? Is this correct or is it good to be merged? :thinking:

keturn:
Ah, this is where I was discovering the challenges of doing asynchronous-collaboration on a refactoring branch. I hadn't laid out my entire thought process, outside some brief commit messages that got sketchier as the night went on.

This has had some push-me-pull-you action. I expect some of that results in a good compromise, I worry some of it is pulling in different directions in counterproductive ways. Want to review to sort which is which. At the moment, this is on the top of my queue, so hopefully Soon.

keturn:
other note: I think this review process has been _far_ too drawn-out for a single-file refactoring PR, but since we crossed that line long ago, I figured this might as well be in the slow lane and use it as a refactoring discussion opportunity.

pollend:
There is no direct impact to the code-base so we can take time. @keturn

keturn (commented):
some refactoring conversation, but also one more concrete question

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -52,64 +64,59 @@ public String getMessage() {
 
     @Override
     public boolean step() {
-        EntityManager entityManager = context.get(EntityManager.class);
-        ChunkProvider chunkProvider = context.get(ChunkProvider.class);
+        this.entityManager = context.get(EntityManager.class);
+        this.worldGenerator = context.get(WorldGenerator.class);
+        this.config = context.get(Config.class);
+        this.chunkProvider = context.get(ChunkProvider.class);
```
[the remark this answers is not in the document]
pollend: the plan here is to drive this all through dependency injection. all these loadprocess objects should be transient so they are built used and torn down. not sure when would be a good time to look at gestalt-DI but I think we can preemptively move these to fields and later drop usage of the context when running these steps.

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -55,26 +61,27 @@ public String getMessage() {
...
+            useConfigurationFromConfig(worldEntity);
         }
 
         return true;
     }
 
-    private boolean worldEntityExists() {
-        EntityManager entityManager = context.get(EntityManager.class);
-        Iterable<EntityRef> worldEntityIterator = entityManager.getEntitiesWith(WorldComponent.class);
-        return worldEntityIterator.iterator().hasNext();
```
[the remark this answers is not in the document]
pollend: I think I removed it because it added more logic to something that seemed straightforward to me. Its reused logic for the sake of decomposing a class.  I would probably opt more for adding comments why hasNext is necessary. i.e `if there is no world entity then we need to create one` then trying to derive it from more descriptive method names.  that also personal preference.

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/CreateWorldEntity.java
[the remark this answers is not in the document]
keturn: we're putting them in fields because it'll use field injection? or constructor injection, which will need to store them in fields somewhere? or does it provide injection on non-constructor methods too?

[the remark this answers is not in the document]
pollend: either will work but I would opt for field injection in this case. gestalt-DI will throw an exception if it can't find an object to inject unless you use Optional<>

jdrueckert:
[Refactor CreateWorldEntity](https://trello.com/c/yxdSZzXf/66-refactor-createworldentity)

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/CreateWorldEntity.java
```diff
@@ -52,64 +65,63 @@ public String getMessage() {
...
             // replace the world generator values from the components in the world entity
-            WorldGenerator worldGenerator = context.get(WorldGenerator.class);
-            WorldConfigurator worldConfigurator = worldGenerator.getConfigurator();
-            Map<String, Component> params = worldConfigurator.getProperties();
-            for (Map.Entry<String, Component> entry : params.entrySet()) {
-                Class<? extends Component> clazz = entry.getValue().getClass();
-                Component comp = worldEntity.getComponent(clazz);
-                if (comp != null) {
-                    worldConfigurator.setProperty(entry.getKey(), comp);
+            checkNotNull(worldConfigurator).getProperties().forEach((key, currentComponent) -> {
```
skaldarnar, line 82: this looks like a random null check - how does `checkNotNull` behave if it is null? just give a better error message than regular NPE?

pollend: it throws an exception if worldConfigurator is null.

keturn: Oh, I probably tossed that in there at some point when I got a null during development. If you give `checkNotNull` a second argument with a message, _then_ it's better than a Java < 14 NPE, but I am not sure if it's really better in this single-argument form.

skaldarnar (approved):
I do think what I see on the right is an improvement to what I see on the left, thus approving this PR.

I think there was a version somewhere in-between with more named methods which I personally liked (better). 

@keturn @pollend you've got an approval here, please merge at your convenience or revert to that other commit with more structuring by methods and ping me again.

keturn:
merged, as it doesn't seem like leaving this open was facilitating further discussion. Though I think we do still need to resolve the questions that came up here for some kind of code guidelines.

skaldarnar:
Please do update https://github.com/MovingBlocks/Terasology/wiki/Code-Conventions in any way you seem helpful :green_heart:

keturn:
I think working agreements for the team need to be a group decision, rather than decided by whoever edited the wiki last. Otherwise we'll end up editing it in circles same as we did for this PR. 😜

PR #4585, opened 2021-03-17
fix(player): target box not disappearing when target is removed

antag99:
### Contains
Fixes a bug where the target indication box does not disappear when the player points at a block that is not within range.

### How to test
- Open a world, hover over a block and rotate the camera upwards so no block is in range. The target indication box for the most recently targeted block should still be present.
- Apply the PR and repeat. The indication box should instead disappear.


pollend (dismissed):

pollend:
for some reason github shows a conflict but you should easily be able to merge this since the Package has only just moved with the recent change.

pollend (approved):

PR #4590, opened 2021-03-20
Don't move the camera when window is refocused

tolziplohu:
### Contains

Fix the camera position resetting to look at the sky when the window is unfocused and then focused again. It just resets the mouse position whenever the window is focused, which also means that if you unfocus, move the mouse, and focus again, the camera doesn't move.

### How to test

Start the game, move the camera, alt-tab to another window, then alt-tab back. The camera shouldn't move.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/input/lwjgl/LwjglMouseDevice.java
+++ b/engine/src/main/java/org/terasology/engine/input/lwjgl/LwjglMouseDevice.java
@@ -60,8 +60,6 @@ public class LwjglMouseDevice implements MouseDevice, PropertyChangeListener {
         DoubleBuffer mouseY = BufferUtils.createDoubleBuffer(1);
 
         GLFW.glfwGetCursorPos(window, mouseX, mouseY);
-        mouseX.rewind();
-        mouseY.rewind();
 
         double x = mouseX.get(0);
         double y = mouseY.get(0);
@@ -128,8 +126,6 @@ public class LwjglMouseDevice implements MouseDevice, PropertyChangeListener {
             DoubleBuffer mouseY = BufferUtils.createDoubleBuffer(1);
 
             GLFW.glfwGetCursorPos(window, mouseX, mouseY);
-            mouseX.rewind();
-            mouseY.rewind();
 
             xpos = mouseX.get(0);
             ypos = mouseY.get(0);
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/engine/input/lwjgl/LwjglMouseDevice.java
```diff
@@ -123,6 +124,26 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+
+            GLFW.glfwGetCursorPos(window, mouseX, mouseY);
+            mouseX.rewind();
+            mouseY.rewind();
+
+            double x = mouseX.get(0);
+            double y = mouseY.get(0);
+    
+            this.xpos = x;
+            this.ypos = y;
```
keturn, line 143: suggestion: I think these are short enough that you could `xpos = mouseX.get(0)` without the intermediate local variable.

keturn (changes requested):
The explanation makes sense and this does look like the right place to do it.

I'm not familiar with the LWJGL API, but this is at least consistent with the way `update` does it.

engine/src/main/java/org/terasology/engine/input/lwjgl/LwjglMouseDevice.java
```diff
@@ -123,6 +117,21 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+        // reported by GLFW change even if the mouse hasn't moved.
+        // So when the window regains focus, we set the new position as the reference without generating deltas.
+        if (isFocused) {
+            DoubleBuffer mouseX = BufferUtils.createDoubleBuffer(1);
+            DoubleBuffer mouseY = BufferUtils.createDoubleBuffer(1);
+
+            GLFW.glfwGetCursorPos(window, mouseX, mouseY);
+
+            xpos = mouseX.get(0);
+            ypos = mouseY.get(0);
```
pollend, line 131: can this be replaced with a Vector2d;

pollend (changes requested):

engine/src/main/java/org/terasology/engine/input/lwjgl/LwjglMouseDevice.java
tolziplohu: Yes, it can. That does make it shorter, as long as we're fine with `newPos.sub(pos, posDelta)` being somewhat harder to read than `xposDelta = x - this.xpos`.

keturn (approved):

engine/src/main/java/org/terasology/engine/input/lwjgl/LwjglMouseDevice.java
keturn: It would also require corresponding changes to the `update` method which uses those fields.

I know I've encouraged a bit of scope creep here by bringing up the code duplication in `update` and the use of `rewind`, but we should probably not stack _all_ of the possible clean-ups we can think of in to this bugfix PR.

skaldarnar (approved):

pollend (approved):

PR #4592, opened 2021-03-21
Fixing spelling.

mahela97:
File - https://github.com/mahela97/Terasology/blob/2aa5cbbf6b08bdf8826acc3b3f8517815df9eb5c/engine/src/main/java/org/terasology/engine/utilities/procedural/SimplexNoise.java

Line -73

### Contains

Fix spelling errors in the documentation. The correct name of the variable is TILEABLE1DMAGICNUMBER.

- Before it was  ->
@see SimplexNoise#tileable1DMagicNumber

- Now ->
@see SimplexNoise#TILEABLE1DMAGICNUMBER

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -12,5 +12,5 @@
     </writeAnnotations>
   </component>
   <component name="ExternalStorageConfigurationManager" enabled="true" />
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" default="true" project-jdk-name="11" project-jdk-type="JavaSDK" />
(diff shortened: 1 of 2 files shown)


.idea/misc.xml
```diff
@@ -12,5 +12,5 @@
     </writeAnnotations>
   </component>
   <component name="ExternalStorageConfigurationManager" enabled="true" />
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="11" project-jdk-type="JavaSDK" />
-</project>
```
pollend, line 16: can you revert this change.

pollend (changes requested):

.idea/misc.xml
mahela97: sure

PR #4597, opened 2021-03-23
feat(entitySystem): Introduce DoNotPersist annotation

antag99:
### Contains

Useful for components that store information relevant only to the running session.
The "null" component serialization check is renamed to "default", and the EntityInfoComponent is marked with the new annotation instead of being a special case.

### How to test

Run the unit tests in EntitySerializerTest.


pollend:
If were going to introduce another annotation I rather make it apart of gestalt. At some point we should start using the entity system in gestalt.

https://github.com/MovingBlocks/gestalt

antag99:
> If were going to introduce another annotation I rather make it apart of gestalt. At some point we should start using the entity system in gestalt.
> 
> https://github.com/MovingBlocks/gestalt

It seems that neither EntitySerializer nor ComponentMetadata is present in gestalt yet. So I think the thing this PR implements would either be a part of the main engine repo, or could be migrated over to gestalt later.

engine/src/main/java/org/terasology/engine/entitySystem/DoNotPersist.java
```diff
@@ -0,0 +1,17 @@
...
+
+package org.terasology.engine.entitySystem;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used to mark components that are not to be stored upon serialization of an entity.
```
skaldarnar, line 12: Is this only about _serialization_ for persistence (e.g., save game) or also about _network serialization_?

skaldarnar (commented):
The changes look pretty clean to me, and removing a special case for a specific component sounds very reasonable to me :+1:

pollend:
> > If were going to introduce another annotation I rather make it apart of gestalt. At some point we should start using the entity system in gestalt.
> > https://github.com/MovingBlocks/gestalt
> 
> It seems that neither EntitySerializer nor ComponentMetadata is present in gestalt yet. So I think the thing this PR implements would either be a part of the main engine repo, or could be migrated over to gestalt later.

I  think we can define the annotation in gestalt-entity system and use it from there potentially. would help later when and if we start using gestalt-entity-system. I guess we have to swap all the imports at some point so maybe this isn't soo bad. what do you think @skaldarnar.

engine/src/main/java/org/terasology/engine/entitySystem/DoNotPersist.java
pollend: I would opt for a @Persisted flag then @DoNotPersist but I do realize that would be a very large change. When I think about serialization I would prefer to opt into persisting something to disk vs not persisting it.

antag99: It is about serialization for persistence, the network serialization does not use the default component serialization check. But considering this I think it is more reasonable to reinstate the "null" component check and introduce a PersistenceComponentSerializeCheck that looks for the annotation. (I will do this on my local branch.)

antag99:
Hmm, I ran those tests locally and they passed the second time after rebuild. Must have missed something. Anyway, the EntityInfoComponent is decidedly a special case and has been reinstated as such :D. (Could also be handled separately in (Network)EntitySerializer).

skaldarnar (approved):

antag99:
It seems a change was overwritten by merge, I have made a fixup commit.

skaldarnar (approved):

PR #4598, opened 2021-03-23
chore: update to TeraNUI 3.0 from 1.5, for gestalt separation

keturn:
This updates Terasology for https://github.com/MovingBlocks/TeraNUI/pull/47

Making less of NUI depend on a specific version of gestalt, which will make our gestalt upgrade easier.

Based on https://github.com/BenjaminAmos/Terasology/tree/nui-gestalt-separation  
I hope @BenjaminAmos will forgive me for making a PR of it.

### How to test


engine/src/main/java/org/terasology/engine/persistence/typeHandling/extensionTypes/UISkinTypeHandler.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Optional;
+
+public class UISkinTypeHandler extends StringRepresentationTypeHandler<UISkin> {
+    @Override
+    public String getAsString(UISkin item) {
+        if (item == null) {
+            return "";
+        }
+        // TODO
+        return item.toString();
```
keturn, line 21: @BenjaminAmos "TODO"? Is this not quite right?

keturn (commented):
testing out well so far

engine/src/main/java/org/terasology/engine/persistence/typeHandling/extensionTypes/UISkinTypeHandler.java
BenjaminAmos: When this was handled by `AssetTypeHandler`, then this method would return the asset urn for the skin.
https://github.com/MovingBlocks/Terasology/blob/cbc07339ea0fdbfaf1cfe31beee8b0b44a25834e/engine/src/main/java/org/terasology/engine/persistence/typeHandling/extensionTypes/AssetTypeHandler.java#L22-L28
When given a `UISkin` though, there is no direct mapping to the resource that generated it. I wasn't sure how to return the source asset urn from there. Any help with that bit would be appreciated.

It's TODO because I have no idea what the consequences of leaving this unimplemented will be but given that it relates to serialisation, it could cause some issues.

jdrueckert: @DarkWeird @eviltak any ideas?

DarkWeird: > @DarkWeird @eviltak any ideas?

1. store reverse link somewhere. 
2. don't save UISkin separately
3. creates common (UISkin/UISkinAsset)TypeHandlerFactory

BenjaminAmos: I've given this another attempt, however it's still a work-around. It involves sub-classing `UISkin` to add a `ResourceUrn` to it and then copying the retrieved skin into an instance of that subclass. It's at https://github.com/BenjaminAmos/Terasology/commit/c8ffd14a9325cd5515156aa1fc07f0f03b2bc128. If I can't think of anything better soon, then I'll send a PR for it. Suprisingly though, I couldn't actually find anywhere where this particular method is called. Components don't tend to store `UISkin` elements directly and so-far I have only found deserialisation usage for this type handler. That may be why having this return invalid values doesn't seem to affect gameplay.

BenjaminAmos: I can't see how a reverse link would work, as two assets could hypothetically contain the same content yet be from different sources. `AssetTypeHandler` handles all asset types, not just `UISkin`. I haven't thought of any other solutions, so, whilst it's not ideal, I've submiited #4604 to potentially resolve this.

DarkWeird (approved):
Looks good for me.
Works all basic screens works in main menu, also works controls.

Not works loading/creating world(not found world generator) - i think that because i have dirty omega workspace

keturn:
I ran omega tests and they are _mostly_ good, except some NameGenerator tests that CI says weren't failing in the last build. 

org.terasology.namegenerator.CreatureNameProviderTest

```
java.lang.ExceptionInInitializerError
	at org.terasology.namegenerator.creature.CreatureNameProvider.<init>(CreatureNameProvider.java:46)
	at org.terasology.namegenerator.CreatureNameProviderTest.testBase(CreatureNameProviderTest.java:47)
Caused by: java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.get(Optional.java:148)
	at org.terasology.namegenerator.creature.CreatureAssetTheme.fetchNameList(CreatureAssetTheme.java:94)
	at org.terasology.namegenerator.creature.CreatureAssetTheme.<init>(CreatureAssetTheme.java:65)
	at org.terasology.namegenerator.creature.CreatureAssetTheme.<clinit>(CreatureAssetTheme.java:34)
	... 93 more
```

and this is one of those things where it's hard to imagine it being NUI-related.

It might be a CI vs local thing instead? should test develop locally to confirm.

Issue #4596, opened 2021-03-23
Can't open the game after we close the game from taskbar when the game is minimized


mahela97:
### What you were trying to do

The game crashed after the splash screen.

### What actually happened

I try to run the game as usual and it gave this exception. The reason for this exception is "windowWidth", "windowHeight" Values in config.cfg(In root directory) is set to 0(zero) after we close the game from the taskbar when the game is minimized.  

java.lang.RuntimeException: Failed to create window
	at org.terasology.engine.core.subsystem.lwjgl.LwjglGraphics.initWindow(LwjglGraphics.java:152)
	at org.terasology.engine.core.subsystem.lwjgl.LwjglGraphics.postInitialise(LwjglGraphics.java:78)
	at org.terasology.engine.core.TerasologyEngine.postInitSubsystems(TerasologyEngine.java:295)
	at org.terasology.engine.core.TerasologyEngine.initialize(TerasologyEngine.java:225)
	at org.terasology.engine.core.TerasologyEngine.initializeRun(TerasologyEngine.java:408)
	at org.terasology.engine.core.TerasologyEngine.run(TerasologyEngine.java:399)
	at org.terasology.engine.Terasology.main(Terasology.java:179)
For more details, see the log files in D:\My works\Terasology\logs\2021-03-23_16-09-36 

### How to reproduce

* Step 1
 Open the game as usual. 
* Step 2
 Minimize the game and close the game from the preview in the taskbar. Do not maximize again.
* Step 3
Run the game again and it will crash after the splash screen.

### Computer details

Asus tuf 504, Running on windows 10

Cervator:
Fixed by #4619

PR #4601, opened 2021-03-25
fix: auto-correct window resolution if set to (0,0)

mahela97:
### Contains

This PR will fix the Bug mentioned here #4596

### How to test

Close the game when it is minimized and open the game again. Game window resolution will set to default values.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/core/subsystem/lwjgl/LwjglGraphics.java
+++ b/engine/src/main/java/org/terasology/engine/core/subsystem/lwjgl/LwjglGraphics.java
@@ -60,6 +60,12 @@ public class LwjglGraphics extends BaseLwjglSubsystem {
         this.engine = gameEngine;
         this.context = rootContext;
         this.config = context.get(Config.class).getRendering();
+        if (config.getWindowHeight()<=0 && config.getWindowWidth()<=0){
+            config.setWindowPosX(379);
+            config.setWindowPosY(235);
+            config.setWindowWidth(1280);
+            config.setWindowHeight(800);
+        }
         lwjglDisplay = new LwjglDisplayDevice(context);
         context.put(DisplayDevice.class, lwjglDisplay);
         logger.info("Initial initialization complete");


keturn:
I'm glad you figured out what was causing this! It seems I can't test it here, because my window manager doesn't trigger the bug, but it sounds likely.

From what you wrote in #4596, the problem comes from zeros being written to those width and height values in the configuration file. Here you're detecting the problem in LwjglGraphics.

Suggestion **A**: We should catch it earlier, because this sounds like a bad configuration for any renderer. Move the checks from here up to a `Config` class so things like `config.getWindowWidth` never return zero. I _think_ the place to do that is probably [RenderingConfig](https://github.com/MovingBlocks/Terasology/blob/46d650d434c02d636a5e8dce460985456eec96de/engine/src/main/java/org/terasology/engine/config/RenderingConfig.java), but I'm not 100% sure because [config stuff has been changing recently](https://github.com/MovingBlocks/Terasology/pull/4599).

Suggestion **B**: You've provided values here (1280×800+379+235) which are better than zero, but it's not clear where those values come from or if they match any other defaults in the system. Is there any variable or method that has default values for configuration options that you could call?

Suggestion **C**: Detecting and adjusting for an invalid config as you've done here is a good step. It makes the system more resilient in a way that will save us regardless of the cause, this closed-when-minified thing or something else. And _also,_ we should try to avoid writing a bad config file in the first place. Can we fix it so minifying the window doesn't end up writing zeros to the config?

(I think you could, if you like, hold off on **C** and do that in a separate pull request. I expect it to involve slightly different work and we don't need to hold up this first part of the bug fix to wait for it.)

engine/src/main/java/org/terasology/engine/core/subsystem/lwjgl/LwjglGraphics.java
```diff
@@ -60,6 +60,12 @@ public void initialise(GameEngine gameEngine, Context rootContext) {
         this.engine = gameEngine;
         this.context = rootContext;
         this.config = context.get(Config.class).getRendering();
+        if (config.getWindowHeight()<=0 && config.getWindowWidth()<=0){
+            config.setWindowPosX(379);
+            config.setWindowPosY(235);
+            config.setWindowWidth(1280);
+            config.setWindowHeight(800);
```
pollend, line 67: these values are pretty arbitrary. you only se to set the width and height since position is irrelevant. probably should define a constant for a default resolution.

engine/src/main/java/org/terasology/engine/core/subsystem/lwjgl/LwjglGraphics.java
```diff
@@ -60,6 +60,12 @@ public void initialise(GameEngine gameEngine, Context rootContext) {
         this.engine = gameEngine;
         this.context = rootContext;
         this.config = context.get(Config.class).getRendering();
+        if (config.getWindowHeight()<=0 && config.getWindowWidth()<=0){
```
pollend, line 63:
```suggestion
        if (config.getWindowHeight() <= 0 && config.getWindowWidth() <= 0) {
```

pollend (changes requested):

engine/src/main/java/org/terasology/engine/core/subsystem/lwjgl/LwjglGraphics.java
mahela97: These are the values that are generated when we delete the config file and reopen the game. So i took them as default values.

DarkWeird:
@keturn 
>Suggestion A: We should catch it earlier, because this sounds like a bad configuration for any renderer. Move the checks from here up to a Config class so things like config.getWindowWidth never return zero. I think the place to do that is probably RenderingConfig, but I'm not 100% sure because config stuff has been changing recently.

Linked PR is PlayerConfig, not RenderingConfig. - and yeah, AutoConfig can handle not saving zero values.

skaldarnar:
I' appreciate if we would make use of the config features here and prevent 0 values (option A) :+1:

mahela97:
> I' appreciate if we would make use of the config features here and prevent 0 values (option A) 👍

It is getting that value from the GLFW library. So we cant change it right? 😕 

from this method

`public static void glfwGetWindowSize(@NativeType("GLFWwindow *") long window, @Nullable @NativeType("int *") int[] width, @Nullable @NativeType("int *") int[] height) {
        long __functionAddress = Functions.GetWindowSize;
        if (CHECKS) {
            check(window);
            checkSafe(width, 1);
            checkSafe(height, 1);
        }
        invokePPPV(window, width, height, __functionAddress);
    }`

engine/src/main/java/org/terasology/engine/core/subsystem/lwjgl/LwjglGraphics.java
keturn: and what generates those? Is there a default config file that gets copied in, or are they generated by some method in the config system? or do we somehow skip this bit entirely when there's no config file, and the defaults end up being something lwjgl picked when we didn't give it any other instructions?

PR #4604, opened 2021-03-27
TeraNUI 3.0 update: Implemented `UISkinTypeHandler.getAsString`

BenjaminAmos:
### Description
This pull request attempts to resolve https://github.com/MovingBlocks/Terasology/pull/4598#discussion_r600276900 through a work-around involving sub-classing `UISkin` into `UISkinWithUrn` in order to associate a `ResourceUrn` with it. When a `UISkin` is deserialised, then it is copied into a new `UISkinWithUrn` instance to associate it with the urn it was deserialised with.
### How to test
I haven't found anywhere where the serialisation part of this code is used yet whilst debugging, so I am not sure how to test this.
### Notes
This is based off #4598 and, if merged, should be merged into it.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/rendering/nui/skin/UISkinWithUrn.java
+++ b/engine/src/main/java/org/terasology/engine/rendering/nui/skin/UISkinWithUrn.java
@@ -1,0 +1,33 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.engine.rendering.nui.skin;
+
+import org.terasology.assets.ResourceUrn;
+import org.terasology.nui.skin.UISkin;
+import org.terasology.nui.skin.UIStyleFamily;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class UISkinWithUrn extends UISkin {
+    private final ResourceUrn urn;
+
+    public UISkinWithUrn(Map<String, UIStyleFamily> skinFamilies, ResourceUrn urn) {
+        super(skinFamilies);
+        this.urn = urn;
+    }
+
+    public static UISkinWithUrn createFromSkin(UISkin skin, ResourceUrn urn) {
+        Map<String, UIStyleFamily> skinFamilies = new HashMap<>();
+        for (String family : skin.getFamilies()) {
+            skinFamilies.put(family, skin.getFamily(family));
+        }
+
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/engine/rendering/nui/skin/UISkinWithUrn.java
```diff
@@ -0,0 +1,33 @@
...
+package org.terasology.engine.rendering.nui.skin;
+
+import org.terasology.assets.ResourceUrn;
+import org.terasology.nui.skin.UISkin;
+import org.terasology.nui.skin.UIStyleFamily;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class UISkinWithUrn extends UISkin {
```
skaldarnar, line 13: Let's add this as a remark here for future reference.

```suggestion
/**
 * [NOTE] Sub-classing UISkin into UISkinWithUrn is a work-around in order to associate a ResourceUrn with a UISkin.
 * When a UISkin is deserialised, then it is copied into a new UISkinWithUrn instance to associate it with the urn it was deserialised with.
 *
 * (2021-03-31) It is not clear whether this feature is used anywhere (in the Omega module space) right now.
 */
public class UISkinWithUrn extends UISkin {
```

skaldarnar (commented):
I'd like to persist the comment we had on this in the code such that we don't forget about it. 

If we're not using the feature right now, and we're not really sure whether it should be a feature at all, maybe we should not implement it and see if it breaks anywhere instead? 

However, I think this change looks reasonable clean, I'd be fine with merging it when the comment is in.

engine/src/main/java/org/terasology/engine/rendering/nui/skin/UISkinWithUrn.java
BenjaminAmos: I've added the suggested comment, also clarifying which that this refers to deserialisation from `UISkinTypeHandler` (the UISkin is deserialised from JSON by `UISkinFormat` into data for a `UISkinAsset` and the `UISkinTypeHandler` fetches the skin from its asset).

keturn:
I looked more at this area of the code and have a better idea of what it's role is.

I'm still wary of code we don't know how to test, but that's not telling you anything you don't already know.

I'm going to go ahead and merge this, especially as the merge target is just the other PR.

PR #4608, opened 2021-03-31
fix: could place blocks inside the player

tolziplohu:
### Contains

Disallows placing blocks inside the player. The restriction on placing blocks inside colliders was broken, so you could place blocks inside of yourself. This was mentioned briefly in the org meeting as a problem with jumping and placing blocks below you.

It turns out it was due to calling `btCollisionObject.getUserPointer()` to get a collider's attached entity, which returns a `long` which casts implicitly into an `Object` through `Long`, instead of getting `.userData`. The entity data is set with `.userData`, so it should be accessed the same way.

The margins of the bounding box had been messed up at some point as well, making it so you couldn't place blocks near the player, so I put them back to a reasonable position.

### How to test

Try looking down and placing blocks inside your feet - it shouldn't work. You should still be able to place blocks near you: try walking up to a block at waist height and placing one on top of it, or placing a block next to the one you're standing on.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/physics/bullet/BulletPhysics.java
+++ b/engine/src/main/java/org/terasology/engine/physics/bullet/BulletPhysics.java
@@ -183,7 +183,7 @@ public class BulletPhysics implements PhysicsEngine {
         List<EntityRef> result = Lists.newArrayList();
         for (int i = 0; i < scanObject.getNumOverlappingObjects(); ++i) {
             btCollisionObject other = scanObject.getOverlappingObject(i);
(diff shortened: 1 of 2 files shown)


pollend:
@tolziplohu impressive :D, glad that you found it. that was pretty subtle as bugs go.

tolziplohu:
> @tolziplohu impressive :D, glad that you found it. that was pretty subtle as bugs go.

Thank you!

pollend (approved):
Just tested this and it seems good to me :D

PR #4609, opened 2021-03-31
feat: allow repeated jumping by key press

mahela97:
### Contains

Previously when a player presses the jump button, the character only jumps once. With this PR, if a player presses once and keeps holding the jump button then the character also keeps jumping. Since Terasology is based on blocks, the character needs to jump frequently, and with this feature, the gameplay experience will improve.

### How to test

Just create a game and keep press and hold the jump button. (Space in PC)


engine/src/main/java/org/terasology/engine/logic/characters/CharacterMoveInputEvent.java
```diff
@@ -80,8 +80,8 @@ public boolean isCrouching() {
         return crouching;
     }
 
-    public boolean isJumpRequested() {
-        return jumpRequested;
+    public boolean isJumping() {
+        return jumping;
```
skaldarnar, line 84: This changes public API and might break modules, for instance. Did you check the Omega workspace that this still works?

skaldarnar (approved):
I did compile and run an Omega workspace, tests out fine.

engine/src/main/java/org/terasology/engine/logic/characters/CharacterMoveInputEvent.java
mahela97: I checked whether any code I changed affects any modules in and it didn't.

PR #4610, opened 2021-04-02
chore: remove DebugPropertySystem

pollend:
Me and @keturn stumbled on this when looking at gestalt v7. looks like some tool to expand an object and look at properties but I don't see this referenced anywhere from my omega workspace so from what I can tell this is pretty much unused in its current state. I don't know what this would be used for exactly. 

The last major change to this system was on 2015 so I guess we just drop this: https://github.com/MovingBlocks/Terasology/commits/66e5f5d86804a849381e3abd8dfe79713e24cc2a/engine/src/main/java/org/terasology/logic/debug/DebugPropertiesSystem.java


keturn:
I guess this provided a pane that was empty by default, but while debugging code you could put some `addProperty` calls in the code you're troubleshooting to show values for that thing?

kinda like print-debugging, but if the values update in real time without causing log spam, I can imagine that would be useful.

and presumably toggling the mouse capturing is to provide a way to scroll or otherwise interact with that panel

engine/src/main/java/org/terasology/engine/logic/players/DebugControlSystem.java
```diff
@@ -126,13 +118,6 @@ public void onKeyDown(KeyDownEvent event, EntityRef entity) {
         }
 
         switch (event.getKey().getId()) {
-            case Keyboard.KeyId.F11:
```
skaldarnar, line 129: Interestingly, the comment in the DebugProperties states that it is accessible via <kbd>F1</kbd> - probably just adds to the confusion, though...

skaldarnar (approved):
I'm fine with removing unused code :+1:

PR #4611, opened 2021-04-02
build: add tasks for 'unit' and 'integration' tests

### LINKED ISSUE #4606: test: separate integration tests to their own gradle task
Some tests are slow and flaky and not safe to run in parallel. We've [added tags](https://github.com/MovingBlocks/Terasology/pull/4605) to these tests and would like to use them to define a separate test run.

We have `TteTest` and `MteTest` tags for these tests. The remainder of the tests (hopefully less slow and flaky) remain untagged. We will be treating `TteTest` and `MteTest` as equivalent for this new test task.

We need one gradle task (possibly the default `test` task?) that **excludes** tests with either of these tags,  
and one gradle task that **includes** _only_ tests with either of these tags.

This needs to be done for
- [ ] `engine` and/or `engine-test` projects in the main repo
- [ ] modules, which are configured by `terasology-module` under `build-logic`

`engine-tests/build.gradle` is probably the easiest place to start.

### References

Terasology tests are written using JUnit 5 and run with the JUnit Platform.

* [Tagging and Filtering—JUnit 5 User Guide](https://junit.org/junit5/docs/current/user-guide/#writing-tests-tagging-and-filtering)
* [Testing in Java Projects: Test Grouping—Gradle 6.8](https://docs.gradle.org/6.8.2/userguide/java_testing.html#test_grouping)

## Naming Questions

- Should we make new tasks for "tests with MTE" and "tests without MTE", and keep `test` in its current behavior of running all JUnit tests? **Or** should we change the default `test` task to be the "without MTE" one?
- What should the "tests with MTE" task be named? Could be less specific than "MTE" to include other implementations in this category. `integrationTest`, `simulationTest`, `flakyTest`?

skaldarnar:
This PR adds two new tasks for both modules and `engine-tests` to run only either unit tests (`gradlew unitTest`) or integration tests (`gradlew integrationTest`).
To run all tests regardless of any tags use the default command `gradlew test`.

The descriptions of all tasks have been adjusted, and the new tasks were added to the "Verification" group (see `gradlew :modules:tasks`).

Resolves #4606


keturn:
@mahela97 Have you had a chance to test this out?

Does it behave as you expect?

Is there anything unclear about how it's written?

engine-tests/build.gradle
```diff
@@ -80,8 +80,40 @@ task copyResourcesToClasses(type:Copy) {
 }
 
 //TODO: Remove it  when gestalt will can to handle ProtectionDomain without classes (Resources)
-test.dependsOn copyResourcesToClasses
-test.dependsOn rootProject.extractNatives
+test {
+    dependsOn copyResourcesToClasses
+    dependsOn rootProject.extractNatives
```
keturn, line 85: I could _not_ figure out what that comment was referring to last time I looked at this PR, but I get it now. It's specifically that `dependsOn(copyResourcesToClasses)`.

Suggest moving the comment inside the braces to make that clearer.

keturn (changes requested):

mahela97:
> @mahela97 Have you had a chance to test this out?
> 
> Does it behave as you expect?
> 
> Is there anything unclear about how it's written?

Yes, I tested it. There was one test case that failed because of gson but it happened because there is a problem with gson library in my environment. So I think it is fine.

keturn (approved):

keturn:
Looks like it's working to me!

From these initial test runs, I expect we'll probably end up re-tagging some things, as the "fast" version of the omega module tests is far from _fast_ right now. We'll figure out those details as we go; this sets us up with a good starting place.

PR #4614, opened 2021-04-04
Allow entities to be rendered like chunks

4Denthusiast:
This PR adds the `ChunkMeshComponent` class, and the `ChunkMeshRenderer` system which passes entities with that component (and a `LocationComponent`) to the rendering engine as though they were chunks. This means that all the rendering effects that are supported for blocks, like shadows, transparency and all the fancy water effects can be used for entities too. The main intended application is for entities that are meant to look like they're made out of blocks, like the falling block group entities from FallingBlocks.

This PR also contains a few other changes intended to make it easier for systems other than the usual one to produce `ChunkMesh`es, and in particular, to allow the associated FallingBlocks PR, https://github.com/Terasology/FallingBlocks/pull/9, to work.

- The lighting data is incorporated into the chunk mesh at the `BlockMeshGenerator` stage, rather than `generateOptimisedBuffers`, so that the mesh can be more easily manipulated before the data is put in a buffer to be sent to the GPU.
- The methods to get `ChunkView`s from a `WorldProvider` have been rearranged, and a more flexible variant added, so that `BlockMeshGenerator`s, which require `ChunkView`s, can be used to make meshes of arbitrary patches of the world.
- The strength of gravity is exposed as a constant, which isn't particularly relevant to this PR, except that the FallingBlocks PR needs it.

Also, in order to make `ChunkMesh` serializable, the options are added to not discard the copy of the `ChunkMesh`'s data that's stored in main memory (as opposed to GPU memory), and to construct a `ChunkMesh` without using a `BufferPool`, simply getting buffers directly from LWJGL instead.

This PR should have basically no effect by itself, as it just provides a system that can be used elsewhere, so the easiest way to test it (apart from just testing that it doesn't break anything) is testing https://github.com/Terasology/FallingBlocks/pull/9.


engine/src/main/java/org/terasology/engine/rendering/world/RenderableWorldImpl.java
```diff
@@ -75,11 +77,11 @@
     private int statIgnoredPhases;
 
 
-    RenderableWorldImpl(Context context, GLBufferPool bufferPool, Camera playerCamera) {
```
pollend, line 78: generating a VBO Id is a trivial cost and the expensive comes when you try to pass data to the buffer. the extra book keeping in maintaining a pool is probably more expensive then just disposing of the buffer.

engine/src/main/java/org/terasology/engine/world/block/shapes/BlockMeshPart.java
```diff
@@ -115,4 +124,103 @@ public BlockMeshPart rotate(Quaternionf rotation) {
 
         return new BlockMeshPart(newVertices, newNormals, texCoords, indices, texFrames);
     }
+
+    private float[] calcLightingValuesForVertexPos(ChunkView chunkView, Vector3f vertexPos, Vector3f normal) {
```
pollend, line 128: whats your reasoning for moving this outside of Chunktessellator?

engine/src/main/java/org/terasology/engine/persistence/typeHandling/extensionTypes/ChunkMeshTypeHandler.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    protected PersistedData serializeNonNull(ChunkMesh value, PersistedDataSerializer serializer) {
+        if (!value.hasVertexElements()) {
+            throw new IllegalStateException("Attempting to serialize a ChunkMesh whose data has already been discarded.");
+        }
+        List<PersistedData> data = new ArrayList<>();
+        for (ChunkMesh.RenderType renderType : ChunkMesh.RenderType.values()) {
+            data.add(serializer.serialize(asByteBuffer(value.getVertexElements(renderType).finalVertices)));
+            data.add(serializer.serialize(asByteBuffer(value.getVertexElements(renderType).finalIndices)));
+        }
```
pollend, line 30: this is quite a bit to serialize. any benefits with just encoding the block id?

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
```diff
@@ -155,8 +156,8 @@ private void generateVBO(RenderType type) {
         VertexElements elements = vertexElements.get(type);
         int id = type.getIndex();
         if (!disposed && elements.finalIndices.limit() > 0 && elements.finalVertices.limit() > 0) {
-            vertexBuffers[id] = bufferPool.get("chunkMesh");
-            idxBuffers[id] = bufferPool.get("chunkMesh");
+            vertexBuffers[id] = bufferPool != null ? bufferPool.get("chunkMesh") : GL15.glGenBuffers();
+            idxBuffers[id] = bufferPool != null ? bufferPool.get("chunkMesh") : GL15.glGenBuffers();
```
pollend, line 160: I don't think we should bother pooling vbo ids. the cost of generating new id's is not that expensive. We end up introducing more traffic to the GPU by populate and depopulate a buffer.  The cost itself occurs when you try write data to a VBO. The commit that introduced this did this to avoid a crash but that is not that case anymore.

pollend (changes requested):

engine/src/main/java/org/terasology/engine/world/block/shapes/BlockMeshPart.java
4Denthusiast: In the FallingBlocks PR, the system generating those meshes first uses the `BlockMeshGenerator`s to generate the mesh data, then edits the data to translate the mesh from world coordinates to local coordinates, then calls `generateOptimisedBuffers`, so if the lighting data were still calculated using the positions available during the `generateOptimisedBuffers` step, it would have been incorrect. This way, it's added during the `BlockMeshGenerator` step, so editing it afterwards works fine.

engine/src/main/java/org/terasology/engine/persistence/typeHandling/extensionTypes/ChunkMeshTypeHandler.java
4Denthusiast: The mesh that a block generates depends on the context in which it occurs: all the adjacent blocks and all their lighting data usually, and sometimes the extraData fields too. In theory it could depend on the block entities too, although I don't know of any examples where that currently happens. In order to reconstruct the mesh, it would be necessary to put all that data into a `ChunkView` in order to re-generate the mesh, which would be pretty complicated.

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
4Denthusiast: I'm fine with getting rid of `BufferPool`. Do you think I should do that in this PR? It would simplify some of it a little.

pollend: depends on how much you have to change. if its easy and simple to do then I think it would make sense here.

pollend (dismissed):

pollend (dismissed):

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
pollend: I think we can do this after in a second PR, might make sense to remove bufferPool in one go.

engine/src/main/java/org/terasology/engine/persistence/typeHandling/mathTypes/AABBfTypeHandler.java
```diff
@@ -33,7 +33,7 @@ protected PersistedData serializeNonNull(AABBf value, PersistedDataSerializer se
             PersistedDataMap map = data.getAsValueMap();
 
             PersistedDataArray minDataArr = map.get(MIN_FIELD).getAsArray();
-            PersistedDataArray maxDataArr = map.get(MIN_FIELD).getAsArray();
+            PersistedDataArray maxDataArr = map.get(MAX_FIELD).getAsArray();
```
pollend, line 36: looks like you found a serialization bug  :+1:

pollend (approved):

jdrueckert:
[Improve LaS team balancing](https://trello.com/c/kawWZtVY/65-improve-las-team-balancing)

PR #4615, opened 2021-04-04
chore: refactor physics voxel world system

pollend:
I noticed @DarkWeird introduced some changes to VoxelWorld so I tweaked it and reintroduce this change: https://github.com/MovingBlocks/Terasology/pull/4502/files . I combine both system and and reduced some of the duplicated logic in both VoxelWorld and VoxelFluidWorld.


engine/src/main/java/org/terasology/engine/physics/bullet/VoxelWorldSystem.java
```diff
@@ -0,0 +1,96 @@
...
+    public void onNewChunk(OnChunkLoaded chunkAvailable, EntityRef worldEntity) {
+        Vector3ic chunkPos = chunkAvailable.getChunkPos();
+        Chunk chunk = chunkProvider.getChunk(chunkPos);
+        ByteBuffer buffer =
+                ByteBuffer.allocateDirect(2 * (Chunks.SIZE_X * Chunks.SIZE_Y * Chunks.SIZE_Z));
+        buffer.order(ByteOrder.nativeOrder());
+        for (int z = 0; z < Chunks.SIZE_Z; z++) {
+            for (int x = 0; x < Chunks.SIZE_X; x++) {
+                for (int y = 0; y < Chunks.SIZE_Y; y++) {
+                    Block block = chunk.getBlock(x, y, z);
```
DarkWeird, line 87: nitpick(non-blocking): I did provide additional method for getting blockId directly... Physics needs to use Block entity at registerBlock only... this is unnecessary work at 99% of time (getting block data from BlockManager is expensive operation, because TShortObjectMap)

engine/src/main/java/org/terasology/engine/physics/bullet/VoxelWorldSystem.java
```diff
@@ -0,0 +1,96 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+package org.terasology.engine.physics.bullet;
+
+import com.google.api.client.util.Lists;
```
DarkWeird, line 5:
```suggestion
import com.google.common.collect.Lists;
```

DarkWeird (changes requested):
Needs small import fix.

Thx for extract this code from old PR !

engine/src/main/java/org/terasology/engine/physics/bullet/VoxelWorldSystem.java
pollend: I don't think we have a method like that. would @4Denthusiast have a clue since he recently refactored Chunk.

DarkWeird (approved):

engine/src/main/java/org/terasology/engine/physics/bullet/VoxelWorldSystem.java
DarkWeird: I introduced it in my PR.
But not-blocking :)

PR #4616, opened 2021-04-07
chore: add primative mesh producer in-place of OpenGLUtils#renderFullscreenQuad

pollend:
I defined a primitive that can be used for rendering a quad to the screen for post processing effects. This is preferable then the current scheme which uses matrix stacks and call list which are deprecated in opengl. I would like to start updating the shader logic and remove:

- ftransform
- glBegin
- glTexCoord2d
- glVertex3i
- glEnd
- glEndList
- glMatrixMode

This change will only influence the FinalPostProcessing node so this should break everything if the node isn't rendering correctly. This drops renderFullscreenQuad() which uses display list. 

-------

- https://github.com/Terasology/CoreRendering/pull/39

we also need a way to define layout qualifiers for meshes: https://www.khronos.org/opengl/wiki/Layout_Qualifier_(GLSL)


engine/src/main/resources/assets/shaders/post_vert.glsl
```diff
@@ -16,7 +16,7 @@
 
 void main()
 {
-	gl_Position = ftransform();
+	gl_Position = gl_Vertex;
+//    gl_FrontColor = gl_Color;
```
DarkWeird, line 20: Commented code?
What about Color now?

DarkWeird (commented):
Can you add some description?

pollend:
> Can you add some description?

It's a draft pr still looking at a couple things.

engine/src/main/resources/assets/shaders/post_vert.glsl
pollend: It's not used, and it doesn't make sense why you need to do this  for a screen buffer affect

pollend:
[Add primitive mesh producer for a quad](https://trello.com/c/gQLwR57g/112-add-primitive-mesh-producer-for-a-quad)

DarkWeird (approved):
Looks nice.
Especially concept: get named asset - draw asset

future related problem: `Assets` uses `CoreRegistry`
current related problem: GLSL330 without GL32's profiles not works at Linux(intel) and Macos(seems). windows should be ok with this. other non-free drivers have (forward/backward) compatibility. free drivers - idk

PR #4619, opened 2021-04-09
fix: avoid writing window size of 0 to the config

mahela97:
Fix according to @keturn suggestion. Sorry I had to made a separate PR because I accidentally deleted the previous branch.


pollend:
just trying to understand why we need to restore where the window was placed on the screen. 

do you still need these PRs:

- https://github.com/MovingBlocks/Terasology/pull/4603
- https://github.com/MovingBlocks/Terasology/pull/4601

keturn:
This is a change to the shutdown method, so I think it replaces #4603 but not the one that changes the initialise method.

The code looks good to me. We'll need someone else to test it because I wasn't able to reproduce this under Ubuntu.

mahela97:
> just trying to understand why we need to restore where the window was placed on the screen. what does GLFW do by default if you don't specify where the window is placed? is that up to the windows manager to drive that or is it by default placed at some fixed value like (0,0).
> 
> do you still need these PRs:
> 
> * #4603
> * #4601

Yes if we don't restore where the window was placed, it will always start with default values in #4160. 

No those PRs won't be ne necessary anymore. Shall I remove those?

mahela97:
> This is a change to the shutdown method, so I think it replaces #4603 but not the one that changes the initialise method.
> 
> The code looks good to me. We'll need someone else to test it because I wasn't able to reproduce this under Ubuntu.

Okay :)

Cervator (approved):
Was able to test this on Windows 10, which it may be specific to due to the preview-via-task-bar feature allowing you to trigger the normal closing of a window that's technically minimized.

Could replicate before this PR and confirm it fixed after! Thank you @mahela97 👍 

Going to let this branch update for final CI testing then anybody can squash-merge after

PR #4621, opened 2021-04-10
feat: Add BlockAreac#distanceSquared

shadowasphodel2919:
### Contains

https://github.com/Terasology/CoreWorlds/issues/27#issue-802749528
Added the `distanceSquared()` function to the `BlockArea` file.
-> Implemented it by creating a calculate distance function


engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
```diff
@@ -354,4 +355,35 @@ public int hashCode() {
...
+        }else{
+            if(py < minY) {
+                return calDis(maxX - px, minY - py);
+            }
+            if(py <= maxY) {
+                return px - maxX;
+            }
+            return calDis(maxX - px, maxY - py);
+        }
+    }
```
pollend, line 388: do we want to use block locations or the closes edge of the area. @skaldarnar 

I'm assuming this is more useful for the actual bounds of the blockarea not just the location of the blocks. 

block area is defined as (minx - .5f ,miny - .5f ) and (minx + .5f ,miny + .5f )

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
```diff
@@ -354,4 +355,35 @@ public int hashCode() {
     public String toString() {
         return "BlockArea[(" + this.minX + ", " + this.minY + ")...(" + this.maxX + ", " + this.maxY + ")]";
     }
+
+    private int calDis(int x, int y){
+        return (int) Math.sqrt(x * x + y * y);
+    }
```
pollend, line 361: I would opt to just inline this rather then have a separate method.

pollend (changes requested):

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
```diff
@@ -354,4 +355,35 @@ public int hashCode() {
...
+            return py - maxY;
+        }else{
+            if(py < minY) {
+                return calDis(maxX - px, minY - py);
+            }
+            if(py <= maxY) {
+                return px - maxX;
+            }
+            return calDis(maxX - px, maxY - py);
+        }
```
pollend, line 387: this logic doesn't look quite right to me. 

here is what is defined in the current Rectanglei class. the current logic in Rectanglei is pretty concise and it works well. 

```
        //     center = min + width / 2 = min + (max - min) / 2
        // 2 * center = 2 * min + max - min = min + max
        int cx2 = minX() + maxX();
        int cy2 = minY() + maxY();

        //     dx = p - center - width / 2
        // 2 * dx = 2 * px - 2 * center - width
        //     dx = (2 * px - 2 * center - width) / 2
        long dx = Math.max(Math.abs(2 * px - cx2) - getSizeX(), 0) / 2;
        long dy = Math.max(Math.abs(2 * py - cy2) - getSizeY(), 0) / 2;
        return dx * dx + dy * dy;

```

pollend (changes requested):

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
```diff
@@ -354,4 +355,35 @@ public int hashCode() {
     public String toString() {
         return "BlockArea[(" + this.minX + ", " + this.minY + ")...(" + this.maxX + ", " + this.maxY + ")]";
     }
+
+    private int calDis(int x, int y){
+        return (int) Math.sqrt(x * x + y * y);
+    }
+    public int distanceSquared(int px, int py){
```
pollend, line 362: I would also add some unit test to validate if this is working as intended.

pollend (changes requested):

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
shadowasphodel2919: Yes, I shall be making the required changes.

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
skaldarnar: Good questions. I guess that we want to return a distance of 0 in case the given point is "touching" the area.

```java
new BlockArea(0, 0, 2, 2).distanceSquared(2, 3) == 0;
new BlockArea(0, 0, 2, 2).distanceSquared(3, 2) == 0;
```

What about if the location "touches" only at the corner? 

```java
new BlockArea(0, 0, 2, 2).distanceSquared(3, 3) == ???;
```

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
skaldarnar: Agreed, see https://github.com/MovingBlocks/Terasology/pull/4621#discussion_r622032468 above.

skaldarnar:
@shadowasphodel2919 thanks for updating the PR. 

Please add a few test cases to https://github.com/MovingBlocks/Terasology/blob/develop/engine-tests/src/test/java/org/terasology/world/block/BlockAreaTest.java to ensure everything is working as expected.

Once that is done, we should add the dual method for BlockRegion in a follow up PR to keep the APIs consistent.

shadowasphodel2919:
@skaldarnar yes I am working upon it.

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
```diff
@@ -354,4 +355,24 @@ public int hashCode() {
...
+        int cx2 = minX() + maxX();
+        int cy2 = minY() + maxY();
+
+        //     dx = p - center - width / 2
+        // 2 * dx = 2 * px - 2 * center - width
+        //     dx = (2 * px - 2 * center - width) / 2
+        long dx = Math.max(Math.abs(2 * px - cx2) - getSizeX(), 0) / 2;
+        long dy = Math.max(Math.abs(2 * py - cy2) - getSizeY(), 0) / 2;
+        return dx * dx + dy * dy;
+    }
```
pollend, line 377: this looks good sorry, for the late response. so this is the square distance from the center of a block to another center. I guess do we want it to be calculated for the edge instead?

@skaldarnar

skaldarnar: Yes, this should be available for the immutable interface in `BlockAreac` as well.

shadowasphodel2919:
I have added the test cases for both the functions along with the changes to BlockAreac.

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
```diff
@@ -354,4 +355,24 @@ public int hashCode() {
...
+        // 2 * center  = 2 * min + max - min = min + max
+        int cx2 = minX() + maxX();
+        int cy2 = minY() + maxY();
+
+        //     dx = p - center - width / 2
+        // 2 * dx = 2 * px - 2 * center - width
+        //     dx = (2 * px - 2 * center - width) / 2
+        long dx = Math.max(Math.abs(2 * px - cx2) - getSizeX(), 0) / 2;
+        long dy = Math.max(Math.abs(2 * py - cy2) - getSizeY(), 0) / 2;
+        return dx * dx + dy * dy;
```
pollend, line 376: could we move these to the edges of the bounding box. maybe it would be more useful to use floating point instead of int px and int py. just need that change and I will approve this.

engine-tests/src/test/java/org/terasology/world/block/BlockAreaTest.java
```diff
@@ -130,4 +131,30 @@ public void testIntersectsArea(BlockArea other, boolean intersecting) {
...
+    @MethodSource("testDistanceSquaredArgs")
+    public void testDistanceSquared(BlockArea area, int x, int y, int val){
+        assertEquals(val,area.distanceSquared(x,y));
+    }
+
+    @Test
+    public void testDistanceSquaredVector(){
+        BlockArea area = new BlockArea(0, 0, 3, 3);
+        assertEquals(8,area.distanceSquared(new Vector2i(-3, -3)));
+    }
```
pollend, line 159: really nice, I like it.

pollend (changes requested):

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
shadowasphodel2919: I did create a test version for float valued arguments and along with changing the return type, there is a considerable difference in the values returned by it. The code is ready to be pushed along with changed test values but would it not affect the results where this function is used? The benefit to calculating distance this way is that this function returns 0 for all the values touching the block.

engine-tests/src/test/java/org/terasology/world/block/BlockAreaTest.java
shadowasphodel2919: Thanks :)

engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
pollend: min[X,Y,Z](), max[X,Y,Z]() are the center of blocks. but the actual boundaries are -.5 and +.5. umm can't look at the code right now to point at an example but there should be some code that shows this behaviour.

jdrueckert:
@skaldarnar @pollend are there any open issues from your side with this or can we resolve the conflicts and try to get this merged?

skaldarnar (approved):

jdrueckert (approved):

PR #4593, opened 2021-03-21
feature: migrate to gestaltv7

pollend:
Here is the initial migration to gestalt-v7.

## Changes

- update gestalt package
  - org.terasology.gestalt.assets.management.AssetManager
  - org.terasology.gestalt.naming.Name
  - org.terasology.gestalt.assets.AssetType;
  - org.terasology.gestalt.assets.management.AssetManager;
  - org.terasology.gestalt.assets.module.ModuleAwareAssetTypeManager;
  - org.terasology.gestalt.assets.module.ModuleAwareAssetTypeManagerImpl;
  - org.terasology.gestalt.module.ModuleEnvironment;
  - org.terasology.gestalt.module.ModuleRegistry;
  - org.terasology.gestalt.module.dependencyresolution.DependencyResolver;
  - org.terasology.gestalt.module.dependencyresolution.ResolutionResult;

## Issues

- [ ] something jmh-related preventing running from IntelliJ?
- [ ] tidy use of `google()` repository
- [x] #4610 Crash on F11 (DebugControlSystem failure to get MouseDevice) 
- [ ] migrate the modules: many have branches under the name `chore/gestalt-v7-migration`

----
## Module Changes

- [ ] https://github.com/Terasology/CoreRendering/pull/36

Issue #4509, opened 2021-02-13
Gestalt V7 Migration


pollend:
# Gestalt v7 Integration
* Changing engine module to package module. ( classpath modules dropped)
 * engine moved in self package (`org.terasology.engine` now)
 * create syntetic package module  for nui (easier then write all in whitelist)
 * Update gestalt to 7.0.5
 * Update gson to 2.8.5
 * Update TeraNui to 2.0.0
 * Update Guava to 29.0-jre
 * Update Reflections to org.terasology:reflections:0.9.12-MB (gestalt build it)
 * Reimplemented Module download (gestalt drop `BaseModule` class)
 * Restore Module loading on gestalt v7
 * Use AutoreloadingAssetManager (seems asset must reload easier now)
 * Replace condition `Module#isSourceModule` to checking Module's files source == `DirectoryFileSource` - must be equal ( affected WorldGeneratorManager)
 * add Gestalt's classes to whitelist
 * remove duplication classes from `org.terasology.input`(`org.terasology.engine.input` now). We have this classes in TeraNUI
 * move `cacheReflections` task on after `classes` task (partial resolve https://github.com/MovingBlocks/Terasology/issues/4126#issuecomment-692232675) (reflections gathering for running from sources now)
 * remove `cleanReflections` after `jar` task
 * renamed `cacheReflections`'s artifact `reflections.cache` to `manifest.json` and serializer changed to Json for using gestaltv7's default values

Relevant PR for reference:
- [Breaking all] Gestalt v7 re-integration (https://github.com/MovingBlocks/Terasology/pull/4144)

keturn:
The big list currently in this issue description is from #4144. There are some things that were in #4144 that could be broken out and done independently, including:

* classpath modules dropped: see #4543
* removing shrinkwrap from engine-tests: reference https://github.com/MovingBlocks/Terasology/pull/4144/commits/eef934f08069b1afd08c5b64f7efc001cdd51448
* update guava: reference https://github.com/MovingBlocks/Terasology/pull/4144/commits/fec847ae3ecbbac09b611b386068e9c28ed356d2
* update gson: straightforward upgrade.
* remove duplication classes from org.terasology.input. We have this classes in TeraNUI: has this been done already?
* move `cacheReflections` task on after classes task: can be done separately.

jdrueckert:
Migration effort done (except for a few follow-ups tracked in https://github.com/orgs/MovingBlocks/projects/20

PR #4622, opened 2021-04-10
chore: migrate to gestalt v7—fork 🅰, keep subsystems separate

keturn:
This is a variant of #4593. It diverges after 20546f1853a8f5cce99fff8018700d81340181d3 in an attempt to find an alternative to moving subsystems back in to the `org.terasology.engine` namespace.

The diff is usable if you only include the later commits, i.e. _after_ all the import changes. https://github.com/MovingBlocks/Terasology/pull/4622/files/94e1e419d3d9cff599a2ec4e7bd68cdf4983ebc3..HEAD  
but this trick is working less well the longer this branch is around, as that diff includes merge commits in a way that diffs from the branch root do not.

To generate a diff from the command line that excludes the `import` clutter:

```sh
git diff --ignore-matching-lines="^import .*"  --ignore-all-space \
    origin/develop origin/feature/migrate-gestalt-v7.a 
```

## Build Instructions

Use `chore/gestalt-v7-migration` branches for `/modules/*`.

## Notes for Reviewers

- The place most related to gestalt-module is the ModuleManager: https://github.com/MovingBlocks/Terasology/blob/feature/migrate-gestalt-v7.a/engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java 
    but there are also changes spread across asset loaders and type handlers.
- Does anyone think it is critical that we keep building reflections.cache at compile time? In play testing, do you notice this being a lot slower to start or join a game?
- Do we need to fix anything about the way we're handling NUI or subsystems? These are libraries that are outside of engine, but define things that need to be discoverable to the ECS or TypeRegistry systems. i.e. they have `@API` annotations or have classes involved in some `getSubtypes` query.
  - is #4593's approach better?

## To Do

- [ ] reinstate cache-reflections build tasks?
- [x] Merge gestalt fix for asset manager race condition: https://github.com/MovingBlocks/gestalt/pull/102
- [x] something jmh-related preventing running from IntelliJ?
- [x] tidy use of `google()` repository
- [x] migrate the modules: many have branches under the name `chore/gestalt-v7-migration`


keturn:
This isn't to the point where it does anything interesting, but it has some gradle chores I wanted to make sure were pushed somewhere.

engine/src/main/java/org/terasology/engine/core/subsystem/lwjgl/LwjglGraphics.java
```diff
@@ -70,6 +58,16 @@ public void registerCoreAssetTypes(ModuleAwareAssetTypeManager assetTypeManager)
...
     }
 
+    @Override
+    public void preInitialise(Context context) {
+        super.preInitialise(context);
+        GameEngine engine = context.get(GameEngine.class);
+
+        // Make sure NUI classes are registered as part of the engine module.
+        // FIXME: kludgy use of method outside of Interface
+        ((TerasologyEngine) engine).addToClassesOnClasspathsToAddToEngine(CanvasRenderer.class);
```
keturn, line 68: FIXME: add this method to GameEngine interface, or figure out alternate interface.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -111,23 +113,35 @@ private void loadModulesFromApplicationPath(PathManager pathManager) {
...
 
+        // We need the class predicate to include classes in subsystems and whatnot. We can't change it in an
+        // existing module, so make a new one based on the one from the moduleFactory.
+        Module engine = new Module(
+                packageModule.getMetadata(),
+                packageModule.getResources(),
+                Collections.emptyList(),
+                packageReflections,
+                packageModule.getClassPredicate().or(clazz -> clazz.getClassLoader().equals(engineClassLoader))
+        );
```
keturn, line 142: This whole method is essentially doing gestalt ModuleFactory stuff. A sign that we might want to add a new method there or expose some of its private methods.

engine/src/main/java/org/terasology/engine/core/subsystem/lwjgl/LwjglGraphics.java
keturn: I'm calling this "add a NUI class to the engine module's classpath" thing from the LwjglGraphics subsystem.

My thinking was that it doesn't necessarily need to happen for headless mode, and LwjglGraphics is a thing you get when not in headless mode?

Lwjgl mostly doesn't depend on NUI, except it does already involve this CanvasRenderer, so it's not a new dependency.

keturn: I also think it's weird for the subsystem to push this information to engine like this.

Engine already adds the subsystem classes themselves; maybe Subsystem should have some interface by which it can specify additional classes engine can pull in.

keturn (commented):
This runs single-player Sample Gameplay now.

It has the same two test failures in engine-test we've been seeing in this branch all along.

keturn:
The diff is usable if you only include the later commits, i.e. _after_ all the import changes. https://github.com/MovingBlocks/Terasology/pull/4622/files/94e1e419d3d9cff599a2ec4e7bd68cdf4983ebc3..HEAD

keturn:
Oops. I was wrong when I said we only had the same two test failures. In addition, we have

* [nui.editor.ContextMenuUtilsTest.testNodeTypes](http://jenkins.terasology.io/teraorg/job/Terasology/job/engine/job/PR-4622/2/testReport/org.terasology.rendering.nui.editor/ContextMenuUtilsTest/testNodeTypes__/). Not surprising given we've been messing with NUI and reflections.
* VectorTypeSerializerTest.testSerializationConstant](http://jenkins.terasology.io/teraorg/job/Terasology/job/engine/job/PR-4622/2/testReport/org.terasology.engine.persistence.serializers/VectorTypeSerializerTest/testSerializationConstant__/). Not so expected. Needs a closer look.

https://github.com/MovingBlocks/Terasology/blob/26c4457d88364b2983f0504995e969728d597166/engine-tests/src/test/java/org/terasology/persistence/serializers/VectorTypeSerializerTest.java#L61-L71

and the log says

> WARN  o.t.p.t.c.RuntimeDelegatingTypeHandler - Cannot find subtype 'org.joml.Vector3f' to deserialize as, deserializing as base type 'interface org.joml.Vector3fc'
> ERROR o.t.p.t.c.ObjectFieldMapTypeHandler - Could not deserialize field v1

🤔

keturn:
> WARN o.t.p.t.c.RuntimeDelegatingTypeHandler - Cannot find subtype 'org.joml.Vector3f' to deserialize as, deserializing as base type 'interface org.joml.Vector3fc'

This is the part caused by this branch.

* The field's type is `Vector3fc`.
* The type according to the serialized data is `Vector3f`.
* RuntimeDelegatingTypeHandler looks for a subtype of `Vector3fc` named `Vector3f` within its ModuleEnvironmentSandbox.
  - none found; its subtypes scanner knows of no subtypes of Vector3fc

> ERROR o.t.p.t.c.ObjectFieldMapTypeHandler - Could not deserialize field v1

This is probably a bug in RuntimeDelegatingTypeHandler? The data is

```json
{"class":"org.joml.Vector3f","content":[1.0,2.0,3.0]}
```

and RDTH passes the whole thing to the deserializer, not just its VALUE_FIELD.

The deserializer looks at that https://github.com/MovingBlocks/Terasology/blob/7143a6c10bd09cd48a96d39538d2cabb99c6719b/engine/src/main/java/org/terasology/engine/persistence/typeHandling/mathTypes/Vector3fcTypeHandler.java#L23-L4

and goes "wtf? this is not an Array, I will throw it away and not tell you what is wrong until you get elbows-deep in the debugger."

keturn:
I've churned through several attempts in my local workspace... but I'm not at the point where they've identifiably _fixed_ the type handler thing, so I haven't sorted out if any of it is worth committing yet.

There are several layers of _stuff_ that get in the way of finding classes:

* the sandbox permissions https://github.com/MovingBlocks/Terasology/blob/26c4457d88364b2983f0504995e969728d597166/engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java#L167-L171
* [ModuleEnvironment.getModuleProviding](https://github.com/MovingBlocks/gestalt/blob/a51ba072f2e7c8e084ddf0b530d73bbb713ae2a3/gestalt-module/src/main/java/org/terasology/gestalt/module/ModuleEnvironment.java#L260) tests the class against each module's "class predicate."
* other things, particularly anything accessed through a "subtypesOf" filter, depend on the types existing in some Reflections manifest.

and it seems very easy for those things to not be aligned. The specific example that test failure hits is `org.joml.Vector3fc` & `org.joml.Vector3f`. 

The TypeHandlerLibrary makes sure they have handlers registered: https://github.com/MovingBlocks/Terasology/blob/26c4457d88364b2983f0504995e969728d597166/engine/src/main/java/org/terasology/engine/persistence/typeHandling/TypeHandlerLibraryImpl.java#L121-L122

but RuntimeDelegatingTypeHandler also constrains things to its sandbox.

and who is responsible for making sure that reflections manifests and module-class-predicates are set up for external classes like org.joml?

keturn:
probably going to leave compile-time Reflections-gathering disabled for a bit in this branch, because figuring out how running Reflections in a gradle task differs from running it in the application is bad for my headache.

I did try to factor the three implementations of `cacheReflections` out to a common one. I'm trying not to fall in to the tar pit of making it perfectly well behaved, as I tried to with #4022, instead attempting to just move the code while keeping the same semi-sketchy behavior.

pollend:
the cache stuff that @DarkWeird added is a lot easier to work with then reflection. will have to wait though :/

keturn:
I've traced my troubles with the VectorTypeSerializerTest to TypeRegistry.initalizeReflections: https://github.com/MovingBlocks/TeraNUI/blob/baeddc049fe42f11f7e035784f6e351507a7a3ca/nui-reflect/src/main/java/org/terasology/reflection/TypeRegistry.java#L92-L104

...is it not going to load across repositories? Okay:

```java
       reflections = new Reflections(
            new ConfigurationBuilder()
                .setScanners(
                    new SubTypesScanner(false),
                    new TypeAnnotationsScanner()
                )
                .addClassLoaders(allClassLoaders)
                .addUrls(ClasspathHelper.forClassLoader(
                    allClassLoaders.stream()
                        .filter(classLoaderFilter)
                        .toArray(ClassLoader[]::new)
                ))
                .filterInputsBy(TypeRegistry::filterWhitelistedTypes)
        );
```

@BenjaminAmos Do you happen to know what the intent is for the argument to `addUrls` here? Or is your name on this only because you had the commit that moved it out of the Terasology repo?

I ask because the result of that `ClasspathHelper.forClassLoader()` expression results in **zero** URLs.

If I swap it out for [ClasspathHelper.forJavaClasspath](http://ronmamo.github.io/reflections/org/reflections/util/ClasspathHelper.html#forJavaClassPath()) then that test case is able to find joml classes again.

keturn:
The catch is, that doesn't fix the TypeSerializerTest failure, which is a _very similar_ test failure.

All of these assertions pass:

```java
        Assertions.assertNotEmpty(typeHandlerLibrary.getTypeHandler(Animal.class));
        Assertions.assertNotEmpty(typeHandlerLibrary.getTypeHandler(Dog.class));

        Assertions.assertNotEmpty(typeRegistry.load("org.terasology.engine.persistence.serializers.TypeSerializerTest$Animal"));
        Assertions.assertNotEmpty(typeRegistry.load("org.terasology.engine.persistence.serializers.TypeSerializerTest$Dog"));

        ModuleEnvironment env = moduleManager.getEnvironment();
        Assertions.assertNotEmpty(env.getSubtypesOf(Animal.class));
```

But _this_ is the bit that fails:

```java
        assertNotEmpty(typeRegistry.getSubtypesOf(Animal.class));
```

and I don't know how to make it work

I guess the next step is to go back and find out what makes it pass in `develop`? blerb.

keturn:
I went to see what happens if I add those extra assertions in develop: https://github.com/keturn/Terasology/blob/46dea1e3937a65a115da3467cf0ad4c9207d535e/engine-tests/src/test/java/org/terasology/engine/persistence/serializers/TypeSerializerTest.java#L107-L109

and, uh, that typeRegistry.getSubtypesOf assertion is false, even while the deserialization appears to be successful

so yeah, I still don't know how anything works

DarkWeird:
> I've traced my troubles with the VectorTypeSerializerTest to TypeRegistry.initalizeReflections: https://github.com/MovingBlocks/TeraNUI/blob/baeddc049fe42f11f7e035784f6e351507a7a3ca/nui-reflect/src/main/java/org/terasology/reflection/TypeRegistry.java#L92-L104
> 
> ...is it not going to load across repositories? Okay:
> 
> ```java
>        reflections = new Reflections(
>             new ConfigurationBuilder()
>                 .setScanners(
>                     new SubTypesScanner(false),
>                     new TypeAnnotationsScanner()
>                 )
>                 .addClassLoaders(allClassLoaders)
>                 .addUrls(ClasspathHelper.forClassLoader(
>                     allClassLoaders.stream()
>                         .filter(classLoaderFilter)
>                         .toArray(ClassLoader[]::new)
>                 ))
>                 .filterInputsBy(TypeRegistry::filterWhitelistedTypes)
>         );
> ```
> 
> @BenjaminAmos Do you happen to know what the intent is for the argument to `addUrls` here? Or is your name on this only because you had the commit that moved it out of the Terasology repo?
> 
> I ask because that `ClasspathHelper.forClassLoader()` expression results in **zero** URLs.
> 
> If I swap it out for [ClasspathHelper.forJavaClasspath](http://ronmamo.github.io/reflections/org/reflections/util/ClasspathHelper.html#forJavaClassPath()) then that test case is able to find joml classes again.

It is Reflections with java 11 issue. it cannot load classloaders properly ( reflections try to load Bootstrap classloader, system classloader and context classloader... also several else)

keturn:
> It is Reflections with java 11 issue

but many of us have been using Java 11 for quite a while now ???

DarkWeird:
> > It is Reflections with java 11 issue
> 
> but many of us have been using Java 11 for quite a while now ???

Developers - mostly.
Build happens at Java 8
Players - part at java 8, part at java 9+
Also we should supports both versions now. ( dreamland with android)

We can found some problem with JavaClasspath(I don't sure): 
1. handle more classes then should - classes from subsystem's dependencies (DiscordRPC)
2. handle less classes then should - classes from modules(we load them dynamically)

keturn:
> > but many of us have been using Java 11 for quite a while now ???
> Developers - mostly.

Right, I know that TeraNUI isn't targeting _only_ Java 11, but given that most developers and everyone who is using Launcher has been using Java 11,

and this method returns nothing with Java 11,

then why hasn't this been a problem in unit tests or gameplay events?

If it does nothing, should we remove this initialization method entirely? If not, why not?

> We can found some problem with JavaClasspath(I don't sure):
>
> - handle more classes then should - classes from subsystem's dependencies (DiscordRPC)

Is the `.filterInputsBy(TypeRegistry::filterWhitelistedTypes)` sufficient to take care of that? If not, why not?

> - handle less classes then should - classes from modules(we load them dynamically)

`.filter(classLoaderFilter)` makes sure that no module classloaders are used in that `addUrls` argument.

One of the reasons I am unclear on the intent of this method. What classloader would you be passing in here if it isn't a module classloader?

keturn:
I still think that TypeRegistry.initialiseReflections method is in trouble,  
but I think this change to ModuleTypeRegsitry is sufficient for now: https://github.com/MovingBlocks/TeraNUI/pull/52

With that, I'm down to only one engine-test failure locally. And though I haven't decided how to fix it yet, at least it makes sense.

keturn:
MTE tests are failing on account of not being able to find its "dummy" world generator.

and there's a lot of this sort of thing in its logs:
> ERROR o.t.g.m.s.JavaModuleClassLoader - Denied access to class (not allowed with this module's permissions): org.terasology.moduletestingenvironment.fixtures.DummySystem

I guess having dealt with the TypeRegistry problems, we are now on to ClassLoader problems.

keturn:
I think the place to look is ModuleManager.setupSandbox.

Guesses:
1. APIScanner is busted somehow.
2. maybe the way MTE is registered as a module during its own tests happens _after_ setupSandbox is run, so it never gets APIScanner run on it.

build-logic/src/main/kotlin/reflections-manifest.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+        val classPaths = mainSourceSet.compileClasspath.map { it.toURI().toURL() }
+        val classLoader = URLClassLoader(classPaths.toTypedArray())
+//        try {
+//            val reflections = Reflections(ConfigurationBuilder()
+//                .setSerializer(JsonSerializer())
+//                .addClassLoader(classLoader)
+//                // .filterInputsBy(FilterBuilder.parsePackages("+org"))
+//                .addUrls(inputs.files.map { it.toURI().toURL() })
+//                .setScanners(TypeAnnotationsScanner(), SubTypesScanner(false)))
+//            reflections.save(manifestFile.toString())
```
keturn, line 31: FIXME: decide if we're re-enabling cacheReflections task or not

engine/src/main/java/org/terasology/engine/core/TerasologyEngine.java
```diff
@@ -181,13 +182,16 @@ public TerasologyEngine(TimeSubsystem timeSubsystem, Collection<EngineSubsystem>
...
+        // register NUI classes with engine module
+        addToClassesOnClasspathsToAddToEngine(UIWidget.class);
     }
 
     /**
      * Provide ability to set additional engine classpath locations.   This must be called before initialize() or run().
      * @param clazz any class that appears in the resource location to treat as an engine classpath.
      */
-    protected void addToClassesOnClasspathsToAddToEngine(Class<?> clazz) {
+    public void addToClassesOnClasspathsToAddToEngine(Class<?> clazz) {
```
keturn, line 194: FIXME: can revert this permission change now, as I abandoned my attempt to register NUI only for a certain subsystem.

keturn:
The good news is that there is a failing test case for APIScanner:

https://github.com/MovingBlocks/gestalt/blob/0832d9709bd80ee40590682e04a8aa174f516e06/gestalt-module/src/test/java/org/terasology/gestalt/module/sandbox/APIScannerTest.java#L41

("good" in that "at least we have a test case for it.")

keturn:
The bad news is that I got that test to pass without changing the implementation. Meaning it doesn't fix our problem here.

engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+class ClasspathCompromisingModuleFactory extends ModuleFactory {
+    @Override
+    public Module createDirectoryModule(ModuleMetadata metadata, File directory) {
```
keturn, line 21: TODO: document

engine-tests/src/test/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactoryTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Test
+    public void directoryModuleContainsClass() {
+        ModuleFactory factory = new ClasspathCompromisingModuleFactory();
+
+        File engineTestDirectory = new File(System.getProperty("user.dir", "."));
+
+        ModuleMetadata metadata = new ModuleMetadata(new Name("unittest"), new Version("1.0.0"));
+
+        Module module = factory.createDirectoryModule(metadata, engineTestDirectory);
```
keturn, line 33: TODO: this could use some explanation

keturn:
iota tests pass!!!!!!!!

keturn:
This is the trick that makes MTE tests work: https://github.com/MovingBlocks/Terasology/blob/8aef6cc3efc3cf9d1d5dbeff715a7744517a0eef/engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java#L20-L27

The catch was that `getModuleProviding` for modules usually works based on the classloader: https://github.com/MovingBlocks/gestalt/blob/a51ba072f2e7c8e084ddf0b530d73bbb713ae2a3/gestalt-module/src/main/java/org/terasology/gestalt/module/ModuleEnvironment.java#L260-L263

but tests (and the module classes they import) are loaded with the standard classloader, so the ModuleEnvironment can't tell.

Ideally, we'd have a test runner that loads the tests in a module environment instead of from the classpath. But that's out-of-scope for this branch, so instead I used the "load classpath modules" flag we put in for MTE to change the module's class predicate.

I have [one test](https://github.com/MovingBlocks/Terasology/blob/8aef6cc3efc3cf9d1d5dbeff715a7744517a0eef/engine-tests/src/test/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactoryTest.java#L26) for it. It probably _should_ have a few additional tests, but writing test cases for classpath-dependent things is a mess.

keturn:
[[GESTALTv7] migrate module & asset](https://trello.com/c/te4uIhaz/46-gestaltv7-migrate-module-asset)

keturn:
Running omega tests on this today. I got a build to finish. Gradle's build scan server probably hates me: https://scans.gradle.com/s/rufyo4hcsjoqg

okay, the build _almost_ finished. need to follow up on:

* FlowingLiquids compilation
* FlexiblePathfinding test compilation
* NameGenerator test failures
* BlockDetector test error
* ItemPipes test error
* Tasks test error

the slowest test suite was the ModuleTestEnvironment tests at just over 5 minutes.

I did _not_ use `--parallel` on this run.

build-logic/src/main/kotlin/reflections-manifest.gradle.kts
pollend: doesn't gestalt depend on reflections? how is it working without reflections.

engine-tests/src/test/java/org/terasology/i18n/TranslationFormatTests.java
```diff
@@ -54,20 +54,20 @@ public void setup() {
...
-    }
+//    @Test
+//    public void testPathMatcher() throws IOException, InvalidAssetFilenameException{
+//        assertFalse(format.getFileMatcher().matches(Paths.get("menu.json")));
+//        assertFalse(format.getFileMatcher().matches(Paths.get("menu.prefab")));
+//
+//        assertTrue(format.getFileMatcher().matches(Paths.get("menu.lang")));
+//        assertTrue(format.getFileMatcher().matches(Paths.get("menu_pl.lang")));
+//        assertTrue(format.getFileMatcher().matches(Paths.get("menu_en-US-x-lvariant-POSIX.lang")));
+//    }
```
pollend, line 70: do we need this test?

pollend:
the intellij branch seems to be working well are you still having this problem @keturn ?

pollend:
for the serialization test its failing to resolve the subtype in RuntimeDelegatingTypeHandler#239, something appears to be missing from the sandbox.

build-logic/src/main/kotlin/reflections-manifest.gradle.kts
keturn: It does depend on Reflections, but if there is not a pre-compiled Reflections cache it will do the scan at runtime: https://github.com/MovingBlocks/gestalt/blob/a51ba072f2e7c8e084ddf0b530d73bbb713ae2a3/gestalt-module/src/main/java/org/terasology/gestalt/module/ModuleFactory.java#L211-L225

keturn: the TypeRegistry in nui.reflections already relies exclusively on the runtime scan: https://github.com/MovingBlocks/TeraNUI/blob/0d6d07dcfbd29f4319eb6084de95d24c81ba53a9/nui-reflect/src/main/java/org/terasology/reflection/TypeRegistry.java#L91-L93

keturn:
> for the serialization test its failing to resolve the subtype in RuntimeDelegatingTypeHandler#239, something appears to be missing from the sandbox.

A RuntimeDelegatingTypeHandlerTest failure? Is it https://github.com/MovingBlocks/Terasology/pull/3992? I've been hoping to get someone to explain that one for eleven months now. Would much appreciate any leads on it you can find.

keturn:
Yesterday's run of `unit` in the omega resulted in all tests passing except for MTE's own tests, which again failed with an out-of-memory error as in https://github.com/Terasology/ModuleTestingEnvironment/issues/49#issuecomment-816926391

engine-tests/src/test/java/org/terasology/i18n/TranslationFormatTests.java
keturn: TranslationFormat still has that Matcher, so I assume so.

Pushed a fix.

keturn:
oops. gotta rebase out all those spam commits. plz hold.

keturn:
Thanks to getting the ModuleTypeRegistry change merged in TeraNUI, the branch now shows **all checks passed** 🟢  (on Java 8).

Cervator:
I've opened this issue several times over the last week, and while the details are beyond me (or at least the average amount of time I have when I come up for air), I do want to express how impressive the detective work and progress has been here. Hopefully getting close :-)

keturn:
NameGenerator tests fixed with addition of LOAD_CLASSPATH_MODULES_PROPERTY

keturn:
lawl. okay, MTE tests on modules-that-depend-on-Inventory work again, with https://github.com/Terasology/Inventory/pull/42

engine-tests/src/test/java/org/terasology/engine/persistence/serializers/TypeSerializerTest.java
```diff
@@ -85,14 +95,50 @@ void testDeserialize() {
...
+        assertNotEmpty(typeHandlerLibrary.getTypeHandler(Animal.class));
+        assertNotEmpty(typeHandlerLibrary.getTypeHandler(Dog.class));
+
+        assertNotEmpty(typeRegistry.load("org.terasology.engine.persistence.serializers.TypeSerializerTest$Animal"));
+        assertNotEmpty(typeRegistry.load("org.terasology.engine.persistence.serializers.TypeSerializerTest$Dog"));
+
+//        ModuleEnvironment env = moduleManager.getEnvironment();
+//        assertNotEmpty(env.getSubtypesOf(Animal.class));
+
+//        assertNotEmpty(typeRegistry.getSubtypesOf(Animal.class));
```
keturn, line 140: FIXME: bunch o clutter added to these tests.

should def remove the commented-code. but should the extra assertions stay? or are they out of scope for the class under test?

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -4,99 +4,90 @@
...
+            moduleFactory = new ClasspathCompromisingModuleFactory();
+        } else {
+            moduleFactory = new ModuleFactory();
+        }
+        moduleFactory.setDefaultLibsSubpath("build/libs");
+
+        Map<String, ModuleMetadataLoader> mmlm = moduleFactory.getModuleMetadataLoaderMap();
+        mmlm.put(TerasologyConstants.MODULE_INFO_FILENAME.toString(), metadataReader);
+        mmlm.put("engine-module.txt", metadataReader);  // FIXME: this should be *only* for engine-module.
+        // FIXME: …or maybe engine-module doesn't need a special filename anymore?
```
keturn, line 84: FIXME: engine-module.txt should only be for engine module, if at all.

keturn (commented):
I combed through the branch and tidied up some of the rough edges from when we were first getting things working.

There are a few things here and there that _could_ be extracted to a separate PR, but they're little debugging things. Null checks, logs, that sort of thing. I don't think extracting them would make the branch significantly easier to review.

keturn:
I tried to do another omega test run this morning.

* Exception in engine-test: NetworkOwnershipTest
* compile failure in ItemPipes that looks like the gestalt branch is out of sync with the Health changes again
* and OutOfMemory error somewhere?

keturn:
:heavy_check_mark: module branches updated again

keturn:
org.terasology...mathTypes.BlockAreaTypeHandlerTest seems to be where it's getting stuck? sometimes?

keturn:
Problem is, these tests don't fail when I run only that test class. I think there must be some resource leak building up when we do a lot of them in one run.

keturn:
# Status Report

This branch still has its problems, but it's getting to the point where it's harder to tell to what extent the problems are unique to this branch (and thus whether fixing them is in scope for this PR).

There's definitely a maintenance cost to keeping this branch and all the associated module branches open, and I think it's reached the tipping point where it may be better to merge it than continuing to accumulate fixes here.

pollend (dismissed):
I glanced through this code it seems fine to me. umm I guess it doesn't hurt to merge it because the state of this PR is already getting pretty large.

keturn:
* ✔ ReallyCrashGameBlock really crashes game
* ✔ can run headless server
* ✔ can connect to headless server from client

I am getting some troubling exceptions when I _disconnect,_ but that's getting back to "I'm not sure which bugs are in branch vs trunk."

engine-tests/src/main/java/org/terasology/engine/testUtil/Assertions.java
```diff
@@ -0,0 +1,25 @@
...
+        if (!obj.isPresent()) {
+            fail("Empty result.");
+        }
+    }
+
+    public static void assertNotEmpty(Iterable<?> obj) {
+        if (Iterables.isEmpty(obj)) {
+            fail("Empty result.");
+        }
+    }
```
skaldarnar, line 24: I'm surprised this is not available by default 🧐 Are these now available for module tests as well?

engine-tests/src/main/java/org/terasology/engine/testUtil/ModuleManagerFactory.java
```diff
@@ -2,41 +2,21 @@
...
-    public static ModuleManager create() throws Exception {
-        return create(false);
+    private ModuleManagerFactory() {
     }
 
-    public static ModuleManager create(boolean loadModulesFromClasspath) throws Exception {
-        ModuleManager moduleManager = new ModuleManager("", Collections.emptyList(), loadModulesFromClasspath);
-        loadUnitTestModule(moduleManager);
+    public static ModuleManager create() throws Exception {
+        // Loading screens, among other things, break when NUI classes are not added to engine.
```
skaldarnar, line 16: I guess this is worth its own issue. At least we should come up with a pattern or some ideas on hardening such that headless envrionments don't run into these problems...

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -4,99 +4,92 @@
...
+    private final ModuleFactory moduleFactory;
     private final ModuleInstallManager installManager;
-    private Module engineModule;
+    private final Module engineModule;
 
     public ModuleManager(String masterServerAddress) {
         this(masterServerAddress, Collections.emptyList());
     }
 
     public ModuleManager(String masterServerAddress, List<Class<?>> classesOnClasspathsToAddToEngine) {
```
skaldarnar, line 69: What is the `masterServer`? Does the module manager connect to a server to fetch modules from there? 

I saw a test where you pass in the empty string `""` - I guess to denote that there is no master server. Why not use `null` to denote that? (It probably was like that before, just trying to understand this).

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
skaldarnar: Is `engine-module.txt` treated any different anywhere? We're obviously using the same reader here, at least...

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -105,49 +98,61 @@ public ModuleManager(String masterServerAddress, List<Class<?>> classesOnClasspa
...
      * tree, and we want to make sure engine is at the root of it.
-     */   
-    private void ensureModulesDependOnEngine(Module engineModule) {
+     */
+    private void ensureModulesDependOnEngine() {
         DependencyInfo engineDep = new DependencyInfo();
         engineDep.setId(engineModule.getId());
         engineDep.setMinVersion(engineModule.getVersion());
-        engineDep.setMaxVersion(engineModule.getVersion().getNextPatchVersion());
+        engineDep.setMaxVersion(engineModule.getVersion().getNextMinorVersion());
```
skaldarnar, line 109: Is this a nother semver related change?

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -105,49 +98,61 @@ public ModuleManager(String masterServerAddress, List<Class<?>> classesOnClasspa
...
+
+        HashSet<Name> engineModules = Sets.newHashSet(engineModule.getId());
+        engineModules.addAll(engineModule.getMetadata().getDependencies().stream().map(DependencyInfo::getId).collect(Collectors.toList()));
```
skaldarnar, line 112: What are "engine modules", i.e., what are we collecting here?

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -105,49 +98,61 @@ public ModuleManager(String masterServerAddress, List<Class<?>> classesOnClasspa
...
-            additionalClassesList.addAll(classesOnClasspathsToAddToEngine); // provide access to any facade-provided classes
-            Class<?>[] additionalClassesArray = new Class[additionalClassesList.size()];
-            additionalClassesArray = additionalClassesList.toArray(additionalClassesArray);
-            engineModule = ClasspathModule.create(metadata, getClass(), additionalClassesArray);
-        } catch (IOException e) {
-            throw new RuntimeException("Failed to read engine metadata", e);
-        } catch (URISyntaxException e) {
-            throw new RuntimeException("Failed to convert engine library location to path", e);
-        }
+        Module packageModule = moduleFactory.createPackageModule("org.terasology.engine");
```
skaldarnar, line 126: What is a "package module"? Does it just collect everything that is within the package? Can other modules add to this engine module package?

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -105,49 +98,61 @@ public ModuleManager(String masterServerAddress, List<Class<?>> classesOnClasspa
...
     }
 
     private Module loadEngineModule(List<Class<?>> classesOnClasspathsToAddToEngine) {
-        Module engineModule;
-        try (Reader reader = new InputStreamReader(getClass().getResourceAsStream("/engine-module.txt"), TerasologyConstants.CHARSET)) {
-            ModuleMetadata metadata = metadataReader.read(reader);
-            List<Class<?>> additionalClassesList = new ArrayList<>(classesOnClasspathsToAddToEngine.size() + 2);
-            additionalClassesList.add(Module.class); // provide access to gestalt-module.jar
-            additionalClassesList.add(Asset.class); // provide access to gestalt-asset-core.jar
-            additionalClassesList.add(UIWidget.class); // provide access to nui.jar
```
skaldarnar, line 134: So how does/did this work? You just need to provide a single entry-point class, and the rest is collected automatically? (Guess I'll need a lecture on reflections and class loading...)

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -105,49 +98,61 @@ public ModuleManager(String masterServerAddress, List<Class<?>> classesOnClasspa
...
+        Set<Scanner> scanners = packageReflections.getConfiguration().getScanners();
+        config.setScanners(scanners.toArray(new Scanner[0]));
 
-        enrichReflectionsWithSubsystems(engineModule);
+        for (Class<?> aClass : classesOnClasspathsToAddToEngine) {
+            config.addUrls(ClasspathHelper.forClass(aClass));
+        }
 
-        return engineModule;
+        // TODO: is this using reflections.cache?
```
skaldarnar, line 138: How can we figure this out?

skaldarnar (commented):
First round of review...

build-logic/src/main/kotlin/reflections-manifest.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+
+    doLast {
+        val classPaths = mainSourceSet.compileClasspath.map { it.toURI().toURL() }
+        val classLoader = URLClassLoader(classPaths.toTypedArray())
+        try {
+            val reflections = Reflections(
+                ConfigurationBuilder()
+                .setSerializer(JsonSerializer())
+                .addClassLoader(classLoader)
+                // .filterInputsBy(FilterBuilder.parsePackages("+org"))
```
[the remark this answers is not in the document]
keturn: I think that was from some attempt to reduce the amount of time this was taking by restricting which packages it considers.

Not that including all `org` packages (but not `com`, etc) makes much sense.

engine-tests/src/main/java/org/terasology/engine/testUtil/Assertions.java
keturn: Right? There are things for Arrays and Lists and other Collections, but I didn't see anything in junit Assertions for Iterable.

engine-tests/src/main/java/org/terasology/engine/testUtil/ModuleManagerFactory.java
keturn: Yeah. And I'm on the fence about whether headless environments should have NUI. There are some "headless" tests where you don't need anything to _render,_ but maybe you're still testing to make sure the controls are there and do the right interactions.

On the other hand, on a real headless server, there's nothing trying to interact with the invisible UI, so putting it through these screens is pretty pointless.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: As you guessed, this is something that is unchanged in this branch. It is a good question that I have not explored the answer to.

But yeah, one of the things the ModuleManager does is set up that `ModuleInstallManager`, which is supposed to handle installation of modules at run-time?

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: oh, it was a workaround to avoid that gestalt#100 problem.

I think it would be okay to not explicitly set a maxVersion here at all, and let the gestalt DependencyInfo make its decisions about what is compatible with the required minVersion.

(But didn't you want to require modules to start specifying their engine compatibility in their module.txt?)

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: > Does it just collect everything that is within the package?

Yep, https://github.com/MovingBlocks/gestalt/blob/f459dac2fd0b073f5a76124b106aa491d5ddc2ba/gestalt-module/src/main/java/org/terasology/gestalt/module/ModuleFactory.java#L268-L274

> Can other modules add to this engine module package?

Worrying question, right? Because there's nothing stopping any jar from containing a class under `org/terasology/engine`.

I think/hope that since module jars aren't loaded from the classpath with the main application classloader, there should/could be something about the ModuleClassLoader that prevents them from being conflated with the package module? But there is an awful lot of "should" and "hope" in that statement.

engine/src/main/java/org/terasology/engine/audio/openAL/staticSound/OpenALSound.java
```diff
@@ -119,18 +119,25 @@ protected void doReload(StaticSoundData newData) {
...
 
-    private static class DisposalAction implements Runnable {
+    public static class DisposalAction implements DisposableResource {
         private final ResourceUrn urn;
         private int bufferId;
-        private final WeakReference<OpenALSound> asset;
+        private WeakReference<OpenALSound> asset;
 
-        DisposalAction(ResourceUrn urn, OpenALSound openALSound) {
+        public DisposalAction(ResourceUrn urn) {
```
pollend, line 127: why is this exposed, shouldn't this just be related to sounds resources. maybe protected?

engine/src/main/java/org/terasology/engine/audio/openAL/staticSound/OpenALSound.java
```diff
@@ -42,17 +43,16 @@
...
-
-    private DisposalAction disposalAction;
+    private final DisposalAction disposalAction;
 
     // TODO: Do we have proper support for unloading sounds (as mods are changed?)
     private int bufferId;
 
-    public OpenALSound(ResourceUrn urn, AssetType<?, StaticSoundData> assetType, StaticSoundData data, OpenALManager audioManager) {
-        super(urn, assetType);
+    public OpenALSound(ResourceUrn urn, AssetType<?, StaticSoundData> assetType, StaticSoundData data, OpenALManager audioManager, OpenALSound.DisposalAction disposalAction) {
```
pollend, line 51: umm, we are exposing this because its passed in through a constructor.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: Oh, I think I did figure that one out: When using ModuleFactory's createDirectoryModule or createArchiveModule, then it will check for a reflections.cache to load. But there's no implicit cache checking when using the Reflections API directly, it only reads from a cache if you call those methods.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: These go through reflections.ClasspathHelper.forClass that returns a URL that "contains the class."

That is often a `.jar` that contains that `.class`, so you pass `UIWidget.class` and then Reflections will do its scanning thing on everything in `nui.jar`.

but I definitely have concerns about how well this works in all cases.

What if you're running under IntelliJ and it has put `classes/` directories on the classpath instead of jars?

What if you're running from a distribution build with `jpackage` and it's bundled everything in to one archive or fat jar or something?

The Reflections stuff and the APIs we've built on top of them poke through the abstraction layer that usually keeps us from worrying too much about how the class files are organized, and I am uneasy about it.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: In _this specific context_ (the `ensureModulesDependOnEngine` method), this is–  

oh, I remember now, this was from the phase of this PR when we made a module for NUI, and so this `engineModules` here was "engine and _any modules engine depends on_" to avoid creating a circular dependency. e.g. engine depends on NUI, we don't want to add a dependency to NUI to make it depend on engine.

In this branch's current state, NUI is not a module, so we don't have to worry about that.

But also in this branch's current state, we have this [ridiculous list of hardcoded NUI packages and classes](https://github.com/MovingBlocks/Terasology/blob/1e9a36c7fcaabc513c9496bcbd32b888c1a163c5/engine/src/main/java/org/terasology/engine/core/module/ExternalApiWhitelist.java#L14), which according to some of the conversation around NUI 3.0 is a thing we should have some other way of doing: https://github.com/MovingBlocks/TeraNUI/pull/27#issuecomment-805639667

I remain very fuzzy on how to configure these "modules" that are libraries not subject to the same constraints we put on the stuff in `/modules/*`.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: I _imagine_ that engine has this `engine-module.txt` has this unique filename because it was loaded from a `class.getResource` call https://github.com/MovingBlocks/Terasology/blob/0fa1272241b43be94e9920c84a41683643c9f240/engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java#L128  
instead of a filesystem API that opens a file in the root of a specific directory or archive.

So maybe it needed a name to distinguish it from a `/module.txt` file that might have been put there by anything else on the classpath?

but now that we've moved it to `org/terasology/engine/` maybe that's enough of a qualifier that it could just be named `module.txt` again?

engine-tests/src/main/java/org/terasology/engine/testUtil/Assertions.java
keturn: oh, and yes, public classes in `engine-tests` are available to module tests, I believe.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
skaldarnar: I see, that makes sense now. 

And yes, I'd like to have modules specify their engine version (range) explicitly as there is currently no way to figure out with which version of the engine a module can be build and run. Not sure where on the timeline that is, though.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
skaldarnar: I see - I was wondering whether these are subsystems or something like that. Maybe somebody needs to explain me the details of what we are doing, what the plan for maintaining it was/is, and what other (feasible) options we have... :see_no_evil:

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
skaldarnar: Alright, at least that confirms both my assumption and suspicion. 

Do we have a diagram of class and module loaders, class paths, and all that somewhere?

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
skaldarnar: Okay, so there are a bunch of assumptions that must hold for this to work (and typically it will work in our workspaces and in distributions).

The classes we mention here, however, are handpicked "representatives" of each of the artifacts we want to allow-list. If we (re-)move that class, things will break, but at least it will break at compilation time (actual class references here, not just string identifiers or such)...

Well, at least there are still people around who understand this :upside_down_face:

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
skaldarnar: How often do we call these methods? Can we do some crude benchmarking for an Omega workspace and run it on a couple of developer machines to get a feeling when the reflection cache helps, and how big the impact of removing it is?

Does this only affect local development, or also our distributions?

keturn: Not called too often. I think it's just once per GameEngine.initialize, which looks like it's once per process.

(not counting tests, which want to make new engines or engine-like things a whole bunch.)

> Does this only affect local development, or also our distributions?

all.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: Wouldn't be hard to do here on the engine side. It'd be a matter of replacing this "add engine dependencies to everything" method with a "assert things declared their own engine dependency" method.

keturn:
In https://github.com/MovingBlocks/Terasology/pull/4622#discussion_r622603164 skaldarnar wrote:
> I see - I was wondering whether these [engineModules] are subsystems or something like that.

[replying here instead of inline because it goes beyond that bit of code.]

Ah, you may have got that idea from the ModuleManager constructor argument named `classesOnClasspathsToAddToEngine`, which is handled in [loadEngineModule](https://github.com/MovingBlocks/Terasology/blob/1e9a36c7fcaabc513c9496bcbd32b888c1a163c5/engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java#L125). Those _are_ subsystems: https://github.com/MovingBlocks/Terasology/blob/1e9a36c7fcaabc513c9496bcbd32b888c1a163c5/engine/src/main/java/org/terasology/engine/core/TerasologyEngine.java#L184-L185

…

The [lists of allowed packages and classes](https://github.com/MovingBlocks/Terasology/blob/1e9a36c7fcaabc513c9496bcbd32b888c1a163c5/engine/src/main/java/org/terasology/engine/core/module/ExternalApiWhitelist.java) are used in a straightforward manner, from what I've been able to tell. You list a thing there, and it gets used in a conditional somewhere that does a direct comparison against the list. If it's allowed, then any module is allowed to load that class and do whatever it wants with it.

The things that get passed to a Reflections configuration, on the other hand, are not so direct. Reflections isn't really there to provide permissions checks. Reflections data is used to answer questions of "What classes have this annotation?" and "What are the subtypes of this class?"

– but we want to make sure the answers to those queries don't return things from modules that aren't active.  
– and failing to get the answer you expected from one of those things can sure _look_ a whole lot like when a thing is missing from the permission list.

We could use clarity on when a class needs to be 
* added to the permission list
* added to classesOnClasspathsToAddToEngine
* be annotated with `@gestalt.module.sandbox.API`
* be added to the permissionSet independent of any of those things? https://github.com/MovingBlocks/Terasology/blob/1e9a36c7fcaabc513c9496bcbd32b888c1a163c5/engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java#L186-L187
* be included by the engine module's `classPredicate` https://github.com/MovingBlocks/gestalt/blob/4b6ea6d50176da0ac0d93a9186f1b44ba2cb12c4/gestalt-module/src/main/java/org/terasology/gestalt/module/Module.java#L60
    **classPredicate**: Predicate to determine what classes to include from the main classpath (classes from the unloaded classpaths will be included automatically)

the Module.classPredicate being a thing that's either new to v7 or was just well-hidden in v5. You can have a class that the security policy and clasloaders all allow you to _load,_ but something may ask ModuleEnvironment.getModuleProviding about it, and if no modules recognize it then it's not usable for some purposes after all.

oh and also these should probably line up somehow? https://github.com/MovingBlocks/gestalt/blob/4b6ea6d50176da0ac0d93a9186f1b44ba2cb12c4/gestalt-module/src/main/java/org/terasology/gestalt/module/Module.java#L56-L57
- **fileSources**:    Any sources of files that compose the module. Must not be null - can be `EmptyFileSource`
- **classpaths**:     Any extra classpaths to load for the module


> Maybe somebody needs to explain me the details of what we are doing, what the plan for maintaining it was/is, and what other (feasible) options we have... 🙈

That sounds super useful! When you get that explanation, can you go back in time to [February](https://github.com/MovingBlocks/Terasology/issues/4509) or earlier and let us know how it works? It would have saved us a _ton_ of time if we knew wtf was going on in here.

keturn:
This test from the Malicious module fails:

```java
    public void testReflectionConstruction() {
        numTests++;
        try {
            PathManager.getInstance();
        } catch (NoClassDefFoundError e) {
            passedTests++;
            return;
        }
        logger.warn("Succeeded in getting an instance.");
```

It's not something we're in the habit of checking regularly, so if it also fails on `develop`, I was going to let it slide for this PR.

Fortunately for `develop` and unfortunately for us here, it _does_ pass on `develop`.

I think that must mean we've defined something in [`loadEngineModule`](https://github.com/MovingBlocks/Terasology/blob/1e9a36c7fcaabc513c9496bcbd32b888c1a163c5/engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java#L125) too broadly.

I guess I'll try to write some unit tests for `loadEngineModule`?

engine/src/main/java/org/terasology/engine/core/bootstrap/EnvironmentSwitchHandler.java
```diff
@@ -53,10 +60,19 @@
...
 
+    private final Collection<CopyStrategyEntry<?>> typesWithCopyConstructors = ImmutableList.of(
+            new CopyStrategyEntry<>(Quaternionf.class, Quaternionf::new),
+            new CopyStrategyEntry<>(Vector2f.class, Vector2f::new),
+            new CopyStrategyEntry<>(Vector2i.class, Vector2i::new),
+            new CopyStrategyEntry<>(Vector3f.class, Vector3f::new),
+            new CopyStrategyEntry<>(Vector3i.class, Vector3i::new),
+            new CopyStrategyEntry<>(Vector4f.class, Vector4f::new),
+            new CopyStrategyEntry<>(Vector4i.class, Vector4i::new)
+    );
```
skaldarnar, line 71: Is this new, or was this moved from somewhere? 

Edit: I see there's a `RegisterCopyStrategy` annotation that was scanned before. Did we deprecate that in favor of explicitly listing them here?

engine/src/main/java/org/terasology/engine/core/bootstrap/EnvironmentSwitchHandler.java
```diff
@@ -113,22 +122,21 @@ public void handleSwitchToGameEnvironment(Context context) {
          */
         unregisterPrefabFormats(assetTypeManager);
         registeredPrefabFormat = new PrefabFormat(componentLibrary, typeHandlerLibrary);
-        assetTypeManager.registerCoreFormat(Prefab.class, registeredPrefabFormat);
+        assetTypeManager.getAssetFileDataProducer(assetTypeManager
+                .getAssetType(Prefab.class)
+                .orElseThrow(() -> new RuntimeException("Cannot get Prefab Asset typee")))
```
skaldarnar, line 127: Typo

```suggestion
                .orElseThrow(() -> new RuntimeException("Cannot get Prefab Asset type")))
```

engine/src/main/java/org/terasology/engine/core/module/ModuleListDownloader.java
```diff
@@ -55,21 +63,20 @@ public ModuleRegistry call() throws IOException {
...
+            for (JsonElement jObject : jsonArray) {
+                String json = gson.toJson(jObject);
 
-            while (reader.hasNext()) {
-                ModuleMetadata meta = metaReader.read(reader);
+                ModuleMetadata meta = metaReader.read(new StringReader(json));
                 logger.debug("Read module {} - {}", meta.getId(), meta.getVersion());
-                RemoteModule remoteModule = new RemoteModule(meta);
-                modules.add(remoteModule);
+                modules.add(new Module(meta, new EmptyFileSource(), Collections.emptyList(), new Reflections(),
```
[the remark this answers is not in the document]
skaldarnar: Was `RemoteModule` coming from gestalt v5, or why don't we use it anymore?

Edit: Just saw it is removed in this PR. Could you please provide the reasoning for that?

skaldarnar (approved):
I still don't get all the details in here, but I couldn't spot anything that obviously "bad" or "broken". I've run my Omega workspace with these changes (and all the module changes) and could start a game just fine.

Therefore, I'm fine with getting this merged to move forward, and fix anything that pops up afterwards in a separate PR. We could even just do a preview release to run it from distribution, throw it on the play-test server and see what happens.

engine/src/main/java/org/terasology/engine/core/bootstrap/EnvironmentSwitchHandler.java
keturn: There's a PR for that https://github.com/MovingBlocks/TeraNUI/pull/50

engine/src/main/java/org/terasology/engine/core/module/ModuleListDownloader.java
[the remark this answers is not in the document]
keturn: @pollend thoughts on RemoteModule?

Cervator (dismissed):
Have not actually tested and unable to go through the diff, but I've been at least following the related chat when able and want to add my support to merging this soon ™️ if not immediately. It is amazing work, should be allowed to dock safely already, then anything outstanding we can keep going with as we can.

skaldarnar (dismissed):

keturn:
if you need to check out things _before_ this was merged, see tag https://github.com/MovingBlocks/Terasology/tree/pre-gestalt-v7-merge

PR #4623, opened 2021-04-12
chore: improve block mesh generator

pollend:
I use a hash for the initial comparison for BlockUri's this avoid a lot of expensive string comparisions. EnumMap use the ordinal for its lookup but its quite a bit more expensive.  looking at the code there is an extra check to verify if the key is a valid class and this is called for every block so its a non-trival amount of code that is introduced. probably better to just use an array which is a lot more lightweight then the enumMap.


engine/src/main/java/org/terasology/engine/world/block/BlockUri.java
```diff
@@ -156,7 +162,8 @@ public boolean equals(Object obj) {
         }
         if (obj instanceof BlockUri) {
             BlockUri other = (BlockUri) obj;
-            return Objects.equal(other.blockFamilyDefinition, blockFamilyDefinition)
+            return this.hashCode == ((BlockUri) obj).hashCode
```
skaldarnar, line 165: The has code is computed from `blockFamilyDefinition`, `shape`, and `blockName`. If we are reasonably sure that we don't have hash collisions here, why check the other fields explicitly in addition?

skaldarnar:
```suggestion
            return this.hashCode == other.hashCode
```

pollend: the comparison falls through most of the time so the cost of checking the other terms is trivial.

pollend:
[Small Optimization for blockmesh generator](https://trello.com/c/h9eJ3fct/109-small-optimization-for-blockmesh-generator)

jdrueckert (approved):

PR #4624, opened 2021-04-13
chore: remove unused gelatinousCube asset

pollend:
While working on this PR: https://github.com/MovingBlocks/Terasology/pull/4620

I noticed this asset and from what I can tell its not used anywhere. I can't find it in my omega workspace and the shader has some kind of warping cos effect on the y axis.  Looks like some test code that was never actually removed from the codebase: https://github.com/MovingBlocks/Terasology/commit/cceb5d4cb15a3c43404f7c80cb1c64dfc1859f33


Cervator (approved):
I can speak to this, not that I've tested the code, but: This is/was the OG proto-Gooey. Once upon a time in the long long ago, in the before time, begla created a gelatinous cube with pure magic, no models, and all was well. They'd bounce around as cubes with another inner cube of sorts. These guys: https://youtu.be/H3qItsrpceY?t=2715

Was the evolution of the super simple slime mob which was the first ever creature, and I think just some leftover of pulling in the old GMC! pack from MC.

Next evolution of that is/was probably https://github.com/Terasology/GelCubes which in turn came out of some of my ancient Portals/Spawning code. Later on yet we got the actual model-based Gooeys

I assume the GelCube module is/was based on this once upon a time, since it uses a hard coded `"renderType" : "GelatinousCube"` mesh thing. Probably broke long ago though, and I don't imagine losing this code in the engine would really hurt. Maybe toss an issue into the module that references this PR just for historical purposes in case anybody is ever curious

PR #4625, opened 2021-04-13
fix: correct colliders for LIQUID

pollend:
looks like I tried setting up the colliders the same way for liquid and solid blocks. this correct the problem and is easily verified by trying to swim in any body of water in the world. 

ref PR: https://github.com/MovingBlocks/Terasology/pull/4615


engine/src/main/java/org/terasology/engine/physics/bullet/world/VoxelBlockFluidWorld.java
```diff
@@ -3,19 +3,61 @@
...
+public class VoxelBlockFluidWorld implements VoxelWorld {
     private final boolean[] registered = new boolean[Short.MAX_VALUE];
+    private final VoxelCollisionAlgorithmWrapper wrapper;
+    private final btRigidBody.btRigidBodyConstructionInfo blockConsInf;
+    private final btVoxelShape worldShape;
+    private final btRigidBody rigidBody;
 
     public VoxelBlockFluidWorld(BulletPhysics physics) {
-        super(physics);
+        btDiscreteDynamicsWorld discreteDynamicsWorld = ((BulletPhysics) physics).getDiscreteDynamicsWorld();
```
keturn, line 36: nitpick: redundant cast?

pollend: yep

pollend, line 36:
```suggestion
        btDiscreteDynamicsWorld discreteDynamicsWorld = physics.getDiscreteDynamicsWorld();
```

engine/src/main/java/org/terasology/engine/physics/bullet/world/VoxelBlockWorld.java
```diff
@@ -3,17 +3,58 @@
...
+public class VoxelBlockWorld implements VoxelWorld {
     private final boolean[] registered = new boolean[Short.MAX_VALUE];
+    private final VoxelCollisionAlgorithmWrapper wrapper;
+    private final btRigidBody.btRigidBodyConstructionInfo blockConsInf;
+    private final btVoxelShape worldShape;
+    private final btRigidBody rigidBody;
 
     public VoxelBlockWorld(BulletPhysics physics) {
-        super(physics);
+        btDiscreteDynamicsWorld discreteDynamicsWorld = ((BulletPhysics) physics).getDiscreteDynamicsWorld();
```
pollend, line 36:
```suggestion
        btDiscreteDynamicsWorld discreteDynamicsWorld =  physics.getDiscreteDynamicsWorld();
```

keturn (commented):
I'm glad you found the cause of this bug; it's an important one and I wouldn't have known where to start!

I'm always telling people they should try to review things even if they don't have domain expertise, so I guess I should walk the talk. I don't know anything about our physics system or bullet but here's what I see:

There's an entire file removed; an abstract superclass. Looks like it's been inlined in two of its subclasses. Why? Is the intention to wholly revert that part of #4615?

well, no, I guess it's not entirely reverting #4615, because it was [even more differenter before that](https://github.com/MovingBlocks/Terasology/pull/4615/files#diff-7c27734b0dfb7770d775df4e6a059dcff6c9bf1456c71f695043581a4db766c4).

As a reviewer, I'm worried that there might be other things we're missing, as it's hard to compare the two `VoxelBlock*World` classes here with each other and their pre-4615 implementations.

I also notice this is a bugfix PR without automated tests. Is this something that it would make sense to add automated tests for? (that would help with my "we might have overlooked something that should be different in the middle of this duplicated code" worries.)

pollend:
> I'm glad you found the cause of this bug; it's an important one and I wouldn't have known where to start!
> 
> I'm always telling people they should try to review things even if they don't have domain expertise, so I guess I should walk the talk. I don't know anything about our physics system or bullet but here's what I see:
> 
> There's an entire file removed; an abstract superclass. Looks like it's been inlined in two of its subclasses. Why? Is the intention to wholly revert that part of #4615?
> 
> well, no, I guess it's not entirely reverting #4615, because it was [even more differenter before that](https://github.com/MovingBlocks/Terasology/pull/4615/files#diff-7c27734b0dfb7770d775df4e6a059dcff6c9bf1456c71f695043581a4db766c4).
> 
> As a reviewer, I'm worried that there might be other things we're missing, as it's hard to compare the two `VoxelBlock*World` classes here with each other and their pre-4615 implementations.
> 
> I also notice this is a bugfix PR without automated tests. Is this something that it would make sense to add automated tests for? (that would help with my "we might have overlooked something that should be different in the middle of this duplicated code" worries.)

I removed the abstract class because I couldn't find a lot that is the same. I guess I could have some method to assign the mask and group. you think they share enough that it would be possible to move it back to the abstract class?

how would we go about testing this?

keturn:
> how would we go about testing this?

Wait, didn't _I_ just ask _you_ that? Oh, no, I guess I asked a slightly different question. 😆 Okay, fine, I'll start:

Being totally unfamiliar with what the interfaces around here are like, the first thing that comes to mind for me is that we could do the big MTE-style integration test: build a world, put some solid and liquid blocks in it, throw entities at them and see where they stop moving.

But we know MTE tests are slow, and involve running _so_ much code that when they do fail, they're not much help in pinpointing exactly which method is failing.

What are some more focused methods we could call to find out if solid blocks are solid and liquid blocks are not?

The `VoxelWorld` interface is giving me _no_ ideas; everything returns void and so I'd just be testing, what, a side-effect of `setBlock` or `registerBlock` has on the blocks I pass to it?

Maybe a call to Physics.rayTrace? That's still a fair amount of work, as we'd still have to build up a world to some extent, but at least we wouldn't have to tick through it to find out what happens.

Is there any thing smaller-scale we could do? Are there ways of testing the difference between these blocks other than a ray trace?

pollend:
> > how would we go about testing this?
> 
> Wait, didn't _I_ just ask _you_ that? Oh, no, I guess I asked a slightly different question. laughing Okay, fine, I'll start:
> 
> Being totally unfamiliar with the interfaces around here, the first thing that comes to mind for me is that we could do the big MTE-style integration test: build a world, put some solid and liquid blocks in it, throw entities at them and see where they stop moving.
> 
> But we know MTE tests are slow, and involve running _so_ much code that when they do fail, they're not much help in pinpointing exactly which method is failing.
> 
> What are some more focused methods we could call to find out if solid blocks are solid and liquid blocks are not?
> 
> The `VoxelWorld` interface is giving me _no_ ideas; everything returns void and so I'd just be testing, what, a side-effect of `setBlock` or `registerBlock` has on the blocks I pass to it?
> 
> Maybe a call to Physics.rayTrace? That's still a fair amount of work, as we'd still have to build up a world to some extent, but at least we wouldn't have to tick through it to find out what happens.
> 
> Is there any thing smaller-scale we could do? Are there ways of testing the difference between these blocks other than a ray trace?

umm seems difficult to write a good test. might be simpler to test it in game then try to write a test to verify a behavior.

pollend:
> > how would we go about testing this?
> 
> Wait, didn't _I_ just ask _you_ that? Oh, no, I guess I asked a slightly different question. laughing Okay, fine, I'll start:
> 
> Being totally unfamiliar with the interfaces around here, the first thing that comes to mind for me is that we could do the big MTE-style integration test: build a world, put some solid and liquid blocks in it, throw entities at them and see where they stop moving.
> 
> But we know MTE tests are slow, and involve running _so_ much code that when they do fail, they're not much help in pinpointing exactly which method is failing.
> 
> What are some more focused methods we could call to find out if solid blocks are solid and liquid blocks are not?
> 
> The `VoxelWorld` interface is giving me _no_ ideas; everything returns void and so I'd just be testing, what, a side-effect of `setBlock` or `registerBlock` has on the blocks I pass to it?
> 
> Maybe a call to Physics.rayTrace? That's still a fair amount of work, as we'd still have to build up a world to some extent, but at least we wouldn't have to tick through it to find out what happens.
> 
> Is there any thing smaller-scale we could do? Are there ways of testing the difference between these blocks other than a ray trace?

do we want to add an MTE test for this now. what kind of MTE test to consider for colliders?

Cervator:
Not much I can add to the review right now since it is somewhat technically dense - but I can say that I've tested it locally, water appears to be wet again rather than solid, and I can still scoop water and fill a tank. Liquid also still flows in JS. Can throw liquid and solid blocks around against other liquid or solid blocks without apparent weirdness (well, beyond that you can _throw_ a water block, which just stays a dropped block, being pushed around by other physics-enabled dropped blocks)

pollend:
[Correct colliders for fluids](https://trello.com/c/1BSU8kQ3/98-correct-colliders-for-fluids)

skaldarnar (approved):

skaldarnar:
I'm going to merge this as it fixes a quite severe bug. 

This does not render @keturn's comment obsolete, and we should find a way to ensure that this does not break that easily with automated tests. However, we may need to learn a lot more about the engine _and_ designing testable software before that becomes such an easy action item that we can expect proper tests for everything (either already in place or added by a PR).

pollend:
> I'm going to merge this as it fixes a quite severe bug.
> 
> This does not render @keturn's comment obsolete, and we should find a way to ensure that this does not break that easily with automated tests. However, we may need to learn a lot more about the engine _and_ designing testable software before that becomes such an easy action item that we can expect proper tests for everything (either already in place or added by a PR).

I've been debating if this is really that great. adds a level of indirection to make something more generic when the use case is very specific.

PR #4630, opened 2021-04-20
Fix: Gitignore some more files that Intellij automatically generates.

4Denthusiast:
IntelliJ automatically generated these files in my workspace, but they are just empty or relate to local display preferences or something. I assume other people have been running into the same issue. Anyway, they aren't actually part of any changes I've made to the project and they don't need to be shared, but they show up as changed in git, which is annoying, so this should fix that.

There was an explicit comment previously about why dictionaries were not gitignored, even though it generates one per user, that the dictionaries are actually meant to be shared. There's some discussion about it at https://youtrack.jetbrains.com/issue/IDEA-208431, which seems to mostly be the Jetbrains staff arguing with their users that it's working as intended, and the users arguing that they don't want it to work this way. Sharing dictionaries doesn't sound like that useful a feature, and I haven't even got any entries in my dictionary but it still generated a file with an empty list, which I then have to ignore manually every time I'm staging changes.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -10,13 +10,8 @@ workspace.xml
 # Shelf is where it put changes you're not ready to share. (it's like git stash)
 /shelf
 
-# Dictionaries: do not ignore!
-#  /dictionaries
-# Yes, it makes a new file for each user, but JetBrains product manager Dmitry Jemerov explains:
-#
-# >    The dictionary is not in fact per-user; the spellchecker uses words from all users'
-# >    dictionaries for checking the spelling.
-#          — https://youtrack.jetbrains.com/issue/WI-45679#focus=streamItem-27-3457290.0-0
+# Dictionaries: automatically generated per user
+/dictionaries
 
 # managed by gradle sync
 /gradle.xml
@@ -33,6 +28,8 @@ workspace.xml
 /uiDesigner.xml
 /modules.xml
 /modules
+/inspectionProfiles
+/libraries-with-intellij-classes.xml
 
 # Ignore Sonar linting, if set up locally
 /sonarlint/
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -10,8 +10,13 @@ workspace.xml
 # Shelf is where it put changes you're not ready to share. (it's like git stash)
 /shelf
 
-# Dictionaries: automatically generated per user
-/dictionaries
+# Dictionaries: do not ignore!
+#  /dictionaries
+# Yes, it makes a new file for each user, but JetBrains product manager Dmitry Jemerov explains:
+#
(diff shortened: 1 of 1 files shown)


keturn:
### dictionaries

> Sharing dictionaries doesn't sound like that useful a feature

Strongly disagree. Projects have all sorts of jargon that is not in the default dictionary for that language. Very much worth sharing.

I don't agree with their choice of storage mechanism here, but it seems workable until we encounter someone who cannot afford to have their username leaked to the git repo that way.

### inspectionProfiles

I think that inspection profiles are at least as useful to share as checkstyle rules, and we have a standard set of those.

I also acknowledge there might be some things in there we don't universally agree upon. If that's the case, I'd be okay with ignoring this until the next time (as yet unspecified) that we talk about coding standards.

### libraries-with-intellij-classes.xml

Ah, it's not just me then! Agreed.

(As near as I can tell, it came in with 2021.1? Or maybe a plugin? Not that it matters for this PR, but I want to file an upstream issue about it.)

4Denthusiast:
@keturn What would you suggest we do about the dictionaries, then? Just have everyone either add a dictionary with their name (even if it's blank) or manually ignore it? Even if a shared dictionary of Terasology jargon to use for Intellij's spellchecking is something we want to have, I think that gitignoring that directory is still the right thing to do. The only file currently in there on Github is kevint.xml (I assume that's yours), which has only 5 words in it, so people clearly haven't been sharing these as it's currently set up. I generally assume that we'll be using so much jargon and general non-standard terminology that setting up a dictionary and attempting to use spell-check just isn't worth the effort.

I don't really know what all the stuff in inspectionProfiles is. I'm assuming that as it was automatically generated and there isn't very much in there, it's fine if people start without it. I think I vaguely remember turning off spell-check at some point, and that setting is stored in there, but the other one have less obvious names.

keturn:
Hrm. I did overlook the "even if it's blank" part of the scenario. Mine has been in there for so long that I _thought_ it only shows up after you add a word to it, but it sounds like I was wrong about that. (Or maybe that's how it used to work, but doesn't anymore.)

I agree that having everyone check in _blank_ files there would be rubbish.

In which case, yeah, if you delete your dictionary file and you find it just re-generates another empty one next time you restart, then I guess we have no choice but to ignore these.

:sigh:

4Denthusiast:
> In which case, yeah, if you delete your dictionary file and you find it just re-generates another empty one next time you restart, then I guess we have no choice but to ignore these.

I tried just deleting it now, and it hasn't reappeared. I wouldn't have thought I would have created and then emptied a dictionary at some point in the past, so I'm not sure what's happening differently this time. It looks like just not adding words to the dictionary might be an alternative option. I'll remove that bit of the change.

keturn (approved):

PR #4632, opened 2021-04-21
logistics: Replace usage of jcenter (shutting down)

### LINKED ISSUE #4582: remove JCenter from all build configurations
JCenter is [shutting down](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/).

They will stop accepting new uploads _this month_ (the end of March 2021) and will stop serving downloads entirely in 2022.

So by 2022, we'll want to have removed all the `jcenter()` entries from the repository configurations in our gradle files. (Both `build.gradle.*` and everything in `build-logic`.)

The reason we didn't just do that _immediately_ is because we don't know if all our dependencies are available on other servers yet!

The process for fixing this is something like:

* remove `jcenter` from our configurations
  - replace it with `mavenCentral()` if that config doesn't have it already
* attempt to build. Note which dependencies fail.
* for each one, investigate
  - has that project moved to another server that we need to add to our configuration?
  - have they announced a _plan_ to move, but aren't finished yet?
  - is the project without a maintainer?
    + has it been picked up by a new organization?
    + should we mirror its old releases to our own repository?
    + should we maintain a fork of the project?
    + should we drop the dependency and use something else?

Related reading:
* general documentation on gradle repository configuration: https://docs.gradle.org/current/userguide/declaring_repositories.html
* gradle's instructions on how to migrate your dependencies: https://blog.gradle.org/jcenter-shutdown

Cervator:
### Contains

Replaces/removes `jcenter` usage. Fixes #4582, for the engine repo at least, both regular Gradle bits and a Groovy hit.

There are / will be a few more PRs for other repos, that's tracked decently in Trello even if the engine issue auto-closes with this PR getting merged.

Seems to work after local testing and http://jenkins.terasology.io/teraorg/job/Nanoware/job/Terasology/job/jCenter/

Also tested it in that develop branch to get a build harness to try with the Sample module, just in case: http://jenkins.terasology.io/teraorg/job/Nanoware/job/TerasologyModules/job/H/job/Sample/job/develop/6/

### How to test

* Make sure the project can still be built from scratch using Gradle, with `--refresh-dependencies` thrown in or the local Gradle cache deleted
* Probably also try a super duper clean build of some sort in IntelliJ? Not sure how to make it force-refresh dependencies tho
* Make sure `groovyw` utility commands using Git actions still work

You can also confirm that Gradle can still fetch all the dependency things by adding this little snippet to, say, the `build.gradle` in the engine dir:

```
task getDeps(type: Copy) {
    from sourceSets.main.runtimeClasspath
    into 'pullRuntimeClasspathDeps/'
}
```

`engine/pullRuntimeClasspathDeps` should end up containing a bunch of things if you run `gradlew engine:getDeps` - including the `trove4j-3.0.3.jar` that had been brought up at some point as possibly causing trouble. It seemed to fetch no problem and does exist at the right version in Maven Central, so I dunno what the problem was there.

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -8,7 +8,7 @@ plugins {
 }
 
 repositories {
-    jcenter()
+    mavenCentral()
 
     maven {
         name = "Terasology Artifactory"
--- a/build-logic/src/main/kotlin/terasology-repositories.gradle.kts
+++ b/build-logic/src/main/kotlin/terasology-repositories.gradle.kts
@@ -5,13 +5,11 @@ import java.net.URI
 
 // We use both Maven Central and our own Artifactory instance, which contains module builds, extra libs, and so on
 repositories {
-    // External libs - jcenter is Bintray and a superset of Maven Central
-    jcenter {
+    mavenCentral {
         content {
             // This is first choice for most java dependencies, but assume we'll need to check our
             // own repository for things from our own organization.
-            // (This is an optimization so gradle doesn't try to find our hundreds of modules
-            // in jcenter.)
+            // (This is an optimization so gradle doesn't try to find our hundreds of modules in 3rd party repos)
             excludeGroupByRegex("""org\.terasology(\..+)?""")
         }
     }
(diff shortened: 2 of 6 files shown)


Cervator:
[Replace jCenter in all repositories](https://trello.com/c/jMtFGegz/59-replace-jcenter-in-all-repositories)

keturn (approved):
Looks good, compiles, runs, and groovyw works as well. :+1:

PR #4634, opened 2021-04-24
feature(rendering)!: add vertex attribute scheme

pollend:
Here is a generalized scheme that maps attributes to a given vertex resource. The attributes would be defined per hardware implementation such as vulkan or opengl. The attribute maps the value correctly into the buffer so hopefully this should generalize well but I could be wrong. I also just write the data straight to the buffer and then pass the contents of the buffer straight to opengl. Should be more efficient since the data is backed by the buffer rather then an array. arbitrary changes to the buffer should let us quickly update the vertex data without rebuilding a new buffer.

Ideally would like to store the vbo pointer with the resource so the buffered data is not duplicated per mesh instance. still not sure if this attribute scheme is the right way to go. 

I reverted some of the work for skeletal mesh. this should shrink the scope of this work. 
https://github.com/MovingBlocks/Terasology/pull/4627

I would like to build up a scheme of these types of resources and they should be a thin shim for the underlying data. would like to make a variant for FBO, textures etc ...


pollend:
[Resource Attribute VBO Scheme](https://trello.com/c/PnL56rG2/97-resource-attribute-vbo-scheme)

engine/src/main/java/org/terasology/engine/rendering/opengl/OpenGLMeshBase.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    default void updateState(VBOContext state) {
+        GL30.glBindBuffer(GL30.GL_ARRAY_BUFFER, state.vbo);
+        for (int x = 0; x < state.entries.length; x++) {
+            if (state.entries[x].version != state.entries[x].resource.getVersion()) {
+                GL30.glBufferSubData(GL30.GL_ARRAY_BUFFER, state.entries[x].offset, state.entries[x].resource.buffer);
+            }
+        }
+        GL30.glBindBuffer(GL30.GL_ARRAY_BUFFER, 0);
```
pollend, line 30: trying to find a mismatch with the marked vbo so the data is correctly updated during the drawing pass.

pollend:
I'm just going to use this as reference. there are some problems that I was not able to address

pollend:
I though about this some more and I decoupled the resource and the binding itself. so the binding informs how the data is written to the buffer. the buffer should autosize to accommodate the binding which I think is a better scheme.

pollend:
[Expirmental rework of VertexData](https://trello.com/c/GQsdSPZI/110-expirmental-rework-of-vertexdata)

pollend:
looks like DirectByteBuffers has close to no difference between a fixed array. so the binding is read and written directly to the bytebuffer and cuts the transfer that happens where we take an array/allocate a buffer and copy it over. there are some benefits to doing it this way. updating the vertex data is a simple direct memory transfer to opengl so there is no need to allocate and prefill a buffer when transferring data. I think we can get a significant performance boost if we setup something similar for chunk data. 

http://imranrashid.com/posts/profiling-bytebuffers/

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/VertexDefinition.java
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.engine.rendering.assets.mesh;
+
+import org.terasology.engine.rendering.assets.mesh.resource.VertexAttribute;
+import org.terasology.engine.rendering.assets.mesh.resource.VertexResource;
+
+/**
+ *
+ */
```
skaldarnar, line 11: This might be obvious for someone who knows their way around graphics programming, but for me Javadoc would be really helpful. 

If this is a very common pattern, maybe we can give cross links (to something like OpenGL specs or docs or whatever).

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/IndexResource.java
```diff
@@ -0,0 +1,58 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.engine.rendering.assets.mesh.resource;
+
+import org.lwjgl.BufferUtils;
+import java.nio.ByteBuffer;
+
+public class IndexResource {
```
skaldarnar, line 9: Same here - what is an "index resource"?

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/VertexAttribute.java
```diff
@@ -0,0 +1,64 @@
...
+import org.terasology.nui.Color;
+
+import java.nio.ByteBuffer;
+
+/**
+ * attribute maps a target object or a primitive data to a {@link VertexResource}
+ *
+ * @param <TARGET> the target object
+ */
+public class VertexAttribute<TARGET> {
```
skaldarnar, line 19: Hm, this reads funny. I guess I'm just not used to having full words :scream: as type parameters... I guess you'd typically just use `T` in Java. What is the reason you chose `TARGET`?

engine/src/main/java/org/terasology/engine/rendering/assets/skeletalmesh/SkeletalMeshDataBuilder.java
```diff
@@ -53,33 +53,33 @@ public SkeletalMeshDataBuilder addWeights(List<BoneWeight> weights) {
...
+//            vertices.add(pos);
+//            weights.add(weight);
+//            uvs.add(new Vector2f(texCoord0.get(i * 2), texCoord0.get(i * 2 + 1)));
+//        }
+//
+//        for (int i = 0; i < meshIndices.size(); i++) {
+//            indices.add(meshIndices.get(i) + weightsStart);
+//        }
+//        return this;
+//    }
```
skaldarnar, line 82: Why is this commented out? Is it no longer needed? If so, just remove it.

engine/src/main/java/org/terasology/engine/rendering/nui/internal/LwjglCanvasRenderer.java
```diff
@@ -86,7 +86,7 @@
     private DisplayDevice displayDevice;
     private float uiScale = 1f;
 
-    private Matrix4fStack modelMatrixStack = new Matrix4fStack(1000);
+    private Matrix4fStack modelMatrixStack = new Matrix4fStack(10000);
```
skaldarnar, line 89: two magic numbers :eyes: what is this change good for?

skaldarnar (commented):
Left a few comments, but I have absolutely not idea about the code changes itself. I would feel better if anybody from the 3d wizardry department had at least a brief look whether the changes "go in the right direction" - maybe @vampcat has a bit of time? 

In other news, I checked out this PR and started it in a local singleplayer game of Iota. Everything looks fine as far as I can tell.

- **OS** Linux Mint 20 Cinnamon
- **Linux Kernel** 5.4.0
- **CPU** AMD Ryzen 7 3700X
- **Memory** 32 GB
- **GPU** GeForce GTX 1070 (450.119.03)

engine/src/main/java/org/terasology/engine/rendering/nui/internal/LwjglCanvasRenderer.java
pollend: Matrix4fStack is preallocated so the matrices are reused. we need to allocate something ahead of time.

engine/src/main/java/org/terasology/engine/core/subsystem/headless/assets/HeadlessMesh.java
```diff
@@ -45,8 +46,13 @@ public AABBfc getAABB() {
...
     @Override
-    public TFloatList getVertices() {
-        return data.getVertices();
+    public VertexAttributeBinding<Vector3f>  getVertices() {
+        return data.verts();
+    }
+
+    @Override
+    public int getVertexCount() {
+        return 0;
```
vampcat, line 56: Why is this always returning 0?

engine/src/main/java/org/terasology/engine/rendering/assets/skeletalmesh/SkeletalMeshDataBuilder.java
pollend: the concept is you can build a static mesh and attach it to an existing bone :? kind of strange and niche and apparently unused so I dropped it. I want to refactor this to use a vertex shader to instead of constantly updating vertex data on the CPU.

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/MeshBuilder.java
```diff
@@ -116,12 +108,14 @@ public MeshBuilder addTexCoord(Vector2fc v) {
     }
 
     public MeshBuilder addIndex(int index) {
-        meshData.getIndices().add(index);
+        meshData.indices.put(index);//.add(index);
```
vampcat, line 111: Remove the commented out part?

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/IndexResource.java
```diff
@@ -0,0 +1,85 @@
...
+    public void reserveElements(int elements) {
+        ensureCapacity(elements * Integer.BYTES);
+    }
+
+    public void rewind() {
+        posIndex = 0;
+    }
+
+    public void put(int value) {
+        ensureCapacity((posIndex + 1) * Integer.BYTES);
```
vampcat, line 55: Is it possible for this function to be called multiple times, say in a loop? If yes, then wouldn't it make more sense to not put it here/have an alternate overload that takes a list and only performs the check once for the entire list?

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/VertexAttribute.java
```diff
@@ -0,0 +1,58 @@
...
+import org.lwjgl.opengl.GL30;
+
+/**
+ * attribute maps a target object or a primitive data to a {@link VertexResource}
+ *
+ * @param <T> the target object
+ */
+public class VertexAttribute<T, TImpl extends T> {
+
+
```
vampcat, line 15: The newlines in this file looks a bit wonky?

vampcat (commented):
Not a review, just a brief gloss-over of the top 20 files or so

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/IndexResource.java
pollend: the check should be quick. the buffer is doubled each time you overflow its content. its dynamic in how the data is getting allocated.

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/VertexAttribute.java
pollend: yep fixed it :D

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/VertexAttribute.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    protected VertexAttribute(Class<TImpl> type, AttributeConfiguration<T, TImpl> attributeConfiguration, TypeMapping mapping, int count) {
+        this.type = type;
+        this.mapping = mapping;
+        this.count = count;
+        this.configuration = attributeConfiguration;
+    }
+
+
+    public enum TypeMapping {
```
DarkWeird, line 43: I think this enum should be external - can this to use by another non-vertex data?

engine/src/main/java/org/terasology/engine/rendering/primitives/Tessellator.java
```diff
@@ -157,22 +124,16 @@ public void setLighting(Vector3f v) {
...
+    public MeshData buildMeshData() {
+        return meshData.clone();
     }
 
     public Mesh generateMesh(ResourceUrn urn) {
         Preconditions.checkNotNull(urn);
-        Mesh result = Assets.generateAsset(urn, meshData, Mesh.class);
-        meshData = new StandardMeshData();
-        return result;
+        return Assets.generateAsset(urn, buildMeshData(), Mesh.class);
```
DarkWeird, line 133: Previous behavior resets Tessellator's state. 
You could reuse tessallator for different objects. you cannot do it now. (needs additional review for usages)

pollend: There is a clove method in build mesh data

DarkWeird: But `clone()` not resets `meshData` state... or resets? if yes - then it is not clone. (imho clone should be pure)

pollend: Umm should I use a different verb. I would need to duplicate the state for the bindings.

jdrueckert:
This currently does not compile. I see the same issues locally that are also reported by the CI

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/Mesh.java
```diff
@@ -22,34 +24,26 @@ protected Mesh(ResourceUrn urn, AssetType<?, MeshData> assetType, DisposableReso
 
     public abstract AABBfc getAABB();
 
-    protected AABBf getBound(MeshData data, AABBf dest) {
-        TFloatList vertices = data.getVertices();
-        int vertexCount = vertices.size() / 3;
-        if (vertexCount == 0) {
+    protected AABBf getBound(AABBf dest) {
+        VertexAttributeBinding<Vector3fc, Vector3f> vertices = this.getVertices();
+        if (vertices.getResource().elements() == 0) {
```
skaldarnar, line 29: Should this use the abstraction for `getVertexCount()` here fore consistency?

```suggestion
        if (this.getVertexCount() == 0) {
```

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/Mesh.java
```diff
@@ -22,34 +24,26 @@ protected Mesh(ResourceUrn urn, AssetType<?, MeshData> assetType, DisposableReso
...
-        dest.minZ = vertices.get(2);
-        dest.maxX = vertices.get(0);
-        dest.maxY = vertices.get(1);
-        dest.maxZ = vertices.get(2);
-
-        for (int index = 1; index < vertexCount; ++index) {
-            dest.union(vertices.get(3 * index), vertices.get(3 * index + 1), vertices.get(3 * index + 2));
-
+        Vector3f pos = new Vector3f();
+        for (int x = 0; x < vertices.getResource().elements(); x++) {
```
skaldarnar, line 39:
```suggestion
        for (int x = 0; x < this.getVertexCount(); x++) {
```

engine/src/main/java/org/terasology/engine/physics/bullet/BulletPhysics.java
```diff
@@ -743,16 +744,16 @@ private btCollisionShape getShapeFor(EntityRef entityRef) {
...
-            FloatBuffer buffer = BufferUtils.createFloatBuffer(hull.sourceMesh.getVertices().size());
-            TFloatIterator iterator = hull.sourceMesh.getVertices().iterator();
-            int numPoints = 0;
-            while (iterator.hasNext()) {
-                numPoints++;
-                buffer.put(iterator.next());
-                buffer.put(iterator.next());
-                buffer.put(iterator.next());
+            VertexAttributeBinding<Vector3fc, Vector3f> positions = hull.sourceMesh.getVertices();
+            FloatBuffer buffer = BufferUtils.createFloatBuffer(positions.getResource().elements() * 3);
```
skaldarnar, line 748: Should this use the abstraction for `getVertexCount()` rather than looking at the resource's elements directly? As this number is used later again, maybe even save it to a local field?

```suggestion
            int vertexCount = hull.sourceMesh.getVertexCount();
            FloatBuffer buffer = BufferUtils.createFloatBuffer(vertexCount * 3);
```

engine/src/main/java/org/terasology/engine/physics/bullet/BulletPhysics.java
```diff
@@ -743,16 +744,16 @@ private btCollisionShape getShapeFor(EntityRef entityRef) {
...
+            Vector3f pos = new Vector3f();
+            for (int i = 0; i < positions.getResource().elements(); i++) {
+                positions.get(i, pos);
+                buffer.put(pos.x);
+                buffer.put(pos.y);
+                buffer.put(pos.z);
             }
-            return new btConvexHullShape(buffer, numPoints, 3 * Float.BYTES);
+            return new btConvexHullShape(buffer, positions.getResource().elements(), 3 * Float.BYTES);
```
skaldarnar, line 756:
```suggestion
            return new btConvexHullShape(buffer, vertexCount, 3 * Float.BYTES);
```

skaldarnar (commented):
first brief round of review, only looked a handful of files yet...

jdrueckert:
With #4735 being merged into this, I left https://github.com/Terasology/CoreRendering/pull/59 still open

engine/src/main/java/org/terasology/engine/core/subsystem/headless/assets/HeadlessMesh.java
```diff
@@ -32,8 +34,13 @@ public AABBfc getAABB() {
...
 
     @Override
-    public TFloatList getVertices() {
-        return data.getVertices();
+    public VertexAttributeBinding<Vector3fc, Vector3f> vertices() {
+        return data.positions();
+    }
+
+    @Override
+    public int elements() {
```
skaldarnar, line 42: I don't think `elements` is a good name for something that returns a number, but maybe this is the right name for this domain 🤷  I'd prefer something like `elementCount` to emphasize that this is a number, opposed to `elements` which, to me, suggests that this method returns the elements of this mesh (as in, vertices, edges, or whatever the "elements" of a mesh are).

engine/src/main/java/org/terasology/engine/rendering/AABBRenderer.java
```diff
@@ -178,25 +178,22 @@ public Vector2fc map(int vertexIndex, float u, float v) {
...
+        Vector3f pos = new Vector3f();
+        meshData.position.put(pos.set(-dimensions.x - offset, -dimensions.y - offset, -dimensions.z - offset)); // 0
+        meshData.position.put(pos.set(+dimensions.x + offset, -dimensions.y - offset, -dimensions.z - offset)); // 1
+        meshData.position.put(pos.set(+dimensions.x + offset, -dimensions.y - offset, +dimensions.z + offset)); // 2
+        meshData.position.put(pos.set(-dimensions.x - offset, -dimensions.y - offset, +dimensions.z + offset)); // 3
+
+        meshData.position.put(pos.set(-dimensions.x - offset, +dimensions.y + offset, -dimensions.z - offset)); // 4
+        meshData.position.put(pos.set(+dimensions.x + offset, +dimensions.y + offset, -dimensions.z - offset)); // 5
+        meshData.position.put(pos.set(+dimensions.x + offset, +dimensions.y + offset, +dimensions.z + offset)); // 6
+        meshData.position.put(pos.set(-dimensions.x - offset, +dimensions.y + offset, +dimensions.z + offset)); // 7
```
skaldarnar, line 194: this passes in the same vector object with changed values each time. with a quick search, it was hard for me to see whether we store this element directly and would override previous states, or whether we do a "copy-on-write" somewhere. It would help a lot if we could find a way to document this somewhere (here, on the `put` method of the `VertexAttribute`, I think).

engine/src/main/java/org/terasology/engine/rendering/AABBRenderer.java
```diff
@@ -215,8 +212,10 @@ private void generateDisplayListWire() {
                 6, 7,
                 7, 4,
         });
+
+        Color color = new Color(Color.black);
         for (int i = 0; i < 8; i++) {
-            meshData.color0.addAll(new float[]{0.0f, 0.0f, 0.0f, 1.0f});
+            meshData.color0.put(color);
```
skaldarnar, line 218: Why do we need to create a new instance of `color` here beforehand, and can't use 
```suggestion
            meshData.color0.put(Color.black);
```
here? Also, do we do a "copy-on-write" here? Otherwise, is it correct that we put the same color instance for all 8 indices?

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/DrawingMode.java
```diff
@@ -0,0 +1,27 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.engine.rendering.assets.mesh.resource;
+
+import org.lwjgl.opengl.GL30;
+import org.lwjgl.opengl.GL33;
+import org.lwjgl.opengl.GL44;
+
+public enum DrawingMode {
```
skaldarnar, line 10: Extracted from `MeshData.java`.

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/Mesh.java
```diff
@@ -22,34 +24,26 @@ protected Mesh(ResourceUrn urn, AssetType<?, MeshData> assetType, DisposableReso
 
     public abstract AABBfc getAABB();
 
-    protected AABBf getBound(MeshData data, AABBf dest) {
-        TFloatList vertices = data.getVertices();
-        int vertexCount = vertices.size() / 3;
-        if (vertexCount == 0) {
+    protected AABBf getBound(AABBf dest) {
+        VertexAttributeBinding<Vector3fc, Vector3f> vertices = this.vertices();
+        if (elements() == 0) {
```
skaldarnar: Please use (or omit) `this` consistently 😉

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/ObjMeshFormat.java
```diff
@@ -50,63 +48,50 @@ public MeshData load(ResourceUrn urn, List<AssetDataFile> inputs) throws IOExcep
...
         }
     }
 
     private StandardMeshData processData(List<Vector3f> rawVertices, List<Vector3f> rawNormals, List<Vector2f> rawTexCoords, List<Vector3i[]> rawIndices) throws IOException {
+        int numIndices = 0;
+        int numVerts = 0;
+        for (int x = 0; x < rawIndices.size(); x++) {
+            numIndices += (rawIndices.get(x).length - 2) * 3;
+            numVerts += rawIndices.get(x).length;
+        }
```
skaldarnar, line 61: How often is this code called/how performance critical is this? I personally would prefer using a more functional approach using the Streams API:

```java
        final int numVerts = rawIndices.stream().map(x -> x.length).reduce(0, Integer::sum);
        final int numIndices = rawIndices.stream().map(x -> (x.length - 2) * 3).reduce(0, Integer::sum);
```

And instead of doing the computation for `numIndices` inside the loop we could do the math and only do it at the end, can't we? This would only do one stream iteration then.

```java
final int numIndices = (numVerts - 2 * rawIndices.size()) * 3;
```

pollend: umm not that much

pollend: lets move this to another PR. its non critical to how this is suppose to work.

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/IndexResource.java
```diff
@@ -0,0 +1,83 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.engine.rendering.assets.mesh.resource;
+
+/**
+ * defines the order of vertices to walk for rendering geometry
+ *
+ * refrence: https://www.khronos.org/opengl/wiki/Primitive
```
skaldarnar, line 9: :+1:

engine/src/main/java/org/terasology/engine/rendering/assets/skeletalmesh/SkeletalMeshDataBuilder.java
```diff
@@ -50,34 +50,6 @@ public SkeletalMeshDataBuilder addWeights(List<BoneWeight> weights) {
         return this;
     }
 
-    public SkeletalMeshDataBuilder addMesh(Bone bone, MeshBuilder builder) {
-        return addMesh(bone, builder.getMeshData());
-    }
-
-    public SkeletalMeshDataBuilder addMesh(Bone bone, StandardMeshData data) {
```
skaldarnar, line 57: where did this go?

engine/src/main/java/org/terasology/engine/rendering/opengl/OpenGLMeshBase.java
```diff
@@ -0,0 +1,79 @@
...
+            state.entries[i].offset = offset;
+
+            int currentOffset = offset;
+            resource.writeBuffer((buffer) -> {
+                GL30.glBufferSubData(GL30.GL_ARRAY_BUFFER, currentOffset, buffer);
+            });
+//            ByteBuffer vertexBuffer = resource.buffer();
+//            vertexBuffer.rewind();
+//            vertexBuffer.limit(resource.inSize());
+//            GL30.glBufferSubData(GL30.GL_ARRAY_BUFFER, offset, vertexBuffer);
```
skaldarnar, line 66: why is this commented?

engine/src/main/java/org/terasology/engine/rendering/primitives/Tessellator.java
skaldarnar: @pollend @DarkWeird is this resolved?

engine/src/main/resources/org/terasology/engine/assets/shaders/shadowMap_frag.glsl
```diff
@@ -1,21 +1,10 @@
...
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+#version 330 core
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+//varying vec4 positionProj;
```
skaldarnar, line 4:
```suggestion
```

skaldarnar (approved):
Tested this together with https://github.com/Terasology/CoreRendering/pull/59 and the games looks quite the same. I could not spot any difference in performance, either. Tested with Iota workspace.

I'll approve this, but @pollend please feel free to do small changes and ping me for a re-approve.

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/StandardMeshData.java
```diff
@@ -16,44 +24,123 @@
     public static final int COLOR0_INDEX = 4;
     public static final int LIGHT0_INDEX = 5;
 
+    public final VertexResource positionBuffer;
+    public final VertexAttributeBinding<Vector3fc, Vector3f> position;
```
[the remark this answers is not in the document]
pollend: - Attribute is what is describing the format of the data
- AttributeBinding is how that data is getting mapped for that attribute
- VertexResource is just a basic buffer for how that data is getting stored. The bare minimum of describing this data i.e a stride a buffer and a size. The idea is for vulkan we would have a set of attributes describing vulkan data types and the binding themselves would stay consistent. have some kind of VertexFactory to build the data type for the end user to use but that is something to worry about later.

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/VertexResourceBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.terasology.engine.rendering.assets.mesh.resource;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * factory that is used to define an attribute binding to the backing {@link VertexResource}
+ */
+public class VertexResourceBuilder {
```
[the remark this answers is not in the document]
pollend: maybe a VertexResourceFactory?

engine/src/main/java/org/terasology/engine/rendering/assets/mesh/resource/TypeMapping.java
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.engine.rendering.assets.mesh.resource;
+
+import org.lwjgl.opengl.GL30;
+
+public enum TypeMapping {
```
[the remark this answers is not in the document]
pollend: I was going to pack more types into this enum for different graphics API's have a unified enum for different mappings but maybe that is the wrong approach.

engine/src/main/java/org/terasology/engine/rendering/assets/skeletalmesh/SkeletalMeshDataBuilder.java
pollend: we don't use it in any capacity It was just a way to add meshes to bones.  I guess I would want to redo this as a skinned mesh so we're not doing this all on the CPU.

pollend:
> Tested this together with [Terasology/CoreRendering#59](https://github.com/Terasology/CoreRendering/pull/59) and the games looks quite the same. I could not spot any difference in performance, either. Tested with Iota workspace.
> 
> I'll approve this, but @pollend please feel free to do small changes and ping me for a re-approve.
> 
> Still to be checked: does this work with the falling block animation?

kind of hard to tell with other larger bottle necks in the pipeline. The pipeline can only run as fast as its slowest stage :/.

skaldarnar (approved):

PR #4635, opened 2021-04-24
chore: remove deprecated ProtobufSerializer

keturn:
ProtobufSerializer has been deprecated for a while, replaced by the generic Serializer interface.

This removes the class and the few places where it was still mentioned in tests.

I figured I could do this split out from the gestalt-v7 branch, but it looks like these tests skip themselves in develop, so :shrug:


pollend:
this seems pretty straightforward.

pollend (approved):

pollend:
@keturn I updated the branch did I do anything incorrectly else I think this could be merged.

PR #4645, opened 2021-05-02
build(modules): improve matching of compatible versions in module deps

keturn:
by avoiding use of gestalt.VersionRange

I'm not too worried about this PR needing review, but having build-related things seen by another set of eyes is always good.

the package-info in engine-test isn't really related, it's a stowaway

### DIFF
--- a/build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
+++ b/build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
@@ -8,6 +8,7 @@ import org.terasology.gestalt.module.ModuleMetadata
 import org.terasology.gestalt.module.ModuleMetadataJsonAdapter
 import org.terasology.gestalt.module.dependencyresolution.DependencyInfo
 import org.terasology.gestalt.naming.Version
+import org.terasology.gestalt.naming.VersionRange
 import java.io.File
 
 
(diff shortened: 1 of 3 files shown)


build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
```diff
@@ -72,7 +73,15 @@ class ModuleMetadataForGradle(private val moduleConfig: ModuleMetadata) {
...
         }
 
-        val gradleDep = GradleDependencyInfo(TERASOLOGY_MODULES_GROUP, gestaltDependency.id.toString(), gestaltDependency.versionRange().toString())
+        val version = if (gestaltDependency.versionPredicate() is VersionRange) {
+            gestaltDependency.versionPredicate().toString()
+        } else {
+            // TODO: gradle-compatible version expressions for gestalt dependencies
+            //     https://github.com/MovingBlocks/gestalt/issues/114
+            gestaltDependency.minVersion.toString();
+        }
```
pollend, line 82: I'm kind of confused how does this affect the logic? seems to output the results toString and discard?

keturn: `if` can be used as an expression in kotlin. Like a more verbose `? :` ternary operator.

pollend: oh I see, that is interesting.

pollend (approved):

build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
skaldarnar: that's super nice, and something I'm totally missing in Java 🤓

Issue #4638, opened 2021-05-01
allow instantiating ModuleManager without scanning `modules/`


keturn:
A lot of work happens in the constructor of [ModuleManager](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java), including scanning everything in `/modules/*`.

This is one reason tests get slow and noisy in a large workspace.

Most tests aren't written with other modules in `/modules/` in mind. Tests should be able to construct a ModuleManager or even a TerasologyEngine without this happening.

Some implementation possibilities:
* add a way to set `PathManager.getModulePaths` to an empty list.
* add a way to inject a different PathManager to ModuleManager (instead of it using PathManager.getInstance)
* …

PR #4648, opened 2021-05-03
test: provide a stub PathManager during tests

keturn:
This is for #4638. Usage looks like this

```java
@ExtendWith({MockitoExtension.class, PathManagerProvider.class})
public class FooTest {

    @Test
    public void testThisDoesThisAndNotThat(PathManager pathManager) {
        when(pathManager.getModulePaths()).thenReturn(Collections.emptyList());
        manager = new ModuleManager("");
```

the implementation feels like a _bunch_ of code for the sake of setting a value and cleaning it up at the end of the test. I kind of hate it and hope someone will tell me how I'm over-complicating things and what we could do instead.


engine-tests/src/main/java/org/terasology/engine/core/PathManagerProvider.java
```diff
@@ -0,0 +1,52 @@
...
+                getClass(), extensionContext.getRequiredTestInstance());
+        PathManager newPathManager = mock(PathManager.class);
+        PathManager oldPathManager = setPathManager(newPathManager);
+
+        extensionContext.getStore(namespace).put("oldThing", new Cleaner(oldPathManager, newPathManager));
+
+        return newPathManager;
+    }
+
+    static class Cleaner implements ExtensionContext.Store.CloseableResource {
```
keturn, line 37: This is to set PathManager.instance back to its original value at the end of the test.

It seems like the polite thing to do, and I'm also worried about the prospect of leaving a Mockito instance there and having it leak to other tests.

engine/src/main/java/org/terasology/engine/core/PathManager.java
```diff
@@ -1,7 +1,7 @@
 // Copyright 2021 The Terasology Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-package org.terasology.engine.core.paths;
+package org.terasology.engine.core;
```
keturn, line 4: oh yeah. All the import changes are because I decided that PathManager doesn't deserve its own sub-package. It was the only thing in `engine.core.paths`.

engine-tests/src/main/java/org/terasology/engine/core/PathManagerProvider.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+
+import static org.mockito.Mockito.mock;
+
+public class PathManagerProvider implements ParameterResolver {
+    private PathManagerProvider() {};
```
keturn, line 14: wait, does this really work with the default constructor being private? that doesn't seem right

pollend: don't you do this if you want to have a singleton object. is there a method to produce an instance :?.

jdrueckert: `PathManager::getInstance` is the method that creates the instance in a singleton fashion I guess.
@pollend why would one do `private PathManagerProvider() {};` if one would want to have a singleton object? (this is just me not knowing much about java I guess)

skaldarnar: :books: _hands over Gang of Four book to Nira ..._

You want to ensure that there is only one instance of this class, so you need to restrict who can create instances. The typical (Java) pattern here is to make the (or rather all) constructor private to that class, and than offer a static method to get the single instance `getInstance`. This static method is allowed to call the private constructor, so you're good to go.

(there are numerous SO threads on the topic on whether this is thread safe, etc)

engine/src/main/java/org/terasology/engine/core/PathManager.java
```diff
@@ -140,6 +140,12 @@ public static PathManager getInstance() {
         return instance;
     }
 
+    static PathManager setInstance(PathManager pathManager) {
```
skaldarnar, line 143: This weakens the property of it being a singleton :thinking: Sure, this is not publicly visible, but I'd at least would like to see a JavaDoc comment to emphasize that this is for testing-only.

I guess the "right thing" to do here would be to not make it a singleton at all but rather have a "dummy factory" that produces exactly on instance in normal operation, but can be used to produce stubs when testing...

engine-tests/src/main/java/org/terasology/engine/core/PathManagerProvider.java
```diff
@@ -0,0 +1,52 @@
...
+package org.terasology.engine.core;
+
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+
+import static org.mockito.Mockito.mock;
+
+public class PathManagerProvider implements ParameterResolver {
```
skaldarnar, line 13: TIL about parameter resolvers being a thing. 

I'd highly appreciate docstrings for this class to explain what it does and how to use it.

engine-tests/src/main/java/org/terasology/engine/core/PathManagerProvider.java
skaldarnar: To follow up on the code at hand: my guess would be on JUnit/Mockito magic around reflection and making impossible things possible for testing :mage_man: ...

keturn:
> [`PathManager.setInstance`] weakens the property of it being a singleton 🤔 Sure, this is not publicly visible, but I'd at least would like to see a JavaDoc comment to emphasize that this is for testing-only.
>
> I guess the "right thing" to do here would be to not make it a singleton at all but rather have a "dummy factory" that produces exactly on instance in normal operation, but can be used to produce stubs when testing...

Possibilities for moving forward:
* continue with the approach currently in this PR (package-private `PathManager.setInstance` used by a JUnit ParameterResolver), adding documentation & etc
* make it not-a-singleton somehow
  + Different factory, as in skaldarnar's suggestion?
  + put a PathManager in the Context, and make sure a context is passed to everything that currently calls PathManager.getInstance?
* something else? ▁▁▁▁▁▁▁▁▁

Having now looked at it, do you have any preferences on one way or the other?

skaldarnar:
I would be fine with keeping the current approach of this PR with documentation and disclaimers. 

Not really sure what the future plans for `Context` and Co are, so maybe we can add a `//TODO:` to schedule it for having the `PathManager` as subject of any DI efforts in the future.

skaldarnar (approved):
looks good to me! :+1:

jdrueckert (approved):

PR #4649, opened 2021-05-07
build(module_build): use same DependencyInfo -> String logic for modules and engine

skaldarnar:
**Problem:** In https://github.com/MovingBlocks/gestalt/pull/115 we changed the semantics of module version resolution to no longer include the SNAPSHOT to satisfy a non-snapshot `minVersion`. We also removed the `DependencyInfo::versionRange` accessor as it seemed to be no longer in use.

However, we missed that it's actually used in Terasology's module builds.

**Solution:** This PR unifies the way a String representation of the version range is derived from `DependencyInfo`, no longer accessing the now missing `versionRange()` method. 

In addition, it cleans up some of the build file (one corner-case handling became obsolete).

- build(module_build): use same DependencyInfo -> String logic for modules and engine
- build(module_build): remove corner case handling for min SNAPHSOT

### DIFF
--- a/build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
+++ b/build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
@@ -48,9 +48,11 @@ class ModuleMetadataForGradle(private val moduleConfig: ModuleMetadata) {
     val group: String = TERASOLOGY_MODULES_GROUP
 
     fun engineVersion(): String {
-        return moduleConfig.dependencies.filterNotNull()
+        return moduleConfig.dependencies
+            .filterNotNull()
             .find { it.id.toString() == ENGINE_MODULE_NAME }
-            ?.versionRange()?.toString() ?: "+"
+            ?.let(this::versionStringFromGestaltDependency)
+            ?: "+"
     }
 
     /**
@@ -59,8 +61,19 @@ class ModuleMetadataForGradle(private val moduleConfig: ModuleMetadata) {
      * @return a list of modules and whether each is optional
      */
     fun moduleDependencies(): List<Pair<GradleDependencyInfo, Boolean>> {
-        val gestaltDeps = moduleConfig.dependencies.filterNotNull().filterNot { it.id.toString() == "engine" }
+        val gestaltDeps = moduleConfig.dependencies.filterNotNull().filterNot { it.id.toString() == ENGINE_MODULE_NAME }
         return gestaltDeps.map { gradleModule(it) }
+    }
+
+    private fun versionStringFromGestaltDependency(gestaltDependency: DependencyInfo): String {
+        val version = if (gestaltDependency.versionPredicate() is VersionRange) {
+            gestaltDependency.versionPredicate().toString()
+        } else {
+            // TODO: gradle-compatible version expressions for gestalt dependencies
+            //     https://github.com/MovingBlocks/gestalt/issues/114
+            "[${gestaltDependency.minVersion},)"                                       
+        }
+        return version;
     }
 
     private fun gradleModule(gestaltDependency: DependencyInfo): Pair<GradleDependencyInfo, Boolean> {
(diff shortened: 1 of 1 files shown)


build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
```diff
@@ -48,9 +48,11 @@ class ModuleMetadataForGradle(private val moduleConfig: ModuleMetadata) {
     val group: String = TERASOLOGY_MODULES_GROUP
 
     fun engineVersion(): String {
-        return moduleConfig.dependencies.filterNotNull()
+        return moduleConfig.dependencies
+            .filterNotNull()
             .find { it.id.toString() == ENGINE_MODULE_NAME }
-            ?.versionRange()?.toString() ?: "+"
+            ?.let(this::versionStringFromGestaltDependency)
+            ?: "+"
```
keturn, line 55: thought: genuinely not sure what to think about this usage of `?.let()`.

I know this sort of construct is one approach to these situations where we're doing chains of "only if not null" operations, but it's an area of Kotlin I still feel much less than fluent in.

The IDE tells me that we could leave off the explicit `this`:
```kotlin
return allThatStuff
    ?.let(::versionStringFromGestaltDependency)
    ?: "+"
```

or expand it back out to the block form:
```kotlin
return allThatStuff
    ?.let { versionStringFromGestaltDependency(it) }
    ?: "+"
```

I find myself preferring the block form to read, but I can definitely also appreciate the functional perspective of "this is calling exactly this function, we don't need to introduce another scope for it!"

build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
```diff
@@ -59,27 +61,23 @@ class ModuleMetadataForGradle(private val moduleConfig: ModuleMetadata) {
...
             // TODO: gradle-compatible version expressions for gestalt dependencies
             //     https://github.com/MovingBlocks/gestalt/issues/114
-            "[${gestaltDependency.minVersion},)"                                                                                                                                                                   
+            "[${gestaltDependency.minVersion},)"                                       
         }
+        return version;
+    }
+
+    private fun gradleModule(gestaltDependency: DependencyInfo): Pair<GradleDependencyInfo, Boolean> {
+        val version = versionStringFromGestaltDependency(gestaltDependency)
```
keturn, line 80: thought: noticing neither of these functions refer to fields of the class, i.e. in Java we would say "method could be static."

Kotlin doesn't have a `static` keyword for this, but classes do have what is essentially a static _companion_ object you can put things in for similar effect. (for example, see the `forProject` function in this class.)

keturn (approved):
Good change, good refactoring for the common code, and I have actually run it to test it this time :see_no_evil:

build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
skaldarnar: I was actually using VS Code for this one - maybe IntelliJ support would have helped me here. 

Personally, the block form is _less_ readable for me :thinking: but I also don't understand whether `let` is the right thing to use here. There are other words to use, I think, such as `apply`, which might fit better...

Actually, I was looking for `map` as it would be called in Scala...

build-logic/src/main/kotlin/org/terasology/gradology/module_build.kt
skaldarnar: Uh, nice! I have to learn more Kotlin! 

Scala has the [same concept](https://docs.scala-lang.org/overviews/scala-book/companion-objects.html), shouldn't be that different I guess :upside_down_face:

keturn:
Functions like `versionStringFromGestaltDependency` could also be made top-level functions, outside of any Companion object.

Or, since it only operates on a single DependencyInfo argument, we could define it as an extension method on DependencyInfo. That _would_ make the usage cleaner:

```
return allThatStuff
    ?.versionStringFromGestaltDependency()
    ?: "+"
```

but I have a nagging feeling that it is possible to go overboard with using extensions for everything. Defining an extension method that's only used once or twice might be more confusing than helpful.

PR #4651, opened 2021-05-08
Bugfix: add excluded entity for rayTrace

pollend:
looks like I might of removed part of the implementation when I migrated the logic. its not pressing to fix this right away since the old logic still work or we could just drop the method entirely. looks like we've been using this to filter out the character collider when the player is interacting with the world but not sure if its still a problem.


skaldarnar:
How would I tell the difference of before and after in-game?

pollend:
the original method was broken without this exclusion check and it didn't seem to break anything :? the only methods that use this try to exclude just the player.

jdrueckert (approved):

PR #4652, opened 2021-05-08
chore: simplify logic for getShapeFor

pollend:
The bulletBoxShape logic is a wrapper that provides the object and passes it back out but the implementation is already self contained within BulletPhyiscs. just easier to instantiate the objects as is then then introduce this extra level of indirection into the code.


skaldarnar (approved):
The changes look reasonable, but I haven't tested this locally...

PR #4657, opened 2021-05-10
chore: replace ViewFrustum with JOML FrustumIntersection

pollend:
I dropped ViewFrustum in favor of FrustumIntersection from JOML. This removes all the strange buffer mapping that we're doing to pass a matrix to ViewFrustum and simplifies some of this logic to some degree. One thing I noticed is the cameras view frustrum is coupled to the player which is strange. There is too much logic bound to the camera for it to be of any general use so we need to decouple some of this. 

ref: https://github.com/Terasology/CoreRendering/pull/40

```
Vector3f playerView = CoreRegistry.get(LocalPlayer.class).getViewPosition(new Vector3f());
```


engine/src/main/java/org/terasology/engine/rendering/cameras/Camera.java
```diff
@@ -229,6 +231,8 @@ public boolean isReflected() {
     }
 
     public boolean hasInSight(AABBfc aabb) {
-        return viewFrustum.intersects(aabb);
+        // TODO: strange offset to shift aabb by player view position
+        Vector3f playerView = CoreRegistry.get(LocalPlayer.class).getViewPosition(new Vector3f());
+        return viewFrustum.testAab(aabb.minX() - playerView.x, aabb.minY() - playerView.y, aabb.minZ() - playerView.z, aabb.maxX() - playerView.x, aabb.maxY() - playerView.y, aabb.maxZ() - playerView.z);
```
pollend, line 236: why :x: :x:

DarkWeird (approved):
Works. looks same.
(ToDo not checked)

PR #4659, opened 2021-05-11
chore: combine NativeHelper and OS

pollend:
I combined NativeHelper and OS since they share a lot of common functionality. instead of methods I made them public static finial so you shouldn't be able to modify these properties. strings are immutable and final prevents you from overwriting the value.


engine/src/main/java/org/terasology/engine/utilities/OS.java
```diff
@@ -13,14 +15,23 @@
...
 
+    public static final String OS_NAME = System.getProperty("os.name").toLowerCase();
+    public static final String ARCHITECTURE = System.getProperty("os.arch");
+    public static final String USER_DIRECTORY = System.getProperty("user.dir");
+    public static final String PATH_SEPARATOR = FileSystems.getDefault().getSeparator();
+
+    public static final boolean IS_WINDOWS = OS_NAME.toLowerCase().contains("win");
+    public static final boolean IS_MACOSX =  OS_NAME.toLowerCase().contains("mac");
+    public static final boolean IS_LINUX =  OS_NAME.toLowerCase().contains("linux") | OS_NAME.toLowerCase().contains("freebsd") || OS_NAME.toLowerCase().contains("sunos") || OS_NAME.toLowerCase().contains("unix");
+    public static final boolean IS_64 = ARCHITECTURE.endsWith("64");
```
skaldarnar, line 26: You changed the logic here to be the one from `NativeHelper` - are we sure this works the same as the previous implementation in this file? The [Platform class from the Launcher](https://github.com/MovingBlocks/TerasologyLauncher/blob/master/src/main/java/org/terasology/launcher/util/Platform.java) uses yet another logic to determine things 🙈 

Is there a best practice way to do it out there?

engine/src/main/java/org/terasology/engine/utilities/OS.java
```diff
@@ -13,14 +15,23 @@
     MACOSX,
     LINUX;
 
+    public static final String OS_NAME = System.getProperty("os.name").toLowerCase();
+    public static final String ARCHITECTURE = System.getProperty("os.arch");
+    public static final String USER_DIRECTORY = System.getProperty("user.dir");
+    public static final String PATH_SEPARATOR = FileSystems.getDefault().getSeparator();
```
skaldarnar, line 21: I don't like that we have to expose this here 🧐 Usually we would use `Paths` to construct the path in a platform-agnostic way. Is this not possible in `OpenVrProvider.java` due to module sandboxing?

engine/src/main/java/org/terasology/engine/rendering/openvrprovider/OpenVRProvider.java
```diff
@@ -279,8 +279,25 @@ private boolean initializeOpenVRLibrary() {
         if (initialized) {
             return true;
         }
-        logger.info("Adding OpenVR search path: " + NativeHelper.getOpenVRLibPath());
-        NativeLibrary.addSearchPath("openvr_api", NativeHelper.getOpenVRLibPath());
+
+        String target = OS.USER_DIRECTORY +  OS.PATH_SEPARATOR + "openvr_natives" + OS.PATH_SEPARATOR;
```
skaldarnar, line 283: I'd prefer to use `Paths...` operations here instead of manually constructing the path string. Would that be possible or is module sandboxing preventing that?

skaldarnar (commented):
@pollend I agree that merging these classes is a good idea 👍 

Left a couple of questions on the usag of file and path utility before approving.

skaldarnar (dismissed):

skaldarnar (approved):

PR #4660, opened 2021-05-11
fix(json-config-loading): Fix handling invalid AutoConfigs and config writing

### LINKED ISSUE #4642: JsonSyntaxException MalformedJsonException from AutoConfigManager.load

Start the game.

### What actually happened

Before getting to the main menu, the game crashes with 

```
[main] ERROR o.t.engine.core.TerasologyEngine - Failed to initialise Terasology
com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 4 path $
        at com.google.gson.Gson.assertFullConsumption(Gson.java:908)
        at com.google.gson.Gson.fromJson(Gson.java:871)
        at org.terasology.engine.persistence.typeHandling.gson.GsonPersistedDataReader.read(GsonPersistedDataReader.java:33)
        at org.terasology.engine.persistence.typeHandling.gson.GsonPersistedDataReader.read(GsonPersistedDataReader.java:16)
        at org.terasology.persistence.serializers.Serializer.deserialize(Serializer.java:84)
        at org.terasology.engine.config.flexible.AutoConfigManager.loadSettingsFromDisk(AutoConfigManager.java:81)
```

there may be other errors shown adjacent to this as things crash.

There's a config file that contains
```json
{}debugEnabled":true}
```

which looks like something probably opened an existing file without truncating it and wrote the empty object `{}`. 

running develop sometime after 4.3.0

DarkWeird:
### Contains

* fix invalid json config writing
* add logging when config invalid
* use defaults when config invalid
* Fixes #4642 

### How to test

Chapter 1 (fix writing, require default settings):
1. Open TS
2. Open Settings -> autoconfig
3. change one of setting (set non-default)
4. exit game
5. Open TS
6. Open Settings -> autoconfig
7. change same settings to default
8. exit game
9. check config file - content should be `{}` (all defaults)

Chapter 2 (logging and invalid config):
1. make Config file as invalid - to corrupt json structure
2. Open TS
3. Check Logs - you should see error log entry about config loading error
4. Check how renders invalid config in autoconfig - should be default
5. exit game
6. check config file - content should be `{}` (all defaults) if you don't change settings at (4)


engine/src/main/java/org/terasology/engine/config/flexible/AutoConfigManager.java
```diff
@@ -84,7 +84,7 @@ public void loadConfigsIn(Context context) {
         try (InputStream inputStream = Files.newInputStream(configPath, StandardOpenOption.READ)) {
             T loadedConfig = (T) serializer.deserialize(TypeInfo.of(configClass), inputStream).get();
             mergeConfig(configClass, loadedConfig, config);
-        } catch (IOException e) {
+        } catch (Exception e) {
```
skaldarnar, line 87: Hm, I think it is usually a "bad practice" to just capture `Exception`. Do we know what exception can occur here? What else other than `IOException` is thrown here?

jdrueckert (changes requested):
This worked nicely against that damned SIGSEGV :tada: 
As described, with a broken config it displays the defaults again.

**However**, when I then change some value, it doesn't clean up the broken config from before but simply overwrites parts of it, resulting in again a broken config, e.g.
`{"height":1.75}83,0,255],"playerName":"Player37343"}"height":1.7}`
Before breaking it and starting Terasology again, that's what it looked like:
`{"color":[194,183,0,255],"playerName":"Player37343","height":1.7}`

DarkWeird:
>I guess keeping the broken one is okay as long as settings are not changed.
If they are changed I'd say we should clean the broken config up before.

It isn't easy. fast way - easy. right way - not so easy.

For example: if you choose equals to default values - then you will have wrong case with changed values will stay, when you returns default values by hand - they not saves.

jdrueckert (approved):
Tests out very nicely :+1:

Issue #4152, opened 2020-09-19
Client may (re)spawn with camera stuck at 0,0,0 under certain circumstances in multiplayer


Cervator:
## What you were trying to do

Play test and assorted testing, dying to respawn, disconnecting to relog, etc. Wanted to be able to respawn successfully or relog at the place I logged out.

### What actually happened

In some cases my player's _camera_ would be stuck at 0,0,0 and able to look around. At the same time an entire player avatar was correctly present where it was supposed to be, and even accepted the input from said players mouse + keyboard, seeming to move, break blocks, etc perfectly normally to an observing player.

@DarkWeird noted some likely clues / theories in #4126 where it seems like a `LocationComponent` _somewhere_ gets set wrong or not set at all. It might also be that the few entities related to a player (client, character, gaze...) get split somehow, as I think we've even seen the avatar _both_ stuck in the ground at 0,0,0 as well as functioning normally at a proper location.

### How to reproduce

* Start a multiplayer game - has been seen in JoshariasSurvival and MetalRenegades for play tests (so several people involved from different sites)
* Connect, disconnect, die, respawn, relog, etc
* Profit!

Don't have a lot more specific details yet. We also tried during a play test where it happened to reset a given player's `config.cfg` to cause creation of a brand new player on the server, which might have fixed it in some cases, but not others? May be a roll of the dice or up to timing whether a player successfully avoids the issue or not.

jdrueckert:
Assumption (@pollend): `isFinite` checks return (0,0,0) in case player position is not finite. Question is why is the player position not finite?

PR #4662, opened 2021-05-12
chore: remove finite check from LocalPlayer

pollend:
we've mentioned in the past having a situation where the player position is placed at (0,0,0). I'm suspecting it has something to do with this logic. so for an invalid position we just return dest or a new Vector3/Quaternion. we shouldn't be just defaulting to 0,0,0. this ends up hiding a larger underlying problem where we see some odd behavior but we've written this logic in a way to hide a large problem. I rather have the game crash with a stacktrace then try poking at thing in the dark because we've written ourselves into a corner. 

https://github.com/MovingBlocks/Terasology/issues/4152


jdrueckert (commented):
I agree that we don't want to hide any bigger problem with that but maybe instead of removing null checks entirely, we could consider throwing a reasonable exception instead.

skaldarnar (approved):
I think not having explicit checks for the one-line methods `getPosition`, `getRotation`, and `getVelocity` is completely fine. The only thing we gain here is a user-readable message, but an NPE already pins it down to a single point of failure.

A similar argument holds for `getViewPosition` and `getViewRotation`. The line in which the NPE would occur indicates well enough what the issue is (at least to the devs).

We might add todo items to revisit this when we have better exception handling (soft and hard errors for crashing to menu or to desktop).

PR #4664, opened 2021-05-14
chore: replace TypeMap with ClassToInstanceMap

pollend:
This replaces TypeMap with just a standard map. TypeMap is a straight wrapping class of Map with only a subset of the possible functions. I think its generally more useful to just use a map and the details of the usage is already hidden away in one level of indirection so the extra level of indirection seems pretty unnecessary given what is lost with just wrapping a parameter like this.


DarkWeird:
I suggest to use `ClassToInstanceMap` from guava

DarkWeird (approved):

PR #4670, opened 2021-05-16
chore: replace map with switch cases for Side

pollend:
these changes are similar to this PR: https://github.com/MovingBlocks/Terasology/pull/4668

Not so sure about some of these changes it does introduce more lines of code and I do prefer moving some of this logic to the calling methods so its easier to follow. I ended up removing some of the boolean checks since the cost doesn't seem too expensive at the moment. I think it will only be relevant if we can see some slow down through profiling the code.


engine/src/main/java/org/terasology/engine/math/Side.java
```diff
@@ -201,95 +129,214 @@ public static Side inHorizontalDirection(double x, double z) {
...
+                }
+                break;
         }
+        return this;
     }
 
     public Side pitchClockwise(int turns) {
-        if (!canPitch) {
-            return this;
-        }
```
pollend, line 252: the optimization seems a bit premature so I dropped it from the class.

PR #4672, opened 2021-05-16
Fix for version file for launcher + increment to next major release

Cervator:
This quick-fixes https://github.com/Terasology/Index/issues/16 and should let the launcher identify newer Omega zips from Jenkins, as the `versionInfo.properties` is again attached from a slightly updated path then retrieved during the Omega zip making.

Goes with https://github.com/Terasology/Index/pull/17

Additionally I think we're overdue for bumping the engine version to the next major, considering the next release will include the _major_ upgrade to Gestalt v7, that definitely deserves a bump.

If no objections we should also adjust the current milestone accordingly: https://github.com/MovingBlocks/Terasology/milestones/45/edit

And I think #4509 should probably be closed as complete as part of that? I know there are still leftovers floating around, but that core issue seems to deserve its finished status 👍 

Omega fix confirmed with https://jenkins.terasology.io/teraorg/job/Nanoware/job/Omega/job/develop/34/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,7 +24,7 @@ node ("ts-engine && heavy && java8") {
     stage('Build') {
         // Jenkins sometimes doesn't run Gradle automatically in plain console mode, so make it explicit
         sh './gradlew --console=plain clean extractConfig extractNatives distForLauncher testDist'
-        archiveArtifacts 'gradlew, gradle/wrapper/*, templates/build.gradle, config/**, facades/PC/build/distributions/Terasology.zip, engine/build/resources/main/org/terasology/version/versionInfo.properties, natives/**, build-logic/src/**, build-logic/*.kts'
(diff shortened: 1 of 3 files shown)


Cervator:
[Omega builds have no versionInfo](https://trello.com/c/Wci782ks/114-omega-builds-have-no-versioninfo)

keturn:
Limit PR scope to a single issue, please! Unless the issues are intertwined, but that doesn't look to be the case here.

I'm picking on this one a bit because while it is often quite practical to include some one-line change along with the other work in a PR as some bit of incidental cleanup, this PR description talks more about the things that _aren't_ the targeted issue than the fix itself.

(despite me picking on it, I recognize there _is_ more description of the issue and its fix on both Index#16 and Index#17)

I do agree that this change to the Jenkinsfile in combination with the linked change to Index/Jenkinsfile should be sufficient to repair things.

Issue #4542, opened 2021-02-25
Square rendering artifacts


skaldarnar:
It is related to the width/height ratio...

![image](https://user-images.githubusercontent.com/1448874/109227615-e1c2d380-77c0-11eb-8088-b1ce076d743c.png)

![image](https://user-images.githubusercontent.com/1448874/109227602-dcfe1f80-77c0-11eb-94a6-78d5fce2fb59.png)

![image](https://user-images.githubusercontent.com/1448874/109272334-06df3280-7811-11eb-9a24-1a9f5d1103f1.png)

Could be related to https://github.com/MovingBlocks/Terasology/commit/1ca680fb80646cd3325cbe48347d084117f8a1d7. The vignette setting as at least an impact on how the artifact looks - with vignette enabled the squares are black, with it disabled the artifacts are roughly the color of the background (varying color).

skaldarnar:
:detective: Some more detective work: when saving a screenshot of the game via <kbd>F12</kbd> the artifacts are not visible. It is definitely related to the window size and/or ratio of width to height.

![image](https://user-images.githubusercontent.com/1448874/118304546-15bbb400-b4e7-11eb-9dcb-89997bfb9dce.png)
_Screen capture with artifacts_

![Terasology-210514190005-1636x982](https://user-images.githubusercontent.com/1448874/118304555-18b6a480-b4e7-11eb-8987-6ed0a6e8c977.png)
_Screenshot as saved by Terasology via <kbd>F12</kbd>_

PR #4674, opened 2021-05-16
fix: UI in game will get scaled with chaning "Resolution Scale" in world

pollend:
Changing viewport scaling while in a world will cause the game UI to also scale. glViewport is a bit un-manged so the state of the viewport can get dirty between draw calls.  the viewport is update in WorldRenderImpl.java and in a couple other places in the engine. for StateInGame worldRender is called followed by the UI so the viewport is changed somewhere within worldRender. This doesn't seem like necessarily the correct fix to the problem though. 

There is some strange non-determinism with this that I can't quite pin down. DAG pipeline seems to be pretty non-deterministic with how its initialized and the order of things changes. can't quite pin down the behavior. 

image

I also managed to run into this issue: https://github.com/MovingBlocks/Terasology/issues/4542 also fiddling with screen scale. some state is getting dirty and causing things to break down the line.


skaldarnar (approved):
I tried this in combination with #4676 and the visual artifacts seem to be gone. Also, I did not see any black squares as described in #4542 - maybe this solves that bug.

Notworthy: the game even started up with proper ui aspect ratio in windowed mode before resizing :sweat_smile:

Issue #4655, opened 2021-05-08
remove module scanning from telemetry.Metrics


keturn:
During #4640, we found engine.telemetry.Metrics has a `fetchMetricsClassFromEnvironment` method that scans over all the modules in ModuleManager's registry to find `@TelemetryCategory` annotations in modules.

In investigating how we were going to test this, we found there _are_ no modules that use the TelemetryCategory annotation, and we had hard time thinking up any excuses to use it in our current modules. [see [#logistics](https://discord.com/channels/270264625419911192/633681387995660303/840690824991211520)]

I propose removing this "modules may define new TelemetryCategories" feature, leaving telemetry with the metrics that are defined by engine.

It looks like that function will still need to do _something_ to get those annotated classes in the engine, but we can do away with the "scan the entire module registry" part.

PR #4675, opened 2021-05-16
chore(telemetry): drop module scanning from telemetry.Metrics

jdrueckert:
Closes https://github.com/MovingBlocks/Terasology/issues/4655

### DIFF
--- a/engine/src/main/java/org/terasology/engine/telemetry/Metrics.java
+++ b/engine/src/main/java/org/terasology/engine/telemetry/Metrics.java
@@ -31,6 +31,7 @@ import org.terasology.engine.telemetry.metrics.Metric;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/engine/telemetry/Metrics.java
```diff
@@ -50,28 +51,19 @@
...
-            if (module.getResources() instanceof DirectoryFileSource) {
-                ResolutionResult result = resolver.resolve(moduleId);
-                if (result.isSuccess()) {
-                    try (ModuleEnvironment environment = moduleManager.loadEnvironment(result.getModules(), false)) {
-                        for (Class<?> holdingType : environment.getTypesAnnotatedWith(TelemetryCategory.class, new FromModule(environment, moduleId))) {
-                            metricsClassSet.add(holdingType);
-                        }
-                    }
-                }
+        try (ModuleEnvironment environment = moduleManager.loadEnvironment(Collections.emptySet(), false)) {
```
keturn, line 61: `loadEnvironment(Collections.emptySet())` is not something I expected to see, but hey, in this current context of "fix the regressions," I'm on board. It gets rid of the "scan the whole registry" thing we wanted to get rid of, but other than that it doesn't change from the previous approach more than necessary.

We already know that this part of the ModuleManager interface is something we want to revisit.

engine/src/main/java/org/terasology/engine/telemetry/Metrics.java
```diff
@@ -50,28 +51,19 @@
...
     public void initialise(Context context) {
 
-        Set<Class> metricsClassSet = fetchMetricsClassFromEnvironemnt(context);
+        Set<Class> metricsClassSet = fetchMetricsClassFromEngineOnlyEnvironment(context);
         initializeMetrics(metricsClassSet, context);
     }
 
-    private Set<Class> fetchMetricsClassFromEnvironemnt(Context context) {
-
+    private Set<Class> fetchMetricsClassFromEngineOnlyEnvironment(Context context) {
```
keturn, line 58: ⭐ appropriate method rename

keturn (approved):
I remain concerned about not knowing how to test this, but it looks good to me and I think the area is low-risk. 👍

PR #4681, opened 2021-05-16
chore: migrate initialPost to glsl 330

pollend:
this is a migration of the initial post processing logic to glsl330. I also removed the chromatic abberation logic, its not configured or used but we can add it back if we need to use it?

this is tested by enabling light shafts and bloom from settings so all paths in the shader are evaluated. 

requires PR:

https://github.com/Terasology/CoreRendering/pull/44


jdrueckert:
So to see everything in effect, I first turned off light shafts and bloom (because I had both enabled already), took a screenshot and then turned both on again. However, I don't seem to get light shafts anymore now... not sure if this is more of a settings issue, but yeah, dunno... :woman_shrugging: 
what do you think @pollend

pollend:
sounds strange, does the correcty property hit the node. there is like an onChange thing

jdrueckert (approved):

jdrueckert:

PR #4687, opened 2021-05-17
chore: migrate downsampler to glsl 330

pollend:
migrate down sampler shader to gls330

downsampler is used for the exposure affect but I'm not sure this even works correctly. it should still work the same way in its current state.  can be enabled from setting exposure. 

PR: https://github.com/Terasology/CoreRendering/pull/49


jdrueckert (approved):

jdrueckert:

PR #4692, opened 2021-05-18
chore: remove unused displaylist in SpriteParticleRenderer

pollend:
From what I can tell this display list is not called within SpriteParticleRenderer so It should be safe to drop this from the class. The only instance I see using this renderer is in Health  when breaking a block.


DarkWeird (approved):
Yeah, I noticed this too.

PR #4694, opened 2021-05-19
fix: don't restrict world generators to a certain type of module source

### LINKED ISSUE #4673: ensure world generators can be loaded from jar modules
From #4640:

> Occurrence in [WorldGeneratorManager.java](https://github.com/MovingBlocks/Terasology/blob/78e608c57468210a001a8b84df0fd51f76c5d71d/engine/src/main/java/org/terasology/engine/world/generator/internal/WorldGeneratorManager.java#L47):
> - `WorldGenManager` is refreshed for instance on saving the world gen config
> - on refreshing, for every module in the registrythat is present as source code we try to resolve it, load the environment and get all world generators using the `RegisterWorldGenerator.class` annotation to add valid world generators to a list with generator information
> 
> Like with the metrics, the intent here might've been to skip asset-only modules as they won't have annotations to scan.

> <hr />

> This sounds like it might actually be a big part of our MTE test failures. Especially in environments like CI module builds that don't have the full module tree checked out, MTE and other dependencies will be jars, not a DirectoryFileSource.

I guess the way to test this from the application would be something along the lines of:

- `groovyw module get CoreSampleGameplay` _without_ also having `CoreWorlds` or other world provider sources present and readable in `modules/`
- `gradlew game`
- assert that gradle _did_ put jar files for the required world modules in `cachedModules/`
- check to see if you can start a game
- make sure you can enable world generators in Advanced Game Setup and make worlds with them on the world creation screen

Once that is all working, other interesting things to explore include:

- run `gradlew integrationTest` on a module with MTE tests (like `Health`) _without_ `modules/ModuleTestingEnvironment` present as source
- run that module's tests in CI
  - to try this before merging your MovingBlocks/Terasology branch with these changes, you'd probably need to stage it at Nanoware/Terasology and run a test job under https://jenkins.terasology.io/teraorg/job/Nanoware/job/TerasologyModules/

keturn:
Fixes #4673.

This seems to have been not as wide in scope as I'd feared. The changes are specific to the UI in Advanced Game Setup.

Instead of changing the conditional to something other than `instanceof DirectoryFileSource`, I removed the conditional entirely. Not sure if the thing that led to the conditional is no longer relevant in gestalt-module v7. Letting it iterate over all modules instead of only some does not seem to have caused immediate problems.

### How to test

Use a workspace that does have gameplay modes (e.g. CoreSampleGameplay) but does **not** have the world generators as source in `modules/`. (If you've checked them out, either remove them or temporary move them someplace else to hide them. gradle will fetch jars for dependent modules and put them in `cachedModules/`)

Start a new game with Advanced Game Setup.

Assert that all the expected types of worlds are listed in the world creation screen. If the only entry in the dropdown is `dummy`, something is wrong.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/world/generator/internal/WorldGeneratorManager.java
+++ b/engine/src/main/java/org/terasology/engine/world/generator/internal/WorldGeneratorManager.java
@@ -9,20 +9,21 @@ import org.slf4j.LoggerFactory;
 import org.terasology.engine.context.Context;
 import org.terasology.engine.core.SimpleUri;
 import org.terasology.engine.core.module.ModuleManager;
-import org.terasology.gestalt.module.Module;
-import org.terasology.gestalt.module.ModuleEnvironment;
-import org.terasology.gestalt.module.dependencyresolution.DependencyResolver;
-import org.terasology.gestalt.module.dependencyresolution.ResolutionResult;
-import org.terasology.gestalt.module.resources.DirectoryFileSource;
-import org.terasology.gestalt.naming.Name;
 import org.terasology.engine.registry.InjectionHelper;
 import org.terasology.engine.world.generator.RegisterWorldGenerator;
 import org.terasology.engine.world.generator.UnresolvedWorldGeneratorException;
 import org.terasology.engine.world.generator.WorldGenerator;
+import org.terasology.gestalt.module.Module;
+import org.terasology.gestalt.module.ModuleEnvironment;
+import org.terasology.gestalt.module.dependencyresolution.DependencyResolver;
+import org.terasology.gestalt.module.dependencyresolution.ResolutionResult;
+import org.terasology.gestalt.naming.Name;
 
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/terasology/engine/world/generator/internal/WorldGeneratorManager.java
```diff
@@ -39,34 +40,32 @@ public WorldGeneratorManager(Context context) {
     }
 
     public void refresh() {
-        ModuleManager moduleManager = context.get(ModuleManager.class);
```
keturn, line 42: Reviewers note: this diff is much easier to read on GitHub if you ⚙️ ☑️  _Hide whitespace changes_

keturn:
Some failures under module unit tests (Pathfinding, MTE), but I think those are failing in develop too.

I don't expect this PR to change module test results if the WorldGeneratorManager is only used in the setup UI.

jdrueckert (approved):

Issue #4640, opened 2021-05-01
audit use of `instanceof DirectoryFileSource`


keturn:
There were a number of places #4622 did:

```diff
-            if (module.isCodeModule()) {
+            if (module.getResources() instanceof DirectoryFileSource) {
```

because `isCodeModule` is not a method in gestalt-v7.

These call sites need to be audited to see if that is a suitable replacement.

* What is it they need to know about the module?
* Does the replacement work in all situations? if module is checked out as source, if it's a jar in cachedModules, if it's in a release bundle, etc.

jdrueckert:
Occurence in [ModuleDownloadListGenerator.java](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/core/module/ModuleDownloadListGenerator.java#L80):
- it tries to find out whether an online version is newer than a local version of a module
- if versions are equal it tries to check if the local module is a source module or a jar
- if it is a source module, the module is not downloaded again

So:
> What is it they need to know about the module?

Whether it's source or jar.

> Does the replacement work in all situations?

No idea, would need to try that out I guess.

keturn:
Ohh, because if you're in a dev workspace and you have sources, you don't want to download a thing that's going to be different from your local sources.

Okay, that makes sense. I am not 100% sure it _works,_ because [mumble mumble], but I understand the use case there.

jdrueckert:
Occurence in [Metrics.java](https://github.com/MovingBlocks/Terasology/blob/78e608c57468210a001a8b84df0fd51f76c5d71d/engine/src/main/java/org/terasology/engine/telemetry/Metrics.java#L64):

- at init time, it tries to fetch the classes implementing metrics from the env
- for every module registered in the module manager it gets the latest version of that module and checks whether it's present as source
- only if it is, it will resolve its dependencies and let the module manager load the environment for the dependency modules it resolved and from there fetch all classes with the `TelemetryCategory` annotation

So: 
> What is it they need to know about the module?

Again, that it's present as source, I guess. My assumption is, that `DependencyResolver` cannot properly resolve dependencies of a jar or `ModuleManager` cannot properly load the environment, although in that case I'd expect it to check the dependencies being present as source :thinking: 
But maybe, I'm missunderstanding, what the following does?
```java
DependencyResolver resolver = new DependencyResolver(moduleManager.getRegistry());
Module module = moduleManager.getRegistry().getLatestModuleVersion(moduleId);
[...]

PR #4695, opened 2021-05-19
chore(telemetry): drop module scanning from telemetry.TelemetryScreen

jdrueckert:
Relates to #4655 and #4640


keturn (approved):
Notes in the #4675 review apply here as well. Approving along the same lines.

PR #4696, opened 2021-05-19
fix(engine): explicitly register `TypeHandler.class` as representative for TypeHandlerLibrary

skaldarnar:
This makes the `@API` annotated classes and packages of the (non-)subsystem TypeHandlerLibrary visible to the APIScanner operating on the aggregated engine module.

Previously, TypeHandlerLibrary fell through the cracks of not being a real subsystem, but also not being part of the main engine source tree.

Fixes https://github.com/Terasology/Health/issues/85


keturn (approved):
This seems like it should be fine. Still wish I could say that with more confidence.

I'm trying to figure out how or where to test it, and I'm at a loss. This one seems to not provoke a failure from the engine itself, only certain modules.

and tests of the form “make sure this thing I just hardcoded in to this list is actually in this list” are not particularly informative tests.

I guess it's the sort of thing that should be an integration test when we get a module in engine-test.

PR #4697, opened 2021-05-19
chore: add doc and todo on module presence as source check

jdrueckert:
Relates to #4640


pollend (approved):
umm I think it make sense :?

skaldarnar (approved):

Issue #4607, opened 2021-03-27
ci: separate stages for unit tests and flaky tests


keturn:
Following #4606, we'll have more than one type of `test` task. We'd like CI to run all the tests. When checking a PR it should require all the well-behaved tests to pass. A failure among the less-reliable tests may mark the build as "unhealthy" but not failed.

If we want to do this for the engine, the place to do it is `Jenkinsfile`.

The Jenkinsfile for modules lives in https://github.com/MovingBlocks/ModuleJteConfig/blob/master/Jenkinsfile

### Open Questions

- [ ] Do we want to do this for engine tests? It does have some `TteTest` cases, but are they flaky enough to demote to lesser importance?

### Things to Watch Out For

- Make sure _all_ the test results are recorded in Jenkins. i.e. splitting them up in to separate tasks doesn't accidentally make it so only the results of one or the other of those tasks are saved.
- People have built tools to do elaborate monitoring and treatment of flaky tests. _**Resist** going down that rabbit hole!_ 🕳  
  We don't want to make it _too_ easy to accept test flakiness too quickly. Save your time for other things, like fixing flaky tests!

jdrueckert:
![](https://github.trello.services/images/mini-trello-icon.png) [Separate Jenkins test stages](https://trello.com/c/fgorgiQY/53-separate-jenkins-test-stages)

keturn:
you can see what all `check` will run by doing `gradlew check --dry-run`

It looks like `gradlew check --exclude-task test` might work if we want to have a "all the checks _except_ tests" build step without explicitly naming each of those checks.

jdrueckert:
Also, to have the reference here (I need to make answering stuff preferably on GitHub a habit):
since https://github.com/MovingBlocks/Terasology/pull/4611 got merged, we now do have three different gradle tasks:
- `unitTest` running only unit tests
- `integrationTest` running only integration tests (MTE, TTE, etc.)
- `test` running both, unit and integration tests - so basically all tests

skaldarnar:
See also changes done previously in https://github.com/MovingBlocks/ModuleJteConfig/commit/e64657b9d83284a6c272dcce265821a4d5024e16

PR #4701, opened 2021-05-22
ci: split Analytics stage

jdrueckert:
Splits the one big 'Analytics' stage of tests, static code analysis, and
javadoc generation into smaller stages with dedicated purpose.

- **Unit Tests** run unit tests and upload report. this is not expected
to fail, thus using default health scale factor of 1.0
- **Integration Tests** run the heavy-weight MTE tests (and alike) and
upload report. we know they are flaky, so keep them after the publish
stage, and keep the health scale factor of 0.0
- **Analytics** run static code checks (spotbugs, PMD, ...) and upload
reports
- **Documentation** generate javadoc and upload result. this could
potentially be the home for a static site generation step in the future

Based on https://github.com/MovingBlocks/ModuleJteConfig/pull/10
Contributes to #4607


jdrueckert:
[Separate Jenkins test stages](https://trello.com/c/fgorgiQY/53-separate-jenkins-test-stages)

Jenkinsfile
```diff
@@ -21,11 +21,18 @@ node ("ts-engine && heavy && java8") {
...
+
     stage('Build') {
         // Jenkins sometimes doesn't run Gradle automatically in plain console mode, so make it explicit
         sh './gradlew --console=plain clean extractConfig extractNatives distForLauncher testDist'
         archiveArtifacts 'gradlew, gradle/wrapper/*, templates/build.gradle, config/**, facades/PC/build/distributions/Terasology.zip, engine/build/resources/main/org/terasology/engine/version/versionInfo.properties, natives/**, build-logic/src/**, build-logic/*.kts'
     }
+
+    stage('Unit Tests') {
+        sh './gradlew unitTest'
+        junit testResults: '**/build/test-results/unitTest/*.xml', allowEmptyResults: true
```
keturn, line 33: You put this stage before Publish because you want to require the tests to pass before publishing, right?

In that case, I think we want to remove `allowEmptyResults` from this one.

keturn (commented):
I like the way the stages are organized. Did you work out the thing about recording results after test failures?

Jenkinsfile
```diff
@@ -46,16 +59,29 @@ node ("ts-engine && heavy && java8") {
             build job: 'Nanoware/Omega/master', wait: false
         }
     }
-    stage('Analytics') {
-        sh "./gradlew --console=plain check spotbugsmain javadoc"
+
+    stage('IntegrationTests') {
```
jdrueckert, line 63:
```suggestion
    stage('Integration Tests') {
```

Jenkinsfile
```diff
@@ -21,11 +21,24 @@ node ("ts-engine && heavy && java8") {
...
+
     stage('Build') {
         // Jenkins sometimes doesn't run Gradle automatically in plain console mode, so make it explicit
         sh './gradlew --console=plain clean extractConfig extractNatives distForLauncher testDist'
         archiveArtifacts 'gradlew, gradle/wrapper/*, templates/build.gradle, config/**, facades/PC/build/distributions/Terasology.zip, engine/build/resources/main/org/terasology/engine/version/versionInfo.properties, natives/**, build-logic/src/**, build-logic/*.kts'
     }
+
+    stage('Unit Tests') {
+        try {
+            sh './gradlew unitTest'
```
jdrueckert, line 33:
```suggestion
            sh './gradlew --console=plain unitTest'
```

Jenkinsfile
```diff
@@ -46,16 +59,29 @@ node ("ts-engine && heavy && java8") {
             build job: 'Nanoware/Omega/master', wait: false
         }
     }
-    stage('Analytics') {
-        sh "./gradlew --console=plain check spotbugsmain javadoc"
+
+    stage('Integration Tests') {
+        try {
+            sh './gradlew integrationTest'
```
jdrueckert, line 65:
```suggestion
            sh './gradlew --console=plain integrationTest'
```

Jenkinsfile
```diff
@@ -46,16 +59,29 @@ node ("ts-engine && heavy && java8") {
...
+    stage('Analytics') {
+        sh './gradlew --console=plain check -x test spotbugsmain'
         recordIssues tool: checkStyle(pattern: '**/build/reports/checkstyle/*.xml')
         recordIssues tool: spotBugs(pattern: '**/build/reports/spotbugs/main/*.xml', useRankAsPriority: true)
         recordIssues tool: pmdParser(pattern: '**/build/reports/pmd/*.xml')
         recordIssues tool: taskScanner(includePattern: '**/*.java,**/*.groovy,**/*.gradle', lowTags: 'WIBNIF', normalTags: 'TODO', highTags: 'ASAP')
     }
+
+    stage('Documentation') {
+        sh './gradlew javadoc'
```
jdrueckert, line 83:
```suggestion
        sh './gradlew --console=plain javadoc'
```

Jenkinsfile
```diff
@@ -46,16 +59,29 @@ node ("ts-engine && heavy && java8") {
...
-        sh "./gradlew --console=plain check spotbugsmain javadoc"
+
+    stage('Integration Tests') {
+        try {
+            sh './gradlew integrationTest'
+        } catch (e) {
+            echo 'Integration Tests failed'
+            throw e
+        } finally {
+            junit testResults: '**build/test-restults/integrationTest/*.xml', allowEmptyResults: true
```
keturn, line 70: In Python, we can do a try/finally without defining a catch at all. Since we're just re-throwing the error, can we do the same in groovy?

jdrueckert: let's try that out :slightly_smiling_face:

keturn:
I think using `try/finally` makes sense as way to do "run the record step whether or not this step passed" if it's being scripted in Groovy.

There was a typo preventing the integration tests from being recorded. Doing a test run now, and if that works out I will merge. 👍

keturn:
Hmm, nope, still says "None of the test reports contained any result." 🙁

keturn (approved):
Confirmed this is now recording results for both unit tests and integration tests. 👍

PR #4702, opened 2021-05-22
style: improve code style

skaldarnar:
Maintenance PR touching nearly **everything** in the engine, trying to bring the CheckStyle warnings down. For this PR the changes are (mostly) limited to _style_ changes, i.e., only formatting the code.

Before: 158 Error / 1,780 Normal
After: 117 Error / 422 Normal

image

## Open Questions

- we probably want to disable the "[HiddenField](http://jenkins.terasology.io/teraorg/job/Terasology/job/engine/job/PR-4702/12/checkstyle/Error/type.1289032088/)" check
- some "[PackageDeclaration](http://jenkins.terasology.io/teraorg/job/Terasology/job/engine/job/PR-4702/12/checkstyle/Error/type.1197137908/)" checks are failing which looks like some files were not moved into the `engine...` folder (mainly in tests)
- CheckStyle discourages "[ExplicitInitialization](http://jenkins.terasology.io/teraorg/job/Terasology/job/engine/job/PR-4702/12/checkstyle/Normal/type.-1468992172/#issuesContent)", e.g., initializing an `int` variable to `0` or an object to `null` - do we agree here?

Co-authored-by: jdrueckert <jd.rueckert@googlemail.com>


pollend (approved):

jdrueckert (approved):

pollend (dismissed):

keturn:
> we probably want to disable the "HiddenField" check

Oh, I see, it's being a bother with method arguments. Seems useful for local variables, though. Would it be sufficient to tell it to ignore parameters for constructors and setter methods? https://checkstyle.org/config_coding.html#HiddenField

> some "PackageDeclaration" checks are failing which looks like some files were not moved into the engine... folder (mainly in tests)

Yeah, tests got overlooked in that big move. I've been moving them over if/when I touch a test, but we could do a batch move to get it over with too.

> CheckStyle discourages "ExplicitInitialization", e.g., initializing an int variable to 0 or an object to null - do we agree here?

Its docs say:
> Rationale: Each instance variable gets initialized twice, to the same value. Java initializes each instance variable to its default value (0 or null) before performing any initialization specified in the code. So there is a minor inefficiency.

I find that unconvincing. I support disabling that one.

keturn:
> style: make constant names all uppercase

This one has been bugging me. Checkstyle defines a _constant_ as "a static and final field or an interface/annotation field, except serialVersionUID and serialPersistentFields."

When I think of constants like this, I think of value types. Numbers, strings, maybe slightly higher-level immutable values like URL.

But
* class references? https://github.com/MovingBlocks/Terasology/pull/4702/commits/ee9b0a414fcdb85ca81badf38b1eb38407195bba#diff-842a86dbf488376395fc2b4925b50faa11e4cb4d349a683be86feafd6ece5314R23
* mutable objects? https://github.com/MovingBlocks/Terasology/pull/4702/commits/ee9b0a414fcdb85ca81badf38b1eb38407195bba#diff-d1d5d704610ac766c8fbf6e7ffde383a8b285aaefde80768f5f61d6f6e0e8f8bR18

while those _can_ be overridden with an extra annotation when they're used, I feel those field-definition places are especially unsightly and distracting to have such extra noise in.

skaldarnar:
@keturn all good points 👍 

Re **HiddenField**: I think we stumbled upon "setters" in the context of the launcher before. We had a builder pattern in place, but were not prefixing methods with `set...`, so this rule kicked in. 😕 Maybe we want to enforce to have that prefix everywhere, or suppress the warning, or just rename the argument (I mostly dislike all of those options...). I'll do another audit of the typical cases we currently have that violate that rule.

Re **ConstantName**: I definitely see your point. Should we suppress single instances of that warning, or would you outright disable that check?

jdrueckert (approved):

jdrueckert:

PR #4703, opened 2021-05-22
test(ModuleManager): new tests for existing behavior

keturn:
adds some test cases to ModuleManagerTest


keturn:
Added some docs.

FYI, this is _not_ an example of what I expect to see the on the average unit test. Ideally the units should be small enough that the interfaces and assertions mostly speak for themselves with only a little help.

In this case they needed _more_ than a little help as it's a tangle, and the effects can be far-reaching.

skaldarnar (approved):

PR #4705, opened 2021-05-22
feature: rework location change event

pollend:
I changed the location change event to be handled on the next frame. if you change the location position then I mark the last position and prevent the last position/rotation from getting changed. only on the next update do I unlatch the flag so the location is relative to the last frame instead of whenever the component is changed then saved. this is more consistent since you could do something like this and have the component for every change.  

```
Entityref ref;

LocationComponent. ee = ref.getComponent(LocationComponent.class);

ref.saveComponent(ee); // LocationChangeEvent
ref.saveComponent(ee);  // LocationChangeEvent
```


engine/src/main/java/org/terasology/engine/logic/location/LocationChangedEvent.java
```diff
@@ -2,67 +2,56 @@
...
-        this(newLocation, newLocation.position, oRotation, newLocation.position, newLocation.rotation);
+        this.lastPosition.set(lastPosition);
+        this.lastRotation.set(lastRotation);
+        this.currentPosition.set(currentPosition);
+        this.currentRotation.set(currentRotation);
     }
 
-    public LocationChangedEvent(LocationComponent newLocation, Quaternionfc oRotation, Quaternionfc nRotation) {
-        this(newLocation, newLocation.position, oRotation, newLocation.position, nRotation);
+    public Vector3fc lastPosition() {
```
skaldarnar, line 33: I think `last` and `current` works well here as naming. In `BeforeAfterEvent` we use `old` and `new` - was just wondering whether we should be consistent here :thinking:

engine/src/main/java/org/terasology/engine/logic/location/LocationChangedEvent.java
```diff
@@ -2,67 +2,56 @@
...
-    }
 
     public Vector3f vectorMoved(Vector3f dest) {
-        return oldPosition != null ? newPosition.sub(oldPosition, dest) : dest.set(0, 0, 0);
+        return currentPosition.sub(lastPosition, dest);
     }
 
     public float distanceMoved() {
-        return oldPosition != null ? newPosition.distance(oldPosition) : 0.0F;
+        return currentPosition.distance(lastPosition);
```
skaldarnar, line 55: Same here.

engine/src/main/java/org/terasology/engine/logic/location/LocationChangedSystem.java
```diff
@@ -2,23 +2,48 @@
...
+        if (lc.isDirty && (!lc.position.equals(lc.lastPosition) || !lc.rotation.equals(lc.lastRotation))) {
+            process.add(entity);
+        } else {
+            lc.isDirty = false;
+        }
+    }
+
+    @Override
+    public void update(float delta) {
+        for (EntityRef entity : process) {
```
skaldarnar, line 40: Would be interesting to collect some metrics here on how many entities we typically process per tick. Could easily be added in a follow up PR, I guess.

skaldarnar (changes requested):

engine/src/main/java/org/terasology/engine/logic/location/LocationChangedEvent.java
```diff
@@ -2,67 +2,56 @@
...
-        oldPosition = new Vector3f(oPosition);
-        oldRotation = new Quaternionf(oRotation);
-        newPosition = new Vector3f(nPosition);
-        newRotation = new Quaternionf(nRotation);
-        component = nComponent;
-    }
 
     public Vector3f vectorMoved(Vector3f dest) {
-        return oldPosition != null ? newPosition.sub(oldPosition, dest) : dest.set(0, 0, 0);
+        return currentPosition.sub(lastPosition, dest);
```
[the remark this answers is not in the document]
pollend: we don't actually use it in any capacity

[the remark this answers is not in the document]
skaldarnar: We can keep it, that's not the problem. 

In particular, I think that 

```java
locationChange.vectorMoved(dest);
// or 
locationChange.positionDelta(dest);
```

is way better to read than doing the computation on receiver side

```java
locationChange.currentPosition().sub(locationChange.lastPosition(), dest);
```

skaldarnar (approved):

PR #4711, opened 2021-05-24
bugfix: set core profile to 3.3

Hume2:
REF: https://github.com/MovingBlocks/Terasology/issues/4698

### Contains

Fixes #4698 

### How to test

Run it on my computer and if it shows anything except a black screen after the loading screen, it works ;)

PR #4715, opened 2021-05-25
rendering: fix canvas fbo render with GL330

pollend:
canvas lets you render to another off screen buffer and feed that back to the UI. the only module that takes advantage of this is minimaps. should be enough to verify that the maps renders correct in a default gameplay template. drops more immediate mode functions that should be relevant to this PR: https://github.com/MovingBlocks/Terasology/pull/4711

PR: https://github.com/MovingBlocks/Terasology/pull/4620

image


skaldarnar (approved):

jdrueckert:

PR #4720, opened 2021-05-28
refactor(rendering): replace defaultTextured with outputPass shader

pollend:
I'm replacing the defaultTextured material with just a straight outputPass shader.  this just takes in a target texture and passes it back out. defaultTextured is suppose to be generic to some degree but this doesn't really fit into this scheme well. I'm not doing any sort of vertex transformation since all the work for a post processing effect is done on the fragment shader. we might want to consider consolidating the vertex portion of the shader and just have something like 

```
sobel.pass
downSampler.pass
...
```

required PR: https://github.com/Terasology/CoreRendering/pull/52


engine/src/main/resources/org/terasology/engine/assets/shaders/outputPass_frag.glsl
```diff
@@ -0,0 +1,19 @@
+#version 330 core
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+in vec2 v_uv0;
+
+uniform sampler2D target;
```
pollend, line 7: we can't use texture since its kind of reserved.

pollend:

jdrueckert (approved):

PR #4721, opened 2021-05-28
refactor(rendering): migrate ssaoBlur for GLSL 330

pollend:
migrate ssaoBlur to GLSL 330.

this can be verified by enable SSAO from settings. 

required PR: https://github.com/Terasology/CoreRendering/pull/53


jdrueckert:
The underwater areas look weird with SSAO enabled :thinking: 
And personally tbh I like the shadows without SSAO better :sweat_smile: 

Left: without SSAO, Right: with SSAO
image

pollend:
> The underwater areas look weird with SSAO enabled thinking
> And personally tbh I like the shadows without SSAO better sweat_smile they seem to give everything more depth
> 
> Left: without SSAO, Right: with SSAO
> image

its a screen space effect so its done as a post processing effect. the opaque materials are rendered first so we could try re-order things but I would need to look at it closer :?.

jdrueckert:

jdrueckert:
> its a screen space effect so its done as a post processing effect. the opaque materials are rendered first so we could try re-order things but I would need to look at it closer :?.

@pollend Can you create an issue for that so we don't forget about it?

jdrueckert (approved):

PR #4722, opened 2021-05-28
chore: rename engine-module.txt to module.txt

### LINKED ISSUE #4646: Don't use `engine-module.txt` as a possible metadata source for all modules
https://github.com/MovingBlocks/Terasology/blob/0d39a3bca5b54d4db49b57d2f8be49a57cd484a9/engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java#L81-L84

I _imagine_ that engine has this unique `engine-module.txt` filename because it was loaded from a `class.getResource` call https://github.com/MovingBlocks/Terasology/blob/0fa1272241b43be94e9920c84a41683643c9f240/engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java#L128  
instead of a filesystem API that opens a file in the root of a specific directory or archive.

So maybe it needed a name to distinguish it from a `/module.txt` file that might have been put there by anything else on the classpath?

but now that we've moved it to `org/terasology/engine/` maybe that's enough of a qualifier that it could just be named `module.txt` again?

[☝ copied from a comment thread on #4622]

This isn't quite release-critical for a first gestalt-v7 release, but it would definitely be good to get done while we're still on the topic before moving down the roadmap to something else.

jdrueckert:
Resolves #4646 

Tbh I have no idea, if this is what's expected to be done to resolve #4646 ...
I just thought I'd go for what I understood and use that as a basis for discussion :woman_shrugging: :sweat_smile:


engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -79,8 +79,7 @@ public ModuleManager(String masterServerAddress, List<Class<?>> classesOnClasspa
 
         Map<String, ModuleMetadataLoader> mmlm = moduleFactory.getModuleMetadataLoaderMap();
         mmlm.put(TerasologyConstants.MODULE_INFO_FILENAME.toString(), metadataReader);
-        mmlm.put("engine-module.txt", metadataReader);  // FIXME: this should be *only* for engine-module.
-        // FIXME: …or maybe engine-module doesn't need a special filename anymore?
+        mmlm.put("module.txt", metadataReader);  // FIXME: this should be *only* for engine-module.
```
keturn, line 82: I think `module.txt` is `MODULE_INFO_FILENAME`, so we can remove this second `mmlm.put`.

keturn (changes requested):
It looks like the one thing you left out of this PR titled "rename engine-module.txt to module.txt" was renaming `engine-module.txt`. 😜

jdrueckert:
> It looks like the one thing you left out of this PR titled "rename engine-module.txt to module.txt" was renaming `engine-module.txt`. stuck_out_tongue_winking_eye

Whoops :sweat_smile:

keturn (approved):

keturn:
Ah, _now_ it runs with passing tests!

I see versionInfo.properties is built with version information, and I believe the test cases will make sure that ModuleManager has a module of the appropriate name.

I think that the only thing left to confirm is that you can actually run a game world in a real process with other (non-engine) modules present. Can't think of anything else in particular to look for beyond that.

pollend:
@jdrueckert do you want to merge this?

jdrueckert:
> @jdrueckert do you want to merge this?

need to test it properly (as keturn mentioned) first, but once that's done: yes. 
any objections? @pollend

jdrueckert:
tests out fine in CoreGameplay, so I'll go ahead and merge this

PR #4725, opened 2021-05-28
fix: ensure absolute path in PathManager

skaldarnar:
This fixes the discovery of duplicate modules, such as
```
22:16:12.763 [main] ERROR o.t.g.module.TableModuleRegistry - Duplicate module CoreAdvancedAssets-1.3.0-SNAPSHOT discovered
22:16:12.763 [main] INFO  o.t.gestalt.module.ModulePathScanner - Discovered duplicate module: CoreAdvancedAssets-1.3.0-SNAPSHOT, skipping
```

This also makes sure to compare the paths with `.equals` instead of checking for reference equality with `==`.

Plus, small style change to make stream operations more readable in `ModuleManager`.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/core/PathManager.java
+++ b/engine/src/main/java/org/terasology/engine/core/PathManager.java
@@ -160,7 +160,7 @@ public final class PathManager {
      * @throws IOException Thrown when required directories cannot be accessed.
      */
     public void useOverrideHomePath(Path rootPath) throws IOException {
-        this.homePath = rootPath;
+        this.homePath = rootPath.toAbsolutePath();
         updateDirs();
     }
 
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/engine/core/PathManager.java
```diff
@@ -355,7 +355,7 @@ private void updateDirs() throws IOException {
         Path homeModPath = homePath.resolve(MODULE_DIR);
         Path modCachePath = homePath.resolve(MODULE_CACHE_DIR);
 
-        if (homePath == installPath) {
+        if (homePath.equals(installPath)) {
```
keturn, line 358: This is a cross-language confusion thing that I probably keep tripping over.

Note to self:

Java `==` compares only instance identity. More like C. Python `is` or JavaScript `===` (I think).

Java does _not_ support operator overloading at all. Defining `.equals` is not like defining Python `__eq__`.

keturn (approved):
LGTM. I think it makes sense for a parameter named `rootPath` to be used as an absolute path.

I wouldn't mind unit tests for this, but I'm not too worried about it.

PR #4726, opened 2021-05-29
fix: adjust 'Land.ogg' sound asset to be mono channel

skaldarnar:
TIL our engine can adjust the volume for audio tracks and sounds, but only if they are using mono channel. This was quite apparent for the landing sound when running the game in multiplayer. Foot steps and digging sounds would fade out with distance, but the landing sound does not.

I now pinned this down to the sound asset being stereo opposed to all other sound assets which are just mono...

Edit: this is a known issue, see #663

## How to test

- Start two instances of the game and host a local multiplayer game.
- Disable the sound effects on one client, and keep them enabled on the other.
- Move around with the client that has the sound effects disabled (so that you can hear that player move away from the other client).

The landing sound after jump/falling off an edge should fade out with distance as all other character sounds do.


jdrueckert (approved):
Tests out fine :+1:

PR #4727, opened 2021-05-29
perf: reuse Vector3f/Quaternionf between game ticks

skaldarnar:
I noticed that the `AudioSystem` creates new instances for player position, velocity, and view rotation with every game tick.

It can just reuse the same instance between game ticks, saving on unnecessary instance creation.


pollend (approved):
looks straightforward the calling method looks safe to do this.

PR #4728, opened 2021-05-29
doc: improve document of PlaySound and PlaySoundForOwner events

skaldarnar:
These are just small additions, but I hope they make the behavior of these events a bit clearer and document what I have learned about them.

Contributes to #663


pollend (approved):

engine/src/main/java/org/terasology/engine/audio/events/PlaySoundEvent.java
```diff
@@ -8,20 +8,29 @@
 import org.terasology.engine.network.BroadcastEvent;
 
 /**
- * This event is sent whenever a sound is to be played.
- * The event is not sent to or played for the client of the entity that sent the event as they have simulated it already.
+ * A <i>trigger event</i> to play a sound.
+ * <p>
+ * The event is not sent to or played for the client of the entity that sent the event as they have simulated it
+ * already.
```
pollend, line 14: I'm not sure what you are trying to say here?

```suggestion
 * The event is not sent to or played for the client of the entity that sent the event as they have simulated it
 * already.
```

engine/src/main/java/org/terasology/engine/audio/events/PlaySoundEvent.java
```diff
@@ -8,20 +8,29 @@
...
+ * <p>
+ * The target entities {@link org.terasology.engine.logic.location.LocationComponent LocationComponent} defines at which
+ * location (world coordinates) the sound is played at. If the entity does not have a location, the sound is played as
+ * global effect.
```
pollend, line 18:
```suggestion
 * The sound played from the target entity is played from the position defined by the {@link org.terasology.engine.logic.location.LocationComponent LocationComponent}.
 * If the entity does not specify a location, then the sound is played as a global sound effect.
```

maybe something like this?

pollend (changes requested):

engine/src/main/java/org/terasology/engine/audio/events/PlaySoundEvent.java
skaldarnar: yeah, me neither - I'll have another look at this...

pollend (approved):

PR #4729, opened 2021-05-29
rendering: migrate sky sphere

pollend:
this migrates on the the more annoying bits of this whole endeavor. this creates a sphere is that pretty much the same as the original display list type sphere.  probably should work out the details of how we want to deal with meshes going forward. The sky sphere should look the same with this PR. This object is a bit off to the side so I wonder what could be done to unify a clear API. 

requires PR: https://github.com/Terasology/CoreRendering/pull/54


pollend:

jdrueckert:
Am I supposed to see these lines and the central point where they meet?
I never noticed that before, so I'm wondering if that's expected.
image

pollend:
> Am I supposed to see these lines and the central point where they meet?
> I never noticed that before, so I'm wondering if that's expected.
> image

its in the other skybox as well before this change I believe. I think there is just a seam in the texture

pollend:
here is an image from develop.

image

engine/src/main/resources/org/terasology/engine/assets/shaders/sky_frag.glsl
```diff
@@ -1,22 +1,15 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+in vec4 v_position;
+in vec2 v_uv0;
+in vec3 v_colorYxy;
+in vec3 v_skyVec;
+
+//varying	vec3 colorYxy;
+//varying vec3 skyVec;
+//varying	vec4 position;
```
jdrueckert, line 12: can this be removed?

engine/src/main/resources/org/terasology/engine/assets/shaders/sky_vert.glsl
```diff
@@ -1,29 +1,26 @@
...
 uniform vec3    zenith;
 
 const vec4 eyePos = vec4(0.0, 0.0, 0.0, 1.0);
 
+uniform mat4 modelViewMatrix;
+uniform mat4 projectionMatrix;
+
 vec3 allWeather(float t, float cosTheta, float cosGamma)
 {
 	float	gamma      = acos ( cosGamma );
```
jdrueckert, line 26: tabs

engine/src/main/resources/org/terasology/engine/assets/shaders/sky_vert.glsl
```diff
@@ -65,12 +62,12 @@ void main(void)
...
-    position        = gl_Vertex;
-    gl_Position     = ftransform();
-    gl_TexCoord[0]  = gl_MultiTexCoord0;
-}
\ No newline at end of file
+	v_skyVec        = v.xyz;
+	v_colorYxy      = allWeatherSky(turbidity, max(v.y, 0.0) + 0.05, lv, l.y);
+	v_position      = vec4(in_vert, 1.0);
+	v_uv0           = in_uv0;
+	gl_Position     = (projectionMatrix * modelViewMatrix) * v_position;
```
jdrueckert, line 72: tabs

jdrueckert:
This currently fails for me with the following error:
```
20:45:06.738 [Chunk-Processing-Reactor] ERROR o.t.e.w.c.p.ChunkProcessingPipeline - Reactor thread was interrupted
java.lang.InterruptedException: null
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.reportInterruptAfterWait(AbstractQueuedSynchronizer.java:2056)
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2090)
	at java.base/java.util.concurrent.PriorityBlockingQueue.take(PriorityBlockingQueue.java:546)
	at java.base/java.util.concurrent.ExecutorCompletionService.take(ExecutorCompletionService.java:200)
	at org.terasology.engine.world.chunks.pipeline.ChunkProcessingPipeline.chunkTaskHandler(ChunkProcessingPipeline.java:104)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

jdrueckert (approved):

PR #4731, opened 2021-05-30
refactor(rendering): rework AABBRenderer with Mesh and update default glsl

pollend:
This replaced the display-list logic in AABBRendering with a StandardMesh. This is verified by verifying the block selection in a standard gameplay template. 

Note: this is implemented as a static mesh so constantly rebuilding the mesh this way might be considerably expensive. constantly rebuilding a VAO is probably a lot more expensive then using a displaylist but we use this in a small enough capacity with a sparse enough meshes where this might not be a problem. Ideally we could just push the data to some kind of buffer and submit all the meshdata per frame. umm will have to work that out when we have to look at this again. 

should help with merging this PR: https://github.com/Terasology/CoreRendering/pull/57

Required PR: https://github.com/Terasology/CoreRendering/pull/58

image


jdrueckert:
@pollend minor observation: I saw that sometimes we select a block although the center of the crosshair is not on it yet. Not sure if we check for the center or for a certain portion of the crosshair area to be on something. For something like a block it's not that big of a deal as that's big enough, but if we talk about torches or fences, this might be more of an annoyance.

Anyway, that's nothing that should block this PR, just some observation I wanted to note on.

jdrueckert (approved):

PR #4732, opened 2021-05-30
fix(rendering): correct order of initilization for GLSLShader

pollend:
I've run into this a few times when testing shaders. so these properties end up with empty strings and the shader end up in a misconfigured state so it fails when some property is called but missing from the shader. why does changing this order resolve the problem?


jdrueckert (approved):

jdrueckert:
As posted by @pollend on Discord: https://stackoverflow.com/questions/13554507/order-of-static-variable-initialization-java

PR #4740, opened 2021-06-04
fix: prevent NPE with DelayManager on remote clients

skaldarnar:
The delay manager is only available on authority systems and is not
initialized on remote clients.

Without this fix, it is impossible for clients to join a game as the
game will crash hard on connecting to the server when trying initialize
the player.

See #4742


subsystems/DiscordRPC/src/main/java/org/terasology/subsystem/discordrpc/DiscordRPCSystem.java
```diff
@@ -85,7 +85,9 @@ public void shutdown() {
     @ReceiveEvent
     public void onPlayerInitialized(LocalPlayerInitializedEvent event, EntityRef localPlayer) {
         /* Adds the periodic action when the player is hosting or playing online to update party size */
-        if (networkSystem.getMode() != NetworkMode.NONE) {
+        //FIXME: The 'delayManager' is only available on the authority system and is not initialized on clients.
+        //       This will fail with a NPE when a clients tries to join a game.
+        if (delayManager != null && networkSystem.getMode() != NetworkMode.NONE) {
```
keturn, line 90: This looks like it will avoid a potential NPE, but it does so by silently disabling this feature. I suggest logging an error here.

keturn (changes requested):
This isn't adding up for me. Git says the previous change to that `onPlayerInitialized` method was included in 4.3.0, have there really been no multiplayer games on 4.3.0 or later???

subsystems/DiscordRPC/src/main/java/org/terasology/subsystem/discordrpc/DiscordRPCSystem.java
skaldarnar: I'm wondering whether this was ever tested in a multiplayer scenario. The [DelayedActionSystem](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/logic/delay/DelayedActionSystem.java#L30-L32) as only implementation for the `DelayManager` interface is clearly marked as authority system, so this code can only ever work if the _client_ is also the _authority_ - which is the case for [single-player mode](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/network/NetworkMode.java#L11-L14) and when [hosting a multi-player](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/network/NetworkMode.java#L21-L24). 

I'm hesitant to log an error here, because it just cannot work, and it is no error that anyone should report to us 🤔 I think we'll need a ticket to track this, either to make it more clear that `DelayManager` can only be used on authority systems, or by providing a client-side implementation somehow... Or by making the delay manager controllable via events that get send against the authority...

jdrueckert (approved):

PR #4741, opened 2021-06-04
chore: change entity dump filename to '<timestamp>-entityDump.json'

skaldarnar:
Entity dumps are now save to files named like this:
```
2021-06-04T20:49:20.923507Z-entityDump.json
```

Previously, every entity dump was created as 'entityDump.txt'. When
working with multiple clients, or when dumping the entity state multiple
times, this required a manual rename each time.

Additionally, using the `.json` file ending helps with modern editors to
use correct syntax highlighting.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/logic/console/commands/CoreCommands.java
+++ b/engine/src/main/java/org/terasology/engine/logic/console/commands/CoreCommands.java
@@ -69,6 +69,8 @@ import org.terasology.nui.skin.UISkinAsset;
 
 import java.io.IOException;
 import java.net.UnknownHostException;
+import java.nio.file.Path;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


keturn:
Do you have a strong preference for `<timestamp>-entityDump` instead of `entityDump-<timestamp>`? I'm more used to seeing the timestamp as a suffix for logfiles and such.

Unless, I suppose, we're thinking of `.entityDump.json` all as one extension describing the file type. 🤔

engine/src/main/java/org/terasology/engine/logic/console/commands/CoreCommands.java
```diff
@@ -468,30 +470,32 @@ public String leave() {
...
                     " Supports list of component names, which will be used to only save entities that contains" +
                     " one or more of those components. Names should be separated by spaces.")
     public String dumpEntities(@CommandParam(value = "componentNames", required = false) String... componentNames) throws IOException {
         int savedEntityCount;
         EngineEntityManager engineEntityManager = (EngineEntityManager) entityManager;
-        PrefabSerializer prefabSerializer = new PrefabSerializer(engineEntityManager.getComponentLibrary(), engineEntityManager.getTypeSerializerLibrary());
+        PrefabSerializer prefabSerializer =
+                new PrefabSerializer(engineEntityManager.getComponentLibrary(), engineEntityManager.getTypeSerializerLibrary());
         WorldDumper worldDumper = new WorldDumper(engineEntityManager, prefabSerializer);
+        Path outFile = PathManager.getInstance().getHomePath().resolve(Instant.now() + "-entityDump.json");
```
keturn, line 483: :+1: de-duplication

keturn (approved):
Code Looks Good to Me.

I haven't tested it myself. How does one trigger an entity dump?

pollend (approved):

skaldarnar:
@keturn no real preference for the pattern, no. The point about log files having a similar pattern makes sense, and we should probably adopt it here (to group all entity dumps together when sorting by file name). 

I was also thinking about including more info here, like some way to identify where the dump came from (client id or something like that), and also to generalize this method such that it can be used in other cases, e.g., when handling fatal game errors and we crash to the main menu. I believe that the entity data might be helpful in debugging gameplay issues in the future. 

Given that, I merged this as is, but I'm looking forward to more changes in this part of the code. As we don't read the files anywhere, renaming the output file is simple and cheap 👍

PR #4744, opened 2021-06-05
fix: correct extra indices sphere builder

pollend:
looks like my index buffer was requesting indexes outside of the range of my vertex buffer.

before:

image

after:

image


skaldarnar (approved):
tests out fine, no weird triangle in the sky

PR #4747, opened 2021-06-05
chore(ModuleManager): refactor and document engine module setup

### LINKED ISSUE #4639: do not allow modules to import unsafe engine classes
The Malicious module detected a regression in #4622 that allows modules to import _any_ class in the `org.terasology.engine` package.

Related: https://github.com/MovingBlocks/gestalt/pull/112

It would help to have a test for this as part of the normal automated test suite that doesn't require manually launching a game and searching the logs.

keturn:
This is the code that has been sitting around in my working directory that was ostensibly to fix #4639, the failures detected by MaliciousModule.

Upon closer inspection, it seems that https://github.com/MovingBlocks/gestalt/pull/112 is sufficient to fix that error on its own. These are changes (mostly refactors) I did while trying to determine how to write tests or fix the problem without reverting the code in gestalt. They may not be functionally _necessary,_ but maybe worth keeping for de-tangling this code.

This doesn't influence the MTE test failures either. That remains an independent problem.

### How to test

- normal game-with-modules stuff
- start game with MaliciousModule and look in log for messages from MaliciousTestSystem. Ensure it says 2/2 tests passed.

- [x] merge https://github.com/MovingBlocks/gestalt/pull/112


engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -128,40 +130,61 @@ private void loadModulesFromApplicationPath(PathManager pathManager) {
...
         // We need the class predicate to include classes in subsystems and whatnot. We can't change it in an
-        // existing module, so make a new one based on the one from the moduleFactory.
-        Module engine = new Module(
+        // existing module, so make a new one based on the one from the factory.
+        // TODO: expand the ModuleFactory interface to make this whole thing less awkward
+        return new Module(
                 packageModule.getMetadata(),
                 packageModule.getResources(),
-                Collections.emptyList(),
+                classPaths,
```
keturn, line 171: This is potentially a functional change: engine's `Module.moduleClasspaths` is no longer an empty list.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -265,6 +290,30 @@ public ModuleFactory getModuleFactory() {
     }
 
     public PermissionProvider getPermissionProvider(Module module) {
-        return permissionProviderFactory.createPermissionProviderFor(module, module.getClassPredicate());
+        // TODO: This has only one usage. Audit to see if it's worth keeping.
+        return permissionProviderFactory.createPermissionProviderFor(
+                module,
+                x -> false  // modules (other than engine) never contain classpath classes
```
keturn, line 296: This is potentially a functional change, but probably not. The method doesn't seem to be much used, and `false` is probably the thing game modules have in this field anyway.

keturn:
Oh, I pushed to have Jenkins run the tests for me, but I guess if I'm building against that gestalt PR I'll have to run them locally too.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -265,6 +290,30 @@ public ModuleFactory getModuleFactory() {
...
+        );
+    }
+
+    /** Convert URL to File, with support for {@code jar} URLs. */
+    private static File urlToFile(URL url) {
```
keturn, line 301: Do we have any existing utility modules where this could go?

(I took a quick peek in o.t.e.utilities, but got spooked by the cobwebs.)

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: It doesn't seem to make a difference for the Malicious results. At least not in a local dev workspace running under gradle, I haven't ruled out the possibility that it's useful for something in other configurations.

keturn:
> does this help with the MTE errors

Nope! That still appears to be an independent issue.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -66,25 +71,10 @@ public ModuleManager(String masterServerAddress) {
     }
 
     public ModuleManager(String masterServerAddress, List<Class<?>> classesOnClasspathsToAddToEngine) {
-        PathManager pathManager = PathManager.getInstance();  // get early so if it needs to initialize, it does it now
```
keturn, line 69: I'm sure there was _something_ that prompted me to do this with the comment, but now I can't imagine what that would have been. Instantiating PathManager doesn't have any effects that can be seen without doing this same `getInstance` call. e.g. it does _not_ change the process's working directory.

So we'll pretend I'm better-informed now than I was then and remove it.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
jdrueckert: @skaldarnar and I also had a short conversation about o.t.e.utils...
I personally am okay with some "utils" or "common" rummage box to throw things into that don't really fit anywhere else and would live a lonely life if put in some area of the code dedicated to them. Naturally such a rummage box needs cleaning up every once in a while in case some of the things in there suddenly become clusterable due to new similar things being dropped in there.
Skal seemed more inclined to have them live a lonely life in an area dedicated to them and similar things that we can envision being added in the mid-term future.

jdrueckert (approved):
I'm not sure I completely followed _everything_ that happens in this PR, but I also didn't see anything that really stood out to me in a negative way.
The added comments certainly do help and I believe the todos are a good basis for future improvements here.

So same as with Michael's rendering PRs: From what I saw I don't think it should break anything (although I'm not entirely sure), but I do think it improves things and goes into the right direction, so I'm approving this.

PR #4749, opened 2021-06-05
build(Jenkinsfile): improve pipeline performance; add docstrings

skaldarnar:
If a quality gate is violated the build becomes unstable. The failed stages are reported as red crosses in Github.

The quality gates for Checkstyle, PDM, and Spotbugs prevent us from making changes which introduce new warnings.

In addition, skipping "blames" speeds up the build, and finger pointing anyways never helped anything.

This also includes the changes from MovingBlocks/TeraConfig#16 to the checkstyle configuration.


skaldarnar:

Jenkinsfile
```diff
@@ -57,25 +57,30 @@ node ("ts-engine && heavy && java8") {
...
+        sh './gradlew --console=plain check -x test'
+        discoverGitReferenceBuild(defaultBranch: 'develop') //TODO: does this also work for PRs with different base branch?
+        recordIssues skipBlames: true, qualityGates: [[threshold: 1, type: 'NEW', unstable: true]], 
+            tool: checkStyle(pattern: '**/build/reports/checkstyle/*.xml')
+        recordIssues skipBlames: true, qualityGates: [[threshold: 1, type: 'NEW', unstable: true]],
+            tool: spotBugs(pattern: '**/build/reports/spotbugs/main/*.xml', useRankAsPriority: true)
+        recordIssues skipBlames: true, qualityGates: [[threshold: 1, type: 'NEW', unstable: true]],
+            tool: pmdParser(pattern: '**/build/reports/pmd/*.xml')
+        recordIssues skipBlames: true, 
+            tool: taskScanner(includePattern: '**/*.java,**/*.groovy,**/*.gradle', lowTags: 'WIBNIF', normalTags: 'TODO', highTags: 'ASAP')
```
keturn, line 70: If you have a bunch of `recordIssues` steps with the same options but different tools, you could refactor to
```yaml
recordIssues skipBlames: true, qualityGates: […], tools: [
  checkStyle(…),
  spotBugs(…),
  …
]
```

keturn:
I'm confused about the difference between Jenkins "quality gate" and "health report," as it looks like those are both things you can base off of the issue counts. I think you've got it right: you can set the quality gate on _new_ issues specifically.

If the goal is to block changes with new issues, should we mark the build as "failed" instead of "unstable"?

Are we treating all checkstyle issues equally, regardless of severity?

Jenkinsfile
```diff
@@ -57,25 +57,30 @@ node ("ts-engine && heavy && java8") {
...
-    }
-
     stage('Analytics') {
-        sh './gradlew --console=plain check -x test spotbugsmain'
-        recordIssues tool: checkStyle(pattern: '**/build/reports/checkstyle/*.xml')
-        recordIssues tool: spotBugs(pattern: '**/build/reports/spotbugs/main/*.xml', useRankAsPriority: true)
-        recordIssues tool: pmdParser(pattern: '**/build/reports/pmd/*.xml')
-        recordIssues tool: taskScanner(includePattern: '**/*.java,**/*.groovy,**/*.gradle', lowTags: 'WIBNIF', normalTags: 'TODO', highTags: 'ASAP')
+        sh './gradlew --console=plain check -x test'
+        discoverGitReferenceBuild(defaultBranch: 'develop') //TODO: does this also work for PRs with different base branch?
```
keturn, line 62: I think that because we run this under a Jenkins multibranch pipeline, we don't need to specify `defaultBranch` here.

skaldarnar: Actually, we do! If left empty, it defaults to `master` which is not the right branch to compare against. In particular, our `develop` (and PRs based on that) have diverged more than 100 commits from `master` and the Git Forensics Plugin cannot resolve a common ancestor build.

Jenkinsfile
skaldarnar: In that case, we would also use the same quality gates for all tools, right?

skaldarnar:
> should we mark the build as "failed" instead of "unstable"?

I started off with that at first. I mainly reverted it, as this would not only fail the build, but stop directly at record stage (severe checkstyle issue? → no record of PMD warnings). This situation becomes better with a single `recordIssues` statement, I guess, but still not really convenient to handle :confused: 

> Are we treating all checkstyle issues equally, regardless of severity?

Yes, currently we treat all issues equally, and not only for checkstyle, but also for PMD and Spotbugs. My idea here is that we should just avoid _any_ new issues (or suppress the warning). Interestingly, we only have ERROR and NORMAL priority warnings, but no HIGH ones :shrug: We probably want to teak the configuration a bit more as in https://github.com/MovingBlocks/TeraConfig/pull/16

Jenkinsfile
keturn: Huh. That's disappointing, as Jenkins multibranch already knows when things are part of a PR, not just some random build happening in a vacuum... oh well. There are very very few PRs based on branches other than develop, so this should be fine.

Jenkinsfile
keturn: I think that's how it would work. In which case you'd leave them separate if you wanted to tune the gate thresholds differently for different tools.

I want there to be some way to change the default for a step's option (e.g. `skipBlames`) for all usages of that step in a stage or pipeline, but it doesn't look like there's actually a general way to do that.

skaldarnar:
I don't know what a good configuration here is =/ Setting the build to "UNSTABLE" if integration tests fail will give `continuous-integration/jenkins/pr-merge` a red :x: but we are currently using that overall result as required check.

If anybody has input on this please let me know. 

As the engine builds are currently all green :heavy_check_mark: we're good here, but I'd like to get module builds into state where we can merge again without getting used to ignoring the :x: See https://github.com/MovingBlocks/ModuleJteConfig/pull/13

jdrueckert:
> Setting the build to "UNSTABLE" if integration tests fail will give continuous-integration/jenkins/pr-merge a red x but we are currently using that overall result as required check.

Just had a look at the test commit for that. The main problem I see is that you don't really see why this build was marked unstable because all stages are green.
I'd prefer removing `pr-merge` from being required and only marking the indiviual stages as required, i.e. on engine:
- `Build`
- `Unit Tests`
- `Analytics`
- `Integration Tests`
For the modules, I'd skip `Integration Tests` and not make it required. This would make the PR mergeable even if the integration tests fail, while keeping the integration test issues visible.
I'm aware that this _could_ result in the side-effects that people "get used to" ignoring failed things, but IMO as long as they are not required that's okay-ish. As soon as we had time and capacities to tackle MTE we can make the integration test stage required for modules, too.

skaldarnar:
Since we have to set this for all modules I'd recommend to synchronize this with API calls. I guess [Update branch protection](https://docs.github.com/en/rest/reference/repos#update-branch-protection) has all the info we need.

What I don't know yet is how the `ci/cloudbees/stage/*` check compares to `ci/cloudbees/tests/*` or `ci/cloudbees/checkstyle`. We want to see all the results, even if one of them fails. It might be that the stage succeeds in the end, but the individual steps fail. We need to figure that out before we set the required checks.

The default configuration might be:
```jsonc
{
    "required_status_checks": {
        "strict": true, // Require branches to be up to date before merging.
        "contexts": [   // The list of status checks to require in order to merge into this branch
            "ci/cloudbees/stage/Build",
            "ci/cloudbees/stage/Analytics",
            "ci/cloudbees/tests/Unit Tests",
            "ci/cloudbees/test/Integration Tests"
        ]
    },
    "required_pull_request_reviews": {
        "dismiss_stale_reviews": true, // Set to true if you want to automatically dismiss approving reviews when someone pushes a new commit.
        "required_approving_review_count": 1 // Specify the number of reviewers required to approve pull requests. Use a number between 1 and 6.
    }
}
```

With the `gh` CLI tool the configuration can be applied via the following command when run inside the respective repo (untested):

```sh
gh api -X PUT /repos/:owner/:repo/branches/develop/protection \
	--header Accept=application/vnd.github.v3+json \
	--input default.json
```

keturn:
Oh, I see, the notes about how things are reported as red or green are about how they show up in _GitHub,_ not Jenkins. I'd missed that on my earlier reading.

Okay, if we're stumbling over how things are reported in GitHub and there's confusion about which checks on the GitHub side line up with which steps and stages in Jenkins, I suggest we pause this here and do the <i><a href="https://docs.cloudbees.com/docs/cloudbees-ci/latest/scm-integration/enabling-scm-reporting#_enabling_github_app_authentication>Use GitHub App Authentication</a></i> thing first.

Why am I suggesting this thing about authentication methods when the current conversation doesn't look like it has anything to do with authentication or credentials? Because, for some esoteric reason I haven't uncovered, the Cloudbees/GitHub integration has features that don't turn on unless it's using that authentication method with the GitHub API.

I expect that will add more detail to what we see on the GitHub side of things, which I hope will help make it clearer about how the stages and checks line up.

I did the proof of concept for that in Nanoware a while back but I stalled out when it came time to translate that proof of concept made with the web UI in to our configuration repo so it would be reproducible.

Jenkinsfile
```diff
@@ -57,25 +72,32 @@ node ("ts-engine && heavy && java8") {
...
     }
 
     stage('Documentation') {
         sh './gradlew --console=plain javadoc'
         step([$class: 'JavadocArchiver', javadocDir: 'engine/build/docs/javadoc', keepAll: false])
-        recordIssues tool: javaDoc()
+        recordIssues skipBlames: true, tool: javaDoc()
+    }
+
+    stage('Integration Tests') {
```
skaldarnar, line 96: I've moved integration tests down as they are long running, and it might be desirable to get fast feedback on the other analytic reports. 

I still think that **Publish** and **Documentation** should be the last stages in the pipeline, but since we decided to publish early and run checks only later I figured that putting the short-running stages first makes sense.

skaldarnar:
I've removed the quality gate for the Warnings Next Generation reports now. In the end, this PR is just a small performance and documentation improvement for the `Jenkinsfile`.

For reference: I've used the following quality gate to prevent any new errors being introduced, regardless of their severity. This gate should not care about the errors that are already in the code base, and should allow us to gradually improve on this. 

I plan to bring this in within the next 1-2 weeks.

```groovy
qualityGates: [[threshold: 1, type: 'NEW', unstable: true]], 
```

keturn (approved):
I think these comments are accurate and things look like they're being recorded correctly on Jenkins. Getting the correct reference build should be a good improvement too. :+1:

PR #4751, opened 2021-06-06
fix: world generation with zones

tolziplohu:
### Contains

When fixing https://github.com/Terasology/TutorialWorldGeneration/pull/19, I noticed that the zoned world generator doesn't work anymore. Only the first zone added to a `WorldBuilder` actually appears in the world. It turns out this is because `WorldBuilder` deduplicates rasterizers based on class (but curiously, only rasterizers which don't have dependencies), so it sees three `Zone`s and decides that they're equivalent and it only needs to rasterize the first one. This is invalid for zones, and it's also invalid for any rasterizer with per-instance configuration, which may not be used now but would be a good thing to support. This PR just gets rid of the deduplication.

### How to test

Checkout https://github.com/Terasology/TutorialWorldGeneration/pull/19, and try creating a world with the `ZonedSimplex` generator. Previously, the ground wouldn't generate and you would fall infinitely; after this PR, it should generate a normal-looking world.

Was there any case where the deduplication that `WorldBuilder` was doing is actually needed? If so, we may want to use some other criteria for deduplication or something.


skaldarnar (approved):
I can't tell what we need the de-duplication for. It might be related to some multi-world setup (not sure how those even work :thinking:).

Let's get this merged! :+1:

tolziplohu:
> Let's get this merged! :+1: 

So um, does that mean I should merge it? I'm new to having write access  :)

PR #4754, opened 2021-06-09
doc: update minimal system requirements for OpenGL 3.3

skaldarnar:
If I understand this correctly, with the latest changes to rendering we now have a **minimum requirement** for a GPU supporting OpenGL 3.3. 
According to Wikipedia, the listed GPUs (both dedicated and integrated) should match that requirement.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -50,15 +50,15 @@ We encourage contributions from anybody and try to keep a warm and friendly comm
     <tr>
         <td>Graphics (GPU)</td>
         <td style="vertical-align:top">
-            Intel HD Graphics (Gen 5)<br/>
-            GeForce 6xxx series or<br/>
-            Radeon HD 2000 series <br/>
-            with OpenGL 2.1*
+            Intel HD Graphics (Gen 7)<br/>
+            GeForce 8xxx series (or higher) or<br/>
+            Radeon HD 2000 series (or higher)<br/>
+            with OpenGL 3.3
         </td>
         <td  style="vertical-align:top">
             GeForce 8xxx series (or higher) or<br/>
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -50,15 +50,15 @@ We encourage contributions from anybody and try to keep a warm and friendly comm
...
+            GeForce 8xxx series (or higher) or<br/>
+            Radeon HD 2000 series (or higher)<br/>
+            with OpenGL 3.3
         </td>
         <td  style="vertical-align:top">
             GeForce 8xxx series (or higher) or<br/>
             Radeon HD 2000 series (or higher)<br/>
-            with OpenGL 3.x
+            with OpenGL 3.3
         </td>
```
jdrueckert, line 62: Minimal and recommended GPU settings are nor almost equal. do we need to keep two different cells here?

skaldarnar: I don't know 🤷‍♂️  we could also bump the recommended GPU part up, although I don't know what would give a player a performance benefit... if we are so unsure about things, maybe we should just keep the minimal requirements and not mention the "recommmended" at all?

DarkWeird: It is hard to calculate.

There needs research (run TS with different GPU and videos settings)
Or .. use standart pair of minimal and recommended. Most games using copy pasted min/rec requirements.

We use only OpenGL 3.3 at any way (before minimal is 2, recommended is 3.3.. after PR about particles)

jdrueckert: I'd say just make one cell with the minimal and span that across the recommended column, too.

jdrueckert (approved):

PR #4719, opened 2021-05-27
style: fix checkstyle findings

jdrueckert:

PR #4758, opened 2021-06-12
style: address checkstyle findings

jdrueckert:
This PR resolves code style issues [reported by checkstyle in our CI](https://jenkins.terasology.io/teraorg/job/Terasology/job/engine/job/develop/531/checkstyle/?cloudbees-analytics-link=scm-reporting%2Fwarnings).
Similar PRs in which we already started this are #4702 or #4719


jdrueckert:

engine/src/main/java/org/terasology/engine/entitySystem/entity/EntityBuilder.java
```diff
@@ -151,9 +151,10 @@ public boolean hasAllComponents(List<Class<? extends Component>> filterComponent
         return component;
     }
 
-    public void addComponents(Iterable<? extends Component> components) {
-        components = (components == null) ? Collections.EMPTY_LIST : components;
-        components.forEach(this::addComponent);
+    public void addComponents(Iterable<? extends Component> maybeComponents) {
+        Optional.ofNullable(maybeComponents).ifPresent(comps -> {
+            comps.forEach(this::addComponent);
+        });
```
skaldarnar, line 157: While `maybeComponents` makes sense for the parameter, I don't like that we have to use `comps` because `components` would hide a name... :confused:

engine/src/main/java/org/terasology/engine/utilities/ReflectionUtil.java
```diff
@@ -265,6 +265,7 @@ public static Method findMethod(Class<?> targetType, String methodName, Class<?>
         return (TypeInfo<E>) TypeInfo.of(elementType);
     }
 
+    @SuppressWarnings("checkstyle:ParameterAssignment")
```
skaldarnar, line 268: Maybe we should add a comment here why this is suppressed.

In this case, we're not sure whether the reassignment is necessary, and/or whether it is a desired property of this method.

engine/src/main/java/org/terasology/engine/input/internal/BindCommands.java
```diff
@@ -14,15 +15,29 @@
...
+            .build();
+    public static final ImmutableMap<Integer, SimpleUri> DVORAK = new ImmutableMap.Builder<Integer, SimpleUri>()
+            .put(KeyId.COMMA, new SimpleUri("engine:forwards"))
+            .put(KeyId.A, new SimpleUri("engine:left"))
+            .put(KeyId.O, new SimpleUri("engine:backwards"))
+            .put(KeyId.E, new SimpleUri("engine:right"))
+            .put(KeyId.C, new SimpleUri("engine:inventory"))
+            .put(KeyId.PERIOD, new SimpleUri("engine:useItem"))
+            .build();
+    public static final ImmutableMap<Integer, SimpleUri> NEO = new ImmutableMap.Builder<Integer, SimpleUri>()
```
skaldarnar, line 33: We should test whether this still works nicely before merge

engine-tests/src/test/java/org/terasology/persistence/typeHandling/gson/GsonTypeHandlerLibraryAdapterFactoryTest.java
```diff
@@ -83,9 +83,14 @@ public boolean equals(Object o) {
             }
             TestClass testClass = (TestClass) o;
             return i == testClass.i
-                    && Objects.equals(color, testClass.color)
-                    && Objects.equals(vector4fs, testClass.vector4fs)
-                    && Objects.equals(rectangleiMap, testClass.rectangleiMap);
+                    && java.util.Objects.equals(color, testClass.color)
+                    && java.util.Objects.equals(vector4fs, testClass.vector4fs)
+                    && java.util.Objects.equals(rectangleiMap, testClass.rectangleiMap);
```
skaldarnar, line 88: Any reason why you're using `java.util.Objects.equals` here instead of `Objects.equal(...)` from Guava?

facades/TeraEd/src/main/java/org/terasology/editor/subsystem/LwjglPortlet.java
```diff
@@ -99,9 +99,10 @@ public void postUpdate(GameState currentState, float delta) {
 
         display.update();
         int frameLimit = context.get(Config.class).getRendering().getFrameLimit();
-        if (frameLimit > 0) {
+        // TODO: do we still need this?
+//        if (frameLimit > 0) {
```
skaldarnar, line 103: ping @pollend - what is the `LwjglPortlet` good for?

engine/src/main/java/org/terasology/engine/entitySystem/event/AbstractValueModifiableEvent.java
```diff
@@ -71,9 +71,7 @@ public float getResultValue() {
      * Calculates the result value from the base value and given modifiers and multipliers.
      * <p>
      * The value is calculated based on the following formula:
-     * <pre>
-     * result = (<baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>
-     * </pre>
```
skaldarnar, line 76: I'm not sure about this change - what was the problem here? 

`{@code ...}` will be an in-line verbatim text, like `this`. However, `<pre>...</pre>` should create a block of "preformatted" text, like
``` 
this
``` 

I intentionally put this formula into a block to be rendered separately.

engine/src/main/java/org/terasology/engine/rendering/dag/Node.java
```diff
@@ -72,13 +72,13 @@
      *
      * In the Console, commands have the following syntax:
      *
-     *     dagCommandNode <nodeUri> <command> <arguments...>
+     * {@code dagCommandNode <nodeUri> <command> <arguments...>}
```
skaldarnar, line 75: this would probably also work better inside `<pre>...</pre>`

engine/src/main/java/org/terasology/engine/rendering/dag/RenderTaskListGenerator.java
```diff
@@ -226,8 +226,8 @@ public void process() { }
         /**
          * Returns a string description of the instance.
          *
-         * @return A string in the form: "----- <message>",
-         *         where <message> is the string passed to the constructor.
+         * @return A string in the form: "----- {@code <message>}",
```
skaldarnar, line 229:
```suggestion
         * @return A string in the form: "{@code----- <message>}",
```

engine/src/main/java/org/terasology/engine/utilities/modifiable/ModifiableValue.java
```diff
@@ -49,9 +49,7 @@ public ModifiableValue postAdd(float amount) {
      * Calculates the result value from the base value and given modifiers and multiplier.
      * <p>
      * The value is calculated based on the following formula:
-     * <pre>
-     * result = (<baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>
-     * </pre>
+     * {@code result = (<baseValue> + Σ <modifier>) * Π <multiplier> + Σ <postModifier>}
```
skaldarnar, line 52: Same comment as before.

engine/src/main/java/org/terasology/engine/entitySystem/event/internal/EventSystemImpl.java
```diff
@@ -134,16 +134,14 @@ public void registerEventHandler(ComponentSystem handler) {
 
     @Override
     public void unregisterEventHandler(ComponentSystem handler) {
-        for (SetMultimap<Class<? extends Component>, EventHandlerInfo> eventHandlers :
-                componentSpecificHandlers.values()) {
-            Iterator<EventHandlerInfo> eventHandlerIterator = eventHandlers.values().iterator();
+        componentSpecificHandlers.values().stream().map(eventHandlers -> eventHandlers.values().iterator()).forEach(eventHandlerIterator -> {
```
skaldarnar, line 137:
```suggestion
        componentSpecificHandlers.values().stream()
            .map(eventHandlers -> eventHandlers.values().iterator())
            .forEach(eventHandlerIterator -> {
```

engine/src/main/java/org/terasology/engine/rendering/nui/layers/mainMenu/SelectGameScreen.java
```diff
@@ -169,8 +170,9 @@ private void loadGame(GameInfo item) {
             final GameManifest manifest = item.getManifest();
             config.getWorldGeneration().setDefaultSeed(manifest.getSeed());
             config.getWorldGeneration().setWorldTitle(manifest.getTitle());
-            CoreRegistry.get(GameEngine.class).changeState(new StateLoading(manifest, (isLoadingAsServer()) ?
-                    NetworkMode.DEDICATED_SERVER : NetworkMode.NONE));
+            Optional.ofNullable(CoreRegistry.get(GameEngine.class))
+                    .orElseThrow(() -> new Exception("Failed to get game engine from CoreRegistry"))
```
skaldarnar, line 174: Throwing a generic `Exception` is usually no good idea. We should probably introduce custom Terasology exceptions for cases such like this, but for a start you could throw [IllegalStateException](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalStateException.html) instead.

skaldarnar (changes requested):

engine/src/main/java/org/terasology/engine/rendering/dag/Node.java
jdrueckert: nope. `pre` blocks don't like `<` and `>`

engine/src/main/java/org/terasology/engine/entitySystem/event/AbstractValueModifiableEvent.java
jdrueckert: `pre` blocks don't like `<` and `>`, that's the problem

engine/src/main/java/org/terasology/engine/utilities/modifiable/ModifiableValue.java
jdrueckert: same response as before

engine-tests/src/main/java/org/terasology/engine/Environment.java
```diff
@@ -128,7 +128,7 @@ protected void setupBlockManager(AssetManager assetManager) {
         // empty
     }
 
-    protected void setupExtraDataManager(Context context) {
+    protected void setupExtraDataManager(Context ctxt) {
```
skaldarnar, line 131: This is another instance of HiddenFieldCheck where I don't like either of the options:

- add a bloated `@SuppressWarnings(...)` annotation
- rename the parameter to `ctxt` or `newContext` or similar, actually making it worse to read 

@keturn how would you resolve these warnings (I think you somewhat defended that check rule before)? Unfortunately, configuration options are limited...

engine-tests/src/test/java/org/terasology/engine/persistence/serializers/TypeSerializerTest.java
```diff
@@ -150,7 +150,8 @@ public String toString() {
         }
     }
 
-    private static final class Animal<T> {
+    @SuppressWarnings("checkstyle:FinalClass")
```
skaldarnar, line 153: why does checkstyle think this should be final? I mean, it literally is extended a couple of lines later... is it because of the `private` constructor (which should probably be `protected` instead)?

engine-tests/src/test/java/org/terasology/persistence/typeHandling/gson/PolymorphicTypeAdapterFactoryTest.java
```diff
@@ -92,7 +92,8 @@ private Capsule(Animal animal) {
     private interface Walker {
     }
 
-    private static final class Animal implements Walker {
+    @SuppressWarnings("checkstyle:FinalClass")
```
skaldarnar, line 95: Same question here...

engine/src/main/java/org/terasology/engine/config/flexible/ui/ConstraintWidgetFactory.java
```diff
@@ -32,8 +32,8 @@ protected C getConstraint() {
         return (C) setting.getConstraint();
     }
 
-    public Optional<UIWidget> buildWidgetFor(Setting<T> setting) {
-        this.setting = setting;
+    public Optional<UIWidget> buildWidgetFor(Setting<T> targetSetting) {
```
skaldarnar, line 35: Oh, I saw this method before. I'm wondering whether it should get a different name, as I think the behavior is pretty confusing. 

I'd like to deprecate this method and instead introduce a fluent API setter (similar to build pattern) like this:

```java
    public ConstraintWidgetFactory<T, C> withSetting(Setting<T> setting) {
        this.setting = setting;
        return this;
    }
```

This would allow to do 

```
factory.withSetting(setting).buildWidget();
```

instead of 
```
factory.buildWidgetFor(setting);
```

What I find confusing here is that building a widget actually modifies the state of the factory... and it seems to be the only way to set the setting in the first place :thinking: 

I'm just confused here...

engine/src/main/java/org/terasology/engine/rendering/assets/skeletalmesh/SkeletalMeshDataBuilder.java
```diff
@@ -32,18 +32,18 @@ public SkeletalMeshDataBuilder addBone(Bone bone) {
         return this;
     }
 
-    public SkeletalMeshDataBuilder addVertices(List<Vector3f> vertices) {
-        this.vertices.addAll(vertices);
+    public SkeletalMeshDataBuilder addVertices(List<Vector3f> verticesToAdd) {
```
skaldarnar, line 35: I think these renames are unnecessary and the code would be better readable if we leave it as is...

engine/src/main/java/org/terasology/engine/rendering/dag/AbstractNode.java
```diff
@@ -94,9 +94,9 @@ public void setOutputConnections(Map<String, DependencyConnection> outputConnect
         this.outputConnections = outputConnections;
     }
 
-    public final void postInit(Context context) {
+    public final void postInit(Context ctxt) {
```
skaldarnar, line 97: :confused:

engine/src/main/java/org/terasology/engine/rendering/dag/ModuleRendering.java
```diff
@@ -78,8 +78,8 @@ public void setInitPriority(int initPriority) {
     }
 
     protected void setProvidingModule(Class implementingClass) {
-        ModuleManager moduleManager = context.get(ModuleManager.class);
-        this.providingModule = moduleManager.getEnvironment().getModuleProviding(implementingClass);
+        ModuleManager manager = context.get(ModuleManager.class);
```
skaldarnar, line 81: Why can't we use the member field here? It is initialized exactly the same way, so this should be unnecessary.

engine/src/main/java/org/terasology/engine/rendering/nui/layers/mainMenu/UniverseSetupScreen.java
```diff
@@ -357,8 +357,8 @@ private int findIndex(List<WorldSetupWrapper> worldsList, String worldName) {
 
     private void initAssets() {
 
-        ModuleEnvironment environment = context.get(ModuleManager.class).getEnvironment();
-        BlockFamilyLibrary library =  new BlockFamilyLibrary(environment, context);
+        ModuleEnvironment env = context.get(ModuleManager.class).getEnvironment();
```
skaldarnar, line 360: We have the `environment` as class member, but it is only written in `setEnvironment`. 

It feels wrong to me that `initAssets` resolves the environment again, or, in case it gets called first, does not initialize the environment. To make this even more weird, we also have the `moduleManager` as class member, yet resolve it here again via the context :see_no_evil:

engine/src/main/java/org/terasology/engine/rendering/nui/layers/mainMenu/moduleDetailsScreen/ModuleDetailsScreen.java
```diff
@@ -286,9 +286,9 @@ public String get() {
         });
     }
 
-    private boolean validateModuleDependencies(Name moduleName) {
+    private boolean validateModuleDependencies(Name module) {
```
skaldarnar, line 289: I do think that nearly all of the renames should be reverted. These renames kind of work and are not bad per se, but it should not be necessary.

Part of my reasoning is that I do think that having a class member `UILabel moduleName` is fine and can happily co-exist with a parameter `Name moduleName`.
I'm not a fan of repeating the type in the name, e.g., naming the class member `moduleNameLabel` instead - there is a type system, why should I put it in the name...

jdrueckert: I reverted the HiddenFields related commit.

skaldarnar (approved):

PR #4760, opened 2021-06-12
chore(rendering): migrate use of ARB bind framebuffer EXT and reworked readBackPixels 

pollend:
I modified the API so you just read back the buffer that was submitted to the API. should make things a bit simpler. avoid an extra copy of the buffer when you are reading it back and if need be it can just be passed to another buffer if the data is used in some other way. 

required PR: https://github.com/Terasology/CoreRendering/pull/64


jdrueckert (approved):
Tested this out in `CoreGameplay`. Everything looked as usual so I don't think this breaks anything.

PR #4762, opened 2021-06-13
chore: reusue Floatbuffer for uniform

pollend:
every allocation of a direct buffer will also zero out the memory. for some of the unfirom operations we end up  doing these operations for a non trivial amount so it makes sense to reuse the buffer between calls. Materials are allocated once so there isn't a heavy cost with allocating a smaller buffer upfront if it can help with performance.


jdrueckert (approved):
Successfully started `CoreGameplay` with this. Everything looked as usual, so I don't think this breaks anything.

PR #4765, opened 2021-06-13
chore: cleanup migrate shaders into CoreRendering

pollend:
I've migrated out all the the post processing shaders and moved them into CoreRendering. should let us keep shaders in sync with CoreRendering module without having to make changes in engine to accommodate the changes to a shader. I would like to work on the DAG pipeline a bit and understand a more consistent and easier to use API.


engine/src/main/resources/org/terasology/engine/assets/shaders/blockSelection.info
```diff
@@ -1,7 +1,7 @@
 {
     "params": [
         {
-            "name": "texture",
+            "name": "tex",
```
jdrueckert, line 4: Why? Is there some name clash with `texture`?

jdrueckert (approved):

engine/src/main/resources/org/terasology/engine/assets/shaders/blockSelection.info
pollend: yep you use texture instead of texture2d

PR #4768, opened 2021-06-14
chore: move SideBitFlags into Side

pollend:
Deprecate most of `SideBitFlag` and offer direct replacements in `Side`, basically merging the two classes. 

Respective replacements are documented in the deprecation docstring.

This prepares for eventually removing `SideBitFlag`.


pollend:
@skaldarnar it doesn't need to happen immediately the old interface is still there.

skaldarnar:
> it doesn't need to happen immediately the old interface is still there.

I know, and I think the benefit here is that we are not blocked by this migration even while it's happening, and that we are also not forced to do everything at once. 

However, with just this PR and no follow ups, we just created a second option where there was only one before, with no examples on how to use it or to motivate why the "new" implementation is better than the old one. 
I'm concerned that this will become just another part of the code base that's in a weird shape because we started one thing we never finish.

Therefore, I'd like to repeat my request to at least start making the according PRs such that we can get this finished (i.e., remove `SideBitFlags`) within the next weeks.

skaldarnar (dismissed):

skaldarnar (approved):

skaldarnar (approved):

skaldarnar (approved):

jdrueckert (approved):

PR #4771, opened 2021-06-16
chore: minor cleanup for ChunkProcessingPipeline

pollend:
remove some uncessary imports and replace a linkedlist with an iterable. this avoids a copy where we're moving a list to a linked list.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
+++ b/engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
@@ -13,13 +13,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terasology.engine.monitoring.ThreadActivity;
 import org.terasology.engine.monitoring.ThreadMonitor;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
```diff
@@ -301,8 +300,8 @@ public boolean isPositionProcessing(Vector3ic pos) {
      *
      * @return copy of processing positions
      */
-    public List<Vector3ic> getProcessingPosition() {
-        return new LinkedList<>(chunkProcessingInfoMap.keySet());
+    public Iterable<Vector3ic> getProcessingPosition() {
```
skaldarnar, line 303: @tolziplohu does this conflict with the changes you did on https://github.com/MovingBlocks/Terasology/compare/feat/order-chunks-on-demand?

tolziplohu: No, I actually didn't touch that method.

tolziplohu (approved):
Looks like a good change, doing less work is always good.

PR #4776, opened 2021-06-18
refactor(bootstrap): remove duplicate registerComponents method

keturn:
IntelliJ noticed this was duplicate code. One of the implementations is on a utility class in the same package, so that seems appropriate for the other class to use.


engine/src/main/java/org/terasology/engine/core/bootstrap/EnvironmentSwitchHandler.java
```diff
@@ -49,7 +44,7 @@
 import java.util.Collection;
 import java.util.Optional;
 
-import static com.google.common.base.Verify.verifyNotNull;
+import static org.terasology.engine.core.bootstrap.EntitySystemSetupUtil.registerComponents;
```
skaldarnar, line 47: Not sure whether I like the static import of a method, but on the other hand IDEs are smart enough to show me where this is coming from, and since it's a static method you can treat is just as a function reference...

skaldarnar (approved):

PR #4780, opened 2021-06-20
perf: replace stream with for loop  when retrieving facets.

pollend:
I believe streams are quite a bit more expensive compared to a simple for loop. chunk generation is evaluated quite heavily so I think this should help a little with chunk generation. probably not substantial though but the logic change is pretty simple overall. by no means is this a good comparison since both runs are in a unique world with different scheduling for each thread so the profiling does not really quanitfy any significant change from either run.  chunk generation is going to be harder to reproduce overall unless we want to write a JMH benchmark that better verifies this. 

before:
image
 
 after:
image


skaldarnar:
I started the VisualVM CPU profiler when creating a plain Core Gameplay world. During the whole process I did not move.

- start with render distance "far"
- wait for all chunks to be generated (no more active chunk gen threads)
- increase render distance to "ultra"
- wait for all chunks to be generated (no more active chunk gen threads)

With that setup it does not seem to make a difference, but my system is running way faster than yours it seems...

**Before**
image

**After**
image

skaldarnar:
Did another run with JS 

- LOD chunks: 0
- render distance: far
- create world (seed `hsNAlXJetn4d8a4Z`) and wait for all chunks to be generated

**Before**
image

**After**
image

skaldarnar:
Did another run with JS 

- LOD chunks: **10**
- render distance: far
- create world (seed `hsNAlXJetn4d8a4Z`) and wait for all chunks to be generated

**Before**
image

**After**
image

skaldarnar (approved):

pollend:
I have a ryzen processor so I have a lot of threads I could throw at the problem. wonder why we only allocate 4 @skaldarnar

PR #4783, opened 2021-06-21
refactor(rendering): rework chunk mesh generation with RXJava

pollend:
This is a rework of the chunk generation logic with reactiveX. I do away with ChunkMeshUpdateManager and remove pending mesh from Chunk since all the state information is contained within the Observable. I also added a way to sync the mesh updates back onto the main thread so all the mesh data is generated on a background thread and passed back over to the main thread to upload the data to opengl. 

Contributes to #4798


engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
```diff
@@ -68,9 +68,6 @@ public int getIndex() {
 
     private boolean disposed;
 
-    /* CONCURRENCY */
-    private ReentrantLock lock = new ReentrantLock();
-
```
pollend, line 73: no need this is all single threaded.

engine/src/main/java/org/terasology/engine/rendering/world/RenderableWorldImpl.java
```diff
@@ -386,9 +372,44 @@ public int queueVisibleChunks(boolean isFirstRenderingStageForCurrentFrame) {
...
             for (Chunk chunk : chunksInProximityOfCamera) {
-                if (isChunkValidForRender(chunk) && (chunk.isDirty() || !chunk.hasMesh())) {
+                if (isChunkValidForRender(chunk) && (chunk.isDirty() || !chunk.hasMesh()) && !chunkMeshProcessing.containsKey(chunk.getPosition())) {
                     statDirtyChunks++;
-                    chunkMeshUpdateManager.queueChunkUpdate(chunk);
+                    final Chunk currentChunk = chunk;
+                    chunkMeshProcessing.put(chunk.getPosition(), Observable
+                            .just(currentChunk)
+                            .observeOn(Schedulers.io())
+                            .delay(500 * processedChunks, TimeUnit.MILLISECONDS) // delay chunks so everything is not all evaluated at once
```
pollend, line 381: stagger the evaluation so all the chunks are not all submitted at once. helps with keeping frametimes up.

engine/src/main/java/org/terasology/engine/core/GameThread.java
```diff
@@ -23,6 +27,21 @@
     private static volatile Thread gameThread;
     private static BlockingDeque<Runnable> pendingRunnables = Queues.newLinkedBlockingDeque();
 
+    private static final Scheduler MAIN;
+
+    static {
+        MAIN = Schedulers.from(new Executor() {
+            @Override
+            public void execute(@NotNull Runnable runnable) {
```
skaldarnar, line 35: This is probably generated. We haven't agreed on any annotations to use, thus I think `@NotNull` should be removed.

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
```diff
@@ -95,46 +92,39 @@ public boolean hasVertexElements() {
...
+        return true;
     }
 
     private void generateVBO(RenderType type) {
         VertexElements elements = vertexElements[type.ordinal()];
         int id = type.getIndex();
         if (!disposed && elements.buffer.elements() > 0) {
-            vertexBuffers[id] = GL15.glGenBuffers();
-            idxBuffers[id] = GL15.glGenBuffers();
+            vertexBuffers[id] = GL30.glGenBuffers();
```
skaldarnar, line 119: Is this a change we should just do separated from this PR?

engine/src/main/java/org/terasology/engine/rendering/world/RenderableWorldImpl.java
skaldarnar: How big is `processedChunks` on average? Delaying by half a second multiplied with that number could be quite some time...

skaldarnar (commented):
I'll mark this PR as "draft" as per discussion in #architecture.

tolziplohu (commented):
This does seem to improve performance, especially along with https://github.com/MovingBlocks/Terasology/pull/4773. The delay on processing chunks is actually noticeable, though, and results in chunks that you can move through but can't see yet. Removing the delay does hurt performance, but at least on my computer, setting it to `50 * processedChunks` instead of `500 * processedChunks` made the delay unnoticeable without impacting performance. Still, that's very machine-dependent; is there some way to do it without a hardcoded delay?

skaldarnar:
@pollend why did you close this PR? I'd like to understand your reasoning here (What is better or worse with this approach than with the other in #4786?)

PR #4784, opened 2021-06-22
refactor: convert Jenkinsfile to declarative pipeline syntax

jdrueckert:


jdrueckert:

Jenkinsfile
```diff
@@ -17,7 +18,7 @@ properties([
 /**
  * Main pipeline definition for building the engine.
  *
- * It uses the Scripted Pipeline Syntax.
+ * It uses the Declarative Pipeline Syntax.
  * See https://www.jenkins.io/doc/book/pipeline/#declarative-versus-scripted-pipeline-syntax
```
skaldarnar, line 22: Since it uses declarative syntax now, I think we can now better link to https://www.jenkins.io/doc/book/pipeline/syntax/

keturn:
This is clearly working. It's running the pipeline, the steps are all labelled, it's recording the results, etc.

but my head hurts because there were some things I _assumed_ must be true about the Declarative Pipeline syntax, and this is refuting my assumptions.

For instance, I assumed that the way for Jenkins to support a different syntax was for that `pipeline {}` to be the only block, then it knows it's supposed to use the declarative parser for that file.

I assumed that the trick of converting these pipelines would be those `specialBranch` and `artifactBuildsToKeep` definitions.

But you just left them in there as-is? And it doesn't complain about it at all?

Sheesh.

keturn:
So does "declarative pipeline syntax" actually use a different syntax at all, or are declarative Jenkinsfiles still evaluated as groovy and those are just a different set of methods, methods that are designed as a builder instead of doing immediate execution?

Jenkinsfile
```diff
@@ -30,84 +31,129 @@ properties([
...
+                    sh '''./gradlew
+                        --console=plain
+                        -Dorg.gradle.internal.publish.checksums.insecure=true
+                        publish
+                        -PmavenUser=${artifactoryUser}
+                        -PmavenPass=${artifactoryPass}
+                    '''
+                }
+                script {
+                    // Trigger the Omega dist job to repackage a game zip with modules
```
keturn, line 101: Was not this inside the `specialBranch` conditional before? I agree it should be!

I'm not sure if there's a disadvantage to using a `script` block like this. I'm okay with assuming it's fine until we learn otherwise.

jdrueckert:
> So does "declarative pipeline syntax" actually use a different syntax at all, or are declarative Jenkinsfiles still evaluated as groovy and those are just a different set of methods, methods that are designed as a builder instead of doing immediate execution?

@keturn according to https://www.jenkins.io/doc/book/pipeline/#declarative-versus-scripted-pipeline-syntax they are rather similar in "many of the individual syntactical components", so I guess from the basic structure they look pretty similar. The key difference seems to be the `pipeline` (declarative) vs the `node` (scripted) blocks.  As those are also the outermost parts, I guess that's where the parser get the info whether it's facing a scripted or declarative pipeline.

keturn (approved):
I've tested to make sure that `artifactBuildsToKeep` setting is honored. Looks good so far and we should have more confirmation in another six minutes or so.

I haven't tested the Publish block (because PR branches don't get published), but I'm optimistic enough to go ahead and see how it works in practice.

Jenkinsfile
jdrueckert: The problem I faced here is that ther is this `when` clause to only run a stage if a specific condition is met, but from what I found this is only available for whole stages, not for individual steps.
The alternative would be to create four different stages, each with the matching `when expression`.
I figured that with the script block we avoid the duplication and additional stages shown in the ui (of which at max one will ever be executed in a pipeline run).

jdrueckert: And no, this was outside of the `specialBranch` conditional also before. Directly after it, to be precise.

keturn:
Suggestions for future improvements:

* Move `discoverGitReferenceBuild` as early as possible.
* Publish stage: Refactor "rebuild Omega after engine trunk build" to not need `script` block
* Analytics and Documentation stages: move `recordIssues' to `post` block
* taskScanner: configure tags (haven't been using `WIBNIF` or `ASAP`)
* add `recordIssues tool: java` (to Build stage?)

jdrueckert:
Let's see in the master build that's being triggered now, whether the publish stage behaves properly.
If everything looks as expected, I'll migrate the module pipeline accordingly.

PR #4786, opened 2021-06-23
chore(rendering): migrate chunk mesh generation Flux

pollend:
This is a rework of the chunk generation logic with Project Reactor. I do away with ChunkMeshUpdateManager and remove pending mesh from Chunk since all the state information is contained within the Observable. I also added a way to sync the mesh updates back onto the main thread so all the mesh data is generated on a background thread and passed back over to the main thread to upload the data to opengl.

Contributes to #4798


engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
```diff
@@ -223,33 +207,28 @@ public int render(RenderPhase type) {
...
-                    id = vaoCount[i];
-                    if (id != 0) {
-                        GL30.glDeleteVertexArrays(id);
-                        vaoCount[i] = 0;
-                    }
+        if (!disposed) {
+            for (int i = 0; i < vertexBuffers.length; i++) {
+                int id = vertexBuffers[i];
+                if (id != 0) {
+                    GL15.glDeleteBuffers(id);
```
skaldarnar, line 214: Why use `GL15` but change other occurrences to `GL30`?

skaldarnar:
What's the difference between this and #4783?

pollend:
> What's the difference between this and #4783?

Just trying something different.

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
pollend: they end up mapping to the same call.

skaldarnar:
This also works quite well, and I don't see any noticeable stuttering or delays. However, the overall performance seems to be lower than of the previous PR, as I can easily reach the border of generated chunk (with `setSpeedMultiplier 5`).

pollend:
> This also works quite well, and I don't see any noticeable stuttering or delays. However, the overall performance seems to be lower than of the previous PR, as I can easily reach the border of generated chunk (with `setSpeedMultiplier 5`).
> 
> I tested this together with #4773 in a Metal Renegades world.

try increasing the parallelism. I have it at 4 but we could go with 5/6

keturn (commented):
I like the idea of getting rid of things like the `tryLock` in `generateVBOs` and fields named `pendingMesh`!

I think the big question I'm going to have with this one is: How is the test coverage?

'cuz I'm looking at that big flux definition in RenderableWorldImpl, and it's defining a lot of logic and also dispatching it out to threads in the class's constructor. There are enough conditionals in there that I think tests would be worthwhile, and I expect we'll need to factor it out of the constructor to make it accessible to testing.

pollend:
> I like the idea of getting rid of things like the `tryLock` in `generateVBOs` and fields named `pendingMesh`!
> 
> I think the big question I'm going to have with this one is: How is the test coverage?
> 
> 'cuz I'm looking at that big flux definition in RenderableWorldImpl, and it's defining a lot of logic and also dispatching it out to threads in the class's constructor. There are enough conditionals in there that I think tests would be worthwhile, and I expect we'll need to factor it out of the constructor to make it accessible to testing.

umm I guess you would need to stub out ChunkTesselator and get around all the opengl calls :?

keturn (commented):
Ideas about what to test in that flux pipeline (currently in the constructor of RenderableWorldImpl):

- `distinct` removes duplicates, but then there's some logic later that removes things from the set of "seen" items after they're processed.
  - test sending events that _should_ be de-duplicated
  - test sending events that should _not_ be de-duplicated
- `setDirty(false)` - I'm not clear on what "dirty" means in this context. Is this set at the right time?
- `chunk -> chunk, mesh`:
  - test non-valid chunk views. (`isValidView` is also documented as "should be checked within a lock")
  - ChunkMonitor: suggest moving this call out of this function, it can listen to the events downstream instead
- chunksInProximityOfCamera: What do we do with things that aren't in this set? We spent CPU time on building the mesh and then we just drop it?
- Consumer<Chunk, Mesh>
  - this runs on the main thread and is important-looking. Worth giving its own name for easier identification in traces and profiling?
  - could test to make sure discard and dispose methods are called. Boring tests, but I'm guessing bad things happen if we leave those out, so it could be worth the redundancy. Unless there's another integration test somewhere else making sure we don't have that type of resource leak.

keturn:
> umm I guess you would need to stub out ChunkTesselator and get around all the opengl calls

If the Tessellator is something you don't want to get involved in these unit tests, this pipeline can be broken down into two parts: 
* the part that filters this list of chunks and maps them to ChunkViews to feed to the tessellator, and
* the part that does stuff with the result of the tessellator

Why do we need to get around the OpenGL calls? Do they throw exceptions when run in unit tests?

engine/src/main/java/org/terasology/engine/rendering/world/RenderableWorldImpl.java
```diff
@@ -68,20 +74,25 @@
...
-    private final RenderingConfig renderingConfig = config.getRendering();
+    private final Config config;
+    private final RenderingConfig renderingConfig;
 
     private int statDirtyChunks;
     private int statVisibleChunks;
     private int statIgnoredPhases;
 
+    private final Set<Vector3ic> chunkMeshProcessing = Sets.newConcurrentHashSet();
+    private final Sinks.Many<Chunk> chunkMeshPublisher = Sinks.many().unicast().onBackpressureBuffer();
```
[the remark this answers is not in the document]
pollend: > > umm I guess you would need to stub out ChunkTesselator and get around all the opengl calls
> 
> If the Tessellator is something you don't want to get involved in these unit tests, this pipeline can be broken down into two parts:
> 
>     * the part that filters this list of chunks and maps them to ChunkViews to feed to the tessellator, and
> 
>     * the part that does stuff with the result of the tessellator
> 
> 
> Why do we need to get around the OpenGL calls? Do they throw exceptions when run in unit tests?



> > umm I guess you would need to stub out ChunkTesselator and get around all the opengl calls
> 
> If the Tessellator is something you don't want to get involved in these unit tests, this pipeline can be broken down into two parts:
> 
>     * the part that filters this list of chunks and maps them to ChunkViews to feed to the tessellator, and
> 
>     * the part that does stuff with the result of the tessellator
> 
> 
> Why do we need to get around the OpenGL calls? Do they throw exceptions when run in unit tests?

I believe it just crashes if you don't load the lwgjl library. not sure what happens on a headless environment if you submit opengl calls without any kind of display driver :?.

keturn:
[Reactor: ReadWriteStorageManager engine (persistence)](https://trello.com/c/Be5t9Wtr/212-reactor-readwritestoragemanager-engine-persistence)

keturn:
[Reactor: ChunkMesh engine (rendering)](https://trello.com/c/7h2mypT9/213-reactor-chunkmesh-engine-rendering)

keturn:
Oh boy: it's just over six weeks until this PR's birthday.

Reviewing the comments above, it looks like the blocker here was that it either didn't have tests or the tests it had weren't passing in CI.

Now I've merged #4987 and we have passing tests!

I'm sure there are follow-up things I would have liked to do (following a "make tests to pin down behavior, then change implementation while supported by tests" pattern), but I don't remember any of those things being necessary for this pass.

I think it's time to re-review this, do a little more hands-on testing to make sure merging #4987 didn't foul things up too badly, and maybe we can finally wrap this thing up.

pollend:
This looks good to me. I don't see anything wrong per say.

keturn (approved):
alright, let's do some merging!

PR #4791, opened 2021-06-24
fix(ci): backslash-escape multi-line shell command

keturn:
fixes failure in Publish step

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -89,11 +89,11 @@ pipeline {
                 withCredentials([usernamePassword(credentialsId: 'artifactory-gooey', \
                                                     usernameVariable: 'artifactoryUser', \
                                                     passwordVariable: 'artifactoryPass')]) {
-                    sh '''./gradlew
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -89,11 +89,11 @@ pipeline {
                 withCredentials([usernamePassword(credentialsId: 'artifactory-gooey', \
                                                     usernameVariable: 'artifactoryUser', \
                                                     passwordVariable: 'artifactoryPass')]) {
-                    sh '''./gradlew
-                        --console=plain
-                        -Dorg.gradle.internal.publish.checksums.insecure=true
-                        publish
-                        -PmavenUser=${artifactoryUser}
+                    sh '''./gradlew \\
```
jdrueckert, line 92: do we really need two? 🤔 
also I noticed, we probably need double-quotes to allow variable substitution for the creds.

jdrueckert (approved):

PR #4792, opened 2021-06-24
fix: failing publish stage

jdrueckert:
- use double-quotes instead of single-quotes to allow variable substitution
- add backslashes to continue command instead of having multiple in the multiline sh block


jdrueckert:
@keturn beat me to it 😁 https://github.com/MovingBlocks/Terasology/pull/4791

PR #4793, opened 2021-06-25
chore: minor cleanup Block

pollend:
just some minor things I noticed when looking through Block.


jdrueckert (approved):
Omega still compiled and Core Gameplay still seemed to work as expected.

PR #4794, opened 2021-06-25
perf: improve performance of mesh generator for chunks

pollend:
some minor performance tweaks to mesh generation. so one bottleneck I found is we evaluate the sides of all the blocks for every block within the chunks so each block will check all its ordinals even if it is lets say its air which has no geometry. API is coupled to the block so mesh information is defined by the block itself. would like this to be more data oriented but that is something to consider later. 

I also swapped out the maps for switch cases for BlockParts which should also help to some extent. 

hard to test all the permutations so just verifying that chunks render correctly for a couple different types of blocks will be sufficient.


engine/src/main/java/org/terasology/engine/rendering/primitives/BlockMeshGeneratorSingleShape.java
```diff
@@ -24,7 +24,10 @@ public BlockMeshGeneratorSingleShape(Block block) {
 
     @Override
     public void generateChunkMesh(ChunkView view, ChunkMesh chunkMesh, int x, int y, int z) {
-        final Block selfBlock = view.getBlock(x, y, z);
+        final BlockAppearance blockAppearance = block.getPrimaryAppearance();
+        if (!blockAppearance.hasAppearance()) {
+            return;
+        }
```
pollend, line 30: we were just evaluating all air blocks so we check the ordinals of all the empty blocks :/

engine/src/main/java/org/terasology/engine/math/Side.java
```diff
@@ -116,10 +116,8 @@ public static Side inHorizontalDirection(double x, double z) {
      * method.
      *
      * @return All available sides'
-     * @deprecated use {@link Side#values()}
      */
-    @Deprecated
-    public static EnumSet<Side> getAllSides() {
+    public static List<Side> getAllSides() {
```
pollend, line 120: looks like values() might add a lot of overhead the only way to make this "immutable" is to always return a new array :/. will have to use ImmutableList.

jdrueckert: @pollend Does this mean, you'll do another commit on this?

pollend: will have to rework what I considered deprecated because it adds a lot of overhead. :/

skaldarnar: [Memory-Hogging Enum.values() Method](https://dzone.com/articles/memory-hogging-enumvalues-method) is a good read on that topic. 

After reading through this, I think we should keep the cached collection of values and add docstrings to explain why we're doing this, and why it is preferable to use `getAllSides()` over `.values()`. 

@pollend why did you change the [EnumSet](https://docs.oracle.com/javase/7/docs/api/java/util/EnumSet.html) to a `List`? Is it to guarantee immutability? If we're concerned about the footprint of the collection, we could also use [`Guava#immutableEnumSet`](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/Sets.html#immutableEnumSet(E,E...))

engine/src/main/java/org/terasology/engine/world/block/BlockPart.java
```diff
@@ -32,38 +29,36 @@
...
+            case BACK:
+                return BACK;
+            case FRONT:
+                return FRONT;
+            case TOP:
+                return TOP;
+            default:
+                break;
+        }
+        return null;
```
skaldarnar, line 49: Do you think returning null is a good idea here? We should at least document this behavior. I'd prefer throwing an exception in the `default` case here, as it should never occur.

```suggestion
            default:
                throw new IllegalArgumentException();
        }
```

**Edit:** Wait, there's the `CENTER` block part which has no corresponding side... :thinking: but I still think that the `default` case cannot appear.

engine/src/main/java/org/terasology/engine/world/block/BlockPart.java
```diff
@@ -19,10 +18,8 @@
     BOTTOM(Side.BOTTOM),
     CENTER;
 
-    private static final EnumMap<Side, BlockPart> SIDE_MAP;
-    private static final BlockPart[] SIDES;
-    private static final List<BlockPart> HORIZONTAL_SIDES;
-
+    private static final List<BlockPart> HORIZONTAL_SIDES = ImmutableList.of(LEFT, RIGHT, FRONT, BACK);
+    private static final List<BlockPart> SIDES = ImmutableList.of(TOP, LEFT, RIGHT, FRONT, BACK, BOTTOM, CENTER);
```
skaldarnar, line 22: This is just the full set of `values()` - maybe using that makes it bit more obvious that this are just all enum values - basically the cached `.values()` as discussed above.

```suggestion
    /**
     * Cached block parts in an immutable list.
     * See https://dzone.com/articles/memory-hogging-enumvalues-method
     */
    private static final List<BlockPart> SIDES = ImmutableList.of(BlockPart.values());
```

engine/src/main/java/org/terasology/engine/rendering/primitives/BlockMeshGeneratorSingleShape.java
```diff
@@ -24,20 +24,22 @@ public BlockMeshGeneratorSingleShape(Block block) {
 
     @Override
     public void generateChunkMesh(ChunkView view, ChunkMesh chunkMesh, int x, int y, int z) {
-        final Block selfBlock = view.getBlock(x, y, z);
+        final BlockAppearance blockAppearance = block.getPrimaryAppearance();
+        if (!blockAppearance.hasAppearance()) {
+            return;
```
skaldarnar, line 29:
```suggestion
            // perf: Skip mesh generation for blocks without appearance, e.g., air blocks.
            return;
```

engine/src/main/java/org/terasology/engine/rendering/primitives/BlockMeshGeneratorSingleShape.java
skaldarnar: this is useful information we should add as docstring comment.

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkTessellator.java
```diff
@@ -43,9 +43,7 @@ public ChunkMesh generateMesh(ChunkView chunkView, float scale, int border) {
             for (int z = 0; z < Chunks.SIZE_Z; z++) {
                 for (int y = 0; y < Chunks.SIZE_Y - border * 2; y++) {
                     Block block = chunkView.getBlock(x, y, z);
-                    if (block != null && block.getMeshGenerator() != null) {
```
skaldarnar, line 46: Why is it safe to remove these checks?

skaldarnar (changes requested):

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkTessellator.java
pollend: there is a default definition and I added a precondition where this can't be null for setMeshGenerator

engine/src/main/java/org/terasology/engine/math/Side.java
```diff
@@ -44,14 +45,26 @@
     /**
      * @return The horizontal sides, for iteration
      */
-    public static ImmutableList<Side> horizontalSides() {
+    public static List<Side> horizontalSides() {
         return Y_TANGENT_SIDE;
     }
 
+    /**
+     * This provides a static EnumSet of all Sides defined in the enumeration. The result contains the same values as
```
skaldarnar, line 53: The method does not return an EnumSet but an ImutableList. Please adjust the docstring.

engine/src/main/java/org/terasology/engine/math/Side.java
```diff
@@ -44,14 +45,26 @@
...
+     * calling {@code Side#values} but this does not create a new copy on every call. <br/>
+     * <b>Warning:</b> Do not change the content of the returned enum set! It will be reflected on all calls to this
```
skaldarnar, line 55: If we're returning an immutable list this warning is not necessary.

engine/src/main/java/org/terasology/engine/math/Side.java
```diff
@@ -44,14 +45,26 @@
...
+     * method.
+     *
+     * @return All available sides'
```
skaldarnar, line 58: typo: stray `'` at the end

skaldarnar (changes requested):

engine/src/main/java/org/terasology/engine/math/Side.java
skaldarnar: @pollend could you answer this, please?

pollend: an enumset tries to pack the data wonder if that has an impact on performance.

skaldarnar:
@pollend did you check the rest of Omega for any broken code with this? At least in a MetalReneages workspace this does not compile:

```
> Task :modules:MetalRenegades:compileJava FAILED
/home/skaldarnar/Code/movingblocks/ts-mr/modules/MetalRenegades/src/main/java/org/terasology/metalrenegades/world/dynamic/roads/RailRasterizer.java:65: error: cannot find symbol
        for (Side side : Side.getAllSides()) {
                             ^
  symbol:   method getAllSides()
  location: class Side
```

skaldarnar (approved):

PR #4795, opened 2021-06-25
chore: remove finite check from LocalPlayer 

pollend:
we've mentioned in the past having a situation where the player position is placed at (0,0,0). I'm suspecting it has something to do with this logic. so for an invalid position we just return dest or a new Vector3/Quaternion. we shouldn't be just defaulting to 0,0,0. this ends up hiding a larger underlying problem where we see some odd behavior but we've written this logic in a way to hide a large problem. I rather have the game crash with a stacktrace then try poking at thing in the dark because we've written ourselves into a corner.

#4152 https://github.com/MovingBlocks/Terasology/pull/4662


pollend:
want to test this with multiplayer but I think I've mostly addressed the broken cases

jdrueckert (approved):

Issue #4757, opened 2021-06-12
MTE tests are broken after gestalt v7 migration


skaldarnar:
Currently the module integration tests using Module Testing Environment (MTE) are all broken. 

Locally, they fail with a `VerifyException` as the testing environment is unable to find a module for a sepcific component. As far as I can observe, this is not always the same component, but the stacktrace looks quite similar every time.

```stacktrace
com.google.common.base.VerifyException: Could not find module for MappedType class org.terasology.unittest.stubs.MappedTypeComponent
	at com.google.common.base.Verify.verify(Verify.java:124)
	at com.google.common.base.Verify.verifyNotNull(Verify.java:500)
	at org.terasology.engine.core.bootstrap.EnvironmentSwitchHandler.registerComponents(EnvironmentSwitchHandler.java:197)
	at org.terasology.engine.core.bootstrap.EnvironmentSwitchHandler.handleSwitchToGameEnvironment(EnvironmentSwitchHandler.java:107)
	at org.terasology.engine.core.TerasologyEngine.initialize(TerasologyEngine.java:227)
	at org.terasology.moduletestingenvironment.ModuleTestingEnvironment.createEngine(ModuleTestingEnvironment.java:409)
	at org.terasology.moduletestingenvironment.ModuleTestingEnvironment.createHeadlessEngine(ModuleTestingEnvironment.java:383)
	at org.terasology.moduletestingenvironment.ModuleTestingEnvironment.createHost(ModuleTestingEnvironment.java:451)
	at org.terasology.moduletestingenvironment.ModuleTestingEnvironment.setup(ModuleTestingEnvironment.java:167)
	at org.terasology.moduletestingenvironment.MTEExtension.beforeAll(MTEExtension.java:93)
```

In CI, the tests also fail but with a different error message. For instance, in [Health build #62](http://jenkins.terasology.io/teraorg/job/Terasology/job/Modules/job/H/job/Health/job/develop/62/testReport/) the tests fail with a `NoClassDefFoundError` for `PathManager`.

```
java.lang.NoClassDefFoundError: org/terasology/engine/core/paths/PathManager
	at org.terasology.moduletestingenvironment.ModuleTestingEnvironment.createEngine(ModuleTestingEnvironment.java:401)
	at org.terasology.moduletestingenvironment.ModuleTestingEnvironment.createHeadlessEngine(ModuleTestingEnvironment.java:383)
[...]

PR #4796, opened 2021-06-25
test: provide a subsystem to register the unittest module

keturn:
Many current ModuleTestingEnvironment errors (#4757) come from a failure to find a module associated with some stub Components included with engine-tests.

This subsystem registers the `unittest` module containing those classes with the ModuleManager before the engine is initialized.

It's used on the MTE side by adding it to the list of subsystems given to the builder: https://github.com/Terasology/ModuleTestingEnvironment/pull/55

## Pros

- Looks like we're on the right track. With this, MTE's own test suite goes from almost-all-failing to 19 pass, 3 failing.
- Very little code to implement and use.

## Cons

- 💄🐷. I still feel this is a kludge that isn't exactly making things simpler. At least it using an existing extension point (subsystems), but it adds a new thing to the mix that isn't usually considered in the module-loading process.
- I worry it's fragile. This makes things work because Subsystem.initialize is called at a useful place in the engine's boot sequence, but it's very possible that something else will call `ModuleManager.loadEnvironment` later—_without_ including `"unittest"` in the environment—and then things will break with an unclear error again. (At least, I think so. Might not. Should test.)
- It leaves the question of "why are we trying to register all components before loading modules?" unanswered.

## Outstanding but maybe not critical?
- Question: If we register the module _without_ entering it in the current module environment, is that sufficient to avoid the error?
- Are there other places in engine-test that should prefer this over their current use of ModuleManagerFactory?

### DIFF
--- a/engine-tests/src/main/java/org/terasology/engine/testUtil/WithUnittestModule.java
+++ b/engine-tests/src/main/java/org/terasology/engine/testUtil/WithUnittestModule.java
@@ -1,0 +1,25 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.engine.testUtil;
+
+import org.terasology.engine.context.Context;
+import org.terasology.engine.core.GameEngine;
+import org.terasology.engine.core.module.ModuleManager;
+import org.terasology.engine.core.subsystem.EngineSubsystem;
+import org.terasology.gestalt.module.Module;
+
+public class WithUnittestModule implements EngineSubsystem {
+    @Override
+    public String getName() {
+        return "Unittest";
+    }
+
+    @Override
+    public void initialise(GameEngine engine, Context rootContext) {
+        EngineSubsystem.super.initialise(engine, rootContext);
+        ModuleManager manager = rootContext.get(ModuleManager.class);
+        Module unittestModule = manager.registerPackageModule("org.terasology.unittest");
+        manager.resolveAndLoadEnvironment(unittestModule.getId());
+    }
+}
--- a/engine-tests/src/main/java/org/terasology/engine/testUtil/WithUnittestModule.java
+++ b/engine-tests/src/main/java/org/terasology/engine/testUtil/WithUnittestModule.java
@@ -9,6 +9,17 @@ import org.terasology.engine.core.module.ModuleManager;
 import org.terasology.engine.core.subsystem.EngineSubsystem;
 import org.terasology.gestalt.module.Module;
 
+/**
+ * This subsystem registers the {@code unittest} module containing stub component classes with the {@link ModuleManager} before the engine
+ * is initialized.
+ * <p>
+ * It is indended to be used on the MTE side by adding it to the list of subsystems given to the builder.
+ * <p>
+ * NOTE: The usage of this subsystem for MTE is potentially fragile as it depends on specific assumptions on when subsystems are initialized.
+ * This makes things work because {@code Subsystem.initialize} is called at a useful place in the engine's boot
+ * sequence, but it's very possible that something else will call {@code ModuleManager.loadEnvironment} later — without including "unittest"
+ * in the environment — and then things might break with an unclear error again.
+ */
 public class WithUnittestModule implements EngineSubsystem {
     @Override
     public String getName() {


engine-tests/src/main/java/org/terasology/engine/testUtil/WithUnittestModule.java
```diff
@@ -0,0 +1,25 @@
...
+
+package org.terasology.engine.testUtil;
+
+import org.terasology.engine.context.Context;
+import org.terasology.engine.core.GameEngine;
+import org.terasology.engine.core.module.ModuleManager;
+import org.terasology.engine.core.subsystem.EngineSubsystem;
+import org.terasology.gestalt.module.Module;
+
+public class WithUnittestModule implements EngineSubsystem {
```
keturn, line 12: any suggestions for the name? should it end it `Subsystem`?

engine-tests/src/main/java/org/terasology/engine/testUtil/WithUnittestModule.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public String getName() {
+        return "Unittest";
+    }
+
+    @Override
+    public void initialise(GameEngine engine, Context rootContext) {
+        EngineSubsystem.super.initialise(engine, rootContext);
```
keturn, line 20: Calling `super` out of habit here, but the default implementation is a no-op. Is that a good practice or should we remove this `super`?

keturn:
Switched to _draft_ until we get the remaining tests fixed— or find out they are failing for an unrelated reason.

keturn:
All three test failures were in [TestEventReceiverTest](https://github.com/Terasology/ModuleTestingEnvironment/blob/2bccc60d48be1ce07628f3bfa4ce9046f27e3447/src/test/java/org/terasology/moduletestingenvironment/TestEventReceiverTest.java#L26) and they all fail in `onDummyEvent`:
```java
@Share(DummySystem.class)
public class DummySystem extends BaseComponentSystem {
    @ReceiveEvent
    public void onDummyEvent(DummyEvent event, EntityRef entity) {
        DummyComponent component = entity.getComponent(DummyComponent.class);
        component.dummy = true;
        entity.saveComponent(component);
    }
}
```
with a NullPointerException suggesting that the `entity.getComponent` call returned null.

keturn:
I am [confused](https://discord.com/channels/270264625419911192/696830442065625118/858128587332255785) about those test failures, but I'm willing to say they're not necessarily within the scope of this PR.

keturn:
so far I have only checked MTE's own test suite.

Still need to check the MTE tests of other modules.

and once this is merged (with MTE#55), then we'll start seeing if that was sufficient for the errors when run under the CI environment, or if that's something else.

keturn:
I ran some module tests. There are some failures, I haven't yet tallied them to see what the failure modes are: https://scans.gradle.com/s/6vj4ag42fug5w/tests/failed

keturn:
Four modules:
- Pathfinding
- NameGenerator
- SimpleFarming
- Tasks

so it's worth noting that a lot of other modules with MTE tests (like Health's RegenTest) passed!

Tasks is the smallest of those four, so that's the thing I'm looking at next.

keturn:
`Tasks` was pulling in an old version of MTE. I fixed that.

The others are using various non-MTE tests and so we can call them out of scope for this PR.

engine-tests/src/main/java/org/terasology/engine/testUtil/WithUnittestModule.java
skaldarnar: A survey of what we currently (including this PR) have shows:

- Usage in extends/implements clause: 22
- 18 with `Subsystem` suffix / 5 without suffix
- classes without suffix are
	- WithUnittestModule
	- OpenVRInput 
	- Headless[Audio|Graphics|Input]
	
Given that, I'd suggest to
- add the `Subsystem` suffix here. Would `RegisterUnittestModuleSubsystem` be a reasonable name?
- (different PR) rename `OpenVRInput` >>> `OpenVrInputSubsystem`
- (different PR) add `Subsystem` suffix to `Headless[Audio|Graphics|Input]`
- (different PR) make sure to consistently use `Subsystem` (we use `SubSystem` in some places)

The latter changes would fit the topic of code style improvements @jdrueckert :wink:

engine-tests/src/main/java/org/terasology/engine/testUtil/WithUnittestModule.java
skaldarnar: I think I would not make a call to `super` when overriding _default_ methods of an interface usually :thinking:
I understand the `EngineSubsystem` interface as something where I get a bunch of hook methods I can choose to implement, but I don't have to. 

If it suddenly becomes necessary to always perform some kind of code here you'd probably change the underlying implementation such that contract stays the same for child classes (e.g., introduce a new "internal init method" that in turn calls the hook method...

tl;dr: I don't think it is necessary, and I would omit it.

engine-tests/src/main/java/org/terasology/engine/testUtil/WithUnittestModule.java
skaldarnar: I'd appreciate a short docstring comment for this subsystem explaining why we have it, and what it is used for.

skaldarnar (approved):

skaldarnar (approved):

PR #4797, opened 2021-06-25
fix: make BlockArea.iterator() use the same order as noise and facet data

tolziplohu:
### Contains

Many facet providers use raw arrays of facet data and precalculated noise, which is more efficient. To get the position as well as the index, it would make sense to use `BlockArea.iterator()` and call `next()` on each loop. The problem is, facet and noise data increase the x coordinate first (`(0, 0), (1,  0), (2, 0)...`), but `BlockArea.iterator()` increases the y coordinate first (`(0, 1), (0, 1), (0, 2)...`). There's no good reason for that that I could find, so this PR switches it to the more intuitive x-first order that everything else uses, making it possible to both get the block positions and use subsampled noise efficiently in a facet provider.

Was there any reason for the y-first order here?


engine/src/main/java/org/terasology/engine/world/block/BlockArea.java
```diff
@@ -116,6 +116,7 @@ public Vector2ic next() {
 
                 if (current.equals(next)) {
                     if (findNext()) {
+                        current.set(next);
```
pollend, line 119: can you add a test for this case. I think this doesn't cover the case if you call next() twice in a row without findNext(). I think I made the same mistake in BlockRegion.

pollend (changes requested):

skaldarnar:
As discussed in the mentor meeting on 2021-06-26 @tolziplohu will polish this PR a bit by 
- adding the requested test
- adding documentation to explicitly state the iterator contract (order of elements)

tolziplohu:
I added the test and the documentation.

@pollend:
> I think this doesn't cover the case if you call next() twice in a row without findNext(). I think I made the same mistake in BlockRegion.

Yeah, I added that line because I noticed that it wasn't working without calling `hasNext()` in between. It does look like BlockRegion has the same problem - should I fix that and add a test for it in this PR, or should that be separate?

pollend (approved):

Issue #4798, opened 2021-06-26
refactor usage of concurrency with ReactiveX


pollend:
some situations the number of threads are overprovisioned and in some cases its under provisioned. Some workloads might benefit with more threads but selective changing the number of threads will be over provisioned on some systems and under provisioned on another system. 

### Current situation

- Chunk-Unloader (allocate 4 threads)
  - int list to queue event for BeforeDeactivateBlocks
  - very light load so threads are mostly idle
  -at any moment a single thread will become active
  
- Chunk-Updater (allocates 8 threads)
 - generates chunk mesh and queue them back on the main thread

- LOD Chunk Generation
 - only triggered for handling different levels of lod

- ThreadManagerSubsystem (allocates 16 threads)
 - loadingGame
 - create new world
 - saving screenshot

- TaskMaster
  - Behaviors (1 thread)
  - FlexiablePathFinding (4 threads)
  - Pathfinding (1 thread)

-----------------------------------

- [ ] drop usage of the ThreadManagerSubsystem easily can be replaced by usage of Observable
- [ ] remove usage of TaskMaster
  - [ ] replace usage of threading in Behaviours
  - [ ] replace usage of threading in FlexiablePathFinding
  - [ ] replace usage of threading in pathfinding
  - [ ] LOD Chunk generation
  - [ ] chunk unloader probably does not benefit with parallelism
- [ ] replace usage of threading in FlexiablePathFinding

keturn:
Do we have anyone familiar enough with the various implementations of the reactive pattern in Java to evaluate [Project Reactor](https://projectreactor.io/) (with `Flux` and `Mono` APIs) in comparison to RxJava2?

skaldarnar:
I tried to sketch out what we're currently doing and using (no guarantee for completeness):
![image](https://user-images.githubusercontent.com/1448874/123555764-0a1b0880-d788-11eb-914e-4b234cfde699.png)

pollend:
@skaldarnar that's right got me confused because ExecutorService just creates a pool of threads and each instance of TaskMaster will introduce its own set of threads. we might get into a situation where the processor is constantly context switching between a lot of computational threads.

pollend:
> Do we have anyone familiar enough with the various implementations of the reactive pattern in Java to evaluate [Project Reactor](https://projectreactor.io/) (with `Flux` and `Mono` APIs) in comparison to RxJava2?

[...]

PR #4799, opened 2021-06-27
refactor: replace usage of ThreadManagerSubsystem with reactor Scheduler

pollend:
This replaces the ThreadManagerSubsystem with Scheduler provided by rxjava. pretty straightforward  since this subsystem is only ever used in the core engine and not to an extent that would make sense. the threads allocated are mostly unused and sit idle. 

toTest:
- verify that game previews are saved
- verify that a screen shot could be taken with F12
- game startup is dependent on some of this threading. 

Contributes to https://github.com/MovingBlocks/Terasology/issues/4798


engine/src/main/java/org/terasology/engine/core/GameThread.java
```diff
@@ -23,9 +29,22 @@
...
+    }
+
+    public static Scheduler newThread() {
+        return AccessController.doPrivileged((PrivilegedAction<Scheduler>) Schedulers::newThread);
+    }
+
+    public static Scheduler computation() {
+        return AccessController.doPrivileged((PrivilegedAction<Scheduler>) Schedulers::computation);
+    }
+
```
pollend, line 47: we need to have these in a privileged scope.

engine/src/main/java/org/terasology/engine/core/GameThread.java
```diff
@@ -18,14 +23,33 @@
  * </ul>
  *
  */
+@API
```
skaldarnar, line 26: Why did the `GameThread` become public API?

engine/src/main/java/org/terasology/engine/core/GameThread.java
skaldarnar: Why do they need to be in privileged scope? Let's document this here for the future.

engine/src/main/java/org/terasology/engine/core/module/ExternalApiWhitelist.java
```diff
@@ -109,6 +109,21 @@
...
+            .add("io.reactivex.rxjava3.functions")
+            .add("io.reactivex.rxjava3.internal")
+            .add("io.reactivex.rxjava3.observables")
+            .add("io.reactivex.rxjava3.observers")
+            .add("io.reactivex.rxjava3.parallel")
+            .add("io.reactivex.rxjava3.plugins")
+            .add("io.reactivex.rxjava3.processors")
+            .add("io.reactivex.rxjava3.schedulers")
+            .add("io.reactivex.rxjava3.subjects")
+            .add("io.reactivex.rxjava3.subscribers")
```
skaldarnar, line 126: How did you decide what to add here, and what is not added (assuming that there must be something we leave out, or otherwise we could have used a single entry `io.reactivex.rxjava3`).

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -153,16 +154,16 @@ public static void main(String[] args) {
                 engine.run(new StateHeadlessSetup());
             } else if (loadLastGame) {
                 engine.initialize(); //initialize the managers first
-                engine.getFromEngineContext(ThreadManager.class).submitTask("loadGame", () -> {
```
skaldarnar, line 156: It looks like with `ThreadManager` we could give the tasks a name such that they are identifiable later, e.g., in logs. Is there a similar concept in RxJava?

skaldarnar (commented):
Please also have a look at the alert reported by LGTM

engine/src/main/java/org/terasology/engine/core/GameThread.java
pollend: do we need a separate API?

engine/src/main/java/org/terasology/engine/core/GameThread.java
pollend: a lot of thread actions are guarded by the sandbox.

engine/src/main/java/org/terasology/engine/core/GameSchedulers.java
```diff
@@ -0,0 +1,7 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.engine.core;
+
+public class GameSchedulers {
```
tolziplohu, line 6: Why does this class exist?

engine/src/main/java/org/terasology/engine/rendering/opengl/ScreenGrabber.java
```diff
@@ -101,15 +101,13 @@ public void saveScreenshot() {
         int width = sceneFinalFbo.width();
         int height = sceneFinalFbo.height();
 
-        Runnable task;
         if (savingGamePreview) {
-            task = () -> saveGamePreviewTask(buffer, width, height);
+            GameScheduler.boundedElastic().schedule(() -> saveScreenshotTask(buffer, width, height));
```
tolziplohu, line 105: It looks like you accidentally changed this to `saveScreenshotTask`, which is why game previews aren't saving.
```suggestion
            GameScheduler.boundedElastic().schedule(() -> saveGamePreviewTask(buffer, width, height));
```

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -153,16 +154,18 @@ public static void main(String[] args) {
...
+
+                Mono.fromRunnable(() -> {
                     GameManifest gameManifest = getLatestGameManifest();
                     if (gameManifest != null) {
                         engine.changeState(new StateLoading(gameManifest, NetworkMode.NONE));
                     }
-                });
+                }).subscribeOn(GameScheduler.boundedElastic()).subscribe();
```
tolziplohu, line 163: I'm curious why this isn't just `GameScheduler.boundedElastic().schedule(() -> { ... })` - aren't they equivalent?

facades/PC/src/main/java/org/terasology/engine/Terasology.java
```diff
@@ -173,7 +176,7 @@ public static void main(String[] args) {
                             }
                             engine.changeState(new StateLoading(gameManifest, NetworkMode.NONE));
                         }
-                    });
+                    }).subscribeOn(GameScheduler.boundedElastic()).subscribe();
```
tolziplohu, line 179: same as above

tolziplohu (dismissed):
This looks good to me, it's good Reactor infrastructure to have.

>   verify that game previews are saved

This is the only thing that appears to not be working for me - it looks like you accidentally made it save a screenshot instead of a preview.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
pollend: this should let us do some tractability if we enable it.

engine/src/main/java/org/terasology/engine/core/module/ExternalApiWhitelist.java
keturn: This particular diff has been changed to replace the rxjava packages with reactor packages, but the same question still applies.

keturn (commented):
The thing we lose by skipping TaskMaster is the integration with our ThreadMonitor: https://github.com/MovingBlocks/Terasology/blob/5f84ee0198c43047d911b71ada8331309ce4a014/engine/src/main/java/org/terasology/engine/utilities/concurrency/TaskProcessor.java#L34

engine/src/main/java/org/terasology/engine/core/GameScheduler.java
```diff
@@ -0,0 +1,39 @@
...
+    static {
+        MAIN = Schedulers.fromExecutor(runnable -> GameThread.asynch(runnable));
+    }
+
+    public static Scheduler gameMain() {
+        return  MAIN;
+    }
+
+    public static Scheduler boundedElastic() {
+        return AccessController.doPrivileged((PrivilegedAction<Scheduler>) Schedulers::boundedElastic);
```
keturn, line 28: A key point for my understanding is that these all return _singletons._ They are _not_ factory methods that return a new Scheduler each time. Makes more sense now.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
keturn: How does that work?

facades/PC/src/main/java/org/terasology/engine/Terasology.java
keturn: It looks like there is a subscriber [Context](https://projectreactor.io/docs/core/release/reference/#context) that is useful for things like this.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
pollend: https://projectreactor.io/docs/core/release/reference/#_tags

pollend:
image

engine/src/main/java/org/terasology/engine/rendering/nui/layers/ingame/metrics/RunningThreadsMode.java
```diff
@@ -35,4 +52,9 @@ public boolean isAvailable() {
     public boolean isPerformanceManagerMode() {
         return false;
     }
+
+    @Override
+    public void close() throws Exception {
+
+    }
```
keturn, line 59: Did you intend to put something here? I see you added an `implements AutoCloseable` to the class.

facades/PC/src/main/java/org/terasology/engine/Terasology.java
keturn: I don't get it. That link talks about how we can add extra tags, but I don't see any of that here?

engine/src/main/java/org/terasology/engine/rendering/nui/layers/ingame/metrics/RunningThreadsMode.java
pollend: opps was intending to have a temporary registry.

engine/src/main/java/org/terasology/engine/rendering/nui/layers/ingame/metrics/RunningThreadsMode.java
```diff
@@ -3,25 +3,42 @@
...
-            builder.append(threads.getName());
-            builder.append(" - ");
-            builder.append(threads.getLastTask());
-            builder.append("\n");
+
+        Search.in(DisplayMetricsMonitor.metricRegistry)
+                .tag("monitor", "display-metric")
+                .timers().forEach(k -> {
+
+            HistogramSnapshot snapshot = k.takeSnapshot();
```
keturn, line 29: I'm trying to figure out how this compares to the previous implementation.

The old one was a list of all _currently active tasks._

If I'm reading this correctly, the new one is a list of timers that had something recorded in the previous reporting period (DisplayMetricsMonitor.captureDuration), with some stats about throughput and latency.

Those are also useful things to know, but they're not the same.

Is that a problem in practice?

- is there more lag this way? i.e. if it only updates once every captureDuration.
  + is it enough to matter? does anyone _want_ updates more than once per second?
- how often are there tasks with duration greater than captureDuration? If this happens, I'm concerned it gives a less accurate notion of tasks that are still processing.

keturn: Also: This is searching the DisplayMetricsMonitor registry.

but the ThreadMonitor timers are registered in GlobalRegistry.

Is this search going to find those timers?

keturn (commented):
Learning lots of things about micrometer! This looks promising.

There's one part where I'm not connecting the dots.

- `saveScreenshot` used ThreadManager. ThreadManager used ThreadMonitor.
- This PR replaced ThreadManager with Reactor schedulers.
  + keturn complained that meant `saveScreenshot` was no longer reporting to ThreadMonitor.
- [`27e8b0e` (#4799)](https://github.com/MovingBlocks/Terasology/pull/4799/commits/27e8b0ec43284bc01d5917a8b80a86136872b8c0) made changes to ThreadMonitor (in SingleThreadMonitorImpl) so that ThreadMonitor reports to micrometer.
- and changed RunningThreadsMode to report only from micrometer, not directly from ThreadMonitor.
- but `saveScreenshot` (and anything else that formerly used ThreadManager) still doesn't use ThreadMonitor,
  + and I don't see any other MetricsMode for the DebugOverlay that reports on tasks that lack ThreadMonitor integration?

engine/src/main/java/org/terasology/engine/rendering/nui/layers/ingame/metrics/RunningThreadsMode.java
pollend: metric registry is where the data is supposed to get sinked to. its an empty registry since we are not pushing these metrics anywhere.  DismplayMetricMonitor metric registry is a blank registry with a 1 second interval for the sampled data. 

- tells you how long it takes to process an task/ the number of entries that are getting processed 
- The capture is driven by when the entry was recorded i.e the task ending. 

I would probably use JMX if you need a better picture. the snapshots have a lot more data then what I can present within the scope of this ticket.

pollend:
> Learning lots of things about micrometer! This looks promising.
> 
> There's one part where I'm not connecting the dots.
> 
>     * `saveScreenshot` used ThreadManager. ThreadManager used ThreadMonitor.
> 
>     * This PR replaced ThreadManager with Reactor schedulers.
>       
>       * keturn complained that meant `saveScreenshot` was no longer reporting to ThreadMonitor.
> 
>     * [`27e8b0e` (#4799)](https://github.com/MovingBlocks/Terasology/pull/4799/commits/27e8b0ec43284bc01d5917a8b80a86136872b8c0) made changes to ThreadMonitor (in SingleThreadMonitorImpl) so that ThreadMonitor reports to micrometer.
> 
>     * and changed RunningThreadsMode to report only from micrometer, not directly from ThreadMonitor.
> 
>     * but `saveScreenshot` (and anything else that formerly used ThreadManager) still doesn't use ThreadMonitor,
>       
>       * and I don't see any other MetricsMode for the DebugOverlay that reports on tasks that lack ThreadMonitor integration?

I was testing this was Pathfinding. just need to add a name and a tag with metrics(). probably want to do something custom with the list :https://projectreactor.io/docs/core/release/reference/#metrics

image

keturn:
# Reporting on In-Progress Tasks

The use case we're looking to address is: How do I know if there is a long-running (or indefinitely stuck) task currently being processed?

A task that has not finished will not be represented in a snapshot of a Timer.

## Possible Approaches

### ExecutorService metrics

ExecutorServices (such as those backing Reactor's schedulers) are [instrumented with meters](https://micrometer.io/docs/ref/jvm). They include `executor.active` and `executor.queued` gauges.

These should be sufficient to let us know if there _are_ active tasks, though it is not enough to distinguish short tasks from long ones, nor will it let us see _which_ tasks they are.

### Long Task Timers

If we have certain operations that are known to take a long time, we can wrap them in a [**long task timer**](https://micrometer.io/docs/concepts#_long_task_timers).

(I'm not sure we have an annotation processor that handles the `@Timed(longTask=true)` annotation, we might have to use the MeterRegistry API instead.)

### GameScheduler API

`GameScheduler` provides a central interface. We could change from returning direct Scheduler references like this

```java
GameScheduler.boundedElastic().schedule(() -> saveScreenshotTask(buffer, width, height));
```

to providing a `schedule` method more like the old `ThreadManager.submitTask`:

```java
GameScheduler.schedule("Write Screenshot", () -> saveScreenshotTask(buffer, width, height));
```

which would give us a chance to wrap the Runnable in something like our ThreadMonitor#startThreadActivity.

engine/src/main/java/org/terasology/engine/rendering/nui/layers/ingame/metrics/RunningThreadsMode.java
keturn: Posted some implementation ideas related to "currently active tasks" in https://github.com/MovingBlocks/Terasology/pull/4799#issuecomment-894694532

pollend:
thats why those tasks are wrapped in a Mono, does the same thing. 
> # Reporting on In-Progress Tasks
> 
> The use case we're looking to address is: How do I know if there is a long-running (or indefinitely stuck) task currently being processed?
> 
> A task that has not finished will not be represented in a snapshot of a Timer.
> ## Possible Approaches
> ### ExecutorService metrics
> 
> ExecutorServices (such as those backing Reactor's schedulers) are [instrumented with meters](https://micrometer.io/docs/ref/jvm). They include `executor.active` and `executor.queued` gauges.
> 
> These should be sufficient to let us know if there _are_ active tasks, though it is not enough to distinguish short tasks from long ones, nor will it let us see _which_ tasks they are.
> ### Long Task Timers
> 
> If we have certain operations that are known to take a long time, we can wrap them in a [**long task timer**](https://micrometer.io/docs/concepts#_long_task_timers).
> 
> (I'm not sure we have an annotation processor that handles the `@Timed(longTask=true)` annotation, we might have to use the MeterRegistry API instead.)
> ### GameScheduler API
> 
> `GameScheduler` provides a central interface. We could change from returning direct Scheduler references like this
> 
> ```java
> GameScheduler.boundedElastic().schedule(() -> saveScreenshotTask(buffer, width, height));
> ```
> 
> to providing a `schedule` method more like the old `ThreadManager.submitTask`:
> 
> ```java
> GameScheduler.schedule("Write Screenshot", () -> saveScreenshotTask(buffer, width, height));
> ```
> 
> which would give us a chance to wrap the Runnable in something like our ThreadMonitor#startThreadActivity.

keturn:
# Security

## GameScheduler

- `GameScheduler` is marked `@API`. `gameMain()` provides access to scheduling on `GameThread`, which is not `@API`.
- What is requiring the `AccessController.doPrivileged` to get the schedulers?
  + Is engine code requiring it, or only modules?
  + Is it for something that reactor…Schedulers does only on the first call (e.g. because it makes a new thread pool), or always?
  + If it is for new threads, does `immediate` need it? Immediate doesn't involve new threads.

## ExternalApiWhitelist

- Packages
  + `reactor.core.scheduler` — probs don't want modules having direct access to `Schedulers`; replace this PACKAGES entry with a few CLASSES?

keturn:
# Document

## GameScheduler
- Link to [Reactor Reference Guide on Schedulers](https://projectreactor.io/docs/core/3.4.8/reference/#schedulers) for general info on what "boundedElastic" is and how it differs from "parallel."
- When programming for Terasology, when should we use each?
  + Is the answer always "parallel"?

pollend:
a boundedElastic scheduler will create threads to fit the needs of the tasks getting submitted to the scheduler. if you can keep all threads busy with 3 threads then elastic scheduler will use three and will split the load. it will expand the number of threads to match the rate of submitted tasks I believe. parallel scheduler is capped to the number of threads the processor supports. difference between IO and CPU bounds tasks.

pollend:
The way the thread monitor works is not how I want to fit this into this micrometer scheme. I would ideally like to build a facade over micrometer. so we can just stub it out for release candidates.

engine/src/main/java/org/terasology/engine/core/GameScheduler.java
```diff
@@ -0,0 +1,53 @@
...
+     * <p>
+     * You can expect this to always return the <em>same</em> scheduler; it does not create a new scheduler instance or thread
+     * on every call.
+     *
+     * @return (singleton)
+     * @see <a href="https://projectreactor.io/docs/core/release/reference/#core-features">Reactor Core Features</a>
+     */
+    public static Scheduler getScheduler() {
+        return AccessController.doPrivileged((PrivilegedAction<Scheduler>) Schedulers::parallel);
+    }
```
pollend, line 52: can we change this to `parallel()` getScheduler is too generic. keep it the same as the reactor API.

keturn: Done. :+1:

engine/src/main/java/org/terasology/engine/monitoring/DisplayMetricsMonitor.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
+public class DisplayMetricsMonitor extends StepMeterRegistry {
+    public static final DisplayMetricsMonitor metricRegistry = new DisplayMetricsMonitor();
+    public static final Duration captureDuration = Duration.ofSeconds(1);
+
+    public DisplayMetricsMonitor() {
+        super(new LoggingRegistryConfig() {
```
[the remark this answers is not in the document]
keturn: I assumed this is some leftover of earlier experimentation. I've replaced it with StepRegistryConfig.

engine/src/main/java/org/terasology/engine/monitoring/impl/SingleThreadMonitorImpl.java
```diff
@@ -102,12 +110,19 @@ public final synchronized void beginTask(String task) {
         if (taskCounters.adjustOrPutValue(task, 1, 1) == 1) {
             tasks.add(task);
         }
+        startTime = Clock.SYSTEM.monotonicTime();
+        this.taskTimer = Timer.builder(task)
+                .tags(Collections.singletonList(Tag.of("monitor", "display-metric")))
+                .register(Metrics.globalRegistry);
```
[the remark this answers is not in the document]
keturn: resolved by declaring micrometer to be out-of-scope for this PR (#4839)

engine/src/main/java/org/terasology/engine/rendering/nui/layers/ingame/metrics/RunningThreadsMode.java
keturn: resolved by declaring micrometer to be out-of-scope for this PR (#4839)

facades/PC/src/main/java/org/terasology/engine/Terasology.java
keturn: #4839 added a `GameScheduler.scheduleParallel` method that takes a name which it uses in the same way as ThreadManager.submitTask did.

It doesn't address future use cases where we take advantage of Flux's model for pipelines, but it suffices for one-shot Runnables like this.

pollend: should be fine for our use cases. i'm not too happy but it should be good for our use cases at this point.

keturn:
> `GameScheduler` is marked `@API`. `gameMain()` provides access to scheduling on `GameThread`, which is not `@API`.

I brought this up in today's meeting, and it wasn't clear to any of us if/when modules need access to this.

skaldarnar's proposal was to leave this implemented, but at a permission level modules wouldn't be able to actually use, with a comment along the lines of "change this if we have a documented use case for it" or something like that.

(I don't think we can implement it _exactly_ as proposed, because I don't think the `@API` system gives us per-method control, but we can make it package-private or something.)

pollend:
> > `GameScheduler` is marked `@API`. `gameMain()` provides access to scheduling on `GameThread`, which is not `@API`.
> 
> I brought this up in today's meeting, and it wasn't clear to any of us if/when modules need access to this.
> 
> skaldarnar's proposal was to leave this implemented, but at a permission level modules wouldn't be able to actually use, with a comment along the lines of "change this if we have a documented use case for it" or something like that.
> 
> (I don't think we can implement it _exactly_ as proposed, because I don't think the `@API` system gives us per-method control, but we can make it package-private or something.)

how do you propose letting a module schedule something back on the main thread?

DarkWeird:
Try to take screenshot:

```
17:48:07.881 [main] ERROR o.t.engine.core.TerasologyEngine - Uncaught exception, attempting clean game shutdown
java.lang.ExceptionInInitializerError: null
	at reactor.core.scheduler.Schedulers.newParallel(Schedulers.java:545)
	at reactor.core.scheduler.Schedulers.lambda$static$4(Schedulers.java:1135)
	at reactor.core.scheduler.Schedulers.cache(Schedulers.java:1168)
	at reactor.core.scheduler.Schedulers.parallel(Schedulers.java:238)
	at org.terasology.engine.core.GameScheduler.scheduleParallel(GameScheduler.java:69)
	at org.terasology.engine.rendering.opengl.ScreenGrabber.saveScreenshot(ScreenGrabber.java:112)
	at org.terasology.corerendering.rendering.dag.nodes.FinalPostProcessingNode.process(FinalPostProcessingNode.java:173)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at org.terasology.engine.rendering.world.WorldRendererImpl.render(WorldRendererImpl.java:350)
	at org.terasology.engine.core.modes.StateIngame.render(StateIngame.java:214)
	at org.terasology.engine.core.subsystem.lwjgl.LwjglGraphics.postUpdate(LwjglGraphics.java:76)
	at org.terasology.engine.core.TerasologyEngine.tick(TerasologyEngine.java:509)
	at org.terasology.engine.core.TerasologyEngine.mainLoop(TerasologyEngine.java:460)
	at org.terasology.engine.core.TerasologyEngine.runMain(TerasologyEngine.java:436)
	at org.terasology.engine.core.TerasologyEngine.run(TerasologyEngine.java:402)
	at org.terasology.engine.Terasology.call(Terasology.java:183)
	at org.terasology.engine.Terasology.call(Terasology.java:67)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1933)
	at picocli.CommandLine.access$1200(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2291)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2159)
	at picocli.CommandLine.execute(CommandLine.java:2058)
	at org.terasology.engine.Terasology.main(Terasology.java:115)
Caused by: java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "modifyThread")
	at java.base/java.security.AccessControlContext.checkPermission(AccessControlContext.java:472)
	at java.base/java.security.AccessController.checkPermission(AccessController.java:897)
	at java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:322)
	at org.terasology.gestalt.module.sandbox.ModuleSecurityManager.checkPermission(ModuleSecurityManager.java:48)
	at java.base/java.util.concurrent.ThreadPoolExecutor.checkShutdownAccess(ThreadPoolExecutor.java:748)
	at java.base/java.util.concurrent.ThreadPoolExecutor.shutdownNow(ThreadPoolExecutor.java:1405)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor.shutdownNow(ScheduledThreadPoolExecutor.java:870)
	at java.base/java.util.concurrent.Executors$DelegatedExecutorService.shutdownNow(Executors.java:693)
	at reactor.core.scheduler.ParallelScheduler.<clinit>(ParallelScheduler.java:57)
	... 25 common frames omitted
```

Need to initialize schedulers eager. Or add warp this place with `doPrivilegeAction`
This is first call to scheduler and it fail because called from CoreRendering module

keturn:
> > `GameScheduler` is marked `@API`. `gameMain()` provides access to scheduling on `GameThread`, which is not `@API`.
> 
> I brought this up in today's meeting, and it wasn't clear to any of us if/when modules need access to this.

[Some discussion](https://discord.com/channels/270264625419911192/876864739945365524/876867401931685930) convinced me we do have a use case that requires this:

We want to allow modules to make OpenGL calls, so we must give them a way to run things on the thread that owns the OpenGL Context. (OpenGL works by pushing things on to a command queue, and you don't want to be mixing that up by doing that from multiple threads.)

While we don't _currently_ have module code that makes use of GameThread for OpenGL this way, it's not much of a stretch for me to see that we might. Especially as we use Reactor to move more work _off_ the main thread.

DarkWeird:
@keturn 
OpenGL thread at my way currently.

CoreRendering is using opengl from modules.
It runs at main thread currently.

keturn:
> Try to take screenshot:

Oh, this is my fault! When I wrote the method that takes the task name, I didn't run it through the other method with the AccessController wrapping.

I'll try the eager initialization.

DarkWeird:
> > Try to take screenshot:
> 
> Oh, this is my fault! When I wrote the method that takes the task name, I didn't run it through the other method with the AccessController wrapping.
> 
> I'll try the eager initialization.

Eager works nice.
In graphics scheduler PR - parallel using for interval flux. For splashscreen render loop.

keturn:
>  If you make changes to the suppression of alerts in a pull request, please note that theses changes will only be effective once that pull request is merged. 

eh? that's weird, LGTM-bot.

keturn (approved):
Okay, I think the permission-modifying code is isolated enough and adequately commented now.

As for the matter of why we never had to add PermissionProperties before, I think it's a combination:
1. The old ThreadManager seems to have been running code outside the security context, so it's possible things were sneaking by, but mostly
2. There aren't that many things called from the module sandbox that require reading system properties.

I'm totally stumped on why wildcards don't work for PermissionProperties, but I am not worried enough about the maintenance cost of that list to put more importance on it for the time being.

engine/src/main/java/org/terasology/engine/core/GameThread.java
keturn: see https://github.com/MovingBlocks/Terasology/pull/4799#issuecomment-904231869

pollend:
> Okay, I think the permission-modifying code is isolated enough and adequately commented now.
> 
> As for the matter of why we never had to add PermissionProperties before, I think it's a combination:
> 
>     1. The old ThreadManager seems to have been running code outside the security context, so it's possible things were sneaking by, but mostly
> 
>     2. There aren't that many things called from the module sandbox that require reading system properties.
> 
> 
> I'm totally stumped on why wildcards don't work for PermissionProperties, but I am not worried enough about the maintenance cost of that list to put more importance on it for the time being.

I already tried wildcards :P

Issue #4801, opened 2021-06-28
ExceptionInInitializerError (NoClassDefFoundError) during Advanced Game Setup (AGS)


KdotJPG:
### What you were trying to do

Create a custom world/universe type to further test some changes.

### What actually happened

Crashes when I click Configure or Continue to Pregeneration

### How to reproduce

* Choose 'Single Player' on the main screen
* Click the 'Create' button
* Click 'Advanced'
* Try any configuration of modules (I've tried with all disabled and all enabled except for Module Testing Environment)
* Click 'Continue to Universe Setup'
* Choose a world generator and click 'Add', 
* Click 'Configure'
  * Alternatively, click 'Continue to Pregeneration'

![image](https://user-images.githubusercontent.com/8829856/123589779-b6281800-d7b7-11eb-8ac5-dab401e1b43a.png)

### Game version

ALPHA-20 from the `develop` branch (latest commit was [#fbeed19](https://github.com/MovingBlocks/Terasology/commit/fbeed19316403e76a88bb2da85f89f0bad2e1e48)) and run with `gradlew game`

### Log details

```
02:22:03.165 [main] WARN  o.t.engine.registry.InjectionHelper - CameraTargetSystem wanted Physics injected but CoreRegistry has none.
02:22:03.169 [main] WARN  o.t.engine.registry.InjectionHelper - CameraTargetSystem wanted BlockEntityRegistry injected but CoreRegistry has none.
02:22:03.169 [main] WARN  o.t.engine.registry.InjectionHelper - CameraTargetSystem wanted WorldRenderer injected but CoreRegistry has none.
02:22:03.341 [main] WARN  o.t.engine.registry.InjectionHelper - CoreCommands wanted BlockManager injected but CoreRegistry has none.
02:22:03.354 [main] WARN  o.t.engine.registry.InjectionHelper - LocalPlayerSystem wanted WorldProvider injected but CoreRegistry has none.
02:22:03.354 [main] WARN  o.t.engine.registry.InjectionHelper - LocalPlayerSystem wanted DelayManager injected but CoreRegistry has none.
02:22:03.354 [main] WARN  o.t.engine.registry.InjectionHelper - LocalPlayerSystem wanted PhysicsEngine injected but CoreRegistry has none.
02:22:03.355 [main] WARN  o.t.engine.registry.InjectionHelper - CameraTargetSystem wanted Physics injected but CoreRegistry has none.
[...]

PR #4802, opened 2021-06-28
feat(noise): Add domain rotation to mitigate Perlin squareness

KdotJPG:
### Contains

An improvement to the existing Perlin noise algorithm.

Perlin noise, since deprecated in this project in favor of Simplex, produces visually significant square bias as a flaw of the algorithm's underlying design. However, the noise is still used in the project in some places. Its visual results can be greatly improved for plane-based voxel world generation, using a simple coordinate-space rotation on the 3D noise that rotates Y to point up the grid's main diagonal (<1, 1, 1>).

Without domain rotation:
image

With domain rotation:
image

### How to test

You can generate the above images by temporarily dropping the following `RenderPerlin.java` into the same directory as `PerlinNoise.java` and running its main method through IntelliJ. The first image was generated without this PR's change, and the second was with it.

```java
package org.terasology.engine.utilities.procedural;

import java.awt.image.BufferedImage;
import javax.swing.*;
import java.awt.Color;

public class RenderPerlin {
    private static final int WIDTH = 512;
    private static final int HEIGHT = 512;
    private static final float FREQ = 1.0f / 32.0f;
    private static final boolean RIDGED = true;

    public static void main(String[] args) {

        PerlinNoise noise = new PerlinNoise(0L);

        // Image
        BufferedImage image = new BufferedImage(WIDTH, HEIGHT, BufferedImage.TYPE_INT_RGB);
        for (int y = 0; y < WIDTH; y++) {
            for (int x = 0; x < HEIGHT; x++) {
                double n = noise.noise(x * FREQ, 0, y * FREQ);
                n = RIDGED ? 1 - Math.abs(n) : n * 0.5 + 0.5;
                if (n < 0) n = 0;
                else if (n > 1) n = 1;
                int v = (int)(n * 255);
                Color color = new Color(v, v, v);
                image.setRGB(x, y, color.getRGB());
            }
        }

        // Display it
        JFrame frame = new JFrame();
        JLabel imageLabel = new JLabel();
        imageLabel.setIcon(new ImageIcon(image));
        frame.add(imageLabel);
        frame.pack();
        frame.setResizable(false);
        frame.setDefaultCloseOperation(WindowConstants.EXIT_ON_CLOSE);
        frame.setVisible(true);
    }
}
```

It is worth noting that I experience a crash (issue #4801) while trying to configure and/or add a custom world type, in this example the Perlin option. The code I provided verifies the modified noise generation itself is sound, but you may wish to perform further testing before merging this PR.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/utilities/procedural/PerlinNoise.java
+++ b/engine/src/main/java/org/terasology/engine/utilities/procedural/PerlinNoise.java
@@ -72,13 +72,22 @@ public class PerlinNoise extends AbstractNoise implements Noise2D, Noise3D {
      */
     @Override
     public float noise(float posX, float posY, float posZ) {
-        int xInt = Math.floorMod(TeraMath.floorToInt(posX), permCount);
-        int yInt = Math.floorMod(TeraMath.floorToInt(posY), permCount);
-        int zInt = Math.floorMod(TeraMath.floorToInt(posZ), permCount);
 
-        float x = posX - TeraMath.fastFloor(posX);
-        float y = posY - TeraMath.fastFloor(posY);
-        float z = posZ - TeraMath.fastFloor(posZ);
+        // Domain rotation removes Perlin's characteristic square artifacts from the XZ planes, by pointing Y up the grid's main diagonal.
+        float xz = posX + posZ;
+        float s2 = xz * -0.211324865405187f;
+        float yy = posY * 0.577350269189626f;
+        float rPosX = posX + (s2 + yy);
+        float rPosY = xz * -0.577350269189626f + yy;
+        float rPosZ = posZ + (s2 + yy);
+
+        int xInt = Math.floorMod(TeraMath.floorToInt(rPosX), permCount);
+        int yInt = Math.floorMod(TeraMath.floorToInt(rPosY), permCount);
+        int zInt = Math.floorMod(TeraMath.floorToInt(rPosZ), permCount);
+
+        float x = rPosX - TeraMath.fastFloor(rPosX);
+        float y = rPosY - TeraMath.fastFloor(rPosY);
+        float z = rPosZ - TeraMath.fastFloor(rPosZ);
 
         float u = TeraMath.fadePerlin(x);
         float v = TeraMath.fadePerlin(y);
--- a/engine/src/main/java/org/terasology/engine/utilities/procedural/PerlinNoise.java
+++ b/engine/src/main/java/org/terasology/engine/utilities/procedural/PerlinNoise.java
@@ -6,12 +6,22 @@ import org.terasology.engine.utilities.random.FastRandom;
 import org.terasology.math.TeraMath;
 
 /**
- * Improved Perlin noise based on the reference implementation by Ken Perlin.
- * @deprecated Prefer using {@link SimplexNoise}, it is comparable to Perlin noise
- * (fewer directional artifacts, lower computational overhead for higher dimensions).
+ * Domain-rotated Perlin noise. Based on reference implementation by Ken Perlin, with domain rotation by K.jpg
  *
+ * Perlin noise is an older form of noise designed without isotropy in mind. It produces significant square-aligned
+ * directional artifacts, when evaluated on planes aligned to its internal coordinate grid. However, when used for
+ * 2D-focused use-cases of 3D such as flat-world voxel terrain with overhangs, its results can be greatly improved
+ * by using a rotated coordinate space. Here in this implementation, Y is rotated to point up the main diagonal of
+ * the noise grid, while X and Z span the planes perpendicular to that. Square bias is effectively hidden, and
+ * visible directional bias is greatly reduced.
+ *
+ * It is worthwhile to note that this technique results in different input coordinates being differently useful
+ * for different purposes. When using the noise in 2D-based 3D voxel world, Y should be the vertical direction.
+ * For a 2D animation, Y should be the time variable. To generate 2D-only noise, use X and Z only, with Y fixed.
+ * This enables best utilization of the orientation of the noise grid.
+ *
+ * Note that this technique requires the 3D noise to produce nice 2D results.
  */
-@Deprecated
 public class PerlinNoise extends AbstractNoise implements Noise2D, Noise3D {
 
     private final int[] noisePermutations;
@@ -28,16 +38,14 @@ public class PerlinNoise extends AbstractNoise implements Noise2D, Noise3D {
 
     /**
      * Init. a new generator with a given seed value and grid dimension.
-     * Supports tileable noise generation
      *
      * @param seed The seed value
-     * @param gridDim gridDim x gridDim will be the size of the perlin's grid of vectors,
-     * noise will be tiled if an input coordinate crosses a multiple of gridDim
+     * @param permCount The size of the permutation table.
      */
-    public PerlinNoise(long seed, int gridDim) {
+    public PerlinNoise(long seed, int permCount) {
         FastRandom rand = new FastRandom(seed);
 
-        permCount = gridDim;
+        this.permCount = permCount;
         noisePermutations = new int[permCount * 2];
         int[] noiseTable = new int[permCount];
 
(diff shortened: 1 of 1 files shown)


pollend:
@KdotJPG hmm, wonder why its deprecated. nothing inherently wrong with it though :?

engine/src/main/java/org/terasology/engine/utilities/procedural/PerlinNoise.java
```diff
@@ -72,13 +72,22 @@ public PerlinNoise(long seed, int gridDim) {
...
-        int xInt = Math.floorMod(TeraMath.floorToInt(posX), permCount);
-        int yInt = Math.floorMod(TeraMath.floorToInt(posY), permCount);
-        int zInt = Math.floorMod(TeraMath.floorToInt(posZ), permCount);
 
-        float x = posX - TeraMath.fastFloor(posX);
-        float y = posY - TeraMath.fastFloor(posY);
-        float z = posZ - TeraMath.fastFloor(posZ);
+        // Domain rotation removes Perlin's characteristic square artifacts from the XZ planes, by pointing Y up the grid's main diagonal.
+        float xz = posX + posZ;
+        float s2 = xz * -0.211324865405187f;
```
skaldarnar, line 78: where do these "magic numbers" come from?

skaldarnar (commented):
The results indeed look better :+1: 

To my understanding Simplex Noise does not have these visual artifacts (in that strength), or does it? I think we deemed Simplex Noise better in all relevant aspects, and thus tried to switch to it. Just nobody ever finished removing all the plain Perlin Noise dependecies/usages.

Giving @tolziplohu a chance to have a look at this :nerd_face:

engine/src/main/java/org/terasology/engine/utilities/procedural/PerlinNoise.java
```diff
@@ -72,13 +72,22 @@ public PerlinNoise(long seed, int gridDim) {
...
     @Override
     public float noise(float posX, float posY, float posZ) {
-        int xInt = Math.floorMod(TeraMath.floorToInt(posX), permCount);
-        int yInt = Math.floorMod(TeraMath.floorToInt(posY), permCount);
-        int zInt = Math.floorMod(TeraMath.floorToInt(posZ), permCount);
 
-        float x = posX - TeraMath.fastFloor(posX);
-        float y = posY - TeraMath.fastFloor(posY);
-        float z = posZ - TeraMath.fastFloor(posZ);
+        // Domain rotation removes Perlin's characteristic square artifacts from the XZ planes, by pointing Y up the grid's main diagonal.
```
tolziplohu, line 76: I still don't really understand what rotation is actually being performed here, and it's not clear what exactly the math below is doing either. I'd appreciate more explanation - for example, a comment here saying how much it's rotating around what axis, and comments below on things like the magic numbers explaining where you got them.

tolziplohu (commented):
It's also my impression that there's no reason to use Perlin noise anymore. It's probably better to switch existing uses of Perlin noise to Simplex than to improve the Perlin noise. If there's actually some reason for something to continue using Perlin noise, then we should un-deprecate it and this change makes sense, but I don't think that's the case.

engine/src/main/java/org/terasology/engine/utilities/procedural/PerlinNoise.java
KdotJPG: They come from the vectors of the rotated coordinate space.

Ordinarily, X can be said to move in the unit vector direction <1, 0, 0>, Y in <0, 1, 0>, and Z in <0, 0, 1>. With this rotation, moving along the input for Y now moves in the unit direction <0.577, 0.577, 0.577> in the noise's internal coordinate space. Perpendicular to that, X and Z move in the directions <0.789, -0.577, -0.211> and <-0.211, -0.577, 0.789>. These vectors form a rotation matrix. The code is a simplification of the multiplication of this rotation matrix by the input coordinate, taking advantage of the many repetitions of 0.577 and the fact that 0.789 = 1-0.211.

In the past, when using this formula, I have left these values directly in the code, because the numerical choices behind the values are deeply entwined with the rotation code itself. If it would be preferable, though, I could pull out the 0.577 into a constant named something like `ROOT3_OVER_3` and the `-0.211` into `XZ_ROTATION_CONSTANT`.

KdotJPG: In addition to the vectors being unit length and perpendicular so as to form a rotation, the choice of <0.577, 0.577, 0.577> is because it points up the main diagonal of the noise grid, moving the XZ planes as far out of alignment with the square cube faces as possible. The <0.789, -0.577, -0.211> and <-0.211, -0.577, 0.789> are chosen out of all possibilities, because they're "the same" up to coordinate order. This reduces any slight probabilistic differences in noise behavior along those two world directions as much as possible.

engine/src/main/java/org/terasology/engine/utilities/procedural/PerlinNoise.java
KdotJPG: > It's also my impression that there's no reason to use Perlin noise anymore, and that Simplex doesn't have these artifacts in the first place. It's probably better to switch existing uses of Perlin noise to Simplex than to improve the Perlin noise. If there's actually some reason for something to continue using Perlin noise, then we should un-deprecate it and this change makes sense, but I don't think that's the case.

I agree the noise isn't great in its unmitigated/unrotated form, and would agree with decisions to phase it out. With the rotation, though, I've founds the results to be quite nice when used for overhang-enabled voxel terrain. I use a variant of it in [this project](https://github.com/KdotJPG/Simply-Improved-Terrain). If including the change and removing the deprecation, I could also rename the noise to "RPerlin" or "RotoPerlin" to distinguish it from the unmitigated algorithm, if that would make sense for this project. Any user-visible references could be updated accordingly to go with that (e.g. the "Perlin" world type). I recently began using this name [here](https://github.com/KdotJPG/Noise-Extras/blob/master/CachedColumnNoise/RotoperlinCachedColumnNoise.java) to refer to implementations of the old Perlin algorithm that have the built-in rotational improvement.

Re: Simplex, there are some artifacts present in the simplex noise given the gradient tables it uses, but they're not of the same strongly grid-aligned nature. I have some OpenSimplex2 code that I'd like to submit that would improve them, I just wound up tackling this first.

> Rotation

I explained a lot of it in [this other comment](https://github.com/MovingBlocks/Terasology/pull/4802#discussion_r660157980). Let me know if there's anything else I can clarify. Happy to add more details to a comment in the actual as well! The rotational axis wouldn't really describe much about what the rotation does, but perhaps I could mention that the rotation rotates the Y vector 45 degrees to point down all three of X, Y, Z equally.

pollend:
@KdotJPG have you joined the discord. the expertise is really well appreciated. :D

engine/src/main/java/org/terasology/engine/utilities/procedural/PerlinNoise.java
skaldarnar: I'm actually wondering whether we should just put this explanation in a code comment here :sweat_smile: Constants help a bit (you can at least give it a name), and maybe this would be a good place to add the docstring...

skaldarnar (commented):
@KdotJPG I think we're nearly ready to merge this, but I'd like to get the comments that we have in Github now into docstrings - could try to finish this up, please?

From the org meeting on 2021-07-03:

> we discussed the topic and will go ahead with merging the changes and un-deprecating Perlin noise

KdotJPG:
> 
> 
> @KdotJPG I think we're nearly ready to merge this, but I'd like to get the comments that we have in Github now into docstrings - could try to finish this up, please?
> 
> From the org meeting on 2021-07-03:
> 
> > we discussed the topic and will go ahead with merging the changes and un-deprecating Perlin noise

Oh - working on it! I had misunderstood hearing about this thinking it had already been merged! Will get these changes in tonight.

skaldarnar (approved):

PR #4807, opened 2021-07-02
test: make parts of PathManagerProvider public for use by MTE

keturn:
We want to be able to run a module's MTE tests without it scanning `/modules/*` every test case, as explained in #4638. We made a `PathManagerProvider` extension to help with this, but when it comes to MTE, I haven't found a way for MTEExtension to declare a dependency on the PathManagerProvider extension.

As a work-around for not having a way to compose Jupiter extensions, MTEExtension wants to call some of these methods directly, which requires making them public.

To help guard against _non_-test code from using these, I've adjusted the dependencies in terasology-module so that a module's `implementation` does not have access to this code in `engine-tests`. That's only for use in `testImplementation`.

There is some risk of that change breaking existing module code, but modules really _shouldn't_ have been using `engine-tests` classes in non-test code in the first place.

### DIFF
--- a/build-logic/src/main/kotlin/terasology-module.gradle.kts
+++ b/build-logic/src/main/kotlin/terasology-module.gradle.kts
@@ -44,7 +44,7 @@ configurations {
 // Set dependencies. Note that the dependency information from module.txt is used for other Terasology modules
 dependencies {
     implementation(group = "org.terasology.engine", name = "engine", version = moduleMetadata.engineVersion())
-    implementation(group = "org.terasology.engine", name = "engine-tests", version = moduleMetadata.engineVersion())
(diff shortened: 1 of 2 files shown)


jdrueckert:
So I just had a look through the omega usages of classes in `engine-tests/src/main/java/org/terasology/`.
- `HeadlessEnvironment` is used in `NameGenerator` module
- `WorldProvidingHeadlessEnvironment` is used in `Behaviors` and `Pathfinding` modules
However, all usages seem to be in test code, so guess we're fine with regards to omega. Everything beyond that we don't really care about that much currently anyway, so... from my side this is good to go.

jdrueckert (approved):

PR #4809, opened 2021-07-03
refactor(GameState): extract common code from MainMenu and HeadlessSetup

keturn:
I was working with the StateHeadlessSetup class and noticing it has a lot of code duplicated from another class. To make it easier to see what the differences between the various classes are, I extracted the common code to some methods.

No change in functionality is intended by this PR, but I did re-order the execution of some of these components, which has the potential to break something.

## Notes to Reviewers

I'm not thrilled about the state of things here, but it's a refactor, so please try to limit the scope of your requests to the refactoring. I'd also like to avoid getting pulled down the rabbit hole of putting _all_ the potentially related refactors we can think of into this PR.

Things it would be useful to have feedback on:

* Should this code really **not** be common between these two classes? If you think some of it was copied from one to the other by mistake and it really shouldn't be in both, this change could be counterproductive.
* _Create a superclass_ is not a great pattern, but since we need to initialize four fields on the instance, I haven't thought of another pattern that would fit without more extensive changes.
* Names of things. `AbstractState` is a particularly unhelpful name.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/core/modes/AbstractState.java
+++ b/engine/src/main/java/org/terasology/engine/core/modes/AbstractState.java
@@ -1,0 +1,84 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.engine.core.modes;
+
+import org.terasology.engine.context.Context;
+import org.terasology.engine.core.ComponentSystemManager;
+import org.terasology.engine.core.bootstrap.EntitySystemSetupUtil;
+import org.terasology.engine.entitySystem.entity.EntityRef;
+import org.terasology.engine.entitySystem.entity.internal.EngineEntityManager;
+import org.terasology.engine.entitySystem.event.internal.EventSystem;
+import org.terasology.engine.logic.console.Console;
+import org.terasology.engine.logic.console.ConsoleImpl;
+import org.terasology.engine.logic.console.ConsoleSystem;
+import org.terasology.engine.logic.console.commands.CoreCommands;
+import org.terasology.engine.logic.players.LocalPlayer;
+import org.terasology.engine.network.ClientComponent;
+import org.terasology.engine.recording.DirectionAndOriginPosRecorderList;
+import org.terasology.engine.recording.RecordAndReplayCurrentStatus;
+import org.terasology.engine.registry.CoreRegistry;
+import org.terasology.engine.rendering.nui.NUIManager;
+import org.terasology.engine.rendering.nui.internal.NUIManagerInternal;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/terasology/engine/core/modes/StateMainMenu.java
```diff
@@ -69,33 +44,15 @@ public StateMainMenu(String showMessageOnLoad) {
...
+        nuiManager = context.get(NUIManager.class);
         eventSystem.registerEventHandler(nuiManager);
-
-        componentSystemManager = new ComponentSystemManager(context);
-        context.put(ComponentSystemManager.class, componentSystemManager);
-
-        // TODO: Reduce coupling between Input system and CameraTargetSystem,
-        // TODO: potentially eliminating the following lines. See Issue #1126
-        CameraTargetSystem cameraTargetSystem = new CameraTargetSystem();
-        context.put(CameraTargetSystem.class, cameraTargetSystem);
```
keturn, line 93: It looks like I've removed this use of CameraTargetSystem in the main menu, but that's not really the case. It still happens within RegisterInputSystem called in this method.

pollend (approved):

engine/src/main/java/org/terasology/engine/core/modes/AbstractState.java
```diff
@@ -0,0 +1,84 @@
...
+
+import static com.google.common.base.Verify.verifyNotNull;
+
+public abstract class AbstractState implements GameState {
+    protected Context context;
+    protected EngineEntityManager entityManager;
+    protected EventSystem eventSystem;
+    protected ComponentSystemManager componentSystemManager;
+
+    protected void initEntityAndComponentManagers() {
```
skaldarnar, line 34: This does a lot with the `context` and strictly requires it. Would it make sense to make the `context` and argument of this method, similar to how we pass it to `createLocalPlayer` explcitly?

skaldarnar (approved):

engine/src/main/java/org/terasology/engine/core/modes/AbstractState.java
keturn: `createLocalPlayer` is the method I extracted first. Passing `context` let me make it `static`, which I figured would make it easier to move the method later if there were more copies of it.

Then I realized if I shuffled some things around I could also extract all this stuff that went in to `initEntityAndComponentManagers`, but because it assigns things to multiple instance fields, it wouldn't work as static.

Upon review, I see that all these fields are direct references to a thing in the context, so it probably could be split in to a "populate `context` with this stuff" static method and then a "pull some references out of context and cache them in fields" non-static method, but I worried that would make it more difficult to review if I included it in the same PR.

PR #4813, opened 2021-07-04
feat(ModuleManager): do not depend on PathManager to load modules from the classpath

keturn:
When modules run tests, the module and its dependencies are in the classpath. The test runner is unlikely to put unrelated modules in the classpath. This PR enables us to load those modules without registering everything in a `modules/` directory.

Required for some MTE improvements (PR pending).

No expected impact in normal execution, this only takes effect when the system property telling ModuleManager to use classpath modules is turned on.

### How to test

I think to thing to check is that module tests that are currently passing are still passing with this branch.

I _expect_ that works; if it doesn't, this may have to wait for merge until the accompanying MTE PR is ready.

- [x] unit tests
- [ ] try with module-tests that have dependencies; source-present and otherwise

### DIFF
--- a/engine-tests/src/test/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactoryTest.java
+++ b/engine-tests/src/test/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactoryTest.java
@@ -3,10 +3,10 @@
 
 package org.terasology.engine.core.module;
 
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.terasology.gestalt.module.Module;
-import org.terasology.gestalt.module.ModuleFactory;
 import org.terasology.gestalt.module.ModuleMetadata;
 import org.terasology.gestalt.module.sandbox.API;
 import org.terasology.gestalt.naming.Name;
@@ -15,17 +15,29 @@ import org.terasology.unittest.ExampleClass;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 
+import static com.google.common.truth.Truth8.assertThat;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class ClasspathCompromisingModuleFactoryTest {
     static final Class<?> someClassOutsideTheModule = ClasspathCompromisingModuleFactory.class;
+    static final String METADATA_NAME = "module.json";
+
+    ClasspathCompromisingModuleFactory factory;
+
+    @BeforeEach
+    public void newFactory() {
+        factory = new ClasspathCompromisingModuleFactory();
+        factory.setDefaultLibsSubpath("build/libs");
+    }
 
     @Test
     public void directoryModuleContainsClass() {
-        ModuleFactory factory = new ClasspathCompromisingModuleFactory();
-
         // This test assumes that the unittest module is under the current working directory (`engine-test/`)
         File engineTestDirectory = new File(System.getProperty("user.dir", "."));
         ModuleMetadata metadata = new ModuleMetadata(new Name("unittest"), new Version("1.0.0"));
@@ -42,8 +54,6 @@ public class ClasspathCompromisingModuleFactoryTest {
     @Test
     @Disabled("TODO: need a jar module containing classes")
     public void archiveModuleContainsClass() throws IOException {
-        ModuleFactory factory = new ClasspathCompromisingModuleFactory();
-
         Module module = factory.createArchiveModule(new File("FIXME.jar"));
 
         Class<?> someClassInTheModule = module.getModuleManifest().getTypesAnnotatedWith(API.class).iterator().next();
@@ -59,5 +69,32 @@ public class ClasspathCompromisingModuleFactoryTest {
         //   - m/build/classes/org/t/Foo.class
         //   - m/build/libs/foo.jar
         // load m as directory module while foo.jar is on classpath
+    }
+
+    @Test
+    public void canGetPathFromJarResource() throws MalformedURLException {
+        // A jar file on the classpath but not in a local build directory.
+        URL jarUrl = new URL("jar:file:/example/Terasology/cachedModules/CoreAssets-2.3.0-SNAPSHOT.jar!/module.json");
+        Path expectedPath = Paths.get("/example/Terasology/cachedModules/CoreAssets-2.3.0-SNAPSHOT.jar");
+
+        assertThat(factory.canonicalModuleLocation(METADATA_NAME, jarUrl)).isEqualTo(expectedPath);
+    }
+
+    @Test
+    public void canGetPathFromLocalJarBuild() throws MalformedURLException {
+        // A jar file on the classpath that is a build directory in a local development workspace
+        URL jarUrl = new URL("jar:file:/example/Terasology/modules/CoreAssets/build/libs/CoreAssets-2.3.0-SNAPSHOT.jar!/module.json");
+        Path expectedPath = Paths.get("/example/Terasology/modules/CoreAssets");
+
+        assertThat(factory.canonicalModuleLocation(METADATA_NAME, jarUrl)).isEqualTo(expectedPath);
+    }
+
+    @Test
+    public void canGetPathFromFilesystemResource() throws MalformedURLException {
+        // A directory on the classpath that is a build directory in a local development workspace
+        URL fileUrl = new URL("file:/example/Terasology/modules/Health/build/classes/module.json");
+        Path expectedPath = Paths.get("/example/Terasology/modules/Health");
+
+        assertThat(factory.canonicalModuleLocation(METADATA_NAME, fileUrl)).isEqualTo(expectedPath);
     }
 }
--- a/engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
+++ b/engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
@@ -94,17 +94,18 @@ class ClasspathCompromisingModuleFactory extends ModuleFactory {
         return findModuleRoot(relativePathFromModuleRoot, path).orElse(path);
     }
 
-    private Path modulePathFromMetadataJarUrl(URL url) {
-        checkArgument(url.getProtocol().equals("jar"), "Not a jar URL: %s", url);
+    private Path modulePathFromMetadataJarUrl(URL jarUrl) {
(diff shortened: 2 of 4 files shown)


keturn:
I think this is in decent shape and is passing the basic smoke tests locally. Ready for review.

and I _think_ it's reviewable independent of any MTE changes that build on it later.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -136,6 +146,35 @@ private void loadModulesFromApplicationPath(PathManager pathManager) {
         scanner.scan(registry, paths);
     }
 
+    private void loadModulesFromClassPath() {
+        ClasspathCompromisingModuleFactory moduleFactory = (ClasspathCompromisingModuleFactory) this.moduleFactory;
+        for (String metadata_name : moduleFactory.getModuleMetadataLoaderMap().keySet()) {
```
skaldarnar, line 151: We usually don't use snake case in our Java code. 
```suggestion
        for (String metadataName : moduleFactory.getModuleMetadataLoaderMap().keySet()) {
```

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -136,6 +146,35 @@ private void loadModulesFromApplicationPath(PathManager pathManager) {
...
+            Enumeration<URL> urls;
+            try {
+                urls = ClassLoader.getSystemResources(metadata_name);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
```
skaldarnar, line 156: What's the reason we turn the `IOException` into a `RuntimeException` here? Usually a runtime exception is something that could have been prevented by the caller - is that the case here?

Or is it just to have an unchecked exception that bubbles up without annotating `throws ...` everywhere?

We should probably start building a common set of exceptions we want to use in the code base, with dedicated error handling and meaningful logging....

skaldarnar: Hm, maybe just using `RuntimeException` is actually good enough because the _cause_ is preserved when handing the `e` through :thinking: 

https://softwareengineering.stackexchange.com/questions/121328/is-it-good-practice-to-catch-a-checked-exception-and-throw-a-runtimeexception

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
```diff
@@ -136,6 +146,35 @@ private void loadModulesFromApplicationPath(PathManager pathManager) {
...
+            }
+            while (urls.hasMoreElements()) {
+                URL url = urls.nextElement();
+                logger.debug("Probably a module in U:{}", url);
+                Path path = moduleFactory.canonicalModuleLocation(metadata_name, url);
```
skaldarnar, line 161:
```suggestion
                Path path = moduleFactory.canonicalModuleLocation(metadataName, url);
```

engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
```diff
@@ -52,6 +63,97 @@ public Module createArchiveModule(ModuleMetadata metadata, File archive) throws
...
+     * Accounts for loading modules from development workspaces that may have their build directories
+     * on the classpath, as is the case when running tests.
+     *
+     * @see #setDefaultCodeSubpath
+     * @see #setDefaultLibsSubpath
+     *
+     * @param metadataName the expected name of the metadata file, as it would appear in {@link #getModuleMetadataLoaderMap()}
+     * @param metadataUrl a URL of a metadata file, such as might be returned from {@link ClassLoader#getSystemResource}
+     * @return the module's base directory, or a jar file if it doesn't look like a local build
+     */
```
skaldarnar, line 78: :+1:

engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
```diff
@@ -52,6 +63,97 @@ public Module createArchiveModule(ModuleMetadata metadata, File archive) throws
...
+        }
+    }
+
+    private Path modulePathFromMetadataFileUrl(String metadata_name, URL url) {
+        Path path = fromUrl(url);
+        // We are considering the location of a resource file, so compare it to the code path.
+        // Include the metadata_name in case it has path components of its own.
+        Path relativePathFromModuleRoot = Paths.get(getDefaultCodeSubpath(), metadata_name);
+        return findModuleRoot(relativePathFromModuleRoot, path).orElse(path);
+    }
```
skaldarnar, line 95:
```suggestion
    private Path modulePathFromMetadataFileUrl(String metadataName, URL url) {
        Path path = fromUrl(url);
        // We are considering the location of a resource file, so compare it to the code path.
        // Include the metadata name in case it has path components of its own.
        Path relativePathFromModuleRoot = Paths.get(getDefaultCodeSubpath(), metadataName);
        return findModuleRoot(relativePathFromModuleRoot, path).orElse(path);
    }
```

engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
```diff
@@ -52,6 +63,97 @@ public Module createArchiveModule(ModuleMetadata metadata, File archive) throws
...
+
+    private Path modulePathFromMetadataJarUrl(URL url) {
+        checkArgument(url.getProtocol().equals("jar"), "Not a jar URL: %s", url);
+        try {
+            JarURLConnection connection = (JarURLConnection) url.openConnection();
+            url = connection.getJarFileURL();
```
skaldarnar, line 101: We're overriding the URL we get passed in here - is that a good practice? The URL we get passed in is not the same as the one we get from `getJarFileURL()`, or is it?

engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
```diff
@@ -52,6 +63,97 @@ public Module createArchiveModule(ModuleMetadata metadata, File archive) throws
...
+    private static Optional<Path> findModuleRoot(Path relativePathFromModuleRoot, Path path) {
+        if (path.endsWith(relativePathFromModuleRoot)) {
+            int relativeDepth = relativePathFromModuleRoot.getNameCount();
+            Path parentPath = path.subpath(0, path.getNameCount() - relativeDepth);
+            if (path.getRoot() != null) {  // TODO: test case
+                parentPath = path.getRoot().resolve(parentPath);
+            }
+            return Optional.of(parentPath);
+        } else {
+            logger.warn(" +- does not seem to be in a build directory {}", path);
```
skaldarnar, line 137: Does the ` +-` prefix in this logging line make more sense in the context this method is called?

engine-tests/src/test/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactoryTest.java
```diff
@@ -15,17 +15,29 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
 
+import static com.google.common.truth.Truth8.assertThat;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class ClasspathCompromisingModuleFactoryTest {
     static final Class<?> someClassOutsideTheModule = ClasspathCompromisingModuleFactory.class;
+    static final String METADATA_NAME = "module.json";
```
skaldarnar, line 29: Do you use a different metadata file name here on purpose from what we use in Terasology? I think we're going with `module.txt`, even though I'd like to use `module.json` for my editors to understand the file format ^^

skaldarnar (commented):
Some minor remarks, overall looks good to me :+1:

engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
keturn: The indentation helped when I had a bunch more logging earlier. but yeah, it's worth a second look to see if it still helps.

engine-tests/src/test/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactoryTest.java
keturn: The trade-off here is not wanting to spend too many lines of code in setUp overriding defaults that aren't significant to the test cases.

and the default name isn't exposed as a constant by gestalt.

but I do agree that the test cases shouldn't look too different than the values we see in practice, which is why I did set the libs path.

so yeah, I'm wishy-washy on this one and open to making changes, just having a hard time settling on what makes things most clear.

engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
keturn: It converts it from `jar:file:/stuff/Foo.jar!/module.txt` to `file:/stuff/Foo.jar`.

but oh, I moved the conditional out of here, that makes things easier. So if we want to keep a practice of treating method arguments as `final`, I can introduce another local variable for that.

engine/src/main/java/org/terasology/engine/core/module/ModuleManager.java
keturn: > Or is it just to have an unchecked exception that bubbles up without annotating throws ... everywhere?

Mostly this.

I've been pretty sloppy about things like this, figuring that we can defer the decision to make well-typed exceptions until we see which ones come up in practice and we have ideas about how to deal with them.

One consideration is how likely they are. An IO error as a result of a file operation to a user-supplied location is one thing. But a read-only `getSystemResources` call? Seems like the environment would have to be pretty broken for that to happen. Doesn't make much sense to me to document that as part of this method's interface.

engine/src/main/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactory.java
skaldarnar: I'd prefer to treat method arguments as `final` (and even mark them as such). 👍

engine-tests/src/test/java/org/terasology/engine/core/module/ClasspathCompromisingModuleFactoryTest.java
skaldarnar: I think the test cases are clear enough, no need to change anything. I was just wondering whether there's a deeper reason for this 🙃

skaldarnar (approved):

skaldarnar:
> try with module-tests that have dependencies; source-present and otherwise
I haven't tests

keturn:
I've been using this locally for some things, and just confirmed I can run the SimpleFarming tests whether its module dependencies are in `modules/` or provided as jars by gradle. That's enough confirmation for me for test-only code.

PR #4814, opened 2021-07-05
perf: avoid calling GLFW.glfwGetWindowSize

pollend:
`GLFW.glfwGetWindowSize` is fairly expensive so this is only updated once per frame. should help a little bit with performance and I'm still looking into some other modifications to the render pipeline.


DarkWeird:
Seems you can place windowsize callback there

skaldarnar (approved):

PR #4818, opened 2021-07-08
feat: allow for ranged item activation

skaldarnar:
This originates from a discussion with @ahv15 and @jdrueckert on allowing ranged activation of items, e.g., to activate the magic staff in LigthAndShadow targeting a player a couple blocks away. We noticed that `ActivateEvent` fits the use case, but the computation of the `target` is bound to the interaction range as defined in the character component.

As a mitigation, this PR adds a generic `RangeComponent` to indicate some kind of ranged property, e.g., the interaction range of an item, an activation range of a tower or trap, ...

Eventually, we probably want to decouple the interaction range from the character component, and instead just use the `RangeComponent` on the player entity itself.

- feat: allow for ranged item activation
- feat: add simple RangeComponent
- feat: allow owned entities to increase interaction range
- doc: add API contracts as docstring to activation-related events

### Follow-Up Questions

These questions may be discussed now, but are not blocking merging this PR.

- [ ] the activation-related events look very similar to each other - maybe we can share code (or at least an interface) between them?
- [x] this does change the semantics of hit normal and hit position on those events. I documented this to state the method contracts, but we should keep an eye open whether this breaks anything.
- [ ] there's a tight coupling between `LocalPlayer` and the `CharacterSystem` for prediction. How can we test this? Can/should they share code?

### DIFF
--- a/engine/src/main/java/org/terasology/engine/logic/players/LocalPlayer.java
+++ b/engine/src/main/java/org/terasology/engine/logic/players/LocalPlayer.java
@@ -10,6 +10,7 @@ import org.terasology.engine.logic.characters.CharacterComponent;
 import org.terasology.engine.logic.characters.CharacterMovementComponent;
 import org.terasology.engine.logic.characters.CharacterSystem;
 import org.terasology.engine.logic.common.ActivateEvent;
+import org.terasology.engine.logic.common.RangeComponent;
 import org.terasology.engine.logic.location.LocationComponent;
 import org.terasology.engine.logic.characters.events.ActivationPredicted;
 import org.terasology.engine.logic.characters.events.ActivationRequest;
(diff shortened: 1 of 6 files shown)


skaldarnar:
@ahv15 @meetcshah19 this might be of interest ... :nerd_face:

engine/src/main/java/org/terasology/engine/logic/players/LocalPlayer.java
```diff
@@ -236,24 +237,27 @@ private boolean activateTargetOrOwnedEntity(EntityRef usedOwnedEntity) {
...
             activatedObject.send(new ActivationPredicted(character, result.getEntity(), originPos, direction,
                     result.getHitPoint(), result.getHitNormal(), activationId));
+
             character.send(new ActivationRequest(character, ownedEntityUsage, usedOwnedEntity, eventWithTarget,
                     result.getEntity(),
                     originPos, direction, result.getHitPoint(), result.getHitNormal(), activationId));
             return true;
-        } else if (ownedEntityUsage) {
-            usedOwnedEntity.send(new ActivationPredicted(character, EntityRef.NULL, originPos, direction,
-                    originPos, new Vector3f(), activationId));
```
skaldarnar, line 252: Instead of `originPos` and `new Vector3f()` we now put `null` for both in case there is no target. This changes semantics, but I added this as docstring to at least help with debugging.

skaldarnar:
`CharacterComponent#interactionRange` is not used in many places as far as I can tell and could easily™ be migrated fully to `RangeComponent` if we want to do that.

jdrueckert (approved):
> the activation-related events look very similar to each other - maybe we can share code (or at least an interface) between them?

This should be tracked outside of this PR.

> there's a tight coupling between LocalPlayer and the CharacterSystem for prediction. How can we test this? Can/should they share code?

I think they should share code, yes. Not sure, where that code should live though. I would argue that the effort to write a test ensuring the two code spots don't diverge is better spent on overhauling the `logic` package. Maybe the separation of players and characters doesn't make sense anymore and there's even more duplication in there that we didn't find yet :shrug: 

Apart from that, this PR looks fine from my side.

PR #4819, opened 2021-07-10
perf: upgrade with proto3 and reworked network packets

pollend:
The main thing I reworked is to use the oneof keyword which means that at any one time the defined type can only have one of that type of message. this should save us some space per request/response from the client and server. optional fields would have to be encode if they are defined otherwise. per packet I think this should be a small amount of space saved.


jdrueckert:
[Protobuf v2 -> v3](https://trello.com/c/zxHzMyzg/169-protobuf-v2-v3)

mkienenb:
Should be split out:   2 -> 3 protobuf upgrade in one PR,   Other stuff in one (or more PRs), so that individual changes can be evaluated and measured separately.

mkienenb:
After a discussion with Michael, the metrics that make the most sense to use for determining the effectiveness of each change are the total time to serialize/compress/build the packets and the amount of final payload data in the packets for a certain set of events, including client->server and server->client.

So
- client -> server  (prepare client-originating data)  [final payload size, time]
- client -> server (unpack server-originating data)  [final payload size, time]
- server -> client  (prepare server-originating data)  [final payload size, time]
- server -> client (unpack client-originating data)  [final payload size, time]

The size of the data packet can be measured either in-code at the time of (channel.write*() methods or using a proxy server to remove metric collection impacts from the client and/or server.

I'm not sure where the time to prepare the packets would be measured, but maybe measuring the total time spent in NetClient.update()  through until             NetData.NetMessage.Builder message = NetData.NetMessage.newBuilder();
            message.setTime(time.getGameTimeInMs());
            sendRegisteredBlocks(message);
            sendChunkInvalidations(message);
            sendNewChunks(message);
            sendRemovedEntities(message);
            sendInitialEntities(message);
            sendDirtyEntities(message);
            sendEvents(message);


There may be additional impacts not handled by these metrics, such as how waiting until a large amount of data is collected impacts game latency/performance.   It's unclear to me how these can be measured.

PR #4820, opened 2021-07-10
perf(network): replace de-/inflate with lz4 for both directions

pollend:
we would need to have a dry run test on a server to verify if this improves the actual network performance. Lz4 is a lot faster compression algorithm for both the consumer and reader. would it be possible to create a one off build to have this on a test server @Cervator. I also run the compress and deflation for both send and receiving packets.   

https://github.com/lz4/lz4

### DIFF
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -59,6 +59,7 @@ dependencies {
     api group: 'net.sf.trove4j', name: 'trove4j', version: '3.0.3'
     implementation group: 'io.netty', name: 'netty-all', version: '4.1.53.Final'
     implementation group: 'com.google.protobuf', name: 'protobuf-java', version: '2.6.1'
+    implementation 'org.lz4:lz4-java:1.8.0'
     implementation group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.2'
     // Javax for protobuf due to @Generated - needed on Java 9 or newer Javas
     // TODO: Can likely replace with protobuf Gradle task and omit the generated source files instead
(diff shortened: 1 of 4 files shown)


engine/build.gradle
```diff
@@ -59,6 +59,7 @@ dependencies {
     api group: 'net.sf.trove4j', name: 'trove4j', version: '3.0.3'
     implementation group: 'io.netty', name: 'netty-all', version: '4.1.53.Final'
     implementation group: 'com.google.protobuf', name: 'protobuf-java', version: '2.6.1'
+    implementation 'org.lz4:lz4-java:1.8.0'
```
skaldarnar, line 62: Please use the same syntax/format for adding this dependency as for the dependencies around it :wink:

engine/src/main/java/org/terasology/engine/network/internal/pipelineFactory/InfoRequestPipelineFactory.java
```diff
@@ -22,7 +24,7 @@
  * A pipeline that requests {@link org.terasology.engine.network.ServerInfoMessage} before it auto-disconnects. This is similar
  * to {@link TerasologyClientPipelineFactory}.
  */
-public class InfoRequestPipelineFactory extends ChannelInitializer {
+public class  InfoRequestPipelineFactory extends ChannelInitializer {
```
skaldarnar, line 27: Please double check for checkstyle alerts and formatting (I can spot an additional space here :yum: )

skaldarnar (commented):
I cannot say much about the semantic content of this PR, it looks like a 1:1 replacement to me. As you said, would be nice to test on the play-test server to see whether this improves anything.

skaldarnar (approved):

PR #4821, opened 2021-07-11
perf: remove extra length prepend

pollend:
here are some small tweaks the stream. looks like we've encoded the length of the packet like 3 times. prepend a size before the compressed frame, lz4frame prepends a length to the compressed packet, followed by another length encoding for the protobuf message. l

rework

this is what this change does. lz4frame will split the payload into multiple frames that are sent over the wire. we need the length to be prepend to the protobuf message so we can decode individual protobuf message. 

[frame_rework.zip](https://github.com/MovingBlocks/Terasology/files/6796293/frame_rework.zip)


jdrueckert (approved):
Nice PR description :+1:

PR #4822, opened 2021-07-14
feat: rework chunk ordering using Reactor

tolziplohu:
### Contains

This is a rewrite of https://github.com/MovingBlocks/Terasology/pull/4773 to use Reactor for concurrency (see https://github.com/MovingBlocks/Terasology/issues/4798 and https://github.com/MovingBlocks/Terasology/pull/4786). It does the same thing as the old PR, but is a lot simpler and has more reliable concurrency. Performace should be about the same.

### How to test

Try flying around the world with this PR - it should be the same as https://github.com/MovingBlocks/Terasology/pull/4773 and much better than normal.

Unlike https://github.com/MovingBlocks/Terasology/pull/4773, in clients connected to a remote server, this PR currently doesn't bother reordering chunks recieved from the server to load the closest ones first. I don't think there's much reason to do that, since the bottleneck will almost always be the server and network, but if it's a problem we can switch back to a PriorityBlockingQueue.

### DIFF
--- a/engine-tests/src/test/java/org/terasology/engine/world/chunks/localChunkProvider/LocalChunkProviderTest.java
+++ b/engine-tests/src/test/java/org/terasology/engine/world/chunks/localChunkProvider/LocalChunkProviderTest.java
@@ -3,6 +3,7 @@
 package org.terasology.engine.world.chunks.localChunkProvider;
 
 import com.google.common.collect.Maps;
+import org.joml.Vector3f;
 import org.joml.Vector3i;
 import org.joml.Vector3ic;
 import org.junit.jupiter.api.AfterEach;
@@ -13,11 +14,11 @@ import org.mockito.ArgumentCaptor;
 import org.terasology.engine.entitySystem.entity.EntityManager;
 import org.terasology.engine.entitySystem.entity.EntityRef;
 import org.terasology.engine.entitySystem.event.Event;
+import org.terasology.engine.logic.location.LocationComponent;
 import org.terasology.engine.world.BlockEntityRegistry;
 import org.terasology.engine.world.block.BeforeDeactivateBlocks;
 import org.terasology.engine.world.block.Block;
 import org.terasology.engine.world.block.BlockManager;
-import org.terasology.engine.world.block.BlockRegion;
 import org.terasology.engine.world.block.OnActivatedBlocks;
 import org.terasology.engine.world.block.OnAddedBlocks;
 import org.terasology.engine.world.chunks.Chunk;
@@ -35,18 +36,15 @@ import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 class LocalChunkProviderTest {
 
-    private static final int WAIT_CHUNK_IS_READY_IN_SECONDS = 30;
+    private static final int WAIT_CHUNK_IS_READY_IN_SECONDS = 5;
 
     private LocalChunkProvider chunkProvider;
     private EntityManager entityManager;
@@ -58,6 +56,8 @@ class LocalChunkProviderTest {
     private Block blockAtBlockManager;
     private TestStorageManager storageManager;
     private TestWorldGenerator generator;
+    private RelevanceSystem relevanceSystem;
+    private EntityRef playerEntity;
 
     @BeforeEach
     public void setUp() {
@@ -81,7 +81,8 @@ class LocalChunkProviderTest {
                 chunkCache);
         chunkProvider.setBlockEntityRegistry(blockEntityRegistry);
         chunkProvider.setWorldEntity(worldEntity);
-        chunkProvider.setRelevanceSystem(new RelevanceSystem(chunkProvider)); // workaround. initialize loading pipeline
+        relevanceSystem = new RelevanceSystem(chunkProvider);
+        chunkProvider.setRelevanceSystem(relevanceSystem); // workaround. initialize loading pipeline
     }
 
     @AfterEach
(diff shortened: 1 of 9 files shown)


pollend:
oi, didn't expect this. I'll have to take a look tonight.

engine-tests/src/test/java/org/terasology/engine/world/chunks/localChunkProvider/LocalChunkProviderTest.java
```diff
@@ -44,7 +44,7 @@
 
 class LocalChunkProviderTest {
 
-    private static final int WAIT_CHUNK_IS_READY_IN_SECONDS = 5;
+    private static final int WAIT_CHUNK_IS_READY_IN_SECONDS = 30;
```
tolziplohu: While 5 is long enough for my computer, it appears that the CI machine does unfortunately need 30 seconds to generate all the chunks. It probably has a lot of threads running at once.

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
```diff
@@ -44,120 +32,162 @@
  */
 public class ChunkProcessingPipeline {
```
pollend, line 33: I think I would like to just do away with this ChunkProcessingPipeline. that could be addressed later

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
```diff
@@ -44,120 +32,162 @@
...
-        };
-        chunkProcessor = new ExecutorCompletionService<>(executor,
-                new PriorityBlockingQueue<>(800, comparable));
-        reactor = new Thread(this::chunkTaskHandler);
-        reactor.setDaemon(true);
-        reactor.setName("Chunk-Processing-Reactor");
-        reactor.start();
+        scheduler = Schedulers.newParallel("chunk processing", NUM_TASK_THREADS);
+        Flux<Chunk> stream = chunkStream.subscribeOn(scheduler);
+        for (int i = 0; i < NUM_TASK_THREADS; i++) {
```
pollend, line 56: use the parallel scheduler this is cpu bounded. we probably also want to limit the number of subscribers to the number of cores on the machine.
```suggestion
        Flux<Chunk> stream = chunkStream.subscribeOn(Schedulers.parallel());
        for (int i = 0; i < NUM_TASK_THREADS; i++) {
```

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
```diff
@@ -44,120 +32,162 @@
...
+    private boolean processChunkTasks() {
+        boolean anyChanged = false;
+        for (ChunkProcessingInfo chunkProcessingInfo : chunkProcessingInfoMap.values()) {
+            ChunkTask chunkTask = chunkProcessingInfo.getChunkTask();
+            if (chunkTask != null) {
+
+                List<Chunk> providedChunks = new ArrayList<>();
+                boolean satisfied = true;
+                for (Vector3ic pos : chunkTask.getRequirements()) {
+                    Chunk chunk = getChunkBy(chunkProcessingInfo.getChunkTaskProvider(), pos);
```
pollend, line 135: what is the reasoning behind this. seems so strange to me. so we have 2 threads pushing chunks to the same concurrent map and at the same time we also attempt to iterate over the collection. there is going to be a lot of thread contention with all this as we lock and unlock this same map.

engine/src/main/java/org/terasology/engine/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -437,10 +405,101 @@ private boolean isChunkReady(Chunk chunk) {
...
+    public void notifyRelevanceChanged() {
+        loadingPipeline.notifyUpdate();
+    }
+
+    private void updateList() {
+        relevanceSystem.neededChunks()
+                .filter(pos -> !chunksInRange.contains(pos))
+                .forEach(chunksInRange::add);
+        chunksInRange.removeIf(x -> !relevanceSystem.isChunkInRegions(x) || isChunkReady(x));
+        chunksInRange.sort(relevanceSystem.createChunkPosComparator().reversed());
```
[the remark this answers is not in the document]
tolziplohu: I don't think so, the only place we add to `chunksInRange` is the line above which `filter`s out duplicates.

engine/src/main/java/org/terasology/engine/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -437,10 +405,101 @@ private boolean isChunkReady(Chunk chunk) {
...
+            }
+            i++;
+        }
+        return anyChanged;
+    }
+
+    private Chunk genChunk(Vector3ic pos) {
+        ChunkStore chunkStore = storageManager.loadChunkStore(pos);
+        Chunk chunk;
+        EntityBufferImpl buffer = new EntityBufferImpl();
```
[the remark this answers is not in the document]
tolziplohu: That's a good idea, I just moved this code from somewhere else in the file but I'll change it.

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
tolziplohu: I am using the parallel scheduler, but with a name and limiting it to `NUM_TASK_THREADS` threads. From the docs, it sounds like that's the only difference between `newParallel` and `parallel`. More than `NUM_TASK_THREADS` workers wouldn't be helpful as there are only that many subscriptions, so are you just talking about the case where the host CPU has less than `NUM_TASK_THREADS` cores?

pollend: just realized this is a consuming operation so maybe push the operation to boundedElastic(). should re-claim the thread when we're finished.

tolziplohu: Do you mean when the player is standing still and chunk generation has finished, or after a world is closed? In the former case, I don't think that usually happens for very long, so I would guess that reclaiming the workers would have more overhead; in the latter case, the whole scheduler is already disposed after the world is closed.

engine/src/main/java/org/terasology/engine/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -437,10 +405,101 @@ private boolean isChunkReady(Chunk chunk) {
...
     }
 
+    public void notifyRelevanceChanged() {
+        loadingPipeline.notifyUpdate();
+    }
+
+    private void updateList() {
+        relevanceSystem.neededChunks()
+                .filter(pos -> !chunksInRange.contains(pos))
+                .forEach(chunksInRange::add);
```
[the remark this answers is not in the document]
tolziplohu: That is a little bit clearer, but I think I would prefer keeping it as is to avoid allocating the extra list.

engine/src/main/java/org/terasology/engine/world/chunks/localChunkProvider/LocalChunkProvider.java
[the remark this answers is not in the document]
tolziplohu: Oh wait, duplicates actually are possible when there are multiple clients connected with overlapping relevance regions, since there will be duplicates in `neededChunks`. However, that shouldn't actually cause any problems because the code to generate chunks checks `currentlyProcessing` first.

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
```diff
@@ -53,10 +53,19 @@
...
+     * Always true in a real game.
+     */
+    protected ChunkProcessingPipeline(Function<Vector3ic, Chunk> chunkProvider, Flux<Chunk> chunkStream, boolean threaded) {
         this.chunkProvider = chunkProvider;
         scheduler = Schedulers.newParallel("chunk processing", NUM_TASK_THREADS);
-        Flux<Chunk> stream = chunkStream.subscribeOn(scheduler);
-        for (int i = 0; i < NUM_TASK_THREADS; i++) {
+        Flux<Chunk> stream = threaded ? chunkStream.subscribeOn(scheduler) : chunkStream;
+        for (int i = 0; i < (threaded ? NUM_TASK_THREADS : 1); i++) {
             stream.subscribe(new BaseSubscriber<Chunk>() {
```
pollend: this is kind of strange, seems like something just so it passes the unit test. I'm not sure this is a good direction that resolves the test?

tolziplohu: It just doesn't use separate threads in the pipeline tests. I agree that that's not ideal, but it's better than adding `Thread.sleep(1000)` after each `Flux.create()`, which is basically what we'd have to do otherwise to ensure that the subscription actually happens before we try to submit chunks (that was what was causing the failing test).

I guess another option would be adding another signal object for waiting until subscription happens, plus a flag to check if it's already subscribed - that would probably work, but I'm not sure it would actually be less intrusive. If you think that would be better, though, I can do that.

pollend: just make  Schduler a parameter then you can use the FIFO scheduler.

tolziplohu: Good idea, I've just implemented that. Unfortunately there isn't an existing scheduler that runs everything immediately on the same thread, but a custom one that does that was just a 10 line anonymous class.

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
```diff
@@ -53,10 +53,19 @@
      * Create ChunkProcessingPipeline.
      */
     public ChunkProcessingPipeline(Function<Vector3ic, Chunk> chunkProvider, Flux<Chunk> chunkStream) {
+        this(chunkProvider, chunkStream, true);
+    }
+
+    /**
+     * @param threaded Whether to use worker threads to generate chunks. If it's false, all processing will be done on the main thread.
+     * It should be set to false in tests that request processing of specific chunks.
+     * Always true in a real game.
```
skaldarnar: I think that doing away with the complexity of multi-threading during tests hopefully helps to fight flakiness and instability. 

However, this also means that we're not testing the "real thing" here. The good part still is that if something goes wrong in the end at runtime we know to start looking at the concurrency parts...

@keturn might this pitfall be also occurring in Pathfinding, even without using Reactor?

tolziplohu: I do agree, it's unfortunate that the concurrent version doesn't work reliably on CI. I'd kind of still like to be able to run it concurrently on my computer, since it does work reliably on a normal computer, but I don't know if there's a nice way to do that.

engine/src/main/java/org/terasology/engine/world/chunks/localChunkProvider/LocalChunkProvider.java
```diff
@@ -437,10 +406,114 @@ private boolean isChunkReady(Chunk chunk) {
...
+                }
+
+                while (positionsPending.size() < numChunks && !chunksInRange.isEmpty()) {
+                    Vector3ic pos = chunksInRange.remove(chunksInRange.size() - 1);
+                    if (currentlyProcessing.contains(pos) || loadingPipeline.isPositionProcessing(pos)) {
+                        continue;
+                    }
+
+                    positionsPending.add(pos);
+                    currentlyProcessing.add(pos);
```
keturn, line 475: Important thing to know (for newbies like me to the Flux API): This emitter function we're passing to Flux.create (on line 460) is called with a new sink _for each subscriber._

That's how we're managing to add multiple chunk-processing subscribers to this without having each subscriber process the _same_ chunks.

keturn (changes requested):
I think I've got the gist of it now, though I haven't looked at all the details.

I fear we're torturing the Flux model with this.

I think that conceptually, we have some sort of publisher of relevant-locations-of-chunks at the top, and something that puts chunk data in to a chunk cache at the bottom, and we want some kind of

```java
f(Flux<Vector3ic> locations) {
   Flux<Chunk> fullChunks = locations
      .parallel()
      .map(chunkProvider);
   fullChunks.subscribe(cacheChunk);
}
```

with maybe some kind of buffering or [windowing operator](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#windowTimeout-int-java.time.Duration-) in there if you think we need to tune `CHUNKS_AT_ONCE`.

I think that the way this implementation of `LocalChunkProvider.chunkFlux` returns a `Flux<Chunk>` that _looks different to every subscriber_ is technically fitting within the bounds of the types, but is contrary to the spirit of a Flux stream.

and I still think that ChunkProcessingPipeline creating a new Scheduler with its own thread pool is a thing we should do only as a last resort. I could be convinced that since chunk generation is such an important and ever-present task, it's worth making an exception for if necessary, but in general that's not a pattern we want systems to use.

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
keturn: A _lot_ of the complexity of this code is there explicitly for the reason of allowing chunk generation to happen in parallel.

If there were only a single-threaded sink processing things from a queue, there's a lot of synchronization and `currentyProcessing` sorts of variables we wouldn't need.

If we are never testing the multi-threaded scenarios, we're seriously undermining the value of our tests.

engine/src/main/java/org/terasology/engine/world/chunks/localChunkProvider/LocalChunkProvider.java
tolziplohu: I agree, named methods would be good and I think `currentlyProcessing` should just be a member variable. I'll change that, and add some comments explaining when things get called too, since I agree that it's not very intuitive.

tolziplohu:
> I think that conceptually, we have some sort of publisher of relevant-locations-of-chunks at the top, and something that puts chunk data in to a chunk cache at the bottom, and we want some kind of
>
>     f(Flux<Vector3ic> locations) {
>       Flux<Chunk> fullChunks = locations
>          .parallel()
>          .map(chunkProvider);
>       fullChunks.subscribe(cacheChunk);
>     }
>
> with maybe some kind of buffering or windowing operator in there if you think we need to tune CHUNKS_AT_ONCE.

That would definitely be nicer, and it's similar to what I tried to do first, but unfortunately it didn't seem to work with our execution model. One problem is that a `Flux<Vector3ic>` and then a `Function<Vector3ic, Chunk>` doesn't work well with the `RemoteChunkProvider`, which would have to maintain another hash map of received chunks. Another is that when there are no more chunks to generate right now and execution pauses, there's no way to wake it up again from the main thread - a possible solution to this is a signal object that the location flux waits on, but that's not very nice either.

> I think that the way this implementation of LocalChunkProvider.chunkFlux returns a Flux<Chunk> that looks different to every subscriber is technically fitting within the bounds of the types, but is contrary to the spirit of a Flux stream.

I'm also wondering about this - there's no mention in the Reactor docs that a flux should always emit the same values to each subscriber, but I also can't find any examples that do that.

keturn:
> Another is that when there are no more chunks to generate right now and execution pauses, there's no way to wake it up again from the main thread

_This_ is the part I keep getting stuck on when trying to think about this design.

I might help if I knew more about what the wake-up triggers would be. Every player movement? Or is there some set of relevance regions that we'd observe changes to?

tolziplohu:
> It might help if I knew more about what the wake-up triggers would be. Every player movement? Or is there some set of relevance regions that we'd observe changes to?

It's woken up whenever the relevance region is updated, by calling `LocalChunkProvider.notifyRelevanceChanged()`. In practice  that's every time the player moves.

pollend:
@skaldarnar anything else blocking this?

pollend (approved):
I realize we lose some testability with this change since the scheduler is limited to a single thread in the test, but wouldn't mind proceeding with the change regardless.

DarkWeird (commented):
Reactor contains Scheduler.immediate scheduler already

tolziplohu:
> Reactor contains Scheduler.immediate scheduler already

Good idea, thank you! I thought I had tried this but apparently I hadn't, because it seems to work.

PR #4823, opened 2021-07-17
feat: add `SetDirection` event to set player orientation

ahv15:
This PR allows players to face a particular direction by making use of an UpdateDirectionEvent.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/logic/players/LocalPlayerSystem.java
+++ b/engine/src/main/java/org/terasology/engine/logic/players/LocalPlayerSystem.java
@@ -14,6 +14,7 @@ import org.terasology.engine.core.subsystem.config.BindsManager;
 import org.terasology.engine.entitySystem.entity.EntityRef;
 import org.terasology.engine.entitySystem.event.ReceiveEvent;
 import org.terasology.engine.entitySystem.systems.BaseComponentSystem;
+import org.terasology.engine.entitySystem.systems.RegisterMode;
 import org.terasology.engine.entitySystem.systems.RenderSystem;
 import org.terasology.engine.entitySystem.systems.UpdateSubscriberSystem;
 import org.terasology.engine.input.InputSystem;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/terasology/engine/logic/players/UpdateDirectionEvent.java
```diff
@@ -0,0 +1,28 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+package org.terasology.engine.logic.players;
+
+import org.terasology.engine.entitySystem.event.Event;
+import org.terasology.engine.network.OwnerEvent;
+
+@OwnerEvent
+public class UpdateDirectionEvent implements Event {
```
skaldarnar, line 9: This event should have some docstring explaining what it is used for and how to use it. Try to categorize it with the event types we mentioned before.

engine/src/main/java/org/terasology/engine/logic/players/LocalPlayerSystem.java
```diff
@@ -263,6 +257,14 @@ public void updateRotationPitch(RotationPitchAxis event, EntityRef entity) {
         event.consume();
     }
 
+    @ReceiveEvent(components = CharacterComponent.class)
+    public void updateDirection(UpdateDirectionEvent event, EntityRef entity) {
+        if (localPlayer.getCharacterEntity().equals(entity)) {
+            lookPitch = event.getPitch();
+            lookYaw = event.getYaw();
+        }
```
skaldarnar, line 265: There are event handlers for `updateRotationYaw` and `updateRotationPitch`. If we were to use the same vocabulary, this would rather be `setRotation` (as `update...` is just setting the delta to be used in the next update).

engine/src/main/java/org/terasology/engine/logic/players/UpdateDirectionEvent.java
skaldarnar: As mentioned in my other comment, maybe we should rather call this `SetDirectionEvent`?

jdrueckert (approved):

PR #4824, opened 2021-07-17
fix: collide with characters instead of walking through

jdrueckert:
Without this PR, the player can walk through other players and also NPC.
With this PR, this should no longer be possible.

Note: For some NPCs you still might be able to walk through them at least partially. Their prefabs might need some adjustment in their collider extents, to avoid that.


pollend:
I did make a fix for the call that allowed us to filter out certain entities that relied on the player collider but didn't verify it. there are some interactions that we have setup for player collision wonder if those interactions might be broken. if we project a ray for lets say hitting another player it will first hit the player colider.

jdrueckert:
> if we project a ray for lets say hitting another player it will first hit the player colider.

@pollend Why would a ray for hitting another player rely on what the CharacterMovement is configured to collide with :thinking:

pollend:
The controlling player is also a player collider

jdrueckert:
Correct, but what does a ray projection have to do with character _movement_?

pollend:
anything related to projecting a ray and using this filter: tandardCollisionGroup.CHARACTER.

jdrueckert:
@pollend I still don't get the problem... I'm only adding the character collision group as "to be collided with" for the _character movement_. So why would this change anything outside of character movement? And if it does, which kinds of character movements do you think this would negatively affect? I don't see how ray projection would be used for movement-related things :thinking: 

In addition, maybe let me ask this the other way round: What's the alternative? How can we have players and NPCs collide _without_ interfering with the interactions you're thinking about?

pollend:
I would verify these interactions.

image

jdrueckert:
@pollend I checked interactions when shooting projectiles and it's completely the same behaviour as before.
The hit box of the player is too wide, but that's the case before this PR as well, so this should be done as a follow-up. I think this PR here at hand is safe.

pollend (approved):

PR #4825, opened 2021-07-18
perf: remove extra locking in pipeline to avoid thread contention

pollend:
I removed the lock on ChunkProcessInfo and replaced it with a concurrent queue along with the locking on the individual processing objects. the problem is this can introduce a lot of contention where the two reactor threads will have contention over locking both the concurrent map and also locking the individual processing objects. I still see a problem of thrashing where we schedule a chunk processor on one thread and then schedule it again on a different thread.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
+++ b/engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
@@ -5,6 +5,7 @@ package org.terasology.engine.world.chunks.pipeline;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Queues;
 import org.joml.Vector3ic;
 import org.reactivestreams.Subscription;
 import org.slf4j.Logger;
@@ -22,6 +23,7 @@ import reactor.core.scheduler.Schedulers;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
 import java.util.Set;
 import java.util.function.Function;
 
@@ -41,16 +43,17 @@ public class ChunkProcessingPipeline {
     private final Function<Vector3ic, Chunk> chunkProvider;
     private final Map<Vector3ic, ChunkProcessingInfo> chunkProcessingInfoMap = Maps.newConcurrentMap();
     private final List<Subscription> subs = new ArrayList<>();
+
     private final Scheduler scheduler;
 
     private final Object completeSignal = new Object();
+    private Queue<ChunkProcessingInfo> processing = Queues.newConcurrentLinkedQueue();
 
     /**
      * Create ChunkProcessingPipeline.
      */
     public ChunkProcessingPipeline(Function<Vector3ic, Chunk> chunkProvider, Flux<Chunk> chunkStream) {
         this.chunkProvider = chunkProvider;
-
         scheduler = Schedulers.newParallel("chunk processing", NUM_TASK_THREADS);
         Flux<Chunk> stream = chunkStream.subscribeOn(scheduler);
         for (int i = 0; i < NUM_TASK_THREADS; i++) {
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
```diff
@@ -112,27 +115,22 @@ private void processNewChunks(List<Chunk> chunks) {
             chunkProcessingInfo.setChunk(chunk);
             chunkProcessingInfo.nextStage(stages);
             chunkProcessingInfo.makeChunkTask();
+            processing.add(chunkProcessingInfo);
             chunkProcessingInfoMap.put(position, chunkProcessingInfo);
         }
-
-        while (processChunkTasks()) {
-            continue;
-        }
```
tolziplohu, line 120: The problem with removing the loop is this situation: let's say the queue has chunk A first and then chunk B, and chunk B was the last chunk to be generated. The next task for chunk A needs chunk B to have finished its task, so when we pop that off the queue we add it to `defer` and put it back on at the end. Then chunk B runs its task, and now if we tried A again it would work, but we don't do that because it's not in a loop and there aren't any more chunks coming.

This would manifest in game as the chunk not appearing until the player moves a little bit to get new chunks to generate, but would only happen very rarely and with the farthest-away chunks, so it's not actually a very big deal.

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
```diff
@@ -141,53 +139,44 @@ private boolean processChunkTasks() {
...
                 }
 
-                // If another thread is running the task, just skip it
-                if (satisfied && chunkProcessingInfo.lock.tryLock()) {
+                if (satisfied) {
                     try {
-                        try (ThreadActivity ignored = ThreadMonitor.startThreadActivity(chunkTask.getName())) {
-                            chunkProcessingInfo.setChunk(chunkTask.apply(providedChunks));
+                        try (ThreadActivity ignored = ThreadMonitor.startThreadActivity(task.getName())) {
+                            info.setChunk(task.apply(providedChunks));
```
tolziplohu, line 145: I wonder if this can cause a race condition if the chunk is updated while another thread is getting it because it's a requirement of a different chunk. This isn't a new problem, though, it was already there, just obscured by all the locking.

tolziplohu (approved):
This is a good idea, besides being simpler and reducing lock usage it also avoids duplicate work of building requirements lists for the same chunk on more than one thread at once. I have a couple of minor concerns, but I'd be happy with merging this into the parent PR and addressing them there if necessary. Thank you!

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
pollend: only one thread has access to the task.

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
pollend: I move them to a defer list and then re-add them to the queue.

engine/src/main/java/org/terasology/engine/world/chunks/pipeline/ChunkProcessingPipeline.java
tolziplohu: No, `getChunkBy()` gets the task out of the map and gets the chunk from it. The task is in the map the whole time.

pollend: oh, yea I see, I guess you could put a lock on the chunk() object instead of locking the whole task.

keturn:
[picocli / POSIX-style command line](https://trello.com/c/0H3q2tUc/146-picocli-posix-style-command-line)

PR #4826, opened 2021-07-20
fix: replicate mesh and material over network

meetcshah19:
### Contains

Fix - Replicates mesh and material of Skeletal Mesh Component over the network.

### How to test

- Play Josharias Survival in multiplayer mode with the latest Manual Labor, with 2 players.
- Give yourself a crude shear using ```give crudeshears``` and spawn a sheep using ```spawnPrefab sheep```.
- Use shear on sheep and check if the model change is replicated for the second player

### DIFF
--- a/engine/src/main/java/org/terasology/engine/rendering/logic/SkeletalMeshComponent.java
+++ b/engine/src/main/java/org/terasology/engine/rendering/logic/SkeletalMeshComponent.java
@@ -20,7 +20,10 @@ import java.util.Map;
 
 @ForceBlockActive
 public class SkeletalMeshComponent implements VisualComponent {
+    @Replicate
     public SkeletalMesh mesh;
+
+    @Replicate
     public Material material;
 
     /**


engine/src/main/java/org/terasology/engine/rendering/logic/SkeletalMeshComponent.java
```diff
@@ -20,7 +20,10 @@
 
 @ForceBlockActive
 public class SkeletalMeshComponent implements VisualComponent {
+    @Replicate
     public SkeletalMesh mesh;
+
+    @Replicate
     public Material material;
```
pollend, line 28: I don't know about straight replicating this component.

pollend:
Not sure about the implications this has on networking. We have problems with the current system. Might exasperate there problem if we do this.

jdrueckert:
If this would be a feature, I would agree... However this is a _fix_.
If this makes problems with the networking more present, that's unfortunate and we have to try to improve on our networking, but this _is_ a desired change to enable proper multiplayer behaviour.

jdrueckert (approved):
Tests out fine :+1:

PR #4828, opened 2021-07-24
feat: restore biome-based grass and foliage tinting

tolziplohu:
### Contains

This is a rewrite of https://github.com/MovingBlocks/Terasology/pull/3802 which uses biome information instead of facet information, which is available to clients, so it won't break multiplayer. It contains multiple PRs for different modules: https://github.com/Terasology/CoreRendering/pull/68, https://github.com/Terasology/BiomesAPI/pull/17, and https://github.com/Terasology/CoreWorlds/pull/41.

### How to test

Look at grass and plants in some different biomes, and make sure the game still works, especially multiplayer.

Terasology-210723180809-1920x1080


engine/src/main/java/org/terasology/engine/rendering/assets/texture/TextureData.java
```diff
@@ -70,6 +72,16 @@ public TextureData(TextureData fromCopy) {
         }
     }
 
+    /**
+     * Look up the color of a given pixel in the base mipmap level.
+     */
+    public Colorc getPixel(int x, int y) {
```
tolziplohu, line 78: I'm not sure if we want this exposed, but the data was already public.

skaldarnar: In which sense was it public before? Could you derive the value manually just by using public methods?

engine/src/main/java/org/terasology/engine/rendering/assets/texture/TextureData.java
```diff
@@ -70,6 +72,16 @@ public TextureData(TextureData fromCopy) {
...
+        ByteBuffer baseMipmap = data[0];
+        int idx = BYTES_PER_PIXEL * (x + y * width);
+        int pixel = baseMipmap.getInt(idx);
+        return new Color(pixel);
```
skaldarnar, line 82: How often is this called (and how often do we believe will this be called)? 

We're creating a new `Color` with each call here. As `Color` is very similar to the JOML vectors we may borrow the same patterns from the math types. For them, we try to avoid creation of new instances, and instead rely on a destination parameter. This basically shifts the responsibility to create instances to the caller site - which hopefully knows better what they can cache or reuse.

tolziplohu: This is called once per block that supports tinting every time a chunk mesh is generated, so that sounds like a good idea.

engine/src/main/java/org/terasology/engine/rendering/assets/texture/TextureData.java
tolziplohu: Yes: `Texture.getData()` is public, and `TextureData.getBuffers()` is public, so this function could be implemented anywhere using only public methods. But `Texture.getData()` has a comment saying `// TODO: Remove when no longer needed`, which is the main reason I'm cautious of this.

skaldarnar (approved):

PR #4829, opened 2021-07-25
fix: resolve issue when loading a world a second time

pollend:
looks like gestalt loads a whole new environment so any assets from one session are not compatible for another session. there are two main things I do, I wipe all the prefabs and let the engine re-resolve them when the world loads and unload any assets that are not loaded in by core.


engine/src/main/java/org/terasology/engine/core/bootstrap/EnvironmentSwitchHandler.java
```diff
@@ -116,6 +124,16 @@ public void handleSwitchToGameEnvironment(Context context) {
...
         unregisterPrefabFormats(assetTypeManager);
+        assetTypeManager.getAssetTypes().forEach(k -> {
+            for (ResourceUrn urn : k.getLoadedAssetUrns()) {
+                if (!urn.getModuleName().equals(TerasologyConstants.ENGINE_MODULE)) {
+                    k.getAsset(urn).ifPresent(Asset::dispose);
+                }
+            }
+        });
+        assetTypeManager.getAssetType(BlockFamilyDefinition.class).ifPresent(AssetType::disposeAll);
+        assetTypeManager.getAssetType(Prefab.class).ifPresent(AssetType::disposeAll);
```
pollend, line 135: umm must be a better way to clean this up :-1:

engine/src/main/java/org/terasology/engine/core/modes/StateLoading.java
```diff
@@ -165,6 +166,7 @@ private void initClient() {
 
     private void initHost() {
         loadProcesses.add(new RegisterMods(context, gameManifest));
+        loadProcesses.add(new InitialiseRendering(context));
```
keturn, line 169: Does this need to be in `initClient` as well?

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/InitialiseRendering.java
```diff
@@ -0,0 +1,33 @@
...
+
+
+    @Override
+    public String getMessage() {
+        return "Initialising Rendering System...";
+    }
+
+    @Override
+    public boolean step() {
+        context.put(RenderingModuleRegistry.class, new RenderingModuleRegistry());
```
keturn, line 25: What makes this better as a LoadProcess instead of a Subsystem?

more explicit ordering of load processes? or we need a different context here than was available for any of the subsystem lifecycle phases?

(and heck it is a lot of lines of code for these two expressions, either way)

keturn (changes requested):
I think my main question here is: can we run the disposal code when we switch _out_ of the game environment? Wouldn't that be a better match to the expected lifecycle of these assets?

engine/src/main/java/org/terasology/engine/core/modes/StateLoading.java
pollend: not sure, I just managed to get this to work for single player still need to test things.

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/InitialiseRendering.java
pollend: we need a new rendering system when you restart a world.

jdrueckert: For my information: Why do we need a new rendering system when restarting a world?

engine/src/main/java/org/terasology/engine/core/bootstrap/EnvironmentSwitchHandler.java
```diff
@@ -116,6 +124,16 @@ public void handleSwitchToGameEnvironment(Context context) {
          * existing then yet.
          */
         unregisterPrefabFormats(assetTypeManager);
+        assetTypeManager.getAssetTypes().forEach(k -> {
+            for (ResourceUrn urn : k.getLoadedAssetUrns()) {
+                if (!urn.getModuleName().equals(TerasologyConstants.ENGINE_MODULE)) {
+                    k.getAsset(urn).ifPresent(Asset::dispose);
+                }
+            }
+        });
```
jdrueckert, line 133: Why do we unregister prefab formats and dispose assets _at this point_?
Shouldn't this be done when exiting the previous world rather than when entering the next? :thinking:

skaldarnar: Same question as https://github.com/MovingBlocks/Terasology/pull/4829#pullrequestreview-714350912 :upside_down_face:

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/InitialiseRendering.java
pollend: its all module drive there really isn't a reason to keep it around as a singleton.

jdrueckert: Okay

jdrueckert:
[Second World Start Bug](https://trello.com/c/iSiODJUC/145-second-world-start-bug)

keturn (approved):

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/InitialiseRendering.java
pollend: yea.. we do this for quite a few loading things.

engine/src/main/java/org/terasology/engine/core/bootstrap/EnvironmentSwitchHandler.java
pollend: we end up loading those classes again on the classpath so the prefabs are not compatible between sessions.

jdrueckert:
I tested this out and it definitely is not crashing anymore, neither when loading nor when starting a second or third world! :tada: 

_However_... I saw the following, so I assume we're still not cleaning up everything properly?
image

I was in my LaS workspace (because testing some other things before) and did the following:
1. Start a new LaS world
2. Load an older LaS world
3. Start a new Core Gameplay world

pollend:
this only broke after the tinting was added I'm sure its a couple line fix but I thought it was out of the scope of this change. @jdrueckert I was planning on looking at this separately.

jdrueckert (commented):
Tests out fine now and I don't see anything in the code that worries me. However, I'd like @keturn and / or @skaldarnar to have a final look, too, so I won't approve this yet.
@keturn, @skaldarnar feel free to approve and merge after you've had a look in case you're good with the changes.

engine/src/main/java/org/terasology/engine/core/bootstrap/EnvironmentSwitchHandler.java
```diff
@@ -99,6 +99,7 @@ public void handleSwitchToGameEnvironment(Context context) {
         context.put(EventLibrary.class, library.getEventLibrary());
         context.put(ClassMetaLibrary.class, new ClassMetaLibraryImpl(context));
 
+
```
skaldarnar, line 102:
```suggestion
```

engine/src/main/java/org/terasology/engine/core/modes/StateIngame.java
```diff
@@ -101,8 +109,20 @@ public Boolean get() {
     @Override
     public void dispose(boolean shuttingDown) {
         ChunkProvider chunkProvider = context.get(ChunkProvider.class);
+        AssetTypeManager assetTypeManager = context.get(ModuleAwareAssetTypeManager.class);
         chunkProvider.dispose();
```
skaldarnar, line 114: For readability, let's do this in blocks, i.e., first we dispose `chunkProvider`, then `assetTypeManager`, ...

```suggestion
        chunkProvider.dispose();

        AssetTypeManager assetTypeManager = context.get(ModuleAwareAssetTypeManager.class);
```

Alternatively, separate the context retrieval from the dispose actions with a new line.

engine/src/main/java/org/terasology/engine/core/modes/StateIngame.java
```diff
@@ -101,8 +109,20 @@ public Boolean get() {
...
 
+        assetTypeManager.getAssetTypes().forEach(k -> {
```
skaldarnar, line 115: What is `k` supposed to mean here? Can we please name it `assetType`?

engine/src/main/java/org/terasology/engine/core/modes/StateIngame.java
```diff
@@ -101,8 +109,20 @@ public Boolean get() {
...
+            for (ResourceUrn urn : k.getLoadedAssetUrns()) {
+                if (!urn.getModuleName().equals(TerasologyConstants.ENGINE_MODULE)) {
+                    k.getAsset(urn).ifPresent(Asset::dispose);
+                }
+            }
+        });
+        assetTypeManager.getAssetType(BlockFamilyDefinition.class).ifPresent(AssetType::disposeAll);
+        assetTypeManager.getAssetType(Prefab.class).ifPresent(AssetType::disposeAll);
```
skaldarnar, line 123: Please add a dosctring why we need to dispose these separately.

I guess this is because they are part of the engine module and are thus skipped by the statement above (https://github.com/MovingBlocks/Terasology/pull/4829/files#diff-b246f2c1e7c1a8f480631ab231b8a25b8b1b7bb5074d6b5b6ed3c6e01e44e0bcR117).

I like to have the info here about
- why do this in a separate statement
- why those two classes/how did we figure that out

engine/src/main/java/org/terasology/engine/core/modes/StateIngame.java
```diff
@@ -101,8 +109,20 @@ public Boolean get() {
...
+
+
```
skaldarnar, line 125: Please avoid more than one empty line
```suggestion

```

engine/src/main/java/org/terasology/engine/core/modes/StateIngame.java
```diff
@@ -141,6 +161,7 @@ public void dispose(boolean shuttingDown) {
         console.dispose();
         GameThread.clearWaitingProcesses();
 
+
```
skaldarnar, line 164: Please avoid more than one empty line
```suggestion
```

engine/src/main/java/org/terasology/engine/world/block/DefaultColorSource.java
```diff
@@ -17,29 +17,23 @@ public Colorc calcColor(int x, int y, int z) {
     COLOR_LUT {
         @Override
         public Colorc calcColor(int x, int y, int z) {
+            ColorProvider colorProvider = CoreRegistry.get(ColorProvider.class);
```
skaldarnar, line 20: Is doing this lookup in `CoreRegistry` fast? I think @tolziplohu put this optimization on purpose to reduce the necessary lookups in the registry with each lookup of a color... :thinking:

engine/src/main/java/org/terasology/engine/world/block/loader/BlockFamilyDefinitionFormat.java
```diff
@@ -318,7 +318,6 @@ private BlockFamilyDefinitionData createBaseData(JsonObject jsonObject) {
         @Override
         public Class<? extends BlockFamily> deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
                 throws JsonParseException {
-
```
skaldarnar, line 321: All these changes are unrelated to this PR. Please revert them.

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/InitialiseRendering.java
skaldarnar: Can we please persist this information in a docstring. Probably the class doc here is a good place to capture this.

skaldarnar (changes requested):

engine/src/main/java/org/terasology/engine/world/block/DefaultColorSource.java
pollend: yea, but we can't save it as a static variable.

engine/src/main/java/org/terasology/engine/core/modes/loadProcesses/InitialiseRendering.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.terasology.engine.core.modes.loadProcesses;
+
+import org.terasology.engine.context.Context;
+import org.terasology.engine.core.modes.SingleStepLoadProcess;
+import org.terasology.engine.core.module.rendering.RenderingModuleRegistry;
+
+/**
+ * Add {@link RenderingModuleRegistry} to the game {@link Context}.
+ */
```
skaldarnar, line 12: @pollend does this description sound right to you?

```suggestion
/**
 * Add {@link RenderingModuleRegistry} to the game {@link Context}.
 * 
 * The rendering system is required whenever a client starts or joins a game. As rendering may fail to re-initialise
 * correctly when it has previously been constructed, this loading process will populate the {@link Context} with a
 * freshly created rendering system.
 * 
 * When switching the game state, the rendering system can just be disposed with the old state.
 */
```

engine/src/main/java/org/terasology/engine/core/modes/StateIngame.java
```diff
@@ -103,6 +111,16 @@ public void dispose(boolean shuttingDown) {
...
         chunkProvider.dispose();
 
+        AssetTypeManager assetTypeManager = context.get(ModuleAwareAssetTypeManager.class);
+        assetTypeManager.getAssetTypes().forEach(k -> {
+            for (ResourceUrn urn : k.getLoadedAssetUrns()) {
+                if (!urn.getModuleName().equals(TerasologyConstants.ENGINE_MODULE)) {
+                    k.getAsset(urn).ifPresent(Asset::dispose);
+                }
+            }
+        });
```
skaldarnar, line 121:
```suggestion
        assetTypeManager.getAssetTypes().forEach(assetType -> {
            for (ResourceUrn urn : assetType.getLoadedAssetUrns()) {
                if (!urn.getModuleName().equals(TerasologyConstants.ENGINE_MODULE)) {
                    assetType.getAsset(urn).ifPresent(Asset::dispose);
                }
            }
        });
```

engine/src/main/java/org/terasology/engine/core/modes/StateIngame.java
skaldarnar:
```suggestion
        // dispose all module assets
        assetTypeManager.getAssetTypes().forEach(k -> {
```

engine/src/main/java/org/terasology/engine/core/modes/StateIngame.java
skaldarnar:
```suggestion
        // dispose engine assets that should not be kept when switching game states
        assetTypeManager.getAssetType(BlockFamilyDefinition.class).ifPresent(AssetType::disposeAll);
        assetTypeManager.getAssetType(Prefab.class).ifPresent(AssetType::disposeAll);

```

skaldarnar (approved):

PR #4830, opened 2021-07-26
test: deprecate WorldProvidingHeadlessEnvironment

keturn:
- Make TerasologyTestingEnvironment internal. It's still widely used in the engine tests, but we don't need to export it to modules.
- Deprecate WorldProvidingHeadlessEnvironment. (There is still some utility code using it in Behaviors and Pathfinding, so we can't yet delete it.)
  - That means Environment and HeadlessEnvironment can't be removed from the package, but we can add a note directing module maintainers to MTE.


keturn:
Note that _most_ of the usages seen in https://github.com/search?q=org%3ATerasology+WorldProvidingHeadlessEnvironment&type=code are addressed by https://github.com/Terasology/Pathfinding/pull/64

pollend (approved):

PR #4833, opened 2021-07-27
feat: add priority to Updates annotation of facet providers

tolziplohu:
### Contains

`WorldBuilder` comes up with the order to run facet providers in based on their annotations - `@Requires`, `@Produces`, and `@Updates`. But right now, there's no way to control order between `@Updates` providers - for example, a river provider should update the elevation after a mountain provider, so that rivers are carved into mountains and not the other way around. This PR adds a mechanism to do that, another field on `@Updates(value = ..., priority = UpdatePriority.PRIORITY_HIGH)` with similar effect to the existing event handler priority mechanism. This also ended up simplifying the provider ordering code.

There's now a circular dependency in `CoreWorlds`, since `SpawnPlateauProvider` requires `SurfaceRoughnessFacet` and `SimplexRoughnessProvider` requires `ElevationFacet`, which can be resolved by explicitly giving `SpawnPlateauProvider` a lower priority - there's even a nice error message that suggests the correct change. Required changes to other world generators should be similarly minor and straightforward.


engine/src/main/java/org/terasology/engine/world/generation/UpdatePriority.java
```diff
@@ -0,0 +1,43 @@
...
+ * Some preset priority values for Updates.priority.
+ */
+public final class UpdatePriority {
+    public static final int PRIORITY_PRODUCES = 250;
+    public static final int PRIORITY_CRITICAL = 200;
+    public static final int PRIORITY_HIGH = 150;
+    public static final int PRIORITY_NORMAL = 100;
+    public static final int PRIORITY_LOW = 50;
+    public static final int PRIORITY_TRIVIAL = 0;
+    public static final int PRIORITY_REQUIRES = -50;
```
pollend, line 16: why not just use an enum?

tolziplohu: I'm just doing the same thing as [`EventPriority`](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/entitySystem/event/EventPriority.java), I'm not sure why that's done that way.

skaldarnar: I'm pretty sure we're not using an enum for event priority because the main goal here is to resolve non-determinism due to same priority. In many cases these default threshold already help to sort out whether something should run with low or high priority, but we may need the fine-grained support to run a `NORMAL + 10` (slightly earlier than the rest of normal prio, but after high prio). 

Given that, I'm wondering whether the defaults for PRODUCES and REQUIRES are good as they are here, or whether they should be way higher/lower (maybe even Integer.MIN` and `Integer.MAX`).

engine/src/main/java/org/terasology/engine/world/generation/WorldBuilder.java
```diff
@@ -200,92 +209,68 @@ public World build() {
...
-                for (FacetProvider fp : providersList) {
-                    // only add @Produces providers to avoid infinite recursion
-                    if (producesFacet(fp, updated.value())) {
-                        if (!scalable || fp instanceof ScalableFacetProvider) {
-                            localOrderedProviders.add(fp);
-                        } else {
-                            return;
-                        }
+        Stream.of(updatedFacets(provider), requiredFacets(provider)).flatMap(Arrays::stream).forEachOrdered(r -> {
+            if (r.value() != providedFacet) {
```
skaldarnar, line 245: Is that one big `if` around the rest, with no `else`? If so, I'd move this out of the `forEach` and use a `.filter(r -> r.value() == providedFacet)` for that.

engine/src/main/java/org/terasology/engine/world/generation/UpdatePriority.java
tolziplohu: > Given that, I'm wondering whether the defaults for PRODUCES and REQUIRES are good as they are here, or whether they should be way higher/lower (maybe even Integer.MINandInteger.MAX`).

I think that makes sense for `PRODUCES`, since it's impossible for anything to update the facet before it's produced. But some things actually want to run after `REQUIRES`, specifically the `SpawnPlateauProvider` from `CoreWorlds`, although those conflicts can often also be fixed by changing the `Requires` to `Updates` on other providers  (`SimplexRoughnessProvider` in this case would pretend to update `ElevationFacet`).

skaldarnar: That totally makes sense to me. Is it possible to annotate the priority in `@Requires` as well?

tolziplohu: No, the idea is that if you need fine-grained control over when something reads the facet, you should just use `@Updates`. Of course, that is somewhat inconsistent with being able to `@Update` after `@Require`-ing like I did with the `SpawnPlateauProvider`, but I'm not sure what the best solution would be - it feels like any way of doing this is bad for modularity.

engine/src/main/java/org/terasology/engine/world/generation/UpdatePriority.java
```diff
@@ -0,0 +1,43 @@
...
+                return "PRIORITY_HIGH";
+            case PRIORITY_NORMAL:
+                return "PRIORITY_NORMAL";
+            case PRIORITY_LOW:
+                return "PRIORITY_LOW";
+            case PRIORITY_REQUIRES:
+                return "PRIORITY_REQUIRES";
+            default:
+                return Integer.toString(priority);
+        }
```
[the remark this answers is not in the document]
pollend: you can use them like ranges

if priority < a
else if priority < b
else if priority < c
else if priority < d
else if priority < e

PRIORITY_<type>(value)

engine/src/main/java/org/terasology/engine/world/generation/WorldBuilder.java
```diff
@@ -200,92 +209,69 @@ public World build() {
...
-            return;
-        }
-        if (!facetCalculationInProgress.add(facet)) {
-            throw new RuntimeException("Circular dependency detected when calculating facet provider ordering for " + facet);
-        }
-        Set<FacetProvider> orderedProviders = Sets.newLinkedHashSet();
-
-        // first add all @Produces facet providers
+    private void addProviderChain(Class<? extends WorldFacet> facet, boolean scalable, int minPriority,
+                                  Set<FacetProvider> orderedProviders) {
```
skaldarnar, line 213: Could you please provide some Javadoc what this method is doing (could also be a general follow-up PR to add docstrings if you prefer). I few things I noticed from just looking at the PR

- `orderedProviders` is a "destination" parameter and is populated in this method
- `providedBy` is updated
- there is a "hidden" recursion via `addRequirements`

skaldarnar (dismissed):

skaldarnar (approved):

PR #4836, opened 2021-08-02
perf: rework with proto3 reworked packing network packets split net messages

pollend:
This is a continuation of these changes: https://github.com/MovingBlocks/Terasology/pull/4819

I split the net packet into individual net messages. I send the messages immediately rather then waiting on the network tick. I could push everything to a queue on the other side and process everything at once with the tick of the heartbeat if this needs to work the same as it was previously. Still need to look into how we are doing lag compensation but I think this should better manage the latency since we are not packing everything into large packets. on the network tick I flush the the channel instead of doing this for every packet.


pollend:
you can use this to simulate network latency on linux just make sure to remove the rule from your interface

for local it should be your loopback interface which is lo.
```
sudo tc qdisc add dev <network-interface> root netem delay 100ms rate 50mbit

sudo tc qdisc delete dev <network-interface> root

```

jdrueckert:
[Protobuf v2 -> v3](https://trello.com/c/zxHzMyzg/169-protobuf-v2-v3)

PR #4839, opened 2021-08-10
refactor(GameScheduler): provide ThreadMonitor-compatible schedule method

keturn:
This is a proposal in hopes of moving us past the controversial point on #4799. It targets that PR's branch.

This is not a _rejection_ of micrometer. Using micrometer instead of some of our homegrown meters is a welcome plan! But I don't think we need to do it in that PR.


engine/src/main/java/org/terasology/engine/rendering/opengl/ScreenGrabber.java
```diff
@@ -101,13 +101,15 @@ public void saveScreenshot() {
...
         if (savingGamePreview) {
-            GameScheduler.parallel().schedule(() -> saveGamePreviewTask(buffer, width, height));
+            task = () -> saveGamePreviewTask(buffer, width, height);
             this.savingGamePreview = false;
         } else {
-            GameScheduler.parallel().schedule(() -> saveScreenshotTask(buffer, width, height));
+            task = () -> saveScreenshotTask(buffer, width, height);
         }
 
+        GameScheduler.scheduleParallel("Write screenshot", task);
```
pollend, line 112: would probably still opt to just use two calls to `scheduleParallel`. no need to be clever.

pollend (approved):

PR #4840, opened 2021-08-12
fix: resolve Advance World Settings issue

pollend:
I reverted a part of the change introduce in this PR: https://github.com/MovingBlocks/Terasology/pull/4829

something is strange about how the context is configured.


skaldarnar (approved):
Just went through the AGS for Metal Renegades with this :+1:

PR #4841, opened 2021-08-13
fix: Better handling of ComponentSystem loading with optional dependencies

DarkWeird:
### Contains

* Fixes hard crash when System using optional dependency's class in method signature
* Better handling error when system crashes at registering
* ahead detection for using optional dependency  at system 

### How to test

Start JS with optional dependencies:
1. without
2. with Genome
3. with BasicCrafting

### DIFF
--- a/engine/src/main/java/org/terasology/engine/core/ComponentSystemManager.java
+++ b/engine/src/main/java/org/terasology/engine/core/ComponentSystemManager.java
@@ -117,7 +117,7 @@ public class ComponentSystemManager {
      * @param id id of system.
      */
     private void rollbackLoading(ComponentSystem system, String id) {
-        if (system != null  // system creates, needs to cleanup.
+        if (system != null  // system was created, needs cleanup.
                 && store.remove(namedLookup.remove(id))// remove system from lookup and store, if already registered.
         ) {
             //TODO unshare if system is shared by InjectionHelper#share()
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/terasology/engine/core/ComponentSystemManager.java
```diff
@@ -82,19 +82,78 @@ public void loadSystems(ModuleEnvironment environment, NetworkMode netMode) {
...
+        }
+    }
+
+    /**
+     * Rollback changes which made loading before fail.
+     * @param system system to rollback.
+     * @param id id of system.
+     */
+    private void rollbackLoading(ComponentSystem system, String id) {
+        if (system != null  // system creates, needs to cleanup.
```
skaldarnar, line 120:
```suggestion
        if (system != null  // system was created, needs cleanup.
```

skaldarnar (commented):
Ah, I see - if we cannot meet all requirements a system has we skip it - so if we have one big system depending on 5 optional modules it won't load if any of those is missing.

To improve the situation, we split systems such that they have high cohesion w.r.t. to the modules they (optionally) depend on.

Will test later today, but the changes look reasonable to me and I think make it better understandable, even if not a perfect solution for the underlying problem...

skaldarnar (approved):
This seems to work nicely, tested with Simple Farming in Metal Renegades.

PR #4842, opened 2021-08-14
chore: adjust default logback configuration

skaldarnar:
Only apply the `logOverrideLevel` to Terasology, but not 3rd party packages we explicitly set the log level for.

Additionally, adds a note how to (temporarily) adjust the configuration to tweak the log output for local testing.


keturn (approved):
👍

PR #4843, opened 2021-08-14
feat: hide NotificationOverlay (chat box) if HUD is hidden

skaldarnar:
While testing some PRs with Metal Renegades I noticed that the chat box (aka notification overlay) is not affected by the "Hide HUD" toggle when hitting <kbd>h</kbd>.

This becomes apparent if someone or something is constantly emitting message to that notification area, which is currently the case with Climate Conditions. 

This is by no means a "fix" for Climate Conditions being quite talkative, but I think that this UI element belongs to the HUD and should therefore be shown/hidden as other HUD elements.


engine/src/main/java/org/terasology/engine/logic/console/ui/NotificationOverlay.java
```diff
@@ -75,6 +82,18 @@ public String get() {
...
+        //FIXME? To bind the 'visible' property which already exists on this screen layer we need to create
+        //       a new binding. To avoid unnecessary lookup calls we use a default binding which we modify
+        //       imperatively based on the property change events we receive from the rendering config.
+        Binding<Boolean> visible = new DefaultBinding<>(true);
+        this.bindVisible(visible);
+        //FIXME? The debug rendering config is properly informed about hiding the HUD, whereas the
+        //       corresponding method on the NUI manager 'NUIManager#isHUDVisible' does not work properly.
+        config.getRendering().getDebug().subscribe(RenderingDebugConfig.HUD_HIDDEN, evt -> {
+            visible.set(!(Boolean) evt.getNewValue());
+        });
```
pollend, line 96: confusing that this is from some kind of debug?

skaldarnar: not really - the only system reacting on that key press is actually the debug system, where "debug system" is a system handling different in-game options for debugging, like rendering chunk bounding boxes.

It's just not propagated anywhere else, and I'm not really sure what the NUI manager property is supposed to express.

pollend (approved):
can you make a ticket describing how to improve this?

skaldarnar:
@pollend here you go: https://github.com/MovingBlocks/Terasology/issues/4866

Issue #4699, opened 2021-05-21
Crash reporter fails to upload the logs


Hume2:
### What you were trying to do
When the game crashes, the crash reporter appears. It asks me to choose where I want to upload the logs. None of Pastebin or Google Drive works.

### What actually happened
When I click Pastebin, it throws an error message saying: Failed to generate paste:

When I click Google Drive, it throws an error message saying: https://jenkins.terasology.org/userContent/GooeyDrive-24506b2f1f34.p12

Sometimes the crash reporter just closes without any apparent reason. I assume, it just crashes.

### How to reproduce

* Step 1: Make Terasology crash
* Step 2: Answer the dialogs until you get to the screen asking for uploading the logs.
* Step 3: Click Pastebin or Google Drive. In both cases, an error message appears.

### Log details and game version
I tested it in versions preview alpha-20+195 (20210519) and release alpha-19+2. It happens in both versions.
I think, the crash reporter is not logged, is it?

### Computer details
OS: Ubuntu 18.04

skaldarnar:
Nargh, a crashing https://github.com/MovingBlocks/CrashReporter is indeed bad 🧐 Crash Reporter is a separate project that hasn't changed for quite some time, so we bundle the same version with our recent releases (thus, no difference between version).

@Cervator the issue about Google Drive seems to be caused by the shutdown of the Jenkins instance at `jenkins.terasology.org`. Maybe we can just move over some configuration and bend over the reference in the Crash Reporter?

First issue about Pastebin might be changed API or changed paths of logs or something like that. We need to investigate that.

skaldarnar:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix CrashReporter](https://trello.com/c/4arhlQws/117-fix-crashreporter)

Cervator:
I'm pretty puzzled why the GDrive option would have anything to do with the old and now gone Jenkins, but sure, that'd do it. I didn't code that piece but surely we must be able to redo it real quick.

Pastebin likewise and at the same time, odds are the API changed ages ago and we didn't update. Looks like https://pastebin.com/doc_api would cover it and I have the credential for our Pro account handy if needed :-)

skaldarnar:
[...]

PR #4844, opened 2021-08-15
chore: add maven repo that contains jpastebin

jdrueckert:
This change is required to allow the CrashReporter to import the [`org.jpastebin`](https://mvnrepository.com/artifact/org/jpastebin/1.0.1) dependency which is required to fix the broken CrashReporter integration with PasteBin.

See issue #4699 
See fix https://github.com/MovingBlocks/CrashReporter/pull/44

### DIFF
--- a/build-logic/src/main/kotlin/terasology-repositories.gradle.kts
+++ b/build-logic/src/main/kotlin/terasology-repositories.gradle.kts
@@ -22,6 +22,11 @@ repositories {
         }
     }
 
+    maven {
+        name = "JBoss Public Maven Repository Group"
+        url = URI("https://repository.jboss.org/nexus/content/repositories/public/")
+    }
+
     // MovingBlocks Artifactory instance for libs not readily available elsewhere plus our own libs
     maven {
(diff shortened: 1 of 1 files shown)


jdrueckert:
[Fix CrashReporter](https://trello.com/c/4arhlQws/117-fix-crashreporter)

keturn (changes requested):
I'm reluctant to add to random repositories, but after looking this one up, it seems to me like it would be okay. i.e. it is maintained by someone who is going to keep it online, probably trustworthy to not let random hacks get added to it, and it is intentionally public.

Suggestions:

* add a comment that mentions this is needed for a crashreporter dependency
  + link to https://developer.jboss.org/docs/DOC-11377
* add the filter (as seen in the Maven Central repo) that makes sure we don't waste time looking for `org.terasology` dependencies in this repository.
  + copy/pasting that twice (maven & google) is maybe okay. Three gets to be noisy. It would be nice to figure out a better way to configure that.
  + alternately, it might be better in this case to filter for _only_ `org:jpastebin`

keturn (approved):
That content filter looks right. 👍

PR #4847, opened 2021-08-16
fix: enable block tinting in LOD chunks

tolziplohu:
### Contains

Block tinting previously queried biome information from the `WorldProvider`. But LOD chunks are never added to the world, just meshed and rendered. All the necessary data is still there, so this switches to passing block tinting systems a `ChunkView` from which to derive necessary information, which really makes more sense anyway.

### How to test

With https://github.com/Terasology/BiomesAPI/pull/18, try enabling LOD chunks in Core Gameplay or Metal Renegades and check that grass and leaves change color in different biomes in LOD chunks.

Here's an example, although it might be hard to see without zooming in:
Terasology-210815201323-1920x1080


keturn:
Oh, I see. https://github.com/MovingBlocks/Terasology/blob/3ae1a395244683de743ce7f1b5e3a5e50bda744a/engine/src/main/java/org/terasology/engine/world/internal/WorldProviderCoreImpl.java#L316-L319

So it's not that asking the WorldProvider and a ChunkView return different things for the same location. The issue is that the ChunkProvider doesn't have that chunk handy, because the ChunkView you have is some approximation for a Chunk that isn't really loaded.

The code that ends up using this is already operating on a ChunkView, so introducing ChunkView objects in to that API isn't a problem in practice.

I am troubled by the fact that it leaves us with multiple ways to try to query this data and won't be obvious to people which to use.

I'm not sure what to make of that. Should BiomeManager _always_ operate on ChunkViews instead of the global WorldProvider?

Should WorldProvider not implement the ChunkView interface, to make sure those queries go to a ChunkView instead?

keturn (approved):
Tested this out, looks like it is working here.

Still would like to sort out the WorldProvider/ChunkView question, but this PR doesn't have to wait for that.

PR #4852, opened 2021-08-17
feat: Add parallel module get

DarkWeird:
### Contains

* reimplement algorithm. gather dependencies before git cloning.
* parallel getting at x1.5-x4 faster then sequence getting

### How to test

Use `time ./gw module init omega` (at Linux)
or use stopwatch for windows.


pollend:
wow that is much faster to clone :D. gj @DarkWeird

PR #4853, opened 2021-08-18
bugfix: move Physics and registered system and add debug overlay 

pollend:
This is not a clean implementation of the physics system but this moves the setup into a component system. This is in the right direction but the way this is split is not entirely ideal. I render the bounding box but would prefer to render the actual collider but this should suffice enough for @skaldarnar.

change this in `config.cfg`
```
...
    "debug": {
      "enabled": false,
      "firstPersonElementsHidden": false,
      "hudHidden": false,
      "wireframe": false,
      "renderChunkBoundingBoxes": false,
      "renderEntityColliders": true,
      "renderSkeletons": false
    }
...
```

image

PR #4855, opened 2021-08-18
feat: add debug bounding box rendering

pollend:
I render the bounding box for the physics object but would prefer to render the actual collider but this should suffice enough for our needs. @skaldarnar I think you wanted to use this to debug metal renegades with this?

change this in `config.cfg`
```
...
    "debug": {
      "enabled": false,
      "firstPersonElementsHidden": false,
      "hudHidden": false,
      "wireframe": false,
      "renderChunkBoundingBoxes": false,
      "renderEntityBoundingBoxes": true,
      "renderSkeletons": false
    }
...
```

image


engine/src/main/java/org/terasology/engine/physics/engine/PhysicsSystem.java
```diff
@@ -164,6 +195,125 @@ public void onItemImpact(ImpactEvent event, EntityRef entity) {
...
+            meshData.indices.rewind();
+            meshData.position.rewind();
+            meshData.color0.rewind();
+
+            Vector3f worldPosition = new Vector3f();
+            Quaternionf worldRot = new Quaternionf();
+            Matrix4f transform = new Matrix4f();
+            AABBf bounds = new AABBf(0, 0, 0, 0, 0, 0);
+
+            for (EntityRef entity : entityManager.getEntitiesWith(RigidBodyComponent.class)) {
```
skaldarnar, line 270: Not all the entities have a `RigidBodyComponent` - should this rather be for `LocationComponent`?

skaldarnar (commented):
As already asked on Discord, it'd be nice to be able to see the bounding boxes for both the different _shape components_ and the actual _collider_ (used for collision and ray casting to determine the target entity).

I'll also add a bunch of debug commands to toggle these things via the in-game console in addition to manually editing the config.

pollend:
need a better way to abstract these debug shapes. 

> As already asked on Discord, it'd be nice to be able to see the bounding boxes for both the different _shape components_ and the actual _collider_ (used for collision and ray casting to determine the target entity).
> 
> I'll also add a bunch of debug commands to toggle these things via the in-game console in addition to manually editing the config.

skaldarnar (commented):
On a second thought, why did you add this to `PyhsicsSystem`?

jdrueckert:
That's pretty cool @pollend !
For fixing the issues we have with the models, though, the collision box would be even more helpful :slightly_smiling_face:

engine/src/main/java/org/terasology/engine/physics/engine/PhysicsSystem.java
```diff
@@ -66,8 +65,6 @@
     @In
     private NetworkSystem networkSystem;
     @In
-    private EntityManager entityManager;
```
skaldarnar, line 69: This was seemingly not used.

engine/src/main/java/org/terasology/engine/config/RenderingDebugConfig.java
```diff
@@ -19,6 +19,8 @@
     public static final String HUD_HIDDEN = "hudHidden";
     public static final String RENDER_CHUNK_BOUNDING_BOXES = "renderChunkBoundingBoxes";
     public static final String RENDER_SKELETONS = "renderSkeletons";
+    public static final String RENDER_ENTITY_BOUNDING_BOXES = "renderEntityBoundingBoxes";
```
skaldarnar, line 22: I renamed this since we're not rendering the entity colliders, but the bounding boxes as described by some component. Those two don't necessarily align, unfortunately.

skaldarnar (dismissed):

engine/src/main/java/org/terasology/engine/physics/engine/PhysicsSystem.java
pollend: this whole class is dropped in here: https://github.com/MovingBlocks/Terasology/pull/4853

skaldarnar: Even better =) I'll drop this change from this PR then to avoid any conflicts :+1:

skaldarnar (approved):

PR #4857, opened 2021-08-18
feat: add commands for visual debugging (bounding boxes, etc.)

skaldarnar:
### Contains

This PR adds a couple of _debug commands_ to toggle debug rendering features such as bounding boxes (entities, chunks) or skeletal meshes. 

All commands are of the form `debug:<settings-name>` to introduce a common, consistent pattern and simplify correlation with the actual settings entry.

### How to test

- `debug:renderEntityBoundingBoxes` entities such as sheep or deer (`spawnPrefab deer` in JS or MR) should be rendered with a white wireframe bounding box. 
- `debug:renderChunkBoundingBoxe` chunk bounding boxes should be rendered in a black wireframe mode (can also be toggled via keyboard shortcut)
- `debug:renderWireframe` the whole world is rendered in wireframe mode (can also be toggled via keyboard shortcut)
- `debug:renderSkeletons` the skeleton (animation rig) of entites should be rendered as white outline. Easy to test alongside the bounding box rendering for entities


pollend (approved):

pollend:
maybe we should just have all extra debugging driven this way. we use all our F keys for debug.

PR #4858, opened 2021-08-19
fix: render (partially) transparent blocks at any distance

tolziplohu:
### Contains

Transparent blocks like leaves and grass are usually invisible beyond a certain distance. This is actually because the game will only render 64 chunks that have transparent blocks in them. At this point there doesn't seem to be any performance reason to do that, and bare trees are annoying, so this PR gets rid of the limit.

### How to test

Look at some trees in the distance, and make sure they still have leaves. I've been testing in Metal Renegades, which might be easier because you can see trees further away.

### DIFF
--- a/engine/src/main/java/org/terasology/engine/rendering/world/RenderableWorldImpl.java
+++ b/engine/src/main/java/org/terasology/engine/rendering/world/RenderableWorldImpl.java
@@ -46,7 +46,6 @@ import java.util.PriorityQueue;
 class RenderableWorldImpl implements RenderableWorld {
 
     private static final int MAX_ANIMATED_CHUNKS = 64;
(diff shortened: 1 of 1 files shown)


jdrueckert:
@tolziplohu there's a few codestyle annotations (in the "Files changed" tab) you might want to check :slightly_smiling_face:

jdrueckert:
For some reason I was not able to reproduce the leaves not being rendered with the limit set :thinking: 
Did I misunderstand the PR description? I tried with limit 32 and view distance "Extreme".

tolziplohu:
> For some reason I was not able to reproduce the leaves not being rendered with the limit set.

Oh, it turns out that only PlantPack has partially transparent leaves; the leaves in CoreAssets are fully opaque. You can try with Metal Renegades, or just look at grass or flowers.

PR #4859, opened 2021-08-20
build(facade): oops, junit was supposed to be a test-only dependency

keturn:
Amendment to #4147. Fixes an overly-broad JUnit dependency and adds some docs.

### DIFF
--- a/facades/PC/src/main/java/org/terasology/engine/DataSizeConverter.java
+++ b/facades/PC/src/main/java/org/terasology/engine/DataSizeConverter.java
@@ -10,21 +10,34 @@ import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+/**
+ * Convert a string describing a size in bytes.
+ * <p>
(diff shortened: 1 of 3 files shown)


keturn:
[stricter limits for memory use on Linux](https://trello.com/c/m8IsoQan/159-stricter-limits-for-memory-use-on-linux)

facades/PC/src/main/java/org/terasology/engine/DataSizeConverter.java
```diff
@@ -10,21 +10,34 @@
...
     enum Unit {
         B(1),
-        K(1 << 10),
-        M(1 << 20),
-        G(1 << 30),
-        T(1L << 40);
+        K(1 << 10),  // kilo
+        M(1 << 20),  // mega
+        G(1 << 30),  // giga
+        T(1L << 40); // tera
```
pollend, line 41: wonder if this could fit somewhere else?

pollend (approved):

PR #4870, opened 2021-08-28
chore: some lintings and suggested simplifications

keturn:
Addressing some of the lint bits that came up in #4786. This is just clearing the way to have fewer distractions in that PR.


engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
```diff
@@ -25,7 +24,6 @@
 /**
  * Chunk meshes store, manipulate and render the vertex data of tessellated chunks.
  */
-@SuppressWarnings("PointlessArithmeticExpression")
```
skaldarnar, line 28: Always a good idea to have pointless expressions :sweat_smile:

skaldarnar (approved):

PR #4878, opened 2021-08-31
chore: remove internal ai system for engine

pollend:
This code is only used within the portal module I think we have other things that can easily replace this logic. need to review the impact and see what can be changed with the portal module. probably have something that can replicate the behavior we are looking for or potentially just shoveling this code into Spwaning?


skaldarnar:
I'm always a fan of removing code :upside_down_face: 

For completeness, I'm pinging 
- @Cervator as main author of Portals (I think)
- @casals as main driving force of AI topics
- (@jdrueckert @keturn as you may have looked at Spawning over the summer)

casals:
Not sure if it's only a matter of transferring it to Spawning but yes, absolutely. As long as the dependencies are solved (just checking a full distribution should be enough), aye.

pollend:
I would opt to move Portals and Spawnable to archive and maybe salvage the code for something else. @skaldarnar

PR #4879, opened 2021-08-31
refactor: do light merging later, remove ChunkProcessingPipeline

naalit:
### Contains

Most of the complexity around `ChunkProcessingPipeline` is because of the need to support chunk tasks which require neighboring chunks. The only task which requires that is light merging, so this PR experiments with moving light merging to after the main chunk processing phase. This makes it so that chunks appear in the world immediately after their generated, but will have their lighting updated when neighboring chunks are generated, instead of waiting for the neighbors to generate to appear at all. It also means `ChunkProcessingPipeline` can be removed entirely, and replaced with fairly simple usage of the Reactor APIs.

### How to test

Try generating a world, loading a world, playing in multiplayer. My multiplayer doesn't really seem to work right now anyway so I haven't tested whether this PR breaks it.

Is there any reason besides light merging to support chunk tasks with multiple requirements? If so, the complexity *might* be worth it.


DarkWeird:
It is incorrect. (Parallel processing incorrect too. But i did it)

Main pros:
1. Light is logical value, which can be used by game logic. Like spawn mobs in the dark.

I always try to do this :(

keturn:
@tolziplohu I know you had [some questions about this](https://discord.com/channels/270264625419911192/702893102481932308/883474730517147728). What should we do with this PR? Put it back in Draft status?

tolziplohu:
> @tolziplohu I know you had [some questions about this](https://discord.com/channels/270264625419911192/702893102481932308/883474730517147728). What should we do with this PR? Put it back in Draft status?

It sounds like if chunks aren't marked ready until light merging happens, that would probably solve the problems @DarkWeird identified. I don't much time currently though, so I'll mark it as a draft for now.

soloturn:
Opened #5363, which ports this idea to current `develop`.

This PR itself can't be rebased — it targets `feat/chunk-ordering-reactor` (#4822), which is unmerged and predates the `org.terasology.engine.*` package restructure (#5192), so it conflicts on its first commit. #5363 is therefore a fresh implementation with no code carried over, and credits @naalit as co-author since the design is theirs: let a chunk go ready on its own and merge light incrementally afterwards.

Two differences worth noting, both found while porting:

- This PR merges on `GameScheduler.scheduleParallel`. After the change the merge touches chunks that are live rather than chunks only the pipeline can see, and deflated light storage can reallocate on write — so #5363 runs it on the main thread under a time budget, matching how `WorldProviderCoreImpl.processPropagation` already handles runtime light changes.
- Neither `LightMerger.merge` nor `LocalChunkView.setValueAt` marks anything dirty. That was harmless while merging ran before the chunk was ready, but once the chunk is already visible the merged light never reaches a mesh — `ChunkMeshWorker` only re-meshes chunks that are `isReady() && isDirty()`. #5363 marks all 27 chunks dirty after merging.

Leaving this open for now rather than closing it, in case you'd rather take it forward yourself — but #5363 supersedes it as far as I can tell.

PR #4880, opened 2021-09-01
chore: replace drain with poll for queue for GameThread

pollend:
This removes an extra array allocation when we process the processing the pending runnable


DarkWeird:
this happens only one time per tick. 
it's not scary

what about adding process during iteration ( multi-thread behavior, like process adds another processes)?

pollend:
lot of this logic is bound to the main game thread @DarkWeird :?

DarkWeird:
> lot of this logic is bound to the main game thread @DarkWeird :?

It should be bound to main game thread. Otherwise you can not use gamethread class :)

I afraid this things:
1. Concurrent modification(don't should be - concurrent queue)
2. Infinity loop (process adding recursion)

PR #4881, opened 2021-09-02
perf: minor adjusments to chunkprocessing pipeline

pollend:
Some minor adjustments and cleanup to ChunkProcessingPipeline. I tried to clean up some of the intent of these methods a bit and introduced some minor performance adjustments. 

- replace set with List for createMulti. None of these collections are used like sets and the set that is used is overly defensive.
- made LightMerger static and migrated instance of Vector3 to Vector3ic
- simplify processChunkTasks the lambdas end up complicating how the function is evaluated. this should be a bit more clearer. i.e only process a chunk if all the required chunks are valid. lets me exit a lot earlier in the evaluation.


tolziplohu (approved):
This looks good to me, I like not creating `LightMerger` instances all the time and the rest of the changes are pretty straightforward cleanup and improvements. You might want to run "optimize imports" though, there are a lot of checkstyle warnings about unused imports.

tolziplohu (approved):

PR #4884, opened 2021-09-03
fix(autoconfig): fix back-animation to empty screen

DarkWeird:
### Contains

Fixes  back-animation to empty screen for AutoConfigScreen

### How to test

1. Tera Main menu
2. enable animated menus
3. go to settings -> Autoconfig
4. click `back`
5. you should returns back


pollend:
that at least makes the menu work correctly.

pollend (approved):

PR #4885, opened 2021-09-04
chore: minor fixes and introduce some test cases for VertexResource

pollend:
some basic test cases for VertexResource.


engine-tests/src/test/java/org/terasology/engine/rendering/assets/mesh/VertexGLAttributeTest.java
```diff
@@ -0,0 +1,259 @@
...
+    public void testFloatBinding() {
+        VertexResourceBuilder builder = new VertexResourceBuilder();
+        VertexFloatAttributeBinding a1 = builder.add(0, GLAttributes.FLOAT_1_VERTEX_ATTRIBUTE);
+        VertexResource resource = builder.build();
+
+        a1.put(10.0f);
+        a1.put(15.0f);
+        a1.put(15.5f);
+        a1.put(2.0f);
+        a1.put(-100.0f);
```
keturn, line 36: Fun ways to test floating-point data structures include: `-0` (negative zero), `NaN`, [`POSITIVE_INFINITY`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Float.html#POSITIVE_INFINITY).

(This may or may not actually be relevant for this application.)

engine-tests/src/test/java/org/terasology/engine/rendering/assets/mesh/VertexGLAttributeTest.java
```diff
@@ -0,0 +1,259 @@
...
+        a1.put(300);
+        a1.put(100);
+
+        assertEquals(4, a1.getPosition());
+        resource.writeBuffer(buffer -> {
+            assertEquals(4 * Byte.BYTES, buffer.limit());
+
+            assertEquals(10, Byte.toUnsignedInt(buffer.get(Byte.BYTES * 0)));
+            assertEquals(150, Byte.toUnsignedInt(buffer.get(Byte.BYTES * 1)));
+            assertEquals(255, Byte.toUnsignedInt(buffer.get(Byte.BYTES * 2)));
```
keturn, line 91: Ah, this `255` case is the one that goes with that `clamp()` you added. I suggest pointing that out with a comment, it's pretty easy to miss when scanning through these lines and numbers that are mostly all the same.

Should this test a negative value too?

keturn (approved):
👍 for tests and docs. The tests look straightforward, making it clear what they're going for.

I notice a lot of tests creating a builder, calling exactly one method on it, and then `build()`, which suggests we might want to add a more convenient API for that later.

I'm not merging this immediately, so you have a chance to decide if there's anything in these comments you want to add, but feel free to merge when ready.

PR #4886, opened 2021-09-04
chore: add test cases for IndexResource and address some problems found through test cases

pollend:
introduce test cases for IndexResource and fix some checkstyle mistakes in the code related to IndexResource. I also fixed some inconsistencies in the API for IndexResouce where the number of indices can be out of sync with the size of the buffer.


keturn (approved):
LGTM.

Tangential question: Why is it named `inSize`?

PR #4887, opened 2021-09-04
chore: cleanup extra qualification for org.joml from termath migration

pollend:
here is some minor cleanup that finishes up the migration from termath. There are some extra qualified org.joml in the code that should help a little bit with cleaning up the existing code.


keturn (approved):
🚮

PR #4888, opened 2021-09-04
chore: remove deprecated Chunk#getPosition(Vector3i)

pollend:
we've had this method depreacted for a while in Chunk, I made sure to have the Vector3i that's allocated in Chunk fixed in place so once allocated we won't be able to modify the position of a chunk.


keturn:
Why are so many of the calls to `chunk.getPosition()` wrapped in `new Vector3i`?

Maybe that's related to your question about how defensive the code needs to be, but I'm not quite connecting the dots.

engine-tests/src/test/java/org/terasology/fixtures/TestStorageManager.java
```diff
@@ -33,7 +32,7 @@ public TestStorageManager(Map<Vector3ic, ChunkStore> chunkStores) {
     }
 
     public void add(Chunk chunk) {
-        chunkStores.put(chunk.getPosition(new Vector3i()), new TestChunkStore(chunk));
+        chunkStores.put(chunk.getPosition(), new TestChunkStore(chunk));
```
skaldarnar, line 35: Hm, we often recommend to create a copy when storing vectors in a map, but I guess for chunk position it is fine to assume the position to be constant and stable for a chunk, not changing ever. :thinking:

engine/src/main/java/org/terasology/engine/monitoring/chunk/ChunkMonitor.java
```diff
@@ -61,17 +61,17 @@ public static void fireChunkCreated(Chunk chunk) {
 
     public static void fireChunkDisposed(Chunk chunk) {
         Preconditions.checkNotNull(chunk, "The parameter 'chunk' must not be null");
-        post(new ChunkMonitorEvent.Disposed(chunk.getPosition(new Vector3i())));
+        post(new ChunkMonitorEvent.Disposed(new Vector3i(chunk.getPosition())));
```
skaldarnar, line 64: Is this conversion an artifact of `ChunkMonitorEvent.Disposed` requiring a mutable `Vector3i` instead of a `Vector3ic` as returned by `Chunk`? If so, is there any need for it to be a mutable vector, or could we adjust the interface to just work with `Vector3ic`?

skaldarnar (commented):
I think we should add to the description in https://github.com/MovingBlocks/Terasology/blob/f85be6d1a036448410d6d7096e6b9305dd4e23f9/engine/src/main/java/org/terasology/engine/world/chunks/Chunk.java#L25-L28 that the resulting vector can be considered stable/constant for a given chunk (at least we are using it that way).

Also, I'm wondering whether some of the `new Vector3i(...)` allocations are just because some other API demands to pass in a mutable `Vector3i` instead of a immutable `Vecor3ic`.
Where possible, we should fix those APIs to work with the immutable type instead. We can likely do that in smaller batches and thereby make better use of `Chunk#getPosition()`.

pollend:
> Why are so many of the calls to `chunk.getPosition()` wrapped in `new Vector3i`?
> 
> Maybe that's related to your question about how defensive the code needs to be, but I'm not quite connecting the dots.

those shouldn't be wrapped in a new

engine/src/main/java/org/terasology/engine/entitySystem/sectors/SectorUtil.java
```diff
@@ -46,7 +46,7 @@ public static SectorRegionComponent createSectorRegionComponent(Collection<Vecto
      */
     public static void addChunksToRegionComponent(SectorRegionComponent regionComponent, Collection<Chunk> chunks) {
         regionComponent.chunks.addAll(chunks.stream()
-                .map(k -> k.getPosition(new Vector3i()))
+                .map(k -> new Vector3i(k.getPosition()))
```
keturn, line 49: Do we need this `new`?

keturn (changes requested):
This does look much better without the extra `new Vector`s. :+1: Just once left I wasn't sure about.

I noticed a signature change on fireChunkTessellated, but it looks like there's only one call site for that and it did get updated, so that part seems good too.

engine/src/main/java/org/terasology/engine/entitySystem/sectors/SectorUtil.java
pollend: its in SectorRegionComponent so you don't know if those might change.

keturn: hmm, I _think_ it would be okay if regionComponent.chunks changed to Vector3i**c**, but I'm okay with saying that doesn't have to be in the scope of this PR. Seems like there's other doubts about SectorRegion's future.

keturn (approved):

keturn:
Currently being held by CI for some unused code checks. I recommend turning on IntelliJ's “optimize imports on the fly“ setting—or, if you don't like it doing stuff on the fly, at least the “optimize imports“ option in the “before commit“ section.

jdrueckert (approved):

PR #4891, opened 2021-09-05
build(IntelliJ): turn on checkstyle.scan-before-checkin

keturn:
Small tweaks to how CheckStyle is setup in gradle and IntelliJ.

The most impactful one is that `scan-before-checkin`.

I was tempted to see if we could get rid of the `samedir` property, but that princess is in another castle.


keturn:
```
* What went wrong:
Execution failed for task ':engine:checkstyleJmh'.
> Unable to store input properties for task ':engine:checkstyleJmh'. Property 'configProperties' with value '{samedir=extension 'checkstyle' property 'configDirectory'}' cannot be serialized.
```

keturn:
fixed. There was some groovy/gradle typing hijinks going on that threw a wrench in the first attempt.

jdrueckert (approved):

PR #4893, opened 2021-09-06
refactor: split chunkMesh to ChunkMeshImpl

pollend:
This introduces an interface for ChunkMesh. should let us mock in the interface when we want to test some rendering code and other graphics API's down the line. still not sure about some of these methods that I'm keeping in the interface. I want to move some of this metric information somewhere else but that is something to consider later. 

PR: https://github.com/MovingBlocks/Terasology/pull/4786


engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
```diff
@@ -19,285 +18,82 @@
...
-    }
+    int render(ChunkMesh.RenderPhase type);
 
-    public boolean hasVertexElements() {
-        return vertexElements != null;
-    }
 
     /**
-     * Generates the VBOs from the pre calculated arrays.
+     * this is only used for
```
keturn, line 58: for what?

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMeshImpl.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+    }
+
+    @Override
+    public DisposableHook disposalHook() {
+        WeakReference<ChunkMeshImpl> chunkMeshWeakReference = new WeakReference<>(this);
+        return new DisposableHook() {
+            private final int[] hookVertexBuffers = vertexBuffers;
+            private final int[] hookIndexBuffers = idxBuffers;
+            private final int[] hookVertexArrayObjects = vaoCount;
```
keturn, line 198: Weak reference to `this`, but direct references to private properties of this? Suspicious.

keturn (commented):
A few quick notes so far. I hope to give it another look later.

Coming from Python, the practice of making an Interface/Class separation purely for the purpose of internal tests always feels weird. We'd just 🦆-type it in Python, right? So I end up second-guessing this sort of thing, even though I often do prefer it to Mockito.

I notice you did add a `HeadlessChunkMesh` implementation of the interface. Is that something a headless server will use? Or does the `HeadlessGraphics` subsystem already skip over all the engine.rendering.`*` stuff?

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMeshImpl.java
pollend: that's because its getting wrapped into a lambda

pollend:
> A few quick notes so far. I hope to give it another look later.
> 
> Coming from Python, the practice of making an Interface/Class separation purely for the purpose of internal tests always feels weird. We'd just duck-type it in Python, right? So I end up second-guessing this sort of thing, even though I often do prefer it to Mockito.
> 
> I notice you did add a `HeadlessChunkMesh` implementation of the interface. Is that something a headless server will use? Or does the `HeadlessGraphics` subsystem already skip over all the engine.rendering.`*` stuff?

I don't think we need the headlesssChunkMesh will just creating a mock object from the interface.

DarkWeird:
I prefer to implement test object over interface than use mockito:
1. Mockito add duration around 200-500ms to every test with using mockito.
2. I think that if you should use mockito - you have architecture problem.
3. Copy pasted mockito code from test to test(also feels noisy)

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
keturn: @pollend, please finish your sentence!

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMeshImpl.java
keturn: What is because what is getting wrapped in what?

keturn (changes requested):

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMeshImpl.java
pollend: this would be in the context of DiposableHook when the context is in the class that is creating the hook.

keturn: As I read this, it looks like ChunkMesh has both a `dispose` method that's visible to the `engine.rendering.primitives` package, as well as this public `disposalHook` method.

And ChunkMeshImpl's implementation of `dispose` uses `disposalHook` (creating a new instance and then immediately invoking it), but **also** does all the same operations itself?

This implementation of `DisposableHook` is written so it _does_ hold references to ChunkMeshImpl's private array fields—this makes sense if there's something holding on to these ChunkMeshImpl.DisposableHooks beyond the time the ChunkMeshImpl objects themselves have been garbage collected, and indeed that's what the ChunkMesh interface's javadoc says, but I'm clearly missing something. What is building and saving this list of DisposableHooks and why is it important for it to do that instead of using ChunkMesh.dispose?

And can we get rid of the code duplication between ChunkMeshImpl.dispose and ChunkMeshImpl.disposalHook().dispose(), or are my sleepy eyes reading it wrong and it's not actually a duplication?

pollend: It used for entities that are associated with a chunkMesh but I can't see it being used anywhere.

pollend:
@keturn we probably want to keep the chunk mesh hook. it will be used for this: https://github.com/Terasology/FallingBlocks/pull/9

keturn:
> we probably want to keep the chunk mesh hook. it will be used for this: [Terasology/FallingBlocks#9](https://github.com/Terasology/FallingBlocks/pull/9)

How is that related to ChunkMesh.disposalHook? The FallingBlocks PR was a bit too big to look at in github web, but I checked it out and searched for "disposal" (or "dispose") in that branch and came up empty.

If we do have a use case for disposalHook, that's fine, but my question from [earlier](https://github.com/MovingBlocks/Terasology/pull/4893#discussion_r716085740) still stands: Can we get rid of the code duplication between ChunkMeshImpl.dispose and ChunkMeshImpl.disposalHook().dispose(), or are my sleepy eyes reading it wrong and it's not actually a duplication?

pollend:
> > we probably want to keep the chunk mesh hook. it will be used for this: [Terasology/FallingBlocks#9](https://github.com/Terasology/FallingBlocks/pull/9)
> 
> How is that related to ChunkMesh.disposalHook? The FallingBlocks PR was a bit too big to look at in github web, but I checked it out and searched for "disposal" (or "dispose") in that branch and came up empty.
> 
> If we do have a use case for disposalHook, that's fine, but my question from [earlier](https://github.com/MovingBlocks/Terasology/pull/4893#discussion_r716085740) still stands: Can we get rid of the code duplication between ChunkMeshImpl.dispose and ChunkMeshImpl.disposalHook().dispose(), or are my sleepy eyes reading it wrong and it's not actually a duplication?

its for the chunkMeshComponent.

keturn:
ah, the pointer to ChunkMeshRenderer system helped, thank you.

This system is already defining an `onDestroyMesh` lifecycler handler for the component. Why do we need to do this reference-tracking stuff instead of doing disposal there?

If we can't use normal component lifecycle events, could we at least make this a WeakReference instead of a PhantomReference, so we can call the normal dispose method on the reference?

(Please include the answers to both of the above in ChunkMeshRenderer's javadoc.)

If we do have a use case for disposalHook, that's fine, but my question from [earlier](https://github.com/MovingBlocks/Terasology/pull/4893#discussion_r716085740) still stands: Can we get rid of the code duplication between ChunkMeshImpl.dispose and ChunkMeshImpl.disposalHook().dispose(), or are my sleepy eyes reading it wrong and it's not actually a duplication?

pollend:
> ah, the pointer to ChunkMeshRenderer system helped, thank you.
> 
> This system is already defining an `onDestroyMesh` lifecycle handler for the component. Why do we need to do this reference-tracking stuff instead of doing disposal there?
> 
> If we can't use normal component lifecycle events, could we at least make this a WeakReference instead of a PhantomReference, so we can call the normal dispose method on the reference?
> 
> (Please include the answers to both of the above in ChunkMeshRenderer's javadoc.)
> 
> If we do have a use case for disposalHook, that's fine, but my question from [earlier](https://github.com/MovingBlocks/Terasology/pull/4893#discussion_r716085740) still stands: Can we get rid of the code duplication between ChunkMeshImpl.dispose and ChunkMeshImpl.disposalHook().dispose(), or are my sleepy eyes reading it wrong and it's not actually a duplication?

better to just keep it simple. I just moved it to use the normal events to drive disposing chunks.

keturn:
> I just moved it to use the normal events to drive disposing chunks.

That does look a lot simpler for the Renderer as well as the dispose method!

CheckStyle has a few notes for you that you may want to address before merging this. I recommend turning on IntelliJ's “optimize imports on the fly“ setting—or, if you don't like it doing stuff on the fly, at least the “optimize imports“ option in the “before commit“ section.

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
```diff
@@ -47,257 +76,18 @@ public int getIndex() {
...
-    }
-
-    public VertexElements getVertexElements(RenderType renderType) {
-        return vertexElements[renderType.ordinal()];
-    }
-
-    public boolean hasVertexElements() {
-        return vertexElements != null;
+    interface DisposableHook {
+        void dispose();
```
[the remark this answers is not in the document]
keturn: removed

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMeshImpl.java
keturn: addressed by f392965a7c931867e11eaa8a94f5a5dcb073afa2

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMesh.java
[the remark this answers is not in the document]
pollend: I removed it :P

engine/src/main/java/org/terasology/engine/rendering/primitives/ChunkMeshImpl.java
```diff
@@ -0,0 +1,235 @@
...
+    public boolean updateMesh() {
+
+        // IMPORTANT: A mesh can only be generated once.
+        if (vertexElements == null || disposed) {
+            return false;
+        }
+
+        // Make sure that if it has already been generated, the previous buffers are freed
+        dispose();
+        disposed = false;
```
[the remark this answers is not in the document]
pollend: the documentation is wrong.

keturn (approved):
Okay, I think that's everything! And checks pass and I can still see chunks in game. Although I did run out of RAM pretty quickly this run.

How would I know if things weren't getting disposed of as they should?

